PR #1193, opened 2020-05-10
Fix getTracerLower behavior

### LINKED ISSUE #1190: TraceID.getLowerLong() returns the less random part of the trace
**Describe the bug**
`TraceId.getLowerLong()` returns the leftmost part of the trace id. This method is expected to return the part of the trace which is more random. According to the [W3C Trace Context Standard](https://www.w3.org/TR/trace-context/#interoperating-with-existing-systems-which-use-shorter-identifiers), this corresponds to the rightmost part of the trace. 

**Steps to reproduce**

```java
TraceId traceId = new TraceId(0x12345L, 0x67890L);
assertThat(traceId.getLowerLong()).isNotEqualTo(0x12345L);
```

thisthat:
This resolves #1190 
Since the method is marked as internal, I also followed the suggestion proposed in the last SIG call and rename it to `getTraceRandomPart()`

### DIFF
--- a/api/src/main/java/io/opentelemetry/trace/TraceId.java
+++ b/api/src/main/java/io/opentelemetry/trace/TraceId.java
@@ -210,8 +210,8 @@ public final class TraceId implements Comparable<TraceId> {
    *
    * @return the lower 8 bytes of the trace-id as a long value, assuming little-endian order.
    */
-  public long getLowerLong() {
-    return (idHi < 0) ? -idHi : idHi;
+  public long getTraceRandomPart() {
+    return (idLo < 0) ? -idLo : idLo;
   }
 
   @Override
--- a/api/src/main/java/io/opentelemetry/trace/TraceId.java
+++ b/api/src/main/java/io/opentelemetry/trace/TraceId.java
@@ -203,15 +203,15 @@ public final class TraceId implements Comparable<TraceId> {
   }
 
   /**
-   * Returns the lower 8 bytes of the trace-id as a long value, assuming little-endian order. This
-   * is used in ProbabilitySampler.
+   * Returns the rightmost 8 bytes of the trace-id as a long value. This is used in
+   * ProbabilitySampler.
    *
    * <p>This method is marked as internal and subject to change.
    *
-   * @return the lower 8 bytes of the trace-id as a long value, assuming little-endian order.
+   * @return the rightmost 8 bytes of the trace-id as a long value.
    */
   public long getTraceRandomPart() {
-    return (idLo < 0) ? -idLo : idLo;
+    return idLo;
   }
 
   @Override
--- a/api/src/test/java/io/opentelemetry/trace/TraceIdTest.java
+++ b/api/src/test/java/io/opentelemetry/trace/TraceIdTest.java
@@ -39,7 +39,7 @@ public class TraceIdTest {
 
   @Test
   public void invalidTraceId() {
-    assertThat(TraceId.getInvalid().getLowerLong()).isEqualTo(0);
+    assertThat(TraceId.getInvalid().getTraceRandomPart()).isEqualTo(0);
   }
 
   @Test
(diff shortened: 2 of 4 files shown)


sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
```diff
@@ -210,7 +210,7 @@ public final Decision shouldSample(
       // while allowing for a (very) small chance of *not* sampling if the id == Long.MAX_VALUE.
       // This is considered a reasonable tradeoff for the simplicity/performance requirements (this
       // code is executed in-line for every Span creation).
-      return Math.abs(traceId.getLowerLong()) < getIdUpperBound()
+      return Math.abs(traceId.getTraceRandomPart()) < getIdUpperBound()
```
bogdandrutu, line 213: We need to make sure every library applies the same logic and extract correctly the random part. This is important because if 2 libraries are configured with the same probability we need both to sample or not sample a given trace id.

api/src/test/java/io/opentelemetry/trace/TraceIdTest.java
```diff
@@ -50,9 +50,13 @@ public void isValid() {
...
-  public void getLowerLong() {
-    assertThat(first.getLowerLong()).isEqualTo(0);
-    assertThat(second.getLowerLong()).isEqualTo(-0xFF00000000000000L);
+  public void testCorrectGetRandomTracePart() {
+    byte[] id = {
+      0x1, 0x2, 0x3, 0x4, 0x5, 0x6, 0x7, 0x8,
+      0x9, 0xA, 0xB, 0xC, 0xD, 0xE, 0xF, 0x0
+    };
+    TraceId traceid = TraceId.fromBytes(id, 0);
+    assertThat(traceid.getTraceRandomPart()).isEqualTo(0x90A0B0C0D0E0F00L);
```
bogdandrutu, line 59: It took me few minutes to check that this is the correct expectation, I would suggest to put a `0` in front of the `9` so having something like `0x090A0B0C0D0E0F00L` so I have 16 hex characters and also a easier way to map every 2 hex characters from here to the bytes.

sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
thisthat: I was looking at other languages how they implemented this, and it's all over the place. Looking at the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#built-in-samplers), the only indication is to use part of the trace for the decision.

api/src/main/java/io/opentelemetry/trace/TraceId.java
```diff
@@ -210,8 +210,8 @@ public boolean equals(@Nullable Object obj) {
    *
    * @return the lower 8 bytes of the trace-id as a long value, assuming little-endian order.
    */
-  public long getLowerLong() {
-    return (idHi < 0) ? -idHi : idHi;
+  public long getTraceRandomPart() {
+    return (idLo < 0) ? -idLo : idLo;
```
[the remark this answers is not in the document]
thisthat: I keep it as it was previously. This basically guarantees the first bit of `idLo` to be zero. However, I don't think this is the intended behavior.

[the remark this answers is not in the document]
bogdandrutu: I saw the problem in the probability sampler we compare this with a positive number so any negative number will pass that condition, so we don't have the guarantees of the rate. Essentially we need to change a bit the probability sampler if we return negative values as well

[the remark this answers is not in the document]
thisthat: You're right! I'll take a look at it!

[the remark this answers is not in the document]
thisthat: The `Probability` sampler already takes the abs value of the result of `getTraceRandomPart()`. Therefore, we don't need to change its logic

api/src/test/java/io/opentelemetry/trace/TraceIdTest.java
```diff
@@ -50,9 +50,36 @@ public void isValid() {
...
 
   @Test
-  public void getLowerLong() {
-    assertThat(first.getLowerLong()).isEqualTo(0);
-    assertThat(second.getLowerLong()).isEqualTo(-0xFF00000000000000L);
+  public void testGetRandomTracePart() {
+    byte[] id = {
+      0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B, 0x0C, 0x0D, 0x0E, 0x0F, 0x00
+    };
+    TraceId traceid = TraceId.fromBytes(id, 0);
```
carlosalberto, line 57: We also have the `TraceId` ctor taking two longs. Could we add a test for that one as well?

carlosalberto:
Overall LGTM.

api/src/test/java/io/opentelemetry/trace/TraceIdTest.java
carlosalberto: Never mind, saw that Bogdan doesn't want us to use that ctor. Resolving.

PR #1201, opened 2020-05-12
Add CounterUpDown implementation in the SDK

bogdandrutu:
This PR will be split into smaller PRs:
* #1202 
* #1200 
* #1208 
* #1210

PR #1202, opened 2020-05-12
Cleanup names in metrics and ensure consistency with specs

bogdandrutu:
Split from https://github.com/open-telemetry/opentelemetry-java/pull/1201

### DIFF
--- a/api/src/main/java/io/opentelemetry/metrics/SyncInstrument.java
+++ b/api/src/main/java/io/opentelemetry/metrics/SyncInstrument.java
@@ -16,7 +16,7 @@
 
 package io.opentelemetry.metrics;
 
-import io.opentelemetry.metrics.InstrumentWithBinding.BoundInstrument;
+import io.opentelemetry.metrics.SyncInstrument.BoundInstrument;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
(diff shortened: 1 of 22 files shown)


api/src/main/java/io/opentelemetry/metrics/SyncInstrument.java
```diff
@@ -26,7 +26,7 @@
  * @since 0.3.0
  */
 @ThreadSafe
-public interface InstrumentWithBinding<B extends BoundInstrument> extends Instrument {
+public interface SyncInstrument<B extends BoundInstrument> extends Instrument {
```
jkwatson, line 29: Even though it's a bit longer, I prefer `SynchronousInstrument`. Also, it would be good to explain in some javadoc what "synchronous" means in this context.

bogdandrutu: Done.

api/src/main/java/io/opentelemetry/metrics/SynchronousInstrument.java
```diff
@@ -16,17 +16,21 @@
...
 package io.opentelemetry.metrics;
 
-import io.opentelemetry.metrics.InstrumentWithBinding.BoundInstrument;
+import io.opentelemetry.metrics.SynchronousInstrument.BoundInstrument;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
- * Base interface for all metrics with bounds defined in this package.
+ * SynchronousInstrument is an interface that defines a type of instruments that are used to report
+ * measurements synchronously, user reports individual measurements when they occur.
```
jkwatson, line 24:
```suggestion
 * measurements synchronously. That is, when the user reports individual measurements as they occur.
```

PR #1203, opened 2020-05-12
Fix remove attributes from span after termination

### LINKED ISSUE #1195: It is possible to remove an attribute from a span that has ended
**Describe the bug**
It is possible to remove an attribute from a span that has ended

**Additional context**
This was introduced in #807 and I believe it was a mistake.

thisthat:
Resolves #1195

### DIFF
--- a/sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
@@ -334,6 +334,10 @@ final class RecordEventsReadableSpan implements ReadableSpan, Span {
     Preconditions.checkNotNull(key, "key");
     if (value == null || (value.getType().equals(STRING) && value.getStringValue() == null)) {
       synchronized (lock) {
+        if (hasEnded) {
+          logger.log(Level.FINE, "Calling setAttribute() on an ended Span.");
+          return;
+        }
         attributes.remove(key);
       }
       return;
--- a/sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
@@ -332,19 +332,13 @@ final class RecordEventsReadableSpan implements ReadableSpan, Span {
   @Override
   public void setAttribute(String key, AttributeValue value) {
     Preconditions.checkNotNull(key, "key");
-    if (value == null || (value.getType().equals(STRING) && value.getStringValue() == null)) {
-      synchronized (lock) {
-        if (hasEnded) {
-          logger.log(Level.FINE, "Calling setAttribute() on an ended Span.");
-          return;
-        }
-        attributes.remove(key);
-      }
-      return;
-    }
     synchronized (lock) {
       if (hasEnded) {
         logger.log(Level.FINE, "Calling setAttribute() on an ended Span.");
+        return;
+      }
+      if (value == null || (value.getType().equals(STRING) && value.getStringValue() == null)) {
+        attributes.remove(key);
         return;
       }
       totalAttributeCount++;
--- a/sdk/src/test/java/io/opentelemetry/sdk/trace/SpanBuilderSdkTest.java
+++ b/sdk/src/test/java/io/opentelemetry/sdk/trace/SpanBuilderSdkTest.java
@@ -272,8 +272,6 @@ public class SpanBuilderSdkTest {
   public void setAttribute_nullAttributeValue_afterEnd() throws Exception {
     Span.Builder spanBuilder = tracerSdk.spanBuilder(SPAN_NAME);
     spanBuilder.setAttribute("emptyString", "");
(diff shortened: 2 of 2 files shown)


sdk/src/test/java/io/opentelemetry/sdk/trace/SpanBuilderSdkTest.java
```diff
@@ -172,6 +172,44 @@ public void setAttribute() {
     }
   }
 
+  @Test
+  public void setAttribute_afterEnd() {
+    Span.Builder spanBuilder = tracerSdk.spanBuilder(SPAN_NAME);
```
carlosalberto, line 177: Do you really need to add 5 attributes for this test? You could have a single item and assert that only that item remains afterwards, and make the test simpler/smaller.

carlosalberto: Oh, never mind. You are exercising all the attribute types.

sdk/src/test/java/io/opentelemetry/sdk/trace/SpanBuilderSdkTest.java
```diff
@@ -230,6 +268,37 @@ public void setAttribute_nullAttributeValue() throws Exception {
     assertThat(span.toSpanData().getAttributes()).isEmpty();
   }
 
+  @Test
+  public void setAttribute_nullAttributeValue_afterEnd() throws Exception {
+    Span.Builder spanBuilder = tracerSdk.spanBuilder(SPAN_NAME);
+    spanBuilder.setAttribute("emptyString", "");
+    spanBuilder.setAttribute("nullString", (AttributeValue) null);
```
carlosalberto, line 275: This will not be added anyway. Maybe we can keep only the values that will be really added? This makes the test more concise.

carlosalberto:
Overall LGTM.

sdk/src/test/java/io/opentelemetry/sdk/trace/SpanBuilderSdkTest.java
thisthat: You're right! This is pointless since it is covered by another test!

sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -334,6 +334,10 @@ public void setAttribute(String key, AttributeValue value) {
     Preconditions.checkNotNull(key, "key");
     if (value == null || (value.getType().equals(STRING) && value.getStringValue() == null)) {
       synchronized (lock) {
+        if (hasEnded) {
```
jkwatson, line 337: why don't we just do this check once up front, rather than repeating the code here and below?

thisthat: I didn't want to introduce another `synchronized` block. But I guess you're right, it is better to do only once upfront!

jkwatson: I'd rather have one slightly larger synchronized block, rather than force jumping in and out of the synchronization twice in rapid succession. As long as you're doing very cheap operations, just putting all the logic in one synchronization context is going to be better.

bogdandrutu: But this one does not jump between synchronized.

jkwatson: I agree. I was just thinking forward to moving the ended check up-front, rather then duplicating that check.

bogdandrutu: Splitting the lock is incorrect. You need to make the check in the same lock sync to ensure that the condition does not change between the two synchronized sections

bogdandrutu: To clarify @jkwatson your proposal is correct but your concern should not be a preference should be a requirement because otherwise the code is incorrect :)

thisthat: You are right @bogdandrutu, I was thinking only to the case where the span already ended.

jkwatson: @bogdandrutu also true! I hadn't actually fully thought through the implementation, but you are absolutely correct.

bogdandrutu: Personally I am happy with the current code, but not opposed to change it based on @jkwatson suggestion. Up to you @thisthat :)

thisthat: I think having everything in a single synchronized block makes it easier to fully understand the method. 7cab1ee addressed this :wink:

Issue #1198, opened 2020-05-11
Span number of attributes doesn't include those from the initial builder


jkwatson:
**Describe the bug**
When we start a span, we don't initialize the number of attributes to the ones set in the builder.

**Steps to reproduce**
```
    Span.Builder spanBuilder = tracerSdk.spanBuilder(SPAN_NAME);
    spanBuilder.setAttribute("attr1", "value1");

    // this should make a copy of the attributes in the builder, and it does, but doesn't keep the count
    RecordEventsReadableSpan span = (RecordEventsReadableSpan) spanBuilder.startSpan();

    //this is correct (it has the attribute from the builder)
    assertThat(span.toSpanData().getAttributes().size()).isEqualTo(1);
    //this is not...it returns 0
    assertThat(span.toSpanData().getTotalAttributeCount()).isEqualTo(1);
```

PR #1207, opened 2020-05-12
Ensure the total number of attributes is counted correctly

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-java/issues/1198


sdk/src/main/java/io/opentelemetry/sdk/trace/AttributesMap.java
```diff
@@ -0,0 +1,80 @@
...
+    throw new RuntimeException("Do not call methods on the map");
+  }
+
+  public AttributeValue replace(String key, AttributeValue value) {
+    throw new RuntimeException("Do not call methods on the map");
+  }
+
+  public boolean replace(String key, AttributeValue oldValue, AttributeValue newValue) {
+    throw new RuntimeException("Do not call methods on the map");
+  }
```
bogdandrutu, line 75: Added as public to override the newly added methods in java8, so when we do the switch and start using them we remember to fix them and count the number of attributes added.

sdk/src/main/java/io/opentelemetry/sdk/trace/AttributesMap.java
```diff
@@ -0,0 +1,86 @@
...
+
+  @Nullable
+  @Override
+  public AttributeValue remove(Object key) {
+    return super.remove(key);
+  }
+
+  // Added as public to override the newly added methods in java8, so when we do the switch and
+  // start using them we remember to fix them and count the number of attributes added.
```
jkwatson, line 66: We should revisit this. Extending HashMap is a pretty bad antipattern in the Java world. I'd rather eat an allocation to wrap a Map and only provide the functionality that we want to expose, rather than having to deal with any arbitrary calls to our Map extension. But, that's beyond the scope of this PR.

bogdandrutu: I would file an issue and we can address later.

bogdandrutu: https://github.com/open-telemetry/opentelemetry-java/issues/1209

jkwatson: yeah, if it were public, I would be much much more adamant about it being fixed ASAP.  :)

PR #1208, opened 2020-05-12
Remove monotonic property from counters

bogdandrutu:
Split from https://github.com/open-telemetry/opentelemetry-java/pull/1201


api/src/main/java/io/opentelemetry/metrics/Counter.java
```diff
@@ -40,17 +40,6 @@
     @Override
     Builder setConstantLabels(Map<String, String> constantLabels);
 
-    /**
```
jkwatson, line 43: Is this Counter.Builder interface even used any more?

jkwatson: In fact, is the `Counter` interface even used?

bogdandrutu: Sorry, my bad, I forgot to delete it :)

Issue #1191, opened 2020-05-08
Improve performance of trace/metrics export pipeline by reducing number of allocs


bogdandrutu:
Currently in the trace and metrics exporters we use SpanData and MetricData, here are some benefits:
1. Both objects are immutable which guarantees that multiple exporters can consume them safely.
2. Another benefit is that for cases where users call into the Span api with a custom `Event` to allow lazy formatting (of attributes or event name) and hold external memory (reference to other objects) we do the formatting when converting to SpanData and remove references to these objects.

I think we should look into providing an alternative implementation that can achieve the same guarantees without having to reallocate all the recorded objects.

One proposal is to have something like "FreezeSpan" on the `ReadableSpanImpl` that will transform all the internal lists/maps into immutable objects as well as will force initialize objects (remove references to the external memory if any, and format names/attributes).

jkwatson:
@bogdandrutu Are you thinking that the "freezing" of the span would happen when the span is ended, or do you think this should live at a different point in the lifecycle?

bogdandrutu:
I think we should do it later and only if needed.

jkwatson:
> I think we should do it later and only if needed.

Are you thinking that it would happen on the first call to "toSpanData" (or the equivalent? Or, a new method on the RERS class that would do the freeze? In that case, who would call the freeze method? SpanProcessor?

bogdandrutu:
Don't know yet. I have not thought all the way to this level of details :).

jkwatson:
Cool. I'm pretty close to having something ready to show.

bogdandrutu:
@jkwatson more ideas:

* Make SpanData.Event interface and have a SpanDataImpl.EventImpl implementation for testing.
* Then make TimedEvent in the SDK implement SpanData.Event, so in the getImmutableTimedEvents we don't create new events for events that are already immutable. Currently we copy all the events.

jkwatson:
@bogdandrutu can we close this now? Or do you have more ideas on how to improve this?

jkwatson:
I think we've got the majority of the low-hanging fruit done for this. Let's open up new issues for any new optimizations that we want to make.

PR #1216, opened 2020-05-13
Option for reducing allocations in the span pipeline

jkwatson:
This is a Draft proposal for one way to approach solving #1191 

@bogdandrutu let me know what you think.


sdk/src/main/java/io/opentelemetry/sdk/trace/data/SpanData.java
```diff
@@ -16,63 +16,47 @@
...
-/**
- * Immutable representation of all data collected by the {@link io.opentelemetry.trace.Span} class.
- *
- * @since 0.1.0
- */
-@Immutable
-@AutoValue
-public abstract class SpanData {
 
+public interface SpanData {
```
bogdandrutu, line 32: Leave the immutable annotation?

jkwatson: yep. will add it back.

jkwatson:
One option would be to return a wrapper of the RERS that implemented the SpanData interface. That's just one allocation (plus the necessary resolving of the links/events). That would be very easy to implement (and memoize in the RERS if we want to only generate it one time). WDYT?

bogdandrutu:
Something about these lines. I think we should start with splitting the SD then deciding on the approach to do with RERS :). But I think we will go with that proposal unless we can think of a better option.

jkwatson:
Ok, I'll put in a PR that just creates the SpanData interface along with the (badly named) SpanDataImpl and keeps the same implementation, then we can move from there.

jkwatson:
closing this, as we're proceeding with this in separate PRs

PR #1217, opened 2020-05-14
Unify config api

### LINKED ISSUE #996: For SpanProcessor/SpanExporter/MetricProducer/MetricExporter use Config
This is a proposal to use a Config object instead of Builders for all the implementations of these classes. This will allow us to have a "getFromEnv" on all these Configs, probably a "getDefault" as well.

This will be easier to initialize compared with Builders in auto-instrumentation as well as when dealing with initializing a pipeline like SpanProcessor -> SpanExporter.

/cc @trask @tylerbenson - what do you think about this from the auto-instrumentation perspective.

thisthat:
This resolves #996

### DIFF
--- a/sdk/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpansProcessorTest.java
+++ b/sdk/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpansProcessorTest.java
@@ -126,14 +126,14 @@ public class BatchSpansProcessorTest {
   @Test
   public void exportMoreSpansThanTheBufferSize() {
     WaitingSpanExporter waitingSpanExporter = new WaitingSpanExporter(6);
-    BatchSpansProcessor bsp =
+    BatchSpansProcessor batchSpansProcessor =
         BatchSpansProcessor.newBuilder(waitingSpanExporter)
             .setMaxQueueSize(6)
             .setMaxExportBatchSize(2)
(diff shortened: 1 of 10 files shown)


sdk/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpansProcessorTest.java
```diff
@@ -329,14 +126,14 @@ public void exportDifferentSampledSpans() {
   @Test
   public void exportMoreSpansThanTheBufferSize() {
     WaitingSpanExporter waitingSpanExporter = new WaitingSpanExporter(6);
-    BatchSpansProcessor.Config config =
-        BatchSpansProcessor.Config.newBuilder()
+    BatchSpansProcessor bsp =
```
jkwatson, line 129: would prefer not abbreviating like this. We should prefer spelling out names to make reading the call sites less confusing.

bogdandrutu: To rephrase what @jkwatson said: we are Java not Go :)))

thisthat: Addressed in 0af5490

bogdandrutu:
Please rebase

thisthat:
> Please rebase

Done!

PR #1218, opened 2020-05-14
Extract SpanData to an interface

jkwatson:
This also moves the SpanData inner classes into the interface, and out of the implementation.


sdk/src/main/java/io/opentelemetry/sdk/trace/data/SpanData.java
```diff
@@ -16,63 +16,54 @@
...
 import io.opentelemetry.trace.TraceId;
 import io.opentelemetry.trace.TraceState;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import javax.annotation.concurrent.Immutable;
+import jdk.nashorn.internal.ir.annotations.Immutable;
```
bogdandrutu, line 31: This is not the right annotation

exporters/zipkin/src/test/java/io/opentelemetry/exporters/zipkin/ZipkinSpanExporterTest.java
```diff
@@ -236,7 +238,7 @@ public void testExport_failed() throws IOException {
     when(mockZipkinCall.execute()).thenThrow(new IOException());
 
     ResultCode resultCode =
-        zipkinSpanExporter.export(Collections.singleton(buildStandardSpan().build()));
+        zipkinSpanExporter.export(Collections.<SpanData>singleton(buildStandardSpan().build()));
```
bogdandrutu, line 241: Use singletonList here?

jkwatson: better than a single-item set? Since it takes Collection, either should work, correct?

sdk/src/main/java/io/opentelemetry/sdk/trace/data/SpanData.java
jkwatson: yep. fixed.

exporters/zipkin/src/test/java/io/opentelemetry/exporters/zipkin/ZipkinSpanExporterTest.java
bogdandrutu: Correct just that list seems lightweight compared to set :)

jkwatson: if you look at the implementation of `singleton`, I think it's about as lightweight as it's possible to be. :)

jkwatson: implementation-wise, they're really just a wrapper for a single data item. I like using `singleton` as it's 4 less characters than `singletonList`. But, honestly, I don't care all that much, either way.

bogdandrutu: Leave it as it is, for me was just reading the code and found this inconsistencies, so was more like this is not what I was expecting.

sdk/src/main/java/io/opentelemetry/sdk/trace/data/SpanDataImpl.java
```diff
@@ -0,0 +1,278 @@
...
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * Immutable representation of all data collected by the {@link io.opentelemetry.trace.Span} class.
+ *
+ * @since 0.1.0
+ */
+@Immutable
+@AutoValue
+public abstract class SpanDataImpl implements SpanData {
```
bogdandrutu, line 43: What about moving this into a `testing` package and maybe rename to `SpanData` or `TestingSpanData`?

sdk/src/main/java/io/opentelemetry/sdk/trace/data/LinkData.java
```diff
@@ -0,0 +1,97 @@
...
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * An immutable implementation of {@link Link}.
+ *
+ * @since 0.1.0
+ */
+@Immutable
+@AutoValue
+public abstract class LinkData implements Link {
```
bogdandrutu, line 36: This is inconsistent, SpanData represents the interface to access recorded properties to the Span, but LinkData represents a concrete implementation of the Link interface.

sdk/src/main/java/io/opentelemetry/sdk/trace/data/SpanDataImpl.java
jkwatson: I agree; I'd like to do that when we stop using it as the representation that's used by the RERS. Until then, it is the only implementation we have.

sdk/src/main/java/io/opentelemetry/sdk/trace/data/LinkData.java
jkwatson: Totally agree. Do you have ideas on better naming? I didn't like "Link" which is what it was before. And, we already have another class called "TimedEvent" for the other case.

jkwatson:
> I am happy with the overall change, a bit struggling with the name of the classes.

Definitely. I know that what we had before, when extracted out of the SpanData implementation, was way too confusing, as the names collided directly with other classes.

What if I move them back into the SpanData interface and just call them `Link` and `Event` ?

sdk/src/main/java/io/opentelemetry/sdk/trace/export/SimpleSpansProcessor.java
```diff
@@ -29,7 +29,7 @@
 import java.util.Properties;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import jdk.nashorn.internal.ir.annotations.Immutable;
```
carlosalberto, line 32: Eh, wondering how we ended up with this one here.

carlosalberto:
> What if I move them back into the SpanData interface and just call them Link and Event ?

I'd be up for this, while keeping the 'old' names (at least for now, while we still have `SpanDataImpl` around). Other than that, this looks great to me!

jkwatson:
> > What if I move them back into the SpanData interface and just call them Link and Event ?
> 
> I'd be up for this, while keeping the 'old' names (at least for now, while we still have `SpanDataImpl` around). Other than that, this looks great to me!

ok, done. PTAL

sdk/src/main/java/io/opentelemetry/sdk/trace/export/SimpleSpansProcessor.java
jkwatson: it's easy to have IDEA import the wrong one. I did it early in this PR, myself. :)

PR #1219, opened 2020-05-14
Remove absolute property from measure

bogdandrutu:
Will followup with renaming.


bogdandrutu:
Holding a bit on this PR until we decide the `AbsoluteValueRecorder` issue.

bogdandrutu:
Actually no reason, we can add the new instrument later.

PR #1231, opened 2020-05-15
Return a wrapped span for SpanData so only the mutable data is copied.

jkwatson:
This is a optimization of the creation of the SpanData implementation. 

master benchmark:
```
SpanPipelineBenchmark.runThePipeline_05Threads                                   thrpt    5  3352.511 ±  53.533  ops/ms
SpanPipelineBenchmark.runThePipeline_05Threads:·gc.alloc.rate                    thrpt    5  7233.932 ± 138.314  MB/sec
SpanPipelineBenchmark.runThePipeline_05Threads:·gc.alloc.rate.norm               thrpt    5  3400.001 ±   0.001    B/op
SpanPipelineBenchmark.runThePipeline_05Threads:·gc.churn.PS_Eden_Space           thrpt    5  7310.190 ± 695.826  MB/sec
SpanPipelineBenchmark.runThePipeline_05Threads:·gc.churn.PS_Eden_Space.norm      thrpt    5  3435.823 ± 315.746    B/op
SpanPipelineBenchmark.runThePipeline_05Threads:·gc.churn.PS_Survivor_Space       thrpt    5     0.232 ±   0.327  MB/sec
SpanPipelineBenchmark.runThePipeline_05Threads:·gc.churn.PS_Survivor_Space.norm  thrpt    5     0.109 ±   0.156    B/op
SpanPipelineBenchmark.runThePipeline_05Threads:·gc.count                         thrpt    5    79.000            counts
SpanPipelineBenchmark.runThePipeline_05Threads:·gc.time                          thrpt    5    50.000                ms
```

this branch benchmark:
```
SpanPipelineBenchmark.runThePipeline_05Threads                                   thrpt    5  3980.803 ± 253.972  ops/ms
SpanPipelineBenchmark.runThePipeline_05Threads:·gc.alloc.rate                    thrpt    5  6935.797 ± 455.120  MB/sec
SpanPipelineBenchmark.runThePipeline_05Threads:·gc.alloc.rate.norm               thrpt    5  2744.001 ±   0.001    B/op
SpanPipelineBenchmark.runThePipeline_05Threads:·gc.churn.PS_Eden_Space           thrpt    5  6984.225 ± 279.329  MB/sec
SpanPipelineBenchmark.runThePipeline_05Threads:·gc.churn.PS_Eden_Space.norm      thrpt    5  2763.550 ± 139.847    B/op
SpanPipelineBenchmark.runThePipeline_05Threads:·gc.churn.PS_Survivor_Space       thrpt    5     0.245 ±   0.330  MB/sec
SpanPipelineBenchmark.runThePipeline_05Threads:·gc.churn.PS_Survivor_Space.norm  thrpt    5     0.097 ±   0.135    B/op
SpanPipelineBenchmark.runThePipeline_05Threads:·gc.count                         thrpt    5    71.000            counts
SpanPipelineBenchmark.runThePipeline_05Threads:·gc.time                          thrpt    5    46.000                ms
```

Or graphically (from a different run): 

Here are the benchmark results from master:  

image

And on this branch:

image

### DIFF
--- a/sdk/src/jmh/java/io/opentelemetry/sdk/trace/SpanPipelineBenchmark.java
+++ b/sdk/src/jmh/java/io/opentelemetry/sdk/trace/SpanPipelineBenchmark.java
@@ -57,7 +57,7 @@ public class SpanPipelineBenchmark {
   @Threads(value = 5)
   @Fork(1)
   @Warmup(iterations = 5, time = 1)
-  @Measurement(iterations = 10, time = 1)
+  @Measurement(iterations = 5, time = 1)
   @OutputTimeUnit(TimeUnit.MILLISECONDS)
   public void runThePipeline_05Threads() {
     doWork();
@@ -73,6 +73,8 @@ public class SpanPipelineBenchmark {
             .startSpan();
     span.setAttribute("longAttribute", 33L);
     span.setAttribute("stringAttribute", "test_value");
+    span.setAttribute("doubleAttribute", 4844.44d);
+    span.setAttribute("booleanAttribute", false);
     span.setStatus(Status.OK);
 
     span.addEvent("testEvent");
--- a/sdk/src/jmh/java/io/opentelemetry/sdk/trace/SpanPipelineBenchmark.java
+++ b/sdk/src/jmh/java/io/opentelemetry/sdk/trace/SpanPipelineBenchmark.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright 2020, OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.sdk.trace;
+
+import io.opentelemetry.common.AttributeValue;
+import io.opentelemetry.sdk.OpenTelemetrySdk;
+import io.opentelemetry.sdk.trace.data.SpanData;
+import io.opentelemetry.sdk.trace.export.SimpleSpanProcessor;
+import io.opentelemetry.sdk.trace.export.SpanExporter;
+import io.opentelemetry.trace.Link;
+import io.opentelemetry.trace.Span;
+import io.opentelemetry.trace.Span.Kind;
+import io.opentelemetry.trace.SpanContext;
+import io.opentelemetry.trace.Status;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Threads;
+import org.openjdk.jmh.annotations.Warmup;
+
+@State(Scope.Benchmark)
+public class SpanPipelineBenchmark {
+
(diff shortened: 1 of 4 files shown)


bogdandrutu:
Can you expand on the idea where are you not happy?

jkwatson:
> Can you expand on the idea where are you not happy?

Mostly, I'm not sure that this solution will have much of an impact in the number of allocations. Do you have a benchmark you've been using so we can compare?

bogdandrutu:
I don't but we can write one to confirm. I did look only at a running up and got a snapshot of heap allocs

jkwatson:
> I don't but we can write one to confirm. I did look only at a running up and got a snapshot of heap allocs

I'll see if I can cook something up today.

jkwatson:
ok. I added a benchmark to this PR. I ran this against both master and this branch. This branch was about 10-12% faster, and had a similar percentage of decrease in allocations.

bogdandrutu:
And I am confident that we can optimize even more :)

sdk/src/jmh/java/io/opentelemetry/sdk/trace/SpanPipelineBenchmark.java
```diff
@@ -0,0 +1,112 @@
...
+  @Threads(value = 5)
+  @Fork(1)
+  @Warmup(iterations = 5, time = 1)
+  @Measurement(iterations = 5, time = 1)
+  @OutputTimeUnit(TimeUnit.MILLISECONDS)
+  public void runThePipeline_05Threads() {
+    doWork();
+  }
+
+  private void doWork() {
```
bogdandrutu, line 66: Add some events too

sdk/src/main/java/io/opentelemetry/sdk/trace/SpanWrapper.java
```diff
@@ -0,0 +1,169 @@
...
+      List<Event> events,
+      Map<String, AttributeValue> attributes,
+      int totalAttributeCount,
+      int totalRecordedEvents,
+      Status status) {
+    return new AutoValue_SpanWrapper(
+        delegate,
+        links,
+        events,
+        new HashMap<>(attributes),
```
bogdandrutu, line 64: Do we need to copy these? Just wrapping in a unmodifiable map should be enough. After the span ends :) before it is needed.

jkwatson: good call. I'll make that change and re-benchmark

jkwatson: Actually, are you suggesting we do this inside the RERS when the span ends? If we did that, we'd need to have an extra field to hold it, unless we created a version of the AttributesMap that was immutable.

bogdandrutu: AttributeMap is immutable after end.

jkwatson: It isn't right now, if we let it escape the RERS. Are you saying we should make it so?

bogdandrutu: Not sure I understand the concern. We call the wrapper using `Collections.unmodifiableMap(attributeMap)`. So nobody can change the attributeMap via SpanData, and nobody can change attributeMap via RERS because hasEnded is true so no changes will happen.

jkwatson: Right, if we do the wrapping here, then everything is fine. I thought maybe you were considering doing the wrapping earlier, immediately on the span ending, rather than on the call to `toSpanData`.

jkwatson: Also, I think we should wrap the links and events, as well.

bogdandrutu: We can do both later. Let's merge it as it is for the moment. I think this is a good start.

bogdandrutu:
@jkwatson please add in the description results for the benchmark.

jkwatson:
> @jkwatson please add in the description results for the benchmark.

will do

bogdandrutu:
Please rebase to pick the codecov change :)

jkwatson:
> Please rebase to pick the codecov change :)

I think this needs some unit tests on the wrapper, which I hadn't written yet.

bogdandrutu:
@jkwatson let's see what the check says once you rebase.

jkwatson:
rebased (also, I had locally added the unmodifiable wrappers and it didn't significantly change the benchmarks, so they slipped in here, as well).

bogdandrutu:
@jkwatson my bad I did not expect that you added the change, it is not correct right now. I will create a PR to fix this.

jkwatson:
> @jkwatson my bad I did not expect that you added the change, it is not correct right now. I will create a PR to fix this.

What's the issue? I can fix it, since I'm in the code adding test coverage.

PR #1233, opened 2020-05-16
Cleanup samplers.

bogdandrutu:


sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
```diff
@@ -68,7 +80,17 @@ public static Sampler alwaysOn() {
    * @since 0.1.0
    */
   public static Sampler alwaysOff() {
-    return ALWAYS_OFF;
+    return AlwaysOffSampler.INSTANCE;
+  }
+
+  /**
+   * Returns a {@link Sampler} that always makes a "no" decision on {@link Span} sampling.
```
jkwatson, line 87: this javadoc needs to be corrected.

sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
```diff
@@ -43,10 +43,22 @@
 @Immutable
 public final class Samplers {
 
-  private static final Sampler ALWAYS_ON = new AlwaysOnSampler();
-  private static final Sampler ALWAYS_OFF = new AlwaysOffSampler();
-  private static final Decision ALWAYS_ON_DECISION = new SimpleDecision(/* decision= */ true);
-  private static final Decision ALWAYS_OFF_DECISION = new SimpleDecision(/* decision= */ false);
+  private static final Decision ALWAYS_ON_DECISION =
```
jkwatson, line 46: Now that this is used for something other than the always-on sampler, I think we should rename this to just `SIMPLE_ON_DECISION` or maybe even just `ON_DECISION`

sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
```diff
@@ -223,57 +261,27 @@ public final String getDescription() {
 
   /** Sampling decision without attributes. */
   @Immutable
-  private static final class SimpleDecision implements Decision {
-
-    private final boolean decision;
-
+  @AutoValue
+  abstract static class DecisionImpl implements Decision {
```
jkwatson, line 265: javadoc on this class isn't correct any more.

sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
Oberon00: Related: #1256. Maybe EMPTY_(NOT_)SAMPLED_DECISION?

sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
bogdandrutu: Obsolete, decided to split into just the cleanup for the moment.

sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
bogdandrutu: Done. Not necessary to have a javadoc for this class.

sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
bogdandrutu: Done.

sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
```diff
@@ -84,8 +96,8 @@ public static Sampler probability(double probability) {
   }
 
   @Immutable
-  private static final class AlwaysOnSampler implements Sampler {
-    AlwaysOnSampler() {}
+  private enum AlwaysOnSampler implements Sampler {
```
jkwatson, line 99: Why change these to enums? This is a non-idiomatic usage of an enum.

bogdandrutu: It is a better solution for singleton classes, read https://dzone.com/articles/java-singletons-using-enum

jkwatson: I'm not sure one graduate student's opinion on DZone should sway us away from more idiomatic solutions. :) 

I don't about the Serialization issue... if anyone is using default Java serialization these days, they are asking for all sorts of troubles.  The reflection issue doesn't matter at all for this case. If someone wanted to create a million sampler instances, they would only be hurting themselves, not the correctness of the code.

jkwatson: Also, it's worth reading the comments on that article. :)

bogdandrutu: That was just an example, there are a lot more articles about this problem and they propose this solution as being they way to do it, do it seems it is common enough.

As for being idiomatic I think it is (Joshua Bloch says this in Effective Java):
https://www.informit.com/articles/article.aspx?p=1216151&seqNum=3
> This approach is functionally equivalent to the public field approach, except that it is more concise, provides the serialization machinery for free, and provides an ironclad guarantee against multiple instantiation, even in the face of sophisticated serialization or reflection attacks. While this approach has yet to be widely adopted, a single-element enum type is the best way to implement a singleton.

carlosalberto: I also vow to not use enums, at least for now. Honestly I don't think it's truly idiomatic, even if it's _relatively_ common (which I'm not happy about, as honestly it _feels_ like a hack around the singleton pattern - and this is probably the second time I've seen it in my life ;) ). 

Let's consider changing this in another PR (as this might apply to `HttpTraceContext`, `B3Propagator`, etc). If we agree, we go and apply this pattern everywhere.

bogdandrutu: @carlosalberto this is not a public class. So I don't see any problem. We also use this pattern already, and as I said I would trust Joshua Bloch more than almost any other java developer :)

carlosalberto: > this is not a public class

Fair enough. As long as we do this only for private classes I'm cool with it.

bogdandrutu: @carlosalberto also publicly I would not expose a variable for a singleton, instead I would go with a static method to retrieve the singleton. That allows us to actually change to not be a singleton if we want, or do other expensive operations (if we want in the future) that should not be part of the static initialization. In other words it gives us more flexibility in the future.

jkwatson: I had completely forgotten that Josh has recommended this approach (probably because I've never actually seen it used in projects I've worked on). Let's go with it, especially as it's a private usage.

bogdandrutu: @jkwatson please press the magic button to approve this.

Issue #1157, opened 2020-04-29
JaegerRemoteSamplerTest fails occasionally


jkwatson:
Very occasionally, this error causes the test to fail.

```
java.lang.AssertionError: Resource ManagedChannelOrphanWrapper{delegate=ManagedChannelImpl{logId=1388, target=directaddress:///my-service}} can not be released in time at the end of test

	at io.grpc.testing.GrpcCleanupRule.teardown(GrpcCleanupRule.java:169)
	at io.grpc.testing.GrpcCleanupRule.access$100(GrpcCleanupRule.java:48)
	at io.grpc.testing.GrpcCleanupRule$1.evaluate(GrpcCleanupRule.java:138)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:53)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:230)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:58)
```

jkwatson:
Fixed by #1276

PR #1235, opened 2020-05-17
Separating the B3 logic for Single and Multiple headers

### LINKED ISSUE #999: Separate single and combined B3 header logic
The B3Propagator includes logic for both injecting and extacting span context from both single and multiple headers. To help simplify the code, the two code paths could be broken into two classes.

ksameersrk:
Fixes #999 

I have separated out the Single and Multiple headers logic for Injector and Extractor.
The structure would be something like this
- `abstract B3PropagatorInjector`
    - `final B3PropagatorInjectorSingleHeader`
    - `final B3PropagatorInjectorMultipleHeaders`

- `abstract B3PropagatorExtractor`
    - `final B3PropagatorExtractorSingleHeader`
    - `final B3PropagatorExtractorMultipleHeaders`


ksameersrk:
I signed it

ksameersrk:
I signed it

carlosalberto:
Overall looks good but it feels slightly weird that the abstract methods are basically part of the `Propagator` interface:

```java
// B3PropagatorExtractor -> is this really part of *only* the B3 propagator format?
<C> Context extract(Context context, C carrier, HttpTextFormat.Getter<C> getter);
```

and at the same time child classes **only use** static methods from the base class.

contrib/trace_propagators/src/main/java/io/opentelemetry/contrib/trace/propagation/B3Propagator.java
```diff
@@ -62,8 +62,13 @@ public B3Propagator() {
    * @param singleHeader whether to use single or multiple headers.
    */
   public B3Propagator(boolean singleHeader) {
```
bogdandrutu, line 64: use a `create` factory method instead of a public ctor.

carlosalberto: Talking of, since the `Propagators` do not keep state, I'm wondering on why we don't expose them as static instances, i.e. `getSingleHeaderPropagator()` and `getMultipleHeaderPropagator()`.

ksameersrk: @bogdandrutu  I can make this change. But not sure why we would need a factor when we just have one argument.

@carlosalberto  Now I think about this, making static methods make sense. But, again I don't know what the implication will be. I think this will cause a breaking change for the clients while upgrading to the next version.

jkwatson: Breaking changes are ok at the moment, since we haven't hit version 1.0 yet.

jkwatson: Yeah, it seems like a `B3Propagators.SINGLE_HEADER` and `B3Propagators.MULTI_HEADER` would be a great way to get access to them.

bogdandrutu: Factory methods are better than constructors because they can be change to return different instances for example.

bogdandrutu: Happy with @jkwatson solution fyi

ksameersrk: I have made changes based on the feedback.

I have made the current constructor as `private`. I have exposed a public static variable for `B3Propagators.SINGLE_HEADER` and `B3Propagators.MULTIPLE_HEADER`.

contrib/trace_propagators/src/main/java/io/opentelemetry/contrib/trace/propagation/B3Propagator.java
```diff
@@ -51,19 +51,31 @@
   private final B3PropagatorInjector b3PropagatorInjector;
   private final B3PropagatorExtractor b3PropagatorExtractor;
 
-  /** Creates a new instance of {@link B3Propagator}. Defaults to use multiple headers. */
-  public B3Propagator() {
-    this(false);
-  }
+  /**
+   * Creates a new instance of {@link B3Propagator}. This instance contains the reference to {@link
```
jkwatson, line 55: This field won't create a new instance. The javadoc should say that it will return an instance and describe it's functionality, rather than the internal implementation.  (same comment goes for the multi-header field)

contrib/trace_propagators/src/main/java/io/opentelemetry/contrib/trace/propagation/B3Propagator.java
```diff
@@ -51,19 +51,31 @@
...
 
   /**
    * Creates a new instance of {@link B3Propagator}.
    *
    * @param singleHeader whether to use single or multiple headers.
    */
-  public B3Propagator(boolean singleHeader) {
-    b3PropagatorInjector = new B3PropagatorInjector(singleHeader);
-    b3PropagatorExtractor = new B3PropagatorExtractor(singleHeader);
+  private B3Propagator(boolean singleHeader) {
```
jkwatson, line 71: since this constructor is private, and only called from the field declarations, you could just inline the boolean logic into the call sites (i.e. get rid of the boolean and just pass in the two implementations that you want in the right constructor calls)

contrib/trace_propagators/src/main/java/io/opentelemetry/contrib/trace/propagation/B3Propagator.java
ksameersrk: Fixed the incorrect JavaDoc.

contrib/trace_propagators/src/main/java/io/opentelemetry/contrib/trace/propagation/B3Propagator.java
ksameersrk: Took care of it.

contrib/trace_propagators/src/main/java/io/opentelemetry/contrib/trace/propagation/B3Propagator.java
carlosalberto: One last thing: do we want to expose these singletons directly, or should we instead expose them through a static methods? We have historically chosen the second option, which I think it's the slightly better approach ;)

jkwatson: no strong opinion from me. For consistency, it's probably better to hide the instances behind static accessors.

ksameersrk: Made change to expose the singletons via static getters.

carlosalberto:
Looking great @ksameersrk FYI the build is failing with your latest changes!

Thanks for iterating on this!

ksameersrk:
@carlosalberto. I just ran `make test verify-format` after setting `enable.docker.tests=true` locally. It ran just fine. Is there any other way to run full suite of tests locally similar to CI?

I see that [opentelemetry-sdk-contrib-jaeger-remote-sampler:test](https://circleci.com/gh/open-telemetry/opentelemetry-java/4162?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link) is failing. Let me check why it is happening just on remote and not locally.

ksameersrk:
It got itself fixed after pushing the dummy commit. The problem seems to be with the environment for earlier commit.

jkwatson:
> @carlosalberto. I just ran `make test verify-format` after setting `enable.docker.tests=true` locally. It ran just fine. Is there any other way to run full suite of tests locally similar to CI?
> 
> I see that [opentelemetry-sdk-contrib-jaeger-remote-sampler:test](https://circleci.com/gh/open-telemetry/opentelemetry-java/4162?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link) is failing. Let me check why it is happening just on remote and not locally.
> 
> Edit: also ran `make test-with-docker` locally and is a success.

Was it this issue:  #1157  ?

ksameersrk:
Yes. That was the precise error.

contrib/trace_propagators/src/main/java/io/opentelemetry/contrib/trace/propagation/B3PropagatorInjector.java
```diff
@@ -18,59 +18,9 @@
...
-              ? B3Propagator.IS_SAMPLED
-              : B3Propagator.NOT_SAMPLED;
-      setter.set(carrier, B3Propagator.COMBINED_HEADER, new String(chars));
-    } else {
-      setter.set(carrier, B3Propagator.TRACE_ID_HEADER, spanContext.getTraceId().toLowerBase16());
-      setter.set(carrier, B3Propagator.SPAN_ID_HEADER, spanContext.getSpanId().toLowerBase16());
-      setter.set(carrier, B3Propagator.SAMPLED_HEADER, sampled);
-    }
-  }
+abstract class B3PropagatorInjector {
```
jkwatson, line 24: What do you think of making this an interface, rather than an abstract class?

contrib/trace_propagators/src/main/java/io/opentelemetry/contrib/trace/propagation/B3PropagatorExtractor.java
```diff
@@ -16,122 +16,32 @@
...
 import io.opentelemetry.trace.TraceState;
-import io.opentelemetry.trace.TracingContextUtils;
-import java.util.Objects;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.concurrent.Immutable;
 
 @Immutable
-final class B3PropagatorExtractor {
+abstract class B3PropagatorExtractor {
```
jkwatson, line 36: What do think about making this an interface, rather than an abstract class?

jkwatson:
Looks good, aside from the fact that I'd rather use interfaces, rather than abstract base classes.

contrib/trace_propagators/src/main/java/io/opentelemetry/contrib/trace/propagation/B3PropagatorExtractor.java
ksameersrk: Okay. If we make them interface then, will we be moving `buildSpanContext`, `isTraceIdValid` and `isSpanIdValid ` to Util class I guess. Not sure if we want to make them `default` method of the interface.

jkwatson: You can still have static methods on your interface.

ksameersrk: I believe it is Java 8 and above right? I don't think we have static and default in Java 7.

jkwatson: oh shoot, you're right. Since they are only used in one place, you can have an inner util class in the interface, like we do in the `SpanData` right now, rather than pulling it out of the interface entirely.

ksameersrk: sounds good

ksameersrk:
Just merged with master since there was conflict.

PR #1237, opened 2020-05-18
use java 8 for tests

### LINKED ISSUE #1187: Use Java 8 or newer for tests
This came up during review of #1176.  We had to use a specific version of ```zipkin-junit``` that uses a 3.x version of the ```okhttp``` lib which can compile with JDK 7.

> yeah it seems odd to not use java 8 in tests. Most projects have build requirements at least 8 if not above these days.. consider when you want to add features that conditionally work with futures etc.

_Originally posted by @adriancole in https://github.com/open-telemetry/opentelemetry-java/pull/1176_

jarebudev:
resolves #1187 

Tests now use Java 8.
I modified the ```HttpTraceContextTest``` to demonstrate that Java 8 is used,  I removed some boilerplate, which is now using lambdas.
I also disabled the ```Java7ApiChecker``` and ```AndroidJdkLibsChecker``` errorprone checks when compiling tests as these will warn if you use e.g. streams in your test.

### DIFF
--- a/api/src/test/java/io/opentelemetry/trace/propagation/HttpTraceContextTest.java
+++ b/api/src/test/java/io/opentelemetry/trace/propagation/HttpTraceContextTest.java
@@ -33,7 +33,6 @@ import io.opentelemetry.trace.TracingContextUtils;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
-import javax.annotation.Nullable;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
@@ -58,21 +57,8 @@ public class HttpTraceContextTest {
       "00-" + TRACE_ID_BASE16 + "-" + SPAN_ID_BASE16 + "-01";
   private static final String TRACEPARENT_HEADER_NOT_SAMPLED =
       "00-" + TRACE_ID_BASE16 + "-" + SPAN_ID_BASE16 + "-00";
-  private static final Setter<Map<String, String>> setter =
-      new Setter<Map<String, String>>() {
-        @Override
-        public void set(Map<String, String> carrier, String key, String value) {
-          carrier.put(key, value);
-        }
-      };
-  private static final Getter<Map<String, String>> getter =
-      new Getter<Map<String, String>>() {
-        @Nullable
-        @Override
-        public String get(Map<String, String> carrier, String key) {
-          return carrier.get(key);
-        }
-      };
+  private static final Setter<Map<String, String>> setter = Map::put;
+  private static final Getter<Map<String, String>> getter = (carrier, key) -> carrier.get(key);
   // Encoding preserves the order which is the reverse order of adding.
   private static final String TRACESTATE_NOT_DEFAULT_ENCODING = "bar=baz,foo=bar";
   private static final String TRACESTATE_NOT_DEFAULT_ENCODING_WITH_SPACES =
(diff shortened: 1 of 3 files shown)


Oberon00:
While this would certainly a QOL improvement, I think if we want to support Java 7 it follows logically that we must not only compile for Java 7 but also test that it actually works on Java 7. Which means that we need at least some tests that run with Java 7.

carlosalberto:
I'm fine with this change and I'm approving but would like to hear if somebody has the same concern as @Oberon00 .

bogdandrutu:
@jkwatson not sure is about "usability", an argument to have `java7` tests is to compile and run unittests using java7 jvm.

jkwatson:
> @jkwatson not sure is about "usability", an argument to have `java7` tests is to compile and run unittests using java7 jvm.

I guess I'm also not concerned with that. I trust that if we compile to java 7 bytecode, it'll run fine on java 7. Are you thinking it would be a good double-check that we haven't messed up the build and are accidentally only building java 8+ bytecode?

jarebudev:
> > @jkwatson not sure is about "usability", an argument to have `java7` tests is to compile and run unittests using java7 jvm.
> 
> I guess I'm also not concerned with that. I trust that if we compile to java 7 bytecode, it'll run fine on java 7. Are you thinking it would be a good double-check that we haven't messed up the build and are accidentally only building java 8+ bytecode?

If this is the concern, could we check for this scenario in an integration test?  E.g. create an example app that's packaged into a Docker image, that uses a Java7 base image, and then test the application using Testcontainers as part of the CI build?

Oberon00:
@jarebudev Testing Java 7, with a real Java 7 JVM in an integration test sounds like a good solution!

bogdandrutu:
Please create an issue to do this.

jarebudev:
> Please create an issue to do this.

created #1249

iNikem:
Not testing with java7 may lead to the situation when we accidentally use JDK API which is not available in java7. Compilation to java7 bytecode will not find this. Thus running at least a representative test suite on java 7 during automatic build is a MUST.

jkwatson:
> Not testing with java7 may lead to the situation when we accidentally use JDK API which is not available in java7. Compilation to java7 bytecode will not find this. Thus running at least a representative test suite on java 7 during automatic build is a MUST.

Won't the animalsniffer already do this for us?

iNikem:
Ah, indeed. animalnsiffer may be enough. Or just follow auto-instr path and compile with java11 and use `-release` option: https://github.com/open-telemetry/opentelemetry-auto-instr-java/pull/390

jkwatson:
Since this is a significant change (if only to tests), I'd like to see all maintainers approve it before merging. @bogdandrutu are you ok with this?

anuraaga:
Not a strong recommendation but one idea is also retrolambda - would allow using lambda in the businesses logic too and that with Guava in tests would probably also be pretty reasonable QOL.

But otherwise +1 to this and a separate PR which switches to JDK11 and the `release` flag that's much simpler than animal sniffer.

api/src/test/java/io/opentelemetry/trace/propagation/HttpTraceContextTest.java
```diff
@@ -58,21 +57,8 @@
...
-  private static final Getter<Map<String, String>> getter =
-      new Getter<Map<String, String>>() {
-        @Nullable
-        @Override
-        public String get(Map<String, String> carrier, String key) {
-          return carrier.get(key);
-        }
-      };
+  private static final Setter<Map<String, String>> setter = Map::put;
+  private static final Getter<Map<String, String>> getter = (carrier, key) -> carrier.get(key);
```
bogdandrutu, line 61: Noooo, I have to learn how to write lambdas :(

PR #1252, opened 2020-05-20
Require coverage 90% for api and sdk

bogdandrutu:


jkwatson:
I'm not generally a fan of having coverage requirements, as I have found that it mostly just causes people to waste time writing useless tests just for the purpose of having coverage, rather than focusing on testing corner cases and the real functionality.

bogdandrutu:
I do agree in general, that's why I went with 95%. But not having a minimum (which should not be 100%) is also bad, people tend to not write test case for error cases like conditions where arguments should not be null. I think having a compromise of a minimum 90 or 95 percent gets the most out of this situation

DotSpy:
Imo we can do this in small steps: set 90% and if we will see a problem with it increase to 95% or further. This 5% will not do much difference when people contributing small PR's, like fixing smth, rather we should review this PR's and see that cases covered with tests

jkwatson:
It looks like we're right around 90% on both the API and SDK right now, if I read the reports from the latest master build correctly. If we set it to 95%, would the build start failing immediately?

bogdandrutu:
@jkwatson happy to go with 90%. In my opinion at least for the API we should have better coverage (maybe for the moment set it as a goal), since this artifact may be included by a lot of libraries it is recommended that we have better confidence on behavior and implementation. What do you think?

carlosalberto:
> happy to go with 90%. In my opinion at least for the API we should have better coverage (maybe for the moment set it as a goal), since this artifact may be included by a lot of libraries it is recommended that we have better confidence on behavior and implementation.

This sounds great to me. 90% is a good start and then we can work our way up (and see how it feels/works for us). Maybe create an issue to get to that goal?

PR #1254, opened 2020-05-20
Configuration for Disruptor and Zipkin

thisthat:
The `Disruptor` and the `Zipkin` exporter went under my radar during #996


exporters/zipkin/src/main/java/io/opentelemetry/exporters/zipkin/ZipkinExporterConfiguration.java
```diff
@@ -68,7 +92,7 @@ public static Builder builder() {
      * <p>This is a primary label for trace lookup and aggregation, so it should be intuitive and
      * consistent. Many use a name from service discovery.
      *
-     * <p>Note: this value, will be superceded by the value of {@link
+     * <p>Note: this value, will be superseded by the value of {@link
```
jkwatson, line 95: note: both of these are valid spellings. I'm ok with either.

sdk/src/main/java/io/opentelemetry/sdk/common/export/ConfigBuilder.java
```diff
@@ -139,4 +139,19 @@ protected static Long getLongProperty(String name, Map<String, String> map) {
...
+
+  /**
+   * Get a {@link String} property from the map, {@code null} if it cannot be found.
+   *
+   * @param name The property name
+   * @param map The map where to look for the property
+   * @return the {@link String} value of the property, {@code null} if the property cannot be found.
+   */
+  @Nullable
+  protected static String getStringProperty(String name, Map<String, String> map) {
```
jkwatson, line 151: do we have a unit test for this new method?

bogdandrutu: I think we should add method for this.

Also this should just be map.get(). I am confused. https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#get-java.lang.Object-

exporters/zipkin/src/main/java/io/opentelemetry/exporters/zipkin/ZipkinExporterConfiguration.java
```diff
@@ -122,5 +146,26 @@ public Builder setEndpoint(String endpoint) {
...
     public abstract ZipkinExporterConfiguration build();
+
+    /**
+     * Sets the configuration values from the given configuration map for only the available keys.
+     *
+     * @param configMap {@link Map} holding the configuration values.
+     * @return this.
+     */
+    @Override
+    protected Builder fromConfigMap(
```
jkwatson, line 157: Can we write unit tests for this?

sdk_contrib/async_processor/src/main/java/io/opentelemetry/sdk/contrib/trace/export/DisruptorAsyncSpanProcessor.java
```diff
@@ -149,6 +183,29 @@ public DisruptorAsyncSpanProcessor build() {
           spanProcessor.isStartRequired(),
           spanProcessor.isEndRequired());
     }
+
+    @Override
+    protected Builder fromConfigMap(
```
jkwatson, line 188: Can you write unit tests for this?

trask:
What do you think about moving `ZipkinExporterConfiguration` functionality to `ZipkinSpanExporter.Builder`, to follow the pattern from `BatchSpanProcessor` where you don't have a separate config object?

sdk/src/main/java/io/opentelemetry/sdk/common/export/ConfigBuilder.java
thisthat: I'll provide tests for this in a separate PR with further cleanups :)

sdk_contrib/async_processor/src/test/java/io/opentelemetry/sdk/contrib/trace/export/DisruptorAsyncSpanProcessorTest.java
```diff
@@ -268,4 +272,26 @@ public void multipleForceFlush() {
...
   }
+
+  abstract static class ConfigBuilderTest extends ConfigBuilder<ConfigBuilderTest> {
+    public static NamingConvention getNaming() {
+      return NamingConvention.DOT;
+    }
+  }
+
+  @Test
+  public void configTest() {
```
jkwatson, line 283: We need a test like this for the zipkin config, as well.

thisthat: Oh thank you! I totally miss it! :facepalm:

Issue #1181, opened 2020-05-06
Unify Javadoc for configuration classes


thisthat:
With #996, [Span|Metric] Processors and Exporters can accept configuration parameters from different sources. The documentation that lists the names for the names used to do the look-up is duplicated among the different classes. This documentation should be unified in the class-level Javadoc and in the package-info.

PR #1255, opened 2020-05-20
Support env vars for TraceConfig initialization

### LINKED ISSUE #1239: Support configuration from env vars for TraceConfig class
TraceConfig to get initialized with configuration properties like `DEFAULT_SAMPLER`, `DEFAULT_SPAN_MAX_NUM_ATTRIBUTES`, etc., from env variables if exists, Instead of configuring it through programmatically, it would be convenient to configure these parameters via env variables  

We can have a config builder reading properties from env similar to what has been achieved in this commit 
https://github.com/open-telemetry/opentelemetry-java/commit/a99a571e7ace327da527e19d706ab7501fc060a0  

There are no alternatives that I could think of

**Additional context**
There is a need for configuring the sampler in auto-instr-java agent too. 
Ref: https://github.com/open-telemetry/opentelemetry-auto-instr-java/issues/374

If you think that this could be a value adding ask, I am more than happy to send a PR for the same.

RashmiRam:
1. Extended ConfigBuilder for building TraceConfig class to support config from env vars
2. Added getDoubleProperty to ConfigBuilder class

Fixes #1239

### DIFF
--- a/sdk/src/main/java/io/opentelemetry/sdk/common/export/ConfigBuilder.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/common/export/ConfigBuilder.java
@@ -151,4 +151,21 @@ public abstract class ConfigBuilder<T> {
   protected static String getStringProperty(String name, Map<String, String> map) {
     return map.get(name);
   }
+
+  /**
+   * Get a double property from the map, {@code null} if it cannot be found or it has a wrong type.
+   *
+   * @param name The property name
+   * @param map The map where to look for the property
+   * @return the {@link Double} value of the property, {@code null} in case of error or if the
+   *     property cannot be found.
+   */
+  @Nullable
+  protected static Double getDoubleProperty(String name, Map<String, String> map) {
+    try {
+      return Double.parseDouble(map.get(name));
+    } catch (NumberFormatException | NullPointerException ex) {
+      return null;
+    }
+  }
 }
--- a/sdk/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
@@ -32,7 +32,7 @@ import javax.annotation.concurrent.Immutable;
  * Class that holds global trace parameters.
  *
  * <p>Note: To update the TraceConfig associated with a {@link
- * io.opentelemetry.sdk.trace.TracerSdkProvider}, you should use the {@link #newBuilder()} method on
+ * io.opentelemetry.sdk.trace.TracerSdkProvider}, you should use the {@link #toBuilder()} method on
  * the TraceConfig currently assigned to the provider, make the changes desired to the {@link
  * Builder} instance, then use the {@link
  * io.opentelemetry.sdk.trace.TracerSdkProvider#updateActiveTraceConfig(TraceConfig)} with the
--- a/sdk/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
@@ -91,16 +91,16 @@ public abstract class TraceConfig {
   private static final int DEFAULT_SPAN_MAX_NUM_ATTRIBUTES_PER_LINK = 32;
 
   /**
-   * Returns a new {@link TraceConfig} reading the configuration values from the environment and
-   * from system properties. System properties override values defined in the environment. If a
-   * configuration value is missing, it uses the default value.
+   * Returns the default {@code TraceConfig}.
    *
    * @return the default {@code TraceConfig}.
    * @since 0.1.0
    */
   public static TraceConfig getDefault() {
-    return newBuilder().readEnvironmentVariables().readSystemProperties().build();
+    return DEFAULT;
   }
+
+  private static final TraceConfig DEFAULT = TraceConfig.newBuilder().build();
 
   /**
    * Returns the global default {@code Sampler} which is used when constructing a new {@code Span}.
--- a/sdk/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
@@ -47,7 +47,8 @@ import javax.annotation.concurrent.Immutable;
  * for the following names:
  *
  * <ul>
- *   <li>{@code otel.config.sampler.probability}: to set the global default sampler for traces.
+ *   <li>{@code otel.config.sampler.probability}: to set the global default sampler which is used
+ *       when constructing a new {@code Span}.
  *   <li>{@code otel.config.max.attrs}: to set the global default max number of attributes per
  *       {@link Span}.
  *   <li>{@code otel.config.max.events}: to set the global default max number of {@link Event}s per
@@ -63,7 +64,8 @@ import javax.annotation.concurrent.Immutable;
  * <p>For environment variables, {@link TraceConfig} will look for the following names:
  *
  * <ul>
- *   <li>{@code OTEL_CONFIG_SAMPLER_PROBABILITY}: to set the global default sampler for traces.
+ *   <li>{@code OTEL_CONFIG_SAMPLER_PROBABILITY}: to set the global default sampler which is used
+ *       when constructing a new {@code Span}.
  *   <li>{@code OTEL_CONFIG_MAX_ATTRS}: to set the global default max number of attributes per
  *       {@link Span}.
  *   <li>{@code OTEL_CONFIG_MAX_EVENTS}: to set the global default max number of {@link Event}s per
@@ -259,7 +261,8 @@ public abstract class TraceConfig {
      * Sets the global default {@code Sampler}. It must be not {@code null} otherwise {@link
      * #build()} will throw an exception.
      *
-     * @param samplerProbability the global default {@code Sampler}.
+     * @param samplerProbability the global default probability used to make decisions on {@link
+     *     Span} sampling.
      * @return this.
      */
     public Builder setSamplerProbability(double samplerProbability) {
@@ -328,11 +331,10 @@ public abstract class TraceConfig {
      * Builds and returns a {@code TraceConfig} with the desired values.
      *
      * @return a {@code TraceConfig} with the desired values.
-     * @throws IllegalArgumentException if any of the max numbers are not positive
+     * @throws IllegalArgumentException if any of the max numbers are not positive.
      */
     public TraceConfig build() {
       TraceConfig traceConfig = autoBuild();
-      Preconditions.checkArgument(traceConfig.getSampler() != null, "sampler");
       Preconditions.checkArgument(
           traceConfig.getMaxNumberOfAttributes() > 0, "maxNumberOfAttributes");
       Preconditions.checkArgument(traceConfig.getMaxNumberOfEvents() > 0, "maxNumberOfEvents");
--- a/sdk/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
@@ -79,13 +79,14 @@ import javax.annotation.concurrent.Immutable;
 @AutoValue
 @Immutable
 public abstract class TraceConfig {
-
+  // These values are the default values for all the global parameters.
+  // TODO: decide which default sampler to use
+  private static final Sampler DEFAULT_SAMPLER = Samplers.alwaysOn();
   private static final int DEFAULT_SPAN_MAX_NUM_ATTRIBUTES = 32;
   private static final int DEFAULT_SPAN_MAX_NUM_EVENTS = 128;
   private static final int DEFAULT_SPAN_MAX_NUM_LINKS = 32;
   private static final int DEFAULT_SPAN_MAX_NUM_ATTRIBUTES_PER_EVENT = 32;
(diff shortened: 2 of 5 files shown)


sdk/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
```diff
@@ -35,17 +38,16 @@
  * io.opentelemetry.sdk.trace.TracerSdkProvider#updateActiveTraceConfig(TraceConfig)} with the
  * resulting TraceConfig instance.
  */
-@AutoValue
```
thisthat, line 38: Please, keep the AutoValue so we don't need to write and maintain trivial code

sdk/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
```diff
@@ -58,76 +60,255 @@ public static TraceConfig getDefault() {
...
   private static final TraceConfig DEFAULT =
-      TraceConfig.newBuilder()
-          .setSampler(DEFAULT_SAMPLER)
-          .setMaxNumberOfAttributes(DEFAULT_SPAN_MAX_NUM_ATTRIBUTES)
-          .setMaxNumberOfEvents(DEFAULT_SPAN_MAX_NUM_EVENTS)
-          .setMaxNumberOfLinks(DEFAULT_SPAN_MAX_NUM_LINKS)
-          .setMaxNumberOfAttributesPerEvent(DEFAULT_SPAN_MAX_NUM_ATTRIBUTES_PER_EVENT)
-          .setMaxNumberOfAttributesPerLink(DEFAULT_SPAN_MAX_NUM_ATTRIBUTES_PER_LINK)
-          .build();
+      TraceConfig.newBuilder().readEnvironmentVariables().readSystemProperties().build();
```
thisthat, line 63: Please, document the behavior env < sys prop in the `getDefualt()` method

sdk/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
```diff
@@ -58,76 +60,255 @@ public static TraceConfig getDefault() {
...
    * Returns a {@link Builder} initialized to the same property values as the current instance.
    *
    * @return a {@link Builder} initialized to the same property values as the current instance.
    */
-  public abstract Builder toBuilder();
+  public Builder toBuilder() {
+    return new Builder(this);
+  }
+
+  public static final class Builder extends ConfigBuilder<Builder> {
```
thisthat, line 152: Also here, please use AutoValue. A reference of how using AutoValue with ConfigBuilder is available [here](https://github.com/dynatrace-oss-contrib/opentelemetry-java/blob/config-builder/exporters/zipkin/src/main/java/io/opentelemetry/exporters/zipkin/ZipkinExporterConfiguration.java#L80)

sdk/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
```diff
@@ -58,76 +60,255 @@ public static TraceConfig getDefault() {
...
+      this.sampler = traceConfig.sampler;
+      this.maxNumberOfEvents = traceConfig.maxNumberOfEvents;
+      this.maxNumberOfLinks = traceConfig.maxNumberOfLinks;
+      this.maxNumberOfAttributesPerLink = traceConfig.maxNumberOfAttributesPerLink;
+      this.maxNumberOfAttributesPerEvent = traceConfig.maxNumberOfAttributesPerEvent;
+    }
+
+    private Builder() {}
+
+    /**
```
thisthat, line 184: In #1181, we decided to get rid of this documentation from the methods and push it to the class and package-info javadoc

RashmiRam:
@thisthat Thanks for the review. Addressed all the comments.

sdk/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
```diff
@@ -34,39 +39,66 @@
...
+ *       {@link Event}.
+ *   <li>{@code OTEL_CONFIG_MAX_LINK_ATTRS}: to set the global default max number of attributes per
+ *       {@link Link}.
+ * </ul>
  */
 @AutoValue
 @Immutable
 public abstract class TraceConfig {
-  // These values are the default values for all the global parameters.
-  // TODO: decide which default sampler to use
```
thisthat, line 42: Please, open an issue or leave it in otherwise, we will forget of this detail ;)

sdk/src/main/java/io/opentelemetry/sdk/trace/config/package-info.java
```diff
@@ -0,0 +1,74 @@
...
+ *   <li>{@code SPAN_MAX_NUM_EVENTS: 128}
+ *   <li>{@code SPAN_MAX_NUM_LINKS: 32}
+ *   <li>{@code SPAN_MAX_NUM_ATTRIBUTES_PER_EVENT: 32}
+ *   <li>{@code SPAN_MAX_NUM_ATTRIBUTES_PER_LINK: 32}
+ * </ul>
+ *
+ * <p>Values for {@link io.opentelemetry.sdk.trace.config.TraceConfig} can be read from system
+ * properties, environment variables, or {@link java.util.Properties} objects.
+ *
+ * <p>For System Properties and {@link java.util.Properties} objects, {@link
```
thisthat, line 40:
```suggestion
 * <p>For system properties and {@link java.util.Properties} objects, {@link
```

sdk/src/main/java/io/opentelemetry/sdk/trace/config/package-info.java
```diff
@@ -0,0 +1,74 @@
...
+ * <ul>
+ *   <li>{@code SAMPLER: Samplers.alwaysOn()}
+ *   <li>{@code SPAN_MAX_NUM_ATTRIBUTES: 32}
+ *   <li>{@code SPAN_MAX_NUM_EVENTS: 128}
+ *   <li>{@code SPAN_MAX_NUM_LINKS: 32}
+ *   <li>{@code SPAN_MAX_NUM_ATTRIBUTES_PER_EVENT: 32}
+ *   <li>{@code SPAN_MAX_NUM_ATTRIBUTES_PER_LINK: 32}
+ * </ul>
+ *
+ * <p>Values for {@link io.opentelemetry.sdk.trace.config.TraceConfig} can be read from system
```
thisthat, line 37:
```suggestion
 * <p>Configuration options for {@link io.opentelemetry.sdk.trace.config.TraceConfig} can be read from system
```

sdk/src/main/java/io/opentelemetry/sdk/trace/config/package-info.java
```diff
@@ -0,0 +1,74 @@
...
+ *       io.opentelemetry.trace.Event}s per {@link io.opentelemetry.trace.Span}.
+ *   <li>{@code otel.config.max.links}: to set the global default max number of {@link
+ *       io.opentelemetry.trace.Link} entries per {@link io.opentelemetry.trace.Span}.
+ *   <li>{@code otel.config.max.event.attrs}: to set the global default max number of attributes per
+ *       {@link io.opentelemetry.trace.Event}.
+ *   <li>{@code otel.config.max.link.attrs}: to set the global default max number of attributes per
+ *       {@link io.opentelemetry.trace.Link}.
+ * </ul>
+ *
+ * <p>For Environment Variable, {@link io.opentelemetry.sdk.trace.config.TraceConfig} will look for
```
thisthat, line 57:
```suggestion
 * <p>For environment variable, {@link io.opentelemetry.sdk.trace.config.TraceConfig} will look for
```

RashmiRam:
@thisthat It would be great if you could let me know when this will be released. TIA

sdk/src/main/java/io/opentelemetry/sdk/common/export/ConfigBuilder.java
```diff
@@ -139,4 +139,21 @@ protected static Long getLongProperty(String name, Map<String, String> map) {
...
+  /**
+   * Get a double property from the map, {@code null} if it cannot be found or it has a wrong type.
+   *
+   * @param name The property name
+   * @param map The map where to look for the property
+   * @return the {@link Double} value of the property, {@code null} in case of error or if the
+   *     property cannot be found.
+   */
+  @Nullable
+  protected static Double getDoubleProperty(String name, Map<String, String> map) {
```
jkwatson, line 152: We should have a unit test for this method (and for this whole class).

sdk/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
```diff
@@ -48,25 +89,17 @@
...
+   * Returns a new {@link TraceConfig} reading the configuration values from the environment and
+   * from system properties. System properties override values defined in the environment. If a
+   * configuration value is missing, it uses the default value.
    *
    * @return the default {@code TraceConfig}.
    * @since 0.1.0
    */
   public static TraceConfig getDefault() {
-    return DEFAULT;
+    return newBuilder().readEnvironmentVariables().readSystemProperties().build();
```
carlosalberto, line 100: Usually `getDefault()` methods return a a static, read-only value. I wonder if we should have the same here, or else have this method use a different name.

carlosalberto: Also: do we really want to use env values as default here? I know that for `Resources` we do environment-variables fetching by default, and I'm wondering if we should do it here as well.

carlosalberto:
Thanks for this PR @RashmiRam ! Overall great, although I'm a little bit unsure about the default values, as mentioned in the comment (hope other reviewers chime in ;) )

sdk/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
jkwatson: Agreed. A method called 'getDefault' should be returning a trace config with the default settings, not returning one that has been configured by the user.

RashmiRam: Yes. Agreed. `getDefault` is less intuitive. If `getDefault` returns trace config with the default settings, anybody who requires the config to be read from either env vars or sys props should again use `updateActiveTraceConfig` as default TraceConfig instance is assigned here. https://github.com/open-telemetry/opentelemetry-java/blob/master/sdk/src/main/java/io/opentelemetry/sdk/trace/TracerSharedState.java#L35

My proposal is to keep the `getDefault` similar to how we do for `Resources` where we fetch env vars by default or if `getDefault` is less intuitive, we can rename it as simply `getInstance`. what do you think?

jkwatson: I would argue the other way. I think we should change the method on Resources to make it clear it's being loaded from the environment, rather than perpetuate the confusing naming to more places.

sdk/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
```diff
@@ -138,6 +255,28 @@ private static Builder newBuilder() {
      */
     public abstract Builder setSampler(Sampler sampler);
 
+    /**
+     * Sets the global default {@code Sampler}. It must be not {@code null} otherwise {@link
+     * #build()} will throw an exception.
+     *
+     * @param samplerProbability the global default {@code Sampler}.
```
jkwatson, line 262: this javadoc is wrong.

sdk/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
carlosalberto: +1 on changing `Resources` to make it clear we are loading stuff from the environment.

RashmiRam: Thanks for the reply. There are 2 things to work on.

- Shall we have one more method `getInstance` to fetch settings from env vars/sys props to use it across resources?
- Is `getDefault` still required? shall we remove/deprecate this? Because, `getInstance` will provide trace config with default settings only in the absence of env vars and sys props.

Please let me know how to proceed further.

sdk/src/main/java/io/opentelemetry/sdk/common/export/ConfigBuilder.java
RashmiRam: Done

sdk/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
RashmiRam: Done

sdk/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
jkwatson: For the moment, revert the implementation of `getDefault` to what it was before. If we want to provide additional methods that bypass the builder, we can make that decision separately, and apply it across the board to all the configurable items in the SDK.

RashmiRam: Done.

carlosalberto:
@RashmiRam Please rebase/merge and fix the conflicts ;)

RashmiRam:
@carlosalberto I am not sure why the built checks are in waiting status for so long. Could you please help?

RashmiRam:
@carlosalberto I did a test commit and the checks seem to be passed. Is there anything else that I could help to get this merged?

PR #1256, opened 2020-05-20
Add public Samplers.emptyDecision(boolean) and decision(boolean, Map).

Oberon00:
This is meant for use by custom Sampler implementations. As of now, you have to use an anonymous inner class or otherwise derive a new class because there is no publicly accessible implementation of Decision.

### DIFF
--- a/sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
@@ -21,6 +21,7 @@ import static java.util.Collections.singletonMap;
 
 import com.google.auto.value.AutoValue;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableMap;
 import io.opentelemetry.common.AttributeValue;
 import io.opentelemetry.sdk.trace.Sampler.Decision;
 import io.opentelemetry.trace.Link;
@@ -33,6 +34,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
@@ -75,16 +77,16 @@ public final class Samplers {
    *
    * @param isSampled The value to return from {@link Decision#isSampled()}.
    * @param attributes The attributes to return from {@link Decision#getAttributes()}. A different
-   *     object instance with the same elements may be returned. The map must not be modified after
-   *     being passed to this function. Use {@link Collections#emptyMap()} for an empty decision.
+   *     object instance with the same elements may be returned.
    * @return A {@link Decision} with the attributes equivalent to {@code attributes} and {@link
    *     Decision#isSampled()} returning {@code isSampled}.
    */
-  public static final Decision decision(boolean isSampled, Map<String, AttributeValue> attributes) {
+  public static final Decision decision(
+      boolean isSampled, @Nonnull Map<String, AttributeValue> attributes) {
     Objects.requireNonNull(attributes, "attributes");
     return attributes.isEmpty()
         ? emptyDecision(isSampled)
-        : DecisionImpl.create(isSampled, attributes);
+        : DecisionImpl.create(isSampled, ImmutableMap.copyOf(attributes));
   }
 
   /**
--- a/sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
@@ -74,9 +74,9 @@ public final class Samplers {
    *     value of the parameter {@code isSampled}.
    */
   public static final Decision emptyDecision(boolean isSampled) {
-    return isSampled ? ALWAYS_ON_DECISION : ALWAYS_OFF_DECISION;
+    return isSampled ? EMPTY_SAMPLED_DECISION : EMPTY_NOT_SAMPLED_DECISION;
   }
-  
+
   /**
    * Returns a {@link Sampler} that always makes a "yes" decision on {@link Span} sampling.
    *
--- a/sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
@@ -32,6 +32,7 @@ import io.opentelemetry.trace.attributes.DoubleAttributeSetter;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
```diff
@@ -51,6 +51,20 @@
...
+   * value of the parameter {@code isSampled}.
+   *
+   * <p>This is meant for use by custom {@link Sampler} implementations.
+   *
+   * @param isSampled The value to return from {@link Decision#isSampled()}.
+   * @return A {@link Decision} with empty attributes and {@link Decision#isSampled()} returning the
+   *     value of the parameter {@code isSampled}.
+   */
+  public static final Decision emptyDecision(boolean isSampled) {
+    return isSampled ? ALWAYS_ON_DECISION : ALWAYS_OFF_DECISION;
```
thisthat, line 65: I would expect custom `Sampler`s to use something like:
```
Samplers.alwaysOn/Off().shouldSample(...)
```
so attributes can be set

Oberon00: I don't understand. Which attributes could be set?

thisthat: That depends on the implementation. For example, jaeger adds their rate limiter to the attributes

Oberon00: Of course, for samplers that needs to add attribute, this function cannot be used.
But at least for the negative decision, attributes will AFAIK be ignored, so at least for this case it could be used by any sampler.

Probably a more full-featured version of this PR could also add `singleAttributeDecision(boolean, key, value)` and `attributedDecision(boolean, Map)`. But for the sampler I'm thinking of I would only need `emptyDecision`.

Oberon00: About your suggestion with `Samplers.alwaysOn/Off().shouldSample(...)`:

First, it is very verbose and IMHO also not as clear. Second, if you need to do something like `someBoolean ? Samplers.alwaysOn... : Samplers.alwaysOff...` it becomes even twice as verbose.

Oberon00: And I still don't understand how attributes could be added using `Samplers.alwaysOn/Off().shouldSample(...)`

bogdandrutu:
Let's merge https://github.com/open-telemetry/opentelemetry-java/pull/1233 before this

Oberon00:
#1233 is merged and I rebased the PR to fix the build and add a unit test. The Java 11 failure is unrelated to my PR, it's the `JaegerRemoteSamplerTest`: `java.lang.AssertionError: Resource ManagedChannelOrphanWrapper{delegate=ManagedChannelImpl{logId=9, target=directaddress:///my-service}} can not be released in time at the end of test`

sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
```diff
@@ -63,6 +63,20 @@
...
+   * Returns a {@link Decision} with empty attributes and {@link Decision#isSampled()} returning the
+   * value of the parameter {@code isSampled}.
+   *
+   * <p>This is meant for use by custom {@link Sampler} implementations.
+   *
+   * @param isSampled The value to return from {@link Decision#isSampled()}.
+   * @return A {@link Decision} with empty attributes and {@link Decision#isSampled()} returning the
+   *     value of the parameter {@code isSampled}.
+   */
+  public static final Decision emptyDecision(boolean isSampled) {
```
bogdandrutu, line 76: What about having just a `decision(bool, map)` and you would call that with a `emptyMap`?

```
public static final Decision decision(boolean isSampled, Map<String, AttributeValue> attributes) {
  Objects.requireNotNull(attributes, "attributes");
  if (attributes.isEmpty()) {
    return isSampled ? EMPTY_SAMPLED_DECISION : EMPTY_NOT_SAMPLED_DECISION;
  }
  return DecisionImpl.create(isSampled, attributes);
}
```

Oberon00: Ah, the DecisionImpl was not there when I created the PR I guess 😃  That totally makes sense.

Oberon00: I added `Samplers.decision` but decided to keep `emptyDecision`, since using the explicit empty map is quite a mouthful (CheckStyle even requires a parameter comment for the sampled boolean) and it is needed in the implementation anyway.

Oberon00:
After this, I wonder if we should even have an interface for Decision, or if the DecisionImpl class renamed to Decision would be enough.

sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
bogdandrutu: What about calling this `decision` as well?

Oberon00: I'm fine with either way. Would you prefer that?

sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
```diff
@@ -63,6 +64,45 @@
...
+   *
+   * <p>This is meant for use by custom {@link Sampler} implementations.
+   *
+   * <p>Using {@link #emptyDecision(boolean)} instead of this method is slightly faster and shorter
+   * if you don't need attributes.
+   *
+   * @param isSampled The value to return from {@link Decision#isSampled()}.
+   * @param attributes The attributes to return from {@link Decision#getAttributes()}. A different
+   *     object instance with the same elements may be returned. The map must not be modified after
+   *     being passed to this function. Use {@link Collections#emptyMap()} for an empty decision.
```
jkwatson, line 79: Just having this note in here, rather than copying the attributes, feels too weak. If this is truly a hard requirement, the API should protect itself, rather than relying on implementors to read this javadoc.

Oberon00: First, I think this is an SDK-only API that will only be used by sampler implementations. An "advanced" API so to speak. If I knew that this function was uselessly copying my map, I would rather write my own replacement for efficiency.

Second, in the currently only usage of the Sampler the attributes are actually copied into the span right-away, so I think ignored this note won't actually do any harm.

jkwatson: Then the word "must" here seems overly strong, and possibly misleading. Can you update the javadoc to reflect the reality of the situation?

Oberon00: Hmm, I wouldn't place too much value on the second point, as the usage of the sampler may change in the future. So anyone who reads the javadoc certainly MUST NOT modify this later.

jkwatson: All I'm saying is: If this is an important constraint, then we should either a) enforce it via code or b) clearly explain what the ramifications of violating it are, and not just mention it offhand in attribute documentation.

Oberon00: Unfortunately I think a) is not possible without losing efficiency. And I think with b) we would over-constrain ourselves regarding future changes.

jkwatson: How about adding a line to the javadoc, not in the attributes section like this:

```
You should not modify the contents of the attributes after passing them to this function. If you modify the contents of the attributes after passing them to this function, behavior is not well defined.
```

bogdandrutu: > a) is not possible without losing efficiency.

Confused about this, you should cache the result if performance is important, otherwise it is good to have your own implementation of the `Decision` where you ensure the immutable part differently.

Having an API that relies on these kind of comments is not safe and we cannot guarantee correctness (we can NLP or other exceptions if the user changes the map while we are reading it).

Oberon00: > you should cache the result if performance is important

Good point actually. I think you convinced me that it is better to copy the attributes after all. I will add this to `Decision.create`.

bogdandrutu: Waiting for this :)

Oberon00: Done! I used `ImmutableMap.copyOf` and also fixed the javadoc.

PR #1257, opened 2020-05-20
Add tests to verify immutability of the SpanData representation

jkwatson:
And, remove an extra wrapping of links which wasn't needed.

### DIFF
--- a/sdk/src/main/java/io/opentelemetry/sdk/trace/SpanWrapper.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/trace/SpanWrapper.java
@@ -59,7 +59,7 @@ abstract class SpanWrapper implements SpanData {
       Status status) {
     return new AutoValue_SpanWrapper(
         delegate,
-        Collections.unmodifiableList(links),
+        links, // note: the caller has already made this unmodifiable
         Collections.unmodifiableList(events),
         Collections.unmodifiableMap(attributes),
         totalAttributeCount,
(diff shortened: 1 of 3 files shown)


sdk/src/main/java/io/opentelemetry/sdk/trace/SpanWrapper.java
```diff
@@ -59,7 +59,7 @@ static SpanWrapper create(
       Status status) {
     return new AutoValue_SpanWrapper(
         delegate,
-        Collections.unmodifiableList(links),
+        links, // note: the caller has already made this unmodifiable
         Collections.unmodifiableList(events),
         Collections.unmodifiableMap(attributes),
```
bogdandrutu, line 64: Let's be consistent and do the wrapping in one place.

jkwatson: How do feel about using guava's ImmutableList here?

jkwatson: Hm. It's also true that, currently, if the span hasn't ended, the attributes will still be modified by the running span, making the unmodifiable wrapper not really immutable. The only way I know to solve this is to copy the Map's contents, rather than just wrapping.

bogdandrutu: That was the bug that I mentioned in the previous PR. You need to have a

```java
private HashMap<> getImmutableAttributes() {
  if (hasEnded) {
    return Collections.unmodifiableMap(attributes)
  }
  return Collections.unmodifiableMap(new HashMap<>(attributes))
}

```

jkwatson: ok. I added a test for this particular case, and made it pass.

jkwatson: (and moved the unmodifiable wrapping to all be in the RERS)

jkwatson:
BTW, I ran the benchmark on this, and it came back slightly better than before, although I suspect that's noise, as I ran the benchmark on my laptop.

PR #1259, opened 2020-05-21
Copy map attributes always to avoid unexpected changes

bogdandrutu:


sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -387,24 +397,24 @@ public void addEvent(io.opentelemetry.trace.Event event, long timestamp) {
     addTimedEvent(TimedEvent.create(timestamp, event));
   }
 
-  private Map<String, AttributeValue> limitEventAttributes(Map<String, AttributeValue> attributes) {
-    if (attributes.size() <= this.maxNumberOfAttributesPerEvent) {
-      return attributes;
+  static Map<String, AttributeValue> copyAndLimitAttributes(
```
jkwatson, line 400: Since this method is also used elsewhere, I'd rather see it moved to a separate utility class, rather than leave it in here.

jkwatson:
Just one little change requested. And, looks like this needs a rebase.

sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
bogdandrutu: Can I do it in a separate PR?

jkwatson: yessir!

PR #1262, opened 2020-05-21
Add overloads for setParent(Context context).

carlosalberto:
This way a `Context` can be used such as:

```java
Context ctx = extractContext(...);
spanBuilder.setParent(ctx); // No need to call TracingContextUtils.getSpan()
```


api/src/main/java/io/opentelemetry/correlationcontext/CorrelationContext.java
```diff
@@ -74,6 +75,28 @@
      */
     Builder setParent(CorrelationContext parent);
 
+    /**
+     * Sets the parent {@link CorrelationContext} to use from the specified {@code Context}. If no
+     * parent is provided, the value of {@link CorrelationContextManager#getCurrentContext()} at
```
jkwatson, line 80: maybe "If no parent context is available in the provided Context, the value of ...."

api/src/main/java/io/opentelemetry/correlationcontext/CorrelationContext.java
```diff
@@ -74,6 +75,28 @@
...
+     * <p>This <b>must</b> be used to create a {@link CorrelationContext} when manual Context
+     * propagation is used.
+     *
+     * <p>If called multiple times, only the last specified value will be used.
+     *
+     * @param context the {@code Context}.
+     * @return this.
+     * @throws NullPointerException if {@code context} is {@code null}.
+     * @see #setNoParent()
+     * @since 0.4.0
```
jkwatson, line 96: 0.5.0

jkwatson:
My main concern with this is that we're making the grpc context even more deeply embedded in the APIs. I think having the helper methods is very useful, but I'm trying to be sensitive to the desire to not expose the grpc context in our APIs.

carlosalberto:
> My main concern with this is that we're making the grpc context even more deeply embedded in the APIs. 

Same feeling. That's why I hadn't worked on this before getting an agreement on the Context-Prop part (that and other work items ;) ). But in any case, I'd like to get things on the discussion table, for a start ;)

jkwatson:
@carlosalberto do you want to move forward with this, even with the expectation that we'll need to revisit when we deal with grpc-context, pre GA?

carlosalberto:
@jkwatson Yes, definitely. Had a small chat with Yuri and he mentioned he would prefer us to keep the explicit way. If/when we change `Context`, we can update this overloads.

So I will massage with code this week in hopes of having it merged ;)

jkwatson:
@carlosalberto any update on this one?

iNikem:
@carlosalberto friendly ping from me as well :) I am contemplating doing some experiments in instrumentation repo and I need this overload.

carlosalberto:
Oh, my bad, totally forgot about this one. Will apply the feedback and hopefully we will be able to roll it soon ;)

carlosalberto:
@open-telemetry/java-approvers Updated this, for your consideration ;)

api/src/main/java/io/opentelemetry/correlationcontext/CorrelationContext.java
jkwatson: @carlosalberto thoughts on my suggestion?

api/src/main/java/io/opentelemetry/correlationcontext/CorrelationContext.java
jkwatson: 0.7.0 now

api/src/main/java/io/opentelemetry/trace/Span.java
```diff
@@ -406,6 +407,23 @@
...
+     * <p>If no {@link Span} is available in the specified {@code Context}, the resulting {@code
+     * Span} will become a root instance, as if {@link #setNoParent()} had been called.
+     *
+     * <p>If called multiple times, only the last specified value will be used. Observe that the
+     * state defined by a previous call to {@link #setNoParent()} will be discarded.
+     *
+     * @param context the {@code Context}.
+     * @return this.
+     * @throws NullPointerException if {@code context} is {@code null}.
+     * @since 0.5.0
```
jkwatson: 0.7.0

carlosalberto: I think there was a race condition between me pushing the changes and your review ;) In any case, all issues have been addressed.

PR #1264, opened 2020-05-21
Add a few missing unit tests

jkwatson:
In this case, for the EnvVarResource and the SimpleSpanProcessor


jkwatson:
And, hey, gets us above 90% on the coverage check!

PR #1266, opened 2020-05-22
Do not allow adding attributes, links after startSpan.

### LINKED ISSUE #1197: It is possible to add a link to a span after it has been built.
**Describe the bug**
If you hold on to the SpanBuilder used to create the span, you can continue to add links after the span has started.

**Steps to reproduce**
```
        Span.Builder spanBuilder = tracerSdk.spanBuilder(SPAN_NAME);
    spanBuilder.addLink(DefaultSpan.getInvalid().getContext());

    //this should lock the links in place
    RecordEventsReadableSpan span = (RecordEventsReadableSpan) spanBuilder.startSpan();
    List<SpanData.Link> originalLinks = span.toSpanData().getLinks();
    int numberOfLinksBeforeAfterStart = originalLinks.size();

    //add another link to the builder
    spanBuilder.addLink(DefaultSpan.getInvalid().getContext());

    //this number is correct
    assertThat(numberOfLinksBeforeAfterStart).isEqualTo(span.toSpanData().getTotalRecordedLinks());
    //but, we actually added a link to the underlying collection, via the builder.
    assertThat(numberOfLinksBeforeAfterStart).isEqualTo(span.toSpanData().getLinks().size());
```

### LINKED ISSUE #1199: You can add attributes to a started span via the builder that created it
**Describe the bug**
When we start a span, we should make a copy of the attributes from the SpanBuilder, but instead we just use a reference to the Builder's collection. This means you can continue to add attributes via the Builder.

**Steps to reproduce**
```
    Span.Builder spanBuilder = tracerSdk.spanBuilder(SPAN_NAME);
    spanBuilder.setAttribute("attr1", "value1");

    // this should make a copy of the attributes from the builder
    RecordEventsReadableSpan span = (RecordEventsReadableSpan) spanBuilder.startSpan();
    Map<String, AttributeValue> startingAttributes = span.toSpanData().getAttributes();

    // add another link to the *builder*
    spanBuilder.setAttribute("attr2", "value2");

    // this number is correct
    assertThat(startingAttributes.size()).isEqualTo(1);
    // but, we actually added an attribute to the underlying collection, via the builder.
    assertThat(span.toSpanData().getAttributes().size()).isEqualTo(1);
```

bogdandrutu:
Fixes #1199 
Fixes #1197

### DIFF
--- a/sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
@@ -84,7 +84,8 @@ final class RecordEventsReadableSpan implements ReadableSpan, Span {
   private final long startEpochNanos;
   // Set of recorded attributes. DO NOT CALL any other method that changes the ordering of events.
   @GuardedBy("lock")
-  private final AttributesMap attributes;
+  @Nullable
+  private AttributesMap attributes;
   // List of recorded events.
   @GuardedBy("lock")
   private final EvictingQueue<TimedEvent> events;
@@ -113,7 +114,7 @@ final class RecordEventsReadableSpan implements ReadableSpan, Span {
       SpanProcessor spanProcessor,
       Clock clock,
       Resource resource,
-      AttributesMap attributes,
+      @Nullable AttributesMap attributes,
       List<io.opentelemetry.trace.Link> links,
       int totalRecordedLinks,
       long startEpochNanos) {
@@ -130,7 +131,6 @@ final class RecordEventsReadableSpan implements ReadableSpan, Span {
     this.hasEnded = false;
     this.clock = clock;
     this.startEpochNanos = startEpochNanos;
-    // TODO: Do not always initialize the attributes.
     this.attributes = attributes;
     this.events = EvictingQueue.create(traceConfig.getMaxNumberOfEvents());
     this.traceConfig = traceConfig;
@@ -199,7 +199,7 @@ final class RecordEventsReadableSpan implements ReadableSpan, Span {
           getImmutableLinks(),
           getImmutableTimedEvents(),
           getImmutableAttributes(),
-          attributes.getTotalAddedValues(),
+          (attributes == null) ? 0 : attributes.getTotalAddedValues(),
           totalRecordedEvents,
           getStatusWithDefault());
     }
@@ -304,8 +304,14 @@ final class RecordEventsReadableSpan implements ReadableSpan, Span {
         return;
       }
       if (value == null || (value.getType().equals(STRING) && value.getStringValue() == null)) {
+        if (attributes == null) {
+          return;
+        }
         attributes.remove(key);
         return;
+      }
+      if (attributes == null) {
+        attributes = new AttributesMap(traceConfig.getMaxNumberOfAttributes());
       }
       attributes.put(key, value);
     }
@@ -510,7 +516,7 @@ final class RecordEventsReadableSpan implements ReadableSpan, Span {
 
   @GuardedBy("lock")
   private Map<String, AttributeValue> getImmutableAttributes() {
-    if (attributes.isEmpty()) {
+    if (attributes == null || attributes.isEmpty()) {
       return Collections.emptyMap();
     }
     return hasEnded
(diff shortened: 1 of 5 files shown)


sdk/src/main/java/io/opentelemetry/sdk/trace/SpanBuilderSdk.java
```diff
@@ -176,10 +173,16 @@
     Objects.requireNonNull(key, "key");
     if (value == null
         || (value.getType() == AttributeValue.Type.STRING && value.getStringValue() == null)) {
+      if (attributes == null) {
+        return this;
+      }
       attributes.remove(key);
-    } else {
-      attributes.put(key, value);
+      return this;
```
arminru, line 180: Could be simplified to `if null { remove } return;`

sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -84,7 +84,8 @@
   private final long startEpochNanos;
   // Set of recorded attributes. DO NOT CALL any other method that changes the ordering of events.
   @GuardedBy("lock")
-  private final AttributesMap attributes;
+  @Nullable
+  private AttributesMap attributes;
```
arminru, line 88: Why are `attributes` nullable but `links` aren't?

sdk/src/test/java/io/opentelemetry/sdk/trace/SpanBuilderSdkTest.java
```diff
@@ -177,6 +177,24 @@ public void changingAttributes_NoEffectAfterAddLink() {
...
+  @Test
+  public void addLink_NoEffectAfterStartSpan() {
+    Span.Builder spanBuilder = tracerSdk.spanBuilder(SPAN_NAME);
+    spanBuilder.addLink(sampledSpanContext);
+    RecordEventsReadableSpan span = (RecordEventsReadableSpan) spanBuilder.startSpan();
+    try {
+      assertThat(span.toSpanData().getLinks())
+          .containsExactly(
+              Link.create(sampledSpanContext, Collections.<String, AttributeValue>emptyMap()));
+      spanBuilder.addLink(sampledSpanContext);
```
arminru, line 189: You might want to use a different SpanContext here, otherwise the missing effect could also come from a check that links are not duplicated (or a set being used internally).

sdk/src/test/java/io/opentelemetry/sdk/trace/SpanBuilderSdkTest.java
```diff
@@ -384,6 +431,27 @@ public void droppingAttributes() {
     }
   }
 
+  @Test
+  public void addAttributes_OnlyViaSampler() {
```
arminru, line 435: 👍 for covering this!

sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
bogdandrutu: Links is a final argument because the `Span` does not have an API to add links (only the builder) so it is fine to just use emptyList (which is immutable) if no links were added to avoid all null checks. In the case of attributes if I use emptyMap does not give me any benefit because that is immutable so I still need to have all checks before adding elements to the map.

sdk/src/main/java/io/opentelemetry/sdk/trace/SpanBuilderSdk.java
bogdandrutu: Done.

sdk/src/test/java/io/opentelemetry/sdk/trace/SpanBuilderSdkTest.java
bogdandrutu: You are welcome :)

sdk/src/test/java/io/opentelemetry/sdk/trace/SpanBuilderSdkTest.java
bogdandrutu: Done, also added a comment to explain.

sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
arminru: Thanks for clarifying!

PR #1268, opened 2020-05-23
Expose RandomIdsGenerator so it can be delegated to.

anuraaga:
When creating a custom `IdsGenerator`, I think it is common to want to delegate to the default random one, e.g., only customize trace ID and use the same code for span ID. I noticed this when writing https://github.com/aws-samples/aws-xray-sdk-with-opentelemetry-sample/blob/both-otel-and-xray/awsagentprovider/src/main/java/com/softwareaws/xray/opentelemetry/exporters/AwsXrayIdsGenerator.java#L29

The `AwsXrayIdsGenerator` in this repo has a huge dependency on the xray SDK which I want to remove (then I don't need to fork there :) ), and this would help when doing that.

~Also made `RandomIdsGenerator` a singleton while here.~

Realized it's more useful to be able to customize Random than make singleton.

### DIFF
--- a/sdk/src/test/java/io/opentelemetry/sdk/trace/RandomIdsGeneratorTest.java
+++ b/sdk/src/test/java/io/opentelemetry/sdk/trace/RandomIdsGeneratorTest.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2020, OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package io.opentelemetry.sdk.trace;
 
 import static com.google.common.truth.Truth.assertThat;
--- a/sdk/src/test/java/io/opentelemetry/sdk/trace/RandomIdsGeneratorTest.java
+++ b/sdk/src/test/java/io/opentelemetry/sdk/trace/RandomIdsGeneratorTest.java
@@ -1,0 +1,80 @@
+package io.opentelemetry.sdk.trace;
+
+import static com.google.common.truth.Truth.assertThat;
+import static org.mockito.Mockito.when;
+
+import java.util.Random;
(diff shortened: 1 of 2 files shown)


anuraaga:
Thanks - in the file I linked to, where we generate IDs for use with X-Ray, you'll see that all the code here has been copied. In particular, there isn't any desire to tweak the logic for the span ID so it would be nice to not have to copy the logic. Some thoughts that came to mind while writing this PR

1) When an SDK provides an interface for extension, it's very common to expose a default implementation as well, to allow for small customization rather than having to replace it completely. In Java 8, it's a bit nicer to be able to keep the class hidden and add a factory method to the interface, but with Java 7 I think there isn't much better than exposing the class.

2) When logic accepts `Random`, it is very common to accept a different type of `Random`. This is often for unit testing, but also allows swapping the `Random` with e.g., `SecureRandom`. In tracing, users may propagate traces to untrusted sources and want more random IDs to prevent reverse engineering request rate, for example.

An alternative is to just copy the code, it's true that it's not so much. I can remove the `RandomSupplier` (along with the unit tests it enables) to reduce a lot of the complexity if that's a better balance, or just close the PR in favor of more copy paste in ID generator implementations, let me know.

bogdandrutu:
@anuraaga thanks for the detailed explanation. I would initially do only the change to expose the class and maybe keep the "default" test you added (without the supplier). That should achieve your first goal and we keep the exposed api minimal.

Does that make sense?

anuraaga:
Sure - simplified to only exposing the class

sdk/src/test/java/io/opentelemetry/sdk/trace/RandomIdsGeneratorTest.java
```diff
@@ -0,0 +1,41 @@
...
+  @Test
+  public void defaults() {
+    RandomIdsGenerator generator = new RandomIdsGenerator();
+
+    // Can't assert values but can assert they're valid.
+    TraceId traceId = generator.generateTraceId();
+    assertThat(traceId).isNotEqualTo(TraceId.getInvalid());
+
+    SpanId spanId = generator.generateSpanId();
+    assertThat(spanId).isNotEqualTo(SpanId.getInvalid());
```
bogdandrutu, line 39: Maybe worth generating 1K ids to check?

bogdandrutu:
Please fix `Execution failed for task ':opentelemetry-sdk:verifyGoogleJavaFormat'.` use `./gradlew goJF`

sdk/src/test/java/io/opentelemetry/sdk/trace/RandomIdsGeneratorTest.java
anuraaga: Sure - for the record, even with 1K test only takes 41ms on my macbook pro

PR #1273, opened 2020-05-25
Document span context extraction precedence in multi format

### LINKED ISSUE #1271: Document precedence order is multiple context propagators are used
```java
    ContextPropagators contextPropagators = DefaultContextPropagators.builder()
        .addHttpTextFormat(new JaegerPropagator())
        .addHttpTextFormat(new HttpTraceContext())
        .build();
    OpenTelemetry.setPropagators(contextPropagators);
```

The `DefaultContextPropagators.Builder` should document precedence order when multiple context propagation formats are used - especially for extraction.  At the moment all formats are extracted and the last one is returned.

https://github.com/open-telemetry/opentelemetry-java/blob/690651b65a938c903c35d65aef3182999fcb1b19/context_prop/src/main/java/io/opentelemetry/context/propagation/DefaultContextPropagators.java#L72

pavolloffay:
Resolves #1271

Signed-off-by: Pavol Loffay <ploffay@redhat.com>

### DIFF
--- a/context_prop/src/main/java/io/opentelemetry/context/propagation/DefaultContextPropagators.java
+++ b/context_prop/src/main/java/io/opentelemetry/context/propagation/DefaultContextPropagators.java
@@ -67,6 +67,9 @@ public final class DefaultContextPropagators implements ContextPropagators {
    *     .build();
    * }</pre>
    *
+   * <p>The first added {@link HttpTextFormat} has the highest precedence when extracting the
+   * {@code SpanContext}.
+   *
    * @since 0.3.0
    */
   public static final class Builder {
@@ -140,7 +143,7 @@ public final class DefaultContextPropagators implements ContextPropagators {
 
     @Override
     public <C> Context extract(Context context, C carrier, Getter<C> getter) {
-      for (int i = 0; i < textPropagators.length; i++) {
+      for (int i = textPropagators.length - 1; i >= 0; i--) {
         context = textPropagators[i].extract(context, carrier, getter);
       }
       return context;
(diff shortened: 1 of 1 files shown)


trask:
Just linking a couple of related issues:

https://github.com/open-telemetry/opentelemetry-auto-instr-java/issues/425
https://github.com/open-telemetry/opentelemetry-specification/issues/496

carlosalberto:
Not sure we want to do this. While researching on this subject, Specification-wise, I realized that in other languages' implementations the last propagator has the highest precedence. So I'd hold this up for now (probably adjusting the auto-instrumentation part) and revisit it when some work on the Specification around this is done.

pavolloffay:
> I realized that in other languages' implementations the last propagator has the highest precedence.

This is exactly what I found confusing here. We should revisit and make it clear for consumers.

jkwatson:
What is the state of this PR? Should we add the "holding on" label pending spec clarification?

carlosalberto:
I'd be up for documenting it and keep the current behavior (that is, the last one wins). This is how this is implemented in other languages. Any opinion on this @pavolloffay ?

(As I said, I'm working on a few changes to the Specification related to context propagation to alleviate this, but we should definitely document parts like this).

jkwatson:
@pavolloffay the title of this PR is off...it's not just documenting it, it's also changing it. What is your plan with this? If we need to wait for spec updates, I'll tag it as `holding_on`.

jkwatson:
bump @pavolloffay

pavolloffay:
I will close it for now. My understanding from @carlosalberto is that we need to handle it first at the spec level.

PR #1274, opened 2020-05-26
Inline generation of XRay trace IDs to remove heavyweight SDK depende…

anuraaga:
…ncy.

The X-Ray SDK is pretty heavyweight, meant for instrumenting, not something as simple as generating IDs. This implements ID generation without the dependency. Also removes Guava since it was only used once so not really worth it.

In a future PR I'll reimplement `EC2Resource` as well to remove the AWS SDK dependency too.

### DIFF
--- a/sdk_contrib/aws_v1_support/build.gradle
+++ b/sdk_contrib/aws_v1_support/build.gradle
@@ -12,8 +12,8 @@ sourceCompatibility = 1.8
 targetCompatibility = 1.8
 
 dependencies {
-    api project(':opentelemetry-api')
-    api project(':opentelemetry-sdk')
+    api project(':opentelemetry-api'),
+            project(':opentelemetry-sdk')
 
     implementation 'com.amazonaws:aws-java-sdk-core:1.11.701'
     implementation 'com.amazonaws:aws-java-sdk-ec2:1.11.701'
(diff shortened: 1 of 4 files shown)


sdk_contrib/aws_v1_support/build.gradle
```diff
@@ -12,12 +12,9 @@ sourceCompatibility = 1.8
 targetCompatibility = 1.8
 
 dependencies {
-    api project(':opentelemetry-api'),
-            project(':opentelemetry-sdk')
```
bogdandrutu, line 16: This is unnecessary change

sdk_contrib/aws_v1_support/build.gradle
```diff
@@ -12,12 +12,9 @@ sourceCompatibility = 1.8
...
+    api project(':opentelemetry-api')
+    api project(':opentelemetry-sdk')
 
-    implementation libraries.guava
-
-    implementation 'com.amazonaws:aws-xray-recorder-sdk-core:2.4.0'
     implementation 'com.amazonaws:aws-java-sdk-core:1.11.701'
```
bogdandrutu, line 18: Can be done in on implementation statement

anuraaga: Sure - I took the opposite approach of making them consistent but went with one statement. But I'd argue it's better style to use multiple statements, there's a significant readability loss when the dependencies for the same configuration don't line up vertically IMO. Food for thought.

Issue #1258, opened 2020-05-21
Change public APIs to not accept Map<String, AttributeValue>


bogdandrutu:
Proposals:
1. Use guava ImmutableMap - probably want to avoid guava dependency in the API.
2. Add our own class that is immutable instead of using a map.

The main motivation is that because the map is not immutable we have to do a copy of the entire map all the time.

jkwatson:
Take a look at #1130 for one proposal to have an `Attributes` class that could probably fulfill this need (although not the specific implementation in that brainstorm)

bogdandrutu:
@jkwatson looked at that, it does more than this issue require. But maybe that's what we need, still thinking about this.

Also: Bonus for labels we have the same problem since the variadic methods can be called with an array (not sure if the compiler copies the array automatically),  may be good to verify that as well.

jkwatson:
There is no secret array-copying when you call a varargs method directly with an array. It just passes it directly to the method.

bogdandrutu:
@jkwatson which means the current labels API suffer from the same problem :(

jkwatson:
It's true. Definitely something to fix.

jkwatson:
Note: I chatted with @trask this morning in the auto-instrumentation meeting, and he mentioned that the auto-instrumentation project never uses any of the methods that take Maps directly. So, maybe we should start by removing the maps from the public API, and figure out an internal, immutable representation as a separate effort.

bogdandrutu:
We discussed this last week, there are other APIs that use map and we cannot completely get rid of all

jkwatson:
This was completed in the 0.6.0 release with the introduction of the Attributes & Labels classes.

PR #1130, opened 2020-04-22
Work in Progress on some Trace API tweaks

jkwatson:
Doing this in Draft PR to make it as open and public as possible. 

Also, doing this via a new contrib module for incubating new APIs.

PR #1278, opened 2020-05-26
Create an Attributes immutable container to replace Map<String, AttributeValue>

jkwatson:
TODO on this PR:

- [x] Make generic in the value, so can be used for metric labels
- [x] Javadoc
- [ ] Builder vs. creation via Map vs Object... varargs


api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -0,0 +1,294 @@
...
+@Immutable
+public abstract class Attributes {
+  private static final Attributes EMPTY = new EmptyAttributes();
+
+  /** javadoc me. */
+  public abstract Set<String> keys();
+
+  /** javadoc me. */
+  @Nullable
+  public abstract AttributeValue getValue(String key);
```
jkwatson, line 44: If we went with a solution that's approximately this shape, I think we'd need to invent an API to iterate over the key-value pairs efficiently, as this one is not, unless we back the implementation with a Map, which is unfortunately kind of a memory-hog.

bogdandrutu:
Exposing an foreach may be an option. Also maybe making this a generic class can help with labels as well

jkwatson:
> Exposing an foreach may be an option. Also maybe making this a generic class can help with labels as well

Yes, I was thinking of building a custom iterator over the keys that would keep track of the index, for efficient access to the values.

jkwatson:
> Also maybe making this a generic class can help with labels as well

So, generic on the value, so we could just have Strings in it, rather than the full AttributeValue?

bogdandrutu:
@jkwatson yes that's what I meant. Nice link to the brave code. That also remind me that we need to sort the keys and dedup them at least for metrics

jkwatson:
I added a simple iterator that returns Map Entries, which might not be perfect, but it does show that it works.

api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -0,0 +1,358 @@
...
+        Arrays.asList(
+            key1, value1,
+            key2, value2,
+            key3, value3,
+            key4, value4,
+            key5, value5));
+  }
+
+  /** javadoc me. */
+  public static Builder newBuilder() {
```
bogdandrutu, line 138: Do we need a Builder?

api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -0,0 +1,358 @@
...
+
+  /** javadoc me. */
+  public abstract Set<String> keys();
+
+  /** javadoc me. */
+  @Nullable
+  public abstract AttributeValue getValue(String key);
+
+  /** javadoc me. */
+  public static Attributes fromMap(Map<String, AttributeValue> values) {
```
bogdandrutu, line 50: Is this useful?

api/src/main/java/io/opentelemetry/common/Attributes.java
bogdandrutu: Do we need this?

api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -0,0 +1,358 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+/** javadoc me. */
+@Immutable
+public abstract class Attributes implements Iterable<Entry<String, AttributeValue>> {
```
bogdandrutu, line 39: I would go with foreach pattern to avoid allocations.

api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -0,0 +1,358 @@
...
+  private static final Attributes EMPTY = new EmptyAttributes();
+
+  /** javadoc me. */
+  public abstract Set<String> keys();
```
bogdandrutu, line 43: Is this useful?

api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -0,0 +1,358 @@
...
+
+    /** javadoc me. */
+    public Builder addAttribute(String key, Boolean... value) {
+      data.add(key);
+      data.add(AttributeValue.arrayAttributeValue(value));
+      return this;
+    }
+  }
+
+  private static class EmptyAttributes extends Attributes {
```
bogdandrutu, line 268: Would use the enum pattern :)

carlosalberto:
Looks great to me overall. My feedback on this (a little on Bogdan's comments) is to avoid adding members we do not need **now**, staying as minimalistic as possible.

jkwatson:
Before removing all the creational methods (and yes, I agree, we should definitely minimize!), I'd like to solicit from feedback from @trask  and @iNikem about what creational pattern(s) the auto-instrumentation project would prefer.

And, we should limit the data-access methods to just those needed for the SDK to operate for now (and perhaps to facilitate testing, if it gets painful to get access to the data to verify it's correct).

api/src/main/java/io/opentelemetry/common/Attributes.java
jkwatson: I think we either need the Builder, or we need a way to pass in the Map. Otherwise, there won't be a way to add an arbitrary number of keys. I'm not sure which is best.

api/src/main/java/io/opentelemetry/common/Attributes.java
jkwatson: I can imagine situations where this is useful, but if we don't think we need it in the SDK (for metrics, maybe?), then I'm happy to remove any and all methods that aren't absolutely necessary for now.

api/src/main/java/io/opentelemetry/common/Attributes.java
jkwatson: I've been struggling with how to provide this without making some sort of wrapper around the key-value pairs. Do you have an API in mind? Maybe I'm not understanding with you mean by a "foreach" pattern here. I was assuming you wanted to be able to do a java "foreach loop" which looks like this:

```
for(Thing value : attributes) {
}
```

But, what is the "Thing" here?

api/src/main/java/io/opentelemetry/common/Attributes.java
jkwatson: removed

api/src/main/java/io/opentelemetry/common/Attributes.java
jkwatson: removed

api/src/main/java/io/opentelemetry/common/Attributes.java
jkwatson: Hm. In this case, since I made Attributes an abstract class (rather than an interface, so I could add static methods to it for java7), I can't use an enum, because they can't extend classes.

trask:
As you mentioned https://github.com/open-telemetry/opentelemetry-java/issues/1258#issuecomment-635488676, my feedback is that we haven't had a use for the setters that take a `Map`, so I'm wondering if this is an important use case that we need to support?

I think the immutability problem becomes easier if we don't expose this `Map` setter, because then we can track the key/value pairs as a list of alternating `String`/`AttributeValue` values inside `Span.Builder` and build the map in `startSpan()`.

iNikem:
From auto-instr POV I don't see any need in providing attributes in a bulk. Unless it will be significantly more performant than setting attributes one by one

api/src/main/java/io/opentelemetry/common/Attributes.java
jkwatson: hmm.  maybe a java 8-style `foreach` method that takes a `BiConsumer<String, AttributeValue>` (or equivalent). A little clunky to use in java 7, but very nice in java 8+!

anuraaga:
Personally would prefer `Iterable` to the consumer since it is more flexible - it supports `forEach` but also normal for loop and can be easily converted to a stream for even more API options. Working with entries isn't ideal but the added flexibility without much API surface seems like a better bang for buck.

anuraaga:
Realized the consumer pattern is to avoid allocation of the `Entry` objects. Makes sense, if the iterator seems useful it can be added later.

jkwatson:
> Realized the consumer pattern is to avoid allocation of the `Entry` objects. Makes sense, if the iterator seems useful it can be added later.

Yes, that was also my plan. We can always add more ways to get at the data, but for now, keeping it very simple, and favoring fewer allocations.

api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -0,0 +1,166 @@
...
+        Arrays.asList(
+            key1, value1,
+            key2, value2,
+            key3, value3,
+            key4, value4,
+            key5, value5));
+  }
+
+  /** javadoc me. */
+  public static <T> Builder<T> newBuilder() {
```
jkwatson, line 125: @bogdandrutu Any further thoughts on having something other than a Builder for the "lots of attributes" use-case?

jkwatson:
@bogdandrutu I added a varargs method to enable creating an arbitrary number of attributes.  The Builder is still here, and I'll delete whichever one we like less before merging.

Otherwise, I think this is ready for review.

api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -0,0 +1,203 @@
...
+import java.util.Map.Entry;
+import java.util.TreeMap;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * An immutable container for attributes. The type parameter denotes the type of the values of the
+ * attributes.
+ */
+@Immutable
+public abstract class Attributes<T> {
```
anuraaga, line 34: Can't an attribute map have values of different types? Or is that only for resource?

anuraaga: For reference here's the class in the collector, for consistency it seems to make sense to have a similar API with the mutators on a builder and accessors on the object

https://github.com/open-telemetry/opentelemetry-collector/blob/master/consumer/pdata/common.go#L311

jkwatson: The variant types for values are encapsulated by the AttributeValue class, so you can declare `Attributes<AttributeValue>` if you need to have the values vary in type. That will be the most common use-case, but for metric labels, we'll want to use `Attributes<String>`.

Remember, the use case here is API surface, so the shape of the API won't necessarily match that in the collector, which doesn't expose an API surface for instrumentation.

anuraaga: Thanks for sharing the examples. I agree that the API often won't match, but for general data containers I figure the principles of creating an easy-to-use API are shared to a large degree. Anyways the collector is mainly a reference point for a similar API.

Some observations from my own usage of these APIs
- Wrapping all operations with `stringAttributeValue` in the Java SDK is tedious and significantly reduces the readability of the code
- Similar manipulation of attributes in the collector was pleasant, I was very happy to see the type-specific mutators
- Most manipulation in the collector was unit tests. I presume unit tests in the SDK will similarly create these data containers a lot, so I would argue that increases the potential benefit in a user-friendly API

I'd generally recommend imagining what user code looks like when using an API when creating it and working backwards from there. I can imagine these

```java
Attributes attributes = Attributes.newBuilder()
  .putString("key1", "value1")
  .putInt("key2", 3)
  .build();
```

vs

```
Attributes<AttributeValue> attributes = Attributes.of(
  "key1", stringAttributeValue("value1"),
  "key2", intAttributeValue(3));
```

and would argue the contents of this latter are significantly obfuscated by the wrappers.

My suggestion would be that having a generic class as a backing store makes sense, and the actual user API could be type-specific, so `Attributes` and `Labels` or something, first is polymorphic second is only String. Former could wrap in `attributeValue` transparent to the user.

Anyways my 2c and sharing some experience with using the current APIs. Hope it helps.

jkwatson: Some of us have actually have proposed an alternative API for this that I think is a bit more fluent for real-life instrumentation use-cases, and saves some allocations in the normal use as well. You can see some of the ideas over in this draft PR:  #1130. In particular, moving the type safety to the key-side of the API I think makes it easier to pre-declare the attribute keys with wrappers, and allows the values to be non-wrapped.

Personally, I find the AttributeValue to be very cumbersome for the most common use-cases, as you have also stated.

anuraaga: Thanks for sharing - yeah the linked PR's API looks quite nice to me :)

jkwatson:
> @jkwatson there are a lot of unresolved comments, please make sure you "resolve" them if they are fixed.

I resolved everything aside from the Builder vs. other for arbitrary attribute sizes.

api/src/main/java/io/opentelemetry/common/Attributes.java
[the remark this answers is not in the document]
jkwatson: My concern with having it be named `*Map`, is that it doesn't implement the `Map` interface, and we don't really want it to. It might add more confusion to have a class that kind-of looks mappish, but doesn't implement the commonly known interface.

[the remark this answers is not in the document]
jkwatson: If we make it package protected, though, I guess that's less of a concern. I wonder if there's another word than `Map` we can use, though, that describes the functionality more accurately.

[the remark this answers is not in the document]
jkwatson: I do like the idea of the sub-classes, but since the creational patterns here are all based on static methods, we'll need to rethink how that works. Any thoughts on that? Do the subclasses need to contain the static `of` methods and a builder, as well? I'm playing around with ideas on this, but if you have some, please share. :)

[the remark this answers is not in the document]
jkwatson: I'm going to open up a new PR with Labels, Attributes, and an ImmutableKeyValuePairs interface they both implement. I'll keep this open in case we think that other PR is garbage. See #1304

jkwatson:
closing, as we're moving on in separate PRs

Issue #575, opened 2019-09-29
remove grpc-context dependency from the api package


adriancole:
I understand why opentelemetry has a bias towards google's gRPC library. This codebase primarily borrows from work that started internally, became google's instrumentation-java, later renamed to opencensus, and now mostly copy/pasted from that. The original primary customer was not just gRPC protocol, but specifically google's library. For example, gRPC is the most primary user of the prior work, and it is helpful to have a user to guide.

Time has passed, though, and what started as an end-to-end instrumentation library has split into a Api + SDK approach, as facilitated by the weaving in of OpenTracing. Yet, code opinons towards google's gRPC implementation still exist in the form of a strict dependency in the api package, and helper code.

While it is understandable that alliances exists, and having good integration with Google's gRPC implementation is important, even gRPC itself has options. For example, Armeria has gRPC protocol support without the need of Google's library. We have already encountered problems where [other google projects drag in gRPC because of Census](https://github.com/googleapis/google-http-java-client/issues/621). This should stop now. OpenTelemetry as a body of people, sure that can advocate for Google. However, the instrumentation api must not pick winners and losers in RPC frameworks.

Please remove all traces of grpc-context from the api library and move the support functions to a context integration package.

This is a follow-up from here: https://github.com/open-telemetry/opentelemetry-java/pull/571#issuecomment-536193532

adriancole:
here are some related issues from grpc-context:

* 2016 [pulling grpc-context into a different library](https://github.com/grpc/grpc-java/pull/2226), but still in grpc repo
* 2017 [modularity problems master issue](https://github.com/grpc/grpc-java/issues/2727) including Java 9 modules and also OSGi

adriancole:
It may also not be obvious why removing it from the api package is important besides bloat. grpc-context is not even an optional dependency. strict dependencies not only cause bloat and pin versions, but they also obscure what apis are actually powering the software.  This means it hides abstraction leaks. For example, you can't tell if an instrumentation site requires grpc-context or not by looking at its dependencies: you have to scan all of its code.

TL;DR; Currently, anything marked as opentelemetry-api might as well be assumed to only work when grpc-context is present.

```bash
$ ./gradlew :opentelemetry-api:dependencies
--snip--

compileClasspath - Compile classpath for source set 'main'.
+--- com.google.auto.value:auto-value-annotations:1.6.2
+--- com.google.errorprone:error_prone_annotations:2.3.2
[...]

PR #1281, opened 2020-05-28
Apply bom imports only if needed

### LINKED ISSUE #1269: Exploded dependencies for API
Compere the following dependencies reports:

```
io.opentelemetry:opentelemetry-api:0.4.1
|    |    |    \--- io.opentelemetry:opentelemetry-context-prop:0.4.1
|    |    |         \--- io.grpc:grpc-context:1.28.0
```
vs
```
io.opentelemetry:opentelemetry-api:0.5.0-SNAPSHOT
|    |    |    +--- io.grpc:grpc-bom:1.28.0
|    |    |    |    \--- io.grpc:grpc-context:1.28.0 (c)
|    |    |    +--- com.google.guava:guava-bom:28.2-android
|    |    |    |    \--- com.google.guava:guava:28.2-android (c)
|    |    |    +--- com.google.protobuf:protobuf-bom:3.11.4
|    |    |    +--- io.zipkin.reporter2:zipkin-reporter-bom:2.12.2
|    |    |    \--- io.opentelemetry:opentelemetry-context-prop:0.5.0-SNAPSHOT
|    |    |         +--- io.grpc:grpc-bom:1.28.0 (*)
|    |    |         +--- com.google.guava:guava-bom:28.2-android (*)
|    |    |         +--- com.google.protobuf:protobuf-bom:3.11.4
|    |    |         +--- io.zipkin.reporter2:zipkin-reporter-bom:2.12.2
|    |    |         \--- io.grpc:grpc-context -> 1.28.0
```

API version 0.5.0 clearly has brought many more dependencies with it. I really hope this is just a publishing bug.

iNikem:
Fixes #1269

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -180,10 +180,7 @@ configure(opentelemetryProjects) {
 
     dependencies {
 
-        implementation platform(boms.grpc),
-                platform(boms.guava),
-                platform(boms.protobuf),
-                platform(boms.zipkin_reporter)
+        implementation platform(boms.grpc)
 
         compileOnly libraries.auto_value_annotation,
                 libraries.errorprone_annotation,
@@ -192,7 +189,8 @@ configure(opentelemetryProjects) {
         testImplementation libraries.junit,
                 libraries.mockito,
                 libraries.truth,
-                libraries.guava_testlib
+                libraries.guava_testlib,
+                platform(boms.guava)
 
         // The ErrorProne plugin defaults to the latest, which would break our
         // build if error prone releases a new version with a new check
--- a/exporters/jaeger/build.gradle
+++ b/exporters/jaeger/build.gradle
@@ -14,6 +14,7 @@ dependencies {
 
     implementation project(':opentelemetry-sdk-contrib-otproto'),
             project(':opentelemetry-sdk'),
+            platform(boms.protobuf),
             libraries.grpc_api,
             libraries.grpc_protobuf,
             libraries.grpc_stub,
--- a/exporters/otlp/build.gradle
+++ b/exporters/otlp/build.gradle
@@ -12,6 +12,7 @@ dependencies {
     api project(':opentelemetry-sdk')
 
     implementation project(':opentelemetry-sdk-contrib-otproto'),
+            platform(boms.protobuf),
             libraries.grpc_api,
             libraries.grpc_protobuf,
             libraries.grpc_stub,
--- a/exporters/prometheus/build.gradle
+++ b/exporters/prometheus/build.gradle
@@ -11,8 +11,7 @@ ext.moduleName = "io.opentelemetry.exporters.prometheus"
 dependencies {
     api project(':opentelemetry-sdk')
 
-    implementation libraries.guava,
-            libraries.prometheus_client
+    implementation libraries.prometheus_client
 
     testImplementation libraries.prometheus_client_common
 
--- a/exporters/zipkin/build.gradle
+++ b/exporters/zipkin/build.gradle
@@ -16,7 +16,8 @@ dependencies {
     annotationProcessor libraries.auto_value
 
     implementation libraries.zipkin_reporter,
-            libraries.zipkin_urlconnection
+            libraries.zipkin_urlconnection,
+            platform(boms.zipkin_reporter)
 
     testImplementation libraries.guava,
             libraries.zipkin_junit
--- a/proto/build.gradle
+++ b/proto/build.gradle
@@ -13,7 +13,8 @@ dependencies {
     api libraries.protobuf,
             libraries.grpc_api,
             libraries.grpc_protobuf,
-            libraries.grpc_stub
+            libraries.grpc_stub,
+            platform(boms.protobuf)
 
     signature "org.codehaus.mojo.signature:java17:1.0@signature"
     signature "net.sf.androidscents.signature:android-api-level-14:4.0_r4@signature"
--- a/sdk/build.gradle
+++ b/sdk/build.gradle
@@ -12,7 +12,8 @@ ext.moduleName = "io.opentelemetry.sdk"
 dependencies {
     api project(':opentelemetry-api')
 
-    implementation libraries.guava
+    implementation libraries.guava,
+            platform(boms.guava)
 
     annotationProcessor libraries.auto_value
 
--- a/sdk_contrib/async_processor/build.gradle
+++ b/sdk_contrib/async_processor/build.gradle
@@ -13,7 +13,8 @@ dependencies {
             project(':opentelemetry-sdk')
 
     implementation libraries.guava,
-            libraries.disruptor
+            libraries.disruptor,
+            platform(boms.guava)
 
     signature "org.codehaus.mojo.signature:java18:1.0@signature"
 }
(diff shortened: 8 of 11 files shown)


build.gradle
```diff
@@ -180,10 +180,7 @@ configure(opentelemetryProjects) {
 
     dependencies {
 
-        implementation platform(boms.grpc),
-                platform(boms.guava),
-                platform(boms.protobuf),
-                platform(boms.zipkin_reporter)
+        implementation platform(boms.grpc)
```
Oberon00, line 183: I don't think we want enforce versions for all grpc components for all artifacts. But probably this is a (serious) issue that will only be resolved with #575. That is, even though we "only" depend on grpc-context, we force every user of the API to only use grpc versions (including other artifacts) that can work together with our grpc-context version. BTW, if grpc also depends on the proposed "propagation.io" context, we will still have the same problem.

bogdandrutu:
@iNikem did you search for any plugin that adds the boms automatically if you depend on any artifact from that bom? That would be a bit more convenient if it exists.

build.gradle
bogdandrutu: Not sure I understand the latest concern. Can you elaborate why it will be the same problem?

Oberon00: Any user of propagation-io-context still needs to use a propagation-io-context version compatible with the grpc version they use. It's not the same problem with BOMs but that fundamental problem remains.

iNikem:
No, I haven't. And I very much doubt one exists. It would mean that for every dependency you have to look for every platform (declared where?) and check if it has that dependency. Performance would be awful, I am afraid.

And I am not sure you want to "magically" resolve your dependencies, as opposed to being very explicit about them.

build.gradle
Oberon00: If the user uses a grpc BOM, that BOM should include propagation-io-context and would override the version declared by opentelemetry-api.

On second thought, as for this PR, I think we should just not use the GRPC bom. BOMs for ensuring that all the versions of related dependencies match. But we want to depend only on a single dependency, namely the context. So there is nothing we could be incompatible with. If the user uses a grpc bom, it will override the API grpc version, which can be bad (see first paragraph of this comment, and my previous comment), but arguably better than opentelemetry-api overriding the grpc version of a user who does not use BOMs (not sure if that's the case).

Oberon00:
There are at least two BOMs now that contain grpc-context for example: The Opentelemetry BOM and the grpc BOM. Which one should you use? How do you know what the "primary" BOM is? Maybe there actually is a mechanism for that, I don't know.

build.gradle
Oberon00:
```suggestion
```

bogdandrutu: we should use the bom for otlp exporter probably but not for API, I think that is what you say. Also user can overwrite all the versions by using their own bom. A bom ensures that all the versions for a set of artifacts are the same, user can still overwrite the bom version.

iNikem: We use several grpc artifacts, e.g. `grpc-protobuf` and `grpc-stub`. So using bom to align their version seems like a totally sensible thing to do. The only concern that I raised and try to solve with this PR, is the unneeded dependency in `api`, which breaks api clients, such as otel auto-instrumentation agent. All other concerns regarding artifacts dependencies can be addressed separately, if they signal the actual problems.

iNikem:
Can somebody please rerun CI job?

build.gradle
Oberon00: I agree we should use the grpc BOM for anything that actually uses grpc. But here it is applied to all artifacts, including API and SDK.

iNikem: api uses `context_prop` which uses grpc.

bogdandrutu:
There is a problem with the testcontainers

build.gradle
Oberon00: From  #575 I got the impression that we should threat grpc-context as "not actually grpc but just in the same group". Anyway, it uses a single artifact from that BOM so I think the following choices are equally valid technically:

1. Use the BOM for api. Then any consumers of api using gprc will be forced (by default; I know now this can be overriden) to use a grpc compatible with the grpc-context that the API uses.
2. Don't use the BOM for api. Then any consumers of api using grpc will use whatever version of grpc they declared. If that version is incompatible with our grpc-context, I'm not sure what will happen.

carlosalberto: > Don't use the BOM for api.

+1 on this. This would simplify things overall (which is exactly what we need for the API case).

iNikem: That would mean removing all shared dependency management related stuff from `api`. I cannot even write `api libraries.grpc_context` then, but have to spell it out `api "io.grpc:grpc-context:${grpcVersion}"`. Are you sure? :)

iNikem:
Approve? Merge? :)

jkwatson:
re-ran the CI workflows

anuraaga:
I think the issue may be coming by not setting `versionMapping` to `fromResolutionResult()`. Can you try that? There's no reason for the POMs that are published from Gradle to be importing BOMs.

https://docs.gradle.org/current/dsl/org.gradle.api.publish.maven.MavenPublication.html#org.gradle.api.publish.maven.MavenPublication:versionMapping(org.gradle.api.Action)

anuraaga:
Also while checking the POMs it's probably a good opportunity to see why there's no Gradle metadata, or if intentional, disabling it explicitly

https://docs.gradle.org/current/userguide/publishing_gradle_module_metadata.html#sub:disabling-gmm-publication

It's strange seeing the "Gradle metadata is published" comment in the POM but no `.module` file in the folder.

https://repo.maven.apache.org/maven2/io/opentelemetry/opentelemetry-api/0.4.1/opentelemetry-api-0.4.1.pom

anuraaga:
I sent #1288 addressing the points I mentioned

DotSpy:
Sorry just saw this issue i totally agree with @anuraaga, boms should just  resolve INNER project version and not leak out

iNikem:
In truth, I don't really care which one of the PRs will be merged, but I am eagerly looking forward to being able to compile auto-instr against latest snapshot of this repo :)

jkwatson:
do we have some evidence that this will fix the issue? [I'm not gradle guru, so it all just looks like magic to me]

iNikem:
> do we have some evidence that this will fix the issue? [I'm not a gradle guru, so it all just looks like magic to me]

My local tests show that yes, it will

DotSpy:
For now i think it will not work with exporters because they will add <dependencyManagement> section to pom and we should stick with @anuraaga solution

iNikem:
Do we need more comments/approves here or this can be merged? Or let us merge #1288 ?

jkwatson:
Carlos and I have approved, so I'll merge this

jkwatson:
@iNikem So.... does this work as desired? What do we do with #1288 ?

anuraaga:
@jkwatson I think this work as desired for some problem. But there were concerns about the BOMs being included in the POMs in the first place. 

https://github.com/open-telemetry/opentelemetry-java/pull/1281#issuecomment-636067348
https://github.com/open-telemetry/opentelemetry-java/pull/1281#discussion_r431910031

If the POMs don't have BOMs included in the first place, don't even have to worry :) So recommend still going ahead with #1288

iNikem:
> @iNikem So.... does this work as desired? What do we do with #1288 ?

Yes, tests pass :)

PR #1284, opened 2020-05-28
feat: examples for metrics

DotSpy:
Please provide feedback what u want to see in samples and i will try to add it


examples/metrics/src/main/java/io/opentelemetry/example/metrics/DoubleMeterExample.java
```diff
@@ -0,0 +1,22 @@
...
+
+import io.opentelemetry.OpenTelemetry;
+import io.opentelemetry.metrics.DoubleCounter;
+import io.opentelemetry.metrics.Meter;
+
+public class DoubleMeterExample {
+
+  public static void main(String[] args) {
+    Meter sampleMeter = OpenTelemetry.getMeterProvider().get("sample", "0.1");
+    DoubleCounter jvmUsageCounter = sampleMeter.doubleCounterBuilder("jvm_memory_usage")
```
jkwatson, line 11: this is probably a bad example of how to use a counter. Counters are intended to be used as a part of a "synchronous" measurement, associated with a span context. Measuring jvm memory usage is a better use-case for an Observer instrument.

examples/metrics/src/main/java/io/opentelemetry/example/metrics/ObserverMeterExample.java
```diff
@@ -0,0 +1,25 @@
+package io.opentelemetry.example.metrics;
+
+import io.opentelemetry.OpenTelemetry;
+import io.opentelemetry.metrics.LongValueObserver;
+import io.opentelemetry.metrics.LongValueObserver.ResultLongValueObserver;
+import io.opentelemetry.metrics.Meter;
+
+public class ObserverMeterExample {
```
jkwatson, line 8: Please add Javadoc explaining the purpose of the example.

examples/metrics/src/main/java/io/opentelemetry/example/metrics/ObserverMeterExample.java
```diff
@@ -0,0 +1,25 @@
...
+
+  public static void main(String[] args) {
+    Meter sampleMeter = OpenTelemetry.getMeterProvider().get("sample", "0.1");
```
jkwatson, line 11: I think it would be helpful to have the meter name reflect reommended meter naming standards, which would mean this should probably be something like "io.opentelemetry.example.metrics" with a version of 0.5, maybe

examples/metrics/src/main/java/io/opentelemetry/example/metrics/ObserverMeterExample.java
```diff
@@ -0,0 +1,25 @@
...
+    LongValueObserver observer = sampleMeter.longValueObserverBuilder("jvm_memory_usage")
+        .setDescription("should meter jvm memory usage")
+        .setUnit("byte")
+        .build();
+
+    observer.setCallback(
+        new LongValueObserver.Callback<LongValueObserver.ResultLongValueObserver>() {
+          @Override
+          public void update(ResultLongValueObserver result) {
+            result.observe(Runtime.getRuntime().totalMemory(), "jvm memory", "free");
```
thisthat, line 21:
```suggestion
            result.observe(Runtime.getRuntime().totalMemory(), "jvm memory", "total");
```

DotSpy:
Thank you guys for comments it's very important to me, i would apply your suggestions after i come home

examples/metrics/src/main/java/io/opentelemetry/example/metrics/DoubleMeterExample.java
DotSpy: @jkwatson maybe u have smth on mind to count wia double counter?

jkwatson: I don't. I'm not sure what a use-case is for counting things that are non-integer valued. Maybe total number of kilobytes processed in some kind of web request?

DotSpy: but they are also int

jkwatson: you can have fractional kb, though, right?

DotSpy: ok i got smth on my mind

jkwatson:
The `.JaegerRemoteSamplerTest` failed again. I thought that had been fixed!

examples/metrics/src/main/java/io/opentelemetry/example/metrics/LongMeterExample.java
```diff
@@ -0,0 +1,38 @@
...
+import io.opentelemetry.OpenTelemetry;
+import io.opentelemetry.metrics.LongCounter;
+import io.opentelemetry.metrics.LongCounter.BoundLongCounter;
+import io.opentelemetry.metrics.Meter;
+import java.util.Random;
+
+/**
+ * Long meter example with bound counter
+ */
+public class LongMeterExample {
```
jkwatson, line 12: This should be renamed to `LongCounterExample`, and it still needs to have a span context surrounding the usage of the counter, to be idiomatically correct.

The javadoc should definitely be more descriptive, and probably include a reference to the relevant section of the metric specifications. It should explain under what circumstances it is appropriate to use a LongCounter, and explain what "bound" means. My expectation is that a newbie who is looking at this code will see an idiomatic example of how to use the instrument being demonstrated and understand why and when they would use such an instrument.

DotSpy: @jkwatson gotch you

examples/metrics/src/main/java/io/opentelemetry/example/metrics/LongCounterExample.java
```diff
@@ -0,0 +1,60 @@
...
+import io.opentelemetry.metrics.Meter;
+import io.opentelemetry.trace.Span;
+import io.opentelemetry.trace.Span.Kind;
+import io.opentelemetry.trace.Tracer;
+import java.io.File;
+import javax.swing.filechooser.FileSystemView;
+
+/**
+ * Example to search all directories for specific file.
+ * Here we will search directory recursively for specific file, using meter to count searched directories.
```
jkwatson, line 16: I think you missed my point here. The javadoc should explain what the example is trying to teach. We aren't trying to teach people how to traverse directory structures. We're trying to teach people how to use a Counter, a LongCounter in particular.

examples/metrics/src/main/java/io/opentelemetry/example/metrics/LongCounterExample.java
```diff
@@ -0,0 +1,60 @@
...
+  private static final Meter sampleMeter = OpenTelemetry.getMeterProvider()
+      .get("io.opentelemetry.example.metrics", "0.5");
+  private static final LongCounter methodCallCounter = sampleMeter
+      .longCounterBuilder("directories_search_count")
+      .setDescription("should count directories searched")
+      .setUnit("unit")
+      .build();
+  //we can use BoundCounters to not specify labels each time
+  private static final BoundLongCounter directoryCounter = methodCallCounter
+      .bind("directory", "searched");
```
jkwatson, line 31: this is a strange label. Why would this be a useful label for this process?

examples/metrics/src/main/java/io/opentelemetry/example/metrics/ObserverMeterExample.java
```diff
@@ -0,0 +1,29 @@
+package io.opentelemetry.example.metrics;
+
+import io.opentelemetry.OpenTelemetry;
+import io.opentelemetry.metrics.LongValueObserver;
+import io.opentelemetry.metrics.LongValueObserver.ResultLongValueObserver;
+import io.opentelemetry.metrics.Meter;
+
+/**
+ * In this example we are setting callback to get asynchronously jvm memory usage
```
jkwatson, line 9: As above, the javadoc should explain what we're trying to teach people with the example.

examples/metrics/src/main/java/io/opentelemetry/example/metrics/LongCounterExample.java
DotSpy: i was thinking to count number of directories that was searched

DotSpy: Shouldn't it be like Directories searched 5 with this label?

examples/metrics/src/main/java/io/opentelemetry/example/metrics/LongCounterExample.java
DotSpy: That example trying to show usage of LongCounter, so as i understand u JD will look like
```Example of using LongCounter and BoundCounters to count searched directories.```

Comment (jkwatson) [reply examples/metrics/src/main/java/io/opentelemetry/example/metrics/LongCounterExample.java:16]: I would also include a link to the relevant specification, and a link to the javadoc for LongCounter class itself, if possible.

File: examples/metrics/src/main/java/io/opentelemetry/example/metrics/LongCounterExample.java (hunk 7 above)
Comment (jkwatson) [reply examples/metrics/src/main/java/io/opentelemetry/example/metrics/LongCounterExample.java:31]: but, the name of the instrument already has this same information in it, so why would you add a redundant label? You might add a bound label for the disk volume, or the host name, or something that would not just repeat the information that is in the instrument name.

File: examples/metrics/src/main/java/io/opentelemetry/example/metrics/LongCounterExample.java (hunk 6 above)
Comment (DotSpy) [reply examples/metrics/src/main/java/io/opentelemetry/example/metrics/LongCounterExample.java]: Applied suggestions, also added span status and catch block

File: examples/metrics/src/main/java/io/opentelemetry/example/metrics/DoubleMeterExample.java [hunk 9]
```diff
@@ -0,0 +1,34 @@
...
+
+import io.opentelemetry.OpenTelemetry;
+import io.opentelemetry.metrics.DoubleCounter;
+import io.opentelemetry.metrics.Meter;
+
+
+/**
+ * Example of using {@link DoubleCounter} to measure execution time of method.
+ */
+public class DoubleMeterExample {
```
jkwatson, line 11: rename to `DoubleCounterExample`

examples/metrics/src/main/java/io/opentelemetry/example/metrics/DoubleMeterExample.java
```diff
@@ -0,0 +1,34 @@
...
+        .get("io.opentelemetry.example.metrics", "0.5");
+    DoubleCounter timeCounter = sampleMeter.doubleCounterBuilder("some_method_time_usage")
+        .setDescription("should measure some method execution time")
+        .setUnit("second")
+        .build();
+    Long timeStart = System.currentTimeMillis();
+    superLongMethod();
+    Long timeEnd = System.currentTimeMillis();
+    Double seconds = (timeEnd.doubleValue() - timeStart.doubleValue()) / 1000;
+    timeCounter.add(seconds, "someWork", "execution time");
```
jkwatson, line 24: this usage should also be in a Span scope, like in the LongCounterExample.

DotSpy:
I also changed 
```
      timeCounter.add(seconds);
```
to show that we can use it without labels

examples/metrics/src/main/java/io/opentelemetry/example/metrics/DoubleCounterExample.java
```diff
@@ -0,0 +1,53 @@
...
+    Span span = tracer.spanBuilder("superLongMethod")
+        .setSpanKind(Kind.INTERNAL)
+        .startSpan();
+    try (Scope scope = tracer.withSpan(span)) {
+      Meter sampleMeter = OpenTelemetry.getMeterProvider()
+          .get("io.opentelemetry.example.metrics", "0.5");
+      DoubleCounter timeCounter = sampleMeter.doubleCounterBuilder("some_method_time_usage")
+          .setDescription("should measure some method execution time")
+          .setUnit("second")
+          .build();
```
jkwatson, line 31: Move the creation of the counter up out of the try block. We only need to use the counter inside a span context, not create it.

jkwatson:
Looking good. Just one little tweak and this will be ready to go.

DotSpy:
> 
> 
> Thanks for the hard work!

I have to thank you for your patience!

examples/metrics/src/main/java/io/opentelemetry/example/metrics/DoubleCounterExample.java
```diff
@@ -0,0 +1,53 @@
...
+
+  public static void main(String[] args) {
+    Span span = tracer.spanBuilder("superLongMethod")
+        .setSpanKind(Kind.INTERNAL)
+        .startSpan();
+    Meter sampleMeter = OpenTelemetry.getMeterProvider()
+        .get("io.opentelemetry.example.metrics", "0.5");
+    DoubleCounter timeCounter = sampleMeter.doubleCounterBuilder("some_method_time_usage")
+        .setDescription("should measure some method execution time")
+        .setUnit("second")
```
arminru, line 29: Is a counter really the right instrument type to use here?
I thought this was only used for things that you actually count, like, the number of requests, errors or total bytes written or something like this. In particular, things where one would be interested in the total sum only and not distinct values or their min/max. Of course one could argue that measuring time entails counting seconds but I thought the ValueRecorder would be the correct instrument to record such distinct measurements.

The [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#valuerecorder) even mentions:

>Example uses for ValueRecorder that are non-additive:
>- capture any kind of timing information

Also the example in the Javadoc even demonstrates how it's used to report timings:
https://github.com/open-telemetry/opentelemetry-java/blob/master/api/src/main/java/io/opentelemetry/metrics/DoubleValueRecorder.java

examples/metrics/src/main/java/io/opentelemetry/example/metrics/DoubleCounterExample.java
```diff
@@ -0,0 +1,53 @@
...
+        .build();
+    try (Scope scope = tracer.withSpan(span)) {
+      Long timeStart = System.currentTimeMillis();
+      superLongMethod();
+      Long timeEnd = System.currentTimeMillis();
+      Double seconds = (timeEnd.doubleValue() - timeStart.doubleValue()) / 1000;
+      //we can add values to counter without addition label key-values pairs
+      timeCounter.add(seconds);
+    } catch (Exception e) {
+      Status status = Status.UNKNOWN.withDescription("Error while finding file");
```
arminru, line 39: Leftover, this example is not about finding any files.

examples/metrics/src/main/java/io/opentelemetry/example/metrics/DoubleCounterExample.java
jkwatson: Do we know what a DoubleCounter would be used to count? Seems like an odd thing to count fractional stuff.

jkwatson: I wonder, if we don't have a good use-case for it, if we should consider removing it from the API.

DotSpy:
i review ur suggestions and kinda modified example, please take a look

examples/metrics/src/main/java/io/opentelemetry/example/metrics/DoubleCounterExample.java
```diff
@@ -0,0 +1,57 @@
...
+import io.opentelemetry.metrics.Meter;
+import io.opentelemetry.trace.Span;
+import io.opentelemetry.trace.Span.Kind;
+import io.opentelemetry.trace.Status;
+import io.opentelemetry.trace.Tracer;
+import java.io.File;
+import javax.swing.filechooser.FileSystemView;
+
+/**
+ * Example of using {@link DoubleCounter} and {@link DoubleCounter.BoundDoubleCounter} to count used space by file  with specific extension.
```
arminru, line 16: Actually you don't use the `BoundDoubleCounter` or labels here at all, or am I missing something? 😅 

The file extension would actually make a great metric dimension/label in your example.

```
disk_space_used[file_extension=dll]:   5 MB
disk_space_used[file_extension=pdf]:   1 MB
disk_space_used[file_extension=png]: 710 MB
```
and so on.

What do you think? 🙂

examples/metrics/src/main/java/io/opentelemetry/example/metrics/DoubleCounterExample.java
```diff
@@ -0,0 +1,57 @@
...
+public class DoubleCounterExample {
+
+  private static final Tracer tracer =
+      OpenTelemetry.getTracer("io.opentelemetry.example.metrics");
+  private static final Meter sampleMeter = OpenTelemetry.getMeterProvider()
+      .get("io.opentelemetry.example.metrics", "0.5");
+  private static final File directoryToCountIn = FileSystemView.getFileSystemView().getHomeDirectory();
+  private static final DoubleCounter methodCallCounter = sampleMeter
+      .doubleCounterBuilder("calculated_used_space")
+      .setDescription("should count space used by file extension")
```
arminru, line 27:
```suggestion
      .setDescription("Counts disk space used by file extension.")
```

It actually does so no need for the subjunctive 😉

examples/metrics/src/main/java/io/opentelemetry/example/metrics/DoubleCounterExample.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+  }
+
+  public void calculateSpaceUsedByFilesWithExtension(String extension, File directory) {
+    File[] files = directory.listFiles();
+    if (files != null) {
+      for (File file : files) {
+        if (file.getName().endsWith("." + extension)) {
+          //we can add values to counter without addition label key-values pairs
+          methodCallCounter.add((double) file.length() / 1024);
```
arminru, line 52:
```suggestion
          methodCallCounter.add((double) file.length() / 1_000_000);
```

`File.length()` returns the size in Bytes but you specified `.setUnit("MB")` above.

examples/metrics/src/main/java/io/opentelemetry/example/metrics/DoubleCounterExample.java
arminru: @jkwatson Fair point. I think it makes sense to consider it but it might only be us not being able to come up with a good example. The example in DoubleCounter's Javadoc is slightly contrived but I could imagine that there are reasonable use cases for it out in the wild. Apart from that, removing it might be surprising to users for the lack of symmetry in the API then.

examples/metrics/src/main/java/io/opentelemetry/example/metrics/DoubleCounterExample.java
DotSpy: True :)

examples/metrics/src/main/java/io/opentelemetry/example/metrics/DoubleCounterExample.java
DotSpy: Sounds good, when i will be at home i'll try to implement it

DotSpy: @arminru when u have a time, pls take a look

examples/metrics/src/main/java/io/opentelemetry/example/metrics/DoubleCounterExample.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  public void calculateSpaceUsedByFilesWithExtension(List<String> extensions, File directory) {
+    File[] files = directory.listFiles();
+    if (files != null) {
+      for (File file : files) {
+        for (String extension : extensions) {
+          if (file.getName().endsWith("." + extension)) {
+            //we can add values to counter without addition label key-values pairs
+            methodCallCounter.add((double) file.length() / 1_000_000, "file_extension", extension);
```
arminru, line 59:
```suggestion
            diskSpaceCounter.add((double) file.length() / 1_000_000, "file_extension", extension);
```

examples/metrics/src/main/java/io/opentelemetry/example/metrics/DoubleCounterExample.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.List;
+import javax.swing.filechooser.FileSystemView;
+
+/**
+ * Example of using {@link DoubleCounter} and {@link DoubleCounter.BoundDoubleCounter} to count disk space used by files with a specific extensions.
+ */
+public class DoubleCounterExample {
+
+  private static final Tracer tracer =
+      OpenTelemetry.getTracer("io.opentelemetry.example.metrics");
```
arminru, line 22:
```suggestion
      OpenTelemetry.getTracer("io.opentelemetry.example.metrics", "0.5");
```
You would usually supply the same library version as for the meter below.

examples/metrics/src/main/java/io/opentelemetry/example/metrics/DoubleCounterExample.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+  }
+
+  public void calculateSpaceUsedByFilesWithExtension(List<String> extensions, File directory) {
+    File[] files = directory.listFiles();
+    if (files != null) {
+      for (File file : files) {
+        for (String extension : extensions) {
+          if (file.getName().endsWith("." + extension)) {
+            //we can add values to counter without addition label key-values pairs
```
arminru, line 58:
```suggestion
            // we can add values to the counter for specific labels
            // the label key is "file_extension", its value is the name of the extension
```

examples/metrics/src/main/java/io/opentelemetry/example/metrics/LongCounterExample.java
```diff
@@ -0,0 +1,64 @@
...
+import java.io.File;
+import javax.swing.filechooser.FileSystemView;
+
+/**
+ * Example of using {@link LongCounter} and {@link LongCounter.BoundLongCounter} to count searched directories.
+ */
+public class LongCounterExample {
+
+  private static final Tracer tracer =
+      OpenTelemetry.getTracer("io.opentelemetry.example.metrics");
```
arminru, line 21:
```suggestion
      OpenTelemetry.getTracer("io.opentelemetry.example.metrics", "0.5");
```

examples/metrics/src/main/java/io/opentelemetry/example/metrics/DoubleCounterExample.java
```diff
@@ -0,0 +1,65 @@
...
+import io.opentelemetry.trace.Span.Kind;
+import io.opentelemetry.trace.Status;
+import io.opentelemetry.trace.Tracer;
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+import javax.swing.filechooser.FileSystemView;
+
+/**
+ * Example of using {@link DoubleCounter} and {@link DoubleCounter.BoundDoubleCounter} to count disk space used by files with a specific extensions.
```
arminru, line 17:
```suggestion
 * Example of using {@link DoubleCounter} to count disk space used by files with specific extensions.
```

examples/metrics/src/main/java/io/opentelemetry/example/metrics/LongCounterExample.java
```diff
@@ -0,0 +1,64 @@
...
+  private static final Meter sampleMeter = OpenTelemetry.getMeterProvider()
+      .get("io.opentelemetry.example.metrics", "0.5");
+  private static final LongCounter methodCallCounter = sampleMeter
```
arminru, line 24:
```suggestion
  private static final LongCounter directoryCounter = sampleMeter
```

examples/metrics/src/main/java/io/opentelemetry/example/metrics/LongCounterExample.java
```diff
@@ -0,0 +1,64 @@
...
+      .longCounterBuilder("directories_search_count")
+      .setDescription("should count directories searched")
+      .setUnit("unit")
+      .build();
+  private static final File homeDirectory = FileSystemView.getFileSystemView().getHomeDirectory();
+  //we can use BoundCounters to not specify labels each time
+  private static final BoundLongCounter directoryCounter = methodCallCounter
```
arminru, line 31:
```suggestion
  private static final BoundLongCounter homeDirectoryCounter = directoryCounter
```

examples/metrics/src/main/java/io/opentelemetry/example/metrics/LongCounterExample.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+  }
+
+  public void findFile(String name, File directory) {
+    File[] files = directory.listFiles();
+    System.out.println("Currently looking at " + directory.getAbsolutePath());
+    if (files != null) {
+      for (File file : files) {
+        if (file.isDirectory()) {
+          directoryCounter.add(1);
```
arminru, line 56:
```suggestion
          // we don't have to specify the value for the "root directory" label again
          // since this is a BoundLongCounter with pre-set labels
          homeDirectoryCounter.add(1);
```

examples/metrics/src/main/java/io/opentelemetry/example/metrics/LongCounterExample.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+public class LongCounterExample {
+
+  private static final Tracer tracer =
+      OpenTelemetry.getTracer("io.opentelemetry.example.metrics");
+  private static final Meter sampleMeter = OpenTelemetry.getMeterProvider()
+      .get("io.opentelemetry.example.metrics", "0.5");
+  private static final LongCounter methodCallCounter = sampleMeter
+      .longCounterBuilder("directories_search_count")
+      .setDescription("should count directories searched")
```
arminru, line 26:
```suggestion
      .setDescription("Counts directories accessed while searching for files.")
```

examples/metrics/src/main/java/io/opentelemetry/example/metrics/LongValueObserverExample.java
```diff
@@ -0,0 +1,32 @@
...
+    LongValueObserver observer = sampleMeter.longValueObserverBuilder("jvm_memory_usage")
+        .setDescription("should meter jvm memory usage")
+        .setUnit("byte")
+        .build();
+
+    observer.setCallback(
+        new LongValueObserver.Callback<LongValueObserver.ResultLongValueObserver>() {
+          @Override
+          public void update(ResultLongValueObserver result) {
+            result.observe(Runtime.getRuntime().totalMemory(), "jvm memory", "total");
```
arminru, line 27:
```suggestion
            result.observe(Runtime.getRuntime().totalMemory(), "kind", "total");
            result.observe(Runtime.getRuntime().freeMemory(), "kind", "free");
```

To demonstrate the advantage of using labels. Otherwise it doesn't really show why you would use them at all here.

One might prefer to report these as two separate metrics rather than metric dimensions, however.
@bogdandrutu, you're more experienced with metrics. WDYT?

DotSpy: we can throw away labels from this example if it misleading

arminru:
@bogdandrutu The build is failing because it misses the `codecov/patch` test coverage target. Can we exclude sample code from this check?

DotSpy:
@bogdandrutu @jkwatson i can think we can merge this one

PR #1285, opened 2020-05-28
Do not inject invalid SpanContext instances.

### LINKED ISSUE #1272: Propagation does not work well with invalid current span.
**Describe the bug**
When an invalid default span (such as returned from `Tracer.getCurrentSpan()` when there is none) is explicitly set e.g. with `Tracer.withSpan`, the HttpTraceContext propagates a header `traceparent: 00-00000000000000000000000000000000-0000000000000000-00`.

**Steps to reproduce**

The following code reproduces the issue (add e.g. to `DefaultTracerTest` in the api test package):

```java
   @Test
  public void testSpanContextPropagationCurrentSpanContext() {
    Context context =
        TracingContextUtils.withSpan(DefaultSpan.create(spanContext), Context.current());
    Scope scope = ContextUtils.withScopedContext(context);
    try {
      Span span = defaultTracer.spanBuilder(SPAN_NAME).startSpan();
      assertThat(span.getContext()).isSameInstanceAs(spanContext);
    } finally {
      scope.close();
    }
  }

  @Test
  public void propagateNoCurrentSpan() {
    final Tracer tracer = OpenTelemetry.getTracer("nospandemo");

    final Setter<Void> printSetter =
        new Setter<Void>() {
          @Override
          public void set(@Nullable Void carrier, String key, String value) {
            System.out.println(key + ": " + value);
          }
        };

    System.out.println("Propagate, no span.");
    OpenTelemetry.getPropagators().getHttpTextFormat().inject(Context.current(), null, printSetter);

    final Span currentSpan = tracer.getCurrentSpan();

    // ... Some code in between, possibly switching threads ...

    try (Scope _scope = tracer.withSpan(currentSpan)) {
      System.out.println("Propagate, invalid span.");
      OpenTelemetry.getPropagators()
          .getHttpTextFormat()
          .inject(Context.current(), null, printSetter);
    }
  }
```

I expect the propagation framework to detect that there is nothing to inject in both cases, hence the following output:

```
Propagate, no span.

Propagate, invalid span.

```

**What did you see instead?**

The invalid span was propagated by the HttpTraceContext propagator.

```
Propagate, no span.

Propagate, invalid span.

traceparent: 00-00000000000000000000000000000000-0000000000000000-00
```

master from a git checkout, commit 84f5a3eab01c269e43c4962c5acc04bfc97de09b

carlosalberto:
Fixes #1272 

I don't imagine we propagating invalid `SpanContext` instances with non-default TraceFlags/TraceState, hence the `isValid()` check, as opposed to compare against `SpanContext.getInvalid()` ;)

### DIFF
--- a/contrib/trace_propagators/src/main/java/io/opentelemetry/contrib/trace/propagation/B3PropagatorInjectorMultipleHeaders.java
+++ b/contrib/trace_propagators/src/main/java/io/opentelemetry/contrib/trace/propagation/B3PropagatorInjectorMultipleHeaders.java
@@ -32,7 +32,7 @@ final class B3PropagatorInjectorMultipleHeaders implements B3PropagatorInjector 
     Objects.requireNonNull(setter, "setter");
 
     Span span = TracingContextUtils.getSpanWithoutDefault(context);
-    if (span == null) {
+    if (span == null || !span.getContext().isValid()) {
       return;
     }
 
--- a/api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java
+++ b/api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java
@@ -81,7 +81,7 @@ public class HttpTraceContext implements HttpTextFormat {
     checkNotNull(setter, "setter");
 
     Span span = TracingContextUtils.getSpanWithoutDefault(context);
-    if (span == null) {
+    if (span == null || !span.getContext().isValid()) {
       return;
     }
 
--- a/api/src/test/java/io/opentelemetry/trace/propagation/HttpTraceContextTest.java
+++ b/api/src/test/java/io/opentelemetry/trace/propagation/HttpTraceContextTest.java
@@ -115,6 +115,22 @@ public class HttpTraceContextTest {
   }
 
   @Test
+  public void inject_invalidContext() {
+    Map<String, String> carrier = new LinkedHashMap<>();
+    httpTraceContext.inject(
+        withSpanContext(
+            SpanContext.create(
+                TraceId.getInvalid(),
+                SpanId.getInvalid(),
+                SAMPLED_TRACE_OPTIONS,
+                TraceState.builder().set("foo", "bar").build()),
+            Context.current()),
+        carrier,
+        setter);
+    assertThat(carrier).hasSize(0);
+  }
+
+  @Test
   public void inject_SampledContext() {
     Map<String, String> carrier = new LinkedHashMap<>();
     Context context =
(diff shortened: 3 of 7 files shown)


contrib/trace_propagators/src/main/java/io/opentelemetry/contrib/trace/propagation/B3PropagatorInjectorMultipleHeaders.java
```diff
@@ -32,7 +32,7 @@
     Objects.requireNonNull(setter, "setter");
 
     Span span = TracingContextUtils.getSpanWithoutDefault(context);
-    if (span == null) {
+    if (span == null || !span.getContext().isValid()) {
```
iNikem, line 35: Wouldn't the code be easier if you replace `TracingContextUtils.getSpanWithoutDefault` with `TracingContextUtils.getSpan` and then drop null check and use only `isValid`?

carlosalberto: Yeah, I was thinking about it too. I guess I prefer to do a simple `null` check before doing a full check, i.e. `getContext()` + `isValid()` ;)

FWIW though, I'm thinking, after cases like these, to remove the `getSpanWithoutDefault()` overload (which I might end up doing after doing some context-prop updates).

Oberon00: There might be propagators that do will be able to meaninfully inject something only based on TraceState, even for invalid trace and span IDs. So while an utility function would be good, I think the final decision should be left to the propagator.

carlosalberto: > There might be propagators that will be able to meaningfully inject something only based on TraceState

Right, exactly, I had totally forgotten about this case :)

bogdandrutu: The w3c standard says that traceparent must be present so I am worried that if that can be the case to only propagate tracestate

Oberon00: Sure, this does not apply to W3C but maybe other propagators (vendor ones for example).

PR #1286, opened 2020-05-28
Remove SpanID from sampling parameters

pjanotti:
Update code per latest spec, see:
https://github.com/open-telemetry/opentelemetry-specification/pull/621

The goal is to avoid any SDK implementation from taking a dependency on it.

### DIFF
--- a/sdk/src/main/java/io/opentelemetry/sdk/trace/SpanBuilderSdk.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/trace/SpanBuilderSdk.java
@@ -196,7 +196,6 @@ final class SpanBuilderSdk implements Span.Builder {
   public Span startSpan() {
     SpanContext parentContext = parent(parentType, parent, remoteParent);
     TraceId traceId;
-    SpanId spanId = idsGenerator.generateSpanId();
     TraceState traceState = TraceState.getDefault();
     if (parentContext == null || !parentContext.isValid()) {
       // New root span.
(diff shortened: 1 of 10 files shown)


sdk/src/main/java/io/opentelemetry/sdk/trace/SpanBuilderSdk.java
```diff
@@ -223,13 +223,7 @@ public Span startSpan() {
...
             .getSampler()
             .shouldSample(
-                parentContext,
-                traceId,
-                spanId,
-                spanName,
-                spanKind,
-                immutableAttributes,
-                immutableLinks);
+                parentContext, traceId, spanName, spanKind, immutableAttributes, immutableLinks);
```
jkwatson, line 226: One of the goals of this change in the spec was to enable deferring the generation of a span id when one wasn't needed. Is it possible to defer the spanId creation, and if not, why not?

carlosalberto: @jkwatson To be clear, you are talking about `SpanId` being crated unconditionally before this line? e.g.

```
SpanId spanId = idsGenerator.generateSpanId();
```

?

carlosalberto:
Thanks for the PR! My only concern is that @jkwatson mention of us still needing to create a `SpanId` for non-sampled values.

sdk/src/main/java/io/opentelemetry/sdk/trace/SpanBuilderSdk.java
arminru: @carlosalberto Exactly, this was mentioned on the original spec PR (https://github.com/open-telemetry/opentelemetry-specification/pull/621) to improve performance.

Oberon00: Follow-up idea (for spec): why not also generate trace ID later? And provide an option for samplers that need it (only one single one: ProbabilitySampler) to optionally generate a trace ID themselves that the SDK will use.

pjanotti: Yeah, I was focusing on avoiding samplers starting to use the `SpanId` and missed the optimization. Let me see what can be done.

pjanotti: Changed the code to use `SpanId.getInvalid()` if the span is not sampled and only generate a new `SpanId` if it is sampled.

bogdandrutu: Please revert the last commit, and don't do the optimization now.

pjanotti: Done, reverted the optimization - didn't see your comment on GH prior to the change...

PR #1287, opened 2020-05-29
Update config / deps to allow building on Java 14.

anuraaga:
I normally use Java 14 and found my build failing due to deprecated source=7 option. I figured suppressing this option would be enough to get the build to work and ended up in a rabbit hole :)

- Suppress options warnings because we are a library, not executable, so warnings about compiler arguments aren't so useful
- Instead of updating check to disable Werror on Java 9+, enabled Werror on all Java. 
- Update error prone to fix buggy check that triggered false positive on Utils.checkNotNull
- Only set error prone javac for pre Java 9 (though I don't think anyone uses such a JDK to build this project)
- Improve javax.annotation-api branch to apply to Java 9+ forever
- Update jacoco and mockito to support Java 14
- Fix error prone warnings
- Fix mockito deprecations

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,11 +38,17 @@ configure(opentelemetryProjects) {
     targetCompatibility = 1.7
 
     tasks.withType(JavaCompile) {
-        // We suppress the "try" warning because it disallows managing an auto-closeable with
-        // try-with-resources without referencing the auto-closeable within the try block.
-        // We suppress the "processing" warning as suggested in
-        // https://groups.google.com/forum/#!topic/bazel-discuss/_R3A9TJSoPM
-        it.options.compilerArgs += ["-Xlint:all", "-Xlint:-try", "-Xlint:-processing"]
+        it.options.compilerArgs += [
+                "-Xlint:all",
+                // We suppress the "try" warning because it disallows managing an auto-closeable with
+                // try-with-resources without referencing the auto-closeable within the try block.
+                "-Xlint:-try",
+                // We suppress the "processing" warning as suggested in
+                // https://groups.google.com/forum/#!topic/bazel-discuss/_R3A9TJSoPM
+                "-Xlint:-processing",
+                // We suppress the "options" warning because it prevents
+                "-Xlint:-options",
+        ]
         it.options.errorprone.disableWarningsInGeneratedCode = true
         it.options.errorprone.allDisabledChecksAsWarnings = true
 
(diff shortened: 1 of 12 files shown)


build.gradle
```diff
@@ -38,11 +38,17 @@ configure(opentelemetryProjects) {
...
-        it.options.compilerArgs += ["-Xlint:all", "-Xlint:-try", "-Xlint:-processing"]
+        it.options.compilerArgs += [
+                "-Xlint:all",
+                // We suppress the "try" warning because it disallows managing an auto-closeable with
+                // try-with-resources without referencing the auto-closeable within the try block.
+                "-Xlint:-try",
+                // We suppress the "processing" warning as suggested in
+                // https://groups.google.com/forum/#!topic/bazel-discuss/_R3A9TJSoPM
+                "-Xlint:-processing",
+                // We suppress the "options" warning because it prevents
```
iNikem, line 49: `prevent` what?

anuraaga: Oops fixed

carlosalberto:
> Only set error prone javac for pre Java 9 (though I don't think anyone uses such a JDK to build this project)

I do ;)

On the other hand, I wonder how many of the developers are are already using Java 14.

anuraaga:
> I do ;)

Nice, do you mind checking out the branch and making sure it builds? I didn't make any changes that should break pre Java 9 either.

Most of the changes also allow building on 12+ which should currently be failing (though didn't try it). I think it's good practice for projects to build on modern JDKs instead of requiring JAVA_HOME troubles to people that keep up to date.

Oberon00:
I use a JDK 8 to build 😄 Since it should also run on old JDKs later...

anuraaga:
JDK 8 seems to still be popular - I must have misunderstood the error prone warnings that probably came from the upgrade and I guess weren't there before. I removed that point from the PR description, hope we can focus on raising the ceiling of the build 😅

build.gradle
```diff
@@ -66,17 +72,17 @@ configure(opentelemetryProjects) {
...
         // Ignore warnings for protobuf and jmh generated files.
         it.options.errorprone.excludedPaths = ".*generated.*"
         // "-XepExcludedPaths:.*/build/generated/source/proto/.*"
 
-        // Enforce errorprone warnings to be errors.
-        if (!JavaVersion.current().isJava9() && !JavaVersion.current().isJava10() && !JavaVersion.current().isJava11()) {
-            // TODO: Enable -Werror for Java 9+
-            it.options.compilerArgs += ["-Werror"]
-        }
+        it.options.compilerArgs += ["-Werror"]
```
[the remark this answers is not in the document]
Oberon00: Why? If it's working (it seems to build at least on Java 11 with our CI), I see no reason to disable this.

PR #1289, opened 2020-05-29
Directly fetch AWS resource metadata without depending on heavyweight…

anuraaga:
… AWS SDK and allow extension for non-EC2 resources like ECS.

The AWS SDK is pretty large, and especially for apps that live in AWS but don't actually use the AWS SDK, it is wasteful to require it for instrumentation. As an added bonus, we can drop the JDK floor raise to Java 8 without that dependency.

This migrates to fetching and parsing the metadata directly using the JDK and jackson-core.

There are also several different resource types for AWS corresponding to plugins [here](https://github.com/aws/aws-xray-sdk-java/tree/master/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/plugins) so this provides an interface to allow adding more of them.

I didn't keep the setting of `ResourceConstraints.HOST_NAME` to the private IP, this didn't seem like the intent of the field. But let me know if it does make sense to keep it.

The v1 in the module doesn't make sense anymore since we don't use the v1 SDK here. Let me know if I should rename the module now and/or what sort of deprecation process we should do.

/cc @kbrockhoff @willarmiros


sdk_contrib/aws_v1_support/src/main/java/io/opentelemetry/sdk/contrib/trace/aws/resource/Ec2ResourcePopulator.java
```diff
@@ -0,0 +1,207 @@
...
+          case "imageId":
+            resourceAttributes.put(ResourceConstants.HOST_IMAGE_ID, stringAttributeValue(value));
+            break;
+          case "accountId":
+            resourceAttributes.put(ResourceConstants.CLOUD_ACCOUNT, stringAttributeValue(value));
+            break;
+          case "region":
+            resourceAttributes.put(ResourceConstants.CLOUD_REGION, stringAttributeValue(value));
+            break;
+          default:
```
thisthat, line 116: It is missing the case for "privateIp" to populate `HOST_NAME`

anuraaga: Yeah asked in my PR description about that - looking at the documentation for HOST_NAME, it didn't seem like it's meant for an IP address so left it out for now. But is it better to set that to the IP address?

thisthat:
Sorry for deleting the comment, I thought it would be saved as a draft and not be published :sweat_smile:

anuraaga:
Thanks for the suggestion. Currently, I am setting `meta-data/hostname`, which is the same as `local-hostname` to `HOST_HOSTNAME`. It would be pure duplication to set it to `HOST_NAME` in addition, do you think it's a good idea to duplicate the attribute?

thisthat:
If they are the same, I'd suggest to set it  `HOST_NAME` as `HOST_HOSTNAME`.

anuraaga:
@thisthat No problem, setting them both now

sdk_contrib/aws_v1_support/src/main/java/io/opentelemetry/sdk/contrib/trace/aws/Ec2Resource.java
```diff
@@ -16,16 +16,16 @@
...
-import java.util.Map;
-import javax.annotation.Nullable;
 
-/** Provides for lookup and population of {@link Resource} labels when running on AWS EC2. */
+/**
+ * Provides for lookup and population of {@link Resource} labels when running on AWS EC2.
+ *
+ * @deprecated Use {@link AwsResourcePopulator}.
+ */
+@Deprecated
```
jkwatson, line 28: since we're pre-1.0, I think we can just delete this if it's not the recommended way to do the work.

sdk_contrib/aws_v1_support/src/main/java/io/opentelemetry/sdk/contrib/trace/aws/resource/AwsResourcePopulator.java
```diff
@@ -0,0 +1,49 @@
...
+  public static Resource createResource() {
+    return createResource(new Ec2ResourcePopulator());
+  }
+
+  @VisibleForTesting
+  static Resource createResource(AwsResourcePopulator... populators) {
+    Map<String, AttributeValue> resourceAttributes = new LinkedHashMap<>();
+
+    for (AwsResourcePopulator populator : populators) {
+      populator.populate(resourceAttributes);
```
jkwatson, line 41: Rather than having a side-effect driven API for this, how about having the populator just return the map, and then do an `putAll` with the result each time?

anuraaga: Yeah since this is just a server singleton makes sense, done.

sdk_contrib/aws_v1_support/src/main/java/io/opentelemetry/sdk/contrib/trace/aws/resource/AwsResource.java
```diff
@@ -0,0 +1,49 @@
...
+   * Returns a {@link Resource} which is filled with attributes describing the current AWS
+   * environment, e.g., metadata for the instance if the app is running on EC2.
+   */
+  public static Resource create() {
+    return create(new Ec2ResourcePopulator());
+  }
+
+  @VisibleForTesting
+  static Resource create(AwsResource... populators) {
+    Map<String, AttributeValue> resourceAttributes = new LinkedHashMap<>();
```
jkwatson, line 38: Is it important that we use an ordered-key map implementation here? If so, please add some documentation as to why.

anuraaga: I tend to use `LinkedHashMap` when the allocation impact isn't large like these singleton-like objects so order mattering doesn't affect things, if not now then down the line. I suspect `Resource.create` does too for similar reasons. If it's helpful to add that as a comment let me know I'll add it.

jkwatson: I do see that Resource.create does wrap the map in a LinkedHashMap implementation, but there's no documentation as to why there, either. If an interface takes a Map, it should definitely assume it won't be ordered, unless there's a very good reason for it.

PR #1292, opened 2020-05-30
added testcontainers integration test to verify jaeger example app works in Java 7

### LINKED ISSUE #1249: Create integration test to ensure Java 7 compatibility
See PR #1237 for context

Our built artefacts target Java 7, the unit tests have been changed to use Java 8.

There was some discussion/concern that as the tests are no longer using Java 7, there's a potential risk that we may not catch any problems/misconfiguration with the build and artefacts could be built using Java 8.

I suggest that an example app is created.  This is packaged into a Docker image which uses a Java 7 base image so we ensure that the app is run using Java 7.  The Docker image containing this example application can then be tested using Testcontainers as part of the CI build to ensure compatibility.

jarebudev:
Here's an initial attempt at this. The test is based on the existing ```JaegerIntegrationTest```.  

I think there's some improvements that can be made to how the examples/integration tests fit into the project, I'll put some ideas into a new issue

resolves #1249


examples/jaeger/src/test/java/io/opentelemetry/example/JaegerExampleIntegrationTest.java
```diff
@@ -0,0 +1,106 @@
...
+          .withNetworkAliases(JAEGER_HOSTNAME)
+          .withExposedPorts(COLLECTOR_PORT, QUERY_PORT)
+          .waitingFor(new HttpWaitStrategy().forPath("/"));
+
+  private static GenericContainer jaegerExampleAppContainer =
+      new GenericContainer(buildJaegerExampleAppImage())
+          .withNetwork(network)
+          .waitingFor(Wait.forLogMessage(".*Bye.*", 1));
+
+  private static ImageFromDockerfile buildJaegerExampleAppImage() {
```
bsideup, line 58: Consider using `withCopyFileToContainer` and `withCmd` on `GenericContainer` instead.
There is no ned to build an image here and it will only add an overhead to the startup :)

examples/jaeger/src/test/java/io/opentelemetry/example/JaegerExampleIntegrationTest.java
```diff
@@ -0,0 +1,106 @@
...
+                        "/app/" + APP_NAME,
+                        "io.opentelemetry.example.JaegerExample",
+                        JAEGER_HOSTNAME,
+                        Integer.toString(COLLECTOR_PORT))
+                    .build());
+  }
+
+  @Test
+  public void testJaegerExampleAppIntegration() {
+    jaegerExampleAppContainer.start();
```
bsideup, line 78: any reason you did not annotate `jaegerExampleAppContainer` with `@Rule`?
The container will get removed, eventually (thanks to Testcontainers' two phase cleanup: JVM shutdown hook and Ryuk - a sidecar container), but here you only start it from this test and then never call `stop`, so it will last until JVM's termination

examples/jaeger/src/test/java/io/opentelemetry/example/JaegerExampleIntegrationTest.java
jarebudev: thats a good spot! thanks!

examples/jaeger/src/test/java/io/opentelemetry/example/JaegerExampleIntegrationTest.java
jarebudev: Ah I was going to change that :) 
As its only doing one test, I've changed it to a ```@Rule``` so that its started before the test, the Jaeger container is still started as a ```@ClassRule``` so to guarantee starting in the correct order.

bsideup: > to guarantee starting in the correct order

In case you need to enforce the order, there is also `GenericContainer#dependsOn` for that :)

examples/jaeger/build.gradle
```diff
@@ -3,12 +3,33 @@ plugins {
...
+}
 
 dependencies {
     compile("io.opentelemetry:opentelemetry-api:${opentelemetryVersion}")
     compile("io.opentelemetry:opentelemetry-sdk:${opentelemetryVersion}")
     compile("io.opentelemetry:opentelemetry-exporters-jaeger:${opentelemetryVersion}")
     compile("io.grpc:grpc-protobuf:1.28.0")
     compile("io.grpc:grpc-netty-shaded:1.28.0")
+
+    implementation platform("org.testcontainers:testcontainers-bom:${testcontainersVersion}")
```
jkwatson, line 26: My understanding of "platform" and boms is weak. This doesn't make the jaeger exporter depend on testcontainers, does it? Should this be `testImplementation` ?

examples/jaeger/src/test/java/io/opentelemetry/example/JaegerExampleIntegrationTest.java
jarebudev: thanks that does the job!

examples/jaeger/build.gradle
jarebudev: That's a good catch - it should be ```testImplementation``` only.  I've changed it :)
I assume you meant jaeger example not jaeger exporter?
If we had left it as ```implementation``` then the bom would be imported and available in compile and runtime scopes as well as test, so you could import into these as well e.g. ```compile("org.testcontainers:testcontainers")```

jkwatson: cool.  ➕

jarebudev:
> Thank you for your contribution! :)
> To fully resolve #1249, this PR misses some changes in the CI script to execute this test using an image with Java 7. However, this would also have the side effect to force us to keep the Jaeger example up to date with the SDK which @bogdandrutu was against. Should we make an exception for this?

I'd be in favour of not changing the CI script in this PR and creating a new issue to put some ideas on how the examples/integration tests could be refactored - there's a lot of overlap/duplicate code with the examples and the testcontainers integration tests.  I was initially thinking that it might be better to combine examples into integration tests so that they can be included as part of the build and verify integration with Jaeger/Zipkin/OTel collector etc with each release.

examples/jaeger/build.gradle
```diff
@@ -3,12 +3,33 @@ plugins {
...
+    testImplementation platform("org.testcontainers:testcontainers-bom:${testcontainersVersion}")
+    testCompile("org.testcontainers:testcontainers")
+    testCompile("org.awaitility:awaitility:3.0.0")
+    testCompile("io.rest-assured:rest-assured:4.2.0")
 }
+
+tasks.withType(Test) {
+    dependsOn fatJar
+    systemProperty 'archive.name', fatJar.outputs.files[0].path
+}
```
bogdandrutu, line 35: new line?

jarebudev:
> Not sure if this is an example, maybe we should have an "integrationtests" directory for this?

that would work, i'll move this under a new 'integrationtests' module.

jarebudev:
I've added a new integration_tests module and put the Java 7 int test in here.
If this gets merged there's some tidying up that can be done in a follow up ticket - the other testcontainers integration tests can be moved here to keep them in one place.

jkwatson:
looks like the build isn't quite right yet.

jkwatson:
@jarebudev Looks like there are a couple of conversations to be resolved. Can you make sure everyone's questions/comments are addressed, and the conversations are resolved? Thanks!

examples/jaeger/build.gradle
jarebudev: > new line?

sorted

jarebudev:
> Not sure if this is an example, maybe we should have an "integrationtests" directory for this?

done

jarebudev:
> Thank you for your contribution! :)
> To fully resolve #1249, this PR misses some changes in the CI script to execute this test using an image with Java 7. However, this would also have the side effect to force us to keep the Jaeger example up to date with the SDK which @bogdandrutu was against. Should we make an exception for this?

the changes i've made since allows the integration tests to be run when the CI build is triggered so the examples are kept distinct.

jarebudev:
ooops wrong button didn't meant to close PR!

> @jarebudev Looks like there are a couple of conversations to be resolved. Can you make sure everyone's questions/comments are addressed, and the conversations are resolved? Thanks!

I think I've covered all the comments now, let us know if there's anything amiss.

PR #1298, opened 2020-06-01
Cleanup metrics SDK. Rename variables, allow different aggregation for async instruments

bogdandrutu:


sdk/src/main/java/io/opentelemetry/sdk/metrics/DoubleSumObserverSdk.java
```diff
@@ -17,47 +17,35 @@
...
       InstrumentDescriptor descriptor,
       MeterProviderSharedState meterProviderSharedState,
       MeterSharedState meterSharedState) {
-    super(descriptor, meterProviderSharedState, meterSharedState);
+    super(
+        descriptor,
+        meterProviderSharedState,
+        meterSharedState,
+        new ActiveBatcher(
+            new ActiveBatcher(
```
jkwatson, line 38: Why are we double-wrapping here?

bogdandrutu: Good catch. Fixed.

PR #1300, opened 2020-06-01
Add implementation for value observers

bogdandrutu:
Currently we use a delta batcher with MinMaxSumCount as the default batcher for this. There are some questions about what should be the start time for this points. We can address this later.

### DIFF
--- a/sdk/src/main/java/io/opentelemetry/sdk/metrics/DoubleValueObserverSdk.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/metrics/DoubleValueObserverSdk.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2020, OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


sdk/src/main/java/io/opentelemetry/sdk/metrics/DoubleValueObserverSdk.java
```diff
@@ -0,0 +1,94 @@
...
+
+  DoubleValueObserverSdk(
+      InstrumentDescriptor descriptor,
+      MeterProviderSharedState meterProviderSharedState,
+      MeterSharedState meterSharedState) {
+    super(
+        descriptor,
+        meterProviderSharedState,
+        meterSharedState,
+        // TODO: Revisit the batcher used here, currently this does not remove duplicate records in
```
jkwatson, line 37: Please add an issue for this.

bogdandrutu: https://github.com/open-telemetry/opentelemetry-java/issues/1302

PR #1304, opened 2020-06-03
Immutable Attributes and Labels

jkwatson:
This introduces two new immutable classes, Attributes and Labels. They both implement a common parameterized abstract class, and share a bit of logic for the creation of an underlying array-based implementation.


api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -0,0 +1,251 @@
...
+      data.add(stringAttributeValue(value));
+      return this;
+    }
+
+    /**
+     * Add a long {@link AttributeValue} to this.
+     *
+     * @return this Builder
+     */
+    public Builder addAttribute(String key, long value) {
```
anuraaga, line 179: Thanks! The API in this PR LGTM. Only comment is may stick with `putAttribute` instead of `add`, more familiar since it's inspired by Java's `Map` and conveys that duplicates are deduped.

carlosalberto: I'm wondering if we should have `setAttribute()` instead, to align this with `Span.setAttribute()`. No strong feeling though.

jkwatson: I'm totally fine, either way. @anuraaga , what do you think about `setAttribute`, rather than `putAttribute` ?

trask: +1 for `setAttribute`, to align with `Span.setAttribute`

jkwatson: ok, I'll assume @anuraaga is ok, and if not, we can change it.

jkwatson:
One downside to the use of the forEach, pre-java 8, is that you have to create an instance of the consumer every time you want to iterate over the data. If the attributes are empty, you have to either a) check for emptiness [we would need to add this method], or create a consumer that is never used.  I guess the iterator is just as bad, though, since you need an instance of the iterator, plus you need the wrap all the entries.

At least with java 8+, this all gets better with lambdas.

jkwatson:
And, I just realized that with a Map, it has to create an iterator for you as well, so what we have here is no worse that a Map from a consumption side, and saves all the extra Entry allocations, so win!

api/src/main/java/io/opentelemetry/common/Attributes.java
jkwatson: done.

jkwatson:
Note: I got rid of the TreeMap sorting/de-duping and replaced it with a little quicksort implementation.

jkwatson:
A further note: I have a branch where I hooked this up to Links and Events to replace the Map usage, and this API worked quite well. I did need to add a `size()` and an `isEmpty()` method to the interface, so that'll be coming later.

The builder ended up being an excellent way to handle things in both the OpenTracing shim and in a couple other places. I imagine that we'll want to replace the usage of the AttributesMap with that Builder inside the RecordEventsReadableSpan, although that also has a "remove" method on it that I'm not sure what we'll do with.

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
```diff
@@ -0,0 +1,105 @@
...
+
+    static <T> List<Object> sortAndFilter(Object[] data) {
+      checkArgument(
+          data.length % 2 == 0, "You must provide an even number of key/value pair arguments.");
+
+      quickSort(data, 0, data.length - 2);
+      return dedupe(data);
+    }
+
+    private static void quickSort(Object[] data, int leftIndex, int rightIndex) {
```
carlosalberto, line 58: Guessing you wrote this you don't want to use `Arrays.sort()`, which uses merge sort (and more memory)? If that's the case, adding a note for further reference would be nice.

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * An immutable set of key-value pairs. Keys are only {@link String} typed. Can be iterated over
+ * using the {@link #forEach(KeyValueConsumer)} method.
+ *
+ * @param <V> The type of the values contained in this.
+ * @see Labels
+ * @see Attributes
+ */
+@Immutable
+public interface ImmutableKeyValuePairs<V> {
```
bogdandrutu, line 35: This can be package protected, no need to expose it public. Also why do you not make this an abstract class with all the utils inside it. It will remove some duplicate code between Labels and Attributes.

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
jkwatson: I don't know of a way to use Arrays.sort() to sort only based on every other value, and keep the pairs together.

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
jkwatson: I'll give it a try. Originally, I was hoping this could be a single-method-interface, and hence implementable as a lambda, but that won't be possible once we add `isEmpty()` and `size()`.

jkwatson: Also, if we make this interface package-protected, we'll need to move the consumer interface up out of it. Not a big deal, but it is a consequence.

api/src/main/java/io/opentelemetry/common/Attributes.java
anuraaga: Sorry for the late SGTM

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
carlosalberto: Oooh, right. I totally forgot about that part ;)

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
bogdandrutu: That can be public correct.

jkwatson: ok, update and pushed. PTaL

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
```diff
@@ -0,0 +1,104 @@
...
+        counter += 2;
+      }
+    }
+
+    quickSort(data, leftIndex, counter - 4);
+    quickSort(data, counter, rightIndex);
+  }
+
+  private static List<Object> dedupe(Object[] data) {
+    List<Object> result = new ArrayList<>(data.length);
```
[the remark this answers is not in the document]
anuraaga: Realized that if sort and dedupe were combined into one insertion sort could do it in-place too. Dupes should be rare I think so copying only for the dupe case would be nice.

api/src/main/java/io/opentelemetry/common/KeyValueConsumer.java
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.common;
+
+/**
+ * Used for iterating over the key-value pairs contained by an {@link ImmutableKeyValuePairs}
```
anuraaga, line 20: ImmutableKeyValuePairs isn't in public API so can't link it. Just some generic word like container should be fine.

api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -0,0 +1,239 @@
...
+  public static class Builder {
+    private final List<Object> data = new ArrayList<>();
+
+    /** Create the {@link Attributes} from this. */
+    public Attributes build() {
+      return sortAndFilterToAttributes(data.toArray());
+    }
+
+    /**
+     * Add a bare {@link AttributeValue} to this.
```
anuraaga, line 141: Sets instead of Add throughout (matches the method name and javadoc should generally be present tense, not command tense)

api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -0,0 +1,239 @@
...
+  @AutoValue
+  @Immutable
+  abstract static class ArrayBackedAttributes extends Attributes {
+    ArrayBackedAttributes() {}
+
+    @Override
+    abstract List<Object> data();
+  }
+
+  /** An {@link Attributes} instance with no attributes. */
```
anuraaga, line 47: `Returns a` for all the factories

api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -0,0 +1,239 @@
...
+import static io.opentelemetry.common.AttributeValue.longAttributeValue;
+import static io.opentelemetry.common.AttributeValue.stringAttributeValue;
+
+import com.google.auto.value.AutoValue;
+import java.util.ArrayList;
+import java.util.List;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * An immutable container for attributes. The type parameter denotes the type of the values of the
```
anuraaga, line 31: No type parameter anymore remove from javadoc

api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -0,0 +1,239 @@
...
+    return sortAndFilterToAttributes(
+        key1, value1,
+        key2, value2,
+        key3, value3,
+        key4, value4,
+        key5, value5);
+  }
+
+  private static Attributes sortAndFilterToAttributes(Object... data) {
+    return new AutoValue_Attributes_ArrayBackedAttributes(sortAndFilter(data));
```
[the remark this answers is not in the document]
jkwatson: The sorting is just for making the equals/hashcode work out right.

api/src/main/java/io/opentelemetry/common/KeyValueConsumer.java
jkwatson: fair enough. will update.

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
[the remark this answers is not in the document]
jkwatson: I think we can save fine-grained optimization like this until after we've got it all up and running. :)

api/src/main/java/io/opentelemetry/common/Attributes.java
jkwatson: whoops! got lost in the refactoring. will fix.

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
```diff
@@ -0,0 +1,104 @@
...
+    Object previousKey = null;
+
+    for (int i = 0; i < data.length; i += 2) {
+      Object key = data[i];
+      Object value = data[i + 1];
+      checkNotNull(key, "You cannot provide null keys for creation of attributes.");
```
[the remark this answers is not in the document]
jkwatson: hmm. null keys seems like a pretty bad problem. what do you suggest we do if some instrumentation puts in a null key?

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
```diff
@@ -0,0 +1,104 @@
...
+ * using the {@link #forEach(KeyValueConsumer)} method.
+ *
+ * @param <V> The type of the values contained in this.
+ * @see Labels
+ * @see Attributes
+ */
+@Immutable
+abstract class ImmutableKeyValuePairs<V> {
+
+  List<Object> data() {
```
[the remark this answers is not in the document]
jkwatson: I wish I could, but if we want to use AutoValue for the implementation, it needs a list. :(

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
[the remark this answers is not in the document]
anuraaga: I agree it's a bad problem but eventually some user will find their app crash because of observability logic and complain even in the most unlikely scenarios :) And I think that's the motivation behind those principles.

Just skipping of course makes finding these bugs hard - how about a configurable check that can be enabled in unit tests? https://github.com/aws/aws-xray-sdk-java/blob/master/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/utils/LooseValidations.java#L24

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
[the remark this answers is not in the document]
anuraaga: Agree though might still consider at least the already-sorted/deduped check in this one that would probably be good enough for most cases anyways and should be relatively simple.

api/src/main/java/io/opentelemetry/common/Attributes.java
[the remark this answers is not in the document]
bogdandrutu: I guess you can do the equals/hash in n^2 and will be as good as sorting + equals check for less than 10 labels. You may want to give a try and a benchmark

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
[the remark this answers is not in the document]
jkwatson: I actually think that the previous incarnation where this was an interface was easier to understand. Even though it ends up with like 6-8 lines of duplicated code, I think it's actually a better design to this be an interface.

api/src/main/java/io/opentelemetry/common/Attributes.java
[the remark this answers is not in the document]
jkwatson: I'd like to save this kind of micro-optimization until after the API lands, and we can parellelize the work with using this for the APIs and doing optimizations. On my branch where I use this for Events and Links, this is already a significant improvement (~10% better) in both speed and memory allocations than the existing implementation.

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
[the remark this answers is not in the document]
jkwatson: How about skipping and logging? I do think this will be a very rare occurrence, and will almost certainly indicate a programming bug.

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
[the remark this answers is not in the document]
jkwatson: See above comments on getting this landed before doing micro-optimizations.

api/src/main/java/io/opentelemetry/common/Attributes.java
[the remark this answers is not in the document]
bogdandrutu: I think just a dump n^2 is simpler than quicksort + dedup on sorted array.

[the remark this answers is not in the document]
jkwatson: ok to save that for a follow-on PR?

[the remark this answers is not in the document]
jkwatson: Just a thought on sorting...if we do need key-lookup in the future, having the keys sorted makes binary searching easy.

api/src/main/java/io/opentelemetry/common/Attributes.java
jkwatson: done

api/src/main/java/io/opentelemetry/common/Attributes.java
jkwatson: done

api/src/main/java/io/opentelemetry/common/Attributes.java
jkwatson: done

api/src/main/java/io/opentelemetry/common/KeyValueConsumer.java
jkwatson: done

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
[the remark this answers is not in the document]
jkwatson: ok, skipped null keys and logged.

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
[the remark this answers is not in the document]
jkwatson: this is going to change in the next PR, with the addition of 2 new methods. will address there.

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
jkwatson: i'm not super happy with the change to an abstract class, but we can address in the future.

jkwatson:
> @jkwatson waiting for you to respond for all outstanding comments :)

ok, got them all. thanks!

api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -0,0 +1,240 @@
...
+  @Immutable
+  abstract static class ArrayBackedAttributes extends Attributes {
+    ArrayBackedAttributes() {}
+
+    @Override
+    abstract List<Object> data();
+  }
+
+  /** Returns a {@link Attributes} instance with no attributes. */
+  public static Attributes empty() {
```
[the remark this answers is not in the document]
jkwatson: I actually didn't use that pattern, quite intentionally, since I think guava's use of the empty `of` is quite annoying and surprising. :)

api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -0,0 +1,240 @@
...
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * An immutable container for attributes.
+ *
+ * <p>The keys are {@link String}s and the values are {@link AttributeValue} instances.
+ */
+@Immutable
+public abstract class Attributes extends ImmutableKeyValuePairs<AttributeValue> {
+  private static final Attributes EMPTY = new Attributes() {};
```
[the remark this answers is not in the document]
jkwatson: Sure. I'm fine either way.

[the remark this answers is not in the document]
jkwatson: done.

PR #1309, opened 2020-06-04
Avoid mockito magic for span processors test, test with empty configmap

bogdandrutu:
Main motivation is to add tests for empty config map, so need to get access to the "default" values and getters, then no need to use mockito for the case where the map is not empty.


jkwatson:
Yeah, I much prefer exposing some getters for immutable data to using mockito's spy functionality.  👍

PR #1317, opened 2020-06-08
Allow setting semantic attributes on Span.Builder

### LINKED ISSUE #1316: Extend `XXXAttributeSetter` API to accept `Span.Builder`
#758 introduced `SemanticAttributes` class which allows for type- and semantic-safe way to set span attributes. It would be amazing to have the same functionality for `Span.Builder` as well. As [it is encouraged](https://github.com/open-telemetry/opentelemetry-specification/blob/4ab7c2fb4964c14fe972b03b5bb6825573bafb24/specification/api-tracing.md#span-creation) to set attributes on a builder whenever possible.

iNikem:
Fixes #1316

### DIFF
--- a/api/src/main/java/io/opentelemetry/trace/attributes/StringAttributeSetter.java
+++ b/api/src/main/java/io/opentelemetry/trace/attributes/StringAttributeSetter.java
@@ -61,4 +61,14 @@ public final class StringAttributeSetter {
   public void set(Span span, @Nullable String value) {
     span.setAttribute(key(), value);
   }
+
+  /**
+   * Sets the attribute on the provided span builder.
+   *
+   * @param spanBuilder the span builder to add the attribute to
+   * @param value the value for this attribute
+   */
+  public void set(Span.Builder spanBuilder, @Nullable String value) {
+    spanBuilder.setAttribute(key(), value);
+  }
 }
--- a/api/src/test/java/io/opentelemetry/trace/attributes/SemanticAttributesTest.java
+++ b/api/src/test/java/io/opentelemetry/trace/attributes/SemanticAttributesTest.java
@@ -34,11 +34,13 @@ import org.junit.runners.JUnit4;
 public class SemanticAttributesTest {
 
   private Span span;
+  private Span.Builder builder;
 
   @Before
   public void setUp() {
     Tracer tracer = OpenTelemetry.getTracer("io.telemetry.api");
     span = tracer.spanBuilder("junit").startSpan();
+    builder = tracer.spanBuilder("junit-builder");
   }
 
   @Test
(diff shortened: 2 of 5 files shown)


iNikem:
I totally admit that this PR is meant to solve my immediate problem in instrumentations repo :)

iNikem:
Anybody be kind and merge this PR? :)

api/src/main/java/io/opentelemetry/trace/attributes/StringAttributeSetter.java
```diff
@@ -61,4 +61,14 @@ public String key() {
...
   }
+
+  /**
+   * Sets the attribute on the provided span builder.
+   *
+   * @param spanBuilder the span builder to add the attribute to
+   * @param value the value for this attribute
+   */
+  public void set(Span.Builder spanBuilder, @Nullable String value) {
+    spanBuilder.setAttribute(key(), value);
```
jkwatson, line 72: would it make sense to return this, or the builder from this call? Or does that not help make a more fluent API?

jkwatson: I guess it doesn't make any sense to return `this`, but would returning the Builder be at all useful?

jkwatson: And, I see that @Oberon00 already made this comment above. sorry for the extra noise.

iNikem: `SemanticAttributes` are now used like this: `SemanticAttribute.HTTP_URL.set(span, value). No quick way to have fluent interface out of this.

api/src/test/java/io/opentelemetry/trace/attributes/SemanticAttributesTest.java
```diff
@@ -51,15 +53,20 @@ public void shouldEnableSetAttributeOnSpan() throws IllegalAccessException {
...
         ((LongAttributeSetter) attribute).set(span, 42L);
+        ((LongAttributeSetter) attribute).set(builder, 42L);
       } else if (attribute instanceof DoubleAttributeSetter) {
         keys.add(((DoubleAttributeSetter) attribute).key());
         ((DoubleAttributeSetter) attribute).set(span, 3.14);
+        ((DoubleAttributeSetter) attribute).set(builder, 3.14);
       } else if (attribute instanceof BooleanAttributeSetter) {
         keys.add(((BooleanAttributeSetter) attribute).key());
         ((BooleanAttributeSetter) attribute).set(span, true);
+        ((BooleanAttributeSetter) attribute).set(builder, true);
```
jkwatson, line 69: This test doesn't actually test anything except that you can loop over the fields of the class. Can we add some actual assertions about the functionality, rather than just making calls that don't test anything?

jkwatson:
> Anybody be kind and merge this PR? :)

I'd like to see the test changed so that it actually tests something before merging this.

api/src/test/java/io/opentelemetry/trace/attributes/SemanticAttributesTest.java
iNikem: Current test indeed only tested that you can call `set` method on every semantic attribute without any errors. But neither `Span` nor `Span.Builder` provides any API for reading attribute values. What else can you test here?

arminru: You would usually mock the builder using Mockito and verify that `builder.setAttribute` was called once with `key()` and the provided value.

jkwatson: You can additionally mock the Span instance and do the same.

carlosalberto: To tell you the truth, we don't do this at all for existing classes (in `SemanticAttributeTest` that is) and thus I feel we should do it in a follow-up PR where we apply the mocking for all of them.

jkwatson: Please create an issue for it, then. We need to make sure our tests are really useful. This isn't going to be available to the instrumentation project until the next release, anyway, so I'm not sure why we're in such a hurry to get this merged.

iNikem: We have a branch against otel-java snapshots :)

iNikem: I can do verification of mock calls, no problem.

iNikem: @jkwatson take a look now, please

jkwatson:
@iNikem Just a heads-up that the project is switching over to EasyCLA, so if you could get that CLA signed soon, that would be great.

iNikem:
> @iNikem Just a heads-up that the project is switching over to EasyCLA, so if you could get that CLA signed soon, that would be great.

Yeah, my manager is dealing with that... :(

PR #1318, opened 2020-06-08
Add an option to specify a Span Kind via the WithSpan annotation.

### LINKED ISSUE #1303: Add the ability to specify a Span Kind using the WithSpan annotation
Add an optional parameter to the `WithSpan` annotation that specifies the span Kind to create.

Doing this manually is possible, but it would be nice to have it built in to the annotation.

jkwatson:
resolves #1303 for the API. Will open an issue with the instrumentation project when this is merged.


trask:
I think I'd prefer for the annotations package to not have any dependencies, and instead have it's own `SpanKind` enum.

@iNikem @tylerbenson wdyt?

iNikem:
I don't see any downsides in depending on API module. And duplicating means syncing, I don't think it is worth the effort.

jkwatson:
@trask  What about specifically excluding grpc-context from the dependencies of the annotations?

jkwatson:
> @trask What about specifically excluding grpc-context from the dependencies of the annotations?

I added this to the PR.

contrib/auto_annotations/build.gradle
```diff
@@ -9,7 +9,7 @@ description = 'OpenTelemetry Contrib Auto Annotations'
 ext.moduleName = "io.opentelemetry.contrib.auto.annotations"
 
 dependencies {
-    implementation project(':opentelemetry-api')
+    implementation project(':opentelemetry-api') exclude group: 'io.grpc', module: 'grpc-context'
```
jkwatson, line 12: @bogdandrutu @carlosalberto any issues with this late-breaking change to minimize the dependencies of the annotations module?

bogdandrutu: First request: Let's do it in a separate PR :)
Second request: I need to understand what you try to achieve :)

jkwatson: 1. sure thing; that's why I asked
2. @trask is concerned that people will need to pull in grpc-context just to use an annotation, and would prefer that not to be the case.

jkwatson: I removed the change from this PR, FYI

jkwatson: I removed the change from this PR, FYI

PR #1323, opened 2020-06-09
[Draft] Initial System metrics proposal

carlosalberto:
This is an initial draft PR for having an SDK-contrib library that automatically reports a few system metrics (initial implementations for other languages exist, such as [Python](https://github.com/open-telemetry/opentelemetry-python/tree/master/ext/opentelemetry-ext-system-metrics)) using observers.

The [Oshi](https://github.com/oshi/oshi) library is used to gather hardware information, and the reported metrics are (in either `bytes` or `seconds`, depending on the metric itself):

*  System memory (total and available).
* System CPU (system, user time and idle time)
* System Network Bytes (sent and received)
* Java Process memory (rss and vms)
* Java Process CPU (system and user time)
* Java GC (only GC count at this moment).

Internally the class is creating its own `IntervalMetricReader`. Also, these metric names will be updated when we have the related semantic conventions.

Tests, configuration support and code lifting code (such as maybe moving the small `Callback` code to their own file) is needed, but wanted to get initial feedback on this feature.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -137,6 +137,7 @@ configure(opentelemetryProjects) {
                 jmh_core                : "org.openjdk.jmh:jmh-core:${jmhVersion}",
                 jmh_bytecode            : "org.openjdk.jmh:jmh-generator-bytecode:${jmhVersion}",
                 jsr305                  : "com.google.code.findbugs:jsr305:${findBugsJsr305Version}",
+                oshi                    : "com.github.oshi:oshi-core:5.1.1",
                 prometheus_client       : "io.prometheus:simpleclient:${prometheusVersion}",
                 prometheus_client_common: "io.prometheus:simpleclient_common:${prometheusVersion}",
                 protobuf                : "com.google.protobuf:protobuf-java",
--- a/sdk_contrib/system_metrics/src/main/java/io/opentelemetry/sdk/contrib/metrics/system/SystemMetrics.java
+++ b/sdk_contrib/system_metrics/src/main/java/io/opentelemetry/sdk/contrib/metrics/system/SystemMetrics.java
@@ -24,8 +24,10 @@ import io.opentelemetry.metrics.AsynchronousInstrument.LongResult;
 import io.opentelemetry.metrics.Meter;
 import io.opentelemetry.sdk.metrics.export.IntervalMetricReader;
 import io.opentelemetry.sdk.metrics.export.MetricExporter;
+import io.opentelemetry.sdk.metrics.export.MetricProducer;
 import java.lang.management.GarbageCollectorMXBean;
 import java.lang.management.ManagementFactory;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
 import java.util.TreeMap;
@@ -44,13 +46,17 @@ public final class SystemMetrics {
   private final IntervalMetricReader intervalReader;
 
   private SystemMetrics(
-      Meter meter, MetricExporter exporter, Map<String, String> labels, int interval) {
+      Collection<MetricProducer> producers,
+      MetricExporter exporter,
+      Map<String, String> labels,
+      int interval) {
 
-    registerObservers(meter, labels);
+    registerObservers(labels);
     intervalReader =
         IntervalMetricReader.builder()
             .setExportIntervalMillis(interval)
             .setMetricExporter(exporter)
+            .setMetricProducers(producers)
             .build();
   }
 
@@ -58,7 +64,8 @@ public final class SystemMetrics {
     intervalReader.shutdown();
   }
 
-  void registerObservers(Meter meter, Map<String, String> labels) {
+  void registerObservers(Map<String, String> labels) {
+    Meter meter = OpenTelemetry.getMeterProvider().get(SystemMetrics.class.getName());
     SystemInfo systemInfo = new SystemInfo();
     OperatingSystem osInfo = systemInfo.getOperatingSystem();
 
@@ -176,14 +183,14 @@ public final class SystemMetrics {
 
   /** Add me. */
   public static final class Builder {
-    private Meter meter = OpenTelemetry.getMeterProvider().get(SystemMetrics.class.getName());
+    private Collection<MetricProducer> producers;
     private MetricExporter exporter;
     private Map<String, String> labels;
     private int interval;
 
     /** Add me. */
-    public Builder setMeter(Meter meter) {
-      this.meter = meter;
+    public Builder setMetricProducers(Collection<MetricProducer> producers) {
+      this.producers = producers;
       return this;
     }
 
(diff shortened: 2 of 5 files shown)


jkwatson:
Shouldn't this be a pure API contribution? Or be a part of the instrumentation project?

carlosalberto:
> why do you need the ask dependency?

Not sure what you mean. Can you elaborate @bogdandrutu ?

carlosalberto:
> Shouldn't this be a pure API contrib-module? Or be a part of the instrumentation project?

That's a good question. We need the SDK only for creating the `IntervalMetricReader`, but otherwise fine - precisely this overlaps with whether you think this should end up in the Instrumentation repo (probably YES). @trask @tylerbenson you guys have an opinion on this?

trask:
> Or be a part of the instrumentation project?

makes sense to me

build.gradle
```diff
@@ -137,6 +137,7 @@ configure(opentelemetryProjects) {
                 jmh_core                : "org.openjdk.jmh:jmh-core:${jmhVersion}",
                 jmh_bytecode            : "org.openjdk.jmh:jmh-generator-bytecode:${jmhVersion}",
                 jsr305                  : "com.google.code.findbugs:jsr305:${findBugsJsr305Version}",
+                oshi                    : "com.github.oshi:oshi-core:5.1.1",
```
anuraaga, line 140: JNA libraries tend to be pretty bloated since they basically duplicate a native interface into Java, and for cross-platform ones have a ton of these. This seems to be 600K along with JNA itself which is 3MB+ (SDK is just 250K for reference). JDK's monitoring beans provide a lot of useful information so maybe just focus on them for now and can consider adding the huge dependency in a followup if the delta in terms of metrics seems useful?

anuraaga: I guess if it's just about system metrics this could be helpful. But it seems more useful to users to focus on JVM metrics first? None of the popular Java metrics libraries seem to use oshi but provide very useful runtime metrics without much binary cost - such metrics should basically "always be enabled" on an instrumented app while this seems much more specialized and should just add the metrics that can't be gotten for free from the JDK.

https://github.com/dropwizard/metrics/tree/release/4.1.x/metrics-jvm/src/main/java/com/codahale/metrics/jvm
https://github.com/prometheus/client_java/tree/master/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot
https://github.com/micrometer-metrics/micrometer/tree/b486e1261cade606eb0ff826fbae318f6d0c9e86/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system

carlosalberto: >  this seems much more specialized and should just add the metrics that can't be gotten for free from the JDK.

That's correct, and this is the precisely the idea: to provide a set of complete system-wide metrics (and was the reason why I wanted to initially put it here, as a separate, extra artifact that some users can use). 

@trask is there a plan to support external instrumentation jars? In that case, this could be provided at runtime. Also, what's your take on having Oshi as a dependency?

anuraaga: Many of the metrics here, like system cpu load, are available from the JDK though. And we definitely wouldn't want double instrumentation of the same metric if people added both, e.g., `instrumentation-jvm` and `instrumentation-system`. So recommendation was to start with the less complex, free stuff since that gives a better picture of what's needed in the complex, dependency-heavy version. So this feels like it's the wrong order to be addressing the issue in IMO. Does that make sense?

carlosalberto: This will be a rather small codebase, so I don't see the need to postpone the discussion on what to use/report ;) I'd argue that it's not a wrong order situation, as I mentioned before, the idea is precisely to provide system-wide metrics.

> we definitely wouldn't want double instrumentation of the same metric if people added both

It's an interesting scenario to have them separated. But in that case I'd rather have the system-wide disabled or something similar. Unless there's a clear path that justifies having them separated (in which case I'd be fine with the separation).

sdk_contrib/system_metrics/src/main/java/io/opentelemetry/sdk/contrib/metrics/system/SystemMetrics.java
```diff
@@ -0,0 +1,221 @@
...
+        .setUnit("seconds")
+        .setConstantLabels(labels)
+        .build()
+        .setCallback(
+            new Callback<DoubleResult>() {
+              @Override
+              public void update(DoubleResult r) {
+                processInfo.updateAttributes();
+                r.observe(processInfo.getUserTime() * 1000, TYPE_LABEL_KEY, "user");
+                r.observe(processInfo.getKernelTime() * 1000, TYPE_LABEL_KEY, "system");
```
trask, line 159: JVM mbeans provide process cpu as a single number (i assume `system` + `user`?), no breakdown of `system`/`user` that i know of

build.gradle
trask: > @trask is there a plan to support external instrumentation jars?

ya, we need to support users writing/bringing their own instrumentation somehow, but we haven't gotten to that yet

> Also, what's your take on having Oshi as a dependency?

well, the auto agent jar is already 28mb, so what's another 3mb? 😭

(side note: maybe we should use the `all` classifier after all, to be clear that it's the kitchen sink of distributions)

I agree that it would be nice to split this into two modules, one that only has only JVM (mbean) metrics, and one that has only system (oshi) metrics.

The JVM metric module could detect if the system metrics module is on the classpath and disable anything that's duplicative with the system metrics (e.g. don't capture cpu `total` via JVM metrics, since system metrics captures cpu `system` + `user`).

anuraaga: Having the jvm module disable metrics if oshi is being used is a nice idea.

Size-wise, let's not forget we are trying to move towards also supporting manual instrumentation, not just agent. Though 10% of the entire agent seems pretty hefty FWIW. System metrics also have the issue that multiple jvms on the same system will also provide redundant metrics, common with containers. That should be taken into account too when evaluating the cost, it does take actual RAM to load large ffi modules. I would expect of packaged into the agent jar it's not enabled by default because of this potentially exploding cost to users.

That's enough high level context, for this PR how about removing the GC metrics and only take stuff from oshi? Can add GC metrics to a separate jvm metrics module

anuraaga: Ah found that we actually already have such a class too

https://github.com/open-telemetry/opentelemetry-java/blob/596617dcbe7763435573a3d02eba4a5986f7dcd8/contrib/runtime_metrics/src/main/java/io/opentelemetry/contrib/metrics/runtime/GarbageCollector.java#L66

carlosalberto: > I agree that it would be nice to split this into two modules, one that only has only JVM (mbean) metrics, and one that has only system (oshi) metrics.

That's fine by me, as long as one module is for JVM-specific metrics (the current `metrics` module in contrib, with a few metric additions) and have this one is for System-wide metrics (which could be disabled by default). However, skipping the Oshi dependency wouldn't be avoided ;) 

Btw, related effort: https://github.com/open-telemetry/opentelemetry-specification/issues/651

carlosalberto:
Closing as we will be moving this one to the java-instrumentation repo.

PR #1325, opened 2020-06-09
Convert Link and Event types to use the new immutable Attributes

jkwatson:


sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -362,24 +363,23 @@ public void addEvent(io.opentelemetry.trace.Event event, long timestamp) {
...
-        temp.put(entry.getKey(), entry.getValue());
-      }
+  static Attributes copyAndLimitAttributes(final Attributes attributes, final int limit) {
+    if (attributes.isEmpty() || attributes.size() <= limit) {
+      return attributes;
     }
 
-    return Collections.unmodifiableMap(temp);
+    final Attributes.Builder temp = Attributes.newBuilder();
+    final AtomicInteger added = new AtomicInteger();
```
bogdandrutu, line 372: why atomic? I think I know why because you need to change it in the lambda. You can have a size() on the builder maybe? Maybe create an issue because this smells a bit :)

carlosalberto:
Overall LGTM, other than the atomic part ;)

sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
jkwatson: yes, entirely because it needed to be updated in the inner class. `size()` on the builder isn't a bad idea at all, though. 

I will note that we're going to need a more sophisticated builder-like class to handle the attributes on the active span itself, so maybe we can roll the cleanup into that change?

sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -362,24 +362,14 @@ public void addEvent(io.opentelemetry.trace.Event event, long timestamp) {
...
-      if (temp.size() < limit) {
-        temp.put(entry.getKey(), entry.getValue());
-      }
+  static Attributes copyAndLimitAttributes(final Attributes attributes, final int limit) {
+    if (attributes.isEmpty() || attributes.size() <= limit) {
+      return attributes;
     }
 
-    return Collections.unmodifiableMap(temp);
+    Attributes.Builder temp = Attributes.newBuilder();
```
anuraaga, line 370: Would usually use something like `result`, not used to temp being as important as this one :)

sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -534,4 +524,22 @@ int getTotalRecordedLinks() {
...
+    private int added;
+
+    public LimitingAttributeConsumer(int limit, Attributes.Builder builder) {
+      this.limit = limit;
+      this.builder = builder;
+    }
+
+    @Override
+    public void consume(String key, AttributeValue value) {
+      if (added++ < limit) {
```
anuraaga, line 540: This works because we don't expect any attributes to have more than `Integer.MAX_VALUE / 2` size (where added will become negative and elements start being appended again), but even so it's probably clearer keep the logic precise instead of the increment shortcut. That said, this might go away soon anyways so no big deal just FYI.

```
if (added < limit) {
  buileder.set(...)
  added++;
}

jkwatson: Are you suggesting the change for a functional reason, or just for clarity? what you suggest is identical functionally. 

I don't understand the Integer.MAX_VALUE comment here. Attributes limits should never come anywhere close to that, so I have zero concern around overflowing the int. Do you have a use-case that I'm not thinking of?

sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
jkwatson: agreed. will pick this up in the next round of changes.

sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
anuraaga: Mostly just for clarity - the fact that attribute limits should never come close is true but also some hidden knowledge needed to know the code works. It's nice to reduce hidden knowledge if it's not a burden, I guess one extra line of code isn't so bad. Also allows the `added` variable to actually be true since the value stops being how many were added, but how many were consumed, after the limit.

Very small nits, just added them for future reference.

jkwatson: ah, yes, I do agree with all of this. I'll be sure to include the cleanup in a future PR

PR #1330, opened 2020-06-10
Convert SpanData to use immutable Attributes

jkwatson:


sdk/src/test/java/io/opentelemetry/sdk/trace/TestUtils.java
```diff
@@ -124,4 +125,32 @@ public static SpanData makeBasicSpan() {
       tracerSdkFactory.updateActiveTraceConfig(originalConfig);
     }
   }
+
+  /** Find an AttributeValue by the key. */
+  public static AttributeValue findByKey(Attributes attributes, String key) {
```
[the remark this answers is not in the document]
jkwatson: I decided to add it, and keep it a simple linear search for now. We can revert if everyone hates it.

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
```diff
@@ -56,6 +57,22 @@ public void forEach(KeyValueConsumer<V> consumer) {
     }
   }
 
+  /**
+   * Returns the value of the given key, or null if the key does not exist.
+   *
+   * <p>Warning: currently implemented via a linear search, so O(n) performance.
```
anuraaga, line 63: +1 to linear search, these should always be small enough where it's faster and opens up optimizations like deferring / not sorting.

api/src/test/java/io/opentelemetry/common/AttributesTest.java
```diff
@@ -149,4 +149,41 @@ public void builder_arrayTypes() {
...
+
+  @Test
+  public void get_Null() {
+    assertThat(Attributes.empty().get("foo")).isNull();
+    assertThat(Attributes.of("key", stringAttributeValue("value")).get("foo")).isNull();
+  }
+
+  @Test
+  public void get() {
+    assertThat(Attributes.of("key", stringAttributeValue("value")).get("key"))
```
anuraaga, line 161: For `Attributes`, having `getString` etc would make these sort of unit tests (thinking user code unit tests less worried about these speifically) nicer.

jkwatson: Not a bad idea. What behavior would you expect if you were asking for a non-string attribute value?

anuraaga: I've seen similar APIs return `null` when it's a different type, it seems fine to me.

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
jkwatson: I actually implemented a binary search, then backed it out because it seemed like total overkill since we're not expecting many attributes. Good exercise to remind myself of some undergrad CS stuff, though (which I never took, since I don't have a CS degree!).

sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -516,13 +515,15 @@ int getTotalRecordedLinks() {
   }
 
   @GuardedBy("lock")
-  private Map<String, AttributeValue> getImmutableAttributes() {
+  private Attributes getImmutableAttributes() {
```
bogdandrutu, line 518: Looking at this I see a lot of allocations happening here. Do you have a plan to change AttributeMap to implement the Attributes so we don't reconstruct the entire map when converting to SpanData?

jkwatson: I would definitely like to tackle this next. So far, I'm a bit stumped on how to approach it, since the AttributeMap is extremely mutable, and Attributes are inherently immutable. Do you have any ideas of how to tackle this one?

api/src/test/java/io/opentelemetry/common/AttributesTest.java
jkwatson: Right now, the corresponding methods on AttributeValue throw an UnsupportedOperationException. I honestly don't know what's the better approach.

sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
bogdandrutu: AttributeMap becomes immutable after end event so we can reuse that object

jkwatson: ok, I've been noodling on this a bit.
If we made an exportable-attributes interface (no idea what the name of this could be), then both our existing Attributes abstract class and the AttributesMap class could both implement that interface. Then, SpanData would expose this interface, rather than the concrete Attributes API class. Is this the sort of thing you're also thinking?

jkwatson: Naming wise, here is my suggestion: 

* Rename the current `Attributes` class to `ImmutableAttributes`, keeping the implementation and the static creation methods there.
* Create a new interface called `Attributes` which will be used as the API interface and for export as well.
* Make the AttributesMap implement the Attributes interface
* Do as you suggest and pass in the AttributeMap directly to the SpanWrapper after the span has ended.

Does this sound right?

jkwatson: Note, I tried this out in a branch, and for the API it ends up feeling parallel to the Map/ImmutableMap APIs :  Attributes/ImmutableAttributes.

api/src/test/java/io/opentelemetry/common/AttributesTest.java
carlosalberto: Observe this is not an API intended to be used by end users, but by 'us', so I'm not so sure the sugar is truly needed.

In any case, if this is something still required, I'd suggest filling an issue and discuss the change there, so we can go ahead and merge this PR soon ;)

sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
carlosalberto: A small question: will `ImmutableAttributes` implement `Attributes` as well?

jkwatson: yes, ImmutableAttributes and AttributesMap will both implement the Attributes interface.

api/src/test/java/io/opentelemetry/common/AttributesTest.java
jkwatson: fair enough, @carlosalberto . I don't think it's needed right now, although it could make exporters easier to write, possibly.

jkwatson: Yeah, we shouldn't do this now; and only add it if it is really needed. Resolving this conversation for now.

jkwatson:
Is there anything blocking reviews on this? I'd like to keep moving on this to get it finished up soon.

api/src/test/java/io/opentelemetry/common/AttributesTest.java
anuraaga: Ok with addressing later. Don't quite understand this being not an end-user API since it would be used to create Resource too?

jkwatson:
> Ok with addressing later. Don't quite understand this being not an end-user API since it would be used to create Resource too?

I think the point was that, in general, API users won't need to be asking what attributes are in a set of attributes. They're writing those values, not reading them. The same applies, I believe, to the Resource.

carlosalberto:
> I'd like to keep moving on this to get it finished up soon.

Shall we merge this and have your upcoming changes in another PR? Is this what you meant? ;)

jkwatson:
> > I'd like to keep moving on this to get it finished up soon.
> 
> Shall we merge this and have your upcoming changes in another PR? Is this what you meant? ;)

Yes. I wasn't planning on changing the interface stuff until the next PR. I could do it all in one bigger PR, but I was trying to limit the scope, so as to not muddy the waters. :)

carlosalberto:
SGTM

PR #1333, opened 2020-06-11
Make the IntervalMetricReader robust to exporter failures

### LINKED ISSUE #1324: IntervalMetricReader should protect itself against exporters throwing exceptions
If a MetricExporter throws an exception, the IntervalMetricReader does not report the exception, and the scheduled thread exception will be halted.

The IMR should make sure to catch and report exceptions thrown by exporters.

jkwatson:
resolves #1324

### DIFF
--- a/sdk/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
@@ -29,6 +29,8 @@ import java.util.Map;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import javax.annotation.concurrent.Immutable;
 
 /**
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
```diff
@@ -202,11 +206,15 @@ private Exporter(InternalState internalState) {
 
     @Override
     public void run() {
-      List<MetricData> metricsList = new ArrayList<>();
-      for (MetricProducer metricProducer : internalState.getMetricProducers()) {
-        metricsList.addAll(metricProducer.getAllMetrics());
+      try {
+        List<MetricData> metricsList = new ArrayList<>();
+        for (MetricProducer metricProducer : internalState.getMetricProducers()) {
```
carlosalberto, line 211: Do we need this section to be in this `try` block too?

jkwatson: Perhaps not, but if anything in here fails, we don't want the IMR to die, so it felt safer just to put everything in the block.

PR #1336, opened 2020-06-15
Introduce an ReadableAttributes interface, with 2 implementations

jkwatson:
This interface is used for the export pipeline via SpanData, with 2 separate implementations, depending on the state of the attributes.


bogdandrutu:
I think having an interface like this is not ideal. We cannot control that user will use the implementation you expect to be used when recording it in Event or Links for example. So you either change all public APIs to use ImmutableAttributes or we name the interface "AttrbituesInterface" or other name and make the current Attributes == ImmutableAttributes.

api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -16,225 +16,28 @@
...
-
-import com.google.auto.value.AutoValue;
-import java.util.ArrayList;
-import java.util.List;
-import javax.annotation.concurrent.Immutable;
+import javax.annotation.Nullable;
 
 /**
- * An immutable container for attributes.
+ * An read-only container for String-keyed attributes.
```
carlosalberto, line 22: * A read-only

jkwatson:
> I think having an interface like this is not ideal. We cannot control that user will use the implementation you expect to be used when recording it in Event or Links for example. So you either change all public APIs to use ImmutableAttributes or we name the interface "AttrbituesInterface" or other name and make the current Attributes == ImmutableAttributes.

Yeah, not being able to rely on the immutable implementation on the API surface is a bummer. 

How about:
1) Keep `Attributes` for the immutable API-facing implementation, and keep the API only using that concrete implementation.
2) Call the interface `ReadableAttributes`, and have that be what the `SpanData` has assigned to it, and hence what exporters get to work with. The other option would be to have the interface still be called `Attributes`, but put it in a different package, but that's pretty ugly, IMO.

I think I can make this work without too much trouble.

jkwatson:
ok! Ready for another look.

api/src/main/java/io/opentelemetry/common/Attributes.java
jkwatson: fixed!

api/src/main/java/io/opentelemetry/common/ReadableAttributes.java
```diff
@@ -0,0 +1,44 @@
...
+  /** Whether there are any attributes contained in this. */
+  boolean isEmpty();
+
+  /** Iterates over all the key-value pairs of attributes contained by this instance. */
+  void forEach(KeyValueConsumer<AttributeValue> consumer);
+
+  /**
+   * Returns the value of the given key, or null if the key does not exist.
+   *
+   * <p>Warning: currently may be implemented via a linear search, depending on implementation, so
```
anuraaga, line 39: Think we can remove this warning now. Seems too scary for what it actually is

jkwatson: Be afraid, be very afraid.

carlosalberto: I'd be up for leaving the comment, but without the "Warning" word, to not scare people (yet have they understood the operation price ;) )

api/src/test/java/io/opentelemetry/common/ImmutableAttributesTest.java
```diff
@@ -31,7 +31,7 @@
 import org.junit.rules.ExpectedException;
 
 /** Unit tests for {@link Attributes}s. */
-public class AttributesTest {
+public class ImmutableAttributesTest {
```
carlosalberto, line 34: This should stay as `AttributesTest`, no?

examples/metrics/src/main/java/io/opentelemetry/example/metrics/LongValueObserverExample.java
```diff
@@ -21,9 +21,9 @@ public static void main(String[] args) {
         .build();
 
     observer.setCallback(
-        new LongValueObserver.Callback<LongValueObserver.ResultLongValueObserver>() {
+        new LongValueObserver.Callback<LongResult>() {
```
carlosalberto, line 24: How did this end up here? ;)

carlosalberto:
Overall good, I like this approach much better. LGTM when the remaining issues are solved ;)

examples/metrics/src/main/java/io/opentelemetry/example/metrics/LongValueObserverExample.java
jkwatson: I have the examples project wired up in my IDE, so I have to fix failures to be able to compile. I can move this to a separate PR, if you prefer.

api/src/test/java/io/opentelemetry/common/ImmutableAttributesTest.java
jkwatson: Yep, will fix.

api/src/main/java/io/opentelemetry/common/ReadableAttributes.java
jkwatson: Fixed.

api/src/test/java/io/opentelemetry/common/ImmutableAttributesTest.java
jkwatson: fixed

examples/metrics/src/main/java/io/opentelemetry/example/metrics/LongValueObserverExample.java
jkwatson: removed. will submit separately

PR #1338, opened 2020-06-16
Convert the usages of the LabelSetSdk to use the Labels class.

jkwatson:
After this, we can consider if we want to change the metric APIs to use Labels, rather than varargs of String.


api/src/main/java/io/opentelemetry/common/Labels.java
```diff
@@ -102,6 +102,10 @@ public static Labels of(
         key5, value5);
   }
 
+  public static Labels of(String[] keyValueLabelPairs) {
+    return sortAndFilterToLabels((Object[]) keyValueLabelPairs);
+  }
+
   private static Labels sortAndFilterToLabels(Object... data) {
     return new AutoValue_Labels_ArrayBackedLabels(sortAndFilter(data));
   }
```
bogdandrutu, line 111: One of this 2 can be removed.

exporters/prometheus/src/main/java/io/opentelemetry/exporters/prometheus/MetricAdapter.java
```diff
@@ -105,10 +105,18 @@ static Type toMetricFamilyType(MetricData.Descriptor.Type type) {
...
+      descriptor
+          .getConstantLabels()
+          .forEach(
+              new KeyValueConsumer<String>() {
+                @Override
+                public void consume(String key, String value) {
+                  finalConstLabelNames.add(toLabelName(key));
+                  finalConstLabelValues.add(value == null ? "" : value);
+                }
+              });
```
bogdandrutu, line 119: Please create an issue to revisit this logic and assign to me. seems we do unnecessary work.

api/src/main/java/io/opentelemetry/common/Labels.java
jkwatson: yeah, it's a good point. If we move the metric APIs to use Labels explicitly, rather than varargs strings, I think we can get rid of the one I just added, altogether. I'll clean this up for this PR, though, since there no reason to have the extra method there for now.

exporters/prometheus/src/main/java/io/opentelemetry/exporters/prometheus/MetricAdapter.java
jkwatson: https://github.com/open-telemetry/opentelemetry-java/issues/1341

api/src/main/java/io/opentelemetry/common/Labels.java
bogdandrutu: We can do in a follow-up. Up to you to decide

jkwatson: yeah, let's do it in a follow-up. It's actually a bunch messier than I first suspected, due to some fun java reasons. :)

PR #1339, opened 2020-06-17
MultiTracePropagator implementation

carlosalberto:
Prototype for supporting multiple Trace formats properly, which is specially handy when the user wants to support a few formats upon extraction (W3C, Jaeger, B3, etc). Intended to solve https://github.com/open-telemetry/opentelemetry-specification/issues/496

Part of the change is having `Propagator` instances NOT set anything when extraction failed (currently we set `DefaultSpan.getInvalid()`). This may 'leak' lingering, old active `Span`s, but this is an error in itself - on the other hand, it means we will be creating less `Context` objects (upon failed extraction, that is).

Also, there's the consideration of whether we need this as a general-purpose propagator, or specific for `tracing` ;)

The interesting piece of code is this:

```java
  public <C> Context extract(Context context, C carrier, Getter<C> getter) {
    for (int i = 0; i < propagators.length; i++) {
      context = propagators[i].extract(context, carrier, getter);
      if (TracingContextUtils.getSpanWithoutDefault(context) != null) {
        break;
      }
    }

    return context;
  }
```

Related to #1273 

PS - Ignore the code style, lack of builder, etc. Will update that once we find agreement ;)


extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/StackPropagator.java
```diff
@@ -0,0 +1,68 @@
...
+    for (int i = 0; i < propagators.length; i++) {
+      propagators[i].inject(context, carrier, setter);
+    }
+  }
+
+  @Override
+  public <C> Context extract(Context context, C carrier, Getter<C> getter) {
+    for (int i = 0; i < propagators.length; i++) {
+      context = propagators[i].extract(context, carrier, getter);
+      if (TracingContextUtils.getSpanWithoutDefault(context) != null) {
```
jkwatson, line 61: This call is a bit opaque. Is there a more readable way to do this work? Even just a method in this class called something like "isContextUsable" would help a lot.

jkwatson:
This seems like a useful addition. 

Another thing I would like: A compound propagator that allowed extraction of a variety of formats, but would only inject a single format (like W3C, for example). This would enable people to start standardizing on one format, but still allow backward compatibility until they've got everything updated.

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/StackPropagator.java
carlosalberto: We should document that, in general, in absence of any extracted `SpanContext` in `Context` we will try the next `Propagator`, and that should be enough IMO. 

Also, `TracingContextUtils` and its `getSpanWithoutDefault()` are rather well known methods.

carlosalberto:
> Another thing I would like: A compound propagator that allowed extraction of a variety of formats, but would only inject a single format (like W3C, for example). This would enable people to start standardizing on one format, but still allow backward compatibility until they've got everything updated.

That's an interesting take. I'm wondering about the trade-offs, but in general sounds like a good backwards compatibility compromise.

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/StackPropagator.java
jkwatson: Well, it took my brain a couple of careful looks at it to understand what you were doing. So, having a clearly named method that explains that would help reduce the cognitive overhead of reading the code. :)

carlosalberto: Fair enough, lets do that.

carlosalberto:
Hey all,

I've updated the code with tests and removing a few things. It's now called `StackTracePropagator`, as it's intended to be used explicitly with trace propagators (had initially intended to use `Text` in the same, to signal it works with `HttpTextFormat` instances, but I doubt we will have to support multiple *binary* formats).

It can also work without the Specification changes mentioned in open-telemetry/opentelemetry-specification#496 but I have left them here for now (will update the code if those changes are rejected).

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/StackTracePropagator.java
```diff
@@ -0,0 +1,184 @@
...
+ * // Extraction will be performed in reverse  order, i.e. starting with the last
+ * // registered propagator (HttpTraceContext in this example).
+ * Context context = OpenTelemetry.getPropagators().getHttpTextFormat()
+ *   .extract(context, carrier, carrierGetter);
+ * }</pre>
+ *
+ * @since 0.6.0
+ */
+@Immutable
+public class StackTracePropagator implements HttpTextFormat {
```
jkwatson, line 60: I think calling this stack trace will confuse many people. How about `StackedTracePropagator`, or `TraceMultiPropagator` or `CompoundPropagator`?

api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java
```diff
@@ -125,6 +125,10 @@
     checkNotNull(getter, "getter");
 
     SpanContext spanContext = extractImpl(carrier, getter);
+    if (!spanContext.isValid()) {
```
anuraaga, line 128: Can we change it so propagator work on `SpanContext`. Is there a reason we use `Context` and copy all this logic into every propagator?

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/StackTracePropagator.java
anuraaga: Yup :P `Composite` is pretty common I think too. Either way, I think the options open up if we don't go in reverse order, which from what I can tell doesn't seem to be that important.

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/StackTracePropagator.java
```diff
@@ -0,0 +1,184 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * A propagator designed to inject and extract multiple trace {@code HttpTextFormat} propagators,
+ * intendended for backwards compatibility with existing services using different formats. It works
+ * in a stack-fashion, starting with the last registered propagator, to the first one.
```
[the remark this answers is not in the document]
jkwatson: Agree with this. I don't see any reason to go in a stack-like order; it'll be much clearer is just to go in order, like @anuraaga suggests.

api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java
jkwatson: I think this arose from the desire to keep context propagation as a separate concern from tracing. I think, looking back on it now, that this might have been a goal that's more academic than useful for end users.

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/StackTracePropagator.java
anuraaga: Ah also remembered we have multispanprocessor. So `MultiPropagator` would be consistent for this codebase.

carlosalberto: `CompoundPropagator` is too generic. `StackedTracePropagator` sounds great to me if we keep the stack-alike behavior, but `TraceMultiPropagator` would work if we go the simple-linear behavior.

`MultiPropagator` is not a good name IMHO as we don't *really* work on all propagators for extraction (we stop as soon as one properly extracts a `SpanContext`).

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/StackTracePropagator.java
```diff
@@ -0,0 +1,184 @@
...
+import io.opentelemetry.trace.TracingContextUtils;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * A propagator designed to inject and extract multiple trace {@code HttpTextFormat} propagators,
+ * intendended for backwards compatibility with existing services using different formats. It works
```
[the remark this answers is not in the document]
carlosalberto: It's the rationale on why this can propagator can be used. I'd rather leave it in place.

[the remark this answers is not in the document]
carlosalberto: Oh never mind, I thought you wanted to remove the `compatibility` word ;)

api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java
carlosalberto: > I think this arose from the desire to keep context propagation as a separate concern from tracing.

This is the case.

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/StackTracePropagator.java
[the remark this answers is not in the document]
carlosalberto: This is based on prior art (also, see #1273 ) but I'm free to change that if we all agree that we want simple linear usage.

cc @pavolloffay

api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java
anuraaga: Doesn't have to be this PR but recommend revisit this. The propagators are in a package called `trace` so decoupling from tracing doesn't seem to make much sense, or otherwise the package is wrong.

carlosalberto: Please open an issue if you think this is important and the reasons. Personally I think there's no problem, but I might be losing some angle ;) (also, see the linked issue).

anuraaga: Thanks will discuss this separately, you're right it doesn't need to be taken care of here.

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/StackTracePropagator.java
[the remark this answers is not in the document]
anuraaga: +1 to simple linear order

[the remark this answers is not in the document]
anuraaga: One observation is this is about how to set propagation order for a Java developer. It seems like this doesn't really need a spec but could be left up to the language? FIFO, LIFO, it may be idiomatic one way or another in different languages. In Java, I think going in-order is idiomatic, we have some famous examples like servlet filter chain. But I can also imagine a function language may use something like b3(w3c) which could be an idiomatic way of expressing the w3c, then b3, order in that language. Environment variables, being a non-programming concept, do make more sense for a spec though.

Just a thought.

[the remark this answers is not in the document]
carlosalberto: Actually I'd like this to be in the spec, but it might take a little bit of time before we can have it there (*if* we actually manage to have it there).

I'm not familiar with servlet filter, but is it something that traverses *all* components, or can it stop right away upon a successful operation? If that's latter, that can help us go forward with this approach (always, we can go linear and change it later on).

For the sake of this PR, I will change to name to `TraceMultiPropagator` now, while we wait for feedback.

jkwatson:
@carlosalberto looks like the build is failing on this. Happy to get it approved and merged with the requested naming changes.

carlosalberto:
@jkwatson Done - however, the order is kept, i.e. the last one wins. Folks at other SIGs informally told me they preferred the last-one-wins approach (the current one here), but @pavolloffay seemed to (based on his PR) prefer to have the first `Propagator` win (also aligned with what you guys think).

If you think we should go with the first-one-wins @jkwatson let me know and I will update the code (guess we could change it later on if/as needed).

jkwatson:
I don't have a strong opinion on first-one-wins vs last-one-wins, so this is fine with me.

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagator.java
```diff
@@ -110,6 +110,9 @@
     Objects.requireNonNull(getter, "getter");
 
     SpanContext spanContext = getSpanContextFromHeader(carrier, getter);
+    if (!spanContext.isValid()) {
```
jkwatson, line 113: can you add unit test cases for this new behavior, for each of the propagators that's changed?

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/TraceMultiPropagator.java
```diff
@@ -0,0 +1,184 @@
...
+@Immutable
+public class TraceMultiPropagator implements HttpTextFormat {
+  private final HttpTextFormat[] propagators;
+  private final List<String> propagatorsFields;
+
+  private TraceMultiPropagator(List<HttpTextFormat> propagatorList) {
+    this.propagators = new HttpTextFormat[propagatorList.size()];
+    propagatorList.toArray(this.propagators);
+
+    List<String> fields = new ArrayList<>();
```
anuraaga, line 68: Should we validate if fields are duped? Or does it not matter? Mainly wondering whether it's ok for this list to have dupes.

carlosalberto: Still not clear how to proceed here (we have a similar case for the base `CompositePropagator` in context-prop). I will fill an issue for that before merging this PR.

(My gut feeling is that we should de-dup them at ctor time).

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagator.java
carlosalberto: Oh my, didn't realize I haven't written those ones. Will do super early tomorrow.

carlosalberto:
@jkwatson Tests added ;)

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/TraceMultiPropagator.java
anuraaga: If you can add a TODO to the issue here would be nice

carlosalberto: An issue is even better ;) https://github.com/open-telemetry/opentelemetry-java/issues/1388

anuraaga: Meant it'd be good to link to that issue here with a code TODO - reduces chance of the issue just sitting in the backlog if it distracts in the code ;)

PR #1340, opened 2020-06-17
Add a ReadableKeyValuePairs and make ReadableAttribute a simple sub-interface

bogdandrutu:
Main advantage is avoid duplicating readable methods, and enforce better hierarchy. Also the consumer interface can now be part of the ReadableKeyValuePairs.


jkwatson:
I do think having a qualified sub-interface for ReadableAttributes still has value, and makes the exporter code a bit easier to read. But, I also agree that having the top interface is a good idea. 

So, yes, and let's re-introduce ReadableAttributes as a sub-interface of ReadableKeyValuePairs. Seems like a smaller change, as well.

bogdandrutu:
>Looks good. ReadableKeyValuePairs.KeyValueConsumer is a bit of a mouthful. Would it make sense to just have the consumer a top-level interface? Having them coupled together is nice, but the style-checker seems to get annoyed when you try to import the sub-interface explicitly.

It is just IDEA that decided to refactor this way when I asked to move that class.

PR #1342, opened 2020-06-17
Extension packages should not use "internal" API package

### LINKED ISSUE #1125: Contrib packages should not use "internal" API package.
Internal package should not be used outside of API (maybe SDK). The reason is called internal is because we don't have the guarantees that we have in the API for the API stability.

iNikem:
Fixes #1125, second attempt.

This PR adds InternalApiProtectionTest which uses ArchUnit to enforce architectural constraints. It only verifies that classes in contrib packages do not use classes in internal package of API module.

Test was added into opentelemetry-all module because the latter depends on all other submodules and therefore sees all classes in the project.

All found violations were fixed.

StringUtils was copied from Guava, so its original copyright notice was preserved with slight formatting changes. Even then I had to change our checkstyle configuration to make it pass.

### DIFF
--- a/extensions/trace_propagators/src/test/java/io/opentelemetry/extensions/trace/propagation/StringUtilsTest.java
+++ b/extensions/trace_propagators/src/test/java/io/opentelemetry/extensions/trace/propagation/StringUtilsTest.java
@@ -17,7 +17,6 @@
 package io.opentelemetry.extensions.trace.propagation;
 
 import static com.google.common.truth.Truth.assertThat;
-import static org.junit.Assert.assertEquals;
 
 import org.junit.Rule;
 import org.junit.Test;
@@ -36,7 +35,7 @@ public class StringUtilsTest {
 
   @Test
   public void padLeft() {
-    assertEquals(StringUtils.padLeft("value", 10), "00000value");
+    assertThat(StringUtils.padLeft("value", 10)).isEqualTo("00000value");
   }
 
   @Test
--- a/extensions/trace_propagators/src/test/java/io/opentelemetry/extensions/trace/propagation/StringUtilsTest.java
+++ b/extensions/trace_propagators/src/test/java/io/opentelemetry/extensions/trace/propagation/StringUtilsTest.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2020, OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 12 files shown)


extensions/trace_propagators/src/test/java/io/opentelemetry/extensions/trace/propagation/StringUtilsTest.java
```diff
@@ -0,0 +1,55 @@
...
+  public void padLeft() {
+    assertEquals(StringUtils.padLeft("value", 10), "00000value");
+  }
+
+  @Test
+  public void padLeft_throws_for_null_value() {
+    try {
+      StringUtils.padLeft(null, 10);
+      fail("Expected exception not");
+    } catch (NullPointerException expected) {
```
jkwatson, line 45: we should use the junit ExpectedException for this, rather than try/fail

bogdandrutu: or simple:

```
    thrown.expect(NullPointerException.class);
    thrown.expectMessage("name");
    StringUtils.padLeft(null, 10);
```

bogdandrutu: @jkwatson there is also an errorprone check for this which we can enable now with tests being java8 about this exact pattern.

jkwatson: yeah, @iNikem disabled it in this PR. that's why it caught my eye.  :)

iNikem: Should we prefer `ExpectedException` or, as we have java8 in tests now, `assertThrows`?

bogdandrutu: I think ExpectedException does a better job for this.

extensions/trace_propagators/src/test/java/io/opentelemetry/extensions/trace/propagation/StringUtilsTest.java
```diff
@@ -0,0 +1,55 @@
...
+  public void isNullOrEmpty() {
+    assertThat(StringUtils.isNullOrEmpty("")).isTrue();
+    assertThat(StringUtils.isNullOrEmpty(null)).isTrue();
+    assertThat(StringUtils.isNullOrEmpty("hello")).isFalse();
+    assertThat(StringUtils.isNullOrEmpty(" ")).isFalse();
+  }
+
+  @Test
+  public void padLeft() {
+    assertEquals(StringUtils.padLeft("value", 10), "00000value");
```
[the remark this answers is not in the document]
bogdandrutu: This was just ignored and marked as resolved, please don't do this in the future, if you disagree say it and don't resolve the comment.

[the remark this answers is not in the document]
iNikem: Oh sorry, I fixed that in another test case below, that's why I marked as resolved. I have totally missed this one.

PR #1344, opened 2020-06-17
Log4j2 extensions

zenmoto:
A small library that allows for adding request correlation information to Log4j 2.x. The two extensions are a context data provider that makes the correlation information available to layout components, and a layout that attempts to follow the [Log Data Model](https://github.com/open-telemetry/oteps/blob/master/text/logs/0097-log-data-model.md), though the JSON representation of that model probably needs to be more strictly defined.


bogdandrutu:
@zenmoto fixing the main build in https://github.com/open-telemetry/opentelemetry-java/pull/1348

api/src/test/java/io/opentelemetry/trace/propagation/HttpTraceContextTest.java
```diff
@@ -58,7 +58,7 @@
   private static final String TRACEPARENT_HEADER_NOT_SAMPLED =
       "00-" + TRACE_ID_BASE16 + "-" + SPAN_ID_BASE16 + "-00";
   private static final Setter<Map<String, String>> setter = Map::put;
-  private static final Getter<Map<String, String>> getter = (carrier, key) -> carrier.get(key);
+  private static final Getter<Map<String, String>> getter = Map::get;
```
bogdandrutu, line 61: This is fixed in master

bogdandrutu:
Please rebase.

extensions/logging/log4j2_extensions/src/test/resources/ContextDataProviderTestConfig.xml
```diff
@@ -0,0 +1,45 @@
...
+      <AppenderRef ref="SyncList"/>
+    </Root>
+    <AsyncLogger name="AsyncContextDataProviderTest" level="debug">
+      <AppenderRef ref="AsyncList"/>
+    </AsyncLogger>
+    <Logger name="JsonContextDataProviderTest" level="debug">
+      <AppenderRef ref="JsonList"/>
+    </Logger>
+  </Loggers>
+</Configuration>
```
bogdandrutu, line 45: new line here?

extensions/logging/log4j2_extensions/src/main/java/io/opentelemetry/contrib/logging/log4j2/TraceContextDataProvider.java
```diff
@@ -0,0 +1,57 @@
...
+   * org.apache.logging.log4j.core.appender.AsyncAppender}
+   *
+   * @return A map containing string versions of the traceid, spanid, and traceflags which can then
+   *     be accessed from layout components
+   * @see OpenTelemetryJsonLayout
+   */
+  @Override
+  public Map<String, String> supplyContextData() {
+    Tracer tracer = OpenTelemetry.getTracerProvider().get("ot_prototype_logging");
+    Span span = tracer.getCurrentSpan();
```
bogdandrutu, line 45: I think you can use `TracingContextUtils.getCurrentSpan()`

Otherwise please make the tracer static member.

extensions/logging/log4j2_extensions/src/main/java/io/opentelemetry/contrib/logging/log4j2/TraceContextDataProvider.java
```diff
@@ -0,0 +1,57 @@
...
+    Map<String, String> map = new HashMap<>();
+    if (span != null && span.isRecording()) {
```
bogdandrutu, line 47: the "isRecording" should be a property of the TraceContextDataProvider to record for non recording spans as well.

extensions/logging/log4j2_extensions/src/main/java/io/opentelemetry/contrib/logging/log4j2/OpenTelemetryJsonLayout.java
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * This class implements a JSON layout for Log4j 2.x that will include request correlation
+ * information in the form of traceid, spanid, and traceflags fields. The format is provisional, but
+ * is designed to mirror the <a
+ * href="https://github.com/open-telemetry/oteps/blob/master/text/logs/0097-log-data-model.md">Log
+ * Data Model</a>
+ */
+@Plugin(name = "OpenTelemetryJsonLayout", category = "Core", elementType = "layout")
+public class OpenTelemetryJsonLayout extends AbstractStringLayout {
```
bogdandrutu, line 42: Can this PR be split between OpenTelemetryJsonLayout and TraceContextDataProvider?

extensions/logging/log4j2_extensions/src/main/java/io/opentelemetry/contrib/logging/log4j2/OpenTelemetryJsonLayout.java
```diff
@@ -0,0 +1,127 @@
...
+
+      generator.writeFieldName("severitytext");
+      generator.writeString(event.getLevel().name());
+
+      generator.writeFieldName("severitynumber");
+      generator.writeNumber(decodeSeverity(event.getLevel()));
+
+      if (event.getContextData().containsKey("traceid")) {
+        writeRequestCorrelation(generator, event.getContextData());
+      }
```
bogdandrutu, line 71: Probably you should cache the value of `event.getContextData()`?

extensions/logging/log4j2_extensions/src/main/java/io/opentelemetry/contrib/logging/log4j2/TraceContextDataProvider.java
```diff
@@ -0,0 +1,57 @@
...
+import io.opentelemetry.trace.Tracer;
+import java.util.HashMap;
+import java.util.Map;
+import org.apache.logging.log4j.core.util.ContextDataProvider;
+
+/**
+ * This ContextDataProvider is loaded via the ServiceProvider facility. {@link #supplyContextData()}
+ * is called when a log entry is created.
+ */
+public class TraceContextDataProvider implements ContextDataProvider {
```
bogdandrutu, line 31: Please investigate if possible to not call `toLowerBase16` on the critical path. Save the current SpanContext in the map and format the strings when exporting:
1. Remove unnecessary formatting and allocations from the critical path
2. Remove unnecessary formatting and allocations if log is dropped in the export pipeline.

extensions/logging/log4j2_extensions/src/main/java/io/opentelemetry/contrib/logging/log4j2/OpenTelemetryJsonLayout.java
zenmoto: I can. The OpenTelemetryJsonLayout depends on the TraceContextDataProvider. Would it be better to submit and merge the context data provider and then submit the other, or would it be better to submit the layout against the data provider branch?

extensions/logging/log4j2_extensions/src/main/java/io/opentelemetry/contrib/logging/log4j2/TraceContextDataProvider.java
zenmoto: Unfortunately one of the main use cases here is use by the PatternLayout class, which won't know to call `toLowerbase16` when used, so this needs to be a string. On the positive side, this code is not called if the logging call is below the current threshold- I just confirmed that if a message is logged at a `trace` level, the message is discarded before the `TraceContextDataProvider` is invoked.

extensions/logging/log4j2_extensions/src/main/java/io/opentelemetry/contrib/logging/log4j2/TraceContextDataProvider.java
zenmoto: Sorry, I'm not following. Do you mean that I should not check on `isRecording`, or that I should check on `TraceContextDataProvider` as well? I'm mostly trying to check to ensure that this is a real span- would it be better to call `span.getContext().isValid()`?

extensions/logging/log4j2_extensions/src/main/java/io/opentelemetry/contrib/logging/log4j2/TraceContextDataProvider.java
```diff
@@ -0,0 +1,57 @@
...
+   * available to a layout, even if the logger is using an {@link
+   * org.apache.logging.log4j.core.appender.AsyncAppender}
+   *
+   * @return A map containing string versions of the traceid, spanid, and traceflags which can then
+   *     be accessed from layout components
+   * @see OpenTelemetryJsonLayout
+   */
+  @Override
+  public Map<String, String> supplyContextData() {
+    Tracer tracer = OpenTelemetry.getTracerProvider().get("ot_prototype_logging");
```
jkwatson, line 44: It's odd to get a handle to a tracer like this without intending to actually do any tracing. If we do end up with a tracer, I agree with @bogdandrutu that it should be a static member of this class. Also, the name of it should match the standard of using the package name where it is used, I think.

extensions/logging/log4j2_extensions/src/test/java/io/opentelemetry/contrib/logging/log4j2/OpenTelemetryJsonLayoutTest.java
```diff
@@ -0,0 +1,74 @@
...
+    Tracer tracer = OpenTelemetry.getTracer("JsonLayoutTest");
+    Span span = tracer.spanBuilder("a_span").startSpan();
+    try (Scope scope = tracer.withSpan(span)) {
+      logger.error("test 2");
+    }
+
+    List<String> messages = appender.getMessages();
+    String first = messages.get(0);
+    Map<?, ?> data = gson.fromJson(first, Map.class);
+    assertEquals("test", data.get("body"));
```
jkwatson, line 59: We prefer using the truth library for assertions, rather than using assertEquals.

extensions/logging/log4j2_extensions/src/main/java/io/opentelemetry/contrib/logging/log4j2/TraceContextDataProvider.java
jkwatson: Yes, `isRecording` refers to if the span is recording events and having events added to it. I think you want to check for a valid span context, as you suggest.

PR #1346, opened 2020-06-17
fix compiler error about lambdas

jkwatson:


api/src/test/java/io/opentelemetry/trace/propagation/HttpTraceContextTest.java
```diff
@@ -58,7 +58,7 @@
   private static final String TRACEPARENT_HEADER_NOT_SAMPLED =
       "00-" + TRACE_ID_BASE16 + "-" + SPAN_ID_BASE16 + "-00";
   private static final Setter<Map<String, String>> setter = Map::put;
-  private static final Getter<Map<String, String>> getter = (carrier, key) -> carrier.get(key);
+  private static final Getter<Map<String, String>> getter = Map::get;
```
bogdandrutu, line 61: You are going to kill me with lambdas :(((

PR #1347, opened 2020-06-17
Change samplers to use the new Attributes class

bogdandrutu:


carlosalberto:
Curious as to why `SamplingDecision` is removed - had we forgotten to do it before?

jkwatson:
> Curious as to why `SamplingDecision` is removed - had we forgotten to do it before?

the changes in the sampling package now make it easier to return these without having the extra class. The class was being used in the jaeger RateLimitingSampler class

PR #1352, opened 2020-06-18
"Start using real java8 in tests."

bogdandrutu:
Depends on https://github.com/open-telemetry/opentelemetry-java/pull/1351


carlosalberto:
Quite a big set of changes, but looks good overall.

PR #1365, opened 2020-06-21
support aws plugins EC2/ECS/Beanstalk

wangzlei:
**Description:** Add AWS EC2/ECS/Beanstack resources support

Can get AWS service resources in Span.


sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/utils/DockerUtil.java
```diff
@@ -0,0 +1,63 @@
...
+
+  /**
+   * Get docker container id from local cgroup file.
+   *
+   * @return docker container ID
+   */
+  @Nullable
+  @SuppressWarnings("DefaultCharset")
+  public String getContainerId() {
+    try (BufferedReader br = new BufferedReader(new FileReader(DEFAULT_CGROUP_PATH))) {
```
anuraaga, line 48: Since this is one-time initialization logic, you may just stick with `Files.readAllLines`

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/utils/DockerUtil.java
```diff
@@ -0,0 +1,63 @@
...
+      String line;
+      while ((line = br.readLine()) != null) {
+        if (line.length() > CONTAINER_ID_LENGTH) {
```
anuraaga, line 51: Can you find some documentation to support this logic? I guess it probably works in practice, but seems fragile to me. It would be less fragile if we at least checked for `/docker/`. Here's an example with my current docker for mac

```
-> % docker run -it --rm bash
Unable to find image 'bash:latest' locally
latest: Pulling from library/bash
df20fa9351a1: Pull complete
f5b2321f64f1: Pull complete
a82fbae36939: Pull complete
Digest: sha256:376bd611ec5dfc1248a2d21d921bd62d0b858344ce1516cd133249c0a814387d
Status: Downloaded newer image for bash:latest
bash-5.0#
bash-5.0#
bash-5.0# cat /proc/self/cgroup
14:name=systemd:/docker/795c2b3af01d8186ff10735a1cc8aec33856b85820a50cbd29539e69d3236fe4
13:rdma:/
12:pids:/docker/795c2b3af01d8186ff10735a1cc8aec33856b85820a50cbd29539e69d3236fe4
11:hugetlb:/docker/795c2b3af01d8186ff10735a1cc8aec33856b85820a50cbd29539e69d3236fe4
10:net_prio:/docker/795c2b3af01d8186ff10735a1cc8aec33856b85820a50cbd29539e69d3236fe4
9:perf_event:/docker/795c2b3af01d8186ff10735a1cc8aec33856b85820a50cbd29539e69d3236fe4
8:net_cls:/docker/795c2b3af01d8186ff10735a1cc8aec33856b85820a50cbd29539e69d3236fe4
7:freezer:/docker/795c2b3af01d8186ff10735a1cc8aec33856b85820a50cbd29539e69d3236fe4
6:devices:/docker/795c2b3af01d8186ff10735a1cc8aec33856b85820a50cbd29539e69d3236fe4
5:memory:/docker/795c2b3af01d8186ff10735a1cc8aec33856b85820a50cbd29539e69d3236fe4
4:blkio:/docker/795c2b3af01d8186ff10735a1cc8aec33856b85820a50cbd29539e69d3236fe4
3:cpuacct:/docker/795c2b3af01d8186ff10735a1cc8aec33856b85820a50cbd29539e69d3236fe4
2:cpu:/docker/795c2b3af01d8186ff10735a1cc8aec33856b85820a50cbd29539e69d3236fe4
1:cpuset:/docker/795c2b3af01d8186ff10735a1cc8aec33856b85820a50cbd29539e69d3236fe4
bash-5.0# echo $HOSTNAME
795c2b3af01d
```

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/BeanstalkResource.java
```diff
@@ -0,0 +1,84 @@
...
+import com.google.common.collect.ImmutableMap;
+import io.opentelemetry.common.AttributeValue;
+import io.opentelemetry.sdk.resources.ResourceConstants;
+import java.io.File;
+import java.io.IOException;
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class BeanstalkResource extends AwsResource {
```
anuraaga, line 33:
```suggestion
class BeanstalkResource extends AwsResource {
```

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EcsResource.java
```diff
@@ -0,0 +1,68 @@
...
+import io.opentelemetry.common.AttributeValue;
+import io.opentelemetry.sdk.extensions.trace.aws.resource.utils.DockerUtil;
+import io.opentelemetry.sdk.resources.ResourceConstants;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class EcsResource extends AwsResource {
```
anuraaga, line 31:
```suggestion
class EcsResource extends AwsResource {
```

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EcsResource.java
```diff
@@ -0,0 +1,68 @@
...
+
+  private static final Logger logger = Logger.getLogger(EcsResource.class.getName());
+
+  private static final String ECS_METADATA_KEY_V4 = "ECS_CONTAINER_METADATA_URI_V4";
+  private static final String ECS_METADATA_KEY_V3 = "ECS_CONTAINER_METADATA_URI";
+  private static final String CONTAINER_ID = "container.id";
+
+  private final DockerUtil dockerUtil;
+
+  public EcsResource(DockerUtil dockerUtil) {
```
anuraaga, line 41: Instead of having `AwsResource` inject this, you could consider a standard pattern of always defining a no-args and only-for-test args constructor

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EcsResource.java
```diff
@@ -0,0 +1,68 @@
...
+    this.dockerUtil = dockerUtil;
+  }
+
+  @Override
+  Map<String, AttributeValue> createAttributes() {
+    String uriV3 = System.getenv(ECS_METADATA_KEY_V3);
+    String uriV4 = System.getenv(ECS_METADATA_KEY_V4);
+    if ((uriV3 == null || uriV3.isEmpty()) && (uriV4 == null || uriV4.isEmpty())) {
```
anuraaga, line 49:
```suggestion
    if (Strings.isNullOrEmpty(uriV3) || Strings.isNullOrEmpty(uriV4)) {
```

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EcsResource.java
```diff
@@ -0,0 +1,68 @@
...
+      return ImmutableMap.of();
+    }
+    ImmutableMap.Builder<String, AttributeValue> resourceAttributes = ImmutableMap.builder();
```
anuraaga, line 52:
```suggestion

    ImmutableMap.Builder<String, AttributeValue> resourceAttributes = ImmutableMap.builder();
```

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EcsResource.java
```diff
@@ -0,0 +1,68 @@
...
+
+    try {
+      String hostName = InetAddress.getLocalHost().getHostName();
```
anuraaga, line 55: While I'm not so familiar with them, there are apparently gotchas with just sticking with JDK, consider a more robust one like this

https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/common/util/SystemInfo.java#L205

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EcsResource.java
```diff
@@ -0,0 +1,68 @@
...
+
+  private final DockerUtil dockerUtil;
+
+  public EcsResource(DockerUtil dockerUtil) {
+    this.dockerUtil = dockerUtil;
+  }
+
+  @Override
+  Map<String, AttributeValue> createAttributes() {
+    String uriV3 = System.getenv(ECS_METADATA_KEY_V3);
```
anuraaga, line 47: `// Check whether we are actually on ECS`

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EcsResource.java
```diff
@@ -0,0 +1,68 @@
...
+    ImmutableMap.Builder<String, AttributeValue> resourceAttributes = ImmutableMap.builder();
+
+    try {
+      String hostName = InetAddress.getLocalHost().getHostName();
+      resourceAttributes.put(ResourceConstants.CONTAINER_NAME, stringAttributeValue(hostName));
+    } catch (UnknownHostException e) {
+      logger.log(Level.WARNING, "Could not get docker container name from hostname.", e);
+    }
+
+    String containerId = dockerUtil.getContainerId();
```
anuraaga, line 61: Looks like there's no ECS-specific logic here other than turning on / off the resource, which we don't really need in practice. Probably for a followup we should look at moving this into the SDK itself.

sdk_extensions/aws_v1_support/src/test/java/io/opentelemetry/sdk/extensions/trace/aws/resource/BeanstalkResourceTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest(BeanstalkResource.class)
+public class BeanstalkResourceTest {
+
+  private BeanstalkResource populator;
+
+  @Before
+  public void setUp() throws Exception {
+    File file = new File("tempFile");
```
anuraaga, line 43: I think this creates a file in the working directory? Use a temporary folder https://junit.org/junit4/javadoc/4.12/org/junit/rules/TemporaryFolder.html

sdk_extensions/aws_v1_support/src/test/java/io/opentelemetry/sdk/extensions/trace/aws/resource/BeanstalkResourceTest.java
```diff
@@ -0,0 +1,69 @@
...
+  public void release() {
+    new File("tempFile").delete();
+  }
+
+  @Test
+  public void testCreateAttributes() {
+
+    Map<String, AttributeValue> metadata = populator.createAttributes();
+
+    assertEquals(3, metadata.size());
```
anuraaga, line 63: Use Truth (similar to assertj) e.g.,

```
assertThat(metadata...).isEqualTo("4")
```

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/utils/DockerUtil.java
```diff
@@ -0,0 +1,63 @@
...
+import java.io.IOException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.annotation.Nullable;
+
+public class DockerUtil {
+
+  private static final Logger logger = Logger.getLogger(DockerUtil.class.getName());
+  private static final int CONTAINER_ID_LENGTH = 64;
+  private static final String DEFAULT_CGROUP_PATH = "/proc/self/cgroup";
```
anuraaga, line 31: Ditto allow passing this path as a parameter so in the test you can use a different file.

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EcsResource.java
```diff
@@ -0,0 +1,68 @@
...
+  private static final String CONTAINER_ID = "container.id";
+
+  private final DockerUtil dockerUtil;
+
+  public EcsResource(DockerUtil dockerUtil) {
+    this.dockerUtil = dockerUtil;
+  }
+
+  @Override
+  Map<String, AttributeValue> createAttributes() {
```
anuraaga, line 46: Inject `Map<String, String>` into constructor instead of calling `System` directly here, then you can avoid powermock.

sdk_extensions/aws_v1_support/src/test/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EcsResourceTest.java
```diff
@@ -0,0 +1,61 @@
...
+    EcsResource populator = new EcsResource(DockerUtil.getInstance());
+    assertEquals(0, populator.createAttributes().size());
+  }
+
+  @Test
+  public void testCreateAttributes() throws UnknownHostException {
+    mockStatic(System.class);
+    when(System.getenv("ECS_CONTAINER_METADATA_URI_V4")).thenReturn("ecs_metadata_uri");
+
+    DockerUtil mockDockerUtil = mock(DockerUtil.class);
```
anuraaga, line 50: Let's use mockito's junit support

```

@Rule
public MockitoRule mocks = MockitoJunit.rule();

@Mock
private DockerUtil dockerUtil;
```

sdk_extensions/aws_v1_support/src/test/java/io/opentelemetry/sdk/extensions/trace/aws/resource/utils/DockerUtilTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest(DockerUtil.class)
+public class DockerUtilTest {
+
+  @Test
+  public void testGetContainerId() throws Exception {
+    assertNull(DockerUtil.getInstance().getContainerId());
+
+    FileReader mockFileReader = mock(FileReader.class);
```
anuraaga, line 41: Going to get so much nicer with an injected path! :)

sdk_extensions/aws_v1_support/src/test/java/io/opentelemetry/sdk/extensions/trace/aws/resource/utils/DockerUtilTest.java
```diff
@@ -0,0 +1,57 @@
...
+    whenNew(FileReader.class).withAnyArguments().thenReturn(mockFileReader);
+    BufferedReader mockBufferReader = mock(BufferedReader.class);
+    when(mockBufferReader.readLine())
+        .thenReturn("dummy")
+        .thenReturn(
+            "11:devices:/ecs/bbc36dd0-5ee0-4007-ba96-c590e0b278d2/"
```
anuraaga, line 47: Test various cases, such as missing file, empty file, file with contents without the ID, etc.

sdk_extensions/aws_v1_support/src/test/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EcsResourceTest.java
```diff
@@ -0,0 +1,61 @@
...
+    when(mockDockerUtil.getContainerId()).thenReturn("0123456789");
```
anuraaga, line 51: Test when no container ID found

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/BeanstalkResource.java
```diff
@@ -0,0 +1,84 @@
...
+  private static final Logger logger = Logger.getLogger(BeanstalkResource.class.getName());
+
+  private static final String DEVELOPMENT_ID = "deployment_id";
+  private static final String VERSION_LABEL = "version_label";
+  private static final String ENVIRONMENT_NAME = "environment_name";
+  private static final String BEANSTALK_CONF_PATH = "/var/elasticbeanstalk/xray/environment.conf";
+  private static final JsonFactory JSON_FACTORY = new JsonFactory();
+
+  @Override
+  Map<String, AttributeValue> createAttributes() {
```
anuraaga, line 44: Might be a good time to migrate from this `Map` to `Attributes`

sdk_extensions/aws_v1_support/build.gradle
```diff
@@ -17,6 +17,8 @@ dependencies {
             platform(boms.guava)
 
     testImplementation 'com.github.tomakehurst:wiremock-jre8:2.26.3'
+    testImplementation 'org.powermock:powermock-module-junit4:2.0.7'
```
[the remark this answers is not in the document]
jkwatson: yes, we should avoid having to use powermock, I definitely agree on this.

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/utils/DockerUtil.java
jkwatson: readAllLines is java 8+, so can't use that in this module, as-is

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/utils/DockerUtil.java
```diff
@@ -0,0 +1,63 @@
...
+          return line.substring(line.length() - CONTAINER_ID_LENGTH);
+        }
+      }
+    } catch (FileNotFoundException e) {
+      logger.log(Level.WARNING, "Failed to read container id, cgroup file does not exist.");
+    } catch (IOException e) {
+      logger.log(Level.WARNING, "Unable to read container id: " + e.getMessage());
+    }
+
+    return null;
```
jkwatson, line 61: If we have to use `null` as a return value here, you should document what conditions will result in null being returned.

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/BeanstalkResource.java
```diff
@@ -0,0 +1,84 @@
...
+      return ImmutableMap.of();
+    }
+
+    ImmutableMap.Builder<String, AttributeValue> resourceAttributes = ImmutableMap.builder();
+
+    try (JsonParser parser = JSON_FACTORY.createParser(configFile)) {
+      parser.nextToken();
+
+      if (!parser.isExpectedStartObjectToken()) {
+        throw new IOException("Invalid Beanstalk config:" + BEANSTALK_CONF_PATH);
```
jkwatson, line 56: don't use exceptions for flow control. Just return the empty attributes here if you need to bail out.

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EcsResource.java
jkwatson: I disagree. Having your dependencies be explicitly defined and required to be set makes the code clearer. I'd rather not hide the fact that this depends on the docker helper class.

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EcsResource.java
jkwatson: You mean the one returned from `System.getenv()`, correct? I agree, this would be better than using powermock.

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EcsResource.java
jkwatson: That's an interesting idea... you're suggesting we have a resource populator in the SDK that's docker aware?

sdk_extensions/aws_v1_support/build.gradle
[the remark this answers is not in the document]
wangzlei: Agree that PowerMock is too invasive, will refactor code.

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/BeanstalkResource.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.logging.Logger;
+
+public class BeanstalkResource extends AwsResource {
+
+  private static final Logger logger = Logger.getLogger(BeanstalkResource.class.getName());
+
+  private static final String DEVELOPMENT_ID = "deployment_id";
+  private static final String VERSION_LABEL = "version_label";
+  private static final String ENVIRONMENT_NAME = "environment_name";
+  private static final String BEANSTALK_CONF_PATH = "/var/elasticbeanstalk/xray/environment.conf";
```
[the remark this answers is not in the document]
wangzlei: ok

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/BeanstalkResource.java
```diff
@@ -0,0 +1,84 @@
...
+
+      if (!parser.isExpectedStartObjectToken()) {
+        throw new IOException("Invalid Beanstalk config:" + BEANSTALK_CONF_PATH);
+      }
+
+      while (parser.nextToken() != JsonToken.END_OBJECT) {
+        String value = parser.nextTextValue();
+        switch (parser.getCurrentName()) {
+          case DEVELOPMENT_ID:
+            value = Long.toString(parser.getLongValue());
```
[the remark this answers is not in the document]
wangzlei: It is following the format from ElasticBeanstalk config.
{
    "deployment_id":4,
    "version_label":"2",
}

_parser.nextTextValue()_ will get null for "deployment_id".

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/utils/DockerUtil.java
wangzlei: Just found some discussions like
https://forums.docker.com/t/get-a-containers-full-id-from-inside-of-itself/37237/4

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EcsResource.java
wangzlei: Still go back to code testability. If we think PowerMock is not acceptable, which means we should avoid using static, singleton. 
Inject docker helper to Resource explicitly might be the right way .

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EcsResource.java
wangzlei: good catch, I was too depend on Apache StringUtils

sdk_extensions/aws_v1_support/src/test/java/io/opentelemetry/sdk/extensions/trace/aws/resource/BeanstalkResourceTest.java
wangzlei: will use @Rule + TempraryFolder instead of

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/BeanstalkResource.java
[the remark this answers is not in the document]
anuraaga: Ah didn't realize that, Ok :) Does it work to use `parser.getText()` instead of `Long.toString`? Otherwise, if the field is not a long for some reason, it will NPE when unboxing so you should handle that case (and add a test for it, while it shouldn't happen in practice it's worth being defensive).

Another option may be to use `parser.nextValue` and `asString`

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EcsResource.java
anuraaga: Yeah - docker is very common, and even if not in SDK itself, it definitely doesn't seem like it needs to be in this AWS module. Hostname is even more generic and probably could go into the SDK.

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EcsResource.java
```diff
@@ -0,0 +1,68 @@
...
+    String uriV3 = System.getenv(ECS_METADATA_KEY_V3);
+    String uriV4 = System.getenv(ECS_METADATA_KEY_V4);
+    if ((uriV3 == null || uriV3.isEmpty()) && (uriV4 == null || uriV4.isEmpty())) {
+      return ImmutableMap.of();
+    }
+    ImmutableMap.Builder<String, AttributeValue> resourceAttributes = ImmutableMap.builder();
+
+    try {
+      String hostName = InetAddress.getLocalHost().getHostName();
+      resourceAttributes.put(ResourceConstants.CONTAINER_NAME, stringAttributeValue(hostName));
```
anuraaga, line 56: I missed this in first review - should this be `ResourceConstants.HOST_HOSTNAME`?

PR #1366, opened 2020-06-22
Add SDK Telemetry resources

### LINKED ISSUE #1311: Resource sdk.version support

Let me know if I'm missing it but I don't think the SDK fills in information about itself as described here

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/semantic_conventions/README.md#telemetry-sdk

The SDK should fill in the SDK attributes. It'll probably need to propagate the current version into a Java resource file.

One thing that comes to mind is it seems to warrant exposing SDK resource attributes. Users may want to add custom attributes like here

https://github.com/aws-samples/aws-xray-sdk-with-opentelemetry-sample/blob/both-otel-and-xray/awsagentprovider/src/main/java/com/softwareaws/xray/opentelemetry/exporters/AwsTraceProvider.java#L40

And this would need some way to access the SDK resource so it can augment it, rather than replace it completely.

**Additional context**

thisthat:
Resolves #1311

### DIFF
--- a/sdk/build.gradle
+++ b/sdk/build.gradle
@@ -8,7 +8,7 @@ plugins {
 
 description = 'OpenTelemetry SDK'
 ext.moduleName = "io.opentelemetry.sdk"
-ext.propertiesDir = "src/main/resources/io/opentelemetry/sdk"
+ext.propertiesDir = "build/generated/properties/io/opentelemetry/sdk"
 
 dependencies {
     api project(':opentelemetry-api')
@@ -27,7 +27,7 @@ dependencies {
 
 sourceSets {
     main {
-        output.dir(propertiesDir, builtBy: 'generateVersionResource')
+        output.dir("build/generated/properties", builtBy: 'generateVersionResource')
     }
 }
 
--- a/sdk/build.gradle
+++ b/sdk/build.gradle
@@ -30,3 +30,10 @@ animalsniffer {
             sourceSets.main
     ]
 }
+
+processResources {
+    new File(
+            projectDir.getAbsolutePath() + File.separator + ["src", "main", "resources"].join(File.separator),
+            "version.properties"
+    ).text = "version=$version"
+}
--- a/sdk/build.gradle
+++ b/sdk/build.gradle
@@ -8,6 +8,7 @@ plugins {
 
 description = 'OpenTelemetry SDK'
 ext.moduleName = "io.opentelemetry.sdk"
+ext.propertiesDir = "src/main/resources/io/opentelemetry/sdk"
 
 dependencies {
     api project(':opentelemetry-api')
@@ -24,6 +25,12 @@ dependencies {
     signature "net.sf.androidscents.signature:android-api-level-24:7.0_r2@signature"
 }
 
+sourceSets {
+    main {
+        output.dir(propertiesDir, builtBy: 'generateVersionResource')
+    }
+}
+
 animalsniffer {
     // Don't check sourceSets.jmh and sourceSets.test
     sourceSets = [
@@ -31,9 +38,11 @@ animalsniffer {
     ]
 }
 
-processResources {
-    new File(
-            projectDir.getAbsolutePath() + File.separator + ["src", "main", "resources"].join(File.separator),
-            "version.properties"
-    ).text = "version=$version"
+task generateVersionResource {
+    doLast {
+        def folder = file(propertiesDir)
+        folder.mkdirs()
+        def propertiesFile = new File(folder.getAbsolutePath(), "version.properties")
+        propertiesFile.write("sdk.version=${project.version}")
+    }
 }
--- a/sdk/src/main/java/io/opentelemetry/sdk/resources/Resource.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/resources/Resource.java
@@ -24,7 +24,6 @@ import io.opentelemetry.common.ReadableAttributes;
 import io.opentelemetry.common.ReadableKeyValuePairs.KeyValueConsumer;
 import io.opentelemetry.internal.StringUtils;
 import io.opentelemetry.internal.Utils;
-import java.io.FileInputStream;
 import java.util.Objects;
 import java.util.Properties;
 import javax.annotation.Nullable;
@@ -57,15 +56,16 @@ public abstract class Resource {
 
   @Nullable
   private static String readVersion() {
+
     Properties properties = new Properties();
     try {
       properties.load(
-          new FileInputStream(Resource.class.getResource("version.properties").getFile()));
+          Resource.class.getResourceAsStream("/io/opentelemetry/sdk/version.properties"));
     } catch (Exception e) {
       // we left the attribute empty
       return null;
     }
-    return properties.getProperty("version");
+    return properties.getProperty("sdk.version");
   }
 
   Resource() {}
--- a/sdk/src/main/java/io/opentelemetry/sdk/resources/Resource.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/resources/Resource.java
@@ -24,7 +24,9 @@ import io.opentelemetry.common.ReadableAttributes;
 import io.opentelemetry.common.ReadableKeyValuePairs.KeyValueConsumer;
 import io.opentelemetry.internal.StringUtils;
 import io.opentelemetry.internal.Utils;
+import java.io.FileInputStream;
 import java.util.Objects;
+import java.util.Properties;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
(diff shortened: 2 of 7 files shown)


sdk/build.gradle
```diff
@@ -30,3 +30,10 @@ animalsniffer {
             sourceSets.main
     ]
 }
+
+processResources {
+    new File(
```
anuraaga, line 35: I don't think this is doing what you expect, this code is run during the Gradle configuration, not during any task which is not good for several reasons.

Instead, add a new task and set it as a `builtBy` source

https://github.com/aws/aws-xray-sdk-java/blob/master/build.gradle.kts#L63
https://github.com/aws/aws-xray-sdk-java/blob/master/build.gradle.kts#L131

sdk/build.gradle
```diff
@@ -30,3 +30,10 @@ animalsniffer {
...
+            projectDir.getAbsolutePath() + File.separator + ["src", "main", "resources"].join(File.separator),
+            "version.properties"
```
anuraaga, line 37: I think we should put this into a package, e.g. io/opentelemetry/sdk/version.properties

sdk/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -45,6 +47,26 @@
...
+              .setAttribute("telemetry.sdk.language", "java")
+              .setAttribute("telemetry.sdk.version", readVersion())
+              .build());
+
+  @Nullable
+  private static String readVersion() {
+    Properties properties = new Properties();
+    try {
+      properties.load(
+          new FileInputStream(Resource.class.getResource("version.properties").getFile()));
```
anuraaga, line 63: I think `FileInputStream` only works when loading from the file system, e.g., `IntelliJ` but not from a JAR. Instead

`Resource.class.getResourceAsAstream("version.properties")` can handle both

sdk/src/main/java/io/opentelemetry/sdk/trace/TracerSdkProvider.java
```diff
@@ -138,7 +138,7 @@ public void forceFlush() {
 
     private Clock clock = MillisClock.getInstance();
     private IdsGenerator idsGenerator = new RandomIdsGenerator();
-    private Resource resource = EnvVarResource.getResource();
+    private Resource resource = Resource.getTelemetrySdk().merge(EnvVarResource.getResource());
```
anuraaga, line 141: So when a user wants to define their own `Resource`, they always have to merge the default one to get these values. I think this is OK, but it could also be good independently of this PR to think if `Resource` would make sense to have its own SPI, so that any `ResourceProvider` in the app are automatically registered without having to manually merge them all. Of course, there could be ordering problems with this, but it's probably still simpler than manual merging in many cases.

sdk/src/test/java/io/opentelemetry/sdk/resource/ResourceTest.java
```diff
@@ -148,4 +150,15 @@ public void testMergeResources_Resource2_Null() {
     Resource resource = DEFAULT_RESOURCE.merge(resource1).merge(null);
     assertThat(resource.getAttributes()).isEqualTo(expectedAttributes);
   }
+
+  @Test
+  public void testSdkTelemetryResources() {
+    Resource resource = Resource.getTelemetrySdk();
+    ReadableAttributes attributes = resource.getAttributes();
+    String[] keys = {"telemetry.sdk.name", "telemetry.sdk.language", "telemetry.sdk.version"};
```
anuraaga, line 158: Think it's better to avoid the forEach

```java
assertThat(attributes.get("telemetry.sdk.name")).isEqualTo(stringAttributeValue("opentelemetry"));
assertThat(attributes.get("telemetry.sdk.language")).isEqualTo(stringAttributeValue("java"));
assertThat(attributes.get("telemetry.sdk.version")).isNotNull();
```

/cc PS @jkwatson Would love to have `getString` :D

sdk/src/main/java/io/opentelemetry/sdk/trace/TracerSdkProvider.java
carlosalberto: Let's definitely address this case in another PR (we have meant to implement the `Resource` part you mention, in some way or another, for some time now, so hopefully we will have it soon).

sdk/src/test/java/io/opentelemetry/sdk/resource/ResourceTest.java
carlosalberto: I'be initially against adding too many public methods to `Attributes`, as this is in general a class expected to be used by the SDK processing, not by end users. `get()` IMHO is enough for this purpose (and having to write more code in *our* side won't kill us ;) )

jkwatson: This test could be made simpler by just building the Attributes object you expect and asserting that what you get back is equal to it.

sdk/src/main/resources/io/opentelemetry/sdk/version.properties
```diff
@@ -0,0 +1 @@
+sdk.version=0.6.0-SNAPSHOT
```
anuraaga, line 1: Can you delete this file? We want to generate into build at build time, not into the codebase. If you remove this and apply the suggestions and tests still pass, then you should be good

sdk/build.gradle
```diff
@@ -8,6 +8,7 @@ plugins {
 
 description = 'OpenTelemetry SDK'
 ext.moduleName = "io.opentelemetry.sdk"
+ext.propertiesDir = "src/main/resources/io/opentelemetry/sdk"
```
anuraaga, line 11: Don't generate to source, generate to build, e.g. `build/generated/properties/io/opentelemetry/sdk`

sdk/build.gradle
```diff
@@ -24,16 +25,24 @@ dependencies {
     signature "net.sf.androidscents.signature:android-api-level-24:7.0_r2@signature"
 }
 
+sourceSets {
+    main {
+        output.dir(propertiesDir, builtBy: 'generateVersionResource')
```
anuraaga, line 30: Here you would register `build/generated/properties` I believe

thisthat:
Thank you @anuraaga for the feedback! I should have addressed all your points now :)

sdk/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -45,6 +46,27 @@
...
+  @Nullable
+  private static String readVersion() {
+
+    Properties properties = new Properties();
+    try {
+      properties.load(
+          Resource.class.getResourceAsStream("/io/opentelemetry/sdk/version.properties"));
+    } catch (Exception e) {
+      // we left the attribute empty
+      return null;
```
anuraaga, line 66: Just noticed `null` value is allowed - @jkwatson does that seem OK? Wouldn't be surprised if it causes a NPE during export.

Either way, what about setting to `unknown`?

thisthat: My choice is based on the following [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#set-attributes).

> Attribute values of null are considered to be not set and get discarded as if that SetAttribute call had never been made.

However, I noticed that the value is currently not discarded and, instead, a StringAttribute with value `null` is created. I was about to investigate a bit before opening an issue :wink:

carlosalberto: Related: to me, it sounds like the build itself should fail if the version is not properly detected.

anuraaga: The build will make sure there is a version in the jar, but due to shading or similar, it may not make it into a user's app. I've seen a similar setup in Armeria cause NPE for users because of not accounting for the possibility before ><

jkwatson: Yes, I would prefer "unknown" to be here, rather than possible nulls to confound an exporter that wasn't expecting it.

thisthat: Done in 6f113c0d

sdk/build.gradle
```diff
@@ -24,9 +25,24 @@ dependencies {
...
     sourceSets = [
             sourceSets.main
     ]
 }
+
+task generateVersionResource {
+    doLast {
+        def folder = file(propertiesDir)
+        folder.mkdirs()
+        def propertiesFile = new File(folder.getAbsolutePath(), "version.properties")
```
jkwatson, line 45: With such a generic filename, it seems like there's a possibility this could collide with other projects version file. What do you think about making this be something super clear like "opentelemetry-sdk-version.properties" ?

jkwatson: oh, never mind. I see that the folder will be included in here, as well. It's all good!

PR #1368, opened 2020-06-22
Change the usage of MMSC aggregations to always be deltas.

jkwatson:
I started looking at the data coming out of our aggregators for ValueRecorders, and realized that the current approach of having our MMSC aggregations be cumulative probably isn't very useful. In particular, the min & max are only useful when considered over a reporting interval, rather than over the lifetime of the process.

So, I propose in this PR to change to having the ValueRecorders use a delta-aggregation for MMSC, so as to make all 4 values produced be useful.


jkwatson:
@carlosalberto @bogdandrutu Can you take a look? I'd like to get this in to the next release.

PR #1369, opened 2020-06-23
support aws plugins EC2/ECS/Beanstalk

wangzlei:
**Description:** Add AWS EC2/ECS/Beanstack resources support
The previous PR https://github.com/open-telemetry/opentelemetry-java/pull/1365 is canceled.

Can get AWS service resources in Span.


sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EcsResource.java
```diff
@@ -0,0 +1,78 @@
...
+  private static final String ECS_METADATA_KEY_V3 = "ECS_CONTAINER_METADATA_URI";
+  private static final String CONTAINER_ID = "container.id";
+
+  private final DockerHelper dockerHelper;
+  private final String uriV3;
+  private final String uriV4;
+
+  EcsResource() {
+    this(
+        new DockerHelper(), System.getenv(ECS_METADATA_KEY_V3), System.getenv(ECS_METADATA_KEY_V4));
```
anuraaga, line 46: This is ok but I would just pass System.getEnv() map. Gives a bit more test coverage

sdk_extensions/aws_v1_support/src/test/java/io/opentelemetry/sdk/extensions/trace/aws/resource/BeanstalkResourceTest.java
```diff
@@ -0,0 +1,70 @@
...
+package io.opentelemetry.sdk.extensions.trace.aws.resource;
+
+import static com.google.common.truth.Truth.assertThat;
+
+import io.opentelemetry.common.AttributeValue;
+import io.opentelemetry.sdk.resources.ResourceConstants;
+import java.io.File;
+import java.io.IOException;
+import java.util.Map;
+import org.apache.commons.io.FileUtils;
```
anuraaga, line 26: Prefer guava (I think this is leaked into our test classpath but in general we don't use Apache commons)

https://guava.dev/releases/21.0/api/docs/com/google/common/io/Files.html#write-java.lang.CharSequence-java.io.File-java.nio.charset.Charset-

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/DockerHelper.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  /**
+   * Get docker container id from local cgroup file.
+   *
+   * @return docker container ID
+   */
+  @SuppressWarnings("DefaultCharset")
+  public String getContainerId() {
+    try (BufferedReader br = new BufferedReader(new FileReader(cgroupPath))) {
```
anuraaga, line 51: I realized we can use Guava's Files readalllines of we want.

sdk_extensions/aws_v1_support/src/test/java/io/opentelemetry/sdk/extensions/trace/aws/resource/BeanstalkResourceTest.java
```diff
@@ -0,0 +1,70 @@
...
+    assertThat(metadata.get(ResourceConstants.SERVICE_VERSION).getStringValue()).isEqualTo("2");
+    assertThat(metadata.get(ResourceConstants.SERVICE_NAMESPACE).getStringValue())
+        .isEqualTo("HttpSubscriber-env");
+  }
+
+  @Test
+  public void testConfigFileMissing() throws IOException {
+    BeanstalkResource populator = new BeanstalkResource("a_file_never_existing");
+    Map<String, AttributeValue> metadata = populator.createAttributes();
+    assertThat(metadata.size()).isEqualTo(0);
```
anuraaga, line 56: There should be assertThat(metadata).isEmpty

sdk_extensions/aws_v1_support/src/test/java/io/opentelemetry/sdk/extensions/trace/aws/resource/BeanstalkResourceTest.java
```diff
@@ -0,0 +1,70 @@
...
+  public void testCreateAttributes() throws IOException {
+    File file = tempFolder.newFile("beanstalk.config");
+    FileUtils.writeStringToFile(
+        file,
+        "{\"noise\": \"noise\", \"deployment_id\":4,\""
+            + "version_label\":\"2\",\"environment_name\":\"HttpSubscriber-env\"}");
+    BeanstalkResource populator = new BeanstalkResource(file.getPath());
+    Map<String, AttributeValue> metadata = populator.createAttributes();
+
+    assertThat(metadata.size()).isEqualTo(3);
```
anuraaga, line 45: Just do assertThat(metadata).containsExactly for all the map assertions in the PR

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EcsResource.java
```diff
@@ -0,0 +1,78 @@
...
+    try {
+      String hostName = InetAddress.getLocalHost().getHostName();
+      resourceAttributes.put(ResourceConstants.CONTAINER_NAME, stringAttributeValue(hostName));
+    } catch (UnknownHostException e) {
+      logger.log(Level.WARNING, "Could not get docker container name from hostname.", e);
+    }
+
+    String containerId = dockerHelper.getContainerId();
+    if (!Strings.isNullOrEmpty(containerId)) {
+      resourceAttributes.put(CONTAINER_ID, stringAttributeValue(containerId));
```
thisthat, line 73: According to the [specifications](https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/resource/semantic_conventions#container) there is not `container.id` field. Do you have a use case for this field?

anuraaga: Ah nice eyes didn't notice this isn't semanticattributes.

@wangzlei shouldn't this be container name and the above hostname?

jkwatson: We should definitely only be putting official semantic attributes into the Resource. And, the keys should be being kept in the `io.opentelemetry.sdk.resource.ResourceConstants` class.

anuraaga: Yeah I think we'll look at proposing container.id as a semantic convention, it seems to be missing (name is usually human readable not the ID).

But wanted to confirm, you think there should only be official semantic attributes here? What about vendor-specific resource attributes? For example, AWS services have a log group ID used for correlation, which is sort of independent from other semantic values like the service name. I was sort of expecting we'd add this as a non-standard attribute on the resource at some point.

anuraaga: Actually nevermind, I guess this log group I'm thinking of is x-ray specific, but not for general AWS users. So yeah doesn't belong here most likely :)

wangzlei: Just realize that my previous reply was not sent out.
Refer to docker documentation [Container identification](https://docs.docker.com/engine/reference/run/#container-identification), Container's `full-id`, `short-id` and `name` are equivalent in docker CLI like 'docker info <id>'. `name` is usually human readable and can be changed by command ['docker rename'](https://docs.docker.com/engine/reference/commandline/rename/#examples) during runtime. 

I will cut a PR for adding `container.id` in semantic conversions. Thanks.

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EcsResource.java
```diff
@@ -0,0 +1,73 @@
...
+
+    } catch (UnknownHostException e) {
+      logger.log(Level.WARNING, "Could not get docker container name from hostname.", e);
+    }
+
+    return attrBuilders.build();
+  }
+
+  private boolean isOnEcs() {
+    if (sysEnv == null) {
```
anuraaga, line 66: Avoid unnecessary null checks, `sysEnv` can't be null (unless a unit test passes it randomly which is something we don't need to worry about in business logic)

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/DockerHelper.java
```diff
@@ -0,0 +1,66 @@
+/*
```
thisthat, line 1: This class is not used after the removal of `container.id`. I would remove it and reintroduce it when your suggested spec changes are accepted :)

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/AwsResource.java
```diff
@@ -31,25 +30,26 @@
...
+    final Attributes.Builder attrBuilder = Attributes.newBuilder();
     for (AwsResource populator : populators) {
-      attrMap.putAll(populator.createAttributes());
-    }
-
-    Attributes.Builder attrBuilder = Attributes.newBuilder();
-    for (Map.Entry<String, AttributeValue> attr : attrMap.entrySet()) {
-      attrBuilder.setAttribute(attr.getKey(), attr.getValue());
+      Attributes attrs = populator.createAttributes();
+      attrs.forEach(
```
jkwatson, line 41: This makes me think that a "merge" function might be useful on the Attributes, like we have on the Resource itself.

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/DockerHelper.java
jkwatson: I agree. let's remove it if we're not using it.

jkwatson:
Looks like just a couple little things to clean up and this will be good to go, @wangzlei

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/AwsResource.java
wangzlei: A ready-to-use `merge` would be better. One more thing I found in unit test is if builder set the duplicated key, the final Attributes has the fist one but not the last one. It seems wired that usually the last one can override the first. Please see my AwsResourceTest.java

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EcsResource.java
wangzlei: removed null check and synEnv null test

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/DockerHelper.java
wangzlei: removed DockerHelper and unit test from PR

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/AwsResource.java
jkwatson: Yes, we currently keep the first, if there are duplicates. If you think this is not correct, please file an issue!

Issue #1314, opened 2020-06-06
Too many wrapper objects?


anuraaga:
**Is your feature request related to a problem? Please describe.**

Currently there are many wrapper objects like `TraceId`, `SpanId`. In other languages these may be cheap but in Java, allocations do affect app performance and instrumentation needs to take efforts to reduce the burden on an app. Lately there's some buzz on the interwebs about this topic folks may have seen :)

**Describe the solution you'd like**

These are mostly attributes of a span's context. It seems like `SpanContext` could store the values like the long IDs directly and provide any needed helpers like hex encoding.

Has this been given any thought? I found this but it doesn't seem to provide concrete motivation for wrappers.

https://github.com/open-telemetry/opentelemetry-java/issues/12

`SpanId` seems particularly aggregious, it's effectively just a `Long`.

https://github.com/open-telemetry/opentelemetry-java/blob/master/api/src/main/java/io/opentelemetry/trace/SpanId.java

Consider how much effort JVMs have put in to avoiding boxing as much as possible, it seems a bit weird to me that we are boxing everything by default.

bogdandrutu:
I agree with your concern, but we try to not expose the internal representation of the SpanId for example. We should think about an API that does not expose the internal details but may achieve what you suggest.

carlosalberto:
Something interesting here is how  we would proceed with cases where `TraceId` and `Spanid` exist by themselves (such as in the sampling API).

bogdandrutu:
@carlosalberto one option is to expose all methods on the SpanContext instead of exposing them on their own objects. I would like to see a prototype before making a decision

trask:
this would help efficiency of auto-instrumentation interop with manual instrumentation as well, where we currently have to wrap all of these objects individually

jkwatson:
> @carlosalberto one option is to expose all methods on the SpanContext instead of exposing them on their own objects. I would like to see a prototype before making a decision

I'd be up for prototyping this idea. Do you have an idea of how to limit the scope of the prototype? This will be a very large change, in general.

bogdandrutu:
@trask 
>this would help efficiency of auto-instrumentation interop with manual instrumentation as well, where we currently have to wrap all of these objects individually

Is there any way you can avoid wrapping and shading the API classes? Not an expert but it seems that a lot of wrapping/shading happens in auto-instrumentation.

bogdandrutu:
@jkwatson for me a successful prototype:
[...]

PR #1371, opened 2020-06-24
what if we just used byte arrays for spanid and traceid?

jkwatson:
This is probably a terrible idea, due to the fact that arrays are 100% mutable, but this experiment seemed like an interesting one to look at, in relation to #1314 

Don't merge this, please.


bogdandrutu:
You have the same problem of too many wrappers, every byte array is a new allocation so you don't resolve the problem of allocations.

jkwatson:
It does solve the issue that @trask brought up with having to deal with wrapper classes, but yes, it doesn't make allocations go away, due to the mutability of arrays.

jkwatson:
closing this, because it's a bad idea. ;)

PR #1373, opened 2020-06-25
enhance OTLP Exporter configuration

malafeev:
OTLP Exporter in [opentelemetry-java-instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation) is not configurable to:

- useTransportSecurity instead of usePlaintext
- insert values into Metadata headers (in ClientInterceptor)
 

From PR https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/567#issuecomment-648600894
Decided to try add configuration directly to OTLP Exporter.

### DIFF
--- a/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcSpanExporter.java
+++ b/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcSpanExporter.java
@@ -290,7 +290,6 @@ public final class OtlpGrpcSpanExporter implements SpanExporter {
       if (metadataValue != null) {
         for (String keyValueString : Splitter.on(';').split(metadataValue)) {
           final List<String> keyValue = Splitter.on('=').splitToList(keyValueString);
-          ;
           if (keyValue.size() == 2) {
             addHeader(keyValue.get(0), keyValue.get(1));
           }
--- a/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcSpanExporter.java
+++ b/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcSpanExporter.java
@@ -35,7 +35,6 @@ import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.ThreadSafe;
 
@@ -227,7 +226,7 @@ public final class OtlpGrpcSpanExporter implements SpanExporter {
      * @param value header value
      * @return this builder's instance
      */
-    public Builder addHeader(@Nonnull String key, @Nonnull String value) {
+    public Builder addHeader(String key, String value) {
       if (metadata == null) {
         metadata = new Metadata();
       }
--- a/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcSpanExporter.java
+++ b/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcSpanExporter.java
@@ -16,13 +16,21 @@
 
 package io.opentelemetry.exporters.otlp;
 
+import static io.grpc.Metadata.ASCII_STRING_MARSHALLER;
+
+import com.google.common.base.Splitter;
 import io.grpc.ManagedChannel;
+import io.grpc.ManagedChannelBuilder;
+import io.grpc.Metadata;
+import io.grpc.Metadata.Key;
+import io.grpc.stub.MetadataUtils;
 import io.opentelemetry.proto.collector.trace.v1.ExportTraceServiceRequest;
 import io.opentelemetry.proto.collector.trace.v1.TraceServiceGrpc;
 import io.opentelemetry.sdk.common.export.ConfigBuilder;
 import io.opentelemetry.sdk.trace.data.SpanData;
 import io.opentelemetry.sdk.trace.export.SpanExporter;
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
@@ -41,6 +49,10 @@ import javax.annotation.concurrent.ThreadSafe;
  * <ul>
  *   <li>{@code otel.otlp.span.timeout}: to set the max waiting time allowed to send each span
  *       batch.
+ *   <li>{@code otel.otlp.endpoint}: to set the endpoint to connect to.
+ *   <li>{@code otel.otlp.use.tls}: to set use or not TLS.
+ *   <li>{@code otel.otlp.metadata} to set key-value pairs separated by semicolon to pass as request
+ *       metadata.
  * </ul>
  *
  * <p>For environment variables, {@link OtlpGrpcSpanExporter} will look for the following names:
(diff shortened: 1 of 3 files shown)


exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcSpanExporter.java
```diff
@@ -41,13 +49,21 @@
...
+ *   <li>{@code otel.otlp.metadata} to set key-value pairs separated by semicolon to pass as request
+ *       metadata.
  * </ul>
  *
  * <p>For environment variables, {@link OtlpGrpcSpanExporter} will look for the following names:
  *
  * <ul>
  *   <li>{@code OTEL_OTLP_SPAN_TIMEOUT}: to set the max waiting time allowed to send each span
  *       batch.
+ *   <li>{@code OTEL_OTLP_ENDPOINT}: to set the endpoint to connect to.
```
jkwatson, line 63: There is a proposal in the specs to standardize environment variables for this sort of thing. Can you please add a comment to https://github.com/open-telemetry/opentelemetry-specification/pull/666/files with the needs here?

malafeev: https://github.com/open-telemetry/opentelemetry-specification/pull/666#issuecomment-649615527

jkwatson:
This looks good. We'll almost certainly need to update all of these env vars when the env var proposal gets merged.

carlosalberto:
Overall good, I definitely prefer to have it here than in the agent code. Two small things:

1) Let's check for non-null values for `setMetadata()` (and maybe `setEndpoint()`? or specifying `null` would clear it up?)
2) We need a clear, documented behavior for how to handle values when the user sets both the `channel` and `endpoint`. Does one override the other one? Does one of them has higher priority? 

My preference would be to simplify things and remove the option to specify a `ManagedChannel`, but probably that's something an advanced user/scenario would need ;)

exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcSpanExporter.java
```diff
@@ -146,11 +162,18 @@ public void shutdown() {
...
   public static class Builder extends ConfigBuilder<Builder> {
     private static final String KEY_SPAN_TIMEOUT = "otel.otlp.span.timeout";
+    private static final String KEY_ENDPOINT = "otel.otlp.endpoint";
+    private static final String KEY_USE_TLS = "otel.otlp.use.tls";
+    private static final String KEY_METADATA = "otel.otlp.metadata";
     private ManagedChannel channel;
     private long deadlineMs = 1_000; // 1 second
+    private String endpoint;
+    private boolean useTls;
+    private Metadata metadata;
```
anuraaga, line 172: `@Nullable`

exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcSpanExporter.java
```diff
@@ -146,11 +162,18 @@ public void shutdown() {
   /** Builder utility for this exporter. */
   public static class Builder extends ConfigBuilder<Builder> {
     private static final String KEY_SPAN_TIMEOUT = "otel.otlp.span.timeout";
+    private static final String KEY_ENDPOINT = "otel.otlp.endpoint";
+    private static final String KEY_USE_TLS = "otel.otlp.use.tls";
+    private static final String KEY_METADATA = "otel.otlp.metadata";
     private ManagedChannel channel;
     private long deadlineMs = 1_000; // 1 second
+    private String endpoint;
```
anuraaga, line 170: `@Nullable`

exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcSpanExporter.java
```diff
@@ -196,6 +274,27 @@ protected Builder fromConfigMap(
...
+      Boolean useTlsValue = getBooleanProperty(KEY_USE_TLS, configMap);
+      if (useTlsValue != null) {
+        this.setUseTls(useTlsValue);
+      }
+
+      String metadataValue = getStringProperty(KEY_METADATA, configMap);
+      if (metadataValue != null) {
+        for (String keyValueString : Splitter.on(';').split(metadataValue)) {
+          final List<String> keyValue = Splitter.on('=').splitToList(keyValueString);
+          ;
```
anuraaga, line 291: Extra semicolon

exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcSpanExporter.java
```diff
@@ -196,6 +274,27 @@ protected Builder fromConfigMap(
...
+
+      Boolean useTlsValue = getBooleanProperty(KEY_USE_TLS, configMap);
+      if (useTlsValue != null) {
+        this.setUseTls(useTlsValue);
+      }
+
+      String metadataValue = getStringProperty(KEY_METADATA, configMap);
+      if (metadataValue != null) {
+        for (String keyValueString : Splitter.on(';').split(metadataValue)) {
+          final List<String> keyValue = Splitter.on('=').splitToList(keyValueString);
```
anuraaga, line 290: Believe you can use `Splitter.withKeyValueSeparator`

anuraaga: Would be good to create a static final with the splitter

`private static final Splitter.MapSplitter METADATA_PROPERTY_SPLITTER = Splitter.on(';').trimResults().omitEmptyStrings().withKeyValueSplitter('=');

malafeev: `withKeyValueSeparator ` is marked with `@Beta`, should we use it?

exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcSpanExporter.java
malafeev: added

exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcSpanExporter.java
malafeev: added

exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcSpanExporter.java
anuraaga: Oof - good catch, I always forget that. Yeah let's not use it.

malafeev:
- @carlosalberto I add @Nullable and @Nonnull annotations.
- it would be great to get rid of explicit channel but who knows if it's needed by some case.
- I added in the javadoc of setChannel method  when channel is ignored and build using endpoint. I'm not sure where we can add more documentation because it's part of Builder only.

exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcSpanExporter.java
malafeev: added

carlosalberto:
Let's try this out - it seems @anuraaga's feedback has been all addressed, so we are ready to merge this? ;)

malafeev:
I think so.

trask:
I'd suggest only annotating with `@Nullable` as a general rule (and treating non-null as the default)

exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcSpanExporter.java
jkwatson: looks like it's still there...

malafeev: removed one more time

malafeev:
@trask I removed `@Nonnull`

PR #1376, opened 2020-06-26
Add semantic attributes for HTTP content length.

anuraaga:
https://github.com/open-telemetry/opentelemetry-specification/pull/641

### DIFF
--- a/api/src/main/java/io/opentelemetry/trace/attributes/SemanticAttributes.java
+++ b/api/src/main/java/io/opentelemetry/trace/attributes/SemanticAttributes.java
@@ -117,7 +117,7 @@ public final class SemanticAttributes {
   public static final StringAttributeSetter HTTP_RESPONSE_CONTENT_LENGTH =
       StringAttributeSetter.create("http.response_content_length");
   /**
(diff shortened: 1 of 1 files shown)


api/src/main/java/io/opentelemetry/trace/attributes/SemanticAttributes.java
```diff
@@ -98,6 +98,30 @@
...
+  public static final StringAttributeSetter HTTP_REQUEST_CONTENT_LENGTH_UNCOMPRESSED =
+      StringAttributeSetter.create("http.request_content_length_uncompressed");
+  /**
+   * The size of the response payload body, in bytes. For payloads using transport encoding, this is
+   * the compressed size.
+   */
+  public static final StringAttributeSetter HTTP_RESPONSE_CONTENT_LENGTH =
+      StringAttributeSetter.create("http.response_content_length");
+  /**
+   * The size of the uncompressed response payload body, in bytes. Only set for response that use
```
arminru, line 120:
```suggestion
   * The size of the uncompressed response payload body, in bytes. Only set for responses that use
```

PR #1378, opened 2020-06-28
Example Module for Zipkin Trace exporter

### LINKED ISSUE #1137: Create an example module for zipkin
It should probably look a lot like the jaeger one, as a starting point.

neeraj97:
Added example module for Zipkin ( issue #1137  ). This module looks a lot like jaeger example.
Fixes #1137

### DIFF
--- a/examples/zipkin/build.gradle
+++ b/examples/zipkin/build.gradle
@@ -1,0 +1,12 @@
+plugins {
+    id 'java'
+}
+
+description = 'OpenTelemetry Examples for Zipkin Exporter'
+ext.moduleName = "io.opentelemetry.examples.zipkin"
+
+dependencies {
+    compile("io.opentelemetry:opentelemetry-api:${opentelemetryVersion}")
+    compile("io.opentelemetry:opentelemetry-sdk:${opentelemetryVersion}")
+    compile("io.opentelemetry:opentelemetry-exporters-zipkin:${opentelemetryVersion}")
+}
--- a/examples/zipkin/src/main/java/io/opentelemetry/example/ZipkinExample.java
+++ b/examples/zipkin/src/main/java/io/opentelemetry/example/ZipkinExample.java
@@ -1,0 +1,87 @@
+package io.opentelemetry.example;
+
+import io.opentelemetry.OpenTelemetry;
+import io.opentelemetry.exporters.zipkin.ZipkinSpanExporter;
+import io.opentelemetry.sdk.OpenTelemetrySdk;
+import io.opentelemetry.sdk.trace.export.SimpleSpanProcessor;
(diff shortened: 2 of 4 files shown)


examples/zipkin/build.gradle
```diff
@@ -0,0 +1,12 @@
...
+}
+
+description = 'OpenTelemetry Examples for Zipkin Exporter'
+ext.moduleName = "io.opentelemetry.examples.zipkin"
+
+dependencies {
+    compile("io.opentelemetry:opentelemetry-api:${opentelemetryVersion}")
+    compile("io.opentelemetry:opentelemetry-sdk:${opentelemetryVersion}")
+    compile("io.opentelemetry:opentelemetry-exporters-zipkin:${opentelemetryVersion}")
+}
```
thisthat, line 12: nit: new line

examples/zipkin/src/main/java/io/opentelemetry/example/ZipkinExample.java
```diff
@@ -0,0 +1,87 @@
...
+    ZipkinExample example = new ZipkinExample(ip, port);
+    example.setupZipkinExporter();
+    example.myWonderfulUseCase();
+    // wait some seconds
+    try {
+      Thread.sleep(1000);
+    } catch (InterruptedException e) {
+    }
+    System.out.println("Bye");
+  }
```
thisthat, line 86: I just realized that instead of waiting, we should call `shutdown()` on the `TracerSdkProvider`. This holds for all the examples we have. This can be done in a separate PR :)

neeraj97: Yes, I think you are right! We should call shutdown( ) on the TracerSdkProvider instead of delaying/waiting. That would make flushes if there any spans and gracefully shutdowns.

jkwatson: while this is open, let's at least change this example to do the right thing, then we can have an issue to fix the rest. @neeraj97 can you create an issue to track it?

neeraj97: Yes I will create the issue to track it.

neeraj97: Created the issue #1382.
Closing this pull request as it would require changes suggested by issue #1382 



> Yes I will create the issue to track it.

neeraj97: > Yes I will create the issue to track it.


Created the issue #1382.
Closing this pull request as it would require changes suggested by issue #1382

PR #1379, opened 2020-06-29
Add semantic attribute for peer.service

anuraaga:
Corresponds to https://github.com/open-telemetry/opentelemetry-specification/pull/652 which is approved and just waiting to be merged. Sending this in the hopes that it speeds up getting both merged :)


jkwatson:
will merge as soon as the spec PR is merged.

PR #1386, opened 2020-06-30
Return Invalid Spans from the no-op Tracer when there is no SpanContext

### LINKED ISSUE #1385: Can we stop generating span and trace ids for no-op spans with no parent context?
We can save 3-5 allocations per no-op span if we can just return an invalid span from the no-op tracer when there is no parent context.

The behavior of generating random trace and span ids was introduced here:  #382 and references this issue: #392 . There is reference to a decision being made to return valid span ids in this case, but no reasoning provided in the issue or the PR as to why.

A note: if we change the behavior to return an Invalid span in this case, no tests fail in the entire project. This leads me to believe that this behavior is not required, but a vestige of an earlier context design which has been obviated by later context rework.

I also don't find any requirements for this behavior in the tracing specs.

If this is necessary for some reason, then we should definitely document it in the code so our future selves will fully understand the rationale.

jkwatson:
Resolves #1385

### DIFF
--- a/api/src/main/java/io/opentelemetry/trace/DefaultSpan.java
+++ b/api/src/main/java/io/opentelemetry/trace/DefaultSpan.java
@@ -18,7 +18,6 @@ package io.opentelemetry.trace;
 
 import io.opentelemetry.common.AttributeValue;
 import io.opentelemetry.common.Attributes;
-import java.util.Random;
 import javax.annotation.concurrent.Immutable;
 
 /**
@@ -32,7 +31,6 @@ import javax.annotation.concurrent.Immutable;
 @Immutable
 public final class DefaultSpan implements Span {
 
-  private static final Random random = new Random();
   private static final DefaultSpan INVALID = new DefaultSpan(SpanContext.getInvalid());
 
   /**
@@ -54,15 +52,6 @@ public final class DefaultSpan implements Span {
    */
   public static DefaultSpan create(SpanContext spanContext) {
     return new DefaultSpan(spanContext);
-  }
-
-  static DefaultSpan createRandom() {
-    return new DefaultSpan(
-        SpanContext.create(
-            TraceId.generateRandomId(random),
-            SpanId.generateRandomId(random),
-            TraceFlags.getDefault(),
-            TraceState.getDefault()));
   }
 
   private final SpanContext spanContext;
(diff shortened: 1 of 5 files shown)


carlosalberto:
The specific use case for this is exactly the one you described in the issue - creating blank-yet-valid `Span`s without a SDK.

I'm up for this change (after @bogdandrutu also weights in, as the `DefaultSpan` is a notion from OC, and he might know of a specific required scenario), but we have to be aware that it _may_ break some users' code.

(FWIW, I think `DefaultSpan.create(SpanContext)` should be good enough for users to consume, instead of the current randomized `DefaultSpan` creation).

jkwatson:
@carlosalberto I'm wracking my brain trying to figure out how this could break anyone's code. This doesn't impact incoming context-propagation...it's only when instrumentation in the process is initiating a new trace, but without an SDK installed. I can't imagine why you would ever want the API to create a phantom trace/span that would get propagated on outgoing calls. I'm willing to listen, but this seems like an categorically bad outcome to me, as the spanId created will never get reported anywhere, except by downstream systems as a parent span id that can never be associated with anything.

carlosalberto:
@jkwatson Fair enough. Breaking user code is mostly for us to be aware, not to not make this change ;)

bogdandrutu:
There are some requirements that I want to make sure we have:
1. Never return null Span to not force user code to always check for null. even if no SDK / implementation installed.
2. Always propagate what we receive - arguably if no implementation/SDK we should or not generate new span ids.
3. For root traces return a non-null Span, in terms of valid IDs (I think this PR changes to use invalid IDs) we can argue both ways - pros can be things like logs are associated with valid IDs, cons are we never going to export these spans and we never going to set recording/sampling bit.

I think we need to make this change consistent across all languages, I tend to be pro this proposal for the 3rd case.

jkwatson:
> 1. Never return null Span to not force user code to always check for null. even if no SDK / implementation installed.

100% - this PR doesn't alter that

> 2. Always propagate what we receive - arguably if no implementation/SDK we should or not generate new span ids.

Propagation should definitely be preserved, even with no SDK. This PR doesn't change that behavior.

> 3. For root traces return a non-null Span, in terms of valid IDs (I think this PR changes to use invalid IDs) we can argue both ways - pros can be things like logs are associated with valid IDs, cons are we never going to export these spans and we never going to set recording/sampling bit.

Logs isn't something I had considered. Would we want imaginary tradeIds/spanIds in logs, if there is no SDK installed, though?

> I think we need to make this change consistent across all languages, I tend to be pro this proposal for the 3rd case.

I looked through the specs and didn't find any reference to this particular case. I'll open up a spec issue to track it.

Thanks for looking!

jkwatson:
Created a spec issue to make sure this behavior is the same, across all languages:  https://github.com/open-telemetry/opentelemetry-specification/issues/689

jkwatson:
Whatever the results of the various discussion in the specs, it doesn't seem like this particular change would conflict with any of them, and seems in line with the most possible broad interpretation of "no-op" for the tracing APIs.

PR #1391, opened 2020-07-01
Start 0.7.0 development cycle

carlosalberto:


carlosalberto:
@jkwatson Yes, that will happen in the 0.6.x branch ;)

jkwatson:
> @jkwatson Yes, that will happen in the 0.6.x branch ;)

yeah, I definitely need to do one of these releases so I grok the whole dance

PR #1395, opened 2020-07-06
Simplified BatchSpanProcessor

### LINKED ISSUE #729: Simplify the BatchSpansProcessor

neeraj97:
Fixes #729 
Used ScheduledExecutorService to schedule background worker to export spans with fixed delays and also forced to execute the worker if the queue size is greater than the halfMaxQueueSize and ScheduledExecutorService  is idle. BlockingQueue is used to avoid manual locks.

### DIFF
--- a/exporters/jaeger/README.md
+++ b/exporters/jaeger/README.md
@@ -6,18 +6,44 @@ This is the OpenTelemetry exporter, sending span data to Jaeger via gRPC.
 
 ## Configuration
 
-Jaeger exporter can be configured by system properties and environmental variables.
-These configuration properties work only for `Jaeger.GrpcSpanExporter.installDefault()`.
+The Jaeger gRPC span exporter can be configured programmatically.
 
-* `JAEGER_ENDPOINT` - agents's gRPC endpoint e.g. `localhost:14250`
-* `JAEGER_SERVICE_NAME` - service name e.g. `my-deployment`
+An example of simple Jaeger gRPC exporter initialization. In this case
+spans will be sent to a Jaeger gRPC endpoint running on `localhost`:
 
-An example of simples Jaeger gRPC exporter initialization. In this case
-spans will be sent to Jaeger agent running on `localhost`:
 ```java
-Builder builder = JaegerGrpcSpanExporter.Builder.fromEnv();
-builder.install(OpenTelemetrySdk.getTracerProvider());
+JaegerGrpcSpanExporter exporter =
+        JaegerGrpcSpanExporter.newBuilder()
+            .setEndpoint("localhost:14250")
+            .setServiceName("my-service")
+            .build();
 ```
+
+Service name and Endpoint can be also configured via environment variables or system properties.
+
+```java
+// Using environment variables
+JaegerGrpcSpanExporter exporter = 
+        JaegerGrpcSpanExporter.newBuilder()
+            .readEnvironmentVariables()
+            .build()
+```
+
+```java
+// Using system properties
+JaegerGrpcSpanExporter exporter = 
+        JaegerGrpcSpanExporter.newBuilder()
+            .readSystemProperties()
+            .build()
+```
+
+The Jaeger gRPC span exporter will look for the following environment variables / system properties:
+* `OTEL_JAEGER_SERVICE_NAME` / `otel.jaeger.service.name`
+* `OTEL_JAEGER_ENDPOINT` / `otel.jaeger.endpoint`
+
+## Compatibility
+
+As with the OpenTelemetry SDK itself, this exporter is compatible with Java 7+ and Android API level 24+.
 
 ## Proto files
 
--- a/exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java
+++ b/exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java
@@ -21,13 +21,13 @@ import io.grpc.ManagedChannelBuilder;
 import io.opentelemetry.exporters.jaeger.proto.api_v2.Collector;
 import io.opentelemetry.exporters.jaeger.proto.api_v2.CollectorServiceGrpc;
 import io.opentelemetry.exporters.jaeger.proto.api_v2.Model;
-import io.opentelemetry.sdk.trace.TracerSdkProvider;
+import io.opentelemetry.sdk.common.export.ConfigBuilder;
 import io.opentelemetry.sdk.trace.data.SpanData;
-import io.opentelemetry.sdk.trace.export.BatchSpanProcessor;
 import io.opentelemetry.sdk.trace.export.SpanExporter;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.util.Collection;
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -37,8 +37,6 @@ import javax.annotation.concurrent.ThreadSafe;
 @ThreadSafe
 public final class JaegerGrpcSpanExporter implements SpanExporter {
   private static final Logger logger = Logger.getLogger(JaegerGrpcSpanExporter.class.getName());
-  private static final String JAEGER_SERVICE_NAME = "JAEGER_SERVICE_NAME";
-  private static final String JAEGER_ENDPOINT = "JAEGER_ENDPOINT";
   private static final String CLIENT_VERSION_KEY = "jaeger.version";
   private static final String CLIENT_VERSION_VALUE = "opentelemetry-java";
   private static final String DEFAULT_JAEGER_ENDPOINT = "localhost:14250";
@@ -165,8 +163,12 @@ public final class JaegerGrpcSpanExporter implements SpanExporter {
   }
 
   /** Builder utility for this exporter. */
-  public static class Builder {
-    private String serviceName;
+  public static class Builder extends ConfigBuilder<Builder> {
+    private static final String KEY_SERVICE_NAME = "otel.jaeger.service.name";
+    private static final String KEY_ENDPOINT = "otel.jaeger.endpoint";
+
+    private String serviceName = UNKNOWN;
+    private String endpoint = DEFAULT_JAEGER_ENDPOINT;
     private ManagedChannel channel;
     private long deadlineMs = 1_000; // 1 second
 
@@ -182,13 +184,26 @@ public final class JaegerGrpcSpanExporter implements SpanExporter {
     }
 
     /**
-     * Sets the managed chanel to use when communicating with the backend. Required.
+     * Sets the managed chanel to use when communicating with the backend. Takes precedence over
+     * {@link #setEndpoint(String)} if both are called.
      *
      * @param channel the channel to use.
      * @return this.
      */
     public Builder setChannel(ManagedChannel channel) {
       this.channel = channel;
+      return this;
+    }
+
+    /**
+     * Sets the Jaeger endpoint to connect to. Optional, defaults to "localhost:14250".
+     *
+     * @param endpoint The Jaeger endpoint URL, ex. "jaegerhost:14250".
+     * @return this.
+     * @since 0.7.0
+     */
+    public Builder setEndpoint(String endpoint) {
+      this.endpoint = endpoint;
       return this;
     }
 
@@ -204,17 +219,25 @@ public final class JaegerGrpcSpanExporter implements SpanExporter {
     }
 
     /**
-     * Creates builder from system properties and environmental variables: {@code JAEGER_ENDPOINT}
-     * e.g. {@code localhost:14250} and {@code JAEGER_SERVICE_NAME} e.g. {@code my-deployment}.
+     * Sets the configuration values from the given configuration map for only the available keys.
      *
-     * @return thes builder's instance
+     * @param configMap {@link Map} holding the configuration values.
+     * @return this.
+     * @since 0.7.0
      */
-    public static Builder fromEnv() {
-      Builder builder = new Builder();
-      String host = getProperty(JAEGER_ENDPOINT, DEFAULT_JAEGER_ENDPOINT);
-      builder.channel = ManagedChannelBuilder.forTarget(host).usePlaintext().build();
-      builder.serviceName = getProperty(JAEGER_SERVICE_NAME, UNKNOWN);
-      return builder;
+    @Override
+    protected Builder fromConfigMap(
+        Map<String, String> configMap, NamingConvention namingConvention) {
+      configMap = namingConvention.normalize(configMap);
+      String stringValue = getStringProperty(KEY_SERVICE_NAME, configMap);
+      if (stringValue != null) {
+        this.setServiceName(stringValue);
+      }
+      stringValue = getStringProperty(KEY_ENDPOINT, configMap);
+      if (stringValue != null) {
+        this.setEndpoint(stringValue);
+      }
+      return this;
     }
 
     /**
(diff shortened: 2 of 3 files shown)


sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -88,8 +91,10 @@
   private static final String EXPORTER_THREAD_NAME =
       BatchSpanProcessor.class.getSimpleName() + "_ExporterThread";
   private final Worker worker;
-  private final Thread workerThread;
   private final boolean sampled;
+  private final ScheduledExecutorService scheduler;
+  private final ScheduledFuture<?> scheduledFutureWithFixedDelay;
+  private Future<?> imperativeSubmission;
```
jkwatson, line 97: I think this reference needs to either be volatile, or implemented with an AtomicReference, since it's assignment needs to be visible across threads.

jkwatson: Also, maybe renamed this to "forcedExportResult" or something along those lines? Not sure exactly what the best name is, but th
is name tripped me up for a bit.

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -163,81 +192,56 @@ public void forceFlush() {
...
-      this.scheduleDelayMillis = scheduleDelayMillis;
       this.maxQueueSize = maxQueueSize;
       this.halfMaxQueueSize = maxQueueSize >> 1;
       this.maxExportBatchSize = maxExportBatchSize;
-      this.spansList = new ArrayList<>(maxQueueSize);
+      this.spansList = new ArrayBlockingQueue<>(maxQueueSize);
       this.exporterTimeoutMillis = exporterTimeoutMillis;
     }
 
+    private static List<SpanData> createSpanDataForExport(
```
jkwatson, line 217: Can you move this back down below where it's used?

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -98,16 +103,22 @@ private BatchSpanProcessor(
...
+  }
+
+  /**
+   * Returns a new Builder for {@link BatchSpanProcessor}.
+   *
+   * @param spanExporter the {@code SpanExporter} to where the Spans are pushed.
+   * @return a new {@link BatchSpanProcessor}.
+   * @throws NullPointerException if the {@code spanExporter} is {@code null}.
+   */
+  public static Builder newBuilder(SpanExporter spanExporter) {
```
jkwatson, line 120: Why move this method as a part of this PR?

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -124,6 +135,17 @@ public void onEnd(ReadableSpan span) {
       return;
     }
     worker.addSpan(span);
+
+    if (worker.spansList.size() >= worker.halfMaxQueueSize) {
```
jkwatson, line 139: Can this logic live down in the worker itself, rather than spreading it across two places?

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -98,16 +103,22 @@ private BatchSpanProcessor(
...
-            spanExporter,
-            scheduleDelayMillis,
-            maxQueueSize,
-            maxExportBatchSize,
-            exporterTimeoutMillis);
-    this.workerThread = new DaemonThreadFactory(WORKER_THREAD_NAME).newThread(worker);
-    this.workerThread.start();
+    this.worker = new Worker(spanExporter, maxQueueSize, maxExportBatchSize, exporterTimeoutMillis);
     this.sampled = sampled;
+    scheduler = Executors.newScheduledThreadPool(1, new DaemonThreadFactory(WORKER_THREAD_NAME));
```
anuraaga, line 108: `newSingleThreadScheduledExecutor`

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
anuraaga: From what I can tell, it's pretty easy to schedule something immediately many times because of the thread safety which will be hard to solve. I think instead of a future, we can just use an `AtomicBoolean`

```java
onEnd() {
  if (forcingExport.compareAndSet(false, true)) {
    scheduler.submit(worker);
  }
}

...

Worker.run() {
...

// Last line
forcingExport.compareAndSet(true, false);
```

anuraaga: That being said, this forced export logic seems suspicious. If the queue isn't keeping up, then it seems the worker will run consecutively forever ignoring `exportTimeoutMillis` completely because `worker.spansList.size() >= worker.halfMaxQueueSize` will always be true any time `forcingExport` is false in the pathological case. I think users would rather have spans dropped than this heavyweight thread running all the time?

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -163,81 +192,56 @@ public void forceFlush() {
...
-          return;
-        }
-        // TODO: Record a gauge for referenced spans.
-        spansList.add(span);
-        // Notify the worker thread that at half of the queue is available. It will take
-        // time anyway for the thread to wake up.
-        if (spansList.size() >= halfMaxQueueSize) {
-          monitor.notifyAll();
-        }
+      if (spansList.size() == maxQueueSize) {
```
anuraaga, line 229: This doesn't look threadsafe

```
if (!spansList.offer(span)) {
  droppedSpans.add(1);
}
```

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -163,81 +192,56 @@ public void forceFlush() {
...
-    private final long scheduleDelayMillis;
     private final int maxQueueSize;
     private final int maxExportBatchSize;
     private final int halfMaxQueueSize;
-    private final Object monitor = new Object();
     private final int exporterTimeoutMillis;
-
-    @GuardedBy("monitor")
-    private final List<ReadableSpan> spansList;
+    private final BlockingQueue<ReadableSpan> spansList;
```
anuraaga, line 202: It's not a list it's a queue, rename to just `spans` to prevent confusion

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -163,81 +192,56 @@ public void forceFlush() {
...
+      this.spansList = new ArrayBlockingQueue<>(maxQueueSize);
       this.exporterTimeoutMillis = exporterTimeoutMillis;
     }
 
+    private static List<SpanData> createSpanDataForExport(
+        List<ReadableSpan> spanList, int startIndex, int numberToTake) {
+      List<SpanData> spanDataBuffer = new ArrayList<>(numberToTake);
+      for (int i = startIndex; i < numberToTake; i++) {
+        spanDataBuffer.add(spanList.get(i).toSpanData());
+        // Remove the reference to the ReadableSpan to allow GC to free the memory.
```
anuraaga, line 222: Hmm, since the list will be GC'd at the same time as the items, I don't think this actually helps

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -247,11 +251,10 @@ private void shutdown() {
     }
 
     private void forceFlush() {
-      ArrayList<ReadableSpan> spansCopy;
-      synchronized (monitor) {
-        spansCopy = new ArrayList<>(spansList);
-        spansList.clear();
-      }
+      // Draining all the batched spans in a separate list
```
anuraaga, line 254: Not sure if we really need this comment, it's obvious for blocking queues

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -163,81 +192,56 @@ public void forceFlush() {
...
       this.maxQueueSize = maxQueueSize;
       this.halfMaxQueueSize = maxQueueSize >> 1;
       this.maxExportBatchSize = maxExportBatchSize;
-      this.spansList = new ArrayList<>(maxQueueSize);
+      this.spansList = new ArrayBlockingQueue<>(maxQueueSize);
       this.exporterTimeoutMillis = exporterTimeoutMillis;
     }
 
+    private static List<SpanData> createSpanDataForExport(
+        List<ReadableSpan> spanList, int startIndex, int numberToTake) {
```
anuraaga, line 218: Fix `numberToTake` to `endIndex`

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -163,81 +192,56 @@ public void forceFlush() {
...
+      List<SpanData> spanDataBuffer = new ArrayList<>(numberToTake);
```
anuraaga, line 219: Due to th edeceptive name `numberToTake`, this looks like a bug, the list gets bigger and bigger and bigger :)

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -163,81 +192,56 @@ public void forceFlush() {
...
-              }
-            } while (spansList.isEmpty());
-          }
-          spansCopy = new ArrayList<>(spansList);
-          spansList.clear();
-        }
-        // Execute the batch export outside the synchronized to not block all producers.
-        exportBatches(spansCopy);
-      }
+      // Draining all the batched spans in a separate list
```
anuraaga, line 239: Ditto

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -163,81 +192,56 @@ public void forceFlush() {
...
+      // avoid blocking the producer thread.
+      ArrayList<ReadableSpan> spansCopy = new ArrayList<>();
+      spansList.drainTo(spansCopy);
+      // Execute the batch export outside the synchronized to not block all producers.
```
anuraaga, line 243: No more synchronized, remove

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
anuraaga: Actually, looks like we can remove this method now it is identical to run

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -303,32 +295,19 @@ public void run() {
...
-    private static final String KEY_SAMPLED = "otel.bsp.export.sampled";
-
     @VisibleForTesting static final long DEFAULT_SCHEDULE_DELAY_MILLIS = 5000;
     @VisibleForTesting static final int DEFAULT_MAX_QUEUE_SIZE = 2048;
     @VisibleForTesting static final int DEFAULT_MAX_EXPORT_BATCH_SIZE = 512;
     @VisibleForTesting static final int DEFAULT_EXPORT_TIMEOUT_MILLIS = 30_000;
     @VisibleForTesting static final boolean DEFAULT_EXPORT_ONLY_SAMPLED = true;
-
+    private static final String KEY_SCHEDULE_DELAY_MILLIS = "otel.bsp.schedule.delay";
+    private static final String KEY_MAX_QUEUE_SIZE = "otel.bsp.max.queue";
```
anuraaga, line 307: Why did you change this ordering? We usually have private before non-private

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
jkwatson: I think this is a good question to ask. I believe it applies to this class on the main branch as well. Maybe this should be addressed first? Can you create an issue for it?

[also, yes, just using an AtomicBoolean seems like a much simpler solution if we *do* want to force exports when the queue is full, before the interval is complete]

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
jkwatson: I think this is the number to take if you look how it's used in the exportBatches method.

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
jkwatson: Agreed. I think this is a holdover from the previous implementation.

jkwatson:
There are at least 2 tests that are broken with this implementation. I would hope that the tests would all still pass with a change just in the implementation, rather than changing the behavior.

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -98,16 +103,22 @@ private BatchSpanProcessor(
...
-            maxQueueSize,
-            maxExportBatchSize,
-            exporterTimeoutMillis);
-    this.workerThread = new DaemonThreadFactory(WORKER_THREAD_NAME).newThread(worker);
-    this.workerThread.start();
+    this.worker = new Worker(spanExporter, maxQueueSize, maxExportBatchSize, exporterTimeoutMillis);
     this.sampled = sampled;
+    scheduler = Executors.newScheduledThreadPool(1, new DaemonThreadFactory(WORKER_THREAD_NAME));
+    scheduledFutureWithFixedDelay =
+        scheduler.scheduleWithFixedDelay(worker, 0, scheduleDelayMillis, TimeUnit.MILLISECONDS);
```
jkwatson, line 110: I don't think you want to schedule an export immediately on startup, do you? I'd recommend delaying by the specified delay before the first export.

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
anuraaga: Really? We iterate from startIndex to this number in the loop

`for (int i = startIndex; i < numberToTake; i++) {`

If it was actually the number to take, it'd be 

`for (int i = startIndex; i < i + numberToTake; i++) {`

wouldn't it?

jkwatson: yikes! This is a really bad bug. Thanks for the sharp eye on that! Good thing no one has been trying to send too many spans with this thing yet!

jkwatson: ok, it wasn't really all that bad a bug, but @anuraaga is correct. The parameter is misnamed, both here and in the caller. This has been fixed on the main branch with #1400 , so if you could fix it here as well, that would be good.

neeraj97:
@jkwatson @anuraaga Thank you for the detailed review. I got some amazing insights into problem. I will work on them.

PR #1399, opened 2020-07-08
Add a toBuilder method to Labels

jkwatson:
While writing some manual instrumentation, I found it very useful to keep a constant Labels instance, but then append to them when actually recording metrics (to add Span status, for instance). This method will help do that more efficiently than having to manually copy the labels from one instance to a new builder.

### DIFF
--- a/api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
+++ b/api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
@@ -136,12 +136,12 @@ abstract class ImmutableKeyValuePairs<V> implements ReadableKeyValuePairs<V> {
         new KeyValueConsumer<V>() {
           @Override
           public void consume(String key, V value) {
-            sb.append(key).append(":").append(value).append(",");
+            sb.append(key).append("=").append(value).append(", ");
           }
         });
     // get rid of that last pesky comma
     if (sb.length() > 1) {
-      sb.deleteCharAt(sb.length() - 1);
+      sb.setLength(sb.length() - 2);
     }
     sb.append("}");
     return sb.toString();
--- a/api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
+++ b/api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
@@ -128,4 +128,22 @@ abstract class ImmutableKeyValuePairs<V> implements ReadableKeyValuePairs<V> {
     data[b] = keyA;
     data[b + 1] = valueA;
   }
+
+  @Override
+  public String toString() {
+    final StringBuilder sb = new StringBuilder("{");
+    forEach(
+        new KeyValueConsumer<V>() {
+          @Override
+          public void consume(String key, V value) {
+            sb.append(key).append(":").append(value).append(",");
+          }
+        });
+    // get rid of that last pesky comma
+    if (sb.length() > 1) {
+      sb.deleteCharAt(sb.length() - 1);
+    }
+    sb.append("}");
+    return sb.toString();
+  }
 }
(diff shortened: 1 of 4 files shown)


api/src/test/java/io/opentelemetry/common/LabelsTest.java
```diff
@@ -124,4 +124,14 @@ public void builder() {
                 "key1", "value1",
                 "key2", "value2"));
   }
+
+  @Test
+  public void toBuilder() {
+    Labels initial = Labels.of("one", "a");
+    Labels second = initial.toBuilder().setLabel("two", "b").build();
+    assertThat(initial.size()).isEqualTo(1);
```
anuraaga, line 132: Looks like we don't really need the size assertions

anuraaga: Well maybe they help - don't think I can find a `toString` for the key value pair classes which would be important for the equals assertions when they fail.

jkwatson: My thinking was that, with such a non-trivial implementation internally, that a few extra assertions wouldn't hurt. The size isn't the size of the arraylist, for example. 

Also, a toString is a really good idea! Not sure why I didn't put that in originally.

jkwatson: I added a toString, just for good measure.

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
```diff
@@ -128,4 +128,22 @@ private static void swap(Object[] data, int a, int b) {
...
+    forEach(
+        new KeyValueConsumer<V>() {
+          @Override
+          public void consume(String key, V value) {
+            sb.append(key).append(":").append(value).append(",");
+          }
+        });
+    // get rid of that last pesky comma
+    if (sb.length() > 1) {
+      sb.deleteCharAt(sb.length() - 1);
```
anuraaga, line 144: Usually see `sb.setLength(sb.length() - 1)` for dealing with pesky commas. Checked `deleteCharAt` and it doesn't seem to avoid `System.arrayCopy` even for this simple case

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
```diff
@@ -128,4 +128,22 @@ private static void swap(Object[] data, int a, int b) {
...
   }
+
+  @Override
+  public String toString() {
+    final StringBuilder sb = new StringBuilder("{");
+    forEach(
+        new KeyValueConsumer<V>() {
+          @Override
+          public void consume(String key, V value) {
+            sb.append(key).append(":").append(value).append(",");
```
anuraaga, line 139:
```suggestion
            sb.append(key).append("=").append(value).append(", ");
```

Nit but maybe we should go ahead and match `Map`? https://github.com/openjdk/jdk/blob/master/test/jdk/java/util/AbstractMap/ToString.java#L46

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
jkwatson: oh! good tip. I hadn't looked that deeply into the implementation of `deleteCharAt`. Probably not a big deal for a debugging toString method, but no reason not to change it, now that I know!

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
jkwatson: Sure. Updated, and added a test to make sure it matches.

jkwatson: Of course, one downside to having this match exactly is that if you had a test that was comparing Labels and a Map, the toString would match, but of course, they will never be equal!

PR #1403, opened 2020-07-09
Update to the latest versions of errorprone, grpc and autovalue

jkwatson:
Mostly, clean up new errorprone warnings

### DIFF
--- a/exporters/inmemory/src/test/java/io/opentelemetry/exporters/inmemory/InMemorySpanExporterTest.java
+++ b/exporters/inmemory/src/test/java/io/opentelemetry/exporters/inmemory/InMemorySpanExporterTest.java
@@ -23,6 +23,10 @@ import io.opentelemetry.sdk.trace.data.SpanData;
 import io.opentelemetry.sdk.trace.data.test.TestSpanData;
 import io.opentelemetry.sdk.trace.export.SimpleSpanProcessor;
 import io.opentelemetry.sdk.trace.export.SpanExporter.ResultCode;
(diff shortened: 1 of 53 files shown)


exporters/inmemory/src/test/java/io/opentelemetry/exporters/inmemory/InMemorySpanExporterTest.java
```diff
@@ -103,12 +107,12 @@ public void export_ReturnCode() {
   static SpanData makeBasicSpan() {
     return TestSpanData.newBuilder()
         .setHasEnded(true)
-        .setTraceId(io.opentelemetry.trace.TraceId.getInvalid())
-        .setSpanId(io.opentelemetry.trace.SpanId.getInvalid())
+        .setTraceId(TraceId.getInvalid())
```
carlosalberto, line 110: Is this style change (use an import instead of the full path for a class) imposed by this errorprone update?

*Personally*, when a class is used only once in a file (such as `SpanId` here), I see no problem using its full path ;)

jkwatson: yep..and this is one of the errorprone warnings that I actually agree with. I don't think the "never import inner classes" is all that great.

jkwatson: also, sounds like you need to join the real IDE generation, @carlosalberto .  ;)

PR #1407, opened 2020-07-10
Call initTracer just once at beginning of main in http example

evantorrie:
This fixes a bug where each time a new HttpClient is instantiated, it adds a new SpanExporter to the
global OpenTelemetry SDK instance.

### DIFF
--- a/examples/http/src/main/java/io/opentelemetry/example/http/HttpClient.java
+++ b/examples/http/src/main/java/io/opentelemetry/example/http/HttpClient.java
@@ -49,7 +49,7 @@ public class HttpClient {
         }
       };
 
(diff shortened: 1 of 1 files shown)


examples/http/src/main/java/io/opentelemetry/example/http/HttpClient.java
```diff
@@ -49,7 +49,7 @@ public void set(HttpURLConnection carrier, String key, String value) {
         }
       };
 
-  private void initTracer() {
+  private static void initTracer() {
```
jkwatson, line 52: Can we also change this to `initTracerSdk()`, since we're not really initializing a tracer here.

PR #1410, opened 2020-07-13
Add semantic attributes for messaging

### LINKED ISSUE #1401: Add SemanticAttributes for messaging systems
A lot attributes defined in https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md are currently missing from `io.opentelemetry.trace.attributes.SemanticAttributes`

iNikem:
Closes #1401

### DIFF
--- a/api/src/main/java/io/opentelemetry/trace/attributes/SemanticAttributes.java
+++ b/api/src/main/java/io/opentelemetry/trace/attributes/SemanticAttributes.java
@@ -201,6 +201,13 @@ public final class SemanticAttributes {
   /** The compressed size of the message payload in bytes. */
   public static final LongAttributeSetter MESSAGING_MESSAGE_PAYLOAD_COMPRESSED_SIZE_BYTES =
       LongAttributeSetter.create("messaging.message_payload_compressed_size_bytes");
+  /**
+   * A string identifying which part and kind of message consumption this span describes: either
+   * "receive" or "process". If the operation is "send", this attribute must not be set: the
+   * operation can be inferred from the span kind in that case.
+   */
+  public static final StringAttributeSetter MESSAGING_OPERATION =
+      StringAttributeSetter.create("messaging.operation");
 
   private SemanticAttributes() {}
 }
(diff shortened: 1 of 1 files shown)


api/src/main/java/io/opentelemetry/trace/attributes/SemanticAttributes.java
```diff
@@ -154,5 +154,53 @@
...
+  /**
+   * The (uncompressed) size of the message payload in bytes. Also use this attribute if it is
+   * unknown whether the compressed or uncompressed payload size is reported.
+   */
+  public static final LongAttributeSetter MESSAGING_MESSAGE_PAYLOAD_SIZE_BYTES =
+      LongAttributeSetter.create("messaging.message_payload_size_bytes");
+  /** The compressed size of the message payload in bytes. */
+  public static final LongAttributeSetter MESSAGING_MESSAGE_PAYLOAD_COMPRESSED_SIZE_BYTES =
+      LongAttributeSetter.create("messaging.message_payload_compressed_size_bytes");
+
```
arminru, line 204:
```suggestion
  /**
   * A string identifying which kind of message consumption this span describes.
   * Either "receive" or "process".
   * For message consumers only.
   */
  public static final StringAttributeSetter MESSAGING_CONVERSATION_ID =
      StringAttributeSetter.create("messaging.operation");

```

iNikem: Thanks! Added.

arminru: Sure. It would make sense to add the "for message consumer only" indication as well. Users should consult the spec anyway but will likely not read up on in if part of it is already mentioned in their IDE quick definition.

arminru: Sure!

It would make sense to add the "for message consumer only" indication as well. Users should consult the spec anyway but will likely not read up on it if part of it is already shown in their IDE quick definition.

iNikem: Doesn't current javadoc says this in first sentence? "A string identifying which part and kind of message **consumption** this span describes"

iNikem:
@open-telemetry/java-approvers Is this good for merge?

api/src/main/java/io/opentelemetry/trace/attributes/SemanticAttributes.java
arminru: I'd personally make it more explicit but it should be fine as it is as well.

PR #1411, opened 2020-07-13
Refactor the creation of instrument batchers to be in a central location

jkwatson:
This should allow us to build a Views API into the SDK.
Note: this PR is intended to be a pure internal refactoring, with no change in current functionality.

### DIFF
--- a/sdk/src/main/java/io/opentelemetry/sdk/metrics/Batchers.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/metrics/Batchers.java
@@ -34,36 +34,38 @@ import java.util.List;
 import java.util.Map;
 
 /** A collection of available Batchers. */
-final class Batchers {
+public final class Batchers {
 
-  static Batcher getNoop() {
+  public static Batcher getNoop() {
     return Noop.INSTANCE;
   }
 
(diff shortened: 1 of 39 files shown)


sdk/src/main/java/io/opentelemetry/sdk/metrics/Batchers.java
```diff
@@ -34,36 +34,38 @@
...
-final class Batchers {
+public final class Batchers {
 
-  static Batcher getNoop() {
+  public static Batcher getNoop() {
     return Noop.INSTANCE;
   }
 
-  static Batcher getCumulativeAllLabels(
+  /** TODO: javadoc me. */
```
carlosalberto, line 43: Were these ones left empty on purpose?

jkwatson: nope! I'll fix that before merging.

jkwatson: done!

jkwatson:
@carlosalberto FYI, I moved the ViewRegistry class up to the main `metrics` package, which cleaned up a bunch of visibility changes on other SDK internals.

PR #1412, opened 2020-07-14
Prototype of a very basic Aggregation-configuration API in the SDK.

jkwatson:
Note: related OTEP: https://github.com/open-telemetry/oteps/pull/126

This prototype supports the following:
  * select instruments to customize by instrument type and name.
  * specify the aggregation and the temporality of the aggregation to be used.

An example of how you could use it:

```java
// get a handle to the MeterSdkProvider
 MeterSdkProvider meterProvider = OpenTelemetrySdk.getMeterProvider();

 // create a selector to select which instruments to customize:
 InstrumentSelector instrumentSelector = InstrumentSelector.newBuilder()
  .instrumentType(InstrumentType.COUNTER)
  .build();

 // create a specification of how you want the metrics aggregated:
 ViewSpecification viewSpecification = 
      ViewSpecification.create(Aggregations.minMaxSumCount(), Temporality.DELTA);

 //register the view with the MeterSdkProvider
 meterProvider.registerView(instrumentSelector, viewSpecification);
```

### DIFF
--- a/sdk/src/main/java/io/opentelemetry/sdk/metrics/MeterSdkProvider.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/metrics/MeterSdkProvider.java
@@ -153,7 +153,9 @@ public final class MeterSdkProvider implements MeterProvider {
    * MeterSdkProvider meterProvider = OpenTelemetrySdk.getMeterProvider();
    *
    * // create a selector to select which instruments to customize:
-   * InstrumentSelector instrumentSelector = InstrumentSelector.create(InstrumentType.COUNTER);
+   * InstrumentSelector instrumentSelector = InstrumentSelector.newBuilder()
+   *   .instrumentType(InstrumentType.COUNTER)
+   *   .build();
    *
    * // create a specification of how you want the metrics aggregated:
    * ViewSpecification viewSpecification =
--- a/sdk/src/main/java/io/opentelemetry/sdk/metrics/MeterSdkProvider.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/metrics/MeterSdkProvider.java
@@ -41,11 +41,12 @@ public final class MeterSdkProvider implements MeterProvider {
 
   private final MeterSdkComponentRegistry registry;
   private final MetricProducer metricProducer;
+  private final ViewRegistry viewRegistry = new ViewRegistry();
 
   private MeterSdkProvider(Clock clock, Resource resource) {
     this.registry =
         new MeterSdkComponentRegistry(
-            MeterProviderSharedState.create(clock, resource), new ViewRegistry());
+            MeterProviderSharedState.create(clock, resource), viewRegistry);
     this.metricProducer = new MetricProducerSdk(this.registry);
   }
 
@@ -140,6 +141,31 @@ public final class MeterSdkProvider implements MeterProvider {
     public MeterSdk newComponent(InstrumentationLibraryInfo instrumentationLibraryInfo) {
       return new MeterSdk(meterProviderSharedState, instrumentationLibraryInfo, viewRegistry);
     }
+  }
+
+  /**
+   * Register a view with the given {@link InstrumentSelector}.
+   * <p>
+   * Example on how to register a view:
+   *
+   * <pre>{@code
+   * // get a handle to the MeterSdkProvider
+   * MeterSdkProvider meterProvider = OpenTelemetrySdk.getMeterProvider();
+   *
+   * // create a selector to select which instruments to customize:
+   * InstrumentSelector instrumentSelector = InstrumentSelector.create(InstrumentType.COUNTER);
+   *
+   * // create a specification of how you want the metrics aggregated:
+   * ViewSpecification viewSpecification = ViewSpecification.create(Aggregations.minMaxSumCount(), Temporality.DELTA);
+   *
+   * //register the view with the MeterSdkProvider
+   * meterProvider.registerView(instrumentSelector, viewSpecification);
+   * </pre>
+   *
+   * @see ViewSpecification
+   */
+  public void registerView(InstrumentSelector selector, ViewSpecification specification) {
+    viewRegistry.registerView(selector, specification);
   }
 
   private static final class MetricProducerSdk implements MetricProducer {
--- a/sdk/src/main/java/io/opentelemetry/sdk/metrics/MeterSdkProvider.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/metrics/MeterSdkProvider.java
@@ -23,6 +23,8 @@ import io.opentelemetry.sdk.internal.ComponentRegistry;
 import io.opentelemetry.sdk.internal.MillisClock;
 import io.opentelemetry.sdk.metrics.data.MetricData;
 import io.opentelemetry.sdk.metrics.export.MetricProducer;
+import io.opentelemetry.sdk.metrics.view.AggregationConfiguration;
+import io.opentelemetry.sdk.metrics.view.InstrumentSelector;
 import io.opentelemetry.sdk.resources.Resource;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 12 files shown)


sdk/src/main/java/io/opentelemetry/sdk/metrics/ViewRegistry.java
```diff
@@ -47,39 +64,71 @@ Batcher createBatcher(
...
-            descriptor, meterProviderSharedState, meterSharedState, aggregation);
-      case VALUE_RECORDER:
-        // TODO: Revisit the batcher used here for value observers,
-        // currently this does not remove duplicate records in the same cycle.
-      case VALUE_OBSERVER:
-        return Batchers.getDeltaAllLabels(
-            descriptor, meterProviderSharedState, meterSharedState, aggregation);
+    Aggregation aggregation = specification.aggregation();
+
+    if (Temporality.CUMULATIVE == specification.temporality()) {
```
anuraaga, line 71: Think we'd usually order `specification.temporality() == ` or use a switch

sdk/src/main/java/io/opentelemetry/sdk/metrics/ViewRegistry.java
```diff
@@ -47,39 +64,71 @@ Batcher createBatcher(
...
+    // If none found, use the defaults:
+    return getDefaultSpecification(descriptor);
+  }
+
+  private static boolean matchesOnType(
+      InstrumentDescriptor descriptor, InstrumentSelector registeredSelector) {
+    if (registeredSelector.instrumentType() == null) {
+      return true;
+    }
+    return Objects.equals(registeredSelector.instrumentType(), descriptor.getType());
```
anuraaga, line 102: Already checked null
```suggestion
    return registeredSelector.instrumentType().equals(descriptor.getType());
```

sdk/src/main/java/io/opentelemetry/sdk/metrics/view/ViewSpecification.java
```diff
@@ -0,0 +1,51 @@
...
+
+@AutoValue
+@Immutable
+public abstract class ViewSpecification {
+
+  public static ViewSpecification create(Aggregation aggregation, Temporality temporality) {
+    return new AutoValue_ViewSpecification(aggregation, temporality);
+  }
+
+  /** Which aggregation should be used for this View. */
```
anuraaga, line 31:
```suggestion
  /** Which {@link Aggregation} should be used for this View. */
```

sdk/src/main/java/io/opentelemetry/sdk/metrics/view/ViewSpecification.java
```diff
@@ -0,0 +1,51 @@
...
+  @Nullable
+  public abstract Aggregation aggregation();
+
+  /** What temporality should be used for this View (delta vs. cumulative). */
```
anuraaga, line 35:
```suggestion
  /** What {@link Temporality} should be used for this View (delta vs. cumulative). */
```

sdk/src/main/java/io/opentelemetry/sdk/metrics/view/ViewSpecification.java
```diff
@@ -0,0 +1,51 @@
...
+  @Nullable
+  public abstract Temporality temporality();
+
+  /** An enumeration which describes the time period over which metrics should be aggregated. */
+  public enum Temporality {
```
anuraaga, line 40: Not sure but it seems like this could be applicable as a top level enum (outside the realm of views)

sdk/src/main/java/io/opentelemetry/sdk/metrics/view/ViewSpecification.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * DELTA means that metrics will be aggregated only over the most recent collection interval.
```
anuraaga, line 42:
```suggestion
     * Metrics will be aggregated only over the most recent collection interval.
```

Current style feels like Go but not idiomatic Java to me

sdk/src/main/java/io/opentelemetry/sdk/metrics/view/ViewSpecification.java
```diff
@@ -0,0 +1,51 @@
...
+     */
+    DELTA,
+    /**
+     * CUMULATIVE means that metrics will be aggregated over the lifetime of the associated {@link
```
anuraaga, line 46:
```suggestion
     * Metrics will be aggregated over the lifetime of the associated {@link
```

sdk/src/main/java/io/opentelemetry/sdk/metrics/ViewRegistry.java
jkwatson: I've seen it both ways...I don't think there's much of a standard, or even agreement on this. Back in the day, I'd use `.equals()` even for enums, since someone could change it from an enum and break code if you use `==`.  :)

sdk/src/main/java/io/opentelemetry/sdk/metrics/view/ViewSpecification.java
jkwatson: possibly, yes. I think this will become a top-level concept eventually, so if/when that happens, we'll definitely make it more first-class.

sdk/src/main/java/io/opentelemetry/sdk/metrics/MeterSdkProvider.java
```diff
@@ -42,11 +44,12 @@
 
   private final MeterSdkComponentRegistry registry;
   private final MetricProducer metricProducer;
+  private final ViewRegistry viewRegistry = new ViewRegistry();
 
   private MeterSdkProvider(Clock clock, Resource resource) {
     this.registry =
         new MeterSdkComponentRegistry(
-            MeterProviderSharedState.create(clock, resource), new ViewRegistry());
+            MeterProviderSharedState.create(clock, resource), viewRegistry);
```
mxiamxia, line 52: Wondering why we don't want ```ViewRegistry``` per ```MeterSdk``` instance level? What if I have Metric instruments with the same type and name but I want it have different custom aggregation & temporality on different ```instrumentationName```?

jkwatson: The SDK user won't in general know what named Meters are available, so I don't know how you would access the specific Meter in order to have a registry at that level. How were you thinking that an operator would know how to use this if the API were available only on the individual Meters?

sdk/src/main/java/io/opentelemetry/sdk/metrics/ViewRegistry.java
```diff
@@ -47,39 +63,71 @@ Batcher createBatcher(
...
-    // todo look up based on fields of the descriptor.
+  // todo: consider moving this method to its own class, for more targetted testing.
+  private AggregationConfiguration findBestMatch(InstrumentDescriptor descriptor) {
+
+    for (Map.Entry<InstrumentSelector, AggregationConfiguration> entry : configuration.entrySet()) {
+      InstrumentSelector registeredSelector = entry.getKey();
+
+      if (matchesOnName(descriptor, registeredSelector)
+          && matchesOnType(descriptor, registeredSelector)) {
+        return entry.getValue();
```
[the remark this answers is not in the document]
jkwatson: sure, happy to make that change. 👍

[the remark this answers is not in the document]
jkwatson: done.

tylerbenson:
@jkwatson what needs to happen to make progress here?  I feel this is a blocker for my work in adding an additional aggregation option.

jkwatson:
> @jkwatson what needs to happen to make progress here? I feel this is a blocker for my work in adding an additional aggregation option.

Well, aside from a rebase... we'd have to decide this was something we wanted to implement, even though it's not really spec'd at this point.

jkwatson:
closing in favor of #2037 (rebasing this proved messy and I no longer work for NR, so didn't want to use their fork as the basis for the PR)

PR #1419, opened 2020-07-15
Add customizable global error handler

MitchellDumovic:
This is a POC solution to https://github.com/open-telemetry/opentelemetry-specification/issues/696 

Much of the inspiration for this PR came from the respective implementation in Go [here](https://github.com/open-telemetry/opentelemetry-go/blob/master/api/global/handler.go#L27)

Adds an error handler delegate to `OpenTelemetry`, which can handle errors through a call to `OpenTelemetry.handleError`. 

The delegate accepts errors in the form of exceptions of type `OpenTelemetryException`, which most often are constructed at the time of an error by wrapping an exception with some additional information about where it occurred. By default, the delegate logs a warning using the native Java logger. A custom delegate can be registered by providing a `ErrorHandlerFactory` class, in a similar way that a custom `TraceProvider` can be registered. 

This PR also updates several locations (primarily in various exporters) to handle an error using the global handler. Some locations I have added net-new reporting, and others I have just replaced the `log.warn` message that previously existed.


jkwatson:
I don't think this belongs in the API. The API is reserved for instrumentation concerns. To me, this is very clearly an operator concern, and should only be a part of the SDK.

carlosalberto:
I see the point on why @jkwatson doesn't want it here - in the end, we probably don't event want to have errors reported when the API *alone* is used. The fact we are consuming such error handler from SDK-level components seem to add weight to that argument.

Personally I'd be fine with adding this very same concept to either `OpenTelemetrySdk` or a similar global component at the SDK level. Would that work for you @jkwatson?

Also, I'm wondering whether we need the factory for the error. I imagine users setting such handler through code directly (when configuring the SDK itself, such as the java-instrumentation code would do), but I'm not totally sure we need a factory here.

jkwatson:
I'm very happy to have an error handling mechanism baked into the SDK, definitely.

MitchellDumovic:
I agree that it makes sense for this to be in the SDK - I've updated this PR to do so, and also took @carlosalberto's suggestion to use a code setter instead of a factory.

sdk/src/main/java/io/opentelemetry/sdk/errorhandler/DefaultErrorHandler.java
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright 2019, OpenTelemetry Authors
```
thisthat, line 2: Nit: 2020

sdk/src/main/java/io/opentelemetry/sdk/errorhandler/ErrorHandler.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright 2019, OpenTelemetry Authors
```
thisthat, line 2: also here the year

sdk/src/main/java/io/opentelemetry/sdk/errorhandler/OpenTelemetryException.java
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright 2019, OpenTelemetry Authors
```
thisthat, line 2: and here

sdk/src/main/java/io/opentelemetry/sdk/errorhandler/DefaultErrorHandler.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * Default implementation of {@link ErrorHandler}. By default, logs the full string of the error
+ * with a level of "Warning".
+ *
+ * @since 0.1.0
```
thisthat, line 27:
```suggestion
 * @since 0.7.0
```

sdk/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
```diff
@@ -18,22 +18,33 @@
...
+ * <p>This class also provides static methods for accessing and setting a custom global error
+ * handler for the OpenTelemetry SDK.
  *
  * @see OpenTelemetry
  */
 @ThreadSafe
 public final class OpenTelemetrySdk {
+  private static final Object mutex = new Object();
+
+  @Nullable private static volatile OpenTelemetrySdk instance;
```
jkwatson, line 45: rather than introducing a whole instance and double-checked locking in the getInstance, why not just have the errorHandler be static, since it's global anyway?

sdk/src/main/java/io/opentelemetry/sdk/errorhandler/DefaultErrorHandler.java
```diff
@@ -0,0 +1,51 @@
...
+   *
+   * @return a {@code ErrorHandler} singleton that is the default implementation for {@link
+   *     ErrorHandler}.
+   */
+  public static ErrorHandler getInstance() {
+    return instance;
+  }
+
+  @Override
+  public void handle(OpenTelemetryException e) {
```
anuraaga, line 46: Just curious do we have a story about controlling the default behavior based on the type of exception? A level flag to `OpenTelemetryException?

The context is I am looking at null / arg checks which we should remove to abide by the error handling specs, that the SDK should not throw errors due to incorrect usage by a user. I think for these logging by default would be problematic since if a simple method like `.setAttribute` was logging it could easily be enough to bring down the app not so unlike an actual exception. But naturally for less common / sporadic errors like export error it would generally be ok to log.

jkwatson: Since this is basically an API for error logging, I'm starting to wonder if adding this additional error handler is overkill. Couldn't users just specify a custom logger if they want special behavior, outside the normal java.util.logging behavior?

That is to say, if we stipulate the all SDK logging will be in the `io.opentelemetry.sdk` namespace, can't we just use normal logging facilities to handle the global error handler use-case?

carlosalberto: > The context is I am looking at null / arg checks which we should remove to abide by the error handling specs, that the SDK should not throw errors due to incorrect usage by a user.

I personally think we shouldn't log wrong parameters, if that's what you are asking. Then again, other implementations log such errors (the Python one does, for example). But we should probably discuss that in the Specification call tomorrow.

carlosalberto: @jkwatson What if the user wants to do something with the error, other than logging?

sdk/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
carlosalberto: +1 for the handler being static (we don't need anything fancy here, at least *for now*).

sdk/src/main/java/io/opentelemetry/sdk/errorhandler/DefaultErrorHandler.java
jkwatson: Well, the logging `Handler` implementation can do whatever it wants, right? It's not required to do any actual logging, is it?

carlosalberto: I think I'm not following ;)

> That is to say, if we stipulate the all SDK logging will be in the io.opentelemetry.sdk namespace, can't we just use normal logging facilities to handle the global error handler use-case?

You mean entirely removing the concept of an error handler, and have user entirely rely for this purpose on `java.util.logging` services?

jkwatson: exactly. @anuraaga 's comment about severity got me thinking about this. My first thought was...why not just add a severity to the global error handler. But, if we're just logging severity, message, throwable, that looks pretty much like every logging API in existence.... so why invent a new one for this purpose?

carlosalberto: And what *if* the user wants to do something other than simple logging? Probably we should look for cases where simple logging is not the principal use case (to go one way or another).

jkwatson: by 'user', do you mean the SDK author generating the "errors" or, the user of the SDK who is doing something with them? A java.util.logging.Handler, I think, can do anything it wants...it doesn't have to be "simple logging".

carlosalberto: So I have mixed feelings about using the Java's logging facility. In theory having a handler is a straightforward, super simple thing to do, albeit limited, while the logging path includes dealing with an entire [LogRecord](https://docs.oracle.com/javase/7/docs/api/java/util/logging/LogRecord.html) object (and what to expect to handle there? Throwable,  message? parameters?), where the user has to override *three* abstract methods for `Handler`.

jkwatson: Yeah, I'm not convinced it was the right way to go, but rather than re-invent an internal logging API, I'd rather use something off the shelf. I can also see arguments for not wanting to generate an entire stack trace just to report some sort of SDK behavior...in which case we're getting dangerously close to implementing a logging API.  :)

MitchellDumovic: Jumping in with my two-cents here… 

Taking a step back,  I think the the high-level goal here is that we want it to be easy for users of the SDK to get visibility into any errors that occur during setup or at runtime that prevent the expected tracing or metrics information from being exported. 

I think that three things that are very important for achieving this high-level goal are: 

(1) Making it easy for the SDK authors to report errors in a standardized way. 
(2) Making it easy for the SDK user (not the SDK author) to have control over where errors get logged and which types of errors get logged 
(3) Having a sane default so that important errors do not get squashed when someone is trying to set up OpenTelemetry. 

I think that both the solution outlined in this PR and using Java’s built-in logging facility technically achieve all of these goals, but I do agree with @carlosalberto that Java’s built-in logging facility potentially adds friction that makes the whole process clunkier to work with. A user has to implement three abstract methods, it’s harder to transmit information about where an error came from and switch on that error, and it’s potentially less obvious to the SDK author that calling `logger.warning` will be correctly handled and propagated to the right place. 

Having said that, I do agree with the notion that we shouldn’t be trying to implement our own logging API from scratch. If we feel like we need a lot of the bells and whistles that the Java logging library provides for our use cases, then I think that we should just go ahead and use that. If we only need a small subset of them, then I think that the overhead of building and maintaining them is worth the increased clarity and ease-of-use that this solution brings.

anuraaga:
I agree that we should have great defaults so users shouldn't have to do anything in general. Some users may want control but if the defaults are good enough, maybe not.

Putting on the user hat, if I had the ability to, I would want

Unexpected, recoverable / transient errors logged as warning
Unrecoverable exceptions like validation issues logged at debug, or logged once to warn
Validation issues thrown as exceptions in unit tests.

If otel provided a flag that I could enable in tests for the extra behavior there, then I'd just use it. If there was a error handler, I'd implement my own flag using it. If it was a jul handler, it would take me some more time to understand how to use jul, but I assume it would work too. None of these options feel bad to me. And if there's no way to have a flag for unit tests, I don't think I'd feel significantly bothered though there probably will be points where I wonder why a tag is missing and the debug to find a null value takes longer than it could have.

carlosalberto:
Interesting points @bogdandrutu 

To me, it feels like an (absolutely beautiful) utopia to try to monitor OpenTelemetry with itself (If I got it right from your comment) - what will you do if you are stuck exporting data, for example? Where can you report such errors as traces/metrics? Also, there's the case of local errors that could easily be diagnosed, without need to do fancy work.

As the description of this PR mentions, Go already such error handler. Would you be against such approach? Asking also as the idea is to have something like this in the Specification itself ;)

Finally, as we *already* do logging for many errors, what do you think about going for @jkwatson's idea about 'formalizing' the usage of Java logging facility to make this possible? Would you be fine with that?

carlosalberto:
Ping @bogdandrutu ;)

MitchellDumovic:
After discussion at the spec SIG this week and in the Otel spec [issue](https://github.com/open-telemetry/opentelemetry-specification/issues/696), it looks like the best approach here is to just leverage the `java.util.logging` library directly to get the customizable handler here that we want. 

From what I understand, we actually don't need to make any code changes for this (besides actually adding additional log messages where they are missing now). However, I'd like to add documentation with examples of how the logger can be customized for those unfamiliar with `java.util.logging`, which I'd like to do in a separate PR as I think this has diverged sufficiently from this PR's original implementation.

With this, OpenTelemetry authors can continue to use `logger.log` messages to log errors, warnings, or info, and it is in the user's hands to accept or filter out any errors output. 

Does this all sound like a reasonable way to move forward? @carlosalberto @bogdandrutu @anuraaga @jkwatson

MitchellDumovic:
I took a stab at adding some documentation for the java logger [here](https://github.com/open-telemetry/opentelemetry-java/pull/1501) - comments are appreciated!

jkwatson:
With #1501 merged, can we close this for now? If we run into a case that logging really can't handle, we can re-open this idea.

PR #1421, opened 2020-07-15
Logging sink api

zenmoto:
This is an attempt at an initial log delivery API. It's not targeted as a replacement for a user-level logging API, instead it's focused on taking log entries and passing them on to an SDK for processing and delivery. The model is based off of open-telemetry/opentelemetry-proto#151.

### DIFF
--- a/api/src/main/java/io/opentelemetry/common/AnyValue.java
+++ b/api/src/main/java/io/opentelemetry/common/AnyValue.java
@@ -1,0 +1,291 @@
+/*
+ * Copyright 2020, OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.common;
+
+import com.google.auto.value.AutoValue;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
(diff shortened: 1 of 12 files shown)


api/src/main/java/io/opentelemetry/common/AnyValue.java
```diff
@@ -0,0 +1,291 @@
...
+/**
+ * A class that represents all the possible values for a data body. An {@code AnyValue} can have 6
+ * types of values: {@code String}, {@code boolean}, {@code int}, {@code double}, {@code array}, or
+ * {@code kvlist}. represented through {@code AnyValue.Type}. A {@code array} or a {@code kvlist}
+ * can in turn hold other {@code AnyValue} instances, allowing for mapping to JSON-like structures.
+ *
+ * @since 0.7.0
+ */
+@Immutable
+public abstract class AnyValue {
```
jkwatson, line 34: I think we have a proto version of this class already. is there a reason we can't use it?

jkwatson:
If this isn't meant to be a user-level logging API, why have it accessible as a user-level instrumentation API? Users will definitely be confused if it's there, but they aren't supposed to use it.

api/src/main/java/io/opentelemetry/common/AnyValue.java
zenmoto: We definitely have it in opentelemetry-proto, but I believe that api does not have a dependency on proto. I could be wrong- I'm trying to mirror how we are doing it with `AttributeValue`.

zenmoto:
It's not that it *can't* be used as a logging API, but it's not going to be ergonomic. This is to enable adapters for existing logging libraries to output through OpenTelemetry. It would also be the foundation for an ergonomic logging library, but I think that's going to be easier to work through when we have functional plumbing.

jkwatson:
I don't think we should add a logging API until it has made it into the official specifications. Especially before GA. I believe the agreement is that we will not GA with a logging API. Please point me somewhere else if I am wrong about that.

bogdandrutu:
@zenmoto to make progress, would strongly suggest to start in one of the extensions directories.

zenmoto:
Thank you @bogdandrutu - I'll close this and take that approach instead.

Issue #1422, opened 2020-07-16
Should exports be async


huntc:
I’m writing an export class for Reactive streams, as I did for [OpenTracing and Dropwizard Metrics](https://github.com/titanclass/reactive-streams-telemetry). 

It occurs me that the export method should be returning a CompletionStage of a result instead of just a result. An export implementation is often performing some long lived action such as communicating over the network. I note that the Jaeger export function actually blocks. 

What’s the appetite for making the export method non-blocking?

jkwatson:
Hi @huntc . I think our general approach is to leave this up to the exporter implementation. For example, the New Relic exporter returns immediately and handles all the processing on a background thread. If you want an async exporter, then my recommendation would be to make yours do the work in the background.

The jaeger exporter should probably be changed to do the same, I'd guess.

huntc:
Thanks, @jkwatson. What's the effect of the returned status indicating failure? 

If knowing whether the export is successful is important then I'd suggest that the method declaration should reflect async possibilities for implementations.

If it isn't important to know if the export is successful then the return type should be Void. 

WDYT?

jkwatson:
We've had this discussion in the specs SIG. At the moment, there is no effect of returning the failure status. We made the method return an enum value in case in the future we want to support a robust retry-system in the SDK. We didn't want to potentially break existing exporters if we did decide to enhance the capabilities of the SDK in this way. Does that make sense?

huntc:
Thanks, John.

Returning a status that yields no current benefit and is generally only attainable by blocking does not make sense to me.

If we may want to have a robust retry system in the future then I'd suggest that we need to use a completion stage now... blocking the thread isn't a great way to go and we know that exporters perform long-lived actions.

That said, returning a `Void` perhaps reflects the correct semantics. I agree that letting the exporters deal with retries in their own way makes sense. Different types of connectivity demand different retry policies e.g. if I'm using an exporter from a node known to have poor internet connectivity and low power requirements, sending on a best-effort basis is fine. If I'm using an exporter within a data center then retrying may make more sense. Then there are circuit-breaker considerations that apply to some environments, but not others...

Now feels the time to introduce API breaking changes before going GM.

Please feel free to point me at another forum if there's a wider conversation to be had. I would have thought it'd be language-specific though given the existence of coroutines etc. in other languages.

jkwatson:
[...]

Issue #1418, opened 2020-07-15
More logging on exporter failures


iNikem:
**Is your feature request related to a problem? Please describe.**
A customer has tried to Otel java instrumentation agent. He has difficulties understanding if he has configured exporters correctly or not. Apparently only Zipkin exporter logs some warnings if data sending fails. All others are silent.

**Describe the solution you'd like**
Provide more feedback if any component fails, e.g. by writing more failure related information to logs.

**Describe alternatives you've considered**
As an alternative make sure that such info is written at least on debug level.

jkwatson:
Looks like the following exporters need at least some amount of error logging:

- [ ] Jaeger
- [ ] OLTP Metrics
- [ ] OLTP Spans

Any others?

iNikem:
Closed by #1480

PR #1428, opened 2020-07-17
Makes Export methods async

huntc:
This PR recognises that the export and flush methods of span and trace exporters can be, and often are, implemented with long-lived operations over networks. We, therefore, see that these method return types are represented using a `CompletableResultCode` type that can be completed asynchronously for either failure or success scenarios. In either instance, the export code can now perform actions on failure and success. In the case of failure, which can also constitute a cancellation by an upstream component (such as a batch span processor), an exporter can now interrupt its threads or perform any other action it needs to. This approach of handling failure is also safer than what now exists, which is for an upstream component to interrupt a thread.

The PR approach is focused on changing the signatures within the export methods now, while we can still introduce API-breaking changes. The implementation of the existing exporters mostly assumes a blocking interaction as before. The remaining internal exporter behaviour can be modified in the future if required and without affecting the API. One exception to this is the `OtlpGrpcMetricExporter` where @anuraaga kindly updated it to become async.

For further background information on this change, see #https://github.com/open-telemetry/opentelemetry-java/issues/1422, which also refers to spec clarification via https://github.com/open-telemetry/opentelemetry-specification/pull/707.

### DIFF
--- a/exporters/inmemory/src/main/java/io/opentelemetry/exporters/inmemory/InMemoryMetricExporter.java
+++ b/exporters/inmemory/src/main/java/io/opentelemetry/exporters/inmemory/InMemoryMetricExporter.java
@@ -16,6 +16,7 @@
 
 package io.opentelemetry.exporters.inmemory;
 
+import io.opentelemetry.sdk.common.export.CompletableResultCode;
 import io.opentelemetry.sdk.metrics.data.MetricData;
 import io.opentelemetry.sdk.metrics.export.MetricExporter;
 import java.util.ArrayList;
@@ -103,12 +104,12 @@ public class InMemoryMetricExporter implements MetricExporter {
    * <p>If this is called after {@code shutdown}, this will return {@code ResultCode.FAILURE}.
    */
   @Override
-  public ResultCode export(Collection<MetricData> metrics) {
+  public CompletableResultCode export(Collection<MetricData> metrics) {
     if (isStopped) {
-      return ResultCode.FAILURE;
+      return CompletableResultCode.ofFailure();
     }
     finishedMetricItems.addAll(metrics);
-    return ResultCode.SUCCESS;
+    return CompletableResultCode.ofSuccess();
   }
 
   /**
@@ -118,8 +119,8 @@ public class InMemoryMetricExporter implements MetricExporter {
    * @return always Success
    */
   @Override
-  public ResultCode flush() {
-    return ResultCode.SUCCESS;
+  public CompletableResultCode flush() {
+    return CompletableResultCode.ofSuccess();
   }
 
   /**
--- a/exporters/inmemory/src/test/java/io/opentelemetry/exporters/inmemory/InMemoryMetricExporterTest.java
+++ b/exporters/inmemory/src/test/java/io/opentelemetry/exporters/inmemory/InMemoryMetricExporterTest.java
@@ -23,7 +23,6 @@ import io.opentelemetry.sdk.common.InstrumentationLibraryInfo;
 import io.opentelemetry.sdk.metrics.data.MetricData;
 import io.opentelemetry.sdk.metrics.data.MetricData.Descriptor;
 import io.opentelemetry.sdk.metrics.data.MetricData.LongPoint;
-import io.opentelemetry.sdk.metrics.export.MetricExporter.ResultCode;
 import io.opentelemetry.sdk.resources.Resource;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -54,7 +53,7 @@ class InMemoryMetricExporterTest {
     metrics.add(generateFakeMetric());
     metrics.add(generateFakeMetric());
 
-    assertThat(exporter.export(metrics)).isEqualTo(ResultCode.SUCCESS);
+    assertThat(exporter.export(metrics).isSuccess()).isTrue();
     List<MetricData> metricItems = exporter.getFinishedMetricItems();
     assertThat(metricItems).isNotNull();
     assertThat(metricItems.size()).isEqualTo(3);
@@ -67,7 +66,7 @@ class InMemoryMetricExporterTest {
     metrics.add(generateFakeMetric());
     metrics.add(generateFakeMetric());
 
-    assertThat(exporter.export(metrics)).isEqualTo(ResultCode.SUCCESS);
+    assertThat(exporter.export(metrics).isSuccess()).isTrue();
     List<MetricData> metricItems = exporter.getFinishedMetricItems();
     assertThat(metricItems).isNotNull();
     assertThat(metricItems.size()).isEqualTo(3);
@@ -84,7 +83,7 @@ class InMemoryMetricExporterTest {
     metrics.add(generateFakeMetric());
     metrics.add(generateFakeMetric());
 
-    assertThat(exporter.export(metrics)).isEqualTo(ResultCode.SUCCESS);
+    assertThat(exporter.export(metrics).isSuccess()).isTrue();
     exporter.shutdown();
     List<MetricData> metricItems = exporter.getFinishedMetricItems();
     assertThat(metricItems).isNotNull();
@@ -98,13 +97,13 @@ class InMemoryMetricExporterTest {
     metrics.add(generateFakeMetric());
     metrics.add(generateFakeMetric());
 
-    assertThat(exporter.export(metrics)).isEqualTo(ResultCode.SUCCESS);
+    assertThat(exporter.export(metrics).isSuccess()).isTrue();
     exporter.shutdown();
-    assertThat(exporter.export(metrics)).isEqualTo(ResultCode.FAILURE);
+    assertThat(exporter.export(metrics).isSuccess()).isFalse();
   }
 
   @Test
   void test_flush() {
-    assertThat(exporter.flush()).isEqualTo(ResultCode.SUCCESS);
+    assertThat(exporter.flush().isSuccess()).isTrue();
   }
 }
--- a/exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java
+++ b/exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java
@@ -21,6 +21,7 @@ import io.grpc.ManagedChannelBuilder;
 import io.opentelemetry.exporters.jaeger.proto.api_v2.Collector;
 import io.opentelemetry.exporters.jaeger.proto.api_v2.CollectorServiceGrpc;
 import io.opentelemetry.exporters.jaeger.proto.api_v2.Model;
+import io.opentelemetry.sdk.common.export.CompletableResultCode;
 import io.opentelemetry.sdk.common.export.ConfigBuilder;
 import io.opentelemetry.sdk.trace.data.SpanData;
 import io.opentelemetry.sdk.trace.export.SpanExporter;
@@ -107,7 +108,7 @@ public final class JaegerGrpcSpanExporter implements SpanExporter {
    * @return the result of the operation
    */
   @Override
-  public ResultCode export(Collection<SpanData> spans) {
+  public CompletableResultCode export(Collection<SpanData> spans) {
     Collector.PostSpansRequest request =
         Collector.PostSpansRequest.newBuilder()
             .setBatch(
@@ -126,10 +127,10 @@ public final class JaegerGrpcSpanExporter implements SpanExporter {
       // for now, there's nothing to check in the response object
       //noinspection ResultOfMethodCallIgnored
       stub.postSpans(request);
-      return ResultCode.SUCCESS;
+      return CompletableResultCode.ofSuccess();
     } catch (Throwable e) {
       logger.log(Level.WARNING, "Failed to export spans", e);
-      return ResultCode.FAILURE;
+      return CompletableResultCode.ofFailure();
     }
   }
 
@@ -139,8 +140,8 @@ public final class JaegerGrpcSpanExporter implements SpanExporter {
    * @return always Success
    */
   @Override
-  public ResultCode flush() {
-    return ResultCode.SUCCESS;
+  public CompletableResultCode flush() {
+    return CompletableResultCode.ofSuccess();
   }
 
   /**
--- a/exporters/logging/src/main/java/io/opentelemetry/exporters/logging/LoggingMetricExporter.java
+++ b/exporters/logging/src/main/java/io/opentelemetry/exporters/logging/LoggingMetricExporter.java
@@ -16,6 +16,7 @@
 
 package io.opentelemetry.exporters.logging;
 
+import io.opentelemetry.sdk.common.export.CompletableResultCode;
 import io.opentelemetry.sdk.metrics.data.MetricData;
 import io.opentelemetry.sdk.metrics.export.MetricExporter;
 import java.util.Collection;
@@ -27,12 +28,12 @@ public class LoggingMetricExporter implements MetricExporter {
   private static final Logger logger = Logger.getLogger(LoggingMetricExporter.class.getName());
 
   @Override
-  public ResultCode export(Collection<MetricData> metrics) {
+  public CompletableResultCode export(Collection<MetricData> metrics) {
     logger.info("Received a collection of " + metrics.size() + " metrics for export.");
     for (MetricData metricData : metrics) {
       logger.log(Level.INFO, "metric: {0}", metricData);
     }
-    return ResultCode.SUCCESS;
+    return CompletableResultCode.ofSuccess();
   }
 
   /**
@@ -41,16 +42,16 @@ public class LoggingMetricExporter implements MetricExporter {
    * @return the result of the operation
    */
   @Override
-  public ResultCode flush() {
-    ResultCode resultCode = ResultCode.SUCCESS;
+  public CompletableResultCode flush() {
+    CompletableResultCode resultCode = new CompletableResultCode();
     for (Handler handler : logger.getHandlers()) {
       try {
         handler.flush();
       } catch (Throwable t) {
-        resultCode = ResultCode.FAILURE;
+        resultCode.fail();
       }
     }
-    return resultCode;
+    return resultCode.succeed();
   }
 
   @Override
--- a/exporters/logging/src/main/java/io/opentelemetry/exporters/logging/LoggingMetricExporter.java
+++ b/exporters/logging/src/main/java/io/opentelemetry/exporters/logging/LoggingMetricExporter.java
@@ -48,7 +48,7 @@ public class LoggingMetricExporter implements MetricExporter {
       try {
         handler.flush();
       } catch (Throwable t) {
-        resultCode.fail();
+        return resultCode.fail();
       }
     }
     return resultCode.succeed();
--- a/exporters/logging/src/main/java/io/opentelemetry/exporters/logging/LoggingSpanExporter.java
+++ b/exporters/logging/src/main/java/io/opentelemetry/exporters/logging/LoggingSpanExporter.java
@@ -16,6 +16,7 @@
 
 package io.opentelemetry.exporters.logging;
 
+import io.opentelemetry.sdk.common.export.CompletableResultCode;
 import io.opentelemetry.sdk.trace.data.SpanData;
 import io.opentelemetry.sdk.trace.export.SpanExporter;
 import java.util.Collection;
@@ -28,11 +29,11 @@ public class LoggingSpanExporter implements SpanExporter {
   private static final Logger logger = Logger.getLogger(LoggingSpanExporter.class.getName());
 
   @Override
-  public ResultCode export(Collection<SpanData> spans) {
+  public CompletableResultCode export(Collection<SpanData> spans) {
     for (SpanData span : spans) {
       logger.log(Level.INFO, "span: {0}", span);
     }
-    return ResultCode.SUCCESS;
+    return CompletableResultCode.ofSuccess();
   }
 
   /**
@@ -41,16 +42,16 @@ public class LoggingSpanExporter implements SpanExporter {
    * @return the result of the operation
    */
   @Override
-  public ResultCode flush() {
-    ResultCode resultCode = ResultCode.SUCCESS;
+  public CompletableResultCode flush() {
+    CompletableResultCode resultCode = new CompletableResultCode();
     for (Handler handler : logger.getHandlers()) {
       try {
         handler.flush();
       } catch (Throwable t) {
-        resultCode = ResultCode.FAILURE;
+        resultCode.fail();
       }
     }
-    return resultCode;
+    return resultCode.succeed();
   }
 
   @Override
--- a/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcMetricExporter.java
+++ b/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcMetricExporter.java
@@ -19,12 +19,18 @@ package io.opentelemetry.exporters.otlp;
 import static io.grpc.Metadata.ASCII_STRING_MARSHALLER;
 
 import com.google.common.base.Splitter;
+import com.google.common.util.concurrent.FutureCallback;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.MoreExecutors;
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
 import io.grpc.Metadata;
 import io.grpc.stub.MetadataUtils;
 import io.opentelemetry.proto.collector.metrics.v1.ExportMetricsServiceRequest;
+import io.opentelemetry.proto.collector.metrics.v1.ExportMetricsServiceResponse;
 import io.opentelemetry.proto.collector.metrics.v1.MetricsServiceGrpc;
+import io.opentelemetry.proto.collector.metrics.v1.MetricsServiceGrpc.MetricsServiceFutureStub;
+import io.opentelemetry.sdk.common.export.CompletableResultCode;
 import io.opentelemetry.sdk.common.export.ConfigBuilder;
 import io.opentelemetry.sdk.metrics.data.MetricData;
 import io.opentelemetry.sdk.metrics.export.MetricExporter;
@@ -65,9 +71,8 @@ public final class OtlpGrpcMetricExporter implements MetricExporter {
 
   private static final Logger logger = Logger.getLogger(OtlpGrpcMetricExporter.class.getName());
 
-  private final MetricsServiceGrpc.MetricsServiceBlockingStub blockingStub;
+  private final MetricsServiceFutureStub metricsService;
   private final ManagedChannel managedChannel;
-  private final long deadlineMs;
 
   /**
    * Creates a new OTLP gRPC Metric Reporter with the given name, using the given channel.
@@ -78,8 +83,11 @@ public final class OtlpGrpcMetricExporter implements MetricExporter {
    */
   private OtlpGrpcMetricExporter(ManagedChannel channel, long deadlineMs) {
     this.managedChannel = channel;
-    this.blockingStub = MetricsServiceGrpc.newBlockingStub(channel);
-    this.deadlineMs = deadlineMs;
+    MetricsServiceFutureStub stub = MetricsServiceGrpc.newFutureStub(channel);
+    if (deadlineMs > 0) {
+      stub = stub.withDeadlineAfter(deadlineMs, TimeUnit.MILLISECONDS);
+    }
+    metricsService = stub;
   }
 
   /**
@@ -89,26 +97,29 @@ public final class OtlpGrpcMetricExporter implements MetricExporter {
    * @return the result of the operation
    */
   @Override
-  public ResultCode export(Collection<MetricData> metrics) {
+  public CompletableResultCode export(Collection<MetricData> metrics) {
     ExportMetricsServiceRequest exportMetricsServiceRequest =
         ExportMetricsServiceRequest.newBuilder()
             .addAllResourceMetrics(MetricAdapter.toProtoResourceMetrics(metrics))
             .build();
 
-    try {
-      MetricsServiceGrpc.MetricsServiceBlockingStub stub = this.blockingStub;
-      if (deadlineMs > 0) {
-        stub = stub.withDeadlineAfter(deadlineMs, TimeUnit.MILLISECONDS);
-      }
+    final CompletableResultCode result = new CompletableResultCode();
+    Futures.addCallback(
+        metricsService.export(exportMetricsServiceRequest),
+        new FutureCallback<ExportMetricsServiceResponse>() {
+          @Override
+          public void onSuccess(@Nullable ExportMetricsServiceResponse response) {
+            result.succeed();
+          }
 
-      // for now, there's nothing to check in the response object
-      // noinspection ResultOfMethodCallIgnored
-      stub.export(exportMetricsServiceRequest);
-      return ResultCode.SUCCESS;
-    } catch (Throwable e) {
-      logger.log(Level.WARNING, "Failed to export metrics", e);
-      return ResultCode.FAILURE;
-    }
+          @Override
+          public void onFailure(Throwable t) {
+            logger.log(Level.WARNING, "Failed to export spans", t);
+            result.fail();
+          }
+        },
+        MoreExecutors.directExecutor());
+    return result;
   }
 
   /**
@@ -117,8 +128,8 @@ public final class OtlpGrpcMetricExporter implements MetricExporter {
    * @return always Success
    */
   @Override
-  public ResultCode flush() {
-    return ResultCode.SUCCESS;
+  public CompletableResultCode flush() {
+    return CompletableResultCode.ofSuccess();
   }
 
   /**
--- a/sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright 2019, OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.sdk.common.export;
+
+import java.util.ArrayList;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.GuardedBy;
+
+/**
+ * The implementation of Export operations are often asynchronous in nature, hence the need to
+ * convey a result at a later time. CompletableResultCode facilitates this.
+ *
+ * <p>This class models JDK 8's CompletableFuture to afford migration should Open Telemetry's SDK
+ * select JDK 8 or greater as a baseline, and also to offer familiarity to developers.
+ */
+public class CompletableResultCode {
+  /** A convenience for declaring success. */
+  public static CompletableResultCode ofSuccess() {
+    return SUCCESS;
+  }
+
+  /** A convenience for declaring failure. */
+  public static CompletableResultCode ofFailure() {
+    return FAILURE;
+  }
+
+  private static final CompletableResultCode SUCCESS = new CompletableResultCode().succeed();
+  private static final CompletableResultCode FAILURE = new CompletableResultCode().fail();
+
+  public CompletableResultCode() {}
+
+  @Nullable
+  @GuardedBy("lock")
+  private Boolean succeeded = null;
+
+  @GuardedBy("lock")
+  private final ArrayList<Runnable> completionActions = new ArrayList<>();
+
+  private final Object lock = new Object();
+
+  /** The export operation finished successfully. */
+  public CompletableResultCode succeed() {
+    synchronized (lock) {
+      if (succeeded == null) {
+        succeeded = true;
+        for (Runnable action : completionActions) {
+          action.run();
+        }
+      }
+    }
+    return this;
+  }
+
+  /** The export operation finished with an error. */
+  public CompletableResultCode fail() {
+    synchronized (lock) {
+      if (succeeded == null) {
+        succeeded = false;
+        for (Runnable action : completionActions) {
+          action.run();
+        }
+      }
+    }
+    return this;
+  }
+
+  /**
+   * Obtain the current state of completion. Generally call once completion is achieved via the
+   * thenRun method.
+   *
+   * @return the current state of completion
+   */
+  public boolean isSuccess() {
+    synchronized (lock) {
+      return succeeded != null && succeeded;
+    }
+  }
+
+  /**
+   * Perform an action on completion. Actions are guaranteed to be called only once.
+   *
+   * @param action the action to perform
+   * @return this completable result so that it may be further composed
+   */
+  public CompletableResultCode whenComplete(Runnable action) {
+    synchronized (lock) {
+      if (succeeded != null) {
+        action.run();
+      } else {
+        this.completionActions.add(action);
+      }
+    }
+    return this;
+  }
+}
--- a/sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
@@ -17,23 +17,24 @@
 package io.opentelemetry.sdk.common.export;
 
 import java.util.ArrayList;
+import java.util.List;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.GuardedBy;
 
 /**
- * The implementation of Export operations are often asynchronous in nature, hence the need to
- * convey a result at a later time. CompletableResultCode facilitates this.
- *
- * <p>This class models JDK 8's CompletableFuture to afford migration should Open Telemetry's SDK
+ * This class models JDK 8's CompletableFuture to afford migration should Open Telemetry's SDK
  * select JDK 8 or greater as a baseline, and also to offer familiarity to developers.
+ *
+ * <p>The implementation of Export operations are often asynchronous in nature, hence the need to
+ * convey a result at a later time. CompletableResultCode facilitates this.
  */
 public class CompletableResultCode {
-  /** A convenience for declaring success. */
+  /** Returns a {@link CompletableResultCode} that has been completed successfully. */
   public static CompletableResultCode ofSuccess() {
     return SUCCESS;
   }
 
-  /** A convenience for declaring failure. */
+  /** Returns a {@link CompletableResultCode} that has been completed unsuccessfully. */
   public static CompletableResultCode ofFailure() {
     return FAILURE;
   }
@@ -48,11 +49,11 @@ public class CompletableResultCode {
   private Boolean succeeded = null;
 
   @GuardedBy("lock")
-  private final ArrayList<Runnable> completionActions = new ArrayList<>();
+  private final List<Runnable> completionActions = new ArrayList<>();
 
   private final Object lock = new Object();
 
-  /** The export operation finished successfully. */
+  /** Complete this {@link CompletableResultCode} successfully if it is not already completed. */
   public CompletableResultCode succeed() {
     synchronized (lock) {
       if (succeeded == null) {
@@ -65,7 +66,7 @@ public class CompletableResultCode {
     return this;
   }
 
-  /** The export operation finished with an error. */
+  /** Complete this {@link CompletableResultCode} unsuccessfully if it is not already completed. */
   public CompletableResultCode fail() {
     synchronized (lock) {
       if (succeeded == null) {
--- a/sdk/src/main/java/io/opentelemetry/sdk/metrics/export/MetricExporter.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/metrics/export/MetricExporter.java
@@ -16,6 +16,7 @@
 
 package io.opentelemetry.sdk.metrics.export;
 
+import io.opentelemetry.sdk.common.export.CompletableResultCode;
 import io.opentelemetry.sdk.metrics.data.MetricData;
 import java.util.Collection;
 
@@ -30,35 +31,23 @@ import java.util.Collection;
 public interface MetricExporter {
 
   /**
-   * The possible results for the export method.
-   *
-   * @since 0.1.0
-   */
-  // TODO: extract this enum and unify it with SpanExporter.ResultCode
-  enum ResultCode {
-    /** The export operation finished successfully. */
-    SUCCESS,
-
-    /** The export operation finished with an error. */
-    FAILURE
-  }
-
-  /**
-   * Exports the collection of given {@link MetricData}.
+   * Exports the collection of given {@link MetricData}. Note that export operations can be
+   * performed simultaneously depending on the type of metric reader being used. However, the {@link
+   * IntervalMetricReader} will ensure that only one export can occur at a time.
    *
    * @param metrics the collection of {@link MetricData} to be exported.
-   * @return the result of the export.
-   * @since 0.1.0
+   * @return the result of the export, which is often an asynchronous operation.
    */
-  ResultCode export(Collection<MetricData> metrics);
+  CompletableResultCode export(Collection<MetricData> metrics);
 
   /**
-   * Exports the collection of {@link MetricData} that have not yet been exported.
+   * Exports the collection of {@link MetricData} that have not yet been exported. Note that flush
+   * operations can be performed simultaneously depending on the type of metric reader being used.
+   * However, the {@link IntervalMetricReader} will ensure that only one export can occur at a time.
    *
-   * @return the result of the flush.
-   * @since 0.4.0
+   * @return the result of the flush, which is often an asynchronous operation.
    */
-  ResultCode flush();
+  CompletableResultCode flush();
 
   /** Called when the associated IntervalMetricReader is shutdown. */
   void shutdown();
--- a/sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
@@ -24,6 +24,7 @@ import io.opentelemetry.metrics.LongCounter;
 import io.opentelemetry.metrics.LongCounter.BoundLongCounter;
 import io.opentelemetry.metrics.Meter;
 import io.opentelemetry.sdk.common.DaemonThreadFactory;
+import io.opentelemetry.sdk.common.export.CompletableResultCode;
 import io.opentelemetry.sdk.common.export.ConfigBuilder;
 import io.opentelemetry.sdk.trace.ReadableSpan;
 import io.opentelemetry.sdk.trace.SpanProcessor;
@@ -32,12 +33,9 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
+import java.util.Timer;
+import java.util.TimerTask;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.concurrent.GuardedBy;
@@ -49,7 +47,8 @@ import javax.annotation.concurrent.GuardedBy;
  * <p>All spans reported by the SDK implementation are first added to a synchronized queue (with a
  * {@code maxQueueSize} maximum size, after the size is reached spans are dropped) and exported
  * every {@code scheduleDelayMillis} to the exporter pipeline in batches of {@code
- * maxExportBatchSize}.
+ * maxExportBatchSize}. Spans may also be dropped when it becomes time to export again, and there is
+ * an export in progress.
  *
  * <p>If the queue gets half full a preemptive notification is sent to the worker thread that
  * exports the spans to wake up and start a new export cycle.
@@ -85,8 +84,8 @@ public final class BatchSpanProcessor implements SpanProcessor {
 
   private static final String WORKER_THREAD_NAME =
       BatchSpanProcessor.class.getSimpleName() + "_WorkerThread";
-  private static final String EXPORTER_THREAD_NAME =
-      BatchSpanProcessor.class.getSimpleName() + "_ExporterThread";
+  private static final String TIMER_THREAD_NAME =
+      BatchSpanProcessor.class.getSimpleName() + "_TimerThread";
   private final Worker worker;
   private final Thread workerThread;
   private final boolean sampled;
@@ -165,8 +164,7 @@ public final class BatchSpanProcessor implements SpanProcessor {
 
     private static final BoundLongCounter droppedSpans;
 
-    private final ExecutorService executorService =
-        Executors.newSingleThreadExecutor(new DaemonThreadFactory(EXPORTER_THREAD_NAME));
+    private final Timer timer = new Timer(TIMER_THREAD_NAME);
 
     private static final Logger logger = Logger.getLogger(Worker.class.getName());
     private final SpanExporter spanExporter;
@@ -176,6 +174,7 @@ public final class BatchSpanProcessor implements SpanProcessor {
     private final int halfMaxQueueSize;
     private final Object monitor = new Object();
     private final int exporterTimeoutMillis;
+    private final AtomicBoolean exportAvailable = new AtomicBoolean(true);
 
     @GuardedBy("monitor")
     private final List<ReadableSpan> spansList;
@@ -191,8 +190,8 @@ public final class BatchSpanProcessor implements SpanProcessor {
       this.maxQueueSize = maxQueueSize;
       this.halfMaxQueueSize = maxQueueSize >> 1;
       this.maxExportBatchSize = maxExportBatchSize;
-      this.spansList = new ArrayList<>(maxQueueSize);
       this.exporterTimeoutMillis = exporterTimeoutMillis;
+      this.spansList = new ArrayList<>(maxQueueSize);
     }
 
     private void addSpan(ReadableSpan span) {
@@ -242,7 +241,7 @@ public final class BatchSpanProcessor implements SpanProcessor {
 
     private void shutdown() {
       forceFlush();
-      executorService.shutdown();
+      timer.cancel();
       spanExporter.shutdown();
     }
 
@@ -277,28 +276,34 @@ public final class BatchSpanProcessor implements SpanProcessor {
     }
 
     // Exports the list of SpanData to the SpanExporter.
+    @SuppressWarnings("BooleanParameter")
     private void onBatchExport(final List<SpanData> spans) {
-      Future<?> submission =
-          executorService.submit(
+      if (exportAvailable.compareAndSet(true, false)) {
+        try {
+          final CompletableResultCode result = spanExporter.export(spans);
+          result.whenComplete(
               new Runnable() {
                 @Override
                 public void run() {
-                  // In case of any exception thrown by the service handlers catch and log.
-                  try {
-                    spanExporter.export(spans);
-                  } catch (Throwable t) {
-                    logger.log(Level.WARNING, "Exception thrown by the export.", t);
+                  if (!result.isSuccess()) {
+                    logger.log(Level.FINE, "Exporter failed");
                   }
+                  exportAvailable.set(true);
                 }
               });
-      try {
-        // wait at most for the configured timeout.
-        submission.get(exporterTimeoutMillis, TimeUnit.MILLISECONDS);
-      } catch (InterruptedException | ExecutionException e) {
-        logger.log(Level.WARNING, "Exception thrown by the export.", e);
-      } catch (TimeoutException e) {
-        logger.log(Level.WARNING, "Export timed out. Cancelling execution.", e);
-        submission.cancel(true);
+          timer.schedule(
+              new TimerTask() {
+                @Override
+                public void run() {
+                  result.fail();
+                }
+              },
+              exporterTimeoutMillis);
+        } catch (Exception e) {
+          logger.log(Level.WARNING, "Exporter threw an Exception", e);
+        }
+      } else {
+        logger.log(Level.FINE, "Exporter busy. Dropping spans.");
       }
     }
   }
--- a/sdk/src/main/java/io/opentelemetry/sdk/trace/export/MultiSpanExporter.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/trace/export/MultiSpanExporter.java
@@ -16,12 +16,11 @@
 
 package io.opentelemetry.sdk.trace.export;
 
-import static io.opentelemetry.sdk.trace.export.SpanExporter.ResultCode.FAILURE;
-import static io.opentelemetry.sdk.trace.export.SpanExporter.ResultCode.SUCCESS;
-
+import io.opentelemetry.sdk.common.export.CompletableResultCode;
 import io.opentelemetry.sdk.trace.data.SpanData;
 import java.util.Collection;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -47,18 +46,20 @@ public final class MultiSpanExporter implements SpanExporter {
   }
 
   @Override
-  public ResultCode export(Collection<SpanData> spans) {
-    ResultCode currentResultCode = SUCCESS;
+  public CompletableResultCode export(Collection<SpanData> spans) {
+    final CompletableResultCode compositeResultCode = new CompletableResultCode();
+    final AtomicInteger completionsToProcess = new AtomicInteger(spanExporters.length);
     for (SpanExporter spanExporter : spanExporters) {
       try {
-        currentResultCode = mergeResultCode(currentResultCode, spanExporter.export(spans));
-      } catch (Throwable t) {
+        final CompletableResultCode singleResult = spanExporter.export(spans);
+        mergeResultCode(compositeResultCode, singleResult, completionsToProcess);
+      } catch (Exception e) {
         // If an exception was thrown by the exporter
-        logger.log(Level.WARNING, "Exception thrown by the export.", t);
-        currentResultCode = FAILURE;
+        logger.log(Level.WARNING, "Exception thrown by the export.", e);
+        compositeResultCode.fail();
       }
     }
-    return currentResultCode;
+    return compositeResultCode;
   }
 
   /**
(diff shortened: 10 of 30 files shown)


exporters/inmemory/build.gradle
```diff
@@ -11,6 +11,8 @@ ext.moduleName = "io.opentelemetry.exporters.inmemory"
 dependencies {
     api project(':opentelemetry-sdk')
 
+    implementation libraries.guava
+
```
huntc, line 15: This felt like an ok dependency to introduce on the exporters given its usage elsewhere. While we don't use Guava in the public signatures of anything, we do use it to build completed futures internally.

exporters/inmemory/src/main/java/io/opentelemetry/exporters/inmemory/InMemoryMetricExporter.java
```diff
@@ -103,12 +105,12 @@ public void reset() {
    * <p>If this is called after {@code shutdown}, this will return {@code ResultCode.FAILURE}.
    */
   @Override
-  public ResultCode export(Collection<MetricData> metrics) {
+  public Future<ResultCode> export(Collection<MetricData> metrics) {
```
anuraaga, line 108: I don't think `Future` is very useful in modern Java, since we target Java 7, we should probably have a callback parameter to the method instead.

exporters/inmemory/src/test/java/io/opentelemetry/exporters/inmemory/InMemoryMetricExporterTest.java
```diff
@@ -51,26 +52,26 @@ private static MetricData generateFakeMetric() {
...
     metrics.add(generateFakeMetric());
     metrics.add(generateFakeMetric());
     metrics.add(generateFakeMetric());
 
-    assertThat(exporter.export(metrics)).isEqualTo(ResultCode.SUCCESS);
+    assertThat(exporter.export(metrics).get()).isEqualTo(ResultCode.SUCCESS);
     List<MetricData> metricItems = exporter.getFinishedMetricItems();
     assertThat(metricItems).isNotNull();
     assertThat(metricItems.size()).isEqualTo(3);
   }
```
huntc, line 65: If an exception is thrown with the export or flush methods, it is probably unrecoverable, hence allowing the exception to propagate. These methods should return a failure status for recoverable failures.

exporters/logging/src/main/java/io/opentelemetry/exporters/logging/LoggingMetricExporter.java
```diff
@@ -50,12 +53,16 @@ public ResultCode flush() {
...
 
   @Override
   public void shutdown() {
     // no-op
-    this.flush();
+    try {
+      this.flush().get();
+    } catch (InterruptedException | ExecutionException e) {
+      logger.log(Level.WARNING, "Metric Exporter threw an Exception", e);
+    }
```
huntc, line 66: This logging is consistent with other internal call sites.

exporters/logging/src/main/java/io/opentelemetry/exporters/logging/LoggingSpanExporter.java
```diff
@@ -50,11 +53,15 @@ public ResultCode flush() {
...
   @Override
   public void shutdown() {
-    this.flush();
+    try {
+      this.flush().get();
+    } catch (InterruptedException | ExecutionException e) {
+      logger.log(Level.WARNING, "Span Exporter threw an Exception", e);
+    }
   }
 }
```
huntc, line 67: Note that if any of the existing exporters throw an exception during export or flush then they are in error in accordance with the API. I've therefore assumed that they don't.

exporters/inmemory/src/main/java/io/opentelemetry/exporters/inmemory/InMemoryMetricExporter.java
huntc: Java 8's `CompletableFuture` implements `Future`. `Future` describes a behaviour that is very much modern Java.

anuraaga: It implement's `Future` for legacy apps :) But `Future` itself only has the possibility of calling `.get`, which is a blocking method, so it's not quite asynchronous. While we weren't really clear on when this result code would actually even be used, but if the goal is to make this fully asynchronous (no blocking for the result at all) while also returning a result, on Java 7 I think it does need to be a callback.

huntc: `Future` is purely an interface that describes a behaviour. Its usage here is legitimate.

huntc: I'm happy if you'd like to change it from a `Future` to a callback yourself though.

huntc: Are there other areas of the API that prefer callbacks over what would otherwise be futures? It is important for this PR to be consistent with them of course.

anuraaga: I think this is the first async API. 

If the goal is to make this a fully async API, `Future` doesn't really cut it. For example, if an event loop callback needs this result, maybe in a reactive stream, it will have to delegate to an I/O threadpool before calling `Future.get`. Perhaps this isn't really an expected use case in which case, we always expect `.get` to called in a synchronous context. If so then `Future` works fine but we have to be clear that it limits the use cases for the API.

huntc: How would you feel about exposing Guava's `ListenableFuture` here? Back in the day, this was always my go-to before Java 8 came along. And, Guava is already on the list of dependencies.

Alternatively, OT could introduce its own `Future` subtype that could host a `ListenableFuture` and migrate to a `CompletableFuture` later on if/when the Java 7 dependency can be dropped.

WDYT?

anuraaga: I like `ListenableFuture` but exposing Guava in the API might be a bit much - interested in others' thoughts.

Definitely, as opposed to a callback parameter, a custom `Future` subtype would work too. As long as it has something like `addListener`, similar to Netty's future, would work well in event loops.

jkwatson:
I honestly don't see how this PR improves anything for anyone. This makes a breaking change to the API for exporters, but with no actual improvement to any functionality. And, the introduction of a guava dependency on a number of exporters seems like a very large downside.

Can you explain what your end-goal is with making this change? Just adding "async"-ness for the sake of adding it doesn't seem like a change that has any value for any actual exporter author or end user.

huntc:
@jkwatson We had a detailed discussion for this PR’s motivation in https://github.com/open-telemetry/opentelemetry-java/issues/1422. I asked you if you wanted me to raise a PR.

In summary, this PR will assist exporters from blocking a thread in order to return a status. The PR introduces an API change now, while we can still get away with API changes. There is no implementation benefit right now, but the opportunity is there in the future. 

The Go and .NET APIs also avoid blocking.

jkwatson:
I guess I had expected there to be some consumption of the Future/Callback in the SDK itself, so that there might actually be a benefit to this change. As things are today, nothing stops an exporter from doing all the work async and returning one of the enum values immediately.

So, I guess my question still stands: how do you see the SDK actually using this future? The Future API as it is also doesn't stop an exporter from making the whole call block and just returning an Future at the end of the blocking process. In fact, I believe that's what you implemented with this PR.

If we're going to make this change, let's make it meaningful, and not just pro-forma. Does that make sense?

huntc:
By returning a reply for export, whether async or not, the SDK can control whether to release memory for metric and span data.

If the export reply remains synchronous then the exporter must also decide whether to release any buffered data as a result of network failure. 

Do we want this logic in both places?

jkwatson:
I get all that. What I don't want, though, is to have this PR land, and then not have the next steps implemented, leaving us with a half-implemented idea. Would you be willing to put together a PR that demonstrates all of this, with changes in the SDK that take advantage of this change? If we like how it all looks, we can circle back to this PR and get it merged as the first step in that larger change.

Also, if you're interested in taking on this scope of work, it would be great to have you join our weekly SIG meeting where we can talk through these plans in real time.

Sound good?

huntc:
Unfortunately, having spent a number of hours already, and due to other work commitments, I’m unable to justify any advancement of this PR.

Feel free to close this PR and it’s associated issue.

jkwatson:
Hmm. I'm sorry to hear that. I would really like to get working on modernizing the codebase (as much as we can given the java 7 requirement). I think these ideas will mesh well with other work-in-progress like #1419 and #1418 . We don't have a good story on export error reporting at the moment, leaving it up to the exporter authors to handle that on their own.

I think that I'd personally prefer a callback, given the cruftiness of the `Future` API (as @anuraaga pointed out). I think that would end up working very well in the context of #1419, and would tie us less to the antiquated `Future` APIs.

huntc:
I don’t see callbacks as being contemporary. 

If you like, I can spend some time over the weekend to look at dropping the Guava dependency and convert the export status enum to represent an async value (without Future). This should be straightforward. 

If there’s no appetite for that, I understand. Sorry that I can’t spend more time than this right now.

jkwatson:
I'm not sure what you mean by "convert the export status enum to represent an async value. " Can you elaborate?

I don't think callbacks are up-to-date, or contemporary, but they don't tie the APIs to what is definitely an antiquated API. At least with callbacks, we completely control the API, soup to nuts. 

@anuraaga What do you think?

huntc:
Perhaps you guys can save me some time by sketching out the interfaces you’d like to see. 

Personally, I still see Future as being correct given that it describes the behaviour we are wanting, it isn’t deprecated and it forms the interface of the Java 8 CompletableFuture. Runtime checks can be made for handling specific types of Future. 

Btw I would like to learn more about why Java 8 isn’t the baseline. It’s causing a bit of a headache here. Have you got a ref to the decision making around this?

anuraaga:
@jkwatson I'm personally OK with either a callback or a custom implementation of `Future` that can have callbacks added. Callback seems a bit simpler to implement and I generally wouldn't avoid it out of principle since pre-Java 8 it was very common for asynchronous programming.

```java
public interface AsyncCallback<T> {
  void onComplete(T result);
  void onError(Throwable t);
}

public void export(Collection<SpanData> spans) {
  export(spans, AsyncCallback.NO_OP);
}

public void export(Collection<SpanData> spans, AsyncCallback<ResultCode> callback);
```

or

```java
public io.opentelemetry.api.Future export(Collection<SpanData> spans);
```

@huntc Java 7 is mostly for Android where it's still common to target older OSes for device compatibility. Not sure about this SDK repo, but the instrumentation repo has some discussion here https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/599

huntc:
Thanks. So, given the decision to drop support for Java 7, let’s sit on this PR until we can use CompletableFuture. Sound good?

anuraaga:
@huntc The decision is to drop support for Java 8 in most instrumentation, but not ones that are common in Android. But the SDK will have to stick to Java 7 regardless, or that would effectively require all instrumentation to be Java 8.

huntc:
Hi @jkwatson and @anuraaga, I've been playing around with an approach that should address all of the concerns to-date:

1. No Guava as a dependency
2. No use of `Future`
3. Leverages existing schedulers to process async results

So far, I've just coded-up a `CompletableResultCode` class and modified the `MetricExporter` calling site, `IntervalMetricReader`. I shall go further if we attain a consensus.

Here's the `CompleteableResultCode`: https://github.com/open-telemetry/opentelemetry-java/pull/1428/files#diff-54755a6ca77edd2993c00462b790c1cfR1-R76, which is then used by `MetricExporter`: https://github.com/open-telemetry/opentelemetry-java/pull/1428/files#diff-35820ddda83973365a3f3012cfdbdd23R19-R52. The `MetricExporter` is called by the `IntervalMetricReader`, which ensures that no more exports occur until the previous export has finished and also avoids blocking: https://github.com/open-telemetry/opentelemetry-java/pull/1428/files#diff-37f07a36d5e4986a125847c65701db0bR192-R228.

My approach bears similarity to Java 8's `CompletableFuture` for the sake of developer familiarity and the potential of future migration.

Thoughts?

exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java
```diff
@@ -126,9 +128,9 @@ public ResultCode export(Collection<SpanData> spans) {
       // for now, there's nothing to check in the response object
       //noinspection ResultOfMethodCallIgnored
       stub.postSpans(request);
-      return ResultCode.SUCCESS;
+      return Futures.immediateFuture(ResultCode.SUCCESS);
```
jkwatson, line 131: if we go with this PR, we should definitely create follow-on issues to make the exporter implementations actually do their work async.

exporters/logging/src/main/java/io/opentelemetry/exporters/logging/LoggingSpanExporter.java
jkwatson: logging is fine here, agreed.

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -285,9 +285,9 @@ private void onBatchExport(final List<SpanData> spans) {
                 public void run() {
                   // In case of any exception thrown by the service handlers catch and log.
                   try {
-                    spanExporter.export(spans);
-                  } catch (Throwable t) {
-                    logger.log(Level.WARNING, "Exception thrown by the export.", t);
+                    spanExporter.export(spans).get();
```
jkwatson, line 288: this is another point at which we should probably have a follow-on issue, so we don't lose track of the fact that we're making something sync that could be async.

sdk/src/main/java/io/opentelemetry/sdk/trace/export/SimpleSpanProcessor.java
```diff
@@ -77,8 +78,8 @@ public void onEnd(ReadableSpan span) {
     }
     try {
       List<SpanData> spans = Collections.singletonList(span.toSpanData());
-      spanExporter.export(spans);
-    } catch (Throwable e) {
+      spanExporter.export(spans).get();
```
jkwatson, line 81: ditto on needing an issue for this one as well.

exporters/inmemory/build.gradle
jkwatson: definitely for the in-memory exporter, this is 100% fine.

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
```diff
@@ -0,0 +1,70 @@
...
+public class CompletableResultCode {
+  public CompletableResultCode() {}
+
+  private volatile boolean succeeded = false;
+  private boolean completed = false;
+  private Runnable action = null;
+  private Executor actionExecutor = null;
+
+  /** The export operation finished successfully. */
+  public synchronized void success() {
```
jkwatson, line 21: I'm not super happy with having a public method be synchronized. There is an unintended consequence to this that if someone externally uses an instance of this class to synchronize on, it will lock this from every executing.

The combination of volatile and public synchronized methods makes me a bit queasy. Can we use some more modern concurrency primitives, like Atomics to make this a little less prone to accidental (or intentional) abuse?

At the very least, if we do need to have old-school synchronization in here, we should synchronize on a private final field that is not accessible outside the class.

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.concurrent.Executor;
+
+/**
+ * The implementation of Export operations are often asynchronous in nature, hence the need to
+ * convey a result at a later time. CompletableResultCode facilitates this.
+ *
+ * <p>This class models JDK 8's CompletableFuture to afford migration should Open Telemetry's SDK
+ * select JDK 8 or greater as a baseline, and also to offer familiarity to developers.
+ */
+public class CompletableResultCode {
```
anuraaga, line 12: We may as well make this generic

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
anuraaga: Agree that we can probably use atomic and cas to avoid synchronization. We're not an asynchronous framework or anything, this is just our export API so we wouldn't expect significant overhead due to extra objects from the atomics

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
```diff
@@ -0,0 +1,70 @@
...
+   * @return the current state of completion
+   */
+  public boolean isSuccess() {
+    return succeeded;
+  }
+
+  /**
+   * Perform an action on completion. Actions are guaranteed to be called only once.
+   *
+   * <p>There should only be one action for this class instance.
```
anuraaga, line 56: If we only support one callback anyways I would definitely prefer just using the callback pattern. How about we just go ahead and maintain a list of runnables here?

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
anuraaga: I realized my comment about list of runnables may make it impractical to avoid a lock.

exporters/inmemory/build.gradle
huntc: It will no longer be required anywhere if I contribute to apply my CompletableExportStatus.

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
huntc: I can replace the use of synchronised with an internal lock. It isn’t possible to use atomics for each field as we require them to be all synchronised simultaneously.

huntc: Also, please avoid wording like “queasy” when providing feedback. The use of volatile where I have it is perfectly reasonable. Please be clearer as to why an approach is unsuitable.

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
huntc: This will add a lot of complexity, which I don’t think is warranted. Ultimately migrating to a CompleteableFuture approach for all async use cases would be best.

I’m not attempting to solve all Open Telemetry async use cases, and I want to ensure that the approach taken here remains reasonable in scope.

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
huntc: I’d prefer to avoid what would be additional complexity, but composing multiple `andThenAsync` methods would be the idiomatic thing to do.

anuraaga: I think we're sort of not past the cliff in terms of complexity. For comparison, the callback pattern is just defining an interface, so it's about as close to no-complexity as changes can get. Here we've added a lot of complexity, having to worry about synchronization to boot, but it seems the only additional feature is a custom executor, which is arguably not that important (our goal was to make the API usable from an async context). So unless we go a bit further, I don't see any reason to take this approach over a normal callback.

Thinking again (thanks for bringing up the point about complexity, I was forgetting it :) ), if we can only imagine a use case where there is a single callback that does something very simple like do a log statement, update a metric, then I'd still lean towards the callback pattern (but I'm still ok with the listenable `Completable` as well if others prefer it). 

It does bring to mind another idea, what about dropping the result code completely? Just a single code seems quite limiting anyways, instead registering a listener on the exporter itself (not on export calls) seems like it could cover all expected use cases, possibly in a more controlled fashion? For example, Zipkin's reporter metrics callback could work well if registered directly on the `Exporter`, and allows consistency in handling success, failure, etc across all exporters.

https://github.com/openzipkin/zipkin-reporter-java/blob/master/core/src/main/java/zipkin2/reporter/ReporterMetrics.java#L44

jkwatson:
As discussed earlier, if we want to drop the result code altogether, we'll need to make a change to the spec.

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
jkwatson: Sorry.. I was just trying to express that what is implemented here has a mix of different concurrency paradigms in it. When I see that, it is a "code smell". And "code smells" make me uncomfortable. When I read this code, I have a hard time understanding why it is written the way it is, and that's a problem.

anuraaga:
While I'm generally scared of filing spec issues due to the risk of taking months to go nowhere, went ahead and filed it :) https://github.com/open-telemetry/opentelemetry-specification/issues/716

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
huntc: I will be applying the CompletableMetricStatus here too.

sdk/src/main/java/io/opentelemetry/sdk/trace/export/SimpleSpanProcessor.java
huntc: This will go away

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
huntc: In essence, using volatile on a boolean is more efficient than using an `AtomicBoolean`. A reason to use `AtomicBoolean` would be to use its CAS operations, which are not required.

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
huntc: I see have a completable status as useful in that the sdk should provide telemetry on a best-effort basis. If a previous export has not yet finished, then we should look at dropping the metrics. I shall include this change so we can then debate its merit.

In terms of complexity, `andThen`, when used with JDK 8's `CompletableFuture` requires the guarantee that any subsequently composed `andThen` occurs after. Therefore, we would have to maintain a list of actions. I can do this.

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
jkwatson: Could we just replace both fields with a single AtomicReference holding the return code itself? Then you could CAS to set it?

huntc: > Could we just replace both fields with a single AtomicReference holding the return code itself? Then you could CAS to set it?

They'd be three fields... We could use a CAS on a state object, but I've now pushed some code that uses a reentrant lock. See what you think of that... it is quite focused on reducing object allocations.

huntc:
@jkwatson @anuraaga I've pushed something which is reasonably complete, but there are test failures because I'm now dropping exports if the previous export hasn't completed. I think this is the correct behaviour, but rather than fixing things up, or possibly introducing a queue to mitigate the loss of telemetry, I thought we should debate this. I'm personally ok with telemetry being exported on a best-effort basis, but I'm very interested to learn of your opinions on this.

sdk/src/main/java/io/opentelemetry/sdk/metrics/export/MetricExporter.java
```diff
@@ -29,36 +30,22 @@
  */
 public interface MetricExporter {
 
-  /**
-   * The possible results for the export method.
-   *
-   * @since 0.1.0
-   */
-  // TODO: extract this enum and unify it with SpanExporter.ResultCode
```
huntc, line 37: This TODO has been actioned.

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -96,15 +90,8 @@ private BatchSpanProcessor(
...
       int maxQueueSize,
-      int maxExportBatchSize,
-      int exporterTimeoutMillis) {
-    this.worker =
-        new Worker(
-            spanExporter,
-            scheduleDelayMillis,
-            maxQueueSize,
-            maxExportBatchSize,
-            exporterTimeoutMillis);
```
huntc, line 107: A bit of a change here is that there's no longer a timeout on export. Given that export is potentially async, we can't reliably cancel. Also, an exporter's contract is to always complete a `CompletableResultCode` - it is a promise that must be fulfilled.

huntc: Also, we don't need to hand tasks off to a worker thread if the exporters are async. They will have their own thread pools. It is important to keep the number of threads used to a minimum. One thread for metrics and another for spans should be all that is required in total.

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
jkwatson: So, as far as I understand this class (and maybe I'm misunderstanding), there are 3 states. 1) incomplete, 2) complete with failure, 3) complete with success. This can be modeled with one AtomicReference that holds the success/failure. AtomicReference is unset: incomplete, AtomicReference is set to failure: complete with failure, AtomicReference is set to success: complete with success.

Am I missing something?

jkwatson:
There's a lot to digest here. I need to spend some real time with this to evaluate it for correctness. Note: removing the export timeout now makes us non-spec-compliant, but maybe that's ok?

jkwatson:
Note: before making changes of this scope, I'd love to have this issue complete so we can validate the changes aren't causing a regression: https://github.com/open-telemetry/opentelemetry-java/issues/1438

carlosalberto:
> Note: removing the export timeout now makes us non-spec-compliant, but maybe that's ok?

Since when being non-compliant is fine? 😄 I started to feel this *might* be discussed in the Specification - then again, maybe not, as maybe this is a very language independent feature.

Haven't been following closely the entire conversation, but after going through the comments, I can't find the exact problem is this fixing in the SDK.  Can you summarize?

huntc:
> Note: removing the export timeout now makes us non-spec-compliant, but maybe that's ok?

Can you please cite where in the spec? I would’ve though this’d might depend on the language implementation.

jkwatson:
> > Note: removing the export timeout now makes us non-spec-compliant, but maybe that's ok?
> 
> Can you please cite where in the spec? I would’ve though this’d might depend on the language implementation.

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#batching-processor

huntc:
> Haven't been following closely the entire conversation, but after going through the comments, I can't find the exact problem is this fixing in the SDK. Can you summarize?

The first para of this PR summary still captures the motivation. However, I will look to enhance it and explain the improvement by avoiding blocking operations.

exporters/inmemory/src/main/java/io/opentelemetry/exporters/inmemory/InMemoryMetricExporter.java
huntc: Future is no longer used. Resolving.

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
huntc: We need to lock both the completion state and the callback list (`actions`) at the same time. Not doing so could lead to a race condition where `action` is added but not executed immediately because the completion has subsequently occurred. You would, therefore, end up with an action never being called.

huntc:
> > > Note: removing the export timeout now makes us non-spec-compliant, but maybe that's ok?
> > 
> > 
> > Can you please cite where in the spec? I would’ve though this’d might depend on the language implementation.
> 
> https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#batching-processor

Thanks for this. I can re-introduce it, but I think having this cancellation timeout is ineffective. For example, could the New Relic export be cancelled given that work has been passed to another thread within the exporter?

Timeouts on exports should perhaps be passed onto the exporter for them to manage. But then, why not also pass on retry configuration? Personally, I think each exporter is responsible for timing itself out and retrying etc as their behaviour is sensitive to the type and nature of network connectivity (if any).

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
jkwatson: Sure, but even if you need the explicit locking to manage the actions, you can still store the complete state in one nullable field, whether you store it in an AtomicReference or not. Currently, you're using two fields, updates to which need to be coordinated, which is inherently more complex that just managing one field with the result in it.

jkwatson:
> > > > Note: removing the export timeout now makes us non-spec-compliant, but maybe that's ok?
> > > 
> > > 
> > > Can you please cite where in the spec? I would’ve though this’d might depend on the language implementation.
> > 
> > 
> > https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#batching-processor
> 
> Thanks for this. I can re-introduce it, but I think having this cancellation timeout is ineffective. For example, could the New Relic export be cancelled given that work has been passed to another thread within the exporter?
> 
> Timeouts on exports should perhaps be passed onto the exporter for them to manage. But then, why not also pass on retry configuration? Personally, I think each exporter is responsible for timing itself out and retrying etc as their behaviour is sensitive to the type and nature of network connectivity (if any).

100% Agree with this. Just for context, the specs (and this codebase) used to have a `FAILED_RETRYABLE` enum value for the ResultCode, but no specification, or implementation that did anything with it. I did manage to get that removed from the spec (and here), at least. :)

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
huntc: Ah, good point. I have now pushed the change.

huntc:
> 100% Agree with this. Just for context, the specs (and this codebase) used to have a `FAILED_RETRYABLE` enum value for the ResultCode, but no specification, or implementation that did anything with it. I did manage to get that removed from the spec (and here), at least. :)

I raised: https://github.com/open-telemetry/opentelemetry-specification/issues/718

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
```diff
@@ -0,0 +1,99 @@
...
+  public static final CompletableResultCode SUCCESS = new CompletableResultCode().success();
+
+  /** A convenience for declaring failure */
+  public static final CompletableResultCode FAILURE = new CompletableResultCode().failure();
+
+  public CompletableResultCode() {}
+
+  private volatile Boolean succeeded = null;
+  private final ArrayList<Runnable> actions = new ArrayList<>();
+  private final Lock state = new ReentrantLock();
```
anuraaga, line 26: Since we have an object for this lock now, is there any need for it to be a `ReentrantLock` as opposed to an `Object` which we use `synchronized` on? IIRC, the latter is a bit faster with a bit less syntax (though maybe not in Java 15 where they're finally removing biased locking)

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
```diff
@@ -0,0 +1,99 @@
...
+
+  /** The export operation finished successfully. */
+  public CompletableResultCode success() {
+    List<Runnable> actionsToRun = new ArrayList<>();
+    state.lock();
+    try {
+      if (succeeded == null) {
+        succeeded = true;
+        actionsToRun.addAll(actions);
```
anuraaga, line 35: Do we need this copy? With `succeeded` always accessed in the lock seems ok without it

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
```diff
@@ -0,0 +1,99 @@
...
+   * <p>There should only be one action for this class instance.
+   *
+   * @param action the action to perform
+   * @return this completable result so that it may be further composed
+   */
+  public CompletableResultCode thenRun(Runnable action) {
+    boolean runActionImmediately = false;
+    state.lock();
+    try {
+      if (succeeded != null && succeeded) {
```
anuraaga, line 86:
```suggestion
      if (succeeded != null) {
```

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -277,28 +258,26 @@ private void exportBatches(ArrayList<ReadableSpan> spanList) {
     }
 
     // Exports the list of SpanData to the SpanExporter.
+    @SuppressWarnings("BooleanParameter")
     private void onBatchExport(final List<SpanData> spans) {
-      Future<?> submission =
-          executorService.submit(
+      if (exportAvailable.compareAndSet(true, false)) {
```
anuraaga, line 263: IIUC this changes the behavior to export essentially being "scheduleAtFixedRate" instead of the current "scheduleAtFixedDelay" since this always returns immediately now. I don't know if this is the intention of this class. Also given a significant refactor in #1395 which will make this class much easier to reason about, maybe we can keep the behavior unchanged for now?

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
huntc: What happens when the object is null?

Also, do you have a ref re. `synchronized` outperforming `ReentrantLock`. I'd imagine that the former is implemented using the latter via pthreads, at least on Unix.

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
huntc: Its paranoia on my part. By copying, we can avoid a deadlock on having this object's API called from within an action handler. Although... we're using a ReentrantLock so we can already avoid that situation! I shall remove the copying.

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
anuraaga: It wouldn't ever be `null` since you'd declare `private final Object lock = new Object()`. We use the pattern in the span itself too, as do many JDK classes, so it'd be a bit more consistent too.

https://github.com/open-telemetry/opentelemetry-java/blob/master/sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java#L73

Not sure about all platforms, but on the main ones Java doesn't use OS for locks, it's all in the JVM using CAS, including crazy stuff that only applies for `synchronized` AFAIK (though it's going away)

https://wiki.openjdk.java.net/display/HotSpot/Synchronization

Here I'd err on consistency with the span and a couple lines of code less if there's no specific need for the `ReentrantLock`.

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
huntc: Pushed changes

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * The implementation of Export operations are often asynchronous in nature, hence the need to
+ * convey a result at a later time. CompletableResultCode facilitates this.
+ *
+ * <p>This class models JDK 8's CompletableFuture to afford migration should Open Telemetry's SDK
+ * select JDK 8 or greater as a baseline, and also to offer familiarity to developers.
+ */
+public class CompletableResultCode {
+  /** A convenience for declaring success */
+  public static final CompletableResultCode SUCCESS = new CompletableResultCode().success();
```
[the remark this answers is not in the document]
huntc: Done. Changes pushed.

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
huntc: Sorry, I misunderstood which object you were referring to.

We actually do need the `ReentrantLock` in case an `andThen` action calls upon other parts of the API, right?

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
huntc: I think we still need the changes here. Contractually, we should guarantee that an export function is not called if it has a request in-flight. It isn't yet clear to me how the proposed changes of #1395 would help.

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
huntc: Actually, your suggestion appears to support reentrancy also (I didn't know! https://docs.oracle.com/javase/tutorial/essential/concurrency/locksync.html). I shall change to be consistent with other parts of the code.

huntc: Changes pushed

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
huntc: Whoops. Thanks. Changed

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
```diff
@@ -0,0 +1,93 @@
...
+  @GuardedBy("lock")
+  private Boolean succeeded = null;
+
+  @GuardedBy("lock")
+  private final ArrayList<Runnable> actions = new ArrayList<>();
+
+  private final Object lock = new Object();
+
+  /** The export operation finished successfully. */
+  public CompletableResultCode success() {
```
anuraaga, line 38: Thanks last nit, these methods looks weird at the call sites since they're not verbs. Something like

`succeed` / `fail`
`complete` / `completeExceptionally`

?

huntc: `succeed`/`fail` ok? They are both completions, and we're not completing with an exception...

huntc: Yes, we need unit tests... I'll put a TODO section in the summary and convert this to a draft PR.

huntc:
Happy to finish this PR, but I’ll need some more encouragement that it is headed in the right direction.

jkwatson:
Before we make a decision one way or another on this approach...  Did you have a chance to take a look at the async disruptor-based exporter-wrapper here?  https://github.com/open-telemetry/opentelemetry-java/tree/master/sdk_extensions/async_processor

huntc:
I’ve not taken a detailed look at the disruptor span processor. However, this PR and it’s associated issue has been mainly focused on having the export method types reflect the desired behaviour.

Even with an improved span processor, the export operations are long lived and shouldn’t block.

jkwatson:
Take a look at the disruptor exporter I linked to. It wraps any other SpanExporter, and provides an async interface to it. We should probably consider something similar for metrics.

anuraaga:
As far as I can tell it transfers tasks to a worker thread to process events. This might be useful for processors that really need to block (JDBC?) but agree with @huntc that it would make sense to make the export methods async to not require that for exporters. For example a gRPC exporter can be written asynchronously trivially with no need for the disruptor which would just add lots of overhead and no benefit.

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -277,28 +258,26 @@ private void exportBatches(ArrayList<ReadableSpan> spanList) {
...
               new Runnable() {
                 @Override
                 public void run() {
-                  // In case of any exception thrown by the service handlers catch and log.
-                  try {
-                    spanExporter.export(spans);
-                  } catch (Throwable t) {
-                    logger.log(Level.WARNING, "Exception thrown by the export.", t);
+                  if (!result.isSuccess()) {
+                    logger.log(Level.FINE, "Exporter failed");
```
jkwatson, line 271: this makes me think that we probably should think about adding an Exception/Throwable to the CompletableResultCode, but that isn't necessary for this PR.

sdk/src/main/java/io/opentelemetry/sdk/trace/export/MultiSpanExporter.java
```diff
@@ -67,18 +68,19 @@ public ResultCode export(Collection<SpanData> spans) {
    * @return the result of the operation
    */
   @Override
-  public ResultCode flush() {
-    ResultCode currentResultCode = SUCCESS;
+  public CompletableResultCode flush() {
+    final CompletableResultCode compositeResultCode = new CompletableResultCode();
+    final CountDownLatch completionsToProcess = new CountDownLatch(spanExporters.length);
```
jkwatson, line 73: is anything waiting on this latch?

jkwatson:
let's do it. This definitely needs a rebase, and to get all the tests passing, and I left a couple more comments, but I'm convinced now this is a good step forward. 

And, I really wish we had java 8, because this would all be so much easier.

huntc:
Great to hear. I shall proceed. Also, I’ll reinstate the processor timeout so that we can remain spec compliant - even though I think it is of limited utility - at least until project loom lands.

As an additional note, I’ll introduce buffering so that there’s less chance of losing telemetry received when an export is in flight. A buffer of max int size would effectively be unbounded, but smaller bounded buffers should make a reasonable default and keep memory managed nicely.

huntc:
Rebased and pushed

sdk/src/main/java/io/opentelemetry/sdk/trace/export/MultiSpanExporter.java
huntc: Nothing needs to wait - See https://github.com/open-telemetry/opentelemetry-java/pull/1428/files#diff-99eb9ce2c0a6fa65d07ccbbee42ba836R107 - when the latch reaches zero, given that there's been a success for the entire composition of result codes, we then succeed the composite.

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
huntc: The contract is for an exporter to either succeed or fail as per the result code. In the case of failure, it could be worth indicating why, with an optional throwable as a parameter. I'd say that it is up to the export class itself to determine what should constitute being worthy of logging here though, as failure may well be "normal" and I don't think we want to pollute the logs.

jkwatson:
> Great to hear. I shall proceed. Also, I’ll reinstate the processor timeout so that we can remain spec compliant - even though I think it is of limited utility - at least until project loom lands.
> 
> As an additional note, I’ll introduce buffering so that there’s less chance of losing telemetry received when an export is in flight. A buffer of max int size would effectively be unbounded, but smaller bounded buffers should make a reasonable default and keep memory managed nicely.

I'd also be ok with making these issues to be resolved with follow-up PRs, to keep the scope of each individual changes more manageable.

carlosalberto:
Please don't work on this more until we have reached an agreement - I know now it's all excitement, but I'm very wary of sexy changes that later we may need to discard 😞 

As usual, I'd like to get an up-to-date summary of what this solves and probably a prototype of an exporter + a case that takes advantage of this new API (this is even the recommended way at the Specification level, where you provide a prototype that shows the advantages). I think a gRPC case was mentioned somewhere, so that could be tried.

Also: I know that @huntc created an issue in the Specification level on a related issue, but we *might* need to discuss this issue there (maybe, maybe not, as this may be an implementation detail for Java).

huntc:
> Please don't work (much) on this more until we have reached an agreement -

Who is “we”, and what constitutes agreement?

> I know now it's all excitement, but I'm very wary of sexy changes that later we may need to discard 

Please don’t assume excitement. My understanding is that a call to an exporter with the simple span processor will suspend program execution. This should not be a possibility. 

Also, recognising that an exporter performs a long lived operation via its types should be obvious. It is at least obvious to both the Go and .Net projects from what I can see.

carlosalberto:
Hey @huntc 

> Is there a consensus required between the maintainers?

Yes (at least we usually try to go that way :) )

> Exporters perform long-lived operations. Their method signatures should reflect that out of correctness.

I see the Disruptor processor has been deemed as a thing to avoid (for some cases, I'm suspecting), but what about using `BatchSpanProcessor`? What's the reason to avoid using that one? I'm still wondering what case the current design fails to address.

On the .Net/Go mention, I'd definitely put Go out of the comparison as it's a different animal IMHO. But I'm curious about the .Net take on this and will check it out.

Also, I'd love to hear @bogdandrutu's opinion, as he was the one originally designing the processor/exporter layer, and he might have something to say.

huntc:
> ... but what about using `BatchSpanProcessor`? What's the reason to avoid using that one? I'm still wondering what case the current design fails to address.

I'm a fan of the simple span processor because there's no thread pool other than any required by the exporter(s). No thread pool == more efficient memory usage. Threads are expensive. The downside of the simple span processor is that if an exporter blocks, so will my program. Exporters should thus never block their calling threads.

Again, getting the method signature for the exporter is my primary concern. The current design fails to convey that exporter operations are long-lived. Types are important.

carlosalberto:
> I'm a fan of the simple span processor because there's no thread pool other than any required by the exporter(s). No thread pool == more efficient memory usage. Threads are expensive.

Oh, that's very interesting. Sounds reasonable, yes, and worth trying out.

> The current design fails to convey that exporter operations are long-lived.

True, but I'd also love to see this change validated with some actual prototype that takes full advantage.

(Hope @bogdandrutu can chime in real soon ;) )

huntc:
> True, but I'd also love to see this change validated with some actual prototype that takes full advantage.

I'm unsure what you mean by "advantage". It's more a question of correctness. If you're looking for an async exporter prototype though, and how exports can be safely canceled then perhaps the following test will be illustrative: https://github.com/open-telemetry/opentelemetry-java/blob/b7cfcab79e4cad7dbc7edf2269e0355a506d39f8/sdk/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorTest.java#L314

Note that prior to this PR, canceling exports via interrupting them is unsafe. So, perhaps here's another good reason then. :-)

But again, my motivation here is mostly on the subject of correctness.

exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcMetricExporter.java
```diff
@@ -89,7 +90,7 @@ private OtlpGrpcMetricExporter(ManagedChannel channel, long deadlineMs) {
    * @return the result of the operation
    */
   @Override
-  public ResultCode export(Collection<MetricData> metrics) {
+  public CompletableResultCode export(Collection<MetricData> metrics) {
```
anuraaga, line 93: @carlosalberto We will need to change this from `MetricsServiceBlockingStub` to `MetricsServiceFutureStub` to see the main result of this PR, making at least the gRPC exporters truly async. I'm ok with that being in a followup but do you want to see it now?

@huntc I guess it's not a huge code change so maybe can knock it out.

sdk/src/main/java/io/opentelemetry/sdk/trace/export/MultiSpanExporter.java
anuraaga: If not waiting on it, may as well use `AtomicInteger` to clarify that it's just for counting.

sdk/src/main/java/io/opentelemetry/sdk/trace/export/SimpleSpanProcessor.java
```diff
@@ -71,15 +74,30 @@ public boolean isStartRequired() {
...
   public void onEnd(ReadableSpan span) {
-    if (sampled && !span.getSpanContext().getTraceFlags().isSampled()) {
-      return;
-    }
-    try {
-      List<SpanData> spans = Collections.singletonList(span.toSpanData());
-      spanExporter.export(spans);
-    } catch (Throwable e) {
-      logger.log(Level.WARNING, "Exception thrown by the export.", e);
+    if (exportAvailable.compareAndSet(true, false)) {
```
anuraaga, line 79: For `SimpleSpanProcessor` I think I'd expect it to be simple and just call export in all cases (if there is a backup in the exporter, that is a reason to use a more complicated span processor like batch span processor).

exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcMetricExporter.java
huntc: That'd be great @anuraaga - could you make a PR on my branch?

anuraaga: @huntc Sure sent https://github.com/huntc/opentelemetry-java/pull/1

sdk/src/main/java/io/opentelemetry/sdk/trace/export/SimpleSpanProcessor.java
huntc: That would then require exporters to become re-entrant. Our assumption at present is that only one export is called at a time (which is probably something we should document - it can simplify the authoring of exporters - wdyt?).

exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcMetricExporter.java
huntc: Merged - thanks @anuraaga. @carlosalberto You've now got a fully async exporter.

sdk/src/main/java/io/opentelemetry/sdk/trace/export/SimpleSpanProcessor.java
jkwatson: I'm not sure what you mean by "re-entrant" here. Do you just mean having more than one export running at a time, or are you actuallyreferring to sending the same spans more than once and having the exporter be able to handle de-duping them?

huntc: Meanwhile, I've added some API docs to clearly state the position on re-entrancy.

huntc: > I'm not sure what you mean by "re-entrant" here. Do you just mean having more than one export running at a time, or are you actuallyreferring to sending the same spans more than once and having the exporter be able to handle de-duping them?

I was referring to the number of `export` or `flush` calls in flight at one time, which is what this code is associated with. We only want one of each of these calls in flight. In fact, we shouldn't really call `export` and `flush` at the same time either. Keeps the programming model simpler for exporters.

huntc: BTW I'm not seeing that the SDK ever calls on `flush`. I'm now wondering about its utility. :-)

jkwatson: The idea with `flush` wasn't for the SDK itself to call it, but for (for example), a lambda to be able to call `flush` explicitly, so that spans wouldn't be left in a queue, unsent, when the lambda shuts down.

huntc: Thanks for the clarification. Should we be marking it as `protected` then, to clarify the intent?

anuraaga: > We only want one of each of these calls in flight.

I dunno about this - since an exporter is a function that takes spans and pushes them to a sink, I'd be surprised that it would generally matter whether there is one or more in flight. An exporter that depended on being called only once sounds like it's more complicated than it should be. I think right now, if `SimpleSpanProcessor` is used and two spans are ended at the same time, maybe as a result of being parallel client RPCs, one of them is guaranteed to be dropped right? I don't think that's the intention of the `SimpleSpanProcessor`.

huntc: Ah - very good point. I'll document them as being re-entrable.

anuraaga: >  Should we be marking it as protected then, to clarify the intent?

By lambda, I think @jkwatson meant AWS Lambda, not Java lambda, or other Google Cloud Functions type of serverless environment where background threads get suspended as soon as a response is returned. So `flush` needs to stay `public` to be called from the app.

huntc: Got it. Thanks.

huntc: I've just made another commit that now clarifies that an `export` or `flush` can be called simultaneously. This thinking has really cleared up the code as I was also able to remove the barrier that enforced one `export` at a time. All tests appear to be working locally for me now.

huntc: @anuraaga I'm worried that the `OtlpGrpc` and other exporters have no appearance of backpressure in place. I'm presuming that when there's a problem trying to connect, many requests will pile up (this could happen prior to this PR).

The exporters should generally handle `CompletableResultCode.whenCompleted`, test for failure and then do what is necessary to cancel any in-flight request. Additionally, an exporter might decide to avoid another export if there's one in-flight and also buffer a small amount.

With Akka-streams, writing an exporter is quite straightforward. I'd create a bounded queue and drop new elements given any back-pressure. This can mostly be achieved with a `Source.queue` call. We don't have Akka-streams here though. :-)

Any thoughts on this?

anuraaga: I think this sort of backpressure is handled by the batch processor which drops when the queue is full.

The simplespanprocessor doesn't but it's why we recommend batch for most apps, simple is more specialized that requires some thought to use.

This seems ok to me but what do you think?

huntc: I think we should document `SimpleSpanProcessor` with a cautionary note. I shall do this now.

huntc: Having said that, the `IntervalMetricReader` has no such back-pressure mechanism.

huntc: So, should I re-instate the ability to only call `export` one at a time for the `IntervalMetricReader` and `BatchSpanProcessor`, and have them drop any new telemetry in the case where an `export` is in-flight? We can leave the `SimpleSpanProcessor` as-is though i.e. without the "in-flight" check. The `SimpleSpanProcessor` will be perfect for those more sophisticated exporters, such as what can be written with Akka-streams.

anuraaga: This SGTM

huntc: I've now pushed a commit with updated documentation for the exporters, and the reinstatement of backpressure handling for the `IntervalMetricReader` and `BatchSpanProcessor`.

BTW Is there an equivalent of `SimpleSpanProcessor` for metrics?

jkwatson: Metrics have a somewhat different pipeline structure. The closest equivalent is probably the `IntervalMetricReader` class for push-based exporters. Pull-based, of course, manage all of that themselves.

huntc:
> (Hope @bogdandrutu can chime in real soon ;) )

FYI All that remains is for me to write a test for the `CompletableResultCode`, then this PR should be done. I hope that @anuraaga 's contribution here with the async OTLP exporter is also illuminating.

exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcMetricExporter.java
```diff
@@ -89,26 +97,29 @@ private OtlpGrpcMetricExporter(ManagedChannel channel, long deadlineMs) {
...
-      stub.export(exportMetricsServiceRequest);
-      return ResultCode.SUCCESS;
-    } catch (Throwable e) {
-      logger.log(Level.WARNING, "Failed to export metrics", e);
-      return ResultCode.FAILURE;
-    }
+          @Override
+          public void onFailure(Throwable t) {
+            logger.log(Level.WARNING, "Failed to export spans", t);
+            result.fail();
```
jkwatson, line 118: This does make me think we should consider adding an optional Throwable/message to the `fail()` method.

anuraaga: I think there's some trade-off (a higher chance if double logging) bug either seems fine for me, including following up in another PR since it's not an API break.

Currently this PR seems to have been blocked by 

https://github.com/open-telemetry/opentelemetry-java/pull/1428#issuecomment-667714059

with no real action item.

huntc:
> Please don't work (much) on this more until we have reached an agreement - I know now it's all excitement, but I'm very wary of sexy changes that later we may need to discard 😞

@carlosalberto Can we please have some resolution on this? I'll happily complete the PR over the weekend, but not if it isn't destined for greatness. :-)

jkwatson:
I'll put this on the agenda for tomorrow's SIG meeting, so it doesn't get lost.

huntc:
> I'll put this on the agenda for tomorrow's SIG meeting, so it doesn't get lost.
 
Thanks. Any resolution?

jkwatson:
> > I'll put this on the agenda for tomorrow's SIG meeting, so it doesn't get lost.
> 
> Thanks. Any resolution?

The decision (which is not a resolution, unfortunately) is that Carlos is going to find some time for a deeper dive into this PR. Meanwhile, the discussion on the original issue (#1422) has some further thoughts from other folks (@Oberon00 in particular)

anuraaga:
@carlosalberto Is it possible to get a commitment to that deep dive over the next few days? I understand you're busy, but this has had many weeks for a closer look and I don't think it's fair to keep it stuck.

carlosalberto:
Let's go with it. 👍

jkwatson:
@huntc We're a go for this. Can you get it rebased? Thanks so much for your patience. You work is appreciated, and sorry that it took a long time to get this landed.

exporters/logging/src/main/java/io/opentelemetry/exporters/logging/LoggingMetricExporter.java
```diff
@@ -41,16 +42,16 @@ public ResultCode export(Collection<MetricData> metrics) {
...
-  public ResultCode flush() {
-    ResultCode resultCode = ResultCode.SUCCESS;
+  public CompletableResultCode flush() {
+    CompletableResultCode resultCode = new CompletableResultCode();
     for (Handler handler : logger.getHandlers()) {
       try {
         handler.flush();
       } catch (Throwable t) {
-        resultCode = ResultCode.FAILURE;
+        resultCode.fail();
```
anuraaga, line 51: Should we early return here? Similar comment elsewhere

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
```diff
@@ -0,0 +1,109 @@
...
+ */
+
+package io.opentelemetry.sdk.common.export;
+
+import java.util.ArrayList;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.GuardedBy;
+
+/**
+ * The implementation of Export operations are often asynchronous in nature, hence the need to
```
anuraaga, line 24: This could be second sentence, first sentences should describe the class. e.g. A result which will be completed at a later time asynchronously.

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
```diff
@@ -0,0 +1,109 @@
...
+ * convey a result at a later time. CompletableResultCode facilitates this.
+ *
+ * <p>This class models JDK 8's CompletableFuture to afford migration should Open Telemetry's SDK
+ * select JDK 8 or greater as a baseline, and also to offer familiarity to developers.
+ */
+public class CompletableResultCode {
+  /** A convenience for declaring success. */
```
anuraaga, line 31: `Returns a {@link CompletableResultCode} that has been completed successfully.`

Ditto for below with `unsuccessfully`

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
```diff
@@ -0,0 +1,109 @@
...
+  @Nullable
+  @GuardedBy("lock")
+  private Boolean succeeded = null;
+
+  @GuardedBy("lock")
+  private final ArrayList<Runnable> completionActions = new ArrayList<>();
+
+  private final Object lock = new Object();
+
+  /** The export operation finished successfully. */
```
anuraaga, line 55: `Completed this {@link CompletableResultCode} successfully.`

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
anuraaga: Realized we can probably just copy paste javadoc from `CompletableFuture`

sdk/src/main/java/io/opentelemetry/sdk/trace/export/SimpleSpanProcessor.java
```diff
@@ -71,15 +74,25 @@ public boolean isStartRequired() {
   }
 
   @Override
+  @SuppressWarnings("BooleanParameter")
```
jkwatson, line 77: I don't think this is needed on this method.

exporters/logging/src/main/java/io/opentelemetry/exporters/logging/LoggingMetricExporter.java
jkwatson: I think it doesn't matter in this particular case, since we might want to flush all the handlers, even if one fails. The call to succeed() below is a no-op if it's been previously failed.

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
```diff
@@ -0,0 +1,109 @@
...
+  public CompletableResultCode succeed() {
+    synchronized (lock) {
+      if (succeeded == null) {
```
jkwatson, line 58: This behavior should be documented, since I think it's non-obvious that it's first-in-wins with succeed/fail.

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
```diff
@@ -0,0 +1,109 @@
...
+  private static final CompletableResultCode FAILURE = new CompletableResultCode().fail();
+
+  public CompletableResultCode() {}
+
+  @Nullable
+  @GuardedBy("lock")
+  private Boolean succeeded = null;
+
+  @GuardedBy("lock")
+  private final ArrayList<Runnable> completionActions = new ArrayList<>();
```
jkwatson, line 51: This can be declared as a simple `List` on the LHS, correct?

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
jkwatson: Also, unit tests to verify it would also be very good.

jkwatson:
I have just a couple of minor issues, mostly just that we should document the first-in-wins behavior of the result code, and write unit tests to verify that behavior.

exporters/logging/src/main/java/io/opentelemetry/exporters/logging/LoggingMetricExporter.java
huntc: I'll add a return to make it clear but, as John mentioned, it makes no diff.

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
huntc: Sentences swapped, although my personal preference is to lead with the rationale as the type name is good enough.

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
huntc: Done

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
huntc: Done

sdk/src/main/java/io/opentelemetry/sdk/trace/export/SimpleSpanProcessor.java
huntc: Hangover from the past. Removed.

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
huntc: Done. New test added also.

sdk/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
huntc: Done

huntc:
> I have just a couple of minor issues, mostly just that we should document the first-in-wins behavior of the result code, and write unit tests to verify that behavior.

PR feedback addressed and is in as a separate commit.

sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
iNikem: Unfortunately, this change from "scheduleAtFixedDelay" to "scheduleAtFixedRate" results in dropped spans. `onBatchExport` can be called in a loop from `exportBatches` method. As this method now returns immediately, calling it in a tight loop guarantees data loss.

huntc: > Unfortunately, this change from "scheduleAtFixedDelay" to "scheduleAtFixedRate" results in dropped spans. `onBatchExport` can be called in a loop from `exportBatches` method. As this method now returns immediately, calling it in a tight loop guarantees data loss.

I think you're right - for situations where the `spansList` cannot be drained fast enough. I shall attend to this. Meanwhile, here's a new ticket: https://github.com/open-telemetry/opentelemetry-java/issues/1566 to capture the problem. It isn't a terrible situation as `BatchSpanProcessor` usage has always dictated that spans can be dropped. However, we can do better.

Issue #1430, opened 2020-07-17
Span.recordException


iNikem:
Implement newly added API method: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#record-exception

Ideally, while solving this issue also add [corresponding semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/exceptions.md) to `SemanticAttributes` class

anuraaga:
I can take this on

PR #1432, opened 2020-07-19
Reuse char[] when constructing trace context header for injection.

anuraaga:
For #1430 I will probably add a similar reusable `StringBuilder` for reducing overhead of recording exceptions. First I figured we should apply it to our existing place to understand the implications.

When constructing strings, it is almost always worth using a reused buffer which avoids zeroing the array and extra garbage generation, and this works well since these buffers do not get exposed outside of the method.

Main concern is putting this in `internal`, we should reduce the number of these `ThreadLocal` that we have in an app since there's no point at all in having multiple, but with `internal` it's not supposed to be used outside, e.g., in the SDK or in propagation extensions. But in `common` it seems to expose it to the public. Does it make sense to allow this sort of class outside the API module? I can add the error prone `RestrictedApi` annotation to formalize that if it does. It's too bad the JDK will never add this sort of class since it's somewhat unsafe, and JDK never exposes unsafe APIs to users. But memory usage of apps would go down if there was one thread local to rule them all...

About 15% faster with less than half of the allocation.
After
```
Benchmark                                                                       Mode  Cnt    Score     Error   Units
HttpTraceContextInjectBenchmark.measureInject                                   avgt   15   70.617 ▒   7.677   ns/op
HttpTraceContextInjectBenchmark.measureInject:╖gc.alloc.rate                    avgt   15  868.646 ▒  82.041  MB/sec
HttpTraceContextInjectBenchmark.measureInject:╖gc.alloc.rate.norm               avgt   15   96.009 ▒   0.001    B/op
HttpTraceContextInjectBenchmark.measureInject:╖gc.churn.G1_Eden_Space           avgt   15  877.322 ▒ 109.523  MB/sec
HttpTraceContextInjectBenchmark.measureInject:╖gc.churn.G1_Eden_Space.norm      avgt   15   97.044 ▒   8.893    B/op
HttpTraceContextInjectBenchmark.measureInject:╖gc.churn.G1_Survivor_Space       avgt   15    0.004 ▒   0.003  MB/sec
HttpTraceContextInjectBenchmark.measureInject:╖gc.churn.G1_Survivor_Space.norm  avgt   15   ? 10?│              B/op
HttpTraceContextInjectBenchmark.measureInject:╖gc.count                         avgt   15   76.000            counts
HttpTraceContextInjectBenchmark.measureInject:╖gc.time                          avgt   15  123.000                ms
```

Before
```
Benchmark                                                                       Mode  Cnt     Score     Error   Units
HttpTraceContextInjectBenchmark.measureInject                                   avgt   15    84.904 ▒  12.105   ns/op
HttpTraceContextInjectBenchmark.measureInject:╖gc.alloc.rate                    avgt   15  1697.155 ▒ 213.918  MB/sec
HttpTraceContextInjectBenchmark.measureInject:╖gc.alloc.rate.norm               avgt   15   224.017 ▒   0.001    B/op
HttpTraceContextInjectBenchmark.measureInject:╖gc.churn.G1_Eden_Space           avgt   15  1692.461 ▒ 228.890  MB/sec
HttpTraceContextInjectBenchmark.measureInject:╖gc.churn.G1_Eden_Space.norm      avgt   15   223.377 ▒  10.128    B/op
HttpTraceContextInjectBenchmark.measureInject:╖gc.churn.G1_Survivor_Space       avgt   15     0.004 ▒   0.003  MB/sec
HttpTraceContextInjectBenchmark.measureInject:╖gc.churn.G1_Survivor_Space.norm  avgt   15     0.001 ▒   0.001    B/op
HttpTraceContextInjectBenchmark.measureInject:╖gc.count                         avgt   15   114.000            counts
HttpTraceContextInjectBenchmark.measureInject:╖gc.time                          avgt   15   205.000                ms
```


api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java
```diff
@@ -98,7 +99,7 @@
     spanContext.getSpanId().copyLowerBase16To(chars, SPAN_ID_OFFSET);
     chars[TRACE_OPTION_OFFSET - 1] = TRACEPARENT_DELIMITER;
     spanContext.getTraceFlags().copyLowerBase16To(chars, TRACE_OPTION_OFFSET);
-    setter.set(carrier, TRACE_PARENT, new String(chars));
+    setter.set(carrier, TRACE_PARENT, new String(chars, 0, TRACEPARENT_HEADER_SIZE));
```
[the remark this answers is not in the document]
anuraaga: `String` constructor always copies input arrays so it should be ok. The extra copy is why recyclable buffers is a popular pattern.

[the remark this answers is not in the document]
bogdandrutu: Makes sense, didn't see the extra copy but makes sense.

iNikem:
Is there a place in the repo for benchmarks as well?

anuraaga:
Yeah they're in the JMH folders. This one had an existing one so it's not in this PR

https://github.com/open-telemetry/opentelemetry-java/blob/master/api/src/jmh/java/io/opentelemetry/trace/propagation/HttpTraceContextInjectBenchmark.java

api/src/main/java/io/opentelemetry/internal/TemporaryBuffers.java
```diff
@@ -0,0 +1,54 @@
...
+package io.opentelemetry.internal;
+
+/**
+ * {@link ThreadLocal} buffers for use when creating new derived objects such as {@link String}s.
+ * These buffers are reused within a single thread - it is _not safe_ to use the buffer to generate
+ * multiple derived objects at the same time because the same memory will be used. In general, you
+ * should get a temporary buffer, fill it with data, and finish by converting into the derived
+ * object within the same method to avoid multiple usages of the same buffer.
+ */
+public final class TemporaryBuffers {
```
jkwatson, line 26: I'm not in love with this name, but it's fine for now. :)

api/src/main/java/io/opentelemetry/internal/TemporaryBuffers.java
```diff
@@ -0,0 +1,54 @@
...
+
+  private static final ThreadLocal<char[]> CHAR_ARRAY = new ThreadLocal<>();
+
+  /**
+   * A {@link ThreadLocal} {@code char[]} of size {@code len}. Take care when using a large value of
+   * {@code len} as this buffer will remain for the lifetime of the thread. The returned buffer will
+   * not be zeroed and may be larger than the requested size, you must make sure to fill the entire
+   * content to the desired value and set the length explicitly when converting to a {@link String}.
+   */
+  public static char[] chars(int len) {
```
jkwatson, line 36: Any thoughts to an optional boolean to zero-out the buffer before handing it out?

anuraaga: I think in general that will remove the majority of the performance benefit. The intention of the temporary buffers is to only use in cases where the filling is complete or controlled. If we come up with a use case for the Boolean though let's add later

jkwatson: fair enough. I was just thinking that in cases where the zeroing out was required, this class could do it more optimally in the initial request, since it knows it's a new char[] being created. No need for it now, though!

PR #1433, opened 2020-07-19
Remove null checks in Span API.

anuraaga:
I was randomly looking through the `Span` API and was surprised to see null checks in the no-op span for sure but also in the real span. It seems to not follow the spec with regards to error handling. "API methods MUST NOT throw unhandled exceptions when used incorrectly by end users. "

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/error-handling.md#basic-error-handling-principles

### DIFF
--- a/sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
@@ -323,16 +323,25 @@ final class RecordEventsReadableSpan implements ReadableSpan, Span {
 
   @Override
   public void addEvent(String name) {
+    if (name == null) {
+      return;
+    }
     addTimedEvent(TimedEvent.create(clock.now(), name, Attributes.empty(), 0));
   }
 
   @Override
   public void addEvent(String name, long timestamp) {
+    if (name == null) {
+      return;
+    }
     addTimedEvent(TimedEvent.create(timestamp, name, Attributes.empty(), 0));
   }
 
   @Override
   public void addEvent(String name, Attributes attributes) {
+    if (name == null) {
+      return;
+    }
     int totalAttributeCount = attributes.size();
     addTimedEvent(
         TimedEvent.create(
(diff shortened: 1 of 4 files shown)


sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -300,7 +299,9 @@ public void setAttribute(String key, boolean value) {
 
   @Override
   public void setAttribute(String key, AttributeValue value) {
-    Preconditions.checkNotNull(key, "key");
+    if (key == null) {
+      return;
```
thisthat, line 303: Before returning, I would add a log to notify the error since the spec requires that key MUST not be null https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#set-attributes
Furthermore, I just noticed that we allow empty strings as a key

thisthat: I have created #1436

anuraaga: I have used a property controlled pattern for validating before, would it make sense here?

https://github.com/aws/aws-xray-sdk-java/blob/master/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/utils/LooseValidations.java#L47

I don't think we should log unconditionally since in many apps log spam can be almost as bad as runtime errors.

thisthat: I think your suggestion is very valuable used in combination with #1419

sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -385,7 +386,9 @@ private void addTimedEvent(TimedEvent timedEvent) {
 
   @Override
   public void setStatus(Status status) {
-    Preconditions.checkNotNull(status, "status");
+    if (status == null) {
```
carlosalberto, line 389: We need new tests for all the SDK changes too.

carlosalberto:
Overall LGTM. Thanks for working on this - this was something that was meant to be done at some point, and now I see we didn't create an issue for tracking this ;(

anuraaga:
Thanks, added tests

PR #1434, opened 2020-07-20
Implement Span.recordException

### LINKED ISSUE #243: Collecting stacktraces on error
When an error occurs, we should capture the full stacktrace on the span, not just the message.  When trying to troubleshoot an error, the full stacktrace is far more useful in providing context than just the error message.

(See comment [here](https://github.com/open-telemetry/opentelemetry-java/pull/229#discussion_r280252988))

### LINKED ISSUE #1430: Span.recordException

anuraaga:
`recordException` accepts a `Throwable` and converts it into the exception semantic attributes. 

Fixes #1430 
Fixes #243

Benchmark result. The custom writer seems to be about 10% faster than Guava + JDK.
```
PrintThrowableBenchmark.customPrintWriter                                   thrpt   60    211.488 ▒    8.383  ops/ms
PrintThrowableBenchmark.customPrintWriter:╖gc.alloc.rate                    thrpt   60   2629.492 ▒  105.163  MB/sec
PrintThrowableBenchmark.customPrintWriter:╖gc.alloc.rate.norm               thrpt   60  19569.316 ▒    0.060    B/op
PrintThrowableBenchmark.customPrintWriter:╖gc.churn.G1_Eden_Space           thrpt   60   2634.753 ▒  113.643  MB/sec
PrintThrowableBenchmark.customPrintWriter:╖gc.churn.G1_Eden_Space.norm      thrpt   60  19610.326 ▒  333.335    B/op
PrintThrowableBenchmark.customPrintWriter:╖gc.churn.G1_Survivor_Space       thrpt   60      0.007 ▒    0.001  MB/sec
PrintThrowableBenchmark.customPrintWriter:╖gc.churn.G1_Survivor_Space.norm  thrpt   60      0.051 ▒    0.008    B/op
PrintThrowableBenchmark.customPrintWriter:╖gc.count                         thrpt   60    649.000             counts
PrintThrowableBenchmark.customPrintWriter:╖gc.time                          thrpt   60   1156.000                 ms
PrintThrowableBenchmark.normalPrintWriter                                   thrpt   60    195.298 ▒   11.995  ops/ms
PrintThrowableBenchmark.normalPrintWriter:╖gc.alloc.rate                    thrpt   60   2431.664 ▒  150.037  MB/sec
PrintThrowableBenchmark.normalPrintWriter:╖gc.alloc.rate.norm               thrpt   60  19601.265 ▒    0.051    B/op
PrintThrowableBenchmark.normalPrintWriter:╖gc.churn.G1_Eden_Space           thrpt   60   2443.735 ▒  161.806  MB/sec
PrintThrowableBenchmark.normalPrintWriter:╖gc.churn.G1_Eden_Space.norm      thrpt   60  19674.502 ▒  332.010    B/op
PrintThrowableBenchmark.normalPrintWriter:╖gc.churn.G1_Survivor_Space       thrpt   60      0.007 ▒    0.001  MB/sec
PrintThrowableBenchmark.normalPrintWriter:╖gc.churn.G1_Survivor_Space.norm  thrpt   60      0.054 ▒    0.009    B/op
PrintThrowableBenchmark.normalPrintWriter:╖gc.count                         thrpt   60    576.000             counts
PrintThrowableBenchmark.normalPrintWriter:╖gc.time                          thrpt   60   1035.000                 ms
PrintThrowableBenchmark.printStream                                         thrpt   60     73.729 ▒    2.933  ops/ms
PrintThrowableBenchmark.printStream:╖gc.alloc.rate                          thrpt   60   2119.750 ▒   85.085  MB/sec
PrintThrowableBenchmark.printStream:╖gc.alloc.rate.norm                     thrpt   60  45275.242 ▒    0.148    B/op
PrintThrowableBenchmark.printStream:╖gc.churn.G1_Eden_Space                 thrpt   60   2136.836 ▒   95.629  MB/sec
PrintThrowableBenchmark.printStream:╖gc.churn.G1_Eden_Space.norm            thrpt   60  45657.127 ▒ 1084.172    B/op
PrintThrowableBenchmark.printStream:╖gc.churn.G1_Survivor_Space             thrpt   60      0.060 ▒    0.006  MB/sec
PrintThrowableBenchmark.printStream:╖gc.churn.G1_Survivor_Space.norm        thrpt   60      1.273 ▒    0.114    B/op
PrintThrowableBenchmark.printStream:╖gc.count                               thrpt   60    554.000             counts
PrintThrowableBenchmark.printStream:╖gc.time                                thrpt   60    966.000                 ms
PrintThrowableBenchmark.stringWriter                                        thrpt   60    199.465 ▒   10.671  ops/ms
PrintThrowableBenchmark.stringWriter:╖gc.alloc.rate                         thrpt   60   2464.810 ▒  132.275  MB/sec
PrintThrowableBenchmark.stringWriter:╖gc.alloc.rate.norm                    thrpt   60  19457.324 ▒    0.022    B/op
PrintThrowableBenchmark.stringWriter:╖gc.churn.G1_Eden_Space                thrpt   60   2467.116 ▒  137.296  MB/sec
PrintThrowableBenchmark.stringWriter:╖gc.churn.G1_Eden_Space.norm           thrpt   60  19474.909 ▒  303.665    B/op
PrintThrowableBenchmark.stringWriter:╖gc.churn.G1_Survivor_Space            thrpt   60      0.008 ▒    0.001  MB/sec
PrintThrowableBenchmark.stringWriter:╖gc.churn.G1_Survivor_Space.norm       thrpt   60      0.060 ▒    0.009    B/op
PrintThrowableBenchmark.stringWriter:╖gc.count                              thrpt   60    615.000             counts
PrintThrowableBenchmark.stringWriter:╖gc.time                               thrpt   60   1084.000                 ms
```


iNikem:
Have you tried this?
```
    ByteArrayOutputStream stream = new ByteArrayOutputStream();
    new Throwable().printStackTrace(new PrintStream(stream));
    stream.toString();
```

Then we don't need to implement our own wrapper...

anuraaga:
Sure why not. It matches my expectation of slowness :) Probably because it UTF-8 encodes / decodes during the print.

`PrintWriter` hasn't had an interface change since Java 1.5 so I'm fairly comfortable with the wrapper, it's well tested and conceivably no real maintenance cost. But the improvement isn't huge so I'm ok with sticking with the Guava approach if there's concern about it.

iNikem:
Interesting... Thanks for benchmark :)

iNikem:
@anuraaga But throwing NPEs is still wrong by the spec, isn't it? I think your initial version was correct and independent from #1433

anuraaga:
I agree about NPEs, but I think either all the methods in the class should throw or none of them. I've changed this to be consistent with master for now since I (optimistically) think this will land first. I'll address all the checks together in #1433. Does that seem ok?

sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -395,6 +396,29 @@ public void setStatus(Status status) {
...
+    SemanticAttributes.EXCEPTION_TYPE.set(attributes, exception.getClass().getCanonicalName());
+    if (exception.getMessage() != null) {
+      SemanticAttributes.EXCEPTION_MESSAGE.set(attributes, exception.getMessage());
+    }
+    StringBuilder buffer = new StringBuilder();
+    StringBuilderPrintWriter writer = new StringBuilderPrintWriter(buffer);
+    exception.printStackTrace(writer);
+    SemanticAttributes.EXCEPTION_STACKTRACE.set(attributes, buffer.toString());
+
+    addEvent("exception", attributes.build(), timestamp);
```
jkwatson, line 419: Should we have a semantic "attribute" constant for this event name, as well? Seems like this event name is special, and canonical, so we should probably put it in the API somewhere.

jkwatson:
Unless I'm reading the benchmarks wrong, the StringWriter and the custom PrintWriter implementations have the same throughput performance, within the error bars. And, the StringWriter has slightly lower allocation rates. 

With that, I'd rather just use the built-in StringWriter, rather than having to maintain a customer PrintWriter implementation.

What version of the JDK/JVM did you run the benchmarks on?

anuraaga:
Ah thanks for noticing I didn't pay enough attention to `StringWriter` because of my plan in a future PR to recycle the buffer. I don't think that's possible for a `StringWriter` (no ability to reset). But I should probably stick to stringwriter in this PR and switch together with that then.

anuraaga:
Thanks, removed the custom writer (for now, muahaha)

sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
anuraaga: Added a constant

sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -416,6 +419,31 @@ public void setStatus(Status status) {
...
+    Attributes.Builder attributes = Attributes.newBuilder();
+    SemanticAttributes.EXCEPTION_TYPE.set(attributes, exception.getClass().getCanonicalName());
+    if (exception.getMessage() != null) {
+      SemanticAttributes.EXCEPTION_MESSAGE.set(attributes, exception.getMessage());
+    }
+    StringWriter writer = new StringWriter();
+    exception.printStackTrace(new PrintWriter(writer));
+    SemanticAttributes.EXCEPTION_STACKTRACE.set(attributes, writer.toString());
+
+    addEvent(SemanticAttributes.EXCEPTION_EVENT_NAME, attributes.build(), timestamp);
```
jkwatson, line 444: 👍

anuraaga:
@carlosalberto @bogdandrutu Could you take a look at this PR? Thanks!

api/src/main/java/io/opentelemetry/trace/Span.java
```diff
@@ -217,6 +217,29 @@
...
+   * <p>Use this method to specify an explicit event timestamp. If not called, the implementation
+   * will use the current timestamp value, which should be the default case.
+   *
+   * <p>Important: this is NOT equivalent with System.nanoTime().
+   *
+   * @param exception the {@link Throwable} to record.
+   * @param timestamp the explicit event timestamp in nanos since epoch.
+   * @since 0.7.0
+   */
+  void recordException(Throwable exception, long timestamp);
```
bogdandrutu, line 241: Would not add this method unless is necessary. Removing methods is harder than adding. I don't see a scenario when an exception happens and somehow later want to record it.

api/src/main/java/io/opentelemetry/trace/attributes/BooleanAttributeSetter.java
```diff
@@ -70,4 +71,14 @@ public void set(Span span, boolean value) {
...
     spanBuilder.setAttribute(key(), value);
   }
+
+  /**
+   * Sets the attribute on the provided {@link Attributes.Builder}.
+   *
+   * @param attributesBuilder the attributes builder to add the attribute to
+   * @param value the value for this attribute
+   */
+  public void set(Attributes.Builder attributesBuilder, boolean value) {
```
bogdandrutu, line 81: Is this change unrelated?

api/src/main/java/io/opentelemetry/trace/attributes/DoubleAttributeSetter.java
```diff
@@ -70,4 +71,14 @@ public void set(Span span, double value) {
...
     spanBuilder.setAttribute(key(), value);
   }
+
+  /**
+   * Sets the attribute on the provided {@link Attributes.Builder}.
+   *
+   * @param attributesBuilder the attributes builder to add the attribute to
+   * @param value the value for this attribute
+   */
+  public void set(Attributes.Builder attributesBuilder, double value) {
```
bogdandrutu, line 81: Unrelated change?

bogdandrutu:
@anuraaga I see in the description that you also added the other methods. Please try to split the PR, thanks.

api/src/main/java/io/opentelemetry/trace/Span.java
anuraaga: I would argue anything with a timestamp needs to have a way to manually set it, because users may have their own clocks, or some similar reason where timing is kept track in their framework and they want to use it. I can remove this method but then I'd argue it's a pre-GA task to reconsider modeling exceptions as events since it means we don't consider the timestamp to be meaningful. @iNikem any thoughts?

bogdandrutu: Actually user using their own clock is the main reason I am against this :). Reason is we do a good job using nanoTime to ensure ordering of the events, but user does not have access to our internal clock so any time recorded by external user will not be order with the other events.

anuraaga: I don't think this is correct in a general sense, many frameworks record timings at a level too deep for us to capture as accurately, at least without bytecode instrumentation. For example a "wire send" event, if it's recorded just before passing a buffer to netty will probably be many nanoseconds more accurate than what our instrumentation could pull off. Note that in case, user is interchangable with instrumentation, the latter also needs to be able to pass on the accurate timings. There is some concern about ordering issues with non-framework generated events, but I'd take the more accurate timings for things like DNS resolution, etc any day.

But I agree this seems far less useful for exceptions. So it makes me wonder whether we really need a timestamp for exceptions in the first place if, similar to a DNS query, we may miss out on a better timing provided by the framework if we choose not to allow that here.

In the meantime, I've removed the method though as it seems like a bigger discussion about why exception is modeled as Event instead of something without a timestamp.

iNikem: I took a look at [API defined by spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#add-events). 

```
An Event is defined by the following properties:

(Required) Name of the event.
(Optional) One or more Attributes with the same restrictions as defined for Span Attributes.
(Optional) Timestamp for the event.
```

Several other API methods accept timestamp as well. Thus this method is totally within spec. Although I too have doubts if anyone will use it.

anuraaga: Thanks for the link to the spec! I didn't check it, just looked at our Java API. I feel our current API doesn't reflect the optional aspect of the Timestamp (at least the javadoc needs to be more explicit about using current time when not provided and pointing at a way to create an event without timestamps). In that case, I'm good with this API I'll investigate this optional-timestamp event case more generally in a separate PR.

anuraaga: Filed #1447 to track that

anuraaga: Ah realized my comment was ambiguous - I meant I'm good with not including the method version that includes a timestamp. As part of #1447 I'll probably switch this to add an event without any timestamp.

iNikem:
@bogdandrutu Do you approve this PR now? Can it be merged?

carlosalberto:
Looks good to me. @bogdandrutu confirm/merge you are satisfied with the changes.

bogdandrutu:
@iNikem is the specs merged? Need to catch a bit with the status of the specs before making a decision

anuraaga:
Yeah they are merged I think these specs?

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#record-exception

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#add-events

PR #1437, opened 2020-07-20
Update Attributes API to comply the spec 

thisthat:
Update attributes API to the latest spec: https://github.com/open-telemetry/opentelemetry-specification/pull/777

### DIFF
--- a/api/src/main/java/io/opentelemetry/common/AttributeValue.java
+++ b/api/src/main/java/io/opentelemetry/common/AttributeValue.java
@@ -341,12 +341,13 @@ public abstract class AttributeValue {
   @AutoValue
   abstract static class AttributeValueStringArray extends AttributeValue {
 
+    private static final List<String> EMPTY = Collections.<String>emptyList();
+
     AttributeValueStringArray() {}
 
     static AttributeValue create(String... stringValues) {
       if (stringValues == null) {
-        return new AutoValue_AttributeValue_AttributeValueStringArray(
-            Collections.<String>emptyList());
+        return new AutoValue_AttributeValue_AttributeValueStringArray(EMPTY);
       }
       return new AutoValue_AttributeValue_AttributeValueStringArray(
           Collections.unmodifiableList(Arrays.asList(stringValues)));
@@ -359,7 +360,7 @@ public abstract class AttributeValue {
 
     @Override
     public boolean isEmpty() {
-      return getStringArrayValue().isEmpty();
+      return getStringArrayValue() == EMPTY;
     }
 
     @Override
@@ -370,12 +371,13 @@ public abstract class AttributeValue {
   @AutoValue
   abstract static class AttributeValueBooleanArray extends AttributeValue {
 
+    private static final List<Boolean> EMPTY = Collections.<Boolean>emptyList();
+
     AttributeValueBooleanArray() {}
 
     static AttributeValue create(Boolean... booleanValues) {
       if (booleanValues == null) {
-        return new AutoValue_AttributeValue_AttributeValueBooleanArray(
-            Collections.<Boolean>emptyList());
+        return new AutoValue_AttributeValue_AttributeValueBooleanArray(EMPTY);
       }
       List<Boolean> values = new ArrayList<>(booleanValues.length);
       values.addAll(Arrays.asList(booleanValues));
@@ -390,7 +392,7 @@ public abstract class AttributeValue {
 
     @Override
     public boolean isEmpty() {
-      return getBooleanArrayValue().isEmpty();
+      return getBooleanArrayValue() == EMPTY;
     }
 
     @Override
@@ -401,11 +403,13 @@ public abstract class AttributeValue {
   @AutoValue
   abstract static class AttributeValueLongArray extends AttributeValue {
 
+    private static final List<Long> EMPTY = Collections.<Long>emptyList();
+
     AttributeValueLongArray() {}
 
     static AttributeValue create(Long... longValues) {
       if (longValues == null) {
-        return new AutoValue_AttributeValue_AttributeValueLongArray(Collections.<Long>emptyList());
+        return new AutoValue_AttributeValue_AttributeValueLongArray(EMPTY);
       }
       List<Long> values = new ArrayList<>(longValues.length);
       values.addAll(Arrays.asList(longValues));
@@ -420,7 +424,7 @@ public abstract class AttributeValue {
 
     @Override
     public boolean isEmpty() {
-      return getLongArrayValue().isEmpty();
+      return getLongArrayValue() == EMPTY;
     }
 
     @Override
@@ -431,12 +435,13 @@ public abstract class AttributeValue {
   @AutoValue
   abstract static class AttributeValueDoubleArray extends AttributeValue {
 
+    private static final List<Double> EMPTY = Collections.<Double>emptyList();
+
     AttributeValueDoubleArray() {}
 
     static AttributeValue create(Double... doubleValues) {
       if (doubleValues == null) {
-        return new AutoValue_AttributeValue_AttributeValueDoubleArray(
-            Collections.<Double>emptyList());
+        return new AutoValue_AttributeValue_AttributeValueDoubleArray(EMPTY);
       }
       List<Double> values = new ArrayList<>(doubleValues.length);
       values.addAll(Arrays.asList(doubleValues));
@@ -451,7 +456,7 @@ public abstract class AttributeValue {
 
     @Override
     public boolean isEmpty() {
-      return getDoubleArrayValue().isEmpty();
+      return getDoubleArrayValue() == EMPTY;
     }
 
     @Override
--- a/api/src/main/java/io/opentelemetry/common/AttributeValue.java
+++ b/api/src/main/java/io/opentelemetry/common/AttributeValue.java
@@ -251,7 +251,7 @@ public abstract class AttributeValue {
    * @return {@code true} if the {@code AttributeValue} is empty.
    * @since 0.8.0
    */
-  public boolean isEmpty() {
+  public boolean isNull() {
     return false;
   }
 
@@ -271,7 +271,7 @@ public abstract class AttributeValue {
     }
 
     @Override
-    public boolean isEmpty() {
+    public boolean isNull() {
       return getStringValue() == null;
     }
 
@@ -341,13 +341,14 @@ public abstract class AttributeValue {
   @AutoValue
   abstract static class AttributeValueStringArray extends AttributeValue {
 
-    private static final List<String> EMPTY = Collections.<String>emptyList();
+    private static final AttributeValue EMPTY =
+        new AutoValue_AttributeValue_AttributeValueStringArray(Collections.<String>emptyList());
 
     AttributeValueStringArray() {}
 
     static AttributeValue create(String... stringValues) {
       if (stringValues == null) {
-        return new AutoValue_AttributeValue_AttributeValueStringArray(EMPTY);
+        return EMPTY;
       }
       return new AutoValue_AttributeValue_AttributeValueStringArray(
           Collections.unmodifiableList(Arrays.asList(stringValues)));
@@ -359,8 +360,8 @@ public abstract class AttributeValue {
     }
 
     @Override
-    public boolean isEmpty() {
-      return getStringArrayValue() == EMPTY;
+    public boolean isNull() {
+      return this == EMPTY;
     }
 
     @Override
@@ -371,13 +372,14 @@ public abstract class AttributeValue {
   @AutoValue
   abstract static class AttributeValueBooleanArray extends AttributeValue {
 
-    private static final List<Boolean> EMPTY = Collections.<Boolean>emptyList();
+    private static final AttributeValue EMPTY =
+        new AutoValue_AttributeValue_AttributeValueBooleanArray(Collections.<Boolean>emptyList());
 
     AttributeValueBooleanArray() {}
 
     static AttributeValue create(Boolean... booleanValues) {
       if (booleanValues == null) {
-        return new AutoValue_AttributeValue_AttributeValueBooleanArray(EMPTY);
+        return EMPTY;
       }
       List<Boolean> values = new ArrayList<>(booleanValues.length);
       values.addAll(Arrays.asList(booleanValues));
@@ -391,8 +393,8 @@ public abstract class AttributeValue {
     }
 
     @Override
-    public boolean isEmpty() {
-      return getBooleanArrayValue() == EMPTY;
+    public boolean isNull() {
+      return this == EMPTY;
     }
 
     @Override
@@ -403,13 +405,14 @@ public abstract class AttributeValue {
   @AutoValue
   abstract static class AttributeValueLongArray extends AttributeValue {
 
-    private static final List<Long> EMPTY = Collections.<Long>emptyList();
+    private static final AttributeValue EMPTY =
+        new AutoValue_AttributeValue_AttributeValueLongArray(Collections.<Long>emptyList());
 
     AttributeValueLongArray() {}
 
     static AttributeValue create(Long... longValues) {
       if (longValues == null) {
-        return new AutoValue_AttributeValue_AttributeValueLongArray(EMPTY);
+        return EMPTY;
       }
       List<Long> values = new ArrayList<>(longValues.length);
       values.addAll(Arrays.asList(longValues));
@@ -423,8 +426,8 @@ public abstract class AttributeValue {
     }
 
     @Override
-    public boolean isEmpty() {
-      return getLongArrayValue() == EMPTY;
+    public boolean isNull() {
+      return this == EMPTY;
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


sdk/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -114,7 +114,31 @@ public static Resource getTelemetrySdk() {
...
+  private static final class NullRemover implements KeyValueConsumer<AttributeValue> {
+    private final Attributes.Builder attrBuilder;
+
+    private NullRemover(Attributes.Builder attrBuilder) {
+      this.attrBuilder = attrBuilder;
+    }
+
+    @Override
+    public void consume(String key, AttributeValue value) {
+      switch (value.getType()) {
```
iNikem, line 131: Simple `if` would be much simpler :)

sdk/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -114,7 +114,31 @@ public static Resource getTelemetrySdk() {
    */
   public static Resource create(Attributes attributes) {
     checkAttributes(Objects.requireNonNull(attributes, "attributes"));
-    return new AutoValue_Resource(attributes);
+    Attributes.Builder attrBuilder = Attributes.newBuilder();
+    attributes.forEach(new NullRemover(attrBuilder));
+    return new AutoValue_Resource(attrBuilder.build());
+  }
+
+  private static final class NullRemover implements KeyValueConsumer<AttributeValue> {
```
jkwatson, line 122: How about "NullValueFilter" ?

jkwatson:
> Shouldn't we just not set `null` in `Attributes` instead of filtering on consumption?
> 
> "Attribute values of null are considered to be not set and get discarded as if that SetAttribute call had never been made"

Since anyone can create a Resource, even outside of this codebase, we wouldn't be able to control that, so having the filter in place seems like a good idea. But, yes, "don't do that" is always a good rule when it comes to nulls. ;)

anuraaga:
@jkwatson Sorry wasn't clear, I meant filtering out null values when creating attributes, e.g. the builder's setters

https://github.com/open-telemetry/opentelemetry-java/blob/master/api/src/main/java/io/opentelemetry/common/Attributes.java#L148

thisthat:
@anuraaga `null` values are meaningful because they are used to delete previously set attributes.

anuraaga:
Hmm if we use null as an implementation detail of emptiness it's ok. In that case shouldn't `forEach` internally filter null? Having all users of `forEach` filter like this seems weird

anuraaga:
On second thought, do we have an API for deleting attributes?

thisthat:
Yes we do have it https://github.com/open-telemetry/opentelemetry-java/blob/master/sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java#L314
However, looking again at the spec, [Events](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#add-events) and [Links](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#add-links) attributes should also offer the possibility to delete existing attributes.

jkwatson:
> Yes we do have it https://github.com/open-telemetry/opentelemetry-java/blob/master/sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java#L314
> However, looking again at the spec, [Events](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#add-events) and [Links](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#add-links) attributes should also offer the possibility to delete existing attributes.


I don't think events and links need to support this. they are created immutable (or lazily so), so you won't be removing attributes from existing links or events.

jkwatson:
Note: the way that attributes are deleted from spans is pretty hacky at the moment (it involves using an extension of HashMap). I'd love to figure out a better way to do it.

sdk/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -114,7 +114,25 @@ public static Resource getTelemetrySdk() {
...
+
+    private NullValueFilter(Attributes.Builder attrBuilder) {
+      this.attrBuilder = attrBuilder;
+    }
+
+    @Override
+    public void consume(String key, AttributeValue value) {
+      if (value.getType() == AttributeValue.Type.STRING && value.getStringValue() == null) {
+        return;
+      }
```
bogdandrutu, line 133: Why other values like null array is not removed?

anuraaga:
So assuming we do need to keep null values to model deletion (seems a bit weird if we support deletion we could have an explicit deletion API? Java Maps have the exact problem @jkwatson mentions of supporting this in a very undefined way), still I wonder why we don't filter in the producer rather than all the consumers, here

https://github.com/open-telemetry/opentelemetry-java/blob/master/api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java#L58

sdk/src/main/java/io/opentelemetry/sdk/resources/Resource.java
anuraaga: I think along these lines, if we're using `null` to model deletion, we should just store `null` directly, not an attribute value with a value of `null`.

thisthat: Because null values inside arrays are meaningful. Since the `Attributes` API does not create `null` attributes, it cannot exist a null value but only a value that contains an `AttributeValue` with null. Does it make sense?

thisthat:
@anuraaga filtering in the producer makes sense only if we do have an explicit API to delete Span attributes and we would need to change the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#set-attributes).

anuraaga:
> @anuraaga filtering in the producer makes sense only if we do have an explicit API to delete Span attributes 

Really? We'd still be setting the attribute with `null`, meaning it's deleted. And if we filter in the producer, it won't be returned to the consumer, fulfilling the requirement of acting as if `SetAttribute` call had never been made. What am I missing? :)

thisthat:
I was missing something :sweat_smile: The `ImmutableKeyValuePairs` is not used to set Span attributes.

sdk/src/main/java/io/opentelemetry/sdk/resources/Resource.java
bogdandrutu: No, it does not make sense. What if the "array" is null, not the value inside the array?

thisthat: The "array" cannot be null. `Resource` uses the `Attributes` class for storing attributes which creates always an object for arrays: https://github.com/open-telemetry/opentelemetry-java/blob/master/api/src/main/java/io/opentelemetry/common/Attributes.java#L204

bogdandrutu: Very strange behavior for arrays then:
https://github.com/open-telemetry/opentelemetry-java/blob/78ed6490ba47568f5241604ea34f18d3ae6f3c4a/api/src/main/java/io/opentelemetry/common/AttributeValue.java#L331

We replace with empty array/list.... Not too much consistency.

jkwatson: General Java guidelines are to never return null for a collection, but use an empty one (per Josh Bloch in Effective Java). So, I think this behavior is a good one.

bogdandrutu: We discussed about this. It is not we return null, it is we get null and add an entry with empty collection, why then not add an entry with empty string?

thisthat: I have uniformed the handling of `null` in 18e3719. The following cases are dropped (or removed if the key was already present):
- null strings, e.g. `(String) null`
- new arrays with 0 length, e.g. `new String[0]`
- null arrays, e.g. `(String[]) null`

Instead, the following are stored as attributes:
- Empty strings
- Arrays with null values as members, e.g. `new String[] {null}`

sdk/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -116,7 +117,45 @@ public static Resource getTelemetrySdk() {
...
+
+    @Override
+    public void consume(String key, AttributeValue value) {
+      boolean ignore = false;
+      switch (value.getType()) {
+        case STRING:
+          ignore = value.getStringValue() == null;
+          break;
+        case LONG_ARRAY:
+          ignore = value.getLongArrayValue().equals(Collections.<Long>emptyList());
```
jkwatson, line 140: It might be easier just to call `isEmpty()` rather than comparing explicitly to an empty list. And, this makes me think we could have an `isEmpty()` method on the AttributeValue, and then you can skip all of the switch branches and have the logic much more simple.

sdk/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -116,7 +116,25 @@ public static Resource getTelemetrySdk() {
    */
   public static Resource create(Attributes attributes) {
     checkAttributes(Objects.requireNonNull(attributes, "attributes"));
-    return new AutoValue_Resource(attributes);
+    Attributes.Builder attrBuilder = Attributes.newBuilder();
+    attributes.forEach(new NullValueFilter(attrBuilder));
+    return new AutoValue_Resource(attrBuilder.build());
+  }
+
+  private static final class NullValueFilter implements KeyValueConsumer<AttributeValue> {
```
anuraaga, line 124: I thought we got on the same page but see this didn't change yet - can't we filter in the producer here?

https://github.com/open-telemetry/opentelemetry-java/blob/master/api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java#L58

thisthat: In the last SIG meeting, we agree on keeping the same behavior for all the interfaces that can create attributes.

anuraaga: I may finally understand, is it that here

https://github.com/open-telemetry/opentelemetry-specification/blob/b4e327c3fbc62be811036cac572e4cc0d57dd560/specification/trace/api.md#set-attributes

where we discuss `Attribute values of null are considered to be not set and get discarded as if that SetAttribute call had never been made.`, this only refers to Spans? But Attributes are more general and could permit null values in other cases? It seems very confusing that we would support `null` attributes (not talking about as an implementation detail for modeling deletion for example but as actually exported value), but if that reading is correct than I agree we wouldn't filter in the producer, and may need to reclarify the spec.

api/src/main/java/io/opentelemetry/common/AttributeValue.java
```diff
@@ -244,6 +244,15 @@ public double getDoubleValue() {
...
   public abstract Type getType();
 
+  /**
+   * Returns {@code true} if the {@code AttributeValue} contains a {@code null} value or it is an
+   * empty array.
+   *
+   * @return {@code true} if the {@code AttributeValue} is empty.
+   * @since 0.8.0
+   */
+  public abstract boolean isEmpty();
```
jkwatson, line 254: we could just have this be concrete and return `false` and only override it for the types that support emptiness.

jkwatson:
I like this change (to add the `isEmpty` on the attribute value for this use). But, it is an API change, so I would like @carlosalberto and @bogdandrutu to chime on that part of this change.

Oberon00:
I haven't been following the full conversation, but if this about discarding empty arrays, I'm against it. We also don't discard empty strings.

See spec https://github.com/open-telemetry/opentelemetry-specification/blob/v0.6.0/specification/trace/api.md#set-attributes:

> Attribute values expressing a numerical value of zero or an empty string are considered meaningful and MUST be stored and passed on to span processors / exporters.

The same should hold for empty arrays IMHO.

jkwatson:
> I haven't been following the full conversation, but if this about discarding empty arrays, I'm against it. We also don't discard empty strings.
> 
> See spec https://github.com/open-telemetry/opentelemetry-specification/blob/v0.6.0/specification/trace/api.md#set-attributes:
> 
> > Attribute values expressing a numerical value of zero or an empty string are considered meaningful and MUST be stored and passed on to span processors / exporters.
> 
> The same should hold for empty arrays IMHO.

This is only for the Resource, not for Span attributes. @thisthat can you update the PR title to reflect that it's just for Resource attributes?

anuraaga:
Thanks for bearing with my confusion!

thisthat:
@open-telemetry/java-approvers I have update the PR to implement the latest specification introduced by https://github.com/open-telemetry/opentelemetry-specification/pull/777 
Resource (`Attributes` API) and Span attributes have the same behavior:
- accept empty arrays
- accept null values within arrays
- drop null or empty keys
- drop null values or delete the attribute if was previously set

api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -155,12 +155,32 @@ public Attributes build() {
       return sortAndFilterToAttributes(data.toArray());
     }
 
+    private boolean checkKeyValue(String key, AttributeValue value) {
```
anuraaga, line 158: How about `shouldAdd` and return false when it shouldn't? Having trouble following this method

api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -155,12 +155,32 @@ public Attributes build() {
...
+      if (key == null || key.length() == 0) {
+        return true;
+      }
+      if (value == null || value.isEmpty()) {
+        int index = data.indexOf(key);
+        if (index == -1) {
+          return true;
+        }
+        // Remove key/value pair
+        data.remove(index);
```
anuraaga, line 168: This is pretty slow but maybe we don't need to worry.

api/src/main/java/io/opentelemetry/common/Attributes.java
thisthat: I would suggest calling it `isInvalidKeyValue`. This way, we don't need to negate its results in the `setAttribute` methods.

anuraaga: doNotAdd could be ok too, though I don't know if I personally find it more readable than a negation. But the main point is to not have key in the method name since it's not only checking the key but doing something special based on the value as well. Just describing the return value (whether to add or not) seems fine to me.

thisthat: done in ed949be

api/src/main/java/io/opentelemetry/common/Attributes.java
thisthat: You are right. However, I don't expect the removal of attributes to be a frequent operation.

api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -155,12 +155,32 @@ public Attributes build() {
...
+        data.remove(index);
```
jkwatson, line 169: Have this method, which looks like a simple query method, also have the side effect of removing data is very surprising. The name of this method should reflect that it is potentially an update operation. Also, how does this handle the fact that the key might be in the data twice at this point? Don't we need to iterate through the entire array to remove all instances of the key?

api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -172,8 +192,12 @@ public Builder setAttribute(String key, AttributeValue value) {
      * @return this Builder
      */
     public Builder setAttribute(String key, String value) {
+      AttributeValue v = stringAttributeValue(value);
+      if (doNotAdd(key, v)) {
```
jkwatson, line 196: rather than repeating this logic over and over again, can we just call the setAttribute method with the attribute value above?

api/src/main/java/io/opentelemetry/common/AttributeValue.java
```diff
@@ -325,12 +341,13 @@ public final Type getType() {
...
 
+    private static final List<String> EMPTY = Collections.<String>emptyList();
+
     AttributeValueStringArray() {}
 
     static AttributeValue create(String... stringValues) {
       if (stringValues == null) {
-        return new AutoValue_AttributeValue_AttributeValueStringArray(
-            Collections.<String>emptyList());
+        return new AutoValue_AttributeValue_AttributeValueStringArray(EMPTY);
```
jkwatson, line 350: could we just return a static final AttributeValue instance, rather than creating a new one for each call?

sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -314,11 +312,10 @@ public void setAttribute(String key, AttributeValue value) {
         logger.log(Level.FINE, "Calling setAttribute() on an ended Span.");
         return;
       }
-      if (value == null || (value.getType().equals(STRING) && value.getStringValue() == null)) {
-        if (attributes == null) {
-          return;
+      if (value == null || value.isEmpty()) {
```
jkwatson, line 315: interesting. now that empty arrays are explicitly allowed, I wonder if "isEmpty" is no longer the right name. Maybe `isMissing` ?

jkwatson: or maybe just `isNull()` since that's literally what we're checking for now.

api/src/main/java/io/opentelemetry/common/AttributeValue.java
thisthat: I've overlooked a simple thing like that :sweat_smile:

api/src/main/java/io/opentelemetry/common/Attributes.java
thisthat: You are right! I have overlooked that dedup is done as the last thing in the builder.

api/src/main/java/io/opentelemetry/common/AttributeValue.java
jkwatson: hey, that's why we review each others code. :). Easy to miss stuff like this when you're in the weeds.

PR #1440, opened 2020-07-21
Forbid empty/null keys

### LINKED ISSUE #1436: Empty strings are wrongly accepted as keys for attributes
**Describe the bug**
It is possible to set an attribute with an empty string, which is forbidden by the specifications https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#set-attributes

thisthat:
Fixes #1436

### DIFF
--- a/api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
+++ b/api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
@@ -21,7 +21,6 @@ import static io.opentelemetry.internal.Utils.checkArgument;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import java.util.logging.Logger;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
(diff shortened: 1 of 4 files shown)


api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
```diff
@@ -105,8 +106,8 @@ private static void quickSort(Object[] data, int leftIndex, int rightIndex) {
     for (int i = 0; i < data.length; i += 2) {
       Object key = data[i];
       Object value = data[i + 1];
-      if (key == null) {
-        logger.warning("Ignoring null key.");
+      if (key == null || "".equals(key)) {
+        logger.warning("Ignoring null/empty key.");
```
jkwatson, line 110: logging here at warning seems like a recipe for a lot of log spam. I'm not sure if we should log this at all, but `warning` definitely seems too aggressive for something in the API.

jkwatson:
Can you add some javadoc to the top of Attributes and Labels to make it clear that null/empty keys will be ignored?

api/src/main/java/io/opentelemetry/common/ImmutableKeyValuePairs.java
thisthat: I also find it strange that we have a warning here. I don't think it adds anything, I will remove it

PR #1444, opened 2020-07-22
Add ParentOrElse Sampler

MitchellDumovic:
Introduce a new kind of composite sampler - a `ParentOrElseSampler` - which is specified in the spec document [here](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#built-in-samplers). A `ParentOrElseSampler` is initialized with a sampler delegate and makes the same sampling decision as the parent. If there is no parent, it uses the provided delegate to make a decision. 

There is a separate question as to whether or not this sampler (with an AlwaysOn delegate) should be used as the default Sampler, which would fix the "bug" outlined in https://github.com/open-telemetry/opentelemetry-java/issues/1445

### DIFF
--- a/sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
@@ -122,15 +122,17 @@ public final class Samplers {
 
   /**
    * Returns a {@link Sampler} that always makes the same decision as the parent {@link Span} to
-   * whether or not to sample. If there is no parent, always makes a "yes" decision on {@code Span}
-   * sampling.
+   * whether or not to sample. If there is no parent, the Sampler uses the provided Sampler delegate to
+   * determine the sampling decision.
    *
-   * @return a {@code Sampler} a sampler that follows its parent span's sampling decisions if the
-   *     parent exists.
+   * @param delegateSampler the {@code Sampler} which is used to make the sampling decisions if the
+   * parent does not exist.
+   * @return a {@code Sampler} that follows the parent's sampling decision if one exists, otherwise
+   * following the delegate sampler's decision.
    * @since 0.7.0
    */
-  public static Sampler followParent() {
-    return FollowParentSampler.INSTANCE;
+  public static Sampler parentOrElse(Sampler delegateSampler) {
+    return new ParentOrElse(delegateSampler);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


bogdandrutu:
@MitchellDumovic see https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#built-in-samplers the AlwaysOn does what specs suggest

sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
```diff
@@ -176,6 +189,32 @@ public String getDescription() {
     }
   }
 
+  @Immutable
+  private enum FollowParentSampler implements Sampler {
+    INSTANCE;
+
+    // If a parent is set, always follows the same sampling decision as the parent.
+    // Otherwise, always makes a "yes" decision on {@link Span} sampling.
```
anuraaga, line 197: It seems like follow parent should be a setting on any sampler, or we should support multiple samplers where any sampler can return something like `NO_DECISION` to allow another sampler to pick it up. Always yes seems to be unrelated to following the parent, right?

anuraaga: Oops just jumped through bogdan's link, yeah shouldn't we use parentorelse instead of adding this sampler?

I would argue ParentOrElse(AlwaysOn) should be the default for the SDK though

sdk/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
```diff
@@ -82,8 +82,7 @@
 @Immutable
 public abstract class TraceConfig {
   // These values are the default values for all the global parameters.
-  // TODO: decide which default sampler to use
-  private static final Sampler DEFAULT_SAMPLER = Samplers.alwaysOn();
+  private static final Sampler DEFAULT_SAMPLER = Samplers.parentOrElse(Samplers.alwaysOn());
```
bogdandrutu, line 85: I know I am annoying, but can we do this in a followup PR? This is an important change and would like to make sure everyone aggress on, just adding the parentElse is trivial to be merged.

sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
bogdandrutu: @anuraaga can we argue in a separate PR :)

sdk/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
MitchellDumovic: That's totally fine! I hadn't realized initially that the default was already set in the spec - I can also create an issue in the spec repo to discuss this instead of doing it in a followup PR here if that makes more sense?

tedsuo:
A change like this should be standardized and recorded in the spec, correct?

bogdandrutu:
@tedsuo which change? This adds parentOrElse which is already defined in specs, that's what I commented to remove the default change

tedsuo:
Sorry, that was the bit I was referring to, that's for splitting it out.

sdk/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
MitchellDumovic: I added an issue to the spec repo to discuss this: https://github.com/open-telemetry/opentelemetry-specification/issues/728 cc @anuraaga 

One thing that I admit to being confused about is how the sampler is meant to interact with the `TraceFlags`, which is slightly inconsistent across languages. If anyone can clear up some of the inconsistencies for me that would be helpful.

PR #1446, opened 2020-07-22
Add version of attribute setters that operate on Attributes.

anuraaga:
They're `AttributeSetter`s so should work on `Attributes` ;) And comes in handy when adding semantic attributes to events and resources.

### DIFF
--- a/api/src/main/java/io/opentelemetry/trace/attributes/BooleanAttributeSetter.java
+++ b/api/src/main/java/io/opentelemetry/trace/attributes/BooleanAttributeSetter.java
@@ -16,6 +16,7 @@
 
 package io.opentelemetry.trace.attributes;
 
+import io.opentelemetry.common.Attributes;
 import io.opentelemetry.trace.Span;
 import javax.annotation.concurrent.Immutable;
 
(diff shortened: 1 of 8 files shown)


api/src/main/java/io/opentelemetry/trace/attributes/SemanticAttributes.java
```diff
@@ -209,5 +209,20 @@
...
+  /** The exception message. */
+  public static final StringAttributeSetter EXCEPTION_MESSAGE =
+      StringAttributeSetter.create("exception.message");
+  /**
+   * A string representing the stacktrace of an exception, as produced by {@link
+   * Throwable#printStackTrace()}.
+   */
+  public static final StringAttributeSetter EXCEPTION_STACKTRACE =
+      StringAttributeSetter.create("exception.stacktrace");
+
```
bogdandrutu, line 226: This is needed in the other PR :)

anuraaga: Oops thanks

bogdandrutu:
This is split from #1432 and already approved. Merging

PR #1448, opened 2020-07-22
Allow Resource to be updated in TracerSdkProvider.

carlosalberto:
This is a small change to allow `Resource` to be updated at any time in `TracerSdkProvider` - we have been discussing doing a more general (single) initialization, but this change makes sense IMHO as we already allow the user to update the tracer config info. For reference - Python also allows this to be updated.

Opinions on this?


bogdandrutu:
The specs says that "Resource" is immutable after the start of the app, TraceConfig doesn't have to be immutable.

carlosalberto:
Bah, had totally missed the part (just confirmed it).

Closing this approach ;(

Issue #962, opened 2020-03-04
Implement SamplingResult according to the spec 


SergeyKanzhelev:
I was looking for which SDKs implemented the Sampling spec as written: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-tracing.md#sampling

It feels like the most of SDKs still implements the previous version with the `Decision` class.

ptravers:
I would like to pick up this issue. I believe the issue needs to be updated to instead reference:
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#sampling

jkwatson:
assigned!

PR #1449, opened 2020-07-22
Refactor decision to sampling result

ptravers:
Fixes https://github.com/open-telemetry/opentelemetry-java/issues/962

### DIFF
--- a/sdk/src/main/java/io/opentelemetry/sdk/trace/Sampler.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/trace/Sampler.java
@@ -86,7 +86,7 @@ public interface Sampler {
   interface SamplingResult {
 
     /**
-     * Return sampling result whether span should be sampled, recorded, both or neither.
+     * Return decision on whether span should be sampled, recorded, both or neither.
      *
      * @return sampling result.
      * @since 0.7.0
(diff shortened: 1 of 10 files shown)


sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
```diff
@@ -259,39 +279,44 @@ public final String getDescription() {
...
+    static SamplingResult create(Decision decision, Attributes attributes) {
+      return new AutoValue_Samplers_SamplingResultImpl(decision, attributes);
     }
 
     @Override
-    public abstract boolean isSampled();
+    public abstract Decision getDecision();
+
+    @Override
+    public boolean isSampled() {
```
bogdandrutu, line 317: Should we deprecate this? Maybe a followup PR, for the moment just create an issue to not forget about?

sdk/src/main/java/io/opentelemetry/sdk/trace/Sampler.java
```diff
@@ -68,27 +68,36 @@ Decision shouldSample(
...
+     * Return decision on whether a span should be recorded, recorded and sampled or not recorded.
      *
-     * @return sampling decision.
-     * @since 0.1.0
+     * @return sampling result.
+     * @since 0.7.0
      */
+    Decision getDecision();
+
     boolean isSampled();
```
bogdandrutu, line 94: Can we remove this?

ptravers: It's no longer required in the spec so it would seem like it's no longer required

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#sampler

PR #1457, opened 2020-07-23
B3: Add Additional Validity Checks and tests for Trace and Span ID

MitchellDumovic:
From the B3 Spec (https://github.com/openzipkin/b3-propagation) 

`Trace identifiers are 64 or 128-bit, but all span identifiers within a trace are 64-bit`

Identifiers must also be implemented as hex-characters. 

Technically these checks were happening before (and these tests would have passed), but an exception was raised elsewhere in the code. It felt more intuitive to move the checks into this validity checking function.

### DIFF
--- a/extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/B3PropagatorExtractor.java
+++ b/extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/B3PropagatorExtractor.java
@@ -18,6 +18,7 @@ package io.opentelemetry.extensions.trace.propagation;
 
 import static io.opentelemetry.extensions.trace.propagation.B3Propagator.MAX_SPAN_ID_LENGTH;
 import static io.opentelemetry.extensions.trace.propagation.B3Propagator.MAX_TRACE_ID_LENGTH;
+import static io.opentelemetry.extensions.trace.propagation.B3Propagator.MIN_TRACE_ID_LENGTH;
 import static io.opentelemetry.extensions.trace.propagation.B3Propagator.TRUE_INT;
 
 import io.grpc.Context;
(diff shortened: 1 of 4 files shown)


extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/B3PropagatorExtractor.java
```diff
@@ -63,12 +65,25 @@ static SpanContext buildSpanContext(String traceId, String spanId, String sample
       }
     }
 
+    private static boolean isHex(String value) {
+      try {
+        new BigInteger(value, 16);
```
iNikem, line 70: I am somewhat concerned about creating one extra object every time we want to check if traceId or spanId are valid

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/B3PropagatorExtractor.java
```diff
@@ -63,12 +65,25 @@ static SpanContext buildSpanContext(String traceId, String spanId, String sample
...
+      } catch (NumberFormatException e) {
+        return false;
+      }
+    }
+
     static boolean isTraceIdValid(String value) {
-      return !(StringUtils.isNullOrEmpty(value) || value.length() > MAX_TRACE_ID_LENGTH);
+      return !(StringUtils.isNullOrEmpty(value)
+          || !isHex(value)
+          || (value.length() != MIN_TRACE_ID_LENGTH && value.length() != MAX_TRACE_ID_LENGTH));
```
anuraaga, line 80: Let's check length before hex since it's a simpler check

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/B3PropagatorExtractor.java
anuraaga: Agree, looping through and checking the character range shouldn't be too much more complex.

carlosalberto:
Code looks great, thanks for the contribution!  LGTM when the suggestions are applied ;)

anuraaga:
Thanks for the change!

PR #1461, opened 2020-07-24
Fix Jaeger exporter system property config

### LINKED ISSUE #1408: Use the now standard ConfigBuilder for JaegerGrpcSpanExporter.Builder
Currently, `JaegerGrpcSpanExporter.Builder` uses the older form:

```
JaegerGrpcSpanExporter.Builder.fromEnv().build();
```

instead of the now standard ConfigBuilder, e.g. `ZipkinSpanExporter`:

```
ZipkinSpanExporter.newBuilder()
    .readEnvironmentVariables()
    .readSystemProperties()
    .build();
```

trask:
Currently, the Jaeger exporter uses system property names that look like env vars `JAEGER_ENDPOINT` / `JAEGER_SERVICE_NAME`, which has caused confusion for a couple of opentelemetry-javaagent users.

This fixes that by using the now standard `ConfigBuilder` for `JaegerGrpcSpanExporter.Builder` (and following the conventions from `ZipkinSpanExporter`).

Closes #1408

### DIFF
--- a/exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java
+++ b/exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java
@@ -184,8 +184,8 @@ public final class JaegerGrpcSpanExporter implements SpanExporter {
     }
 
     /**
-     * Sets the managed chanel to use when communicating with the backend. Required if {@link
-     * Builder#endpoint} is not set. If {@link Builder#endpoint} is set then build the channel.
+     * Sets the managed chanel to use when communicating with the backend. Takes precedence over
+     * {@link #setEndpoint(String)} if both are called.
      *
      * @param channel the channel to use.
      * @return this.
@@ -246,10 +246,8 @@ public final class JaegerGrpcSpanExporter implements SpanExporter {
      * @return a new exporter's instance.
      */
     public JaegerGrpcSpanExporter build() {
-      if (endpoint != null) {
+      if (channel == null) {
         channel = ManagedChannelBuilder.forTarget(endpoint).usePlaintext().build();
-      } else if (channel == null) {
-        throw new IllegalStateException("Must set either endpoint or channel");
       }
       return new JaegerGrpcSpanExporter(serviceName, channel, deadlineMs);
     }
--- a/exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java
+++ b/exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java
@@ -198,6 +198,8 @@ public final class JaegerGrpcSpanExporter implements SpanExporter {
     /**
      * Sets the Jaeger endpoint to connect to. Optional.
      *
+     * <p>Defaults to "localhost:14250".
+     *
      * @param endpoint The Jaeger endpoint URL, ex. "jaegerhost:14250".
      * @return this.
      * @since 0.7.0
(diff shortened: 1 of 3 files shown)


exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java
```diff
@@ -192,6 +193,19 @@ public Builder setChannel(ManagedChannel channel) {
...
+    /**
+     * Sets the jaeger endpoint. This will use the endpoint to assign a {@link ManagedChannel}
+     * instance to this builder.
+     *
+     * @param endpoint The Jaeger endpoint URL, ex. "jaegerhost:14250".
+     * @return this.
+     * @since 0.7.0
+     */
+    public Builder setEndpoint(String endpoint) {
+      setChannel(createChannel(endpoint));
```
bogdandrutu, line 205: I would not create the channel here. I would set the default endpoint in the ctor of the builder. Then in `build()` will check if the channel was set externally otherwise would create a new channel from "endpoint".

It feels to my more natural.

trask: Makes sense, done

exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java
```diff
@@ -182,7 +184,10 @@ public Builder setServiceName(String serviceName) {
     }
 
     /**
-     * Sets the managed chanel to use when communicating with the backend. Required.
+     * Sets the managed chanel to use when communicating with the backend.
+     *
+     * <p>Should use only one of {@link #setChannel(ManagedChannel)} or {@link
```
anuraaga, line 189: Does it make sense to validate this point in build?

exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java
```diff
@@ -223,27 +251,12 @@ public static Builder fromEnv() {
      * @return a new exporter's instance.
      */
     public JaegerGrpcSpanExporter build() {
+      if (channel == null) {
```
anuraaga, line 254: Think we should validate that endpoint is set to provide a better error message, guessing otherwise it's an NPE.

exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java
trask: I updated this to match the OTLP exporter builders behavior and javadoc

exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java
trask: oh, i just realized, endpoint has a default, so no need to validate here i think?

exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java
trask: changed it again, no longer quite matches OTLP exporter behavior/javadoc, but i think a bit better (because of the default on endpoint, it's easier to check if channel was set, and if not fall back to endpoint). i will update my OTLP exporter PR to match

Issue #1462, opened 2020-07-24
Support system property otel.resource.attributes corresponding to OTEL_RESOURCE_ATTRIBUTES


trask:
It looks like currently only way to set resource attributes is via env var `OTEL_RESOURCE_ATTRIBUTES`. Would be nice to also support system property `otel.resource.attributes` similar to other configuration properties which support both env var and system property configuration.

dengliming:
@trask Hi~ It looks good. I submit a PR. Can you review and share your opinion?

trask:
Closed by #1466. Thanks @dengliming!

PR #1466, opened 2020-07-25
Support system property otel.resource.attributes corresponding to OTEL_RESOURCE_ATTRIBUTES

dengliming:
for [https://github.com/open-telemetry/opentelemetry-java/issues/1462](https://github.com/open-telemetry/opentelemetry-java/issues/1462)

### DIFF
--- a/sdk/src/main/java/io/opentelemetry/sdk/resources/EnvVarResource.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/resources/EnvVarResource.java
@@ -18,6 +18,8 @@ package io.opentelemetry.sdk.resources;
 
 import com.google.common.annotations.VisibleForTesting;
 import io.opentelemetry.common.Attributes;
+import io.opentelemetry.sdk.common.export.ConfigBuilder;
+import java.util.Map;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.ThreadSafe;
 
@@ -29,12 +31,11 @@ import javax.annotation.concurrent.ThreadSafe;
  */
 @ThreadSafe
 public final class EnvVarResource {
-  private static final String OTEL_RESOURCE_ATTRIBUTES_ENV = "OTEL_RESOURCE_ATTRIBUTES";
   private static final String ATTRIBUTE_LIST_SPLITTER = ",";
   private static final String ATTRIBUTE_KEY_VALUE_SPLITTER = "=";
 
-  private static final Resource ENV_VAR_RESOURCE =
-      Resource.create(parseResourceAttributes(System.getenv(OTEL_RESOURCE_ATTRIBUTES_ENV)));
+  private static final Resource ENV_VAR_RESOURCE = Resource.create(parseResourceAttributes(
+      new Builder().readEnvironmentVariables().readSystemProperties().getEnvAttributes()));
 
   private EnvVarResource() {}
 
@@ -73,6 +74,37 @@ public final class EnvVarResource {
             keyValuePair[0].trim(), keyValuePair[1].trim().replaceAll("^\"|\"$", ""));
       }
       return attrBuilders.build();
+    }
+  }
+
+  /** Builder utility for this EnvVarResource. */
+  public static class Builder extends ConfigBuilder<Builder> {
+    private static final String OTEL_RESOURCE_ATTRIBUTES_ENV = "OTEL_RESOURCE_ATTRIBUTES";
+    private static final String OTEL_RESOURCE_ATTRIBUTES_KEY = "otel.resource.attributes";
+    private String envAttributes;
+
+    @Override
+    protected Builder fromConfigMap(Map<String, String> configMap,
+        NamingConvention namingConvention) {
+      String envAttributesValue = getStringProperty(OTEL_RESOURCE_ATTRIBUTES_ENV, configMap);
+      if (envAttributesValue != null) {
+        this.setEnvAttributes(envAttributesValue);
+      }
+
+      envAttributesValue = getStringProperty(OTEL_RESOURCE_ATTRIBUTES_KEY, configMap);
+      if (envAttributesValue != null) {
+        this.setEnvAttributes(envAttributesValue);
+      }
+      return this;
+    }
+
+    public String getEnvAttributes() {
+      return envAttributes;
+    }
+
+    public Builder setEnvAttributes(String envAttributes) {
+      this.envAttributes = envAttributes;
+      return this;
     }
   }
 }
--- a/sdk/src/main/java/io/opentelemetry/sdk/resources/EnvVarResource.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/resources/EnvVarResource.java
@@ -34,8 +34,10 @@ public final class EnvVarResource {
   private static final String ATTRIBUTE_LIST_SPLITTER = ",";
   private static final String ATTRIBUTE_KEY_VALUE_SPLITTER = "=";
 
-  private static final Resource ENV_VAR_RESOURCE = Resource.create(parseResourceAttributes(
-      new Builder().readEnvironmentVariables().readSystemProperties().getEnvAttributes()));
+  private static final Resource ENV_VAR_RESOURCE =
+      Resource.create(
+          parseResourceAttributes(
+              new Builder().readEnvironmentVariables().readSystemProperties().getEnvAttributes()));
 
   private EnvVarResource() {}
 
@@ -84,8 +86,8 @@ public final class EnvVarResource {
     private String envAttributes;
 
     @Override
-    protected Builder fromConfigMap(Map<String, String> configMap,
-        NamingConvention namingConvention) {
+    protected Builder fromConfigMap(
+        Map<String, String> configMap, NamingConvention namingConvention) {
       String envAttributesValue = getStringProperty(OTEL_RESOURCE_ATTRIBUTES_ENV, configMap);
       if (envAttributesValue != null) {
         this.setEnvAttributes(envAttributesValue);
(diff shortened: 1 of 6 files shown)


sdk/src/main/java/io/opentelemetry/sdk/resources/EnvVarResource.java
```diff
@@ -75,4 +78,35 @@ static Attributes parseResourceAttributes(@Nullable String rawEnvAttributes) {
       return attrBuilders.build();
     }
   }
+
+  /** Builder utility for this EnvVarResource. */
+  public static class Builder extends ConfigBuilder<Builder> {
+    private static final String OTEL_RESOURCE_ATTRIBUTES_ENV = "OTEL_RESOURCE_ATTRIBUTES";
```
trask, line 84: the nice thing about using `ConfigBuilder` is that you only need to handle system property style names, and it will automatically translate those to env var style names when reading env vars (via `NamingConvention`)
```suggestion
```

sdk/src/main/java/io/opentelemetry/sdk/resources/EnvVarResource.java
```diff
@@ -75,4 +78,35 @@ static Attributes parseResourceAttributes(@Nullable String rawEnvAttributes) {
...
+    private String envAttributes;
+
+    @Override
+    protected Builder fromConfigMap(
+        Map<String, String> configMap, NamingConvention namingConvention) {
+      String envAttributesValue = getStringProperty(OTEL_RESOURCE_ATTRIBUTES_ENV, configMap);
+      if (envAttributesValue != null) {
+        this.setEnvAttributes(envAttributesValue);
+      }
+
```
trask, line 95: see comment above about automatic env var name handling by `ConfigBuilder`
```suggestion
```

sdk/src/main/java/io/opentelemetry/sdk/resources/EnvVarResource.java
```diff
@@ -75,4 +78,35 @@ static Attributes parseResourceAttributes(@Nullable String rawEnvAttributes) {
...
+      envAttributesValue = getStringProperty(OTEL_RESOURCE_ATTRIBUTES_KEY, configMap);
+      if (envAttributesValue != null) {
+        this.setEnvAttributes(envAttributesValue);
+      }
+      return this;
+    }
+
+    public String getEnvAttributes() {
```
trask, line 103: I'd suggest renaming this to `build` and doing the parsing and returning `Resource` here
```suggestion
    public Resource build() {
```

sdk/src/main/java/io/opentelemetry/sdk/resources/EnvVarResource.java
```diff
@@ -75,4 +78,35 @@ static Attributes parseResourceAttributes(@Nullable String rawEnvAttributes) {
       return attrBuilders.build();
     }
   }
+
+  /** Builder utility for this EnvVarResource. */
+  public static class Builder extends ConfigBuilder<Builder> {
```
trask, line 83: I'd suggest making this `private` (or package-protected if you need `@VisibleForTesting`). it doesn't seem like public builder is needed at this time at lease
```suggestion
  private static class Builder extends ConfigBuilder<Builder> {
```

dengliming:
> thx for taking this issue @dengliming!
> 
> a couple of suggestions below

Thanks. Fixed at lastest commit.

sdk/src/main/java/io/opentelemetry/sdk/resources/EnvVarResource.java
```diff
@@ -18,6 +18,8 @@
 
 import com.google.common.annotations.VisibleForTesting;
 import io.opentelemetry.common.Attributes;
+import io.opentelemetry.sdk.common.export.ConfigBuilder;
```
carlosalberto, line 21: We most likely need to later update the code for `ConfigBuilder` to live right under `common`, instead of `common.export`, as this configuration pattern is useful in other scenarios (cc @thisthat )

sdk/src/test/java/io/opentelemetry/sdk/resources/EnvVarResourceTest.java
```diff
@@ -67,4 +68,15 @@ public void parseResourceAttributes_quotes() {
     Attributes result = EnvVarResource.parseResourceAttributes("value=\"foo\"");
     assertThat(result).isEqualTo(Attributes.of("value", stringAttributeValue("foo")));
   }
+
+  @BeforeClass
+  public static void init() {
+    System.setProperty("otel.resource.attributes", "value = foo");
```
carlosalberto, line 74: We should have this set right in the test itself (the rest of the tests don't need this part).

carlosalberto:
Overall good - I have minor doubts on whether we need to use the Config Builder pattern here, as we only rely on a single field, but making the Builder private should keep us safe for the time being.

sdk/src/test/java/io/opentelemetry/sdk/resources/EnvVarResourceTest.java
dengliming: Hi~ This is because I put this static final field. The following code：
```
private static final Resource ENV_VAR_RESOURCE =
      new Builder().readEnvironmentVariables().readSystemProperties().build();
```
(it may inited before)

anuraaga: Hmm, isn't that true if the class happens to be loaded before this test is run? It seems like a lucky coincidence that isn't currently happening in the build.

I think to make this test robust we should stick with calling `new Builder().readEnvironmentVariables().readSystemProperties().build();` in the test instead of relying on the static instance being loaded after the test.

Also, make sure you reset the value of the property after the test is complete.

sdk/src/main/java/io/opentelemetry/sdk/resources/EnvVarResource.java
```diff
@@ -29,12 +31,11 @@
  */
 @ThreadSafe
 public final class EnvVarResource {
```
iNikem, line 33: The name of the class seems to be incorrect now. At least `Var` part.

trask: Ah, now that it supports system properties also, got it (took me some time 😂)

carlosalberto: @iNikem Oh, indeed, I was meant to mention the rename as well.I suggest it being renamed to something like `EnvAutodetectResource` or `AutodetectResource`, and update its javadoc to mention that now we do env vars + system properties detection.

sdk/src/test/java/io/opentelemetry/sdk/resources/EnvVarResourceTest.java
dengliming: agree. fixed at lastest commit. Thanks.

sdk/src/main/java/io/opentelemetry/sdk/resources/EnvVarResource.java
dengliming: Agree. @iNikem @trask @thisthat WDYT?

sdk/src/test/java/io/opentelemetry/sdk/resources/EnvVarResourceTest.java
```diff
@@ -67,4 +67,15 @@ public void parseResourceAttributes_quotes() {
     Attributes result = EnvVarResource.parseResourceAttributes("value=\"foo\"");
     assertThat(result).isEqualTo(Attributes.of("value", stringAttributeValue("foo")));
   }
+
+  @Test
+  public void getResourceAttributes_properties() {
```
thisthat, line 72: Please, add also a test for environment variables. An example is available here: https://github.com/open-telemetry/opentelemetry-java/blob/46f0c5f0b7b5044029fb55b105bf00576665316b/sdk/src/test/java/io/opentelemetry/sdk/common/export/ConfigBuilderTest.java#L270

sdk/src/main/java/io/opentelemetry/sdk/resources/EnvVarResource.java
thisthat: I would prefer `EnvAutodetectResource` over `AutodetectResource` since we only look for a specific env var/sys property.

sdk/src/test/java/io/opentelemetry/sdk/resources/EnvVarResourceTest.java
dengliming: fixed at lastest commit.

thisthat: Thank you :)

sdk/src/main/java/io/opentelemetry/sdk/resources/EnvVarResource.java
dengliming: Already renamed class name EnvVarResource to EnvAutodetectResource and update doc.

dengliming:
> Could we make this package private and add a static factory in `Resource`, maybe `Resource.createDefault()`? That reduces the impact of the name.

I'm not sure I fully understand what you mean. Do you mean like that?
```
private static final Resource ENV_VAR_RESOURCE = createDefault();
public static Resource createDefault() {
  return new Builder().readEnvironmentVariables().readSystemProperties().build();
}

```
and the package change to private
`protected static class Builder extends ConfigBuilder<Builder> {`

anuraaga:
I think in `Resource.java` we can have a `createDefault` method that returns the result of the auto detect. In that case, we can keep `EnvAutoDetectResource` class itself package private. Hides that away as an implementation detail.

PR #1475, opened 2020-07-29
Use OkHttpSender by default in ZipkinSpanExporter

### LINKED ISSUE #1409: Use OkHttpSender by default in ZipkinSpanExporter
See https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/674#discussion_r453431290

> except for when the JAR size is extremely limited, Zipkin always recommends using OkHttp in production since performance is way better

And https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/375#issuecomment-630845065

> The most common http reporter used by Zipkin users is the OkHttp one, it is the default in spring-cloud-sleuth. The URLConnection one exists mostly in the case of classpath issues.

trask:
Unfortunately, this makes the `zipkin-sender-okhttp3` dependency required.

I'm not sure if it's possible to make `OkHttpSender` the default while also giving users the option to shed it from their dependencies if they need to use something small (`URLConnectionSender`).

Maybe it's not a good idea to have any default Sender, and instead require users to bring their own?

Would close #1409

### DIFF
--- a/exporters/zipkin/src/main/java/io/opentelemetry/exporters/zipkin/ZipkinSpanExporter.java
+++ b/exporters/zipkin/src/main/java/io/opentelemetry/exporters/zipkin/ZipkinSpanExporter.java
@@ -45,7 +45,7 @@ import zipkin2.Span;
 import zipkin2.codec.BytesEncoder;
 import zipkin2.codec.SpanBytesEncoder;
 import zipkin2.reporter.Sender;
-import zipkin2.reporter.urlconnection.URLConnectionSender;
+import zipkin2.reporter.okhttp3.OkHttpSender;
 
 /**
  * This class was based on the OpenCensus zipkin exporter code at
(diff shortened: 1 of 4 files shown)


trask:
> Thanks to Java's lazy classloading, I think it would be possible for a user to use `setSender` in an advanced case where they want to swap the sender in their build dependencies. Does that sound right?

👍 and seems reasonable for an advanced case.

Most people will definitely have a better experience with OkHttpSender as the default.

exporters/zipkin/src/main/java/io/opentelemetry/exporters/zipkin/ZipkinSpanExporter.java
```diff
@@ -380,6 +383,9 @@ protected Builder fromConfigMap(
      * @since 0.4.0
      */
     public ZipkinSpanExporter build() {
+      if (sender == null) {
+        sender = OkHttpSender.create(endpoint);
```
thisthat, line 387: I would prefer this to be inside the `setEndpoint()` method so we fail earlier in case of error

bogdandrutu: Haha, we commented that. It will be in the same statement anyway. I prefer this way to ensure that setEndpoint always is the second option (I know that you can still achieve it with your suggestion but makes code a bit unclear)

Issue #1472, opened 2020-07-27
Hide EnvAutodetectResource


carlosalberto:
Consider hiding `EnvAutodetectResource` (formerly `EnvVarResource`) and expose its functionality as a static method in `Resource` (such as `createDefault()` or `createFromEnv()`).

(Follow up from #1466 )

dengliming:
So we can use Resource.createDefault() instead of EnvAutodetectResource.getResource()?

anuraaga:
Yeah that's my idea. The less public classes, the more flexibility we have going forward. For example, we may change it to not detect or use env or something with no restrictions.

dengliming:
If so, the method EnvAutodetectResource.parseResourceAttributes and builder should stay on or just put all to Resource.java?

anuraaga:
As long as it's all package private, so hidden from user, having the implementation in a separate class is still fine :) Can prevent Resource from getting too long and it can have more focused tests.

dengliming:
@anuraaga I'm still a little confused. If i do like this. seems make no sense.
```
 public static Resource createDefault() {
    return new EnvAutodetectResource.Builder().readEnvironmentVariables().readSystemProperties().build();
  }
```

dengliming:
@anuraaga Could you please help to see if the modification below make sense?
in `Resource.java` add below method
```
public static Resource createDefault() {
    return new EnvAutodetectResource.Builder().readEnvironmentVariables().readSystemProperties().build();
  }
```
and keep class `EnvAutodetectResource  ` package protected
`final class EnvAutodetectResource {`

anuraaga:
@dengliming Yup that seems good to me, thanks!

jkwatson:
closing via #1477

PR #1477, opened 2020-07-29
Hide EnvAutodetectResource ro reduce public classes

dengliming:
for [https://github.com/open-telemetry/opentelemetry-java/issues/1472](https://github.com/open-telemetry/opentelemetry-java/issues/1472)  
@carlosalberto @anuraaga Please take a look.

### DIFF
--- a/sdk/src/main/java/io/opentelemetry/sdk/resources/Resource.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/resources/Resource.java
@@ -53,7 +53,7 @@ public abstract class Resource {
               .setAttribute("telemetry.sdk.language", "java")
               .setAttribute("telemetry.sdk.version", readVersion())
               .build());
-  private static final Resource DEFALUT_AUTODETECT_RESOURCE =
+  private static final Resource DEFAULT =
       new EnvAutodetectResource.Builder().readEnvironmentVariables().readSystemProperties().build();
 
   @Nullable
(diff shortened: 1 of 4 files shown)


sdk/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -117,6 +117,19 @@ public static Resource create(Attributes attributes) {
     return new AutoValue_Resource(attributes);
   }
 
+  /**
+   * Returns a {@link Resource}. This resource information is loaded from the
+   * OTEL_RESOURCE_ATTRIBUTES environment variable or otel.resource.attributes system properties.
+   *
+   * @return a {@code Resource}.
+   */
+  public static Resource createDefault() {
```
carlosalberto, line 126: I'd suggest creating this value only once, for simplicity and to keep the invariant `EnvAutodetectResource` has - also, for that case, rename it to `getDefault()` too.

dengliming: Thanks. Fixed at lastest commit.

sdk/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -53,6 +53,8 @@
               .setAttribute("telemetry.sdk.language", "java")
               .setAttribute("telemetry.sdk.version", readVersion())
               .build());
+  private static final Resource DEFALUT_AUTODETECT_RESOURCE =
```
carlosalberto, line 56: DEFAULT ;)

dengliming: done. thx.

PR #716, opened 2019-12-18
provide an API method to return an empty LabelSet

jkwatson:
resolves #713

PR #1480, opened 2020-07-29
Do simple logging for failed export.

carlosalberto:
We already log other situations (such as failing to do a clean shutdown), so it makes sense to add logging for failed export.


iNikem:
Will that solve #1480 ?

carlosalberto:
@iNikem You referenced this same PR ;)

iNikem:
Bah! I mean #1418

carlosalberto:
I think this should fix #1418 yes, mostly because we are already logging the other important failure case (failure to properly shutdown).

The rest of the errors do not seem unrecoverable (they are all parameters-handling related), so I think this simple PR is enough.

exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java
```diff
@@ -126,6 +126,7 @@ public ResultCode export(Collection<SpanData> spans) {
       stub.postSpans(request);
       return ResultCode.SUCCESS;
     } catch (Throwable e) {
+      logger.log(Level.SEVERE, "Failed to export spans", e);
```
anuraaga, line 129: We don't seem to use `SEVERE` anywhere else yet - spans can fail to export for transient reasons (backend is restarting), and observability is not critical to the app anyways, so I'd go with WARNING if not INFO.

iNikem: Please at least WARN :)

iNikem:
@anuraaga Taking into account https://github.com/open-telemetry/opentelemetry-specification/issues/716, I would say logging in each exporter makes more sense :)

anuraaga:
@iNikem Well I'm definitely ok with removing resultcode instead :) But it seems like one or the other makes sense, either exporter is left to their own ways or are controlled by the span processor.

exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java
carlosalberto: WARN works for me. Will update.

carlosalberto:
On top of the possibility of having the export result code being removed, there's the case of using a `MultiSpanProcessor` and having a specific backend fail (imagine that Jaeger exported its data fine, but Zipkin didn't). Also, in each processor's export step you have access to the actual `Exception`, which *could be* useful.

anuraaga:
`MultiSpanProcessor` is also a `SpanProcessor`, if we think that's the right layer to log exporter failures it could take care of it too (sorry wasn't clear with my reference, I linked to batch but simple, multi would also probably have a similar change if going with it).

> Also, in each processor's export step you have access to the actual Exception, which could be useful.

The exception could be propagated up too - it seems weird to return a failure status without the reason. Shouldn't we either leave everything to the exporter, or otherwise have it propagate everything? I personally think either is fine, but a mix seems strange, it's hard to know who has responsibility when it comes to exporting spans.

carlosalberto:
>  Shouldn't we either leave everything to the exporter

Going forward, we should 'normalize' things with Exception/Exporter Code being returned or being totally gone. For the time being, I suggest we keep these warnings and tune this part once we decide how to proceed (most likely after we come to an agreement on whether we want #716 or not).

PR #1481, opened 2020-07-30
Fix too noisy B3 propagator

iNikem:
If incoming request does not have a B3 header, then `B3PropagatorExtractorMultipleHeaders` used to print a message that `Invalid TraceId in B3 header`. This is very confusing for an end user and just misleading.

### DIFF
--- a/extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/B3PropagatorExtractorMultipleHeaders.java
+++ b/extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/B3PropagatorExtractorMultipleHeaders.java
@@ -49,8 +49,11 @@ final class B3PropagatorExtractorMultipleHeaders implements B3PropagatorExtracto
   private static <C> SpanContext getSpanContextFromMultipleHeaders(
       C carrier, HttpTextFormat.Getter<C> getter) {
     String traceId = getter.get(carrier, TRACE_ID_HEADER);
+    if (StringUtils.isNullOrEmpty(traceId)) {
+      return SpanContext.getInvalid();
+    }
     if (!Common.isTraceIdValid(traceId)) {
-      logger.info(
+      logger.warning(
           "Invalid TraceId in B3 header: " + traceId + "'. Returning INVALID span context.");
       return SpanContext.getInvalid();
     }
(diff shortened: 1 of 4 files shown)


extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/B3PropagatorExtractorMultipleHeaders.java
```diff
@@ -49,8 +49,11 @@
   private static <C> SpanContext getSpanContextFromMultipleHeaders(
       C carrier, HttpTextFormat.Getter<C> getter) {
     String traceId = getter.get(carrier, TRACE_ID_HEADER);
+    if (StringUtils.isNullOrEmpty(traceId)) {
+      return SpanContext.getInvalid();
+    }
     if (!Common.isTraceIdValid(traceId)) {
-      logger.info(
+      logger.warning(
```
trask, line 56: the other propagators log with `info`
```suggestion
      logger.info(
```

iNikem: Maybe we should change them? If there is a non-empty header, but its format is wrong, shouldn't this be WARN?

anuraaga: If upstream sends a bad header, I'd be unhappy to have warning logs spammed without much I can do about it. In general, I'd say tracing should avoid spamming logs as much as possible (even info is a bit questionable, warning much more).

iNikem: But without logs, how do you know about that bad header?

iNikem:
But it does not actually matter to me. I can change all such logs to debug, info or warn. As maintainers wish :)

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/B3PropagatorExtractorMultipleHeaders.java
anuraaga: I'd generally recommend metrics or using a lower than default level which can be enabled for tests.

Log spam does have reasonable potential to bring down an app and no one likes it when observability kills the app.

jkwatson: Yes, definitely. If an unknown upstream actor can creating an arbitrary amount of log spam just by setting some bad headers, that's a DOS attack waiting to happen.

iNikem: Changed all log levels to `fine`

PR #1482, opened 2020-07-30
Fix SpanWrapper snapshot being incomplete.

### LINKED ISSUE #1479: SpanWrapper (SpanData) does not snapshot the value of hasEnded.
**Describe the bug**
`io.opentelemetry.sdk.trace.SpanWrapper` does not store the value of `hasEnded`, updates to the originial Span will be inadvertedly reflected in the SpanData.

**Steps to reproduce**

```java
SpanData sd = someSpan.toSpanData();
assertThat(sd.hasEnded()).isFalse();
someSpan.end();
assertThat(sd.hasEnded()).isFalse(); // Fails!
```

No assertion failure -- `SpanData` should be immutable.

**What did you see instead?**
Assertion failure -- `SpanData` was modified.

Master @ 8005c7eddd456b91023c661e8e9a9e37f641c9bb

**Additional context**
This can cause trouble when working with un-ended spans and multiple threads.

Oberon00:
Fixes #1479: `io.opentelemetry.sdk.trace.SpanWrapper` does not store the value of `hasEnded`, `name` or `endEpochNanos`, so updates to the originial Span would be inadvertedly reflected in the SpanData.

Also adds an unit test to check some other immutability properties of SpanWrapper.

### DIFF
--- a/sdk/src/main/java/io/opentelemetry/sdk/trace/SpanWrapper.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/trace/SpanWrapper.java
@@ -33,9 +33,8 @@ import javax.annotation.concurrent.Immutable;
 /**
  * Immutable class that stores {@link SpanData} based on a {@link RecordEventsReadableSpan}.
  *
- * <p>This is a variant of the "flyweight" design pattern, to reduce overhead. We store a reference
- * to a mutable {@link RecordEventsReadableSpan} ({@code delegate}) where we use only the immutable
- * parts from and a copy of all the mutable parts.
+ * <p>This class stores a reference to a mutable {@link RecordEventsReadableSpan} ({@code delegate})
+ * which it uses only the immutable parts from, and a copy of all the mutable parts.
  *
  * <p>When adding a new field to {@link RecordEventsReadableSpan}, store a copy if and only if the
  * field is mutable in the {@link RecordEventsReadableSpan}. Otherwise retrieve it from the
--- a/sdk/src/main/java/io/opentelemetry/sdk/trace/SpanWrapper.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/trace/SpanWrapper.java
@@ -30,6 +30,17 @@ import io.opentelemetry.trace.TraceState;
 import java.util.List;
 import javax.annotation.concurrent.Immutable;
 
+/**
+ * Immutable class that stores {@link SpanData} based on a {@link RecordEventsReadableSpan}.
+ *
+ * <p>This is a variant of the "flyweight" design pattern, to reduce overhead. We store a reference
+ * to a mutable {@link RecordEventsReadableSpan} ({@code delegate}) where we use only the immutable
+ * parts from and a copy of all the mutable parts.
+ *
+ * <p>When adding a new field to {@link RecordEventsReadableSpan}, store a copy if and only if the
+ * field is mutable in the {@link RecordEventsReadableSpan}. Otherwise retrieve it from the
+ * referenced {@link RecordEventsReadableSpan}.
+ */
 @Immutable
 @AutoValue
 abstract class SpanWrapper implements SpanData {
--- a/sdk/src/main/java/io/opentelemetry/sdk/trace/SpanWrapper.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/trace/SpanWrapper.java
@@ -47,6 +47,12 @@ abstract class SpanWrapper implements SpanData {
 
   abstract Status status();
 
+  abstract String name();
+
+  abstract long endEpochNanos();
+
+  abstract boolean hasEnded();
+
   /**
    * Note: the collections that are passed into this creator method are assumed to be immutable to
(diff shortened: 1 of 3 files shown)


sdk/src/test/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpanTest.java
```diff
@@ -253,6 +253,32 @@ public void toSpanData_WithInitialAttributes() {
     assertThat(spanData.getTotalAttributeCount()).isEqualTo(attributes.size() + 1);
   }
 
+  @Test
+  public void toSpanData_snapshot() {
```
iNikem, line 257: Can you please provide more descriptive name for the test? What does it actually test?

Oberon00:
```suggestion
  public void toSpanData_spanDataDoesNotChangeWhenModifyingSpan() {
```

would that be better?

Oberon00: I also updated the PR description.

Oberon00: Improved in 9ac9a16.

sdk/src/main/java/io/opentelemetry/sdk/trace/SpanWrapper.java
```diff
@@ -58,9 +64,21 @@ static SpanWrapper create(
...
       int totalAttributeCount,
       int totalRecordedEvents,
-      Status status) {
+      Status status,
+      String name,
+      long endEpochNanos,
+      boolean hasEnded) {
     return new AutoValue_SpanWrapper(
-        delegate, links, events, attributes, totalAttributeCount, totalRecordedEvents, status);
+        delegate,
```
anuraaga, line 72: I think a small comment can help

`// We reduce overhead by using the delegate for all immutable fields. Take care when adding something new to this class when determining whether to delegate or not.`

Oberon00: I'm hesistant to add this in this bugfix PR. It might cause lots of discussion on how to word it. That aspect you want me to call out did not really change with this PR.

anuraaga: Well the main reason it came to mind is while reviewing this I had to make sure we didn't miss other fields. I think with this PR you put in the thought so documenting that is worthwhile. If this sort of small comment can cause a bikeshed we're in trouble :(

Oberon00: OK, I will add it 😄

Oberon00: Added a javadoc in cc79fdb.

anuraaga: Thanks, I really like it!

sdk/src/main/java/io/opentelemetry/sdk/trace/SpanWrapper.java
```diff
@@ -30,6 +30,17 @@
 import java.util.List;
 import javax.annotation.concurrent.Immutable;
 
+/**
+ * Immutable class that stores {@link SpanData} based on a {@link RecordEventsReadableSpan}.
+ *
+ * <p>This is a variant of the "flyweight" design pattern, to reduce overhead. We store a reference
```
jkwatson, line 36: I don't think this is actually a flyweight in the GoF sense, which is more like a multi-singleton. This is a pure delegate.

Oberon00:
@jkwatson According to [Wikipedia](https://en.wikipedia.org/wiki/Flyweight_pattern):

> A flyweight is an object that minimizes memory usage by sharing as much data as possible with other similar objects

Which this one fits. OK, having many different SpanWrappers for the same span is uncommon, but you can actually do that. I never associated flyweights with singletons, the Wikipedia article is the first time I have seen this. But if you think that association is common, I can remove the reference.

jkwatson:
> @jkwatson According to [Wikipedia](https://en.wikipedia.org/wiki/Flyweight_pattern):
> 
> > A flyweight is an object that minimizes memory usage by sharing as much data as possible with other similar objects
> 
> Which this one fits. OK, having many different SpanWrappers for the same span is uncommon, but you can actually do that. I never associated flyweights with singletons, the Wikipedia article is the first time I have seen this. But if you think that association is common, I can remove the reference.
> 
> EDIT: Removed the reference in [c0e0809](https://github.com/open-telemetry/opentelemetry-java/commit/c0e080925652c7228dfab8daea86f281ff2f74e7).

Not to get too far off into the weeds, but the GoF flyweight canonical example is having a word processor use the same object for every 'a' in a document, rather than having a separate copy for each instance of the letter.  :)

Issue #1452, opened 2020-07-23
Modernize unit tests


anuraaga:
I'd like to propose a couple of options for large refactoring of the unit tests. I'm committed to take on the grunt work for both of them if others are ok :)

### Migrate to JUnit5

JUnit5 has many nice features compared to JUnit4 such as native parameterized tests, repeated tests, and smarter extension API. It's the actively developed version and it'd be nice for us to take advantage of any improvements that come down the line.

In practice, the majority of this refactor will be changing the imports and removing `public` modifier and removing the `@RunWith(JUnit4.class)`, which we could do anyways since I believe that's a relic from Google's build system.

### Migrate to AssertJ

AssertJ is extremely popular in the Java ecosystem while Truth is mostly used by Google projects. While these numbers are likely not extremely precise, I think the multiple orders of magnitude difference is probably generally accurate in indicating the trend.

https://github.com/joel-costigliola/assertj-core/network/dependents
https://github.com/google/truth/network/dependents

I think it would improve our new-contributor experience if we switch from Truth to AssertJ since I'd wager far more Java developers are familiar with that based on these numbers.

In practice, the majority of this refactor will be mechanical, and often just changing imports since the APIs are pretty similar.

jkwatson:
I'm very much in favor of upgrading to junit5, especially since we can, since we updated our tests to use java 8. 100% agreed with removing the `@RunWith` annotations, as well, as they just add noise, IMO.

I have less of of an opinion on AssertJ vs. Truth. I've never used AssertJ and I hadn't used Truth before this project. :)

anuraaga:
I'll tackle the junit5 first - thanks for the vote :)

jkwatson:
@anuraaga I think we can close this now, yes?

Issue #529, opened 2019-09-12
Truth or AssertJ: Pick one


elharo:
Truth and AssertJ do basically the same thing. It's weird to depend on both. At least one, probably AssertJ, should be removed.

https://googlecloudplatform.github.io/cloud-opensource-java/JLBP-1.html

bogdandrutu:
Please go with Truth :) as a lot of the code is using truth :)

PR #1493, opened 2020-08-03
Migrate from Truth to AssertJ

anuraaga:
See #1452 for motivation, this should be a benefit for new contributor experience.

I used IntelliJ replace in path to replace imports and some obvious differences like `.isSameInstanceAs` -> `.isSameAs` and manually updated map assertions for `entry` (assertj is typesafe) and fixed failing list assertions to assert with any order, truth defaults to allowing any order and assertj defaults to exact order.

### DIFF
--- a/api/src/test/java/io/opentelemetry/trace/TraceStateTest.java
+++ b/api/src/test/java/io/opentelemetry/trace/TraceStateTest.java
@@ -16,7 +16,7 @@
 
 package io.opentelemetry.trace;
 
-import static com.google.common.truth.Truth.assertThat;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
 import com.google.common.testing.EqualsTester;
@@ -52,7 +52,7 @@ class TraceStateTest {
     assertThat(secondTraceState.getEntries())
         .containsExactly(TraceState.Entry.create(SECOND_KEY, SECOND_VALUE));
     assertThat(multiValueTraceState.getEntries())
-        .containsExactly(
+        .containsExactlyInAnyOrder(
             TraceState.Entry.create(FIRST_KEY, FIRST_VALUE),
             TraceState.Entry.create(SECOND_KEY, SECOND_VALUE));
   }
@@ -193,7 +193,7 @@ class TraceStateTest {
                 .set(SECOND_KEY, FIRST_VALUE) // add a new entry
                 .build()
                 .getEntries())
-        .containsExactly(
+        .containsExactlyInAnyOrder(
             TraceState.Entry.create(FIRST_KEY, SECOND_VALUE),
             TraceState.Entry.create(SECOND_KEY, FIRST_VALUE));
   }
--- a/api/src/test/java/io/opentelemetry/trace/TraceStateTest.java
+++ b/api/src/test/java/io/opentelemetry/trace/TraceStateTest.java
@@ -194,9 +194,9 @@ class TraceStateTest {
                 .set(SECOND_KEY, FIRST_VALUE) // add a new entry
                 .build()
                 .getEntries())
-        .containsExactlyInAnyOrder(
-            TraceState.Entry.create(FIRST_KEY, SECOND_VALUE),
-            TraceState.Entry.create(SECOND_KEY, FIRST_VALUE));
+        .containsExactly(
+            TraceState.Entry.create(SECOND_KEY, FIRST_VALUE),
+            TraceState.Entry.create(FIRST_KEY, SECOND_VALUE));
   }
 
   @Test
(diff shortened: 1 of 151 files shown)


carlosalberto:
Hey hey -is there any technical advantage for this? I see a few comments from the referenced issue but it sounds like popularity is the most important factor here?

> AssertJ is extremely popular in the Java ecosystem while Truth is mostly used by Google projects. 

Well, OC (and a lot of components we use) were created by Google, and I see no problem with that, unless they are inferior in quality or have been deprecated). To me, migrating to JUnit 5 is a great thing, but this change feels like a matter of personal taste (not to mention AssertJ and Truth are very similar API-wise, and also that, I think, new contributors won't care much about whether we using one or the another).

For example: I went and looked for a comparison and found issues like [this issue](https://github.com/checkstyle/checkstyle/issues/7239) or [this article](https://truth.dev/comparison.html), for example, to have a counter point for migrating to AssertJ.

anuraaga:
There's no technical reason. Which is why popularity should be a major factor here. Because the number of forks gets counted it's not a prcise measurements but seeing _two orders of magnitude_ difference in dependants is a trend I think. We can expect this because in Java Spring is huge, and Spring has adopted AssertJ. Without a significant technical reason to pick something else, we should lean towards what has a higher probability for contributors to be familiar with.

> new contributors won't care much about whether we using one or the another

I cared because I've used assertj when contributing to several other projects and Truth, only at Google - remembering the differences in API was extra effort. I can't speak for everyone but the idea is the trend indicates there is high probability of this happening more widely. I would not make such a change out of personal preference only, my opinion is that using Truth is personal preference and missing the larger trends of the community.

In #529, the only reason given to stick to truth was the prevalence in this codebase. But now you have a PR that migrates it all so that's not a reason. Please take such efforts into account

kenfinnigan:
From a long time open-source contributor to numerous Java projects, I can attest that AssertJ is widely used among Java frameworks and applications.

Until I saw this thread, I'd never heard of Truth before.

Oberon00:
What I like about AssertJ is that you can use `as` and `withFailureMessage` to get nicer exception messages. But I don't think that's a strong argument.

carlosalberto:
> There's no technical reason. Which is why popularity should be a major factor here.

I'm on the boat of "if it's not broken, don't fix it". On the other hand, given the fact also that Google has and is putting resources into OTel (like the recent zPages support), I see a good thing their developers can come and simply use what they are used to, as a counter-example.

Testing is not affecting final users, and we have only so many contributors at this moment. Given that @jkwatson has being vocal on the lack of cycles we have, I'd personally have put effort into other items myself ;)

I'm not against this PR but I'm definitely not a fan of it, as to me, it's a matter of personal taste (and I'm afraid of a Google contributor coming in the future to switch back to Truth, because of his own personal taste as well). Please go ahead and merge when/if other approvers are fine with it though ;)

jkwatson:
I'd rather make the repo more friendly to the community at large, rather than cater to a single vendor's preference, so I'm in favor of switching to AssertJ because it is better known to the majority of the community.

bogdandrutu:
@jkwatson 

>I'd rather make the repo more friendly to the community at large, rather than cater to a single vendor's preference, so I'm in favor of switching to AssertJ because it is better known to the majority of the community.

I don't think we are talking about any vendor here, it is just a library preference, as you can see the Truth was a huge influence to the `AssertJ` which changed their style to be like "assertThat.().isFoo()". So we are talking about using the library that makes the code more readable, as you can see I approved this because I feel with the latest changes `AssertJ` made huge improvements in readability

api/src/test/java/io/opentelemetry/trace/TraceStateTest.java
```diff
@@ -52,7 +52,7 @@ void getEntries() {
     assertThat(secondTraceState.getEntries())
         .containsExactly(TraceState.Entry.create(SECOND_KEY, SECOND_VALUE));
     assertThat(multiValueTraceState.getEntries())
-        .containsExactly(
+        .containsExactlyInAnyOrder(
```
jkwatson, line 55: isn't the order actually important here?

sdk/build.gradle
```diff
@@ -18,6 +18,7 @@ dependencies {
     annotationProcessor libraries.auto_value
 
     testAnnotationProcessor libraries.auto_value
+    testCompileOnly libraries.auto_value_annotation
```
jkwatson, line 21: any idea why this was needed to be added? seems unrelated to the PR, on its face.

api/src/test/java/io/opentelemetry/trace/TraceStateTest.java
anuraaga: Confirmed that the order seems to be changing here - filed #1499 will look into it in a follow up.

sdk/build.gradle
anuraaga: I think Truth was leaking this in their POM so the test dependency on Truth got this transitively.

api/src/test/java/io/opentelemetry/trace/TraceStateTest.java
jkwatson: I think the order is supposed to be the opposite of what you put in, if you look at the javadoc on the TraceState.Builder class. I think this test just wasn't asserting that fact.

anuraaga: Ah you did the digging for me! Thanks I went ahead and updated the test here.

jkwatson: I do like that assertJ makes it super clear when you aren't enforcing the order, though!

Issue #1483, opened 2020-07-30
Lots of Javadoc warnings when building


Oberon00:
For example
```
...\api\src\main\java\io\opentelemetry\common\ReadableKeyValuePairs.java:28: warning: no @return
  int size();
      ^
```
or
```
...\api\src\main\java\io\opentelemetry\common\Labels.java:86: warning: no @param for value4
  public static Labels of(
```

dengliming:
Can you assigne to me?

iNikem:
Not sure it makes sense to write javadoc for obvious params or return values... Is there any profit in such warnings?

anuraaga:
+1 to changing the defaults to not require param / return, we can make sure javadoc is useful through code review. But I wouldn't block if anyone else feels strongly about them. FWIW, if I had to choose between the style of the JDK or @trustin I'll generally side with the later :)

Oberon00:
Personally I'm especially annoyed by having to write (and read!) `@return` docs that are literally copy and pase from the summary section.

jkwatson:
100% agreed with loosening up the javadoc requirements a bit. Duplicated useless javadoc doesn't help anyone.

iNikem:
You now even require javadoc for benchmarks??

jkwatson:
> You now even require javadoc for benchmarks??

I don't think it was intentional...just got swept up in the madness.

carlosalberto:
+1 to this. We should do (later) a pass to remove things that may be annoying to new contributors.

jkwatson:
@Oberon00 Can we close this? I think this is done now with #1494

Oberon00:
I think so!

PR #1494, opened 2020-08-03
Remove javadoc warnings

dengliming:
for [https://github.com/open-telemetry/opentelemetry-java/issues/1483](https://github.com/open-telemetry/opentelemetry-java/issues/1483)

@jkwatson @iNikem @carlosalberto @Oberon00 Would you mind checking to see if it works?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -250,7 +250,7 @@ configure(opentelemetryProjects) {
     if (JavaVersion.current().isJava8Compatible()) {
         allprojects {
             tasks.withType(Javadoc) {
-                options.addStringOption('Xdoclint:none', '-quiet')
+                options.addBooleanOption('Xdoclint:all,-missing', true)
             }
         }
     }
--- a/build.gradle
+++ b/build.gradle
@@ -245,14 +245,7 @@ configure(opentelemetryProjects) {
         source = "8"
         encoding = "UTF-8"
         links 'https://docs.oracle.com/javase/8/docs/api/'
-    }
-
-    if (JavaVersion.current().isJava8Compatible()) {
-        allprojects {
-            tasks.withType(Javadoc) {
-                options.addBooleanOption('Xdoclint:all,-missing', true)
-            }
-        }
+        addBooleanOption('Xdoclint:all,-missing', true)
     }
 
     afterEvaluate {  // Allow subproject to add more source sets.
--- a/build.gradle
+++ b/build.gradle
@@ -247,6 +247,14 @@ configure(opentelemetryProjects) {
         links 'https://docs.oracle.com/javase/8/docs/api/'
     }
 
+    if (JavaVersion.current().isJava8Compatible()) {
+        allprojects {
+            tasks.withType(Javadoc) {
+                options.addStringOption('Xdoclint:none', '-quiet')
+            }
+        }
+    }
+
     afterEvaluate {  // Allow subproject to add more source sets.
         tasks.googleJavaFormat {
             source = sourceSets*.allJava


Oberon00:
This PR looks like it would suppress more warnings than it should. I think we only want to surpress missing `@param` and `@return` and the other warnings are fine.

dengliming:
It doesn't seem to be working at CI build. Don't know whether or not we can suppress missing @param and @return

jkwatson:
This builds locally for me, but I agree with @Oberon00 that this is probably too broadly removing checks. Can we do what he suggests, that is to just remove requirements for parameters and return tags? 

Also, it would be great to completely disable javadoc checks for the jmh modules, since that's not relevant at all.

carlosalberto:
>  Can we do what he suggests, that is to just remove requirements for parameters and return tags?

+1. Let's reduce scope and only go with the removal of requirements for parameters and return tag.

dengliming:
This config should works.
-Xdoclint
all,-group : enable all except group checks
missing : Checks for missing Javadoc comments or tags (for example, a missing comment or class, or a missing @return tag or similar tag on a method).
see [https://docs.oracle.com/javase/8/docs/technotes/tools/windows/javadoc.html#CHDFDACB](https://docs.oracle.com/javase/8/docs/technotes/tools/windows/javadoc.html#CHDFDACB)

dengliming:
@Oberon00 @jkwatson @carlosalberto Can you help to check again?

Oberon00:
You can look at the build here: https://app.circleci.com/pipelines/github/open-telemetry/opentelemetry-java/2752/workflows/9153c0c6-4499-4730-b44a-008a331748bf/jobs/5590

Currently we seem to have an issue there, I had this once locally but I don't remember what I did to resolve it (probably some gradlew clean sort of thing):

```
> Task :opentelemetry-api:javadoc
Picked up _JAVA_OPTIONS: -Xmx3g
/home/circleci/project/api/src/main/java/io/opentelemetry/metrics/DoubleUpDownCounter.java:79: error: cannot find symbol
  interface BoundDoubleUpDownCounter extends BoundInstrument {
                                             ^
  symbol:   class BoundInstrument
  location: interface DoubleUpDownCounter
/home/circleci/project/api/src/main/java/io/opentelemetry/metrics/LongUpDownCounter.java:79: error: cannot find symbol
  interface BoundLongUpDownCounter extends BoundInstrument {
                                           ^
  symbol:   class BoundInstrument
  location: interface LongUpDownCounter
```

Curiously, this does not make the build fail.

build.gradle
```diff
@@ -247,6 +247,14 @@ configure(opentelemetryProjects) {
         links 'https://docs.oracle.com/javase/8/docs/api/'
     }
 
+    if (JavaVersion.current().isJava8Compatible()) {
+        allprojects {
```
Oberon00, line 251: You have an allprojects nested in the "loop" that already walks over all relevant projects. This is probably not what you want.

dengliming: You want to specify a project? like api?

build.gradle
```diff
@@ -247,6 +247,14 @@ configure(opentelemetryProjects) {
...
+            tasks.withType(Javadoc) {
```
Oberon00, line 252: Instead of tasks.withType(Javadoc), why not add the addBooleanOption line to the existing `javadoc.options` block above?

build.gradle
Oberon00: No, that's not what I mean. What I mean is that we are already in a code block that gets executed for all projects. If you nest allprojects inside, you will get your code executed like

* outer code (api)
  * code in allprojects(api)
  * code in allprojects(sdk)
  * code in allprojects(propagation)
* outer code (sdk)
  * code in allprojects(api)
  * code in allprojects(sdk)
  * code in allprojects(propagation)

See also my other comment https://github.com/open-telemetry/opentelemetry-java/pull/1494#discussion_r465105481

Oberon00: See here: https://github.com/dengliming/opentelemetry-java/blob/v0.7.0/build.gradle#L19

dengliming: 👍 thanks.

dengliming:
@Oberon00 I'm curiously too. I run locally. but no this warning.
```
> Task :opentelemetry-api:javadoc
Picked up _JAVA_OPTIONS: -Xmx3g
/home/circleci/project/api/src/main/java/io/opentelemetry/metrics/DoubleUpDownCounter.java:79: error: cannot find symbol
  interface BoundDoubleUpDownCounter extends BoundInstrument {
                                             ^
  symbol:   class BoundInstrument
  location: interface DoubleUpDownCounter
/home/circleci/project/api/src/main/java/io/opentelemetry/metrics/LongUpDownCounter.java:79: error: cannot find symbol
  interface BoundLongUpDownCounter extends BoundInstrument {
                                           ^
  symbol:   class BoundInstrument
  location: interface LongUpDownCounter
...
> Task :opentelemetry-api:javadoc
2 warnings
```

build.gradle
dengliming: 👍 you were right.

PR #1500, opened 2020-08-03
Prototype code on what might look like to remove the AttributeValue from the API

jkwatson:
The goal of this experiment is to figure out how to efficiently write & read strongly-typed attributes into and out of an immutable collection-like structure. This shouldn't be taken as a proposal, but just the result of my noodlings on how to you could build an API like this.

Notes: 
* Ignore the naming..the names aren't intended to be final..just something I could have live next to the existing interfaces and classes.
* I had to give up the `of(...)` constructs, since you would have no type-safety on the values, and the implementation would have to try to figure out how to divine the types and reject non-conforming values if we wanted to just have `of(String, Object)` as the style. So, I just kept the builder, instead.
* On the reading side of the equation, things are not pretty, since I can only hand out the type and the object, and force the reader to either cast to the right type or use some helper methods to do the cast for them.

Other approaches I considered (in no order...none of them are great):
* Add union types to the JLS ( :trollface: )
* Make the AttributeConsumer interface have 8 separate methods on it to get the various typed values out, hence forcing readers to implement all 8 methods in order to read anything.
* Drop value-type-safety altogether, forcing consumers to figure it out for themselves
* Revisit my strongly-typed *key* proposal from a few months back..this wouldn't help out the auto-instrumentation code, though, as it would just move the wrapping over to the key.

Here is some sample code that utilizes the hacky API I built here:
```java
public class Main {

  public static void main(String[] args) {
    CleanAttributes attributes =
        CleanAttributes.newBuilder()
            .setString("string", "I'm a String")
            .setBoolean("boolean", true)
            .setDouble("double", 33.444d)
            .setLong("long", 34333L)
            .setStringArray("stringArray", "one", "two", "three")
            .setBooleanArray("booleanArray", true, false, true)
            .setLongArray("longArray", 33L, 55L, 99L)
            .setDoubleArray("doubleArray", 123.33, 6655.33, 339393.33, 3434.33)
            .build();

    System.out.println("attributes = " + attributes);
    System.out.println();
    System.out.println("Processing with casts:");
    processRaw(attributes);

    System.out.println();
    System.out.println("Processing with types:");
    processTyped(attributes);
  }

  public static void processRaw(CleanReadableAttributes attributes) {
    attributes.forEach(
        new RawAttributeConsumer() {
          @Override
          public void consume(String key, AttributeType type, Object value) {
            switch (type) {
              case STRING:
                String stringValue = (String) value;
                System.out.println("stringValue = " + stringValue);
                break;
              case BOOLEAN:
                boolean booleanValue = (boolean) value;
                System.out.println("booleanValue = " + booleanValue);
                break;
              case LONG:
                long longValue = (long) value;
                System.out.println("longValue = " + longValue);
                break;
              case DOUBLE:
                double doubleValue = (double) value;
                System.out.println("doubleValue = " + doubleValue);
                break;
              case STRING_ARRAY:
                List<String> stringArrayValue = (List<String>) value;
                System.out.println("stringArrayValue = " + stringArrayValue);
                break;
              case BOOLEAN_ARRAY:
                List<Boolean> booleanArrayValue = (List<Boolean>) value;
                System.out.println("booleanArrayValue = " + booleanArrayValue);
                break;
              case LONG_ARRAY:
                List<Long> longArrayValue = (List<Long>) value;
                System.out.println("longArrayValue = " + longArrayValue);
                break;
              case DOUBLE_ARRAY:
                List<Double> doubleArrayValue = (List<Double>) value;
                System.out.println("doubleArrayValue = " + doubleArrayValue);
                break;
            }
          }
        });
  }

  private static void processTyped(CleanAttributes attributes) {
    attributes.forEach(
        new TypedAttributeConsumer() {
          @Override
          public void consumeString(String key, String value) {
            System.out.println(key + " = " + value);
          }

          @Override
          public void consumeLong(String key, long value) {
            System.out.println(key + " = " + value);
          }

          @Override
          public void consumeDouble(String key, double value) {
            System.out.println(key + " = " + value);
          }

          @Override
          public void consumeBoolean(String key, boolean value) {
            System.out.println(key + " = " + value);
          }

          @Override
          public void consumeStringArray(String key, List<String> value) {
            System.out.println(key + " = " + value);
          }

          @Override
          public void consumeLongArray(String key, List<Long> value) {
            System.out.println(key + " = " + value);
          }

          @Override
          public void consumeDoubleArray(String key, List<Double> value) {
            System.out.println(key + " = " + value);
          }

          @Override
          public void consumeBooleanArray(String key, List<Boolean> value) {
            System.out.println(key + " = " + value);
          }
        });
  }
}


```

### DIFF
--- a/api/src/main/java/io/opentelemetry/common/CleanReadableAttributes.java
+++ b/api/src/main/java/io/opentelemetry/common/CleanReadableAttributes.java
@@ -16,6 +16,7 @@
 
 package io.opentelemetry.common;
 
+import java.util.List;
 import javax.annotation.Nullable;
 
 /**
@@ -32,10 +33,30 @@ public interface CleanReadableAttributes {
   /** Iterates over all the key-value pairs of attributes contained by this instance. */
   void forEach(AttributeConsumer consumer);
 
+  void forEach(TypedAttributeConsumer typedConsumer);
+
   @Nullable
   Object get(String key);
 
   interface AttributeConsumer {
     void consume(String key, AttributeType type, Object value);
+  }
+
+  interface TypedAttributeConsumer {
+    void consumeString(String key, String value);
+
+    void consumeLong(String key, long value);
+
+    void consumeDouble(String key, double value);
+
+    void consumeBoolean(String key, boolean value);
+
+    void consumeStringArray(String key, List<String> value);
+
+    void consumeLongArray(String key, List<Long> value);
+
+    void consumeDoubleArray(String key, List<Double> value);
+
+    void consumeBooleanArray(String key, List<Boolean> value);
   }
 }
--- a/api/src/main/java/io/opentelemetry/common/CleanReadableAttributes.java
+++ b/api/src/main/java/io/opentelemetry/common/CleanReadableAttributes.java
@@ -16,7 +16,6 @@
 
 package io.opentelemetry.common;
 
-import java.util.List;
 import javax.annotation.Nullable;
 
 /**
@@ -33,26 +32,10 @@ public interface CleanReadableAttributes {
   /** Iterates over all the key-value pairs of attributes contained by this instance. */
   void forEach(AttributeConsumer consumer);
 
-  Boolean getBooleanValue(Object value);
-
-  String getStringValue(Object value);
-
-  Double getDoubleValue(Object value);
-
-  Long getLongValue(Object value);
-
-  List<Boolean> getBooleanArrayValue(Object value);
-
-  List<String> getStringArrayValue(Object value);
-
-  List<Double> getDoubleArrayValue(Object value);
-
-  List<Long> getLongArrayValue(Object value);
-
   @Nullable
   Object get(String key);
 
   interface AttributeConsumer {
-    void consume(String key, AttributeValue.Type type, Object value);
+    void consume(String key, AttributeType type, Object value);
   }
 }
--- a/api/src/main/java/io/opentelemetry/common/CleanReadableAttributes.java
+++ b/api/src/main/java/io/opentelemetry/common/CleanReadableAttributes.java
@@ -31,14 +31,18 @@ public interface CleanReadableAttributes {
   boolean isEmpty();
 
   /** Iterates over all the key-value pairs of attributes contained by this instance. */
-  void forEach(AttributeConsumer consumer);
+  void forEach(RawAttributeConsumer consumer);
 
+  /**
+   * Iterates over all the key-value pairs of attributes contained by this instance, with a
+   * type-safe access pattern.
+   */
   void forEach(TypedAttributeConsumer typedConsumer);
 
   @Nullable
   Object get(String key);
 
-  interface AttributeConsumer {
+  interface RawAttributeConsumer {
     void consume(String key, AttributeType type, Object value);
   }
 
--- a/api/src/main/java/io/opentelemetry/common/CleanReadableAttributes.java
+++ b/api/src/main/java/io/opentelemetry/common/CleanReadableAttributes.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2020, OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.common;
+
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * A read-only container for String-keyed attributes.
+ *
+ * <p>See {@link Attributes} for the public API implementation.
+ */
+public interface CleanReadableAttributes {
+
+  int size();
+
+  boolean isEmpty();
+
+  /** Iterates over all the key-value pairs of attributes contained by this instance. */
+  void forEach(AttributeConsumer consumer);
+
+  Boolean getBooleanValue(Object value);
+
+  String getStringValue(Object value);
+
+  Double getDoubleValue(Object value);
+
+  Long getLongValue(Object value);
+
+  List<Boolean> getBooleanArrayValue(Object value);
+
+  List<String> getStringArrayValue(Object value);
+
+  List<Double> getDoubleArrayValue(Object value);
+
+  List<Long> getLongArrayValue(Object value);
+
+  @Nullable
+  Object get(String key);
+
+  interface AttributeConsumer {
+    void consume(String key, AttributeValue.Type type, Object value);
+  }
+}
--- a/api/src/main/java/io/opentelemetry/common/Main.java
+++ b/api/src/main/java/io/opentelemetry/common/Main.java
@@ -41,8 +41,11 @@ public class Main {
             .setDoubleArray("doubleArray", 123.33, 6655.33, 339393.33, 3434.33)
             .build();
 
+    System.out.println("attributes = " + attributes);
+    System.out.println();
     System.out.println("Processing with casts:");
     process(attributes);
+
     System.out.println();
     System.out.println("Processing with types:");
     processTyped(attributes);
@@ -94,7 +97,6 @@ public class Main {
   }
 
   public static void process(CleanReadableAttributes attributes) {
-    System.out.println("attributes = " + attributes);
     attributes.forEach(
         new CleanReadableAttributes.AttributeConsumer() {
           @Override
@@ -102,7 +104,11 @@ public class Main {
             switch (type) {
               case STRING:
                 String stringValue = type.asString(value);
+                // or (although you could put any java Class on the LHS here and it would still
+                // compile):
+                String alsoStringValue = type.cast(value);
                 System.out.println("stringValue = " + stringValue);
+                System.out.println("alsoStringValue = " + alsoStringValue);
                 break;
               case BOOLEAN:
                 boolean booleanValue = type.asBoolean(value);
--- a/api/src/main/java/io/opentelemetry/common/Main.java
+++ b/api/src/main/java/io/opentelemetry/common/Main.java
@@ -16,6 +16,7 @@
 
 package io.opentelemetry.common;
 
+import io.opentelemetry.common.CleanReadableAttributes.TypedAttributeConsumer;
 import java.util.List;
 
 @SuppressWarnings({
@@ -40,10 +41,59 @@ public class Main {
             .setDoubleArray("doubleArray", 123.33, 6655.33, 339393.33, 3434.33)
             .build();
 
+    System.out.println("Processing with casts:");
     process(attributes);
+    System.out.println();
+    System.out.println("Processing with types:");
+    processTyped(attributes);
   }
 
-  public static void process(final CleanReadableAttributes attributes) {
+  private static void processTyped(CleanAttributes attributes) {
+    attributes.forEach(
+        new TypedAttributeConsumer() {
+          @Override
+          public void consumeString(String key, String value) {
+            System.out.println(key + " = " + value);
+          }
+
+          @Override
+          public void consumeLong(String key, long value) {
+            System.out.println(key + " = " + value);
+          }
+
+          @Override
+          public void consumeDouble(String key, double value) {
+            System.out.println(key + " = " + value);
+          }
+
+          @Override
+          public void consumeBoolean(String key, boolean value) {
+            System.out.println(key + " = " + value);
+          }
+
+          @Override
+          public void consumeStringArray(String key, List<String> value) {
+            System.out.println(key + " = " + value);
+          }
+
+          @Override
+          public void consumeLongArray(String key, List<Long> value) {
+            System.out.println(key + " = " + value);
+          }
+
+          @Override
+          public void consumeDoubleArray(String key, List<Double> value) {
+            System.out.println(key + " = " + value);
+          }
+
+          @Override
+          public void consumeBooleanArray(String key, List<Boolean> value) {
+            System.out.println(key + " = " + value);
+          }
+        });
+  }
+
+  public static void process(CleanReadableAttributes attributes) {
     System.out.println("attributes = " + attributes);
     attributes.forEach(
         new CleanReadableAttributes.AttributeConsumer() {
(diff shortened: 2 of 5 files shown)


anuraaga:
> This shouldn't be taken as a proposal, but just the result of my noodlings on how to you could build an API like this.

Not sure whether you want feedback or not ;) But if you do, maybe you can temporarily copy in the `Main` code you have there into the PR? Would be easier to add comments about the API.

jkwatson:
> > This shouldn't be taken as a proposal, but just the result of my noodlings on how to you could build an API like this.
> 
> Not sure whether you want feedback or not ;) But if you do, maybe you can temporarily copy in the `Main` code you have there into the PR? Would be easier to add comments about the API.

I would love feedback/thoughts/ideas/etc. I added the Main class/method.

api/src/main/java/io/opentelemetry/common/Main.java
```diff
@@ -0,0 +1,89 @@
...
+})
+public class Main {
+
+  public static void main(String[] args) {
+    CleanAttributes attributes =
+        CleanAttributes.newBuilder()
+            .setAttribute("string", "I'm a String")
+            .setAttribute("boolean", true)
+            .setAttribute("double", 33.444d)
+            .setAttribute("long", 34333L)
```
anuraaga, line 36: I realize this is an aspect of the current API too. Wonder if we need to use `.setLong`, `.setDouble`, etc instead of `.setAttribute` to prevent the easy bug of storing a number as the wrong type.

api/src/main/java/io/opentelemetry/common/Main.java
```diff
@@ -0,0 +1,89 @@
...
+
+  public static void process(final CleanReadableAttributes attributes) {
+    System.out.println("attributes = " + attributes);
+    attributes.forEach(
+        new CleanReadableAttributes.AttributeConsumer() {
+          @Override
+          public void consume(String key, AttributeValue.Type type, Object value) {
+            switch (type) {
+              case STRING:
+                String stringValue = attributes.getStringValue(value);
```
anuraaga, line 54: This could just be `(String) value`, there doesn't seem to be much value to delegate the cast.

api/src/main/java/io/opentelemetry/common/Main.java
```diff
@@ -0,0 +1,89 @@
...
+                break;
+              case LONG:
+                Long longValue = attributes.getLongValue(value);
+                System.out.println("longValue = " + longValue);
+                break;
+              case DOUBLE:
+                Double doubleValue = attributes.getDoubleValue(value);
+                System.out.println("doubleValue = " + doubleValue);
+                break;
+              case STRING_ARRAY:
```
anuraaga, line 69: It's too bad there's no way to do `instanceof List<String>`, Java 15 would look quite nice

```java
if (value instanceof String) {
  String foo = value;  // Automatically cast;
} else if (value instanceof List<String>) // Unfortunately can't :(
```

api/src/main/java/io/opentelemetry/common/CleanReadableAttributes.java
```diff
@@ -0,0 +1,58 @@
...
+public interface CleanReadableAttributes {
+
+  int size();
+
+  boolean isEmpty();
+
+  /** Iterates over all the key-value pairs of attributes contained by this instance. */
+  void forEach(AttributeConsumer consumer);
+
+  Boolean getBooleanValue(Object value);
```
anuraaga, line 36: Mentioned it below, but these seem to not provide much value. But `boolean getBoolean(String key)`, now that sounds nice :)

thisthat: I think we should demand to the users the type retrieval. This way, we do not have to define what should happen if we call `getBoolean(key)` on an attribute that was stored as `long`, for example.

anuraaga: I think such methods are exactly demanding users to do the type retrieval - they are explicitly picking the method to call. But it gives us the ability to provide a UX possibly better for users. If we throw an exception, it's no different than any other option. But we could also treat it as an empty value if it makes sense - I'd lean towards this given we generally don't want to crash user's apps. With it in the javadoc it's pretty straightforward.

api/src/main/java/io/opentelemetry/common/Main.java
jkwatson: Easy change to make, although a bit orthogonal to this particular bit of work. :)

api/src/main/java/io/opentelemetry/common/Main.java
jkwatson: yeah, I have gone back and forth on this one. I agree with you, but I also like having the option for the non-obvious ones (like the "array" values that are actually Lists). You can definitely just have the consumer do the casts manually, for sure.

api/src/main/java/io/opentelemetry/common/Main.java
jkwatson: agreed. or kotlin. :)

api/src/main/java/io/opentelemetry/common/CleanReadableAttributes.java
jkwatson: Yes, this is probably the use-case that seems most beneficial with having these extra helper methods. We can do some intelligent default behavior when the wrong one is called, and make that uniform and documented. And, if people don't want to use them, they can do the cast themselves and handle the bad conditions however they desire.

This is one of the main values of having the typed-keys..having a type-safe API for retrieval, because then you can't ask for a boolean value from a String-typed key.

thisthat: If we don't use primitive types and we return `null` value in case of the wrong type, I agree. It would make a better UX for users.

jkwatson: One option I did consider was to have the API not give the consumer the Object value, but require them to make a `get` call based on the type and the value, which could enforce that the right type is passed in for the right method. This would necessitate, probably, doing some optimization of the `get` code path, which currently just traverses the entire list to find the key. I wasn't sure how exactly to do that. One option would be to provide an Iterator that kept the index to the currently location in the list, rather than doing it via the `forEach` and an consumer.

api/src/main/java/io/opentelemetry/common/Main.java
anuraaga: Can you clarify "non-obvious"? I guess just that it looks a little gross (which I agree with but not enough to make the indirect API I think)

`List<String> stringArrayValue = (List<String>) attributes.getStringArrayValue(value);`

That being said, I think requiring casts does have the downside of requiring the suppresswarnings "unchecked" too if those are enabled. I guess if we provide a helper whose purpose is a cast, we could make it static instead of going through the attributes instance.

iNikem: `getStringArrayValue` implies `String[]` not `List<String>`

api/src/main/java/io/opentelemetry/common/CleanReadableAttributes.java
iNikem: I agree with Anuraag that `boolean getBoolean(String key)` etc would make a nice API for single key retrieval.

api/src/main/java/io/opentelemetry/common/Main.java
iNikem: How about `String AttributeValue.Type.STRING.cast(Object value)` and similar for other types? That will clean up the API of `Attributes`

anuraaga: Would the signature be `<T> T cast(Object value)`? I guess it needs to be same signature on all the enum values. This seems ok to me though as it's pretty readable.

iNikem: It doesn't need to be the same:
```
    STRING {
      public String cast(Object value){
        return (String) value;
      }
    },
    BOOLEAN{
      public boolean cast(Object value){
        return (boolean) value;
      }
    },
```

anuraaga: Ah yeah haven't used that pattern much before - great trick for this case!

jkwatson: excellent. something like this has been tickling the back of my brain, but I couldn't quite put my finger on it. I will try this out and add this API to this PR to see how it feels/looks.

jkwatson: hmm. I'm not sure how you get a handle to call those methods. They don't appear to be visible to be called anywhere. Java doesn't let you add static members to enum instances, either, unfortunately.

jkwatson: yeah, you can't call methods like that on enum values. See here for the JEP to make it possible:  http://openjdk.java.net/jeps/301

api/src/main/java/io/opentelemetry/common/Main.java
jkwatson: I made the change here, since it was easy to do.

api/src/main/java/io/opentelemetry/common/Main.java
jkwatson: I moved the methods to do the casts into a new AttributeType class, which just moves the issue to a different place. I haven't been able to figure out a way to make it so you can only call the right one, yet.

api/src/main/java/io/opentelemetry/common/CleanReadableAttributes.java
jkwatson: easy to add, but wouldn't help with the iteration use-case, which I think will be the most common one, unfortunately.

api/src/main/java/io/opentelemetry/common/Main.java
jkwatson: I also added an 8-way typed consumer interface, just to see how that worked out. With the requirement for a consumer to do the casts with the single method, 8 methods doesn't look too bad. But, also doesn't have the advantage of being a SMI for implementation via a lambda.

api/src/main/java/io/opentelemetry/common/Main.java
```diff
@@ -0,0 +1,139 @@
...
+    System.out.println("Processing with casts:");
+    process(attributes);
+    System.out.println();
+    System.out.println("Processing with types:");
+    processTyped(attributes);
+  }
+
+  private static void processTyped(CleanAttributes attributes) {
+    attributes.forEach(
+        new TypedAttributeConsumer() {
```
anuraaga, line 53: I like this a bit better than the switch. No noise (`asString`, etc) just pure consumption

api/src/main/java/io/opentelemetry/common/Main.java
```diff
@@ -0,0 +1,139 @@
...
+
+  public static void process(CleanReadableAttributes attributes) {
+    System.out.println("attributes = " + attributes);
+    attributes.forEach(
+        new CleanReadableAttributes.AttributeConsumer() {
+          @Override
+          public void consume(String key, AttributeType type, Object value) {
+            switch (type) {
+              case STRING:
+                String stringValue = type.asString(value);
```
anuraaga, line 104: Not strong but I might find this a bit more readable

```java
enum AttributeType {
  STRING, BOOLEAN, FOO, CAT;

  public <T> T cast(Object dog) {
    return (T) dog;
  }

case STRING:
  String stringValue = STRING.cast(value);  // Having the same STRING seems easy to read
```

iNikem: I still like my original proposal more :) This code and Anuraag's suggestion are equal to me.

jkwatson: I don't know how to implement your proposal. Can you add a PR to this one to show me?

jkwatson: @anuraaga it's more readable, for sure, but 100% not type-safe. The user can do the wrong thing trivially, which seems to defeat the purpose, to me. I have this locally, though, and will commit it. :)

api/src/main/java/io/opentelemetry/common/Main.java
jkwatson: me, too, although implementing a 8-method interface is not super ergonomic, either. Especially in the cases where you don't care about the types, and just want to gather raw data [if this is a use-case?]

iNikem: Easy :) Copy all attributes to another `Attributes` before you add your own attributes to it :)

My problem is the opposite: I cannot imagine when do you to iterate over all attributes in a type-safe manner.

jkwatson: any exporter needs to do this. See all the exporters in the repo that have to do this work.

api/src/main/java/io/opentelemetry/common/Main.java
anuraaga: Is it easier to do the wrong thing with that vs `type.asString`? They still need to call the correct method in either case I think, but using the same `STRING` twice is more obvious I think for a reader of the code since there's symmetry there.

api/src/main/java/io/opentelemetry/common/Main.java
anuraaga: @jkwatson Maybe we should knock out `Attributes.toBuilder` now since looks like @iNikem really wants it :P

api/src/main/java/io/opentelemetry/common/Main.java
jkwatson: Well, using the type capture, you could put any old value on the LHS and it would still compile (`BigDecimal`, `SpanContext`, you name it). So, with the `type.as...` methods, at least they only have a limited number of options?

Unfortunately, at this point we just discussing the least bad option, I'll admit. Java's type system just doesn't have the facilities to make a really good API, without going back to `AttributeValue`, as far as I can see. 

I could probably get behind any of these options, TBH. I think the least-bad option is the 8-way interface, because at least in that case they're forced to consider all the possible types. We can also have the single one and require the manual casting along with that, so at least both options are available, depending on the use-case.

api/src/main/java/io/opentelemetry/common/Main.java
jkwatson: I think that's great to have, no matter what, so yes, definitely.

api/src/main/java/io/opentelemetry/common/Main.java
anuraaga: +1 vote to 8-way interface (and optionally also single casting variant, but with `toBuilder` the latter becomes less important perhaps)

jkwatson:
I've pruned this down to 2 options: 1) The 8-way typed consumer interface, and 2) the completely open consumer interface.

I'd really like to make a decision on weather we want to go down this route. If we do, I'll prep a PR that converts the repo over to this. @bogdandrutu @carlosalberto Your opinions would be very useful here.

jkwatson:
closing this, as I've now got an actual PR submitted that I think most of us like.

PR #1501, opened 2020-08-04
Add documentation about Java Logger for Error Handling

MitchellDumovic:
This PR follows from discussion on https://github.com/open-telemetry/opentelemetry-java/pull/1419. 

This is intended to be a an "implementation" of https://github.com/open-telemetry/opentelemetry-specification/issues/696

In the future, authors should continue to use `java.util.logging` with an appropriate log level to log messages, so that users have full control over the logging output by OpenTelemetry.

### DIFF
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -16,6 +16,8 @@
   * [Sampler](#sampler)
   * [Span Processor](#span-processor)
   * [Exporter](#exporter)
+- [Logging And Error Handling](#logging-and-error-handling)
+  * [Examples](#examples)
 <!-- tocstop -->
 
 OpenTelemetry can be used to instrument code for collecting telemetry data. For more details, check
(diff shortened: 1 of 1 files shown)


QUICKSTART.md
```diff
@@ -409,3 +411,59 @@ tracerProvider.addSpanProcessor(BatchSpanProcessor.newBuilder(
 [OpenTelemetry Website]: https://opentelemetry.io/
 [Obtaining a Tracer]: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#obtaining-a-tracer
 [Semantic Conventions]: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions
+
+## Logging and Error Handling 
+
+OpenTelemetry uses [java.util.logging](https://docs.oracle.com/javase/8/docs/api/java/util/logging/package-summary.html)
```
thisthat, line 417: I don't know how much the API changes from 7 to 8, so this suggestion might not make sense. 
```suggestion
OpenTelemetry uses [java.util.logging](https://docs.oracle.com/javase/7/docs/api/java/util/logging/package-summary.html)
```

carlosalberto: Most of the logging we do (all?) happens in the SDK level, where we use Java 8+ ;)

Oberon00: The SDK also uses Java 7 AFAIK. If not I would be in trouble. 😄

Oberon00: See https://github.com/open-telemetry/opentelemetry-java/blob/v0.7.0/build.gradle#L37-L38. We only use Java 8 for (all) unit tests. But users might be able to use the full Java 8 (or 11 or newer) logging API for configuration, even though the SDK uses only the Java 7 subset for producing the logs.

MitchellDumovic: From what I can tell the API does not change significantly from 7 to 8 - all the examples here should still be relevant. Though I agree that it makes the most sense to link the most accurate docs here, I will change this.

PR #1505, opened 2020-08-04
Update the runtime metrics to match the OTEP that sets the semantic conventions

jkwatson:
OTEP is here:  https://github.com/open-telemetry/oteps/blob/master/text/0119-standard-system-metrics.md

Note: I think the Prometheus exporter needs to be updated to not put the meter name as a prefix to the metric name.


kenfinnigan:
I know this is already merged but had a quick process question as I'm still figuring out how everything works.

From reading the OTEP process I thought the OTEP needed to be integrated with the specification before issues were opened in each language implementation? As far as I could tell, this OTEP isn't present in the specification yet.

If I missed where it is, could someone point me in the right direction?

jkwatson:
> I know this is already merged but had a quick process question as I'm still figuring out how everything works.
> 
> From reading the OTEP process I thought the OTEP needed to be integrated with the specification before issues were opened in each language implementation? As far as I could tell, this OTEP isn't present in the specification yet.
> 
> If I missed where it is, could someone point me in the right direction?

no, that's totally right. But, what we had here before was definitely violating the spec in principle (especially around the meter names being used). This gets us a lot close to what we want (and what is specced for general instrument/meter names), and we'll tweak the specific names if they get updated in the official spec to be slightly differently. 

Does that makes sense?

kenfinnigan:
Ok thanks.

If there's a concern around "runtime." prefix on things like JVM metrics, is the best place to open an issue for that in the spec repo, or this one?

jkwatson:
> Ok thanks.
> 
> If there's a concern around "runtime." prefix on things like JVM metrics, is the best place to open an issue for that in the spec repo, or this one?

Yes, or when the spec is written to match OTEP 119, that would be a great place to chime in on that particular convention.

PR #1507, opened 2020-08-05
Add a DelegatingSpanData that can be used to wrap SpanData with different data.

### LINKED ISSUE #1321: Allow exporters to decorate SpanData

anuraaga:
Alternative to #1502.

It's unfortunate since it means we have three implementations of `SpanData` - `SpanWrapper`, `TestSpanData`, `DecoratingSpanData`. Perhaps it allows squeezing a bit of performance but wonder if the wrapping concept is really worth it here in the first place.

Fixes #1321 

/cc @iNikem

### DIFF
--- a/sdk/src/main/java/io/opentelemetry/sdk/trace/data/DelegatingSpanData.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/trace/data/DelegatingSpanData.java
@@ -1,0 +1,153 @@
+package io.opentelemetry.sdk.trace.data;
+
+import static java.util.Objects.requireNonNull;
+
+import io.opentelemetry.common.ReadableAttributes;
+import io.opentelemetry.sdk.common.InstrumentationLibraryInfo;
+import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.trace.Span.Kind;
+import io.opentelemetry.trace.SpanId;
+import io.opentelemetry.trace.Status;
+import io.opentelemetry.trace.TraceFlags;
+import io.opentelemetry.trace.TraceId;
+import io.opentelemetry.trace.TraceState;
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/io/opentelemetry/sdk/trace/data/DelegatingSpanData.java
```diff
@@ -0,0 +1,257 @@
...
+ *   private final ReadableAttributes attributes;
+ *
+ *   SpanDataWithClientType(SpanData delegate) {
+ *     super(delegate);
+ *     String clientType = ClientConfig.parseUserAgent(
+ *       delegate.getAttributes().get(SemanticAttributes.HTTP_USER_AGENT).getStringValue());
+ *     Attributes.Builder newAttributes = Attributes.newBuilder();
+ *     delegate.getAttributes().forEach(newAttributes::setAttribute);
+ *     newAttributes.setAttribute("client_type", clientType);
+ *     attributes = newAttributes.build();
```
jkwatson, line 35: yeah, this was what I was thinking, @iNikem . It's not super pretty, but could be made much nicer with a `toBuilder` on the Attributes class, like we have on Labels.

PR #1510, opened 2020-08-05
Remove the instrumentation library name from the prometheus metric names

jkwatson:
Now that we've decided that the instrument name is effectively globally scoped, we shouldn't be prefixing the instrumentation library name onto the prometheus metric names.

### DIFF
--- a/exporters/prometheus/src/test/java/io/opentelemetry/exporters/prometheus/PrometheusCollectorTest.java
+++ b/exporters/prometheus/src/test/java/io/opentelemetry/exporters/prometheus/PrometheusCollectorTest.java
@@ -68,7 +68,7 @@ class PrometheusCollectorTest {
     return ImmutableList.of(
         MetricData.create(
             Descriptor.create(
-                "name",
+                "grpc.name",
                 "long_description",
                 "1",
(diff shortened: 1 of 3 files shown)


exporters/prometheus/src/test/java/io/opentelemetry/exporters/prometheus/PrometheusCollectorTest.java
```diff
@@ -68,7 +68,7 @@ void registerToDefault() throws IOException {
     return ImmutableList.of(
         MetricData.create(
             Descriptor.create(
-                "name",
+                "grpc.name",
```
anuraaga, line 71: Is this the idiomatic name for these metrics (so before we automatically prefixed grpc, now users do)? If not, it might be clearer to have the tests match what we expect users should be doing as some free documentation.

jkwatson: None of the metric semantic conventions for these things have landed yet. So, not sure at this point. This PR doesn't change what was there before, but it's a good question, and we should definitely make the tests match the semantic conventions when they exist.

PR #1512, opened 2020-08-06
Move TestSpanData to a testing package instead of in the published SDK.

anuraaga:
Extracted this out of #1502


huntc:
Any thoughts on how authors of exporters external to this project can now create test span data?

anuraaga:
That's a good callout - I think in addition to `TestSpanData`, there are many potential utilities we could provide (assertj assertion class for `Span`?). @jkwatson any thoughts on publishing an `opentelemetry-testing` artifact?

huntc:
> That's a good callout - I think in addition to `TestSpanData`, there are many potential utilities we could provide (assertj assertion class for `Span`?). @jkwatson any thoughts on publishing an `opentelemetry-testing` artifact?

Yeah... like a test kit in general...

jkwatson:
I think publishing an opentelemetry-testing artifact would be great. I'll create an issue for it.

jkwatson:
https://github.com/open-telemetry/opentelemetry-java/issues/1647

PR #1513, opened 2020-08-06
Remove log4j extension which was moved to instrumentation repo.

anuraaga:
The major difference is that we haven't implemented a JSON layout there yet. FWIU, the default JSON layout can include context data variables automatically so I'm not sure if it's required, but we can reimplement it there right away if it's being used.

https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/735
https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/751


thisthat:
Is the implementation in the instrumentation repo the same as the one in this repo?
If not, I would add to the PR description a small summary of the differences.

anuraaga:
@thisthat Thanks for the reminder

@zenmoto could you also ack this? Thanks :)

PR #1514, opened 2020-08-06
Migrate most remaining tests to JUnit5.

anuraaga:
Last ones are complicated extensions with no jupiter support yet will work upstream with them to get it.

### DIFF
--- a/exporters/jaeger/src/test/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporterTest.java
+++ b/exporters/jaeger/src/test/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporterTest.java
@@ -22,11 +22,12 @@ import static org.mockito.AdditionalAnswers.delegatesTo;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
 
+import com.google.common.io.Closer;
 import io.grpc.ManagedChannel;
+import io.grpc.Server;
 import io.grpc.inprocess.InProcessChannelBuilder;
 import io.grpc.inprocess.InProcessServerBuilder;
 import io.grpc.stub.StreamObserver;
-import io.grpc.testing.GrpcCleanupRule;
 import io.opentelemetry.exporters.jaeger.proto.api_v2.Collector;
 import io.opentelemetry.exporters.jaeger.proto.api_v2.Collector.PostSpansRequest;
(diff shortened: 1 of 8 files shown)


exporters/jaeger/src/test/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporterTest.java
```diff
@@ -44,39 +45,44 @@
...
 import org.mockito.Mockito;
 
-/** Unit tests for {@link JaegerGrpcSpanExporter}. */
-public class JaegerGrpcSpanExporterTest {
+class JaegerGrpcSpanExporterTest {
   private static final String TRACE_ID = "00000000000000000000000000abc123";
   private static final String SPAN_ID = "0000000000def456";
 
-  @Rule public final GrpcCleanupRule grpcCleanup = new GrpcCleanupRule();
+  private final Closer closer = Closer.create();
```
thisthat, line 58: Hopefully, this will be less flaky :smile:

sdk_extensions/jaeger_remote_sampler/src/test/java/io/opentelemetry/sdk/extensions/trace/jaeger/sampler/JaegerRemoteSamplerTest.java
```diff
@@ -106,20 +112,20 @@ public void connectionWorks() {
 
     // verify
     verify(service).getSamplingStrategy(requestCaptor.capture(), ArgumentMatchers.any());
-    Assert.assertEquals(SERVICE_NAME, requestCaptor.getValue().getServiceName());
-    Assert.assertTrue(sampler.getSampler() instanceof RateLimitingSampler);
-    Assert.assertEquals(
+    Assertions.assertEquals(SERVICE_NAME, requestCaptor.getValue().getServiceName());
+    Assertions.assertTrue(sampler.getSampler() instanceof RateLimitingSampler);
```
Oberon00, line 116: These should be changed to use assertThat().isInstanceOf, the error message we would get from assertTrue would be very unhelpful.

anuraaga: Yeah - there are still quite a few places that use legacy junit assertions. Eventually want to migrate those too - but I don't mind doing partial migration when pointed out :)

PR #1516, opened 2020-08-07
Add Attributes.toBuilder

anuraaga:
It's fairly easy to imagine wanting to add attributes to an existing `Attributes`, maybe storing a base in a config and tweaking at runtime.

Side effect is `Builder` constructor is made non-public which was probably intended.

### DIFF
--- a/api/src/main/java/io/opentelemetry/common/Attributes.java
+++ b/api/src/main/java/io/opentelemetry/common/Attributes.java
@@ -44,6 +44,11 @@ public abstract class Attributes extends ImmutableKeyValuePairs<AttributeValue>
 
     @Override
     abstract List<Object> data();
+
+    @Override
+    public Builder toBuilder() {
+      return new Builder(new ArrayList<>(data()));
+    }
   }
 
   /** Returns a {@link Attributes} instance with no attributes. */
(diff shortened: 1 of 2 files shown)


jkwatson:
One use-case that I suspect might be common would be to *change* a value from the original attributes. Since we don't have any guarantees about behavior when duplicate keys are added, we can't currently support this use-case. This PR seems like a fine API addition, though, and we can address the key-override use-case as a separate issue, I think.

jkwatson:
@carlosalberto @bogdandrutu can you take a look, since this is an addition to the API?

api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -44,6 +44,11 @@
 
     @Override
     abstract List<Object> data();
+
+    @Override
+    public Builder toBuilder() {
```
bogdandrutu, line 49: Would be good to keep the Labels in sync and add similar functionality there?

jkwatson: Labels actually got it implemented first.  :). https://github.com/open-telemetry/opentelemetry-java/blob/7ca698f8452260c8b8da067149444bb01a04da29/api/src/main/java/io/opentelemetry/common/Labels.java#L114

PR #1518, opened 2020-08-07
fix splitting metadata key-value substring with more than one '=' sign

malafeev:
cc @carlosalberto

### DIFF
--- a/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcSpanExporter.java
+++ b/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcSpanExporter.java
@@ -291,7 +291,7 @@ public final class OtlpGrpcSpanExporter implements SpanExporter {
       String metadataValue = getStringProperty(KEY_METADATA, configMap);
       if (metadataValue != null) {
         for (String keyValueString : Splitter.on(';').split(metadataValue)) {
-          final List<String> keyValue = Splitter.on('=').splitToList(keyValueString);
+          final List<String> keyValue = Splitter.on('=').limit(2).splitToList(keyValueString);
           if (keyValue.size() == 2) {
             addHeader(keyValue.get(0), keyValue.get(1));
           }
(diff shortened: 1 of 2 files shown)


malafeev:
@anuraaga I modified existing test to check values with more than one `=`

exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcSpanExporter.java
```diff
@@ -291,7 +291,7 @@ protected Builder fromConfigMap(
       String metadataValue = getStringProperty(KEY_METADATA, configMap);
       if (metadataValue != null) {
         for (String keyValueString : Splitter.on(';').split(metadataValue)) {
-          final List<String> keyValue = Splitter.on('=').splitToList(keyValueString);
+          final List<String> keyValue = Splitter.on('=').limit(2).splitToList(keyValueString);
```
anuraaga, line 294: Should we also add `trimResults` `omitEmptyStrings`? These are HTTP headers so at least keys cannot have spaces or be empty. I'm less sure about trimming the values or the key only, but I think usually a trailing space on a value will be a bug not a feature.

malafeev: added

exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcSpanExporter.java
```diff
@@ -291,7 +291,8 @@ protected Builder fromConfigMap(
       String metadataValue = getStringProperty(KEY_METADATA, configMap);
       if (metadataValue != null) {
         for (String keyValueString : Splitter.on(';').split(metadataValue)) {
-          final List<String> keyValue = Splitter.on('=').splitToList(keyValueString);
+          final List<String> keyValue = Splitter.on('=').limit(2).trimResults().omitEmptyStrings()
```
anuraaga, line 294: Thanks! If you can add two more test cases, one with trimmed results and one with empty strings, that'd be great.

malafeev: added

Issue #1135, opened 2020-04-24
Why is there a DefaultSpan?


yurishkuro:
There is no such concept in the specification, but this class is present in the Java API.

https://github.com/open-telemetry/opentelemetry-java/blob/efc49522ac4fd01bcf94afa60789264515aa75cd/api/src/main/java/io/opentelemetry/trace/DefaultSpan.java

Can it not be hidden inside the SDK as implementation detail?

jkwatson:
@carlosalberto Can probably answer this more completely than I can, but the purpose of the DefaultSpan is to provide an implementation when there is no SDK installed. The implementation does context propagation but nothing else.

yurishkuro:
I don't think that's the only reason, because such default implementation can be done without exposing the DefaultSpan as a public class in the API.

jkwatson:
Looks like the only usage that's keeping it from being made package-access is in the `HttpTraceContext` implementation. @carlosalberto would probably have the most insight into why that leaked in therel

bogdandrutu:
@yurishkuro I am happy to work on making the default implementations not public. Would also be good to make this call across different languages, currently looking for example at go https://godoc.org/go.opentelemetry.io/otel/api/trace there a Noop public types. I think it is good to be consistent.

carlosalberto:
Well, having `DefaultSpan` accessible (or at least, allowing the user to create it) allows us to do some in-propagation stuff. 

I'd like us to offer a way to keep this, if possible (by keeping a method somewhere, that returns an empty `Span` out of a `SpanContext`).

yurishkuro:
Can you be specific? What does it allow?

carlosalberto:
Hey @yurishkuro 

So it allows users to use a `SpanContext` as a `Span`. Moreover, with the latest changes from OTEP 66, this is the only way to store an active `SpanContext` there:

```java
Context context = TracingContextUtils.withSpan(DefaultSpan.create(spanContext)); // OR
Scope scope = TracingContextUtils.currentContextWithSpan(DefaultSpan.create(spanContext));
```

But perhaps that's the only important case we need to support. I see two alternatives: 

1) Keep a method somewhere to create empty `Span`s out of `SpanContext` values (as it's currently the case with `DefaultSpan.create()`), **or**

2) Add `TracingContextUtils` method overloads taking a `SpanContext` instead of `Span` (which internally would use a `DefaultSpan` anyway):

```java
[...]

PR #1519, opened 2020-08-07
Return Span from DefaultSpan factories.

anuraaga:
#1135 talks about hiding `DefaultSpan` completely. I couldn't quite find a good home for the factory methods (maybe `DefaultTracer`?) but it seems like a good idea.

In the meantime, it's more important to make the return type of the factories generic because users should definitely not have to care whether a span is a `DefaultSpan` or not. In practice, the return value of `DefaultSpan` makes it difficult to bridge them in auto instrumentation.

https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/764


carlosalberto:
LGTM. I've long wondered whether there's any reason to expose `DefaultSpan` publicly. @bogdandrutu may know better (although I can't think of one at the moment).

PR #1520, opened 2020-08-07
Update to latest RPC semantic attributes.

### LINKED ISSUE #1402: Missing SemanticAttributes for RPC?
I'm using v0.6.0 of the Java OpenTelemetry API. In [`SemanticAttributes`](https://github.com/open-telemetry/opentelemetry-java/blob/master/api/src/main/java/io/opentelemetry/trace/attributes/SemanticAttributes.java) the only attribute related to RPC I can find is `rpc.service`. Yet the [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/rpc.md) lists a couple of other RPC-related attirbutes, such as `rpc.system` and `rpc.method`. (In fact `rpc.system` seems to have been present in OpenTracing as "component", so it's not really new.)

anuraaga:
Fixes #1402

### DIFF
--- a/api/src/main/java/io/opentelemetry/trace/attributes/SemanticAttributes.java
+++ b/api/src/main/java/io/opentelemetry/trace/attributes/SemanticAttributes.java
@@ -134,6 +134,7 @@ public final class SemanticAttributes {
   public static final StringAttributeSetter RPC_SERVICE =
       StringAttributeSetter.create("rpc.service");
 
+/* The name of the method being called, must be equal to the $method part in the span name */
   public static final StringAttributeSetter RPC_METHOD = StringAttributeSetter.create("rpc.method");
 
   /** The name of a gRPC span event to populate for each message sent / received. */
--- a/api/src/main/java/io/opentelemetry/trace/attributes/SemanticAttributes.java
+++ b/api/src/main/java/io/opentelemetry/trace/attributes/SemanticAttributes.java
@@ -127,20 +127,30 @@ public final class SemanticAttributes {
    */
   public static final LongAttributeSetter HTTP_RESPONSE_CONTENT_LENGTH_UNCOMPRESSED =
       LongAttributeSetter.create("http.response_content_length_uncompressed");
-  /** The service name, must be equal to the $service part in the span name. */
+
+  /** A string identifying the remoting system, e.g., "grpc", "java_rmi" or "wcf". */
+  public static final StringAttributeSetter RPC_SYSTEM = StringAttributeSetter.create("rpc.system");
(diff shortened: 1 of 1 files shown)


api/src/main/java/io/opentelemetry/trace/attributes/SemanticAttributes.java
```diff
@@ -127,20 +127,30 @@
...
+
+  /** A string identifying the remoting system, e.g., "grpc", "java_rmi" or "wcf". */
+  public static final StringAttributeSetter RPC_SYSTEM = StringAttributeSetter.create("rpc.system");
+  /** The full name of the service being called, including its package name, if applicable. */
   public static final StringAttributeSetter RPC_SERVICE =
       StringAttributeSetter.create("rpc.service");
-  /** RPC span event attribute with value "SENT" or "RECEIVED". */
-  public static final StringAttributeSetter MESSAGE_TYPE =
+
+  public static final StringAttributeSetter RPC_METHOD = StringAttributeSetter.create("rpc.method");
```
thisthat, line 137: I would still add the comment 
```suggestion
/* The name of the method being called, must be equal to the $method part in the span name */
  public static final StringAttributeSetter RPC_METHOD = StringAttributeSetter.create("rpc.method");
```

api/src/main/java/io/opentelemetry/trace/attributes/SemanticAttributes.java
```diff
@@ -127,20 +127,31 @@
...
-  /** The service name, must be equal to the $service part in the span name. */
+
+  /** A string identifying the remoting system, e.g., "grpc", "java_rmi" or "wcf". */
+  public static final StringAttributeSetter RPC_SYSTEM = StringAttributeSetter.create("rpc.system");
+  /** The full name of the service being called, including its package name, if applicable. */
   public static final StringAttributeSetter RPC_SERVICE =
       StringAttributeSetter.create("rpc.service");
-  /** RPC span event attribute with value "SENT" or "RECEIVED". */
-  public static final StringAttributeSetter MESSAGE_TYPE =
+
```
anuraaga, line 136:
```suggestion
```

api/src/main/java/io/opentelemetry/trace/attributes/SemanticAttributes.java
anuraaga: Thanks somehow missed this one

thisthat: Thanks to you for keeping this up2date :)

PR #1522, opened 2020-08-08
Add RecordEventsReadableSpan toString

trask:
This would be very helpful for debugging our sporadic test failures in opentelemetry-java-instrumentation.

Let me know if there are other fields you'd like included. I just included the ones I thought would be most useful.


sdk/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -589,6 +589,46 @@ private ReadableAttributes getImmutableAttributes() {
     return builder.build();
   }
 
+  @Override
+  public String toString() {
+    synchronized (lock) {
+      return "RecordEventsReadableSpan{"
```
iNikem, line 595: Is it Ok for such string concatenation inside synchronised section?

PR #1523, opened 2020-08-10
Add JmxMetrics extension

rmfitzpatrick:
These additions provide a convenient method of creating metrics from queried MBeans in custom groovy scripts.  Currently only synchronous instruments are supported, though the plan is to provide asynchronous helpers in the near future.


extensions/jmx_metrics/README.md
```diff
@@ -0,0 +1,116 @@
...
+MBeans from a JMX server.  It loads a custom Groovy script establishes a helpful bound `otel`
+object with methods for obtaining MBeans and constructing synchronous OpenTelemetry instruments:
+
+### Usage
+
+```bash
+$ java -cp <opentelemetry-java jars> io.opentelemetry.extensions.metrics.jmx.JmxMetrics -config ./config.json
+```
+
+##### `config.json` example
```
jrcamp, line 21: Would it be better to use Java properties and allow them to be specified on the command line? It may be easier for some cases instead of having to generate a file? e.g. `java -jar whatever.jar -Dotel.intervalSeconds=5`. Could also use properties file instead of having to include dependencies.

jkwatson:
This looks interesting, but I think it would be better as a contribution to the instrumentation project, rather than the API/SDK itself. The instrumentation project is here: https://github.com/open-telemetry/opentelemetry-java-instrumentation

rmfitzpatrick:
@jkwatson considering this isn't using bytebuddy or generating trace data is introducing metric helpers an overall intention of that project?  This seems closer to the current runtime metric extension than anything in java-instrumentation.

jkwatson:
> @jkwatson considering this isn't using bytebuddy or generating trace data is introducing metric helpers an overall intention of that project? This seems closer to the current runtime metric extension than anything in java-instrumentation.

We're actually planning on moving the runtime metrics over to the instrumentation repo. It includes (or will include) both auto-and-non-auto instrumentation code.

jkwatson:
It's a little hidden over in the instrumentation repo, but this document outlines the scope for that repo:  https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/docs/scope.md

We should probably add something similar over here!

rmfitzpatrick:
@jkwatson thanks, will port this over.  Did you happen to have a prepared location for things like this or is ~`/instrumentation/jmx-metrics` fine?

jkwatson:
I would ask in https://gitter.im/open-telemetry/opentelemetry-java-instrumentation. I don't spend much time over in that repo yet, but the folks in that channel would definitely have the info.

PR #1525, opened 2020-08-10
Separate the SDK into separate publishable artifacts

### LINKED ISSUE #1460: Separate the SDK into separate publishable artifacts
Combining the tracing and metrics API aspects into a single artifact and accessible through the same API class, [OpenTelemetry](https://github.com/open-telemetry/opentelemetry-java/blob/master/api/src/main/java/io/opentelemetry/OpenTelemetry.java), makes it difficult to utilize different solutions for metrics or tracing.

For instance, the [Micrometer](https://micrometer.io/) project is utilized by Spring Boot, Micronaut, and other Java libraries/frameworks for metrics. If they want to continue using Micrometer for metrics, but want to use tracing from OpenTelemetry, it duplicates classes and code that can cause confusion for developers.

I would like to see separate artifacts for tracing and metrics that can be consumed individually if projects so desire.

There can still be a separate artifact that includes the all-encompassing [OpenTelemetry](https://github.com/open-telemetry/opentelemetry-java/blob/master/api/src/main/java/io/opentelemetry/OpenTelemetry.java) for those developers who want to utilize both aspects of OpenTelemetry.

kenfinnigan:
Closes #1460 

I've separated the existing `sdk` module into the following structure:

```
/sdk
  /all
  /common
  /correlation_context
  /metrics
  /tracing
```


sdk/all/build.gradle
```diff
@@ -0,0 +1,54 @@
...
+description = 'OpenTelemetry SDK'
+ext.moduleName = "io.opentelemetry.sdk"
+ext.propertiesDir = "build/generated/properties/io/opentelemetry/sdk"
+
+dependencies {
+    api project(':opentelemetry-api'),
+            project(':opentelemetry-sdk-common'),
+            project(':opentelemetry-sdk-correlation-context'),
+            project(':opentelemetry-sdk-metrics'),
+            project(':opentelemetry-sdk-tracing')
```
jkwatson, line 18: I am not a gradle expert, by any means, but I think we might want to actually have this module have multiple source sets, so downloading it from maven central, or wherever, is a single artifact, rather than multiple artifacts, which is what having dependent modules will result in.

bogdandrutu:
What is the motivation?

sdk/all/build.gradle
kenfinnigan: That's just for "all" which is the new "sdk" right?

I'm ok with doing that if someone can explain/show how.

kenfinnigan:
@bogdandrutu the motivation is that in the Java space [Micrometer](https://micrometer.io/) is an extremely popular and well-maintained framework for capturing and exporting metrics.

To that end, [Quarkus](https://quarkus.io/) would like the ability to use Micrometer with OTeL while still retaining the tracing and other aspects of OTeL. For background, we currently use OpenTracing.

If the "sdk" is split into separate pieces, it's easier to replace the OTeL metrics implementation with Micrometer. Yes it would be possible to just not use the Metrics classes, but as mentioned on the issue having classes present but not used is an anti-pattern for Quarkus, and we want to minimize dead code. Without that option, we'd need to consider developing a separate sdk implementation that used Micrometer for metrics, but we don't want to do that if it can be avoided.

In addition, with the "sdk" split into its constituent parts, it's easier to prevent concerns from different parts from bleeding into another.

sdk/all/build.gradle
jkwatson: Here's an example of this being done: https://github.com/newrelic/newrelic-telemetry-sdk-java/pull/214/files

jkwatson: and, yes, just for the 'all' artifact.

anuraaga: I generally find multiple sources to have a higher chance of biting later (if both source files end up in the final app which is reasonably likely). Is there an advantage to the fat jar for an SDK? Don't feel that strongly one way or another but would lean towards letting Maven do it's thing, it's pretty good at it

kenfinnigan: Can someone confirm whether or not I should make this change?

jkwatson: I don't feel super strongly about it, so if someone cares a lot, we can have it as a follow-up issue/PR.

kenfinnigan:
I added a BOM for the SDK

sdk/common/src/main/java/io/opentelemetry/sdk/common/InstrumentationLibraryInfo.java
```diff
@@ -1,72 +1,71 @@
...
+package io.opentelemetry.sdk.common;
+
+import com.google.auto.value.AutoValue;
+import io.opentelemetry.trace.Tracer;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * Holds information about the instrumentation library specified when creating an instance of {@link
+ * Tracer} using io.opentelemetry.sdk.trace.TracerSdkProvider.
```
jkwatson, line 26: for some unknown reason this line no longer uses the imported class, but the fully-qualified one. I think it'll make the diff slightly cleaner if that change got reverted.

kenfinnigan: I had to do that because the class is in the "tracing" module, which isn't a dependency of "common".

As it was for Javadoc purposes, I thought it would be ok.

Are there alternatives? I couldn't find an easy way to reference a class in another module without adding a bunch of things to the javadoc command line

jkwatson: hmm. good point. I might get rid of the reference altogether to the concrete class, and just reference something like "Tracer or Meter Provider".

kenfinnigan: Will do

sdk/bom/build.gradle
```diff
@@ -0,0 +1,25 @@
+plugins {
```
anuraaga, line 1: We have the opentelemetry bom, is there a benefit in having another bom @jkwatson ?

jkwatson: I actually have no idea. The last I heard, the bom wasn't being properly published. Not a bom expert here.

jkwatson: https://github.com/open-telemetry/opentelemetry-java/issues/1413 for reference

anuraaga: Yeah - I use the bom from the snapshots repo right now and it's fine, we'll need to figure out what's up with maven central, but I wouldn't add another bom which is just a subset of it.

kenfinnigan: On #1460 @bogdandrutu had requested a bom for the SDK pieces.

If there's agreement it's not needed, I can take it out. I'm fine with whatever the majority want

jkwatson: cool. if the existing bom encompasses this already, then I think we don't need another. 👍

kenfinnigan: I removed the SDK BOM

anuraaga:
Thanks for helping with this!

kenfinnigan:
My pleasure @anuraaga.

Thanks to everyone providing feedback and guidance

carlosalberto:
Approving but I wan to hold this from being merged, and wait to

1) Have a final resolution on the related issue in the Specification
2) Have @bogdandrutu approve (the changes are big enough, package organization-wise).

kenfinnigan:
@bogdandrutu any additional concerns with this?

@carlosalberto what's necessary for the resolution at the specification? Anything I can help with?

jkwatson:
Note related spec issue:  https://github.com/open-telemetry/opentelemetry-specification/issues/779

There haven't been any objections in that spec issue, so I think we should move ahead with this change. @kenfinnigan It looks like there is a conflict that has arisen in the sdk build.gradle. Can you get this rebased? Thanks!

jkwatson:
@kenfinnigan as soon as you get this rebased and the build passing, we're ready to merge. Thanks!

kenfinnigan:
Thanks @jkwatson, will rebase shortly

kenfinnigan:
@jkwatson quick question, as the JMH code is now actually running, it's found some compile issues.

What's the preferred approach between keeping the JMH tests in their separate places (metrics and tracing) but removing the use of `OpenTelemetrySdk` from the code, or move all JMH tests into the /sdk/all module where the SDK class is present?

kenfinnigan:
I moved all the JMH code under _/sdk/all/_ as it seemed to be the most logical location

jkwatson:
oh no. I think the changes to the exporter API getting merged has necessitated another rebase. :(. @huntc beat you to the merge race.

kenfinnigan:
Rebased again

kenfinnigan:
@jkwatson I've done the latest rebase, barring new changes, it should be ok.

PR #1528, opened 2020-08-11
Return SDK attributes in default Resource.

anuraaga:
It seems like `getDefault` should return all of our default attributes. The split was probably when we initialized `EnvVarResource` directly, but with that replaced by this generic API it looks like we can provide a more complete default.


jkwatson:
This looks great. Are there any unit tests that would make sense to add for this change?

anuraaga:
Yeah added tests

anuraaga:
Hmm - my test in https://github.com/open-telemetry/opentelemetry-java/pull/1528/commits/f2eaa7c62abcda0e69dec9e66e78ff78aea9da05 seems OK and runs fine locally but failed on CI, so went ahead and deleted it for now.

PR #1529, opened 2020-08-11
Update Gradle Wrapper to 6.6

friederbluemle:
This updates the Gradle Wrapper to the latest version 6.6

https://docs.gradle.org/6.6/release-notes.html


friederbluemle:
I know this is my first PR here, so I shouldn't be complaining, but I have to leave this here: The "EasyCLA" process is by far the most awkward and complicated CLA process I've seen. A fully mailing address?! Why do dozens of other open source projects have a simply 1 or 2 click solution in place? There surely must be a more elegant way.. 

Just something to think about... ;)

jkwatson:
> I know this is my first PR here, so I shouldn't be complaining, but I have to leave this here: The "EasyCLA" process is by far the most awkward and complicated CLA process I've seen. A fully mailing address?! Why do dozens of other open source projects have a simply 1 or 2 click solution in place? There surely must be a more elegant way..
> 
> Just something to think about... ;)

Nobody disagrees. This was a CNCF decision, not specifically us. :(

PR #1535, opened 2020-08-12
Fix for Issue with bound instruments generating datapoints from no recordings.

### LINKED ISSUE #1453: Bound Metric Instruments create empty MetricData
**Describe the bug**
When binding a synchronous instrument, it records empty `MetricData` even though no metrics are recorded. This does not occur with the `BoundValueRecorder`.

**Steps to reproduce**
```
        Meter meter = OpenTelemetry.getMeterProvider().get("test","test:1.0.0");
        DoubleCounter counter = meter
                .doubleCounterBuilder("byte_received")
                .build();
        DoubleCounter.BoundDoubleCounter boundCounter = byteCounter.bind(Labels.empty());
        
        Collection<MetricData> data = OpenTelemetrySdk.getMeterProvider().getMetricProducer().collectAllMetrics();
        for(MetricData md : data) {
            System.out.println(md);
        }
```

As with other instruments, points should be empty. 

**What did you see instead?**
Points contains a `DoublePoint` with the value 0.0.

Artifacts: `opentelemetry-sdk`
Version: `v0.6.0`
How did you reference these artifacts?  `compile('io.opentelemetry:opentelemetry-sdk:0.6.0-SNAPSHOT')`

**Environment**
Compiler: n/a
OS: Windows 10

**Additional context**
None

jkwatson:
Fixes #1453 

The fix is to introduce an AtomicBoolean in the AbstractAggregator that tracks whether recordings have been made since the last reset was done. The ActiveBatcher queries the aggregator and if no recordings have been made, it skips generating metric points.

### DIFF
--- a/sdk/src/main/java/io/opentelemetry/sdk/metrics/aggregator/AbstractAggregator.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/metrics/aggregator/AbstractAggregator.java
@@ -16,14 +16,12 @@
 
 package io.opentelemetry.sdk.metrics.aggregator;
 
-import java.util.concurrent.atomic.AtomicBoolean;
-
 abstract class AbstractAggregator implements Aggregator {
   // Note: This is not 100% thread-safe. There is a race condition where recordings can
   // be made in the moment between the reset and the setting of this field's value. In those
   // cases, it is possible that a recording could be missed in a given recording interval, but
   // it should be picked up in the next, assuming that more recordings are being made.
-  private final AtomicBoolean hasRecordings = new AtomicBoolean(false);
+  private volatile boolean hasRecordings = false;
 
   @Override
   public void mergeToAndReset(Aggregator other) {
@@ -31,7 +29,7 @@ abstract class AbstractAggregator implements Aggregator {
       return;
     }
     doMergeAndReset(other);
-    hasRecordings.set(false);
+    hasRecordings = false;
   }
 
   /**
@@ -47,8 +45,8 @@ abstract class AbstractAggregator implements Aggregator {
 
   @Override
   public final void recordLong(long value) {
-    hasRecordings.set(true);
     doRecordLong(value);
+    hasRecordings = true;
   }
 
   /**
@@ -62,8 +60,8 @@ abstract class AbstractAggregator implements Aggregator {
 
   @Override
   public final void recordDouble(double value) {
-    hasRecordings.set(true);
     doRecordDouble(value);
+    hasRecordings = true;
   }
 
   /**
@@ -77,6 +75,6 @@ abstract class AbstractAggregator implements Aggregator {
 
   @Override
   public boolean hasRecordings() {
-    return hasRecordings.get();
+    return hasRecordings;
   }
 }
--- a/sdk/src/main/java/io/opentelemetry/sdk/metrics/aggregator/AbstractAggregator.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/metrics/aggregator/AbstractAggregator.java
@@ -16,7 +16,14 @@
 
 package io.opentelemetry.sdk.metrics.aggregator;
 
+import java.util.concurrent.atomic.AtomicBoolean;
+
 abstract class AbstractAggregator implements Aggregator {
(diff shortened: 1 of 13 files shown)


sdk/src/main/java/io/opentelemetry/sdk/metrics/aggregator/AbstractAggregator.java
```diff
@@ -16,14 +16,22 @@
...
 package io.opentelemetry.sdk.metrics.aggregator;
 
+import java.util.concurrent.atomic.AtomicBoolean;
+
 abstract class AbstractAggregator implements Aggregator {
+  // Note: This is not 100% thread-safe. There is a race condition where recordings can
+  // be made in the moment between the reset and the setting of this field's value. In those
+  // cases, it is possible that a recording could be missed in a given recording interval, but
+  // it should be picked up in the next, assuming that more recordings are being made.
+  private final AtomicBoolean hasRecordings = new AtomicBoolean(false);
```
anuraaga, line 26: Since we're not using CAS, this can just be `volatile boolean`.

sdk/src/main/java/io/opentelemetry/sdk/metrics/aggregator/AbstractAggregator.java
```diff
@@ -38,12 +46,37 @@ public void mergeToAndReset(Aggregator other) {
   abstract void doMergeAndReset(Aggregator aggregator);
 
   @Override
-  public void recordLong(long value) {
-    throw new UnsupportedOperationException("This Aggregator does not support long values");
+  public final void recordLong(long value) {
+    hasRecordings.set(true);
```
anuraaga, line 50: Isn't it better to set after recording since that's when it actually "has recordings"?

sdk/src/main/java/io/opentelemetry/sdk/metrics/aggregator/Aggregator.java
```diff
@@ -57,4 +57,7 @@
    * @param value the new {@code double} value to be added.
    */
   void recordDouble(double value);
+
+  /** Whether there have been any recordings since this aggregator has been reset. */
+  boolean hasRecordings();
```
anuraaga, line 62: No particular preference but wondered if `isEmpty` is more idiomatic

sdk/src/main/java/io/opentelemetry/sdk/metrics/aggregator/AbstractAggregator.java
jkwatson: I've gone back and forth several times on this, trying to figure out which leads to the least-bad race condition when it happens. I think you're right, though; I'll swap it around.

sdk/src/main/java/io/opentelemetry/sdk/metrics/aggregator/AbstractAggregator.java
jkwatson: true enough. I tend to usually not use volatile, rather than the Atomics, but I don't have a really strong feeling either way. I think the performance is pretty much equal, IIRC. Although I have this feeling that that changed with java 14 and Atomics got much worse (/me goes to read his Java Specialists newsletters).

sdk/src/main/java/io/opentelemetry/sdk/metrics/aggregator/Aggregator.java
jkwatson: In this case, I don't know that aggregators necessarily have a universal concept of "emptiness", so I went with something very semantically clear and domain-specific.

sdk/src/main/java/io/opentelemetry/sdk/metrics/aggregator/AbstractAggregator.java
jkwatson: done. thanks for the close look!

sdk/src/main/java/io/opentelemetry/sdk/metrics/aggregator/AbstractAggregator.java
jkwatson: updated.

Issue #1533, opened 2020-08-12
API convenience methods for when empty labels are used


kenfinnigan:
**Is your feature request related to a problem? Please describe.**
Currently, to use an unbounded metric without labels it's necessary to do something like:

```
doubleCounter.add(2, Labels.empty())
```

**Describe the solution you'd like**
It would be great if any `add()`, `record()`, or other methods where `Labels` is a parameter for recording an unbounded metric,
that the API provided a convenience method that delegates to the current one.

For `DoubleCounter` it would mean adding:

```
default void add(double increment) {
  add(increment, Labels.empty());
}
```

Allowing the developer to:
```
doubleCounter.add(2)
```

dengliming:
If so, should we update sourceCompatibility  to 1.8? 
```
sourceCompatibility = 1.8
    targetCompatibility = 1.8
```

jkwatson:
Unfortunately, we can't do this with default methods, as the API still needs to support java 7.

kenfinnigan:
So we can do the convenience, just not with a default method?

Is the rationale for requiring JDK 7 documented anywhere? Would be interested to read it

dengliming:
@kenfinnigan sure

jkwatson:
The rationale boils down to Android & grpc wanting to be able to use the APIs for java 7-compatible Android versions.

kenfinnigan:
Thanks @jkwatson

thisthat:
Closing this since #1538 got merged. Please, reopen the issue if you think this was not fully addressed.

PR #1538, opened 2020-08-13
Add convenience methods for when empty labels are used.

dengliming:
for [https://github.com/open-telemetry/opentelemetry-java/issues/1533](https://github.com/open-telemetry/opentelemetry-java/issues/1533)
cc @kenfinnigan @jkwatson

### DIFF
--- a/api/src/main/java/io/opentelemetry/metrics/DoubleCounter.java
+++ b/api/src/main/java/io/opentelemetry/metrics/DoubleCounter.java
@@ -64,6 +64,14 @@ public interface DoubleCounter extends SynchronousInstrument<BoundDoubleCounter>
    */
   void add(double increment, Labels labels);
 
+  /**
+   * Adds the given {@code increment} to the current value. The values cannot be negative.
+   *
+   * <p>The value added is associated with the current {@code Context} and with empty labels.
+   *
+   * @param increment the value to add.
+   * @since 0.8.0
+   */
   void add(double increment);
 
   @Override
--- a/api/src/main/java/io/opentelemetry/metrics/DoubleCounter.java
+++ b/api/src/main/java/io/opentelemetry/metrics/DoubleCounter.java
@@ -64,6 +64,8 @@ public interface DoubleCounter extends SynchronousInstrument<BoundDoubleCounter>
    */
   void add(double increment, Labels labels);
 
+  void add(double increment);
+
   @Override
   BoundDoubleCounter bind(Labels labels);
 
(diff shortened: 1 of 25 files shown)


api/src/main/java/io/opentelemetry/metrics/DoubleUpDownCounter.java
```diff
@@ -67,6 +67,8 @@
    */
   void add(double increment, Labels labels);
 
+  void add(double increment);
```
jkwatson, line 70: these methods all need javadoc, please

jkwatson:
This seems like a good user-friendly API change. Can you add some tests for the SDK that assert that the no-label versions have the same result as the Labels.empty() versions? 

Thanks!

dengliming:
> This seems like a good user-friendly API change. Can you add some tests for the SDK that assert that the no-label versions have the same result as the Labels.empty() versions?
> 
> Thanks!

sure. will fix it later. thanks

api/src/main/java/io/opentelemetry/metrics/DoubleUpDownCounter.java
dengliming: fixed at lastest commit.

api/src/main/java/io/opentelemetry/metrics/DoubleCounter.java
```diff
@@ -64,6 +64,16 @@
...
 
+  /**
+   * Adds the given {@code increment} to the current value. The values cannot be negative.
+   *
+   * <p>The value added is associated with the current {@code Context} and with empty labels.
+   *
+   * @param increment the value to add.
+   * @since 0.8.0
+   */
+  void add(double increment);
```
anuraaga, line 75: Not related to this PR at all - @jkwatson I was surprised to see here that we don't allow negative additions

jkwatson: The UpDownCounters are the ones that allow negative increments.

anuraaga: Ah got it! I think... do we need two types of counters? Just food for thought no strong opinion, mostly just used to things like `AtomicInteger` in terms of API.

jkwatson: There was intense debate in the metric specs about this. The rationale (for good or ill) was that having different instruments that reflect the different semantics of the counting was valuable for instrumentation, to make it very clear what sort of counting you were doing. We went the opposite way from OpenCensus, which (I think?) has one instrument for everything, and lots of options.

jkwatson:
I'm going to merge this. If @carlosalberto or @bogdandrutu have a problem with this small API change, we can revisit.

PR #1540, opened 2020-08-14
Clarify timestamps of events in javadoc.

anuraaga:
I think this captures the thoughts in https://github.com/open-telemetry/opentelemetry-specification/issues/796 let me know


carlosalberto:
Not sure that we need all that description (specially the part regarding how to compute the tstamp), but definitely better than what we have now.

PR #1542, opened 2020-08-17
Fixes Name of Cassandra Keyspace attribute key

pjan:


api/src/main/java/io/opentelemetry/trace/attributes/SemanticAttributes.java
```diff
@@ -195,7 +195,7 @@
    * For db.system == cassandra, the name of the keyspace being accessed. To be used instead of the
    * generic db.name attribute.
    */
-  public static final StringAttributeSetter CASSANDRA_NAMESPACE =
+  public static final StringAttributeSetter CASSANDRA_KEYSPACE =
```
anuraaga, line 198: Doh, thanks!

jkwatson:
Thanks for the contribution!

Issue #1439, opened 2020-07-20
If more than one consumer is hooked up to an SDK's MetricProducer, they will not all see all the metric datapoints.


jkwatson:
Since the MetricProducer provided by the MeterProvderSdk consumes all the metrics whenever they are requested, it's a race to see who actually gets the points.

If this is intentional, we should definitely document this behavior very clearly.
If it's not intentional, we should fix it.

jkwatson:
Closing this, now that it's properly documented. We can open a new issue if/when we want to address a redesign which will make this a non-issue.

PR #1546, opened 2020-08-17
Add documentation to make it clear that the MetricProducer cannot be shared across exporters

jkwatson:
It doesn't exactly fix the underlying issue in #1439 (that will probably take a redesign), but it at least documents it more clearly to reduce surprise.

### DIFF
--- a/sdk/src/main/java/io/opentelemetry/sdk/metrics/MeterSdkProvider.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/metrics/MeterSdkProvider.java
@@ -64,7 +64,7 @@ public final class MeterSdkProvider implements MeterProvider {
    * MeterSdkProvider}.
    *
    * <p>WARNING: A MetricProducer is stateful. It will only return changes since the last time it
-   * was access. This means that if more than one {@link
+   * was accessed. This means that if more than one {@link
    * io.opentelemetry.sdk.metrics.export.MetricExporter} has a handle to this MetricProducer, the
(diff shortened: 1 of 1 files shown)


sdk/src/main/java/io/opentelemetry/sdk/metrics/MeterSdkProvider.java
```diff
@@ -63,6 +63,11 @@ public MeterSdk get(String instrumentationName, String instrumentationVersion) {
    * Returns the {@link MetricProducer} that can be used to retrieve metrics from this {@code
    * MeterSdkProvider}.
    *
+   * <p>WARNING: A MetricProducer is stateful. It will only return changes since the last time it
+   * was access. This means that if more than one {@link
```
anuraaga, line 67:
```suggestion
   * was accessed. This means that if more than one {@link
```

sdk/src/main/java/io/opentelemetry/sdk/metrics/MeterSdkProvider.java
```diff
@@ -63,6 +63,11 @@ public MeterSdk get(String instrumentationName, String instrumentationVersion) {
...
+   * io.opentelemetry.sdk.metrics.export.MetricExporter} has a handle to this MetricProducer, the
+   * two exporters will not receive the same metric data to export.
```
anuraaga, line 69:
```suggestion
   * two exporters will not receive copies of the same metric data to export.
```

Issue #1454, opened 2020-07-23
Being able to specify Resource manually


carlosalberto:
Currently the only way to set `Resource`s for the global `OpenTelemetry` is through environment variables - but we need to allow users to set them manually (through code) or through system properties (in a similar fashion to the configuration supported for `BatchSpanProcessor` and siblings).

I see two possible options to do this:

1) Set the `Resource` in the loaded `TracerSdkProvider` and siblings **once**, and report errors if the user tries to do this again.
2) Allow `TracerSdkProvider` and siblings to be set manually, in addition to the SPI path. For this, @Oberon00's [PR](https://github.com/open-telemetry/opentelemetry-java/pull/1058/files) would help us achieve this purpose (while also allowing global initialization, of course).

Option 2) is the most reasonable to me, and can help get it to the finish line, but want to get initial feedback (mostly from @bogdandrutu )

EDIT: Updated the description to specifically address the global case.

Oberon00:
> Currently the only way to set Resources is through environment variables

That's technicylly not true: There is `TracerProviderSdk.Builder.setResource` https://github.com/open-telemetry/opentelemetry-java/blob/v0.6.0/sdk/src/main/java/io/opentelemetry/sdk/trace/TracerSdkProvider.java#L174
It is cumbersome to use without #1058 but still possible if you provide your own SPI.

anuraaga:
Yeah as an example I'm doing this

https://github.com/aws-samples/aws-xray-sdk-with-opentelemetry-sample/blob/both-otel-and-xray/awsagentprovider/src/main/java/com/softwareaws/xray/opentelemetry/exporters/AwsTracerProviderFactory.java#L35

Because we have a pattern of generating many different `Resource` and merging them, an SPI that collects all the `Resource` and merges them would be very natural as a default. Manual initialization probably needs to provide two options of merging into such a default or replacing it completely. I'd expect most users to merge into the default with their own domain-specific attributes though. 

Is there a difference between this issue and https://github.com/open-telemetry/opentelemetry-java/issues/1149?

carlosalberto:
> That's technically not true

Well, there's also the possibility of use reflection to achieve that ;) But in all seriousness, being able to set `Resource` easily for the common case, i.e. the global `OpenTelemetry` part.

I honestly cannot imagine users having to extend the SPI part to *only* override `Resource`.

carlosalberto:
[...]

PR #1548, opened 2020-08-18
Add ResourceProvider SPI to allow for custom Resources

iNikem:
Partially addresses #1454

### DIFF
--- a/sdk/src/main/java/io/opentelemetry/sdk/resources/Resource.java
+++ b/sdk/src/main/java/io/opentelemetry/sdk/resources/Resource.java
@@ -26,6 +26,7 @@ import io.opentelemetry.internal.StringUtils;
 import io.opentelemetry.internal.Utils;
 import java.util.Objects;
 import java.util.Properties;
+import java.util.ServiceLoader;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
@@ -58,11 +59,11 @@ public abstract class Resource {
           .readEnvironmentVariables()
           .readSystemProperties()
           .build()
-          .merge(TELEMETRY_SDK);
+          .merge(TELEMETRY_SDK)
+          .merge(readResourceFromProviders());
 
   @Nullable
   private static String readVersion() {
-
     Properties properties = new Properties();
     try {
       properties.load(
@@ -72,6 +73,14 @@ public abstract class Resource {
       return "unknown";
     }
     return properties.getProperty("sdk.version");
+  }
+
+  private static Resource readResourceFromProviders() {
+    Resource result = Resource.EMPTY;
+    for (ResourceProvider resourceProvider : ServiceLoader.load(ResourceProvider.class)) {
+      result = result.merge(resourceProvider.create());
+    }
+    return result;
   }
 
   Resource() {}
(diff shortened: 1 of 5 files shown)


sdk/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -74,6 +75,14 @@ private static String readVersion() {
     return properties.getProperty("sdk.version");
   }
 
+  private static Resource readResourceFromProviders() {
+    Resource result = Resource.EMPTY;
+    for (ResourceProvider resourceProvider : ServiceLoader.load(ResourceProvider.class)) {
```
anuraaga, line 80: We should probably add a property with a CSV list of implementation class name to exclude to allow users to disable resources they don't want to include automatically.

sdk/src/main/java/io/opentelemetry/sdk/resources/ResourceProvider.java
```diff
@@ -0,0 +1,37 @@
...
+ * <p>Resources specified via system properties or environment variables will take precedence
+ * over any value supplied via {@code ResourceProvider}.
+ *
+ * @see EnvAutodetectResource
+ */
+@ThreadSafe
+public abstract class ResourceProvider {
+
+  public Resource create() {
+    final Attributes.Builder attrBuilder = Attributes.newBuilder();
```
anuraaga, line 24: Attributes is immutable guess you don't need to copy?

iNikem: oi, indeed

sdk/src/main/java/io/opentelemetry/sdk/resources/Resource.java
iNikem: Cannot the same be achieved via controlling application classpath?

anuraaga: If every resource provider is in a separate artifact I think so but will that be the case? Not sure - but I sort of figured that SDK will have one extension artifact with a few providers so disabling one of them only by classpath manipulation could be tough.

iNikem: The only use-case for this that I currently know of is for auto-instrumentation to provide `telemetry.auto.version`. Which, btw, I wouldn't want end-users to disable :)

anuraaga: There are many resources we haven't implemented in Java yet, like docker container ID or process. I figure they should also be implemented with this SPI

https://github.com/open-telemetry/opentelemetry-specification/pull/811#discussion_r470931890

We'll probably have a lot of these providers for full data coverage but for startup time or possible privacy reasons presumably they should be disablable and classpath probably won't be enough. In agent we would probably bundle everything and need a property anyways I think?

iNikem: Ok, fair enough.

- Should I do this in this very same PR?
- Should new config property go to `TraceConfig`?

anuraaga: Think follow-up PR is fine.

Probably not `TraceConfig` - we'd have to stop using the static instance, and I guess this doesn't need to be programmatic (can just programmatically set Resource instead) so it's only for control from command line.

iNikem: Added #1551

iNikem:
I will be happy if this gets merged today :) Then I can finish implementing this in instrumentation repo (European) tomorrow.

iNikem:
@jkwatson @tylerbenson @carlosalberto ^^

PR #1549, opened 2020-08-18
Add Getter.Keys() with Jaeger Baggage support.

### LINKED ISSUE #967: Add Jaeger propagator
Similar to https://github.com/open-telemetry/opentelemetry-java/issues/853 but for Jaeger clients.

* https://www.jaegertracing.io/docs/1.17/client-libraries/#propagation-format
* https://github.com/jaegertracing/jaeger-client-java/blob/master/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java

Does OTEL client support exporting/injecting of TraceContext/SpanContext into two formats? At the same time use W3C TraceContext and Jaeger? The use case is to allow gradual migration of existing deployments using Jaeger prop. format to W3C.

carlosalberto:
Fixes #967 

Implements the https://github.com/open-telemetry/opentelemetry-specification/pull/825 update, and implements Jaeger Baggage support to verify that the API is correct/useful (based on [their own propagator](https://github.com/jaegertracing/jaeger-client-java/blob/master/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/TextMapCodec.java))

- Used `Iterable` for `keys()`.
- Had to remove the usage of anonymous classes in the integration test's `trace-context` section, as it broke some versions of Java 11 (see [this](https://stackoverflow.com/questions/54775253/jdk-11-0-2-compilation-fails-with-javac-npe-on-anonymous-parameterized-class-typ)).
- We can play with removing the (most likely) slow regex in a follow-up.
- Don't put an empty `Baggage` in `Context` in case there was not Baggage to be extracted at all.

For Jaeger reviewers:

- Didn't encode the Baggage values. In Jaeger there's two `TextMap` propagators registered by default (encoded and non-encoded propagation). We need to define the encoding case in a follow-up (somehow).

@open-telemetry/java-approvers 
cc @pavolloffay @jpkrohling


richardstartin:
The problem with making this a collection is it will often need to be materialised, even if you can get away with calling `Map.keySet` in a lot of cases. Making it an `Iterable` would be lighter-weight in most cases, e.g. Kafka client, old enumeration based context carriers (e.g. JMS, `HttpServletRequest`)

carlosalberto:
I'd be fine to have `keys()` return an `Iterable` or `Iterator`.

jkwatson:
Can you add the concrete example of how the Jaeger propagator will use this new feature to this PR? I think that will be very useful for the spec PR, as well, to demonstrate how you're doing to use it.

context_prop/src/main/java/io/opentelemetry/context/propagation/TextMapPropagator.java
```diff
@@ -115,6 +116,14 @@
    */
   interface Getter<C> {
 
+    /**
+     * Returns all the keys in the given carrier.
+     *
+     * @param carrier carrier of propagation fields, such as an http request.
```
jkwatson, line 122:
```suggestion
     * @param carrier The carrier of propagation fields, such as an http request.
```

context_prop/src/main/java/io/opentelemetry/context/propagation/TextMapPropagator.java
```diff
@@ -115,6 +116,14 @@
...
+     * @since 0.10.0
```
jkwatson, line 123: I think we've gotten rid of all the `@since` tags for now; we'll set them all to 1.0 when we're ready to release.

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagator.java
```diff
@@ -187,6 +208,42 @@ public static JaegerPropagator getInstance() {
...
+        }
+        builder.put(key.substring(BAGGAGE_PREFIX.length()), getter.get(carrier, key));
+      } else if (key.equals(BAGGAGE_HEADER)) {
+        builder = parseBaggageHeader(getter.get(carrier, key), builder);
+      }
+    }
+    return builder == null ? null : builder.build();
+  }
+
+  @SuppressWarnings("StringSplitter")
```
jkwatson, line 231: not relevant for this PR in particular, but I think we should turn off this warning globally, since it's telling us to use guava, which we don't to do for the API, and won't, ever.

extensions/trace_propagators/src/test/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagatorTest.java
```diff
@@ -133,12 +145,49 @@ void inject_NotSampledContext() {
...
+                BAGGAGE_PREFIX + "meta", "meta-value"));
+  }
+
   @Test
   void extract_Nothing() {
     // Context remains untouched.
     assertThat(
             jaegerPropagator.extract(
-                Context.current(), Collections.<String, String>emptyMap(), Map::get))
+                Context.current(), Collections.<String, String>emptyMap(), getter))
```
jkwatson, line 190: might be a little safer/cleaner to just use `Context.root()` here, so we know that we don't have a pre-polluted context.

jkwatson:
This looks ok (although it makes me sad that our Getter is no longer a SMI). Based on the discussion above, I had thought you would go with an Iterable/Iterator, rather Collection as the return type for `keys()`. Did something change your mind?

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagator.java
carlosalberto: +1

context_prop/src/main/java/io/opentelemetry/context/propagation/TextMapPropagator.java
carlosalberto: Oh, I totally forgot about it. Will do.

extensions/trace_propagators/src/test/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagatorTest.java
carlosalberto: Fair enough, will update.

carlosalberto:
> I had thought you would go with an Iterable/Iterator, rather Collection as the return type for keys(). Did something change your mind?

Wanted to give a last try at `Collection` and to verify it's still fine ;) But if you think we should try enumerable instead (as mentioned in the previous related conversation), then I will go with enumerable for sure.

jpkrohling (commented):
The Jaeger specific parts are looking good to me, but I'm not an expert in the propagation format. Perhaps @objectiser or @yurishkuro would want to double-check as well?

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagator.java
```diff
@@ -108,16 +118,27 @@ public static JaegerPropagator getInstance() {
...
 
     SpanContext spanContext = getSpanContextFromHeader(carrier, getter);
-    if (!spanContext.isValid()) {
-      return context;
+    if (spanContext.isValid()) {
+      context = context.with(Span.wrap(spanContext));
+    }
+
+    Baggage baggage = getBaggageFromHeader(carrier, getter);
+    if (baggage != null) {
```
anuraaga, line 137: Does it make sense to add empty baggage similar to how we add invalid span?

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagator.java
```diff
@@ -187,6 +208,42 @@ public static JaegerPropagator getInstance() {
     return buildSpanContext(traceId, spanId, flags);
   }
 
+  private static <C> Baggage getBaggageFromHeader(C carrier, Getter<C> getter) {
+    Baggage.Builder builder = null;
+
+    for (String key : getter.keys(carrier)) {
+      if (key.startsWith(BAGGAGE_PREFIX)) {
+        if (key.length() <= BAGGAGE_PREFIX.length()) {
```
anuraaga, line 216: Since we already checked startsWith I think this can be equals check

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagator.java
```diff
@@ -187,6 +208,42 @@ public static JaegerPropagator getInstance() {
...
+      } else if (key.equals(BAGGAGE_HEADER)) {
+        builder = parseBaggageHeader(getter.get(carrier, key), builder);
+      }
+    }
+    return builder == null ? null : builder.build();
+  }
+
+  @SuppressWarnings("StringSplitter")
+  private static Baggage.Builder parseBaggageHeader(String header, Baggage.Builder builder) {
+    for (String part : header.split("\\s*,\\s*")) {
```
anuraaga, line 233: Regex split looks like it could be very slow. Best could be to use indexOf to avoid split array allocation but if too tedious can we at least just split on the single chars?

extensions/trace_propagators/src/test/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagatorTest.java
```diff
@@ -313,6 +362,78 @@ void extract_UrlEncodedContext() throws UnsupportedEncodingException {
...
+    carrier.put(BAGGAGE_HEADER, "nometa+novalue");
+
+    assertThat(getBaggage(jaegerPropagator.extract(Context.current(), carrier, getter)))
+        .isEqualTo(Baggage.empty());
+  }
+
+  @Test
+  void extract_baggageOnly_withHeader_andPrefix() {
+    Map<String, String> carrier = new LinkedHashMap<>();
+    carrier.put(BAGGAGE_HEADER, "nometa=nometa-value,meta=meta-value");
```
anuraaga, line 425: Can you add some test cases with invalid, and with whitespace?

anuraaga (commented):
Mostly looks nice thanks

yurishkuro:
> In Jaeger there's two TextMap propagators registered by default (encoded and non-encoded propagation). Which one is the default? 

Neither is default. In OT, HTTP and TextMap are different, so Jaeger url-encodes baggage for HTTP, but not for TextMap (eg in Kafka headers there's no need to url-encode).

> The jaeger-baggage header support only happens at extraction, not injection, which makes me believe it's deprecated? Should we mention this in the docs?

That is correct. The official baggage format (in and out) is uberctx-{key}. jaeger-baggage header was invented later for manual injection via curl.

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagator.java
carlosalberto: That's a tricky one (I will let @pavolloffay @jpkrohling comment), but I'd imagine we wanting to keep whatever `Baggage` we had before, in case there was no baggage *at all*.

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagator.java
carlosalberto: Let's do that in a follow-up. This code is ported directly from the Jaeger Propagator, but yes, we could try to optimize things in follow ups (we did the same for OC's TraceContext propagator, for example).

extensions/trace_propagators/src/test/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagatorTest.java
carlosalberto: Oh, yes, totally missed that.

context_prop/src/main/java/io/opentelemetry/context/propagation/TextMapPropagator.java
```diff
@@ -115,6 +116,14 @@
    */
   interface Getter<C> {
 
+    /**
+     * Returns all the keys in the given carrier.
+     *
+     * @param carrier The carrier of propagation fields, such as an http request.
+     * @since 0.10.0
+     */
+    Collection<String> keys(C carrier);
```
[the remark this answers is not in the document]
carlosalberto: Fair enough. Will update to go the iterable way.

carlosalberto:
> Neither is default. In OT, HTTP and TextMap are different, so Jaeger url-encodes baggage for HTTP, but not for TextMap (eg in Kafka headers there's no need to url-encode).

Oh, interesting. @yurishkuro , what do you think we should do here? Maybe have two versions, one that does encoding, one that does not? Or maybe something else? e.g.

```
  private JaegerPropagator(bolean doBaggageEncoding) {
     this.doBaggageEncoding = doBaggageEncoding; // Depending on this, baggage will be encoded or not.
  }
```

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagator.java
```diff
@@ -187,6 +208,42 @@ public static JaegerPropagator getInstance() {
...
+        if (key.length() <= BAGGAGE_PREFIX.length()) {
+          continue;
+        }
+
+        if (builder == null) {
+          builder = Baggage.builder();
+        }
+        builder.put(key.substring(BAGGAGE_PREFIX.length()), getter.get(carrier, key));
+      } else if (key.equals(BAGGAGE_HEADER)) {
+        builder = parseBaggageHeader(getter.get(carrier, key), builder);
```
yurishkuro, line 225: Jaeger SDKs would combine baggage from `uberctx-{key}` and `jaeger-baggage` headers, you're overriding it instead.

carlosalberto: Actually they are getting merged, i.e. see [this test](https://github.com/open-telemetry/opentelemetry-java/pull/1549/files#diff-4e39f9e60626e3859be5734ea004b64bbbc1530e9a3d8e685122164aebfe27cdR423) - let me know otherwise :)

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagator.java
carlosalberto: I'd rather stick to check the length, as it only depends on a numerical comparison.

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagator.java
yurishkuro: L225 replaces the builder, so they are only merged if jaeger-baggage header is first in the list, as it is in the unit test, moving this header to the end of the list will ignore uberctx entries, as far as I can tell.

carlosalberto: Verified that it works correctly by putting HEADER at the end (of the Map), i.e. L225 takes any **existing** builder and uses it if already exists.

context_prop/src/main/java/io/opentelemetry/context/propagation/TextMapPropagator.java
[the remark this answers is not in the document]
carlosalberto: Done.

carlosalberto:
@open-telemetry/java-approvers Please review the latest updates.

@yurishkuro Updated the code based on your feedback. In a follow up we can add the encoded-values case, if that works for you.

carlosalberto:
@yurishkuro Also, there's the question on whether, upon **no existing baggage** at all in the carrier, we should put an empty `Baggage` in `Context` (for example, if `SpanContext` was extracted but no **baggage** was detected at alll).

yurishkuro (approved):

yurishkuro:
> upon no existing baggage at all in the carrier, we should put an empty Baggage in Context

I would try to avoid that. Even if there is an immutable singleton `Baggage.empty()`, adding it into an immutable context still requires re-allocation of the context, which is a malloc for little benefit. If would just make `Baggage.fromContext(ctx)` return empty singleton when key is not present.

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagator.java
anuraaga: Ah I meant an equal check on the length - if it starts with it, then the length is guarateed to be at least that length and the `<` here looks redundant

jkwatson:
looks like this needs a rebase, @carlosalberto

anuraaga (approved):

jkwatson (approved):

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagator.java
carlosalberto: Done.

carlosalberto:
@jkwatson Rebased ;)

PR #1555, opened 2020-08-19
Implement OS and some process resource attributes.

anuraaga:
Hinted at this in #1548 and wanted to kick it off so people can have a place to put non-vendor-specific resource attributes.


sdk/common/src/main/java/io/opentelemetry/sdk/resources/ResourceConstants.java
```diff
@@ -25,6 +25,36 @@
  */
 public class ResourceConstants {
 
+  /** The operating system type, such as {@code "WINDOWS"}, {@code "DARWIN"}, {@code "LINUX"}. */
+  public static final String OS_NAME = "os.name";
```
iNikem, line 29: Are these attributes defined by semantic convention? If yes, I think a link to it will be useful here.

iNikem (approved):

sdk/common/src/main/java/io/opentelemetry/sdk/resources/ResourceConstants.java
anuraaga: It's in the class javadoc - but thanks to the reminder I found it was a dead link :)

sdk_extensions/resources/src/main/java/io/opentelemetry/sdk/extensions/resources/OsResource.java
```diff
@@ -0,0 +1,88 @@
...
+      // Ignore
+    }
+    String osDescription = version != null ? os + ' ' + version : os;
+    attributes.setAttribute(ResourceConstants.OS_DESCRIPTION, osDescription);
+
+    return attributes.build();
+  }
+
+  @Nullable
+  private static String getOs(String os) {
```
thisthat, line 61: Would `contains` make the code more robust? I am not familiar with all the cases so I might be wrong.

sdk_extensions/resources/src/main/java/io/opentelemetry/sdk/extensions/resources/ProcessResource.java
```diff
@@ -0,0 +1,78 @@
...
+    if (atIndex >= 0) {
+      String pidString = runtimeName.substring(0, atIndex);
+      try {
+        pid = Long.parseLong(pidString);
+      } catch (NumberFormatException ignored) {
+        // Ignore parse failure.
+      }
+    }
+
+    attributes.setAttribute(ResourceConstants.PROCESS_PID, pid);
```
thisthat, line 46: I would avoid reporting the attribute if no process id was available.

sdk_extensions/resources/src/main/java/io/opentelemetry/sdk/extensions/resources/ProcessResource.java
```diff
@@ -0,0 +1,78 @@
...
+import java.lang.management.ManagementFactory;
+import java.lang.management.RuntimeMXBean;
+
+/** {@link ResourceProvider} which provides information about the current running process. */
+public class ProcessResource extends ResourceProvider {
+  @Override
+  protected Attributes getAttributes() {
+    Attributes.Builder attributes = Attributes.newBuilder();
+
+    // TODO(anuraaga): Use reflection to get more stable values on Java 9+
```
jkwatson, line 32: We should create an issue for this, so the TODO doesn't lie dead in the code.

sdk_extensions/resources/src/main/java/io/opentelemetry/sdk/extensions/resources/ProcessResource.java
```diff
@@ -0,0 +1,78 @@
...
+    RuntimeMXBean runtime = ManagementFactory.getRuntimeMXBean();
+    long pid = -1;
+    String runtimeName = ManagementFactory.getRuntimeMXBean().getName();
+    int atIndex = runtimeName.indexOf('@');
```
jkwatson, line 36: could you add a little comment about what the structure of this name looks like, so the code doesn't just look like magic. Thanks!

sdk_extensions/resources/src/main/java/io/opentelemetry/sdk/extensions/resources/ProcessResource.java
jkwatson: agreed. I'd rather have no value, rather than a -1

sdk_extensions/resources/src/main/java/io/opentelemetry/sdk/extensions/resources/ProcessResource.java
```diff
@@ -0,0 +1,78 @@
...
+      javaHome = System.getProperty("java.home");
+      osName = System.getProperty("os.name");
+    } catch (SecurityException e) {
+      // Ignore
+    }
+    if (javaHome != null) {
+      StringBuilder executablePath = new StringBuilder(javaHome);
+      executablePath
+          .append(File.pathSeparatorChar)
+          .append("bin")
```
jkwatson, line 60: interesting assumption. This is definitely the most common location for a java executable, but is it universal enough to be encoded here?

sdk_extensions/resources/src/main/java/io/opentelemetry/sdk/extensions/resources/ProcessResource.java
```diff
@@ -0,0 +1,78 @@
...
+          .append("java");
+      if (osName != null && osName.toLowerCase().startsWith("windows")) {
+        executablePath.append(".exe");
+      }
+
+      attributes.setAttribute(ResourceConstants.PROCESS_EXECUTABLE_PATH, executablePath.toString());
+
+      StringBuilder commandLine = new StringBuilder(executablePath);
+      for (String arg : runtime.getInputArguments()) {
+        commandLine.append(' ').append(arg);
```
jkwatson, line 71: I'm nervous about blindly adding all the command lines arguments from a java process.This could end up being gigantic, and also leaking senstivie information that is being passed in via system properties. Are we sure this is a good idea?

jkwatson: I guess if this isn't something we're putting in by default in the SDK, it's probably fine. I think we should definitely have some warnings in the module README, or perhaps in the javadoc, or somewhere. :)

anuraaga: Checked the output and it seems to only have system properties / normal command line, and not classpath for example which would be huge.

I think this point applies to the semantic convention itself so we'll probably need to find some guidelines on how to deal with this attribute there.

sdk_extensions/resources/src/main/java/io/opentelemetry/sdk/extensions/resources/ProcessResource.java
anuraaga: Oops forgot, thanks!

Unfortunately, don't think there's any way for me to write a unit test for it.

sdk_extensions/resources/src/main/java/io/opentelemetry/sdk/extensions/resources/ProcessResource.java
jkwatson: oh good. I was worried about the classpath, which would be gigantic. The point still holds for system properties and security, and you're right about the conventions probably needing some clarification. I'm still ok with this for now, since it's an opt-in extension module.

anuraaga:
Oops forgot to push

sdk_extensions/resources/src/main/java/io/opentelemetry/sdk/extensions/resources/ProcessResource.java
jkwatson: Let's go with it for now. I'm sure a user on some obscure platform will let us know when it's wrong. :)

jkwatson: I wonder what this does on Android, for instance. I also hope no one is running this SDK on android, since we haven't been (and don't intend to be) optimizing for that runtime environment.

anuraaga: Currently it's guaranteed to crash because java management isn't available anyways :) I added a note to the README and note to the issue that we can reflectionize this too to unbreak.

thisthat (approved):

Issue #1526, opened 2020-08-10
When using the env vars to set the TraceConfig, the SDK should wrap the probability sampler with a ParentOrElse sampler.


jkwatson:
When we introduced the ParentOrElse sampler as the default, we didn't update the TraceConfig builder code to wrap the probability sampler with a ParentOrElse. I think this is probably unexpected behavior, since the ParentOrElse should continue being the default wrapper Sampler.

If/when we change the Probability sampler to not consider the parent sampling decision, then this bug would be more severe.

bogdandrutu:
This whole config needs a revision

PR #1557, opened 2020-08-19
Remove link and parent sampling decision from Probability sampler

### LINKED ISSUE #1225: Probability sampler only respects positive parent decision
~According to [the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/18b2752ebe6c7f0cdd8c7b2bcbdceb0ae3f5ad95/specification/trace/sdk.md#built-in-samplers), the probability sampler's "default behavior should be to trust the parent `SampledFlag`"~ (EDIT: See https://github.com/open-telemetry/opentelemetry-java/issues/1225#issuecomment-668224260). However, in Java, the decision is only respected if sampled=true. If sampled=false, the trace is potentially restarted with the sampler's probability. IMHO, notwithstanding any spec-conformance, this behavior is very weird and surprising.

The responsible condition is this:
https://github.com/open-telemetry/opentelemetry-java/blob/d7f6d5a64176b44752bbd4bd403716a0f150b3e6/sdk/src/main/java/io/opentelemetry/sdk/trace/Samplers.java#L195

thisthat:
Fixes #1225

### DIFF
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
@@ -473,18 +473,6 @@ public final class Samplers {
         Kind spanKind,
         ReadableAttributes attributes,
         @Nullable List<Link> parentLinks) {
-      // If the parent is sampled keep the sampling samplingResult.
-      if (parentContext.getTraceFlags().isSampled()) {
-        return EMPTY_RECORDED_AND_SAMPLED_SAMPLING_RESULT;
-      }
-      if (parentLinks != null) {
-        // If any parent link is sampled keep the sampling samplingResult.
-        for (Link parentLink : parentLinks) {
-          if (parentLink.getContext().getTraceFlags().isSampled()) {
-            return EMPTY_RECORDED_AND_SAMPLED_SAMPLING_RESULT;
-          }
-        }
-      }
       // Always sample if we are within probability range. This is true even for child spans (that
       // may have had a different sampling samplingResult made) to allow for different sampling
       // policies,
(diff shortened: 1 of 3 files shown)


anuraaga (approved):
Nice

jkwatson:
I think this is a good change, but I have a concern. Right now, if you configure the trace config with a probability, it won't wrap it in a ParentOrElse sampler, even though it starts out that way. This means that a simple config change to change the default probability will have a knock-on effect of also no longer considering the parent's sampling state. 

I think we need to get that bit fixed before we can merge this change. Either, we need to not use the ParentOrElse sampler by default, or we need to always use it when the config is used to change the sampling probability.

jkwatson:
This is the blocking issue:  https://github.com/open-telemetry/opentelemetry-java/issues/1526

thisthat:
@jkwatson I see your point but I think this is not strictly related to the changes introduced by this PR. This PR is about the probability sampler behavior alone whereas #1526 is about changing either the returned object by `Samplers.probability(double)` or https://github.com/open-telemetry/opentelemetry-java/blob/master/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java#L303

jkwatson:
If we merge this, we will break existing users who are expecting the parent-logic to work, with a non-100% probability sampler. I'd like to not leave the system in a state where you can't get the existing behavior.

thisthat:
> If we merge this, we will break existing users who are expecting the parent-logic to work, with a non-100% probability sampler. I'd like to not leave the system in a state where you can't get the existing behavior.

Fair point! :)

jkwatson:
@thisthat I think that #1595 should unblock this. Please take a look.

thisthat:
Thank you @jkwatson! I'll have a look and update the PR accordingly

thisthat:
@jkwatson I have rebased and update the PR. Please, have another look if everything is correctly in place now :)

jkwatson (approved):
Thanks!

PR #1559, opened 2020-08-19
BatchSpanProcessor now starts daemon timer thread

### LINKED ISSUE #1558: JVM fails to exit
After [this change](https://github.com/open-telemetry/opentelemetry-java/commit/d6b53e60f4b7251d7a5845f3e3f9b9e57a5dc8dc#diff-df5688a9dcb936f33790f852e8283adeR167) SDK starts a non-daemon thread which prevents JVM from exiting. This breaks auto-instrumentation tests and will break any other application with running Otel SDK.

### LINKED ISSUE #1588: Deal with the TODOs in the BatchSpanProcessorTest
They should either be resolved, or issues should be created for them.

iNikem:
Fixes #1558

### DIFF
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
@@ -164,7 +164,7 @@ public final class BatchSpanProcessor implements SpanProcessor {
 
     private static final BoundLongCounter droppedSpans;
 
-    private final Timer timer = new Timer(TIMER_THREAD_NAME, /* isDaemon= */true);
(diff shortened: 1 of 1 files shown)


sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -164,7 +164,7 @@ public void forceFlush() {
 
     private static final BoundLongCounter droppedSpans;
 
-    private final Timer timer = new Timer(TIMER_THREAD_NAME);
+    private final Timer timer = new Timer(TIMER_THREAD_NAME, /* isDaemon= */ true);
```
jkwatson, line 167: ugh. I'd like to get rid of the usage of Timer completely and replace it with a ScheduledExecutor

Issue #876, opened 2020-02-18
Do not validate tracestate


justinfoote:
The opentelemetry-java implementation of W3C TraceContext is currently validating every tracestate entry (See [here](https://github.com/open-telemetry/opentelemetry-java/blob/1fced4898fd82a7c9e237c65de4dc235e7085758/api/src/main/java/io/opentelemetry/trace/TraceState.java#L207-L213)).  An invalid tracestate entry will cause an `IllegalArgumentException`, via `Utils::checkArgument`, which will bubble up to `HttpTraceContext::extract`, [here](https://github.com/open-telemetry/opentelemetry-java/blob/1fced4898fd82a7c9e237c65de4dc235e7085758/api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java#L155).  That exception causes the entire W3C TraceContext payload to be discarded!

The W3C spec allows for tracestate validation, with this text:
> The vendor MAY validate the tracestate header. If the tracestate header cannot be parsed the vendor MAY discard the entire header. Invalid tracestate entries MAY also be discarded.

But I don't believe it intends for vendors to discard traceparent and restart the trace.  
In addition, there is an open issue (https://github.com/w3c/trace-context/issues/384) to change that to a _SHOULD NOT_, rather than a _MAY_.

I propose that this project remove all validation of tracestate entries.

Or, at the very least, I propose that this project drop only invalid tracestate entries, rather than the entire W3C TraceContext payload.

jkwatson:
See https://github.com/open-telemetry/opentelemetry-specification/issues/478 for possible spec change.

carlosalberto:
Related to #895

bogdandrutu:
@jkwatson lazy parsing and validation of the tracestate is a nice addition to our project. I am happy to see a PR around that idea.

Oberon00:
I think a better design might be to propagate and store the tracestate more independent of trace+spanid on the Context. That would be a spec change though.

jkwatson:
I'd like to move this to post-GA, if that's ok with folks. Please let me know if you want me to move it back.

iNikem:
I would like to point out, that there are 2 different use cases here. One is while reading TraceState from the wire, another is during manipulating TraceState before propagation/export. The second use case is now explicitly addressed [by the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#tracestate):

> Every mutating operations MUST validate input parameters

[...]

PR #1561, opened 2020-08-19
Update the TraceState key validations to more closely match the w3c spec

### LINKED ISSUE #1027: validation glitches in TraceState.validateKey
I noticed that `TraceState.validateKey` is loose compared to the ABNF in the spec. There are a couple missing tests:

* when tenant format is used (`@`), the right-side can be 1-14 characters. this is not validated
* only when tenant format is used (`@`), the first character can be a digit. Otherwise, it must be a letter

There might be some other problems, but these I noticed. As the spec text doesn't elaborate on the above points, I tried to correct it FWIW https://github.com/w3c/trace-context/pull/386

jkwatson:
Resolves #1027

### DIFF
--- a/api/src/main/java/io/opentelemetry/trace/TraceState.java
+++ b/api/src/main/java/io/opentelemetry/trace/TraceState.java
@@ -254,24 +254,18 @@ public abstract class TraceState {
       if (isNotLegalKeyCharacter(c)) {
         return false;
       }
-      // the location of the '@' sign, or -1 if it isn't there.
-      int atPosition = -1;
       if (c == '@') {
         // you can't have 2 '@' signs
         if (isMultiTenantVendorKey) {
           return false;
         }
-        atPosition = i;
         isMultiTenantVendorKey = true;
         // tenant id (the part to the left of the '@' sign) must be 240 characters or less
         if (i > MAX_TENANT_ID_SIZE) {
           return false;
         }
-      }
-      if (isMultiTenantVendorKey) {
         // vendor id (the part to the right of the '@' sign) must be 13 characters or less
-        int numberOfCharactersPastTheAt = i - atPosition;
-        if (numberOfCharactersPastTheAt > MAX_VENDOR_ID_SIZE) {
+        if (key.length() - i > MAX_VENDOR_ID_SIZE) {
           return false;
         }
       }
--- a/api/src/main/java/io/opentelemetry/trace/TraceState.java
+++ b/api/src/main/java/io/opentelemetry/trace/TraceState.java
@@ -46,6 +46,8 @@ public abstract class TraceState {
   private static final int VALUE_MAX_SIZE = 256;
   private static final int MAX_KEY_VALUE_PAIRS = 32;
   private static final TraceState DEFAULT = TraceState.builder().build();
+  private static final int MAX_TENANT_ID_SIZE = 240;
+  public static final int MAX_VENDOR_ID_SIZE = 13;
 
   /**
    * Returns the default {@code TraceState} with no entries.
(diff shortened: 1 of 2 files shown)


carlosalberto (approved):

thisthat (approved):

api/src/main/java/io/opentelemetry/trace/TraceState.java
```diff
@@ -235,26 +235,53 @@ public static Entry create(String key, String value) {
...
+          && c != '_'
+          && c != '-'
+          && c != '@'
+          && c != '*'
+          && c != '/') {
         return false;
       }
-      if ((c == '@') && (++atSeenCount > 1)) {
-        return false;
+      int atPosition = 0;
```
anuraaga, line 258: I'm having trouble grokking this code.

anuraaga: Sorry accidentally hit Add single comment, continuing below

api/src/main/java/io/opentelemetry/trace/TraceState.java
```diff
@@ -235,26 +235,53 @@ public static Entry create(String key, String value) {
...
+      if (c == '@') {
+        atPosition = i;
+        // tenant id must be 240 characters or less
+        if (i > 240) {
+          return false;
+        }
+        atSeenCount++;
+        if (atSeenCount > 1) {
+          return false;
+        }
```
anuraaga, line 268: Instead of having `atPosition` is it ok to do something like this in this branch?

```java
// remaining is vendor id, must be 13 characters or less
if (key.length() - i > 13) {
  return false;
}
```

api/src/main/java/io/opentelemetry/trace/TraceState.java
```diff
@@ -235,26 +235,53 @@ public static Entry create(String key, String value) {
...
       }
+      if (atSeenCount == 1) {
+        // vendor id must be 13 characters or less
+        if ((i - atPosition) > 13) {
```
anuraaga, line 272: I may just be misgrokking, but isn't `atPosition` `0` for most of the loop rather than the actual `atPosition`? If my reading is right, my above suggestion would probably solve it but it also means there's a unit test missing for the current buggy state.

api/src/main/java/io/opentelemetry/trace/TraceState.java
```diff
@@ -235,26 +235,53 @@ public static Entry create(String key, String value) {
...
+          return false;
+        }
+      }
+    }
+    if (atSeenCount == 0) {
+      // if it's not the vendor format (with an '@' sign), the key must start with a letter.
+      return !Character.isDigit(key.charAt(0));
```
anuraaga, line 279: Also, guess we should avoid unicode comparisons like this and stick with simple ones we implement

api/src/main/java/io/opentelemetry/trace/TraceState.java
thisthat: `atSeenCount` is incremented only when `atPosition` changes to `i`. This way, this if is checked only for `atPosition` > 0.

anuraaga: Yeah but `atPosition` is declared inside the loop, which I think is the original bug. So except for when key[i] == '@', `atPosition == 0` for the remainder of the loop since it's initialized to `0`. Sorry if I'm missing it though I've tried to stare long and this is what I came up with.

thisthat: Oh! You are totally right! I miss that bit! We are indeed missing a test here!

jkwatson: I'm not following the problem. Can you give me a key that we're not handling correctly, so I can add a unit test for the issue?

api/src/main/java/io/opentelemetry/trace/TraceState.java
jkwatson: that's fair. I'll extract the digit logic into a separate method and us that.

api/src/main/java/io/opentelemetry/trace/TraceState.java
jkwatson: Note, I'm rewriting this chunk of code to make it clearer. But, having an example key that we're not handling correctly would be very useful (because I don't see the issue).

jkwatson: I've pushed the rewrite. Let me know if that helps at all with understanding what's going on in here. 

And, please give me a concrete key example case we're not handling correctly if there is one.  :)

anuraaga: I'm pretty sure `12345678901234567890@nr` is valid but fails when I added this to your branch. Basically any time the tenant id is longer than 13 characters, since the current logic is resetting the atPosition to the beinning of the string.

My [suggestion](https://github.com/open-telemetry/opentelemetry-java/pull/1561#discussion_r473635930) of not bothering with `atPosition` still stands :) 

```


  @Test
  void testValidLongVendorKey() {
    TraceState result = EMPTY.toBuilder().set("12345678901234567890@nr", FIRST_VALUE).build();
    assertThat(result.get("12345678901234567890@nr")).isEqualTo(FIRST_VALUE);
  }
```

Oberon00:
I think we should revert this change. We should follow the "be liberal in what you accept" guideline here, otherwise any deployed opentelemetry-java SDK will break tracestate entries that make use of the relaxed grammar in the current draft spec (once it graduates from being a draft).

jkwatson:
> I think we should revert this change. We should follow the "be liberal in what you accept" guideline here, otherwise any deployed opentelemetry-java SDK will break tracestate entries that make use of the relaxed grammar in the current draft spec (once it graduates from being a draft).

If we want to truly "be liberal in what we accept", we should drop all validations of the trace state entries and just pass them through un-molested. See this issue: https://github.com/open-telemetry/opentelemetry-java/issues/876

PR #1563, opened 2020-08-19
prevent parent spanContext from being null in Samplers

### LINKED ISSUE #1487: Sampler.shouldSample can get a parent SpanContext of null
Samplers get `null` for their parent `SpanContext` in `shouldSample` if the parent is "invalid". This seems to be the only place in the whole SDK+API where a SpanContext can be null. Everywhere else we pass an invalid/empty context in such a case. I think the sampling API should behave the same way.

`null` assigned here if the parent is invalid:

https://github.com/open-telemetry/opentelemetry-java/blob/78ed6490ba47568f5241604ea34f18d3ae6f3c4a/sdk/src/main/java/io/opentelemetry/sdk/trace/SpanBuilderSdk.java#L214

then passed to `shouldSample` here:

https://github.com/open-telemetry/opentelemetry-java/blob/78ed6490ba47568f5241604ea34f18d3ae6f3c4a/sdk/src/main/java/io/opentelemetry/sdk/trace/SpanBuilderSdk.java#L231-L232

jarebudev:
resolves #1487

### DIFF
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/SpanBuilderSdk.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/SpanBuilderSdk.java
@@ -213,11 +213,9 @@ final class SpanBuilderSdk implements Span.Builder {
     TraceId traceId;
     SpanId spanId = idsGenerator.generateSpanId();
     TraceState traceState = TraceState.getDefault();
-    if (parentContext == null || !parentContext.isValid()) {
+    if (!parentContext.isValid()) {
       // New root span.
       traceId = idsGenerator.generateTraceId();
-      // This is a root span so no remote or local parent.
-      parentContext = null;
     } else {
       // New child span.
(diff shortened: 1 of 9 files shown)


sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/SpanBuilderSdk.java
```diff
@@ -295,12 +293,11 @@ private static Clock getClock(Span parent, Clock clock) {
     }
   }
 
-  @Nullable
   private static SpanContext parent(
       ParentType parentType, Span explicitParent, SpanContext remoteParent) {
     switch (parentType) {
       case NO_PARENT:
-        return null;
+        return SpanContext.getInvalid();
```
anuraaga, line 300: :+1:

anuraaga (approved):

thisthat (approved):
Thank you! :)

carlosalberto (approved):

PR #1565, opened 2020-08-20
Fix to export entire batch sequentially instead of dropping spans.

anuraaga:
The `BatchSpanProcessor` has two aspects, a queue size and a batch size. When time comes, it tries to export the entire queue, sending a batch at a time. Because of our exporter availability check, though, batches within one flush of the queue are mostly guaranteed to not be sent, instead we should make sure these batches are sent out sequentially.

### DIFF
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
@@ -272,6 +272,7 @@ public final class BatchSpanProcessor implements SpanProcessor {
         }
       } else {
         logger.log(Level.FINE, "Exporter busy. Dropping spans.");
+        droppedSpans.add(spanList.size());
       }
     }
 
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
@@ -35,6 +35,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Timer;
 import java.util.TimerTask;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -255,12 +257,21 @@ public final class BatchSpanProcessor implements SpanProcessor {
       exportBatches(spansCopy);
     }
 
+    @SuppressWarnings("BooleanParameter")
     private void exportBatches(ArrayList<ReadableSpan> spanList) {
       // TODO: Record a counter for pushed spans.
-      for (int i = 0; i < spanList.size(); ) {
-        int lastIndexToTake = Math.min(i + maxExportBatchSize, spanList.size());
-        onBatchExport(createSpanDataForExport(spanList, i, lastIndexToTake));
-        i = lastIndexToTake;
+      if (exportAvailable.compareAndSet(true, false)) {
+        try {
+          for (int i = 0; i < spanList.size(); ) {
+            int lastIndexToTake = Math.min(i + maxExportBatchSize, spanList.size());
+            onBatchExport(createSpanDataForExport(spanList, i, lastIndexToTake));
+            i = lastIndexToTake;
+          }
+        } finally {
+          exportAvailable.set(true);
+        }
+      } else {
+        logger.log(Level.FINE, "Exporter busy. Dropping spans.");
       }
     }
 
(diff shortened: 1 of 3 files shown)


sdk/tracing/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorTest.java
```diff
@@ -171,15 +172,23 @@ void exportMoreSpansThanTheBufferSize() {
...
-        .containsExactly(
-            span1.toSpanData(),
-            span2.toSpanData(),
-            span3.toSpanData(),
-            span4.toSpanData(),
-            span5.toSpanData(),
-            span6.toSpanData());
+
+    // Give time for the BatchSpanProcessor to attempt to export spans.
+    Thread.sleep(500);
```
anuraaga, line 177: I generally don't like `Thread.sleep` in tests but this is the best I could come up with here.

iNikem (approved):

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -276,34 +287,34 @@ private void exportBatches(ArrayList<ReadableSpan> spanList) {
...
+            });
+        timer.schedule(
+            new TimerTask() {
+              @Override
+              public void run() {
+                result.fail();
+              }
+            },
+            exporterTimeoutMillis);
+        latch.await(exporterTimeoutMillis, TimeUnit.MILLISECONDS);
```
huntc, line 309: In essence, we're re-introducing blocking behaviour. I see a better way, but need think a bit more...

anuraaga: Yeah, the way I think about it is BSP is itself an asynchronous span processor since it uses a background thread to process a queue. It has two main properties, a queue size, and a export batch size. If the queue is full, spans are dropped. If the backend cannot support large batches, the export batch size can be made smaller than the queue size. This is to prevent too much pressure on the backend, so blocking is probably the right behavior (sending all the mini batches concurrently is presumably at least as bad as the large batch size the setting tries to avoid), we literally want to slow down the export for the backend. And if it's too slow, then the queue will fill up and spans will be dropped - but that's the balancing act of the backend performance vs ingestion rate. Because BSP uses a background thread, it's impossible for this blocking to leak into the app itself.

If we remove the batch size knob, then this wouldn't be an issue, but I don't know if we need to worry enough to remove the knob.

Anyways, that's my current context, of course happy to find an even better way :) This seemed like the only way to materialize these properties of the BSP though.

huntc:
@anuraaga I've put together the following without any testing, mostly to provoke some further conversation: https://github.com/open-telemetry/opentelemetry-java/pull/1567

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -255,12 +257,21 @@ private void forceFlush() {
       exportBatches(spansCopy);
     }
 
+    @SuppressWarnings("BooleanParameter")
```
jkwatson, line 260: was this yelling for usage of the AtomicBoolean? :(

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -255,12 +257,21 @@ private void forceFlush() {
...
+          for (int i = 0; i < spanList.size(); ) {
+            int lastIndexToTake = Math.min(i + maxExportBatchSize, spanList.size());
+            onBatchExport(createSpanDataForExport(spanList, i, lastIndexToTake));
+            i = lastIndexToTake;
+          }
+        } finally {
+          exportAvailable.set(true);
+        }
+      } else {
+        logger.log(Level.FINE, "Exporter busy. Dropping spans.");
```
jkwatson, line 274: we should increment the droppedSpans counter when this happens.

sdk/tracing/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorTest.java
jkwatson: is there a way to use awaitility to get rid of this? I assume there isn't anything obvious or you would have done it. :)

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
huntc: If you want a really quick fix then just block waiting for the CompleteableResultCode to finish. Nothing else is required. 

Blocking is a poor outcome here that can be easily avoided as per my other PR.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
anuraaga: Yeah :(

sdk/tracing/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorTest.java
anuraaga: Couldn't find anything myself, was hoping someone else would find it but this may be best...

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
anuraaga: Ah because we only have a single thread, if we're blocking then being unavailable isn't possible. Removed that check

anuraaga:
Closing, let's focus on the overall BSP simplification :)

Issue #1560, opened 2020-08-19
Replace usages of java.util.Timer with a ScheduledExecutor in the BatchSpanProcessor


jkwatson:
The Timer has known issues with clock changes. We should use ScheduledExecutors as they are the more modern construct for doing timed executions.

anuraaga:
Is this a dupe of #729?

malafeev:
I can do it, please assign to me to proceed.

iNikem:
There are already 2 PRs addressing this issue :)

PR #1567, opened 2020-08-20
Only drop elements exported by BatchSpanProcessor

### LINKED ISSUE #1566: Leverage BatchSpanProcessor's span dropping functionality

huntc:
This PR takes an approach where only the actual elements passed onto an exporter is removed from the internal buffer of spans within a `BatchSpanProcessor`. Any pressure on the buffer becoming full will result in the `BatchSpanProcessor` dropping elements as per its original design.

Prior to this PR, spans could be dropped by the `BatchSpanProcessor` without being reported.

Along the way, it occurred to me that the worker thread can be replaced by using the timer thread that we have to have. I've now affected code changes there too, which I think simplifies the `BatchSpanProcessor`.

Fixes #1566 

cc @anuraaga

### DIFF
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
@@ -47,8 +47,7 @@ import javax.annotation.concurrent.GuardedBy;
  * <p>All spans reported by the SDK implementation are first added to a synchronized queue (with a
  * {@code maxQueueSize} maximum size, after the size is reached spans are dropped) and exported
  * every {@code scheduleDelayMillis} to the exporter pipeline in batches of {@code
- * maxExportBatchSize}. Spans may also be dropped when it becomes time to export again, and there is
- * an export in progress.
+ * maxExportBatchSize}.
  *
  * <p>If the queue gets half full a preemptive notification is sent to a worker that exports the
  * spans to wake up and start a new export cycle.
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
@@ -232,10 +232,12 @@ public final class BatchSpanProcessor implements SpanProcessor {
             } while (spansList.isEmpty());
           }
           spansCopy = new ArrayList<>(spansList);
-          spansList.clear();
         }
         // Execute the batch export outside the synchronized to not block all producers.
-        exportBatches(spansCopy);
+        int taken = onBatchExport(spansCopy);
+        synchronized (monitor) {
+          spansList.retainAll(spansList.subList(taken, spansList.size()));
+        }
       }
     }
 
@@ -249,18 +251,11 @@ public final class BatchSpanProcessor implements SpanProcessor {
       ArrayList<ReadableSpan> spansCopy;
       synchronized (monitor) {
         spansCopy = new ArrayList<>(spansList);
-        spansList.clear();
       }
       // Execute the batch export outside the synchronized to not block all producers.
-      exportBatches(spansCopy);
-    }
-
-    private void exportBatches(ArrayList<ReadableSpan> spanList) {
-      // TODO: Record a counter for pushed spans.
-      for (int i = 0; i < spanList.size(); ) {
-        int lastIndexToTake = Math.min(i + maxExportBatchSize, spanList.size());
-        onBatchExport(createSpanDataForExport(spanList, i, lastIndexToTake));
-        i = lastIndexToTake;
+      int taken = onBatchExport(spansCopy);
+      synchronized (monitor) {
+        spansList.retainAll(spansList.subList(taken, spansList.size()));
       }
     }
 
@@ -277,10 +272,12 @@ public final class BatchSpanProcessor implements SpanProcessor {
 
     // Exports the list of SpanData to the SpanExporter.
     @SuppressWarnings("BooleanParameter")
-    private void onBatchExport(final List<SpanData> spans) {
+    private int onBatchExport(final List<ReadableSpan> spans) {
       if (exportAvailable.compareAndSet(true, false)) {
+        int exportSize = Math.min(maxExportBatchSize, spans.size());
         try {
-          final CompletableResultCode result = spanExporter.export(spans);
+          final CompletableResultCode result =
+              spanExporter.export(createSpanDataForExport(spans, 0, exportSize));
           result.whenComplete(
               new Runnable() {
                 @Override
@@ -302,8 +299,9 @@ public final class BatchSpanProcessor implements SpanProcessor {
         } catch (Exception e) {
           logger.log(Level.WARNING, "Exporter threw an Exception", e);
         }
+        return exportSize;
       } else {
-        logger.log(Level.FINE, "Exporter busy. Dropping spans.");
+        return 0;
       }
     }
   }
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
@@ -82,6 +82,7 @@ import javax.annotation.concurrent.GuardedBy;
  */
 public final class BatchSpanProcessor implements SpanProcessor {
 
+  // FIXME This worker thread can be eliminated by leveraging the timer
   private static final String WORKER_THREAD_NAME =
       BatchSpanProcessor.class.getSimpleName() + "_WorkerThread";
   private static final String TIMER_THREAD_NAME =
@@ -217,7 +218,6 @@ public final class BatchSpanProcessor implements SpanProcessor {
         // avoid blocking the producer thread.
         ArrayList<ReadableSpan> spansCopy;
         synchronized (monitor) {
-          // If still maxExportBatchSize elements in the queue better to execute an extra
           do {
             // In the case of a spurious wakeup we export only if we have at least one span in
             // the batch. It is acceptable because batching is a best effort mechanism here.
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
@@ -23,13 +23,13 @@ import io.opentelemetry.internal.Utils;
 import io.opentelemetry.metrics.LongCounter;
 import io.opentelemetry.metrics.LongCounter.BoundLongCounter;
 import io.opentelemetry.metrics.Meter;
-import io.opentelemetry.sdk.common.DaemonThreadFactory;
 import io.opentelemetry.sdk.common.export.CompletableResultCode;
 import io.opentelemetry.sdk.common.export.ConfigBuilder;
 import io.opentelemetry.sdk.trace.ReadableSpan;
 import io.opentelemetry.sdk.trace.SpanProcessor;
 import io.opentelemetry.sdk.trace.data.SpanData;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -83,12 +83,9 @@ import javax.annotation.concurrent.GuardedBy;
 public final class BatchSpanProcessor implements SpanProcessor {
 
   // FIXME This worker thread can be eliminated by leveraging the timer
-  private static final String WORKER_THREAD_NAME =
-      BatchSpanProcessor.class.getSimpleName() + "_WorkerThread";
   private static final String TIMER_THREAD_NAME =
       BatchSpanProcessor.class.getSimpleName() + "_TimerThread";
   private final Worker worker;
-  private final Thread workerThread;
   private final boolean sampled;
 
   private BatchSpanProcessor(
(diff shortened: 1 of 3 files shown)


sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -232,10 +232,12 @@ public void run() {
             } while (spansList.isEmpty());
           }
           spansCopy = new ArrayList<>(spansList);
-          spansList.clear();
         }
         // Execute the batch export outside the synchronized to not block all producers.
-        exportBatches(spansCopy);
+        int taken = onBatchExport(spansCopy);
+        synchronized (monitor) {
+          spansList.retainAll(spansList.subList(taken, spansList.size()));
```
huntc, line 239: There should be a better way of doing this. In Scala, I'd resort to using `drop(taken)`.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -277,10 +272,12 @@ private void exportBatches(ArrayList<ReadableSpan> spanList) {
 
     // Exports the list of SpanData to the SpanExporter.
     @SuppressWarnings("BooleanParameter")
-    private void onBatchExport(final List<SpanData> spans) {
+    private int onBatchExport(final List<ReadableSpan> spans) {
```
huntc, line 275: I'm thinking of renaming `onBatchExport` to `exportBatch` if we proceed with these changes

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -302,8 +299,9 @@ public void run() {
         } catch (Exception e) {
           logger.log(Level.WARNING, "Exporter threw an Exception", e);
         }
+        return exportSize;
       } else {
-        logger.log(Level.FINE, "Exporter busy. Dropping spans.");
+        return 0;
```
huntc, line 304: Returning zero is key - this is what'll cause the internal buffer to mount up and for the existing span-drop functionality of logging to kick in.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -232,10 +232,12 @@ public void run() {
             } while (spansList.isEmpty());
           }
           spansCopy = new ArrayList<>(spansList);
-          spansList.clear();
         }
         // Execute the batch export outside the synchronized to not block all producers.
-        exportBatches(spansCopy);
+        int taken = onBatchExport(spansCopy);
```
anuraaga, line 237: I might be missing it, but I think this will cause multiple batches to be sent concurrently? If so, I think it's not respecting the spirit of `maxExportBatchSize`. To be honest, I don't know exactly what the spirit of `maxExportBatchSize` is but I think it's one of 1) send batches sequentially or 2) wait `scheduledDelayMillis` between each chunk of `maxExportBatchSize` no matter what. #1565 aims for 1), neither of our PRs aim for 2) :( But I guess if we're sending concurrently that's not good for BSP, that's the SSP's role. Let me know if I'm confused though.

anuraaga (commented):
Because of

anuraaga:
Because of https://github.com/open-telemetry/opentelemetry-java/pull/1567#discussion_r474003638 my current recommendation is to merge #1565 since we do need a quick fix for BSP dropping spans that shouldn't be and it seems the simplest. All of the related PRs have made me believe we need to prioritize #1560 #729 before doing a deep dive here

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
huntc: It isn’t sending an export concurrently. Only one export is permitted at a time within the onExportBatch function.

huntc:
> Because of #1567 (comment) my current recommendation is to merge #1565 since we do need a quick fix for BSP dropping spans that shouldn't be and it seems the simplest. All of the related PRs have made me believe we need to prioritize #1560 #729 before doing a deep dive here

I see this PR as being much simpler than 1565. The BSP’s contract acknowledges that spans can be dropped, so we’ve not violated that contract. Let’s get this right rather than look for a quick fix.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -232,10 +232,12 @@ public void run() {
             } while (spansList.isEmpty());
           }
           spansCopy = new ArrayList<>(spansList);
-          spansList.clear();
```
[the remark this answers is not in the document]
huntc: I believe that with this PR the contract from the spec that you reference is being met. I think the BSP is actually in good shape.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
huntc: Exporting “no matter what” on schedule is what things look like before this PR. ;-)

anuraaga: Ah I see how the return 0 works. Isn't this effectively a spinlock because we'll keep on looping, find the queue is larger than batch size, try to export, find it's not available, and continue until it's available again? I wouldn't expect to see spinning to wait for I/O, is it better than parking?

huntc: Good point. It would indeed spin until the number of buffer elements drops below the max buffer size. I think we should move the condition on the wait also... that way it wouldn't spin. The batch would then just wake up on its timer no matter what and then try another export. This would ultimately represent the park that you suggested.

I will push another commit now along with some other small changes to improve comprehension.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -177,7 +177,7 @@ public void forceFlush() {
     private final AtomicBoolean exportAvailable = new AtomicBoolean(true);
 
     @GuardedBy("monitor")
-    private final List<ReadableSpan> spansList;
+    private final ArrayList<ReadableSpan> spansList;
```
huntc, line 180: Exposing an array list here as the operations we perform on it assumes an array in terms of bulk copying and therefore, o(1) element access.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -82,6 +82,7 @@
  */
 public final class BatchSpanProcessor implements SpanProcessor {
 
+  // FIXME This worker thread can be eliminated by leveraging the timer
   private static final String WORKER_THREAD_NAME =
```
huntc, line 86: I just noticed that the worker thread can now be eliminated as we can leverage the timer thread. I'll do this in a separate PR if you accept this one. ;-)

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -217,25 +218,24 @@ public void run() {
...
-                Thread.currentThread().interrupt();
-                return;
-              }
-            } while (spansList.isEmpty());
-          }
+          do {
+            // In the case of a spurious wakeup we export only if we have at least one span in
+            // the batch. It is acceptable because batching is a best effort mechanism here.
+            try {
+              monitor.wait(scheduleDelayMillis);
```
anuraaga, line 225: I think this causes the wait to happen between each export batch, rather than between each "processing of the queue". It effectively means the queue size parameter is not having much effect if I'm not mistaken. 

FWIW, I am interested in flushing in an async manner :) I think the way to do it is to have a separate Flusher. What I'm expecting is for it to look like something like

```java
ExecutorService executor;

init() executor.schedule(wakeup);

wakeup {
  if (spans.empty)
    executor.schedule(wakeup)

  drained = spans.drain

  new Flusher(drained).flush();

Flusher {
  pending

  Flusher(spans) pending = spans
  
  flush() {
    if (pending.isEmpty()) {
      executor.schedule(wakeup)
      return
    }

    taken = pending.take(batchSize)
    result = exporter.export(taken)
    result.whenComplete(this::flush)
}
```

I think this has the properties that we want - run through entire queue when exporting, export a batch at a time, and delay between each queue processing (not between every batch as in the right side here, nor potentially no delay ever like left side here). And consecutive flushes without blocking.

So my plan is #1565, then #729, then add AsyncFlusher. @jkwatson has wanted us to migrate to an `ExecutorService` for a long time and I agree :) I think it will make our lives simpler when actually addressing the asyncronous export. Does this plan make sense?

huntc: @anuraaga I need to absorb your comments above more and will do so now. However, #1565 does not make sense to me as it introduces additional complexity that is avoidable. Back soon.

huntc: So, if I've got this right... you're making a distinction here between the time interval in terms of processing the queue vs the time interval between exporting to an exporter. You're also favouring the former.

You want to see exporting happen continually if there are more elements in the queue.

Correct?

anuraaga: Yeah - at least that's my understanding of the spec / goal of the class, but I could be wrong in my reading of course :)

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#batching-processor

A second reading makes me think actually this PR may have the right semantics since it says scheduledDelayMillis = delay between export, and maxExportBatchSize = maximum size of an export. I'm having trouble invisioning how I would actually set maxExportBatchSize different from queueSize but this PR seems ok to me too now. Interested in hearing what others think the BSP is "supposed to do" though :)

huntc: OK, I've just pushed another commit.

The worker thread has been removed and we now instead use the timer for scheduling work, thus saving us a thread. In addition, if there remains work to be done once the export has completed then we queue it for running immediately.

A test has changed slightly in that it was previously relying on the worker thread becoming blocked; which it now no longer does.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -302,10 +296,19 @@ public void run() {
...
       }
     }
+
+    private static void dropSpans(ArrayList<ReadableSpan> spansList, int count) {
+      int remaining = spansList.size() - count;
+      ReadableSpan[] retainedSpanList = new ReadableSpan[remaining];
+      System.arraycopy(spansList.toArray(), count, retainedSpanList, 0, remaining);
+      spansList.clear();
+      spansList.addAll(Arrays.asList(retainedSpanList));
+    }
```
huntc, line 311: There has to be a better way of dropping elements from an array... Given that `clear` traverses the array and nulls elements, we've effectively got three traverses of the array. Now in practical terms, `retainedSpanList ` should be empty most times, or at least only a few elements. So the array copying at L308 and L310 is going to be negligible.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -289,6 +276,13 @@ public void run() {
                     logger.log(Level.FINE, "Exporter failed");
                   }
                   exportAvailable.set(true);
+
+                  synchronized (monitor) {
+                    // If there are still more to process then schedule our task immediately
+                    if (!spansList.isEmpty()) {
+                      timer.schedule(new WorkerTask(), 0);
+                    }
+                  }
```
huntc, line 285: We now steal more work if there's some more available.

huntc:
`stressTest_withBatchSpanProcessor` is failing - I shall look at that next

sdk/tracing/src/test/java/io/opentelemetry/sdk/trace/TracerSdkTest.java
```diff
@@ -151,8 +151,13 @@ void stressTest_withBatchSpanProcessor() {
...
     }
 
+    // Needs to correlate with the BatchSpanProcessor.Builder's default, which is the only thing
+    // this test can guarantee
+    final int defaultMaxQueueSize = 2048;
+
     stressTestBuilder.build().run();
-    assertThat(countingSpanExporter.numberOfSpansExported.get()).isEqualTo(8_000);
+    assertThat(countingSpanExporter.numberOfSpansExported.get())
+        .isGreaterThanOrEqualTo(defaultMaxQueueSize);
```
huntc, line 160: This needed to change given that we are now fully async. We can't guarantee that all spans get through. This is ok though as it is in line with the `BatchSpanProcessor`'s contract.

In reality, this test does see several thousand get through.

anuraaga:
@huntc Chatted with @jkwatson today about this and we decided that because the current behavior is only in snapshot, not released, we don't have to rush a fix in. So going to close my #1565 :-) He also think we really should go ahead and clean up the processor before (or perhaps while) fixing it rather than continuing work with the current monitor / worker pattern. I'm working on a PR for that. @huntc does that sound OK to you? I'm keeping this PR in mind instead of adding any blocking - I think the code will be dramatically simplified and hope it helps us with future possible improvements.

huntc:
Closing given the work of #1571

PR #1571, opened 2020-08-21
Make sure forceFlush / shutdown can have callers wait for them to be done by returning CompletableResultCode.

### LINKED ISSUE #729: Simplify the BatchSpansProcessor

### LINKED ISSUE #1560: Replace usages of java.util.Timer with a ScheduledExecutor in the BatchSpanProcessor

### LINKED ISSUE #1566: Leverage BatchSpanProcessor's span dropping functionality

### LINKED ISSUE #1584: Non-blocking SpanProcessor interface
During code reviews of #1579 #1571 the opinion was expressed that `SpanProcessor` interface should allow for non-blocking usage. More precisely `SpanProcessor.forceFlush` and `SpanProcessor.shutdown` should return some sort of deferred result, such as `CompletableResultCode` or, if java8+ is available, `CompletableFuture`. This will allow for implementations more suitable for reactive environments which do not play well with blocking code.

`SpanExporter#shutdown` should also be changed in the same way.

I think this does not contradict spec, it is purely an implementation detail.

anuraaga:
/cc @huntc

Fixes #1584


sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -132,180 +160,72 @@ public boolean isEndRequired() {
...
-              },
-              exporterTimeoutMillis);
-        } catch (Exception e) {
-          logger.log(Level.WARNING, "Exporter threw an Exception", e);
-        }
-      } else {
-        logger.log(Level.FINE, "Exporter busy. Dropping spans.");
-      }
-    }
+  private ScheduledFuture<?> scheduleNextExport() {
```
iNikem, line 219: Shouldn't this always update `nextExport`?

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -132,180 +160,72 @@ public boolean isEndRequired() {
...
-      for (int i = startIndex; i < endIndex; i++) {
-        spanDataBuffer.add(spanList.get(i).toSpanData());
-        // Remove the reference to the ReadableSpan to allow GC to free the memory.
-        spanList.set(i, null);
-      }
-      return Collections.unmodifiableList(spanDataBuffer);
-    }
+        });
+
+    // Timeout if result doesn't complete soon enough.
```
iNikem, line 207: Not specific to this PR: Why this timeout is the configuration of BSP and not of the exporter?

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -132,180 +160,72 @@ public boolean isEndRequired() {
...
-      this.maxQueueSize = maxQueueSize;
-      this.halfMaxQueueSize = maxQueueSize >> 1;
-      this.maxExportBatchSize = maxExportBatchSize;
-      this.exporterTimeoutMillis = exporterTimeoutMillis;
-      this.spansList = new ArrayList<>(maxQueueSize);
+  // Timeout is best effort don't need to handle the future return value.
+  @SuppressWarnings("FutureReturnValueIgnored")
+  private void exportBatch() {
+    List<ReadableSpan> batch = new ArrayList<>(Math.min(maxExportBatchSize, queue.size()));
+    if (queue.drainTo(batch, maxExportBatchSize) == 0) {
```
iNikem, line 181: Should benchmark this, but we can save on allocations if we take elements by one, convert them to `SpanData` and only then put to outgoing `forExport`. Then we don't need `batch`.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -132,180 +160,72 @@ public boolean isEndRequired() {
...
-        }
-        // TODO: Record a gauge for referenced spans.
-        spansList.add(span);
-        // Notify the worker thread that at half of the queue is available. It will take
-        // time anyway for the thread to wake up.
-        if (spansList.size() >= halfMaxQueueSize) {
-          monitor.notifyAll();
-        }
-      }
+    List<SpanData> forExport = new ArrayList<>(batch.size());
```
iNikem, line 186: Can we reuse this list and not re-allocate it on every export?

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -132,180 +160,72 @@ public boolean isEndRequired() {
...
+        new Runnable() {
+          @Override
+          public void run() {
+            if (!result.isSuccess()) {
+              logger.log(Level.FINE, "Exporter failed");
+            }
+            if (queue.size() >= maxExportBatchSize) {
+              exportBatch();
+            } else {
+              scheduleNextExport();
```
iNikem, line 202: I am concerned, that if any error situation prevents the runtime to reach this line for one export attempt, then the whole BSP will stop working. It seem that current design (schedule next export when this one finishes) tries to achieve "fixed delay". If we switch to "fixed rate" we can simplify it. I don't see why "fixed rate" is worse than "fixed delay".

anuraaga:
Went ahead and changed get to join and don't throw exceptions from it. The majority of this PR is actually making sure `forceFlush` can actually be waited on. Without changing the API for `SpanProcessor.forceFlush`, if we were to stick with void, it's not possible to do an async flush so increasing the flexibility does seem to make sense. But it's definitely a far-reaching change.

anuraaga:
My current thinking after comparing with #1579 is it could make sense to keep the default `BatchSpanProcessor` a simple implementation that blocks between exports. In my own app, I'd probably swap it out with a custom non-blocking span processor that uses a fastutil MpscQueue and a netty event loop for scheduling anyways (I've done this for my usage of brave for a long time).

One of the big problems I have with this PR is avoiding a stack overflow - in practice even when directly exporting the next batch of spans, it would not result in stack overflows for proper asynchronous exporters. But there's no way to guarantee the exporter is asynchronous, and `BatchSpanProcessor` should be generic enough to work with synchronous exporters too. I think the workaround I did to use `executor.submit` without a schedule to immediately execute the next batch probably has at least as much overhead as just a normal blocking call within a loop. 

I think we would need the API change to `SpanProcessor` regardless though since it's not possible to even swap in a non-blocking span processor right now where `forceFlush` can't be invoked asynchronously.

iNikem:
I don't understand why `flush` should return a future (or any analogue). What are those 2 different use-cases where you want both blocking and non-blocking flush? The purpose of flush is to handle FaaS. I think that use-case requires blocking flush. What are others?

anuraaga:
They're both FaaS :) But it depends on the implementation of the FaaS runtime, whether the only way to prevent freezing is blocking (which is the norm right now) or it exposes a callback to be notified when to freeze - the latter would allow asynchronous apps (e.g. something using just reactive streams) which isn't allowed by a blocking void `flush` command.

iNikem:
Are there such FaaS providers already? Or we just speculate and prepare for a future (pun intended) :)

anuraaga:
I've heard such a future is coming ;) We could also add `flushAsync` later if we want to punt on it, but I think we may still at least need to speculatively move `CompletableResultCode` up to allow that.

sdk/tracing/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorTest.java
```diff
@@ -171,25 +174,35 @@ void exportMoreSpansThanTheBufferSize() {
...
-        .containsExactly(
-            span1.toSpanData(),
-            span2.toSpanData(),
-            span3.toSpanData(),
-            span4.toSpanData(),
-            span5.toSpanData(),
-            span6.toSpanData());
+
+    // Give time for the BatchSpanProcessor to attempt to export spans.
+    Thread.sleep(500);
```
huntc, line 179: Using `Thread.sleep` is fragile in tests

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -82,31 +83,61 @@
...
-    this.workerThread.start();
+    this.maxExportBatchSize = maxExportBatchSize;
+    this.scheduleDelayMillis = scheduleDelayMillis;
+    this.exporterTimeoutMillis = exporterTimeoutMillis;
     this.sampled = sampled;
+
+    this.spanExporter = spanExporter;
+    queue = new ArrayBlockingQueue<>(maxQueueSize);
+    executor =
+        Executors.newSingleThreadScheduledExecutor(new DaemonThreadFactory(WORKER_THREAD_NAME));
```
huntc, line 136: Can we pass in the executor here so that the BSP can share resources and leave the executor as an outside concern?

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -131,181 +164,160 @@ public boolean isEndRequired() {
...
+    closed = true;
+    nextExport.cancel(false);
+
+    final CompletableResultCode result = new CompletableResultCode();
+    final CompletableResultCode flushResult = forceFlush();
+    flushResult.whenComplete(
+        new Runnable() {
+          @Override
+          public void run() {
+            executor.shutdown();
```
huntc, line 177: The executor should be an outside concern

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -131,181 +164,160 @@ public boolean isEndRequired() {
...
-      exportBatches(spansCopy);
+  @SuppressWarnings("FutureReturnValueIgnored")
+  private void exportNextBatch(
+      final List<ReadableSpan> spans, final int offset, final CompletableResultCode result) {
+    int num = Math.min(maxExportBatchSize, spans.size() - offset);
+    final int limit = offset + num;
+    List<SpanData> forExport = new ArrayList<>(num);
+    for (int i = offset; i < limit; i++) {
+      forExport.add(spans.get(i).toSpanData());
     }
```
huntc, line 259: I'm thinking that a native array copy would be more efficient here... and then form an array list around that for calling the exporter. WDYT?

huntc (approved):
Looking good! I particularly like the async `forceFlush` now.

huntc:
> I don't understand why `flush` should return a future (or any analogue). What are those 2 different use-cases where you want both blocking and non-blocking flush? The purpose of flush is to handle FaaS. I think that use-case requires blocking flush. What are others?

FaaS isn't the only reason for flushing. Shutting down causes an implicit flush and then a join. Keep the joins (blocking) to a minimum otherwise it will become abused.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
anuraaga: TBH I've never compared performance of preallocated `ArrayList` with `.add` vs allocating an array and using `Arrays.asList`. I'll need to write a JMH to confirm the difference, since `i` doesn't start at `0` there is some extra cognitive load so I'd like to try that after confirming the perf difference.

sdk/tracing/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorTest.java
anuraaga: I agree - I couldn't find a way around it for this test though.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
huntc: I don't think there should be a fallback executor. The executor should always be an external concern IMHO. Users should be encouraged to think about these things... :-)

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -131,181 +164,160 @@ public boolean isEndRequired() {
...
-  public void shutdown() {
-    workerThread.interrupt();
-    worker.shutdown();
-  }
+  public CompletableResultCode shutdown() {
+    closed = true;
+    nextExport.cancel(false);
+
+    final CompletableResultCode result = new CompletableResultCode();
+    final CompletableResultCode flushResult = forceFlush();
```
jkwatson, line 172: It feels like there should be a way to compose the things in here more elegantly, but it could be done as a follow-on PR.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -131,181 +164,160 @@ public boolean isEndRequired() {
...
-      synchronized (monitor) {
-        spansCopy = new ArrayList<>(spansList);
-        spansList.clear();
-      }
-      // Execute the batch export outside the synchronized to not block all producers.
-      exportBatches(spansCopy);
+  @SuppressWarnings("FutureReturnValueIgnored")
+  private void exportNextBatch(
+      final List<ReadableSpan> spans, final int offset, final CompletableResultCode result) {
+    int num = Math.min(maxExportBatchSize, spans.size() - offset);
```
jkwatson, line 254: maybe `numberToExport`, rather than just `num`?

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -131,181 +164,160 @@ public boolean isEndRequired() {
...
-      }
+    final CompletableResultCode exportResult;
+    try {
+      exportResult = spanExporter.export(forExport);
+    } catch (Exception e) {
+      logger.log(Level.FINE, "Exporter failed", e);
+      maybeContinueBatch(spans, limit, result);
+      return;
     }
+    exportResult.whenComplete(
```
jkwatson, line 268: What do you think about putting this inside the try block? Seems like it might read a little more cleanly not to have to pre-declare the `exportResult`.

jkwatson:
This is more complex than I hoped, but I think it looks pretty solid. Needs a rebase, after the addition of the ReadWriteSpan.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
huntc: Lambdas ;-)

sdk/tracing/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorTest.java
huntc: > I agree - I couldn't find a way around it for this test though.

I had the same problem in my PR. The test should assert only what it can guarantee ie that it’ll flush the max buffer and nothing else. If we aren’t happy with that then we should remove the test entirely, as the required outcome can’t be achieved. 

Flaky tests will cause headaches for some unfortunate soul down the track.

huntc:
> This is more complex than I hoped, but I think it looks pretty solid. Needs a rebase, after the addition of the ReadWriteSpan.

FWIW I don’t use the BatchSpanProcessor. For me, batching and other concerns should as thread usage belong within the Exporter. I therefore use the SimpleSpanProcessor. I suspect that as other exporters become more sophisticated then they will also encourage the use of SimpleSpanProcessor. 

FYI I’ve quite a replacement for the OTel exporter that uses Akka streams.

jkwatson:
@anuraaga Looks like this needs to be updated after #1579 got merged.

anuraaga:
@jkwatson Updated

Issue #1572, opened 2020-08-21
Rename ParentOrElse to ParentBased and generalize to support all cases


jkwatson:
See spec updates from this: https://github.com/open-telemetry/opentelemetry-specification/pull/610

PR #1577, opened 2020-08-22
Rename ParentOrElse to ParentBased and generalize to support all cases

dengliming:
closes  [https://github.com/open-telemetry/opentelemetry-java/issues/1572](https://github.com/open-telemetry/opentelemetry-java/issues/1572)

### DIFF
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
@@ -249,13 +249,13 @@ public final class Samplers {
   @Immutable
   static class ParentBased implements Sampler {
     private final Sampler root;
-    private Sampler remoteParentSampled = alwaysOn();
-    private Sampler remoteParentNotSampled = alwaysOff();
-    private Sampler localParentSampled = alwaysOn();
-    private Sampler localParentNotSampled = alwaysOff();
+    private final Sampler remoteParentSampled;
+    private final Sampler remoteParentNotSampled;
+    private final Sampler localParentSampled;
+    private final Sampler localParentNotSampled;
 
     ParentBased(Sampler root) {
-      this.root = root;
+      this(root, alwaysOn(), alwaysOff(), alwaysOn(), alwaysOff());
     }
 
     ParentBased(
@@ -265,18 +265,12 @@ public final class Samplers {
         Sampler localParentSampled,
         Sampler localParentNotSampled) {
       this.root = root;
-      if (remoteParentSampled != null) {
-        this.remoteParentSampled = remoteParentSampled;
-      }
-      if (remoteParentNotSampled != null) {
-        this.remoteParentNotSampled = remoteParentNotSampled;
-      }
-      if (localParentSampled != null) {
-        this.localParentSampled = localParentSampled;
-      }
-      if (localParentNotSampled != null) {
-        this.localParentNotSampled = localParentNotSampled;
-      }
+      this.remoteParentSampled = remoteParentSampled == null ? alwaysOn() : remoteParentSampled;
+      this.remoteParentNotSampled =
+          remoteParentNotSampled == null ? alwaysOff() : remoteParentNotSampled;
+      this.localParentSampled = localParentSampled == null ? alwaysOn() : localParentSampled;
+      this.localParentNotSampled =
+          localParentNotSampled == null ? alwaysOff() : localParentNotSampled;
     }
 
     // If a parent is set, always follows the same sampling decision as the parent.
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
@@ -356,7 +356,7 @@ public final class Samplers {
             this.remoteParentSampled,
             this.remoteParentNotSampled,
             this.localParentSampled,
-            localParentNotSampled);
+            this.localParentNotSampled);
       }
 
       private Builder(Sampler root) {
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
@@ -298,7 +298,8 @@ public final class Samplers {
       private Sampler localParentNotSampled;
 
       /**
-       * Sets the {@link Sampler} to use when there is a remote parent that was sampled.
+       * Sets the {@link Sampler} to use when there is a remote parent that was sampled. If not set,
+       * defaults to always sampling if the remote parent was sampled.
        *
        * @return this Builder
        */
@@ -308,7 +309,8 @@ public final class Samplers {
       }
 
       /**
-       * Sets the {@link Sampler} to use when there is a remote parent that was not sampled.
+       * Sets the {@link Sampler} to use when there is a remote parent that was not sampled. If not
+       * set, defaults to never sampling when the remote parent isn't sampled.
        *
        * @return this Builder
        */
@@ -318,7 +320,8 @@ public final class Samplers {
       }
 
       /**
-       * Sets the {@link Sampler} to use when there is a local parent that was sampled.
+       * Sets the {@link Sampler} to use when there is a local parent that was sampled. If not set,
+       * defaults to always sampling if the local parent was sampled.
        *
        * @return this Builder
        */
@@ -328,7 +331,8 @@ public final class Samplers {
       }
 
       /**
-       * Sets the {@link Sampler} to use when there is a local parent that was not sampled.
+       * Sets the {@link Sampler} to use when there is a local parent that was not sampled. If not
+       * set, defaults to never sampling when the local parent isn't sampled.
        *
        * @return this Builder
        */
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
@@ -147,14 +147,47 @@ public final class Samplers {
    * whether or not to sample. If there is no parent, the Sampler uses the provided Sampler delegate
    * to determine the sampling decision.
    *
-   * @param delegateSampler the {@code Sampler} which is used to make the sampling decisions if the
-   *     parent does not exist.
+   * @param root the {@code Sampler} which is used to make the sampling decisions if the parent does
+   *     not exist.
    * @return a {@code Sampler} that follows the parent's sampling decision if one exists, otherwise
-   *     following the delegate sampler's decision.
+   *     following the root sampler's decision.
    * @since 0.7.0
    */
-  public static Sampler parentBased(Sampler delegateSampler) {
-    return new ParentBased(delegateSampler);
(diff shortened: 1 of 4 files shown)


jkwatson:
@dengliming can you either 1) remove the `closes` from the PR description or 2) address making sure the implementation fully addresses the spec?

Thanks!

dengliming:
@jkwatson Sorry. just forgot it. : )

jkwatson:
> @jkwatson Sorry. just forgot it. : ) . I will do the implementation fully addresses the spec recently. Can you assign to me?

Done, and thanks for taking that on!

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
```diff
@@ -215,10 +248,35 @@ public String getDescription() {
 
   @Immutable
   static class ParentBased implements Sampler {
-    private final Sampler delegateSampler;
+    private final Sampler root;
+    private Sampler remoteParentSampled = alwaysOn();
+    private Sampler remoteParentNotSampled = alwaysOff();
+    private Sampler localParentSampled = alwaysOn();
+    private Sampler localParentNotSampled = alwaysOff();
+
```
jkwatson, line 256: I'd prefer if these were `final` and assigned in the constructor. Then, the single-arg constructor can just pass these values in to the other constructor.

dengliming: agree 👍 . thanks.

sdk/tracing/src/test/java/io/opentelemetry/sdk/trace/SamplersTest.java
```diff
@@ -231,13 +261,75 @@ void parentOrElseSampler_AlwaysOn() {
...
+                    invalidSpanContext,
+                    traceId,
+                    SPAN_NAME,
+                    SPAN_KIND,
+                    Attributes.empty(),
+                    Collections.emptyList())
+                .getDecision())
+        .isEqualTo(Decision.RECORD_AND_SAMPLED);
+
+    // Sampled remote parent
```
jkwatson, line 277: What do you think about breaking this up into smaller tests, so they will fail/pass individually, rather than having one giant test?

dengliming: sure :)

jkwatson:
This is looking pretty good to me. Thanks for putting in the work!

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
```diff
@@ -147,14 +147,47 @@ public static Sampler alwaysOff() {
...
+   *     decisions if the remote parent exist and is not sampled.
+   * @param localParentSampled the optional {@code Sampler} which is used to make the sampling
+   *     decisions if the local parent exist and is sampled.
+   * @param localParentNotSampled the optional {@code Sampler} which is used to make the sampling
+   *     decisions if the local parent exist and is not sampled.
+   * @return a {@code Sampler} that follows the parent's sampling decision if one exists, otherwise
+   *     following the root sampler and other optional sampler's decision.
+   * @since 0.8.0
+   */
+  public static Sampler parentBased(
```
bogdandrutu, line 179: I would use a `Builder` pattern:
```
public static Builder newBuilder(Sampler root)
```

Where Builder allows to set any other `Sampler`. this way user don't have to configure all the others default samplers.

dengliming: 👍 what about `newParentBasedBuilder` ?

jkwatson: seems ok for a start, yes

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
```diff
@@ -147,14 +147,27 @@ public static Sampler alwaysOff() {
...
+  /**
+   * Returns a {@link ParentBased.Builder} that follows the parent's sampling decision if one
+   * exists, otherwise following the root sampler and other optional sampler's decision.
+   *
+   * @param root the required {@code Sampler} which is used to make the sampling decisions if the
+   *     parent does not exist.
+   * @return a {@code ParentBased.Builder}
+   * @since 0.8.0
+   */
+  public static ParentBased.Builder newParentBasedBuilder(Sampler root) {
```
anuraaga, line 169: Also to @jkwatson I think we should either have `new` on both or `new` on neither the non-builder and builder factories, they seem equally new. wdyt?

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
```diff
@@ -231,19 +259,109 @@ public SamplingResult shouldSample(
...
+      private Sampler localParentNotSampled;
+
+      /**
+       * Sets the remoteParentSampled. Optional.
+       *
+       * @param remoteParentSampled used to make the sampling decisions if the remote parent exist
+       *     and is sampled.
+       * @return this.
+       */
+      public Builder withRemoteParentSampled(Sampler remoteParentSampled) {
```
anuraaga, line 307: Our other builders like `SpanBuilder` and `Attributes.Builder` use `set` instead of `with`, can you use set here too?

anuraaga: Alternatively, `when` seems like it could fit as a special semantics for this class, in particular I think if the variable was named `Sampler whenRemoteParentSampled` it's really easy to read.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
```diff
@@ -231,19 +259,109 @@ public SamplingResult shouldSample(
...
+
+    static class Builder {
+      private final Sampler root;
+      private Sampler remoteParentSampled;
+      private Sampler remoteParentNotSampled;
+      private Sampler localParentSampled;
+      private Sampler localParentNotSampled;
+
+      /**
+       * Sets the remoteParentSampled. Optional.
```
anuraaga, line 301: Sets the {@link Sampler} to use when there is a remote parent that was sampled.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
```diff
@@ -231,19 +259,109 @@ public SamplingResult shouldSample(
...
+       *
+       * @param remoteParentSampled used to make the sampling decisions if the remote parent exist
```
anuraaga, line 303: I think we leave `@param` out when it basically matches the javadoc, e.g. https://github.com/open-telemetry/opentelemetry-java/blob/master/api/src/main/java/io/opentelemetry/common/Attributes.java#L204

anuraaga (commented):
Mostly looks good, thanks

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
dengliming: > named Sampler whenRemoteParentSampled it's really easy to read.
+1

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
dengliming: thanks. fixed at lastest commit.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
dengliming: thanks. fixed at lastest commit.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
```diff
@@ -231,19 +259,101 @@ public SamplingResult shouldSample(
...
+
+    static class Builder {
+      private final Sampler root;
+      private Sampler remoteParentSampled;
+      private Sampler remoteParentNotSampled;
+      private Sampler localParentSampled;
+      private Sampler localParentNotSampled;
+
+      /**
+       * Sets the {@link Sampler} to use when there is a remote parent that was sampled.
```
anuraaga, line 301: If not set, defaults to always sampling if the remote parent was sampled.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
```diff
@@ -231,19 +259,101 @@ public SamplingResult shouldSample(
...
+       *
+       * @return this Builder
+       */
+      public Builder withRemoteParentSampled(Sampler remoteParentSampled) {
+        this.remoteParentSampled = remoteParentSampled;
+        return this;
+      }
+
+      /**
+       * Sets the {@link Sampler} to use when there is a remote parent that was not sampled.
```
anuraaga, line 311: If not set, defaults to never sampling when the remote parent isn't sampled.

Similar for local

anuraaga (commented):
Sorry noticed one more small comment on the javadoc

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
dengliming: Usually, If we provide `builder ` factories, we should keep the `constructor ` private. right?

jkwatson: @anuraaga you were suggesting that the `parentBased(Sampler)` method should be `newParentBased(Sampler)`, if we want them to match, correct?

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
jkwatson: I think I'd go with `set` for consistency with the other APIs.

dengliming: thanks. Let's use `set`  instead of `with`.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
anuraaga: Ah yeah I was just talking about the factory names, not about changing constructor visibility. Indeed `newParentBased(Sampler)`. Alternatively, rename the builder returning one to `parentBasedBuilder`

dengliming: @anuraaga Gotcha 👍 thanks. what about this?
```
parentBased(Sampler root)
parentBasedBuilder(Sampler root)
```

anuraaga: I'm good with that :)

anuraaga (approved):
Thanks!

PR #1579, opened 2020-08-23
More aggresive version of BatchSpanProcessor

### LINKED ISSUE #729: Simplify the BatchSpansProcessor

### LINKED ISSUE #1560: Replace usages of java.util.Timer with a ScheduledExecutor in the BatchSpanProcessor

### LINKED ISSUE #1566: Leverage BatchSpanProcessor's span dropping functionality

iNikem:
Alternative to #1571

Fixes #729
Fixes #1566
Fixes #1560

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -70,6 +70,8 @@ configure(opentelemetryProjects) {
         it.options.errorprone.disable("UnnecessaryAnonymousClass")
         // "-Xep:UnnecessaryAnonymousClass:OFF"
 
+        it.options.errorprone.disable("UnusedNestedClass")
+
         it.options.encoding = "UTF-8"
 
         // Ignore warnings for protobuf and jmh generated files.
--- a/build.gradle
+++ b/build.gradle
@@ -70,8 +70,6 @@ configure(opentelemetryProjects) {
         it.options.errorprone.disable("UnnecessaryAnonymousClass")
         // "-Xep:UnnecessaryAnonymousClass:OFF"
 
-        it.options.errorprone.disable("UnusedNestedClass")
-
         it.options.encoding = "UTF-8"
 
         // Ignore warnings for protobuf and jmh generated files.
--- a/sdk/all/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorDroppedSpansBenchmark.java
+++ b/sdk/all/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorDroppedSpansBenchmark.java
@@ -43,8 +43,7 @@ import org.openjdk.jmh.annotations.TearDown;
 import org.openjdk.jmh.annotations.Threads;
 import org.openjdk.jmh.annotations.Warmup;
 
-// @AuxCounters(AuxCounters.Type.EVENTS)
-public class BatchSpanProcessorBenchmark {
+public class BatchSpanProcessorDroppedSpansBenchmark {
 
   private static class DelayingSpanExporter implements SpanExporter {
     @SuppressWarnings("FutureReturnValueIgnored")
@@ -61,8 +60,6 @@ public class BatchSpanProcessorBenchmark {
     @Override
     public void shutdown() {}
   }
-
-  //  private List<Span> spans;
 
   @State(Scope.Benchmark)
   public static class BenchmarkState {
@@ -88,7 +85,6 @@ public class BatchSpanProcessorBenchmark {
     @TearDown(Level.Iteration)
     public final void recordMetrics() {
       allMetrics = metricProducer.collectAllMetrics();
-      //      System.out.println(allMetrics);
     }
   }
 
@@ -110,6 +106,8 @@ public class BatchSpanProcessorBenchmark {
       if (total == 0) {
         return 0;
       } else {
+        // Due to peculiarities of JMH reporting we have to divide this by the number of the
+        // concurrent threads running the actual benchmark.
         return (double) dropped / total / 5;
       }
     }
--- a/sdk/all/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorDroppedSpansBenchmark.java
+++ b/sdk/all/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorDroppedSpansBenchmark.java
@@ -100,8 +100,8 @@ public class BatchSpanProcessorDroppedSpansBenchmark {
 
     /** Burn, checkstyle, burn. */
     public double dropRatio() {
-      long exported = getMetric("exportedSpans");
-      long dropped = getMetric("droppedSpans");
+      long exported = getMetric(true);
+      long dropped = getMetric(false);
       long total = exported + dropped;
       if (total == 0) {
         return 0;
@@ -113,22 +113,28 @@ public class BatchSpanProcessorDroppedSpansBenchmark {
     }
 
     public long exportedSpans() {
-      return getMetric("exportedSpans");
+      return getMetric(true);
     }
 
     public long droppedSpans() {
-      return getMetric("droppedSpans");
+      return getMetric(false);
     }
 
-    private long getMetric(String metricName) {
+    private long getMetric(boolean dropped) {
+      String labelValue = String.valueOf(dropped);
       for (MetricData metricData : allMetrics) {
-        if (metricData.getDescriptor().getName().equals(metricName)) {
+        if (metricData.getDescriptor().getName().equals("processedSpans")) {
           List<Point> points = new ArrayList<>(metricData.getPoints());
           if (points.isEmpty()) {
             return 0;
           } else {
-            LongPoint point = (LongPoint) points.get(points.size() - 1);
-            return point.getValue();
+            // Find latest point with given value of dropped label
+            for (int i = points.size() - 1; i >= 0; i--) {
+              LongPoint point = (LongPoint) points.get(i);
+              if (labelValue.equals(point.getLabels().get("dropped"))) {
+                return point.getValue();
+              }
+            }
           }
         }
       }
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
@@ -31,28 +31,23 @@ import io.opentelemetry.sdk.trace.ReadableSpan;
 import io.opentelemetry.sdk.trace.SpanProcessor;
 import io.opentelemetry.sdk.trace.data.SpanData;
 import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
 import java.util.Map;
-import java.util.Timer;
-import java.util.TimerTask;
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.Objects;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import javax.annotation.concurrent.GuardedBy;
+import javax.annotation.Nullable;
 
 /**
  * Implementation of the {@link SpanProcessor} that batches spans exported by the SDK then pushes
  * them to the exporter pipeline.
  *
  * <p>All spans reported by the SDK implementation are first added to a synchronized queue (with a
- * {@code maxQueueSize} maximum size, after the size is reached spans are dropped) and exported
- * every {@code scheduleDelayMillis} to the exporter pipeline in batches of {@code
- * maxExportBatchSize}. Spans may also be dropped when it becomes time to export again, and there is
- * an export in progress.
- *
- * <p>If the queue gets half full a preemptive notification is sent to the worker thread that
- * exports the spans to wake up and start a new export cycle.
+ * {@code maxQueueSize} maximum size, if queue is full spans are dropped). Spans are exported either
+ * when there are {@code maxExportBatchSize} pending spans or {@code scheduleDelayMillis} has passed
+ * since the last export finished.
  *
  * <p>This batch {@link SpanProcessor} can cause high contention in a very high traffic service.
  * TODO: Add a link to the SpanProcessor that uses Disruptor as alternative with low contention.
@@ -85,10 +80,7 @@ public final class BatchSpanProcessor implements SpanProcessor {
 
   private static final String WORKER_THREAD_NAME =
       BatchSpanProcessor.class.getSimpleName() + "_WorkerThread";
-  private static final String TIMER_THREAD_NAME =
-      BatchSpanProcessor.class.getSimpleName() + "_TimerThread";
   private final Worker worker;
-  private final Thread workerThread;
   private final boolean sampled;
 
   private BatchSpanProcessor(
@@ -102,11 +94,11 @@ public final class BatchSpanProcessor implements SpanProcessor {
         new Worker(
             spanExporter,
             scheduleDelayMillis,
-            maxQueueSize,
             maxExportBatchSize,
-            exporterTimeoutMillis);
-    this.workerThread = new DaemonThreadFactory(WORKER_THREAD_NAME).newThread(worker);
-    this.workerThread.start();
+            exporterTimeoutMillis,
+            new ArrayBlockingQueue<ReadableSpan>(maxQueueSize));
+    Thread workerThread = new DaemonThreadFactory(WORKER_THREAD_NAME).newThread(worker);
+    workerThread.start();
     this.sampled = sampled;
   }
 
@@ -133,7 +125,6 @@ public final class BatchSpanProcessor implements SpanProcessor {
 
   @Override
   public void shutdown() {
-    workerThread.interrupt();
     worker.shutdown();
   }
 
@@ -144,9 +135,6 @@ public final class BatchSpanProcessor implements SpanProcessor {
 
   // Worker is a thread that batches multiple spans and calls the registered SpanExporter to export
   // the data.
-  //
-  // The list of batched data is protected by an explicit monitor object which ensures full
-  // concurrency.
   private static final class Worker implements Runnable {
 
     static {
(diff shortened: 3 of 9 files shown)


sdk/common/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
```diff
@@ -107,4 +109,33 @@ public CompletableResultCode whenComplete(Runnable action) {
...
+          @Override
+          public void run() {
+            latch.countDown();
+          }
+        });
+    latch.await();
+  }
+
+  /** Calls {@link #fail()} if times out. */
+  public void join(long timeout, TimeUnit unit) throws InterruptedException {
```
anuraaga, line 127: `CompletableFuture.join` doesn't throw interrupted exception so I think it's ok to squash that here too (don't know if it's best practice for interruptions but may as well copy the JDK)

sdk/common/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
```diff
@@ -107,4 +109,33 @@ public CompletableResultCode whenComplete(Runnable action) {
     }
     return this;
   }
+
+  /** Waits until this instances completes, regardless of the result. */
+  public void join() throws InterruptedException {
```
anuraaga, line 114: I think we can leave out this method, joining without timeout is rarely a good idea.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -164,146 +152,121 @@ public void forceFlush() {
...
-            } while (spansList.isEmpty());
+      updateNextExportTime();
+
+      while (continueWork) {
+        try {
+          if (flushRequested != null) {
+            flush();
+          }
+
+          ReadableSpan lastElement = queue.poll(100, TimeUnit.MILLISECONDS);
```
anuraaga, line 199: Is there a reason to wait for a constant here instead of until the next export time?

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -164,146 +152,121 @@ public void forceFlush() {
...
+          if (lastElement != null) {
+            batch.add(lastElement.toSpanData());
+          }
```
anuraaga, line 202: It's probably a nice win to add a `drainTo` here to get as many elements as possible. In the case where there are constantly spans, we'll be able to handle them with less queue lock acquisitions and tighter loop. `drainTo` has this very tight loop inside the lock which JVM should deal well with.

```java
while (i < n) {
                    E e = (E) items[take];
                    c.add(e);
                    items[take] = null;
                    if (++take == items.length) take = 0;
                    i++;
                }
```

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -164,146 +152,121 @@ public void forceFlush() {
...
+          Thread.currentThread().interrupt();
+          return;
+        }
+      }
+    }
+
+    private void flush() {
+      int spansToFlush = queue.size();
+      while (spansToFlush > 0) {
+        ReadableSpan span = queue.poll();
```
anuraaga, line 219: Similar to above, it might be better if we can structure this to take advantage of drainTo. I guess you can consecutively drainTo and subtract the number drained from spansToFlush instead of individual poll without a huge change in structure.

anuraaga:
By the way I added this benchmark https://github.com/open-telemetry/opentelemetry-java/pull/1571/files#diff-85f8a2fe8ea17909437002d022e86fab - no clue if it's a very good benchmark but maybe it can be useful :)

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -164,146 +152,121 @@ public void forceFlush() {
...
-                public void run() {
-                  result.fail();
-                }
-              },
-              exporterTimeoutMillis);
-        } catch (Exception e) {
-          logger.log(Level.WARNING, "Exporter threw an Exception", e);
+      try {
+        final CompletableResultCode result = spanExporter.export(batch);
+        result.join(exporterTimeoutMillis, TimeUnit.MILLISECONDS);
```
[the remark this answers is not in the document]
iNikem: I am not sure about that. We have to prevent concurrent calls to export and we have to have timeouts. IMO timed joins is the easiest way to achieve this. At least until benchmarking shows that this is too expensive.

sdk/common/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
iNikem: You mean catch that inside the method?

sdk/common/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
iNikem: I use it `forceFlush`. There is no good option how long should I wait there. And if you say that `forceFlush` should be async, I will just change my argument to `shutdown` :)

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
iNikem: Hm... I will think about that. My first concern is the introduction of `System.nanoTime` on essentially every `poll`.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
iNikem: I will try to benchmark this class today

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
iNikem: Poll certainly cannot block for `scheduleDelayMillis` if it is already almost time to export, but we have partial batch.

jkwatson:
> By the way I added this benchmark https://github.com/open-telemetry/opentelemetry-java/pull/1571/files#diff-85f8a2fe8ea17909437002d022e86fab - no clue if it's a very good benchmark but maybe it can be useful :)

What do you think about putting in that benchmark as a stand-alone PR, so we can test various solutions on the same footing?

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
iNikem: A, main reason I cannot use `drain` here is that I immediately convert to `SpanData`. And I don't want to drain to one collection, then iterate and convert element by element into another collection.

iNikem:
Pushed some polish and my take on @anuraaga 's benchmark. I was more interested in the dropped spans count. Unfortunately could not figure out how to properly measure that, but current benchmark gives some indication. We would greatly benefit from utility to emit new spans with given rate...

Anyway, how should we proceed and choose between different proposal for BSP? Let me remind you, that java instrumentation fails tests with current otel-java snapshot, because of BSP.

build.gradle
```diff
@@ -70,6 +70,8 @@ configure(opentelemetryProjects) {
         it.options.errorprone.disable("UnnecessaryAnonymousClass")
         // "-Xep:UnnecessaryAnonymousClass:OFF"
 
+        it.options.errorprone.disable("UnusedNestedClass")
```
jkwatson, line 73: what is the use-case for an unused nested class that we wouldn't want to warn about it?

iNikem: Damn... My benchmark class has a `ThreadState` inner class that error prone disliked. I will take a look if I can remove this.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
iNikem: See https://github.com/open-telemetry/opentelemetry-java/pull/1579#discussion_r475713805

build.gradle
iNikem: Removed this line

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -160,150 +147,135 @@ public void forceFlush() {
...
+          }
+
+          ReadableSpan lastElement = queue.poll(100, TimeUnit.MILLISECONDS);
+          if (lastElement != null) {
+            batch.add(lastElement.toSpanData());
           }
-          spansCopy = new ArrayList<>(spansList);
-          spansList.clear();
+
+          if (batch.size() >= maxExportBatchSize || System.nanoTime() >= nextExportTime) {
```
jkwatson, line 213: this is very similar to what I did when I took a swing at this class 6 months ago, so obviously I think it's a great approach. ;)

iNikem:
After discussion with @anuraaga we have decided to move forward with this PR. In addition I will file a new issue for making `SpanProcessor.forceFlush` and `SpanProcessor.shutdown` methods non-blocking.

iNikem:
Just in case I remind that this and #1585 are required to let instrumentation upgrade to the latest otel-java version. And we need that to move forward

sdk/common/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
```diff
@@ -107,4 +109,33 @@ public CompletableResultCode whenComplete(Runnable action) {
...
+  /** Returns whether this {@link CompletableResultCode} has completed. */
+  public boolean isDone() {
+    synchronized (lock) {
+      return succeeded != null;
+    }
+  }
+
+  /** Calls {@link #fail()} if times out. */
+  public CompletableResultCode join(long timeout, TimeUnit unit) {
+    final CountDownLatch latch = new CountDownLatch(1);
```
anuraaga, line 122: If you don't mind copying in #1583 that'd be cool, I think the synchronous case optimization and less branching (same lines of code to just all `fail()` twice) is worth it. Then no worries closing that PR in favor of this one

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -160,150 +153,135 @@ public void forceFlush() {
...
-                // Preserve the interruption status as per guidance and stop doing any work.
-                Thread.currentThread().interrupt();
-                return;
-              }
-            } while (spansList.isEmpty());
+      updateNextExportTime();
+
+      while (continueWork) {
+        try {
+          if (flushRequested.get() != null) {
```
anuraaga, line 210: AINTD, we still have the issue that a flush can be delayed by a long 100ms. Do you mind filing an issue for it? Flush is mostly used on FaaS so they pay $$ for the time they use, we should put in the effort to save that even if it adds some complexity.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -160,150 +153,135 @@ public void forceFlush() {
...
+            flush();
+          }
+
+          ReadableSpan lastElement = queue.poll(100, TimeUnit.MILLISECONDS);
```
anuraaga, line 214: I think it's only this line that can produce a `InterruptedException`, can you narrow down the scope of the try / catch to make that more obvious?

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -160,150 +153,135 @@ public void forceFlush() {
...
     }
 
-    private void shutdown() {
-      forceFlush();
-      timer.cancel();
-      spanExporter.shutdown();
+    private void flush() {
+      int spansToFlush = queue.size();
+      while (spansToFlush > 0) {
+        ReadableSpan span = queue.poll();
```
anuraaga, line 234: As we discussed in chat, I still have reservations about the lock contention between `.poll` and `.offer`. It'd be nice if we can file an issue about performance benchmarking / optimization of BSP. If I understood correctly, we're ok with adding complexity if we start with numbers and show the numbers with it, which makes sense to me.

anuraaga (approved):
I'm good with this approach. I consider it the simplest, but it's likely not the most optimal. We should iterate on it with some clear numeric targets, and I think it's reasonable for us to have extension artifacts such as an `opentelemetry-exporter-rxjava` that uses a nice async library for handling it even better. Of course, such extensions are blocked on #1584 which is why that's an important one which I'll rebase #1571 to address.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
iNikem: I will do that after this PR is merged. It will be easier to refer to the code concerned.

sdk/common/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
iNikem: Will do.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
iNikem: #1587 1587

sdk/all/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorDroppedSpansBenchmark.java
```diff
@@ -0,0 +1,151 @@
...
+  @AuxCounters(AuxCounters.Type.EVENTS)
+  public static class ThreadState {
+    private Collection<MetricData> allMetrics;
+
+    @TearDown(Level.Iteration)
+    public final void recordMetrics(BenchmarkState benchmarkState) {
+      allMetrics = benchmarkState.allMetrics;
+    }
+
+    /** Burn, checkstyle, burn. */
```
jkwatson, line 101: 😂

sdk/common/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
```diff
@@ -107,4 +109,42 @@ public CompletableResultCode whenComplete(Runnable action) {
...
+   * it times out or is interrupted, the {@link CompletableResultCode} is failed.
+   *
+   * @return this {@link CompletableResultCode}
+   */
+  public CompletableResultCode join(long timeout, TimeUnit unit) {
+    synchronized (lock) {
+      if (succeeded != null) {
+        return this;
+      }
+    }
```
jkwatson, line 131: same comment as on @anuraaga 's PR: can we just do `if (isDone()) return this;`  instead of duplicating the code?

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -160,150 +153,134 @@ public void forceFlush() {
       droppedSpans =
           droppedSpansCounter.bind(
               Labels.of("spanProcessorType", BatchSpanProcessor.class.getSimpleName()));
+      LongCounter exportedSpansCounter =
+          meter
+              .longCounterBuilder("exportedSpans")
```
jkwatson, line 158: what would you think about having only one counter called "processedSpans", with a boolean label for dropped or not?

jkwatson:
Generally looking good. Just 2 small questions/changes requested.

sdk/common/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
iNikem: Rebased and took all changes from Anuraag PR

anuraaga (approved):
🙇

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -142,169 +132,148 @@ public void forceFlush() {
...
     static {
       Meter meter = OpenTelemetry.getMeter("io.opentelemetry.sdk.trace");
-      LongCounter droppedSpansCounter =
+      LongCounter processedSpansCounter =
           meter
-              .longCounterBuilder("droppedSpans")
+              .longCounterBuilder("processedSpans")
               .setUnit("1")
               .setDescription(
                   "The number of spans dropped by the BatchSpanProcessor due to high throughput.")
```
jkwatson, line 152: this description isn't quite right any more. Can be fixed in a follow-on PR, to get you unblocked.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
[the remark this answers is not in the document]
huntc: Join shouldn’t have a timeout. There’s nothing you can really do if you can’t join. This is why CompleteableFuture and even back to pthreads has no join.

If an exporter doesn’t return then there is nothing that can be done here other than fail it’s result. Exporters can then act on a result failure and halt their activity.

huntc (commented):
It’s really disappointing that we finally reinstated blocking. It is unnecessary to block.

iNikem:
@huntc Are you talking about blocking when calling `exporter.export` method? Can you explain to me why is this bad? I mean that method is called from the dedicated worker thread and we have to ensure that `export` method is not called concurrently. I don't see why blocking is bad in this situation?

huntc:
> @huntc Are you talking about blocking when calling `exporter.export` method? Can you explain to me why is this bad? I mean that method is called from the dedicated worker thread and we have to ensure that `export` method is not called concurrently. I don't see why blocking is bad in this situation?

I was hoping for an executor that can be shared across many components. This is no longer possible without hogging a thread from an executor used in other places. Reusing executors saves on resources.

iNikem:
You mean not using a dedicated BSP-specific worker thread, but a shared executor for all internal concurrent needs? This is a good idea :) And you are absolutely correct, if this idea is ever to be implemented, we will need to change this implementation to non-blocking one.

PR #1583, opened 2020-08-25
Add helpers to CompletableResultCode to wait for a result.

anuraaga:
In #1579 #1571 etc, we found that we do need to allow users to be able to wait for a result, perhaps in a FaaS scenario or just in a unit test. Extracted this out since I ended up needing it in another PR. Added a test for interruption since that PR too. @iNikem

### DIFF
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
@@ -18,6 +18,8 @@ package io.opentelemetry.sdk.common.export;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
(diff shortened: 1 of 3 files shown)


iNikem (approved):

sdk/common/src/main/java/io/opentelemetry/sdk/common/export/CompletableResultCode.java
```diff
@@ -107,4 +109,42 @@ public CompletableResultCode whenComplete(Runnable action) {
...
+   * it times out or is interrupted, the {@link CompletableResultCode} is failed.
+   *
+   * @return this {@link CompletableResultCode}
+   */
+  public CompletableResultCode join(long timeout, TimeUnit unit) {
+    synchronized (lock) {
+      if (succeeded != null) {
+        return this;
+      }
+    }
```
jkwatson, line 131: could we replace this block with `if (isDone()) return this;` ?

anuraaga: Nice, thanks

PR #1585, opened 2020-08-25
Timeout fix in OTLP exporters

iNikem:
`withDeadlineAfter` sets deadline _from the time of its call_. Which means that is has to be called write before the actual call to `export`.


anuraaga (approved):
Doh my bad - possible to add a test?

iNikem:
Added test

iNikem:
Just in case I remind that this and #1579 are required to let instrumentation upgrade to the latest otel-java version. And we need that to move forward

Issue #1580, opened 2020-08-23
Proposal: Remove AttributeValue class and implement attribute value type-safety with a type-indicating Key.


jkwatson:
**Is your feature request related to a problem? Please describe.**

1. Users have expressed some confusion about the way the Semantic conventions for Attribute keys/values are implemented. The usage of the AttributeSetter classes is awkward, and when you aren't using a pre-defined instance of the SemanticAttributes class, it's awkward for API users to have to define their own AttributeSetters to make their API usage consistent.
2. The use of AttributeValue requires, in almost every case, an additional instance allocation to wrap the value that is being set in the Attributes. Most commonly, in instrumentation, the attribute Keys are the thing that is known ahead of time and can be statically allocated, requiring no additional runtime allocation overhead, after the instrumentation (or the API itself, in the case of pre-defined semantic attributers) is initialized. Our current implementation requires the wrapping of every value, at runtime, in the hot path of the application of the instrumentation.

**Describe the solution you'd like**
Rather than provide the type-safe wrapper the value side of the attribute, I propose to make the type safety be enforced by providing type-information on the _key_ side. The API would look something like (for the `Span` interface):

```java
 void setAttribute(StringValuedKey key, String value);
 void setAttribute(LongValuedKey key, long value);
 void setAttribute(DoubleValuedKey key, double value);
 void setAttribute(BooleanValuedKey key, boolean value);
...
```

As mentioned above, this would have the significant allocation advantage of not needing to do any extra allocations at runtime for the values of the attributes, at least for the String valued attributes (which make up the majority of our semantic attributes). The implementation would probably need to wrap the primitives in boxed versions in the storage, so there would be a net zero gain in runtime allocations for the 3 primitive-typed values [again, assuming the keys are statically pre-defined, either in the instrumentation, or in the API's semantic attribute definitions].

A version of this solution was prototyped in this PR: https://github.com/open-telemetry/opentelemetry-java/pull/1097

**Describe alternatives you've considered**
Several other alternatives have been prototyped and proposed:
* Removing AttributeValue altogether, and not replacing it with anything: https://github.com/open-telemetry/opentelemetry-java/pull/1500
* Much discussion has taken place on this issue: https://github.com/open-telemetry/opentelemetry-java/issues/1081
* And some on this one as well: https://github.com/open-telemetry/opentelemetry-java/issues/1314

**Additional context**
[...]

PR #1589, opened 2020-08-25
Convert Attributes and AttributeValue into interfaces

jkwatson:
The creational methods are now in inner classes, which is definitely an API breaking change.

Note: had to pull out the other inner classes up a level, otherwise they would inherit the interfaces `public` access modifier.

The rationale for this is to make life easier for the instrumentation agent code, where it is more desirable to deal with interfaces than abstract classes.

Relevant Issues:
* https://github.com/open-telemetry/opentelemetry-java/issues/1580
* https://github.com/open-telemetry/opentelemetry-java/issues/1314
* https://github.com/open-telemetry/opentelemetry-java/issues/1191
* https://github.com/open-telemetry/opentelemetry-java/issues/1081

Other PRs that strive to achieve the same goal in different ways:
* https://github.com/open-telemetry/opentelemetry-java/pull/1500
* https://github.com/open-telemetry/opentelemetry-java/pull/1097

### DIFF
--- a/api/src/main/java/io/opentelemetry/common/ArrayBackedAttributes.java
+++ b/api/src/main/java/io/opentelemetry/common/ArrayBackedAttributes.java
@@ -14,16 +14,28 @@
  * limitations under the License.
  */
 
-package io.opentelemetry.sdk.resources;
+package io.opentelemetry.common;
 
-import io.opentelemetry.common.AttributeValue;
-import io.opentelemetry.common.Attributes;
+import com.google.auto.value.AutoValue;
+import java.util.ArrayList;
+import java.util.List;
+import javax.annotation.concurrent.Immutable;
 
-public class TestResourceProvider extends ResourceProvider {
+@AutoValue
+@Immutable
+abstract class ArrayBackedAttributes extends ImmutableKeyValuePairs<AttributeValue>
+    implements Attributes {
+  ArrayBackedAttributes() {}
+
+  static Attributes create(List<Object> data) {
+    return new AutoValue_ArrayBackedAttributes(data);
+  }
 
   @Override
-  protected Attributes getAttributes() {
-    return Attributes.Factory.of(
-        "providerAttribute", AttributeValue.Factory.longAttributeValue(42));
+  abstract List<Object> data();
+
+  @Override
+  public Attributes.Builder toBuilder() {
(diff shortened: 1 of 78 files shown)


Oberon00:
Can you link the corresponding issue or otherwise state a rationale?

jkwatson:
> Can you link the corresponding issue or otherwise state a rationale?

Sorry about that. I probably should have marked this as a draft. This is another attempt, in a long series, at providing a way for the java agent to not have to do a bunch of bending-over-backwards to do extra wrapping of concrete classes that exist in the API. I'll update the description with pointers to the many discussions.

trask:
Ugh. I tried implementing both the unshaded and shaded AttributeValue interfaces at the same time, but `getType()` method can't return two different enums. If Type was interface we could use covariant return to return something that implements both interfaces. But that would be a poor API choice.

bogdandrutu:
@trask I think that is a downgrade in usability because cannot use switch statement.

jkwatson:
> Ugh. I tried implementing both the unshaded and shaded AttributeValue interfaces at the same time, but `getType()` method can't return two different enums. If Type was interface we could use covariant return to return something that implements both interfaces. But that would be a poor API choice.

oh, bummer. Pretty stuck if we need to keep the Type around. If we got rid of AttributeValue altogether, and switched out for the typed keys, it might work, though. I don't remember if we still need the Type enum in that case. I'll take a look tomorrow.

bogdandrutu:
@jkwatson want to make sure we don't do all the decisions based on the ability to shade or not. It is very important the agent use-case but API should be developed for usability as well. That being said I am not oppose to search for new ways that are user friendly as well as agent friendly

jkwatson:
> @jkwatson want to make sure we don't do all the decisions based on the ability to shade or not. It is very important the agent use-case but API should be developed for usability as well. That being said I am not oppose to search for new ways that are user friendly as well as agent friendly

Absolutely. I always favor the user in these decisions (although our primary user is the instrumentation folks at the moment!). Trying to find the balance between developer ergonomics and performance in the agent is the name of the game.

jkwatson:
Closing this as a failed experiment, since it makes a less friendly API (with the inner Factory classes) and doesn't actually make it easier for shading in the agent.

PR #1590, opened 2020-08-25
1556 change resourceconstants to resourceattributes

### LINKED ISSUE #1556: ResourceConstants -> ResourceAttributes
We should use the same `*AttributeSetter` pattern for Resource as we do for span attributes for consistency. Since it's public API we need to knock it out soonish.

jarebudev:
resolves #1556


anuraaga (approved):
Thanks!

PR #1594, opened 2020-08-26
SpanContext hides the TraceId/SpanId implementations

jkwatson:
This removes the usage of wrapper classes for the ids. 
Changes the SpanContext creation methods to accept `CharSequence`s and an offset for the ids.
Changes the IdGenerator interface to generate String ids.

Relevant issue: #1314
Extensive discussion about this took place in this draft PR: #1374


jkwatson:
I apologize for the size of this PR. Most of it is mechanical changes in types, so hopefully review will go quickly.

api/src/main/java/io/opentelemetry/trace/BigendianEncoding.java
```diff
@@ -165,5 +174,24 @@ private static void byteToBase16(byte value, char[] dest, int destOffset) {
     dest[destOffset + 1] = ENCODING[b | 0x100];
   }
 
+  public static boolean isValidBase16String(CharSequence value) {
+    for (int i = 0; i < value.length(); i++) {
+      char b = value.charAt(i);
+      // 48..57 && 97..102 are valid
+      if (!Character.isDigit(b) && (b < 97 || b > 102)) {
```
anuraaga, line 181: It might affect other places as well, but while we're here, let's use a non-unicode check for `isDigit`

api/src/main/java/io/opentelemetry/trace/BigendianEncoding.java
```diff
@@ -138,6 +138,15 @@ static void byteToBase16String(byte value, char[] dest, int destOffset) {
     byteToBase16(value, dest, destOffset);
   }
 
+  static byte[] bytesFromBase16(CharSequence value, int srcOffset, int charactersToRead) {
```
anuraaga, line 141:
```suggestion
  static byte[] bytesFromBase16(CharSequence value, int offset, int length) {
```

I think this pattern is very prevelant in the JDK so good to follow the idiom

api/src/main/java/io/opentelemetry/trace/BigendianEncoding.java
```diff
@@ -138,6 +138,15 @@ static void byteToBase16String(byte value, char[] dest, int destOffset) {
...
+    CharSequence part = value.subSequence(srcOffset, srcOffset + charactersToRead);
```
anuraaga, line 142: Think the loop is still very easy even without the subsequence

api/src/main/java/io/opentelemetry/trace/SpanContext.java
```diff
@@ -75,25 +76,88 @@ public static SpanContext create(
...
+      int spanIdOffset,
+      TraceFlags traceFlags,
+      TraceState traceState,
+      boolean remote) {
+    return new AutoValue_SpanContext(
+        traceId.subSequence(traceIdOffset, traceIdOffset + 32).toString(),
+        spanId.subSequence(spanIdOffset, spanIdOffset + 16).toString(),
+        traceFlags,
+        traceState,
+        /* remote$=*/ remote);
```
anuraaga, line 113:
```suggestion
        /* remote=*/ remote);
```

Isn't it?

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/Common.java
```diff
@@ -52,8 +52,8 @@ static SpanContext buildSpanContext(String traceId, String spanId, String sample
               : NOT_SAMPLED_FLAGS;
 
       return SpanContext.createFromRemoteParent(
-          TraceId.fromLowerBase16(StringUtils.padLeft(traceId, MAX_TRACE_ID_LENGTH), 0),
-          SpanId.fromLowerBase16(spanId, 0),
+          StringUtils.padLeft(traceId, MAX_TRACE_ID_LENGTH),
+          spanId,
```
anuraaga, line 56: We can replace some of the methods in this class for checking validity I think

api/src/main/java/io/opentelemetry/trace/SpanContext.java
```diff
@@ -75,25 +76,88 @@ public static SpanContext create(
...
+  abstract String getSpanId();
+
   /**
    * Returns the trace identifier associated with this {@code SpanContext}.
    *
    * @return the trace identifier associated with this {@code SpanContext}.
    * @since 0.1.0
    */
-  public abstract TraceId getTraceId();
+  public String getTraceIdAsBase16() {
```
anuraaga, line 126:
```suggestion
  public String getTraceIdAsHexString() {
```

Our spec now uses hex too and I think it's more familiar to people than base 16
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#retrieving-the-traceid-and-spanid

api/src/main/java/io/opentelemetry/trace/SpanContext.java
```diff
@@ -75,25 +76,88 @@ public static SpanContext create(
...
+   * @return a new {@code SpanContext} with the given identifiers and options.
+   * @since 0.1.0
+   */
+  @SuppressWarnings({
+    "InconsistentOverloads",
+    "UngroupedOverloads",
+    "OverloadMethodsDeclarationOrder"
+  })
+  public static SpanContext create(
+      CharSequence traceId,
```
anuraaga, line 101: Since our accessors (e.g. `getTraceId()` use `String`, I don't think there's an advantage to using `CharSequence` in the factories. Can we use `String` everywhere?

api/src/main/java/io/opentelemetry/trace/SpanId.java
```diff
@@ -16,43 +16,25 @@
...
 
+  private static final ThreadLocal<char[]> charBuffer = new ThreadLocal<>();
   private static final int SIZE = 8;
   private static final int BASE16_SIZE = 2 * SIZE;
-  private static final long INVALID_ID = 0;
-  private static final SpanId INVALID = new SpanId(INVALID_ID);
 
-  // The internal representation of the SpanId.
-  private final long id;
+  public static final String INVALID = "0000000000000000";
```
anuraaga, line 35:
```suggestion
  private static final String INVALID = "0000000000000000";
```

api/src/main/java/io/opentelemetry/trace/TraceId.java
```diff
@@ -17,55 +17,42 @@
...
-public final class TraceId implements Comparable<TraceId> {
+public final class TraceId {
+  private static final ThreadLocal<char[]> charBuffer = new ThreadLocal<>();
 
-  private static final int SIZE = 16;
+  private static final int SIZE_IN_BYTES = 16;
   private static final int BASE16_SIZE = 2 * BigendianEncoding.LONG_BASE16;
-  private static final long INVALID_ID = 0;
-  private static final TraceId INVALID = new TraceId(INVALID_ID, INVALID_ID);
+  public static final String INVALID = "00000000000000000000000000000000";
```
anuraaga, line 35:
```suggestion
  private static final String INVALID = "00000000000000000000000000000000";
```

api/src/main/java/io/opentelemetry/trace/SpanContext.java
```diff
@@ -75,25 +76,88 @@ public static SpanContext create(
...
+      int traceIdOffset,
+      CharSequence spanId,
+      int spanIdOffset,
+      TraceFlags traceFlags,
+      TraceState traceState,
+      boolean remote) {
+    return new AutoValue_SpanContext(
+        traceId.subSequence(traceIdOffset, traceIdOffset + 32).toString(),
```
anuraaga, line 109: I don't know if the offset versions provide enough user benefit to have (not much harder to call subsequence than grok this long method signature). There doesn't seem to be a performance benefit since we always need to take the substring anyways.

If we only had two factories, than I think we could also provide two symmetric factories that accept `byte[]` and we don't need public APIs like `TraceId.toLowerBase16(byte[])`

anuraaga: Though realized a byte[] version of this factory would have benefit from an offset version :)

api/src/main/java/io/opentelemetry/trace/SpanId.java
```diff
@@ -65,46 +47,38 @@ public static int getSize() {
...
-   * @param destOffset the starting offset in the destination buffer.
-   * @throws NullPointerException if {@code dest} is null.
-   * @throws IndexOutOfBoundsException if {@code destOffset+SpanId.getSize()} is greater than {@code
-   *     dest.length}.
-   * @since 0.1.0
-   */
-  public void copyBytesTo(byte[] dest, int destOffset) {
-    BigendianEncoding.longToByteArray(id, dest, destOffset);
+  /** Generate a valid {@link SpanId} from the given long value. */
+  public static String fromLong(long id) {
```
anuraaga, line 69: Any thoughts on moving these to reflect they're only used for ID generation, maybe `IdGeneratorUtil`, or we could make `IdGenerator` an abstract class with protected helper methods for generating IDs from longs?

api/src/main/java/io/opentelemetry/trace/TraceId.java
```diff
@@ -193,32 +163,7 @@ public boolean equals(@Nullable Object obj) {
...
-  @Override
-  public int compareTo(TraceId that) {
-    if (idHi == that.idHi) {
-      if (idLo == that.idLo) {
-        return 0;
-      }
-      return idLo < that.idLo ? -1 : 1;
-    }
-    return idHi < that.idHi ? -1 : 1;
+  public static long getTraceIdRandomPart(CharSequence traceId) {
```
anuraaga, line 166:
```suggestion
  public static long getRandomPart(CharSequence traceId) {
```

We don't repeat the classname in the others, don't think we need to here either.

anuraaga: Though this seems like it doesn't need to be public API, it even documents itself as being used in one specific sampler?

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/B3PropagatorInjectorSingleHeader.java
```diff
@@ -50,9 +50,13 @@
     SpanContext spanContext = span.getContext();
 
     char[] chars = new char[COMBINED_HEADER_SIZE];
-    spanContext.getTraceId().copyLowerBase16To(chars, 0);
+    String traceId = spanContext.getTraceIdAsBase16();
+    System.arraycopy(traceId.toCharArray(), 0, chars, 0, TRACE_ID_HEX_SIZE);
```
anuraaga, line 54:
```suggestion
    traceId.getChars(0, traceId.length(), chars, 0);
```

Similar throughout

api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java
```diff
@@ -94,9 +94,19 @@
     chars[0] = VERSION.charAt(0);
     chars[1] = VERSION.charAt(1);
     chars[2] = TRACEPARENT_DELIMITER;
-    spanContext.getTraceId().copyLowerBase16To(chars, TRACE_ID_OFFSET);
+
+    CharSequence traceId = spanContext.getTraceIdAsBase16();
```
anuraaga, line 98: Looks like a lot can be updated to use `String` / getChars instead of `CharSequence`

anuraaga (commented):
Thanks!

api/src/main/java/io/opentelemetry/trace/SpanContext.java
jkwatson: you would think that, but it seems to shift back and forth in ways I can't predict. If you get it wrong, checkstyle/errorprone complains.

api/src/main/java/io/opentelemetry/trace/TraceId.java
jkwatson: This was the method name before, so I kept it as-is.  And, it's public so any Sampler could get access to it, I believe.

api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java
jkwatson: yeah, it's true. I hadn't tracked them all down. I don't think it's actually harmful to declare these as CharSequence, so I wasn't aggressive at finding them all.

api/src/main/java/io/opentelemetry/trace/BigendianEncoding.java
jkwatson: definitely. I thought I had caught all these, but I missed this one. will fix.

api/src/main/java/io/opentelemetry/trace/BigendianEncoding.java
jkwatson: happy to fix it. Are there non-ascii digits out there?

api/src/main/java/io/opentelemetry/trace/SpanContext.java
jkwatson: My thinking (and it wasn't very strong thinking) was that to be more accepting of inputs, and more restrictive on outputs. But, I definitely don't feel too strongly about that.

api/src/main/java/io/opentelemetry/trace/SpanContext.java
jkwatson: I'll go through and change it.

api/src/main/java/io/opentelemetry/trace/SpanId.java
jkwatson: I liked having all the transformations "close to home" in the relevant helpers, but it wouldn't be hard to convince me to move them. Let's do that as a follow-up, if that's ok.

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/Common.java
jkwatson: You mean use the ones on TraceId/SpanId?

api/src/main/java/io/opentelemetry/trace/SpanContext.java
jkwatson: This was a method signature that @bogdandrutu had requested. I don't know that it's necessary at the moment, though.

api/src/main/java/io/opentelemetry/trace/SpanId.java
jkwatson: Also, there are quite a few usages in test code at the moment, so I think I'd want to hold off for a follow-up PR, for sure.

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/Common.java
anuraaga: Yeah

jkwatson: done

api/src/main/java/io/opentelemetry/trace/BigendianEncoding.java
jkwatson: done

api/src/main/java/io/opentelemetry/trace/BigendianEncoding.java
jkwatson: done

api/src/main/java/io/opentelemetry/trace/SpanContext.java
anuraaga: Got it - I'm ok with either, but have a small feeling that `String` feels more familiar to developers

api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java
jkwatson: I'm going to go through and get rid of the CharSequence usage and go to strictly String based API. We can always loosen it up later, if we just support String for now.

api/src/main/java/io/opentelemetry/trace/SpanContext.java
jkwatson: I think we should get rid of the offsets for now; we can always add another option later; we can't take it back if we leave it in.

api/src/main/java/io/opentelemetry/trace/SpanContext.java
jkwatson: I think occurs when you use the `@Memoized` annotation, due to the way that generates the underlying implementations.

api/src/main/java/io/opentelemetry/trace/SpanContext.java
jkwatson: I got rid of them. Smaller API is better for now.

api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java
jkwatson: done

jkwatson:
@anuraaga I think I got all of your comments addressed. Thanks so much for taking the time to give it a thorough review!

api/src/main/java/io/opentelemetry/trace/BigendianEncoding.java
```diff
@@ -165,5 +173,24 @@ private static void byteToBase16(byte value, char[] dest, int destOffset) {
     dest[destOffset + 1] = ENCODING[b | 0x100];
   }
 
+  public static boolean isValidBase16String(CharSequence value) {
+    for (int i = 0; i < value.length(); i++) {
+      char b = value.charAt(i);
+      // 48..57 && 97..102 are valid
+      if ((b < 48 || b > 57) && (b < 97 || b > 102)) {
```
anuraaga, line 180: I'm pretty sure they're the same, and maybe it's just me but I find writing the same code as the comment would be more readable 

```suggestion
      if (!(b >= 48 && b <= 57) || !(b >= 97 && b <= 102)) {
```

api/src/test/java/io/opentelemetry/trace/TraceIdTest.java
```diff
@@ -75,56 +76,42 @@ void testGetRandomTracePart_NegativeLongRepresentation() {
...
+    assertThat(TraceId.getTraceIdRandomPart(traceid)).isEqualTo(0xFF0A0B0C0D0E0F00L);
   }
 
   @Test
   void fromLowerBase16() {
-    assertThat(TraceId.fromLowerBase16("00000000000000000000000000000000", 0))
-        .isEqualTo(TraceId.getInvalid());
-    assertThat(TraceId.fromLowerBase16("00000000000000000000000000000061", 0)).isEqualTo(first);
-    assertThat(TraceId.fromLowerBase16("ff000000000000000000000000000041", 0)).isEqualTo(second);
+    assertThat(
```
anuraaga, line 85: Think we can remove `.toString()` in this file

exporters/logging/src/test/java/io/opentelemetry/exporters/logging/LoggingSpanExporterTest.java
```diff
@@ -60,8 +60,8 @@ void returnCode() {
     SpanData spanData =
         TestSpanData.newBuilder()
             .setHasEnded(true)
-            .setTraceId(new TraceId(1234L, 6789L))
-            .setSpanId(new SpanId(9876L))
+            .setTraceId(TraceId.fromLongs(1234L, 6789L))
```
anuraaga, line 63: A note for potentially following up with moving `fromLongs` to ID generation, there isn't a great reason for unit tests to use the long version vs string version, presumably assertion messages matching the input is an improvement.

exporters/zipkin/src/main/java/io/opentelemetry/exporters/zipkin/ZipkinSpanExporter.java
```diff
@@ -132,16 +133,16 @@ static Span generateSpan(SpanData spanData, Endpoint localEndpoint) {
 
     final Span.Builder spanBuilder =
         Span.newBuilder()
-            .traceId(spanData.getTraceId().toLowerBase16())
-            .id(spanData.getSpanId().toLowerBase16())
+            .traceId(spanData.getTraceId().toString())
```
anuraaga, line 136: Can remove `toString`

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/B3PropagatorInjectorMultipleHeaders.java
```diff
@@ -39,8 +39,8 @@
     SpanContext spanContext = span.getContext();
     String sampled = spanContext.getTraceFlags().isSampled() ? Common.TRUE_INT : Common.FALSE_INT;
 
-    setter.set(carrier, B3Propagator.TRACE_ID_HEADER, spanContext.getTraceId().toLowerBase16());
-    setter.set(carrier, B3Propagator.SPAN_ID_HEADER, spanContext.getSpanId().toLowerBase16());
+    setter.set(carrier, B3Propagator.TRACE_ID_HEADER, spanContext.getTraceIdAsHexString());
+    setter.set(carrier, B3Propagator.SPAN_ID_HEADER, spanContext.getSpanIdAsHexString().toString());
```
anuraaga, line 43:
```suggestion
    setter.set(carrier, B3Propagator.SPAN_ID_HEADER, spanContext.getSpanIdAsHexString());
```

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagator.java
```diff
@@ -189,11 +198,13 @@ private static SpanContext buildSpanContext(String traceId, String spanId, Strin
       int flagsInt = Integer.parseInt(flags);
       TraceFlags traceFlags = ((flagsInt & 1) == 1) ? SAMPLED_FLAGS : NOT_SAMPLED_FLAGS;
 
+      String otelTraceId = StringUtils.padLeft(traceId, MAX_TRACE_ID_LENGTH);
+      String otelSpanId = StringUtils.padLeft(spanId, MAX_SPAN_ID_LENGTH);
+      if (!TraceId.isValid(otelTraceId) || !SpanId.isValid(otelSpanId)) {
+        return SpanContext.getInvalid();
+      }
       return SpanContext.createFromRemoteParent(
```
anuraaga, line 206: For another PR, we might consider moving the validity checks into `createFromRemoteParent` so propagators don't have to worry about it.

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/OtTracerPropagator.java
```diff
@@ -71,8 +74,8 @@ public static OtTracerPropagator getInstance() {
     if (!spanContext.isValid()) {
       return;
     }
-    setter.set(carrier, TRACE_ID_HEADER, spanContext.getTraceId().toLowerBase16());
-    setter.set(carrier, SPAN_ID_HEADER, spanContext.getSpanId().toLowerBase16());
+    setter.set(carrier, TRACE_ID_HEADER, spanContext.getTraceIdAsHexString().toString());
```
anuraaga, line 77:
```suggestion
    setter.set(carrier, TRACE_ID_HEADER, spanContext.getTraceIdAsHexString());
```

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/OtTracerPropagator.java
```diff
@@ -71,8 +74,8 @@ public static OtTracerPropagator getInstance() {
...
+    setter.set(carrier, SPAN_ID_HEADER, spanContext.getSpanIdAsHexString().toString());
```
anuraaga, line 78:
```suggestion
    setter.set(carrier, SPAN_ID_HEADER, spanContext.getSpanIdAsHexString());
```

opentracing_shim/src/test/java/io/opentelemetry/opentracingshim/testbed/TestUtils.java
```diff
@@ -179,8 +179,12 @@ public static void assertSameTrace(List<SpanData> spans) {
     for (int i = 0; i < spans.size() - 1; i++) {
       // TODO - Include nanos in this comparison.
       assertTrue(spans.get(spans.size() - 1).getEndEpochNanos() >= spans.get(i).getEndEpochNanos());
-      assertEquals(spans.get(spans.size() - 1).getTraceId(), spans.get(i).getTraceId());
-      assertEquals(spans.get(spans.size() - 1).getSpanId(), spans.get(i).getParentSpanId());
+      assertEquals(
+          spans.get(spans.size() - 1).getTraceId().toString(),
```
anuraaga, line 183: Don't need `toString()` (presumably it's a better assertion without it too). Next few files have it

anuraaga (approved):
Just nits, thanks!

api/src/main/java/io/opentelemetry/trace/BigendianEncoding.java
jkwatson: heh. I agree with you. However, what you pasted isn't correct (IDEA says it's always true!). I'll figure out hot to make it what you intended. :)

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagator.java
jkwatson: That seems reasonable!

jkwatson:
ok, all nits have been de-nitted!

api/src/main/java/io/opentelemetry/trace/BigendianEncoding.java
jkwatson: check out the update. I think it's now much easier to understand.

anuraaga (approved):
LGTM thanks

jkwatson:
I'm going to hold off a few days to merge this, both to delay to after the 0.8.0 release, and to give others (@carlosalberto and @bogdandrutu) to chime in with any issue they might have, as this is a significant API change.

api/src/main/java/io/opentelemetry/trace/SpanContext.java
```diff
@@ -75,25 +81,55 @@ public static SpanContext create(
...
   public static SpanContext createFromRemoteParent(
-      TraceId traceId, SpanId spanId, TraceFlags traceFlags, TraceState traceState) {
-    return new AutoValue_SpanContext(traceId, spanId, traceFlags, traceState, /* remote=*/ true);
+      String traceId, String spanId, TraceFlags traceFlags, TraceState traceState) {
+    return create(traceId, spanId, traceFlags, traceState, /* remote=*/ true);
   }
 
+  abstract String getTraceId();
+
+  abstract String getSpanId();
```
tylerbenson, line 90: return `CharSequence` instead?

api/src/main/java/io/opentelemetry/trace/SpanId.java
```diff
@@ -65,46 +47,38 @@ public static int getSize() {
...
-   * Returns the invalid {@code SpanId}. All bytes are 0.
+   * Returns the length of the base16 (hex) representation of the {@code SpanId}.
    *
-   * @return the invalid {@code SpanId}.
-   * @since 0.1.0
+   * @since 0.8.0
    */
-  public static SpanId getInvalid() {
-    return INVALID;
+  public static int getBase16Length() {
```
bogdandrutu, line 54: Hex? If you used hex in SpanContext method names.

api/src/main/java/io/opentelemetry/trace/SpanId.java
```diff
@@ -65,46 +47,38 @@ public static int getSize() {
...
-  public void copyBytesTo(byte[] dest, int destOffset) {
-    BigendianEncoding.longToByteArray(id, dest, destOffset);
+  /** Generate a valid {@link SpanId} from the given long value. */
+  public static String fromLong(long id) {
+    char[] result = getBuffer();
+    BigendianEncoding.longToBase16String(id, result, 0);
+    return new String(result);
+  }
+
+  private static char[] getBuffer() {
```
bogdandrutu, line 75: s/getBuffer/getTemporaryBuffer?

api/src/main/java/io/opentelemetry/trace/SpanId.java
```diff
@@ -119,23 +93,8 @@ public void copyBytesTo(byte[] dest, int destOffset) {
...
-   *
-   * @param dest the destination buffer.
-   * @param destOffset the starting offset in the destination buffer.
-   * @throws IndexOutOfBoundsException if {@code destOffset + 2 * SpanId.getSize()} is greater than
-   *     {@code dest.length}.
-   * @since 0.1.0
-   */
-  public void copyLowerBase16To(char[] dest, int destOffset) {
-    BigendianEncoding.longToBase16String(id, dest, destOffset);
+  public static byte[] bytesFromLowerBase16(String src, int srcOffset) {
```
bogdandrutu, line 96: Same, use Hex if you decided to use that in the SC

api/src/main/java/io/opentelemetry/trace/SpanContext.java
```diff
@@ -59,8 +60,13 @@ public static SpanContext getInvalid() {
    * @since 0.1.0
    */
   public static SpanContext create(
-      TraceId traceId, SpanId spanId, TraceFlags traceFlags, TraceState traceState) {
-    return new AutoValue_SpanContext(traceId, spanId, traceFlags, traceState, /* remote=*/ false);
+      String traceId, String spanId, TraceFlags traceFlags, TraceState traceState) {
```
bogdandrutu, line 63: name the params traceIdHex, spanIdHex

api/src/main/java/io/opentelemetry/trace/SpanId.java
jkwatson: yes! should fix this. good catch.

api/src/main/java/io/opentelemetry/trace/SpanId.java
jkwatson: agreed. 👍

api/src/main/java/io/opentelemetry/trace/SpanContext.java
jkwatson: I think we discussed this, and decided to keep it as String for now. We can always open it up to CharSequence later, since it's a broader interface, I think.

api/src/main/java/io/opentelemetry/trace/SpanId.java
jkwatson: doen

api/src/main/java/io/opentelemetry/trace/SpanId.java
jkwatson: done

api/src/main/java/io/opentelemetry/trace/SpanId.java
jkwatson: done

api/src/main/java/io/opentelemetry/trace/SpanContext.java
jkwatson: done

api/src/main/java/io/opentelemetry/trace/SpanContext.java
iNikem: You can change return types that easily. If clients expect `String` they will break on `CharSequence`. You can widen input, not output

api/src/main/java/io/opentelemetry/trace/SpanContext.java
jkwatson: switched all to String now. :)

api/src/main/java/io/opentelemetry/trace/SpanContext.java
jkwatson: true, but this isn't a public method!

carlosalberto (approved):

carlosalberto:
Amazing job @jkwatson sorry for the late review!

anuraaga:
Awesome!

PR #1595, opened 2020-08-26
Preserve the ParentOrElse behavior when changing sampling probability via configuration

### LINKED ISSUE #1526: When using the env vars to set the TraceConfig, the SDK should wrap the probability sampler with a ParentOrElse sampler.

jkwatson:
This seems like the bare minimum to get #1557 unblocked and preserve at least the parenting logic when a user configures a probability sampler. 

Resolves #1526 

Note: we definitely need to figure out a more robust configuration story for samplers, but that's out of scope for this PR.


anuraaga (approved):
LGTM

PR #1598, opened 2020-08-27
Prototype/rough draft of the typed-key proposal

jkwatson:
This is some actual code that demonstrates the ideas for #1580 

I only implemented a few key types, just to keep things concise and easier to see.

### DIFF
--- a/api/src/main/java/io/opentelemetry/common/experimental/KeyedAttributeDemo.java
+++ b/api/src/main/java/io/opentelemetry/common/experimental/KeyedAttributeDemo.java
@@ -16,13 +16,20 @@
 
 package io.opentelemetry.common.experimental;
 
+import static io.opentelemetry.common.experimental.KeyedAttributes.stringKey;
+
 import io.opentelemetry.common.experimental.KeyedAttributes.BooleanArrayKey;
 import io.opentelemetry.common.experimental.KeyedAttributes.BooleanKey;
+import io.opentelemetry.common.experimental.KeyedAttributes.CompoundKey;
 import io.opentelemetry.common.experimental.KeyedAttributes.KeyImpl;
+import io.opentelemetry.common.experimental.KeyedAttributes.MultiAttribute;
 import io.opentelemetry.common.experimental.KeyedAttributes.StringArrayKey;
 import io.opentelemetry.common.experimental.KeyedAttributes.StringKey;
 import io.opentelemetry.common.experimental.ReadableKeyedAttributes.AttributeConsumer;
 import java.math.BigDecimal;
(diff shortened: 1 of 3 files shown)


api/src/main/java/io/opentelemetry/common/experimental/KeyedAttributeDemo.java
```diff
@@ -0,0 +1,65 @@
...
+        KeyedAttributes.newBuilder()
+            .set(HTTP_HEADERS, "x-stuff", "x-api-key")
+            .set(HTTP_METHOD, "PUT")
+            .set(ERROR_HINT, false)
+            .set(NET_HOST_NAME, "localhost")
+            .build();
+
+    // iterate over them like this, using the readable interface.
+    keyedAttributes.forEach(
+        new AttributeConsumer() {
```
tylerbenson, line 48: I think this bothers me a bit. I'd suggest having a consumer only care about a single generic type and skip anything not of that type. If you need to iterate over everything you'd declare the root type.

jkwatson: I agree. I was pondering a catch-all `consume` method for the unknown keys/values.

jkwatson: I added the catch-all method, just for you. ;)

jkwatson: @tylerbenson Just to see what an API might look like, I added support for setting custom attribute types, as well. I'm not sure how an SDK might use these attributes, or how you might implement your "compound attribute" idea with this API, but I thought it was interesting to see how you could store them and retrieve them in this model.

jkwatson: And, I added a `CompoundAttribute` as well, to see how that would work. Incidentally, I think it would be pretty easy to add a `CompoundAttributeValue` to the existing APIs, so I don't think this particular feature is a value-add with the typed-key approach.

jkwatson: @tylerbenson re-reading your comment, I realized I address half of what you were suggesting. Can you suggest how your proposed API would work, with the way that java generics function? I'm confused about how it would work, exactly.

jkwatson:
Interested parties should note that I updated this. I realized over the weekend that I could keep most of the current API as-is, only changing the set method that takes an AttributeValue. This will make manual instrumentation easier to grok, and still enable statically allocated keys when possible.

PR #1600, opened 2020-08-28
Update opentelemetry-proto to latest version.

anuraaga:
Currently, java instrumentation smoke tests fail because we use JSON. collector `file` exporter writes as JSON and the tests read the JSON. But currently the JSON is not compatible between collector and Java because of renamed enums. Updating the proto will allow us to continue to develop against new versions of the collector.

I'm guessing the intent of the metrics protocol change is for the SDK to also adopt similar data structures. I'm not doing that in this PR since I only want to make the SDK JSON-compatible with the latest collectors - at the same time I've probably made them somewhat compatible for metrics too. The conversion of summary to histogram is clearly not ideal, converting double values into long counts, but I think semantically it may capture something reasonable. Anyways given before the change, the SDK isn't compatible at all for metrics with the latest collector, this is probably an improvement until rewriting the SDK.

### DIFF
--- a/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/MetricAdapter.java
+++ b/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/MetricAdapter.java
@@ -16,18 +16,23 @@
 
 package io.opentelemetry.exporters.otlp;
 
+import static io.opentelemetry.proto.metrics.v1.AggregationTemporality.AGGREGATION_TEMPORALITY_CUMULATIVE;
+import static io.opentelemetry.proto.metrics.v1.AggregationTemporality.AGGREGATION_TEMPORALITY_DELTA;
+import static io.opentelemetry.proto.metrics.v1.AggregationTemporality.AGGREGATION_TEMPORALITY_UNSPECIFIED;
+
 import io.opentelemetry.common.Labels;
 import io.opentelemetry.common.ReadableKeyValuePairs.KeyValueConsumer;
 import io.opentelemetry.proto.common.v1.StringKeyValue;
+import io.opentelemetry.proto.metrics.v1.AggregationTemporality;
 import io.opentelemetry.proto.metrics.v1.DoubleDataPoint;
+import io.opentelemetry.proto.metrics.v1.DoubleHistogram;
+import io.opentelemetry.proto.metrics.v1.DoubleHistogramDataPoint;
+import io.opentelemetry.proto.metrics.v1.DoubleSum;
 import io.opentelemetry.proto.metrics.v1.InstrumentationLibraryMetrics;
-import io.opentelemetry.proto.metrics.v1.Int64DataPoint;
+import io.opentelemetry.proto.metrics.v1.IntDataPoint;
+import io.opentelemetry.proto.metrics.v1.IntSum;
 import io.opentelemetry.proto.metrics.v1.Metric;
-import io.opentelemetry.proto.metrics.v1.MetricDescriptor;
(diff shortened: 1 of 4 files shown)


exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/MetricAdapter.java
```diff
@@ -228,40 +239,22 @@ private static Temporality mapToTemporality(Descriptor descriptor) {
...
+      List<MetricData.ValueAtPercentile> valueAtPercentiles,
+      DoubleHistogramDataPoint.Builder builder) {
+
     for (MetricData.ValueAtPercentile valueAtPercentile : valueAtPercentiles) {
-      result.add(
-          ValueAtPercentile.newBuilder()
-              .setPercentile(valueAtPercentile.getPercentile())
-              .setValue(valueAtPercentile.getValue())
-              .build());
+      // TODO(jkwatson): Value of histogram should be long?
```
anuraaga, line 247: @jkwatson :)

jkwatson:
What version of the proto is this updated to? I haven't seen a release since 0.4.0.

anuraaga:
Just latest master 

32a0c7b519d6e26749a44d7ca8b5a39b08d9ef00

Collector seems to track master so I don't know if the proto has a real release process. But if collector moves with master I think SDK needs to as well to keep up.

jkwatson:
I would prefer not to try to chase master with this. I think there are still changes happening rapidly to the metrics protobuf definitions. I would greatly prefer to only used released versions of the proto.

carlosalberto:
> I would greatly prefer to only used released versions of the proto.

+1 on this.

anuraaga:
@bogdandrutu Will you be doing a proto release soon? We should unblock the trace proto incompatibility soon.

bogdandrutu:
Release will happen Monday, you only need to resync and regenerate the code because of https://github.com/open-telemetry/opentelemetry-proto/pull/214, but Java code will not change

anuraaga:
JFYI, I removed the dependency on go / java being protocol compatible in the instrumentation repo's smoke tests so we don't have a rush on fixing this with regards to instrumentation.

jkwatson:
0.5.0 was just released:  https://github.com/open-telemetry/opentelemetry-proto/releases/tag/v0.5.0

So, please update and we'll get this into the 0.8.0 release tomorrow!

bogdandrutu:
@jkwatson hopefully the collector is changed by tonight so we can give a try before releasing

jkwatson:
> @jkwatson hopefully the collector is changed by tonight so we can give a try before releasing

If not, I think we're ok delaying a day. Should we hold off on merging this, after it's been updated for 0.5.0, do you think?

bogdandrutu:
Before releasing this change I would give a try to the collector. I am getting very close to finish that, is under review. After that gets merged will announce on the metrics gitter

bogdandrutu:
All collector code is merged, waiting for the CI to publish a docker image for that build

anuraaga:
Updated submodule to match the tag version.

exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/MetricAdapter.java
jkwatson: I have no remembrance of what this might have meant.

jkwatson:
Should we merge this, or can you test it against the latest collector before merging, @anuraaga ?

anuraaga:
I haven't really been using metrics yet it'd be quite a chore for me to test it. Maybe merge and give a chance to allow someone (@bogdandrutu ?) to test with the snapshot before releasing?

PR #881, opened 2020-02-19
Change Sampler interface

thisthat:
Fixes #880 changing the signature of `io.opentelemetry.sdk.trace.Sampler#shouldSample`

PR #875, opened 2020-02-18
Rename MeterSharedState with MeterProviderSharedState

bogdandrutu:

PR #856, opened 2020-02-13
Add MetricProducer interface and explain how to use

bogdandrutu:
Signed-off-by: Bogdan Cristian Drutu <bogdandrutu@gmail.com>

PR #1611, opened 2020-09-01
Implement spec change to only accept Context as span parent.

Oberon00:
Partial implementation of (now merged) https://github.com/open-telemetry/opentelemetry-specification/pull/875

This is a prototype for spec PR https://github.com/open-telemetry/opentelemetry-specification/pull/875 (spec issue https://github.com/open-telemetry/opentelemetry-specification/issues/510).

To be more minimal, it does not currently add the Context parameter to SpanProcessor.OnStart, but that would be a trivial follow-up.

----

OLD OUTDATED DESCRIPTION

It replaces the parentSpanId + hasRemoteParent in RecordEventsReadableSpan with a full io.grpc.Context. ~~That was the simple part. The part that is a bit funky is that DefaultSpan now has also has a parent Context in addition to still having a SpanContext for itself.~~

All in all, this removes 24 lines according to `git diff --stat`, so as stated in the spec PR, the total complexity does not really increase here.

The most interesting changes are in [SpanBuilderSdk](https://github.com/open-telemetry/opentelemetry-java/pull/1611/files#diff-434694ecb36099f347e4fb895016c33b), [RecordEventsReadableSpan](https://github.com/open-telemetry/opentelemetry-java/pull/1611/files#diff-c803e8dcb044e6f106decf334b55f991) ~~and [DefaultSpan](https://github.com/open-telemetry/opentelemetry-java/pull/1611/files#diff-0013653671a9ada36dbe95609698de7c)~~.

END OLD OUTDATED DESCRIPTION


api/src/main/java/io/opentelemetry/trace/Span.java
```diff
@@ -283,6 +283,14 @@
    */
   SpanContext getContext();
 
+  /**
+   * Returns the parent {@code Context} associated with this {@code Span}.
+   *
+   * @return the parent {@code Context} associated with this {@code Span}.
+   * @since 0.8.0
+   */
+  Context getParent();
```
Oberon00, line 292: This is only used in the OpenTracingShim so far and the usage there is questionable, so we could remove this change if we want to minimize the PR.

api/src/main/java/io/opentelemetry/trace/DefaultTracer.java
```diff
@@ -67,43 +67,32 @@ static NoopSpanBuilder create(String spanName) {
...
-      spanContext = remoteParent;
-      return this;
+      return new DefaultSpan(TracingContextUtils.getSpan(parent).getContext(), parent);
     }
 
     @Override
     public NoopSpanBuilder setParent(Context context) {
       Utils.checkNotNull(context, "context");
-      spanContext = TracingContextUtils.getSpan(context).getContext();
+      isRootSpan = false; // TODO port this fix to mainline
```
Oberon00, line 87: This is a bug I found when writing this. setNoParent and then calling setParent(Context) will ignore the context.

api/src/main/java/io/opentelemetry/trace/Span.java
Oberon00: I simplified the PR in e102deff, but the state at a7fccf470dd250441820c43ae00bca840a1fdd48 is also still interesting.

api/src/main/java/io/opentelemetry/trace/DefaultTracer.java
jkwatson: worth fixing in a separate PR?

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -508,7 +501,7 @@ private Status getStatusWithDefault() {
   }
 
   SpanId getParentSpanId() {
-    return parentSpanId;
+    return TracingContextUtils.getSpan(parent).getContext().getSpanId();
```
anuraaga, line 504: I don't quite understand what we're getting from the change in this PR. It seems we round-trip through `Context` API instead of much more efficiently accessing fields directly?

api/src/main/java/io/opentelemetry/trace/DefaultTracer.java
```diff
@@ -67,43 +67,32 @@ static NoopSpanBuilder create(String spanName) {
...
-      spanContext = parent.getContext();
-      return this;
-    }
-
-    @Override
-    public NoopSpanBuilder setParent(SpanContext remoteParent) {
-      Utils.checkNotNull(remoteParent, "remoteParent");
-      spanContext = remoteParent;
-      return this;
+      return new DefaultSpan(TracingContextUtils.getSpan(parent).getContext());
```
anuraaga, line 81: For example, for a root span this looks almost like this series of expensive no-ops 

`TracingContextUtils.getSpan(TracingContextUtils.withSpan(DefaultSpan.getInvalid(), Context.current())).getContext()`

vs

`DefaultSpan.getInvalid()`

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/data/SpanData.java
```diff
@@ -77,6 +78,15 @@
...
   SpanId getParentSpanId();
 
+  /**
+   * Returns the parent {@code Context}. If the {@code Span} is a root {@code Span}, the Context
+   * will contain no Span or an invalid span.
+   *
+   * @return the parent {@code SpanId} or an invalid SpanId if this is a root {@code Span}.
+   * @since 0.8.0
+   */
+  Context getParent();
```
anuraaga, line 88: Context isn't data, it's a way to move data around in-process, not out-of-process

anuraaga (commented):
I feel this is conflating two unrelated ideas, the parent / child relationship of spans (a logical connection of IDs) and a mechanism for propagating data within a process, often across threads (Context). The end result is contexts have spans, and spans also have contexts - it's not quite circular but it's not a logical relationship I think.

If someone needs to propagate data along with the span, there's no need to do it with as part of parent/child relationship I think, the data can be added to the Context next to the span, e.g., after creating the span, `Context.withValue(valuetopropagate, Context.current())`. If we really need a simpler API for adding arbitrary information to go along with the span (that is how I'm reading the rationale in #875), we could have a first-class field for extra data like [Brave](https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/TraceContext.java#L196). But I find it confusing if we try to use the span parent for this sort of data.

Oberon00:
@anuraaga 

> I feel this is conflating two unrelated ideas, the parent / child relationship of spans (a logical connection of IDs) and a mechanism for propagating data within a process, often across threads (Context).

Please, can we discuss such fundamental design concerns on the spec PR open-telemetry/opentelemetry-specification#875?

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
Oberon00: This spec change (this PR is an implementation of) has two purposes:

1. Enable passing arbitrary information between propagators inject, extract and exporters (and in the future, after #856 and #881 the samplers)
2. Encourage users to properly propagate the context (it's easier to use Tracer.withSpan than doing the whole dance of `TracingContextUtils.contextWithSpan(parentSpan, Context.current()` which is intentionally made more cumbersome. For example, see the [Actor propagation change](https://github.com/open-telemetry/opentelemetry-java/pull/1611/files#diff-fb2061671d30874e4251c2ff1f7c41baR48) which previously contained what I would call a bug, as it lost all information in the Context except for the span.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/data/SpanData.java
Oberon00: I disagree. The inflexibility of SpanContext/TraceState leaves no other choice to transmit nonstandard information from propagators to exporters. See https://github.com/open-telemetry/opentelemetry-java/pull/1611#discussion_r481846420

api/src/main/java/io/opentelemetry/trace/DefaultTracer.java
Oberon00: Fixed in 0eb98299.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/data/SpanData.java
anuraaga: I guess they are data, but not span data which may be what's confusing me. The context could have who knows what, maybe from the user, or some other library, and much of it wouldn't even be related to the span. I don't think it's a good idea to send everything in the Context to the exporter, Context is a bag for known values. If `SpanContext` had `List<Object>` that would make much more sense since it's know to be span data for sure.

api/src/main/java/io/opentelemetry/trace/DefaultTracer.java
Oberon00: I created #1617

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/data/SpanData.java
Oberon00: > If SpanContext had `List<Object>` that would make much more sense since it's know to be span data for sure.

I actually do like that idea. Or probably a `io.grpc.Context` on the SpanContext instead since it has a nice API for retrieving keys in a type-safe manner.

Oberon00: > The context could have who knows what, maybe from the user, or some other library, and much of it wouldn't even be related to the span. I don't think it's a good idea to send everything in the Context to the exporter

I don't see a problem with that though. The exporter would just retrieve any values it knows (often only the parent span context) and ignore the rest.

anuraaga:
I realized that context APIs don't allow iterating over the context, so my concern of leaking unrelated information into the exporter isn't really a concern.

So still about lifecycle, how about an example with the root span. Want to propagate some information with it.

```
Context context = Context.current().withValue(cat, meow);
Span span = tracer.spanBuilder().parent(context).start();
```

I want to add data to the span, so I have to add it to the context. And I add a parent even though it's a root span. This API seems very confusing to me. Perhaps builder has helpers to set the values, but then this would create an internal context forked from current when building the span that probably can't be mounted. I think we would have to change `start` to return `Context` instead of `Span` with such an approach to make clear that the old context needs to be invalidated by propagating this new one. This might make some sense and makes me more comfortable with the change, but it seems like a huge change too.

anuraaga:
Doh sorry wrong PR!

api/src/main/java/io/opentelemetry/trace/DefaultTracer.java
jkwatson: thanks!

jkwatson:
@Oberon00 What would you think about narrowing the scope of this PR to only update the API, rather than also change it to move the Context all the way down into the exporter layer? I think your spec issue has had the scope narrowed recently to only include the API changes.

Oberon00:
> I think your spec issue has had the scope narrowed recently to only include the API changes.

Indeed, this was changed in today's SIG spec meeting. I will update the PR (will probably remove more than half the changes) but I didn't get around to it yet.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/data/SpanData.java
Oberon00: Context is no longer considered storeable data now.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
Oberon00: This no longer applies.

Oberon00:
I updated the PR. Currently it is even more minimal than the spec PR (no Context parameter for OnStart), but that would be trivial follow-up.

jkwatson:
> I updated the PR. Currently it is even more minimal than the spec PR (no Context parameter for OnStart), but that would be trivial follow-up.

thanks!

api/src/test/java/io/opentelemetry/trace/SpanBuilderTest.java
```diff
@@ -32,8 +33,8 @@
   void doNotCrash_NoopImplementation() {
     Span.Builder spanBuilder = tracer.spanBuilder("MySpanName");
     spanBuilder.setSpanKind(Kind.SERVER);
-    spanBuilder.setParent(DefaultSpan.getInvalid());
-    spanBuilder.setParent(DefaultSpan.getInvalid().getContext());
+    spanBuilder.setParent(TracingContextUtils.withSpan(DefaultSpan.create(null), Context.ROOT));
```
jkwatson, line 36: youch. this API sure is ugly! Remind me again how this is going to make things easier/better for our users?

sdk_extensions/testbed/src/test/java/io/opentelemetry/sdk/extensions/trace/testbed/concurrentcommonrequesthandler/RequestHandler.java
```diff
@@ -30,13 +29,13 @@
 
   private final Tracer tracer;
 
-  private final SpanContext parentContext;
+  private final io.grpc.Context parentContext;
```
jkwatson, line 32: I wonder if it might be better to rename the `Context` class in this package to `TestContext` or something, so we don't have the name collision in here.

api/src/test/java/io/opentelemetry/trace/SpanBuilderTest.java
Oberon00: This is a unit test. Normally you would just call setNoParent.

jkwatson: fair enough, but I think having to do parenting via the TCU is going to be pretty counter-intuitive to most users. :(

opentracing_shim/src/main/java/io/opentelemetry/opentracingshim/SpanBuilderShim.java
```diff
@@ -192,11 +195,15 @@ public Span start() {
...
       builder.setNoParent();
     } else if (parentSpan != null) {
-      builder.setParent(parentSpan.getSpan());
+      // Note: We ignore the (potentially stored) parentSpan's (grand)parent context here.
+      builder.setParent(TracingContextUtils.withSpan(parentSpan.getSpan(), Context.current()));
       SpanContextShim contextShim = spanContextTable().get(parentSpan);
       distContext = contextShim == null ? null : contextShim.getCorrelationContext();
     } else if (parentSpanContext != null) {
-      builder.setParent(parentSpanContext.getSpanContext());
+      // TODO: This might be wonky
```
jkwatson, line 203: can you explain the TODO here? Do we need an issue to track a change? I'd prefer not to leave a cryptic TODO in the code like this, since it's not clear what the issue is.

api/src/test/java/io/opentelemetry/trace/SpanBuilderTest.java
Oberon00: I fully agree, but at the same time, I don't think users should normally do that. Usually you would just use the implicit current parent, or sometimes you need to capture a context with `Context.current()` and forward that to somewhere else. If you start a span at one thread and need to continue it at another thread while *not* setting it as active in the current thread is the only place you would need TracingContextUtils. I also think there should be some may to make that more easy. I could imagine `span.inCurrentContext()` (Java 8 default methods?), or `TracingContextUtils.spanInCurrent(span)` to make this easier.

> Remind me again how this is going to make things easier/better for our users?

As you said, using TracingContextUtils is rather unintuitive, and thus I think most users did not do it, they just passed around spans without the remaining context. However, a custom propagator should be able to set a value in the context in extract and rely on it still being available in inject.

BTW, using `inject` is a place where you already have to often muck around with TracingContextUtils currently.

opentracing_shim/src/main/java/io/opentelemetry/opentracingshim/SpanBuilderShim.java
Oberon00: I would need @carlosalberto to check this. It just felt a bit strange, but I don't have enough knowledge of the OpenTracing shim to verify if there is a problem.

Oberon00: Maybe SpanContextShim should store a Context.

Oberon00: I created #1699 and will change this to use ROOT instead of current (that way later changes are less likely to be breaking).

Oberon00: Done in 9f618e0

sdk_extensions/testbed/src/test/java/io/opentelemetry/sdk/extensions/trace/testbed/concurrentcommonrequesthandler/RequestHandler.java
Oberon00: I renamed it to RequestHandlerContext in 6eea962. Is that name OK?

jkwatson (approved):
Thanks!

api/src/test/java/io/opentelemetry/trace/SpanBuilderTest.java
kenfinnigan: Is it safe to assume the implicit current parent is accurate?

For instance, when dealing with asynchronous methods the thread being executed on is usually not the same thread that might have initiated/received the original request. In that situation, the implicit current parent from `ThreadLocal` is more than likely not the parent you want

Oberon00: No, in that case you need to capture a Context (ideally you can use Context.current()) at he parent side. There is no fundamental change here with this PR.

kenfinnigan: If the APIs are geared towards users utilizing "implicit" parents, is it defined anywhere in what situations it results in problems and alternatives should be implemented, such as you describe?

Oberon00: > If the APIs are geared towards users utilizing "implicit" parents

Hmm, I'm not sure yet why you think the API is more geared towards implicit parent's than before? I mean, I agree that it *is* a bit more cumbersome to have a span that was not active be the parent span at another thread, but the difference is only this:

```java
Span capturedParent = span;
```

vs.

```
Context capturedParent = TracingContextUtils.withSpan(span, Context.current());
```

In case the parent is active on the original thread, you can even do

```
Context capturedParent = Context.current();
```

>  is it defined anywhere in what situations it results in problems

It currently uses grpc Context, so unless some other component does some automatic in-process context propagation, it is just a thread local. I guess you should only use an explicit parent if (a) you extract one from a remote process or (b) your parent comes from a remote thread, i.e. this is the first span that continues some asynchronous operation in another "execution context" (think thread) than the logical parent span.

Oberon00: Maybe we can provide some convenience function (default interface implementation?), so one can do ` Context capturedParent = span.withCurrentContext()`.

kenfinnigan: Personally, it's a lot messier, but more importantly way less obvious that that's what is needed to achieve the desired result.

With respect to a default interface implementation, that wouldn't be possible with the JDK 7 requirement, correct?

carlosalberto: (FWIW, we need more convenience functions in `TracingContextUtils`, or in whatever new name it ends up having ;) )

anuraaga (approved):
Thanks @Oberon00!

api/src/test/java/io/opentelemetry/trace/SpanBuilderTest.java
Oberon00: @kenfinnigan Java 7 was dropped recently, Java 8 is now required.

kenfinnigan: Thanks @Oberon00, don't follow things for a few weeks, and everything changes!

PR #1616, opened 2020-09-01
Create a simple example of using the OTLP exporters

jkwatson:
This also includes a docker-compose setup for the collector and friends, so the entire thing is quasi-self-contained.


examples/otlp/docker/app/main.go
```diff
@@ -0,0 +1,128 @@
...
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"math/rand"
+	"os"
+	"time"
+
+	"contrib.go.opencensus.io/exporter/ocagent"
```
bogdandrutu, line 26: I am a bit lost on why the demo app from collector is copied here :)

jkwatson: yeah...I don't know that we need all the pieces of the demo app. I hadn't trimmed them all out yet. It was a great starting point to have a fully-functional collector back-end with visualizations built in. I'll remove the load generator. :)

jkwatson: removed

examples/otlp/docker/otel-agent-config.yaml
```diff
@@ -0,0 +1,38 @@
...
+  opencensus:
+  zipkin:
+    endpoint: :9411
+  jaeger:
+    protocols:
+      thrift_http:
+
+
+exporters:
+  opencensus:
```
anuraaga, line 11: Since this is the java repo and the example is named otlp, should we just use otlp between the two? Or do we even need to run two collectors for this demo?

examples/otlp/docker/otel-collector-config-demo.yaml
```diff
@@ -0,0 +1,54 @@
+receivers:
+  opencensus:
+  otlp:
+    protocols:
+        grpc:
+  otlp/withendpoint:
```
anuraaga, line 6: Unused I think, if it's just for documentation can you add a comment describing it?

examples/otlp/src/main/java/io/opentelemetry/example/OtlpExporterExample.java
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * Example code for setting up the OTLP exporters.
+ *
+ * <p>If you wish to use this code, you'll need to run a copy of the collector locally, on the
+ * default port. There is a docker-compose configuration for doing this in the docker subdirectory
+ * of this module.
+ */
+public class OtlpExporterExample {
+  public static void main(String[] args) throws InterruptedException {
+    // this will make sure that a proper service.name attribute is set on all the spans/metrics.
```
anuraaga, line 42: If I'm not mistaken, it's a must for this to be called before anything under `io.observability` since resource is immutable. Let's add a comment with that warning.

examples/otlp/docker/otel-agent-config.yaml
jkwatson: yeah, I think we can remove the agent for this demo. I'll pull it out.

examples/otlp/docker/otel-collector-config-demo.yaml
jkwatson: I think I can remove this. will do

examples/otlp/src/main/java/io/opentelemetry/example/OtlpExporterExample.java
jkwatson: Yeah, it's a terrible hack to even have to do this. I'll add some comments.

examples/otlp/src/main/java/io/opentelemetry/example/OtlpExporterExample.java
```diff
@@ -0,0 +1,85 @@
...
+    System.setProperty("otel.resource.attributes", "service.name=OtlpExporterExample");
+
+    // set up the span exporter and wire it into the SDK
+    OtlpGrpcSpanExporter spanExporter = OtlpGrpcSpanExporter.getDefault();
+    SimpleSpanProcessor spanProcessor = SimpleSpanProcessor.newBuilder(spanExporter).build();
```
iNikem, line 47: Are you afraid that people will copy-paste this into their production code? May be use BSP?

examples/otlp/docker/docker-compose.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - "13133:13133" # health_check extension
+      - "55678"       # OpenCensus receiver
+      - "55681:55679" # zpages extension
+      - "55680:55680" # otlp receiver
+    depends_on:
+      - jaeger-all-in-one
+      - zipkin-all-in-one
+
+  # Agent
+  otel-agent:
```
iNikem, line 37: It may be confusing for new users, what is Collector and what is Agent and why do they need both. Some documentation may be of help.

examples/otlp/src/main/java/io/opentelemetry/example/OtlpExporterExample.java
jkwatson: 🤷  There are use-cases for both kinds of span processors. If you think it's a worry, I can change it to a BSP.

examples/otlp/docker/docker-compose.yaml
jkwatson: Hm. I hear what you're saying, but this example isn't supposed to be documentation about the collector. It's just supposed to be a plug-and-play way to show things working. Not even the collector demo itself has this documentation (although I think that one definitely should!)

examples/otlp/docker/otel-collector-config-demo.yaml
jkwatson: I'm actually not sure what this is for, even.

examples/otlp/src/main/java/io/opentelemetry/example/OtlpExporterExample.java
jkwatson: done

iNikem (approved):

examples/otlp/docker/docker-compose.yaml
jkwatson: removed the agent to make it less confusing.

examples/otlp/src/main/java/io/opentelemetry/example/OtlpExporterExample.java
jkwatson: switched to BSP. we'll see if this encourages people to use it. :)

examples/otlp/docker/otel-collector-config-demo.yaml
jkwatson: removed

examples/otlp/docker/otel-collector-config-demo.yaml
```diff
@@ -0,0 +1,50 @@
+receivers:
+  opencensus:
```
anuraaga, line 2:
```suggestion
```

anuraaga (approved):
Thanks!

PR #1617, opened 2020-09-02
DefaultTracer: Fix noParent being ignored after explicit parent.

Oberon00:
See https://github.com/open-telemetry/opentelemetry-java/pull/1611#discussion_r481282746

### DIFF
--- a/api/src/test/java/io/opentelemetry/trace/DefaultTracerTest.java
+++ b/api/src/test/java/io/opentelemetry/trace/DefaultTracerTest.java
@@ -115,6 +115,22 @@ class DefaultTracerTest {
   }
 
   @Test
+  void testSpanContextPropagation_fromContextAfterNoParent() {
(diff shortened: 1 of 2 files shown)


api/src/test/java/io/opentelemetry/trace/DefaultTracerTest.java
```diff
@@ -114,6 +114,22 @@ void testSpanContextPropagation_fromContext() {
     assertThat(span.getContext()).isSameAs(spanContext);
   }
 
+  @Test
+  void testSpanContextPropagation_fromContextAfterNoParent() {
```
Oberon00, line 118: I believed this test would fail before this PR initially, but it's the other way round. Still adding this test to cover this case too.

anuraaga (approved):
Thanks!

thisthat (approved):

iNikem (approved):

arminru (approved):

PR #1619, opened 2020-09-02
Add a generic type for the key on the ImmutableKeyValuePairs.

jkwatson:
This is step 1 toward implementing keyed-attributes (#1580). The plan is to use the existing storage for the new style of Attributes. Because the keys will now not be Strings, but be typed-key wrappers for Strings, we need to make the key in the ImmutableKeyValuePairs be generic as well, since I was not planning on changing the Labels implementation to be keyed by anything but plain Strings.

I actually think the addition of the helper interfaces for LabelConsumer and AttributeConsumer is a good change, even if we don't end up implementing keyed-attributes.

In order to not cause a conflict with the AttributesMap as a HashMap extension, I made the AttributesMap wrap a map, rather than extend it.

This also introduced convenience sub-interfaces for consuming labels and attributes, so consumers don't have to worry about the types if they don't want to.

### DIFF
--- a/api/src/main/java/io/opentelemetry/common/AttributeConsumer.java
+++ b/api/src/main/java/io/opentelemetry/common/AttributeConsumer.java
@@ -16,6 +16,10 @@
 
 package io.opentelemetry.common;
 
-/** Convenience interface for consuming {@link ReadableAttributes}. */
+/**
+ * Convenience interface for consuming {@link ReadableAttributes}.
+ *
+ * @since 0.9.0
+ */
 public interface AttributeConsumer
     extends ReadableKeyValuePairs.KeyValueConsumer<String, AttributeValue> {}
--- a/api/src/main/java/io/opentelemetry/common/AttributeConsumer.java
+++ b/api/src/main/java/io/opentelemetry/common/AttributeConsumer.java
@@ -14,21 +14,8 @@
  * limitations under the License.
  */
 
-package io.opentelemetry.sdk.resources;
+package io.opentelemetry.common;
 
-import static org.assertj.core.api.Assertions.assertThat;
-
-import io.opentelemetry.common.AttributeValue;
-import org.junit.jupiter.api.Test;
-
-class ResourceProvidersTest {
-
-  @Test
-  void default_resource_includes_attributes_from_providers() {
-    Resource resource = Resource.getDefault();
-
-    AttributeValue providerAttribute = resource.getAttributes().getValue("providerAttribute");
-    assertThat(providerAttribute).isNotNull();
-    assertThat(providerAttribute.getLongValue()).isEqualTo(42);
-  }
-}
+/** Convenience interface for consuming {@link ReadableAttributes}. */
+public interface AttributeConsumer
+    extends ReadableKeyValuePairs.KeyValueConsumer<String, AttributeValue> {}
--- a/api/src/main/java/io/opentelemetry/common/LabelConsumer.java
+++ b/api/src/main/java/io/opentelemetry/common/LabelConsumer.java
@@ -18,5 +18,15 @@ package io.opentelemetry.common;
 
 import io.opentelemetry.common.ReadableKeyValuePairs.KeyValueConsumer;
 
-/** Convenience interface for consuming {@link Labels}. */
-public interface LabelConsumer extends KeyValueConsumer<String, String> {}
+/**
+ * Convenience interface for consuming {@link Labels}.
+ *
+ * <p>This interface should be considered to be a FunctionalInterface in the java 8+ meaning of that
+ * term.
+ *
+ * @since 0.9.0
+ */
+public interface LabelConsumer extends KeyValueConsumer<String, String> {
+  @Override
+  void consume(String key, String value);
+}
--- a/api/src/main/java/io/opentelemetry/common/LabelConsumer.java
+++ b/api/src/main/java/io/opentelemetry/common/LabelConsumer.java
@@ -14,21 +14,9 @@
  * limitations under the License.
  */
 
-package io.opentelemetry.sdk.resources;
+package io.opentelemetry.common;
 
-import static org.assertj.core.api.Assertions.assertThat;
+import io.opentelemetry.common.ReadableKeyValuePairs.KeyValueConsumer;
 
-import io.opentelemetry.common.AttributeValue;
-import org.junit.jupiter.api.Test;
-
-class ResourceProvidersTest {
-
-  @Test
-  void default_resource_includes_attributes_from_providers() {
-    Resource resource = Resource.getDefault();
-
-    AttributeValue providerAttribute = resource.getAttributes().getValue("providerAttribute");
-    assertThat(providerAttribute).isNotNull();
-    assertThat(providerAttribute.getLongValue()).isEqualTo(42);
-  }
-}
+/** Convenience interface for consuming {@link Labels}. */
+public interface LabelConsumer extends KeyValueConsumer<String, String> {}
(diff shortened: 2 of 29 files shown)


api/src/main/java/io/opentelemetry/common/AttributeConsumer.java
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.common;
+
+/** Convenience interface for consuming {@link ReadableAttributes}. */
```
anuraaga, line 19: I think for these sort of interfaces, at least until we're on Java 8 someday, we can add a comment like. `This is a FunctionalInterface and will not have methods added to it - it is safe to use as a lambda without chance of API breakage. The reason we don't add the annotation is we target Java 7.`

api/src/main/java/io/opentelemetry/common/LabelConsumer.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.common;
+
+import io.opentelemetry.common.ReadableKeyValuePairs.KeyValueConsumer;
+
+/** Convenience interface for consuming {@link Labels}. */
```
anuraaga, line 21: Ditto

api/src/main/java/io/opentelemetry/common/ReadableAttributes.java
```diff
@@ -21,4 +21,6 @@
  *
  * <p>See {@link Attributes} for the public API implementation.
  */
-public interface ReadableAttributes extends ReadableKeyValuePairs<AttributeValue> {}
+public interface ReadableAttributes extends ReadableKeyValuePairs<String, AttributeValue> {
+  AttributeValue getValue(String key);
```
anuraaga, line 25: I don't quite follow the PR description about clashing with `Map` - we don't inherit from `Map` so there's no clash right?

api/src/main/java/io/opentelemetry/common/ReadableKeyValuePairs.java
```diff
@@ -16,39 +16,28 @@
...
 
-import javax.annotation.Nullable;
-
 /**
  * A read-only container for String-keyed attributes.
  *
  * @param <V> The type of the values contained in this.
  */
-public interface ReadableKeyValuePairs<V> {
+public interface ReadableKeyValuePairs<K, V> {
```
anuraaga, line 24: I can't find this type parameter used in #1598. Why do we need a type parameter for the key? Don't we always use `String` for the key? Accessors may use specialized types for type safety though

```
Attributes {
  void set(StringKey, String value)
  void set(BooleanKey, boolean value)
  void set(Key<T>, T value)

  String getString(StringKey);
  Boolean getBoolean(BooleanKey);
}
```

I don't see a generic parameter for the key needed here, am I missing anything?

Also, I realized the typed keys isn't great for getters 😅 It might be a reason why in grpc-context, the "setters" (e.g., Context.withValue) accept key as a parameter, and getting is something like `MyKey.get(context)`. Since we don't have getters though not a big deal just noticed it.

api/src/main/java/io/opentelemetry/common/AttributeConsumer.java
jkwatson: A good point. But, if I'm going ahead with changing the API for #1580 , this one won't be so nice and simple any more, but have lots of methods on it. The LabelConsumer is a different story, though, as I think it will stay simple.

api/src/main/java/io/opentelemetry/common/ReadableAttributes.java
jkwatson: The issue is with the AttributeMap used in the RecordEventsReadableSpan. It does extend Map, and the `get` method on Map clashes with what this has now.

api/src/main/java/io/opentelemetry/common/ReadableKeyValuePairs.java
jkwatson: This will be needed when we switch the key type, since they won't be strings any more. The prototype didn't mess with the underlying implementation, but just stuck with Maps for demo simplicity (note the TODOs in the prototype code).

jkwatson: I updated the PR description with some details that I realized was missing for people other than me. :)

api/src/main/java/io/opentelemetry/common/ReadableAttributes.java
jkwatson: I updated the PR description with some details that I realized was missing for people other than me. :)

api/src/main/java/io/opentelemetry/common/LabelConsumer.java
jkwatson: I added some javadoc here. Let me know what you think.

api/src/main/java/io/opentelemetry/common/ReadableAttributes.java
jkwatson: Maybe a better answer is to finally rip off the bandaid, and make the AttributesMap not extend HashMap.

jkwatson: @anuraaga I made AttributesMap not extend HashMap anymore, and restored the `get` name. Please take a look!

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/AttributesMap.java
```diff
@@ -17,84 +17,83 @@
...
-  private static final long serialVersionUID = 42L;
 
   AttributesMap(long capacity) {
     this.capacity = capacity;
   }
 
-  @Nullable
-  @Override
-  public AttributeValue put(String key, AttributeValue value) {
+  public void put(String key, AttributeValue value) {
```
anuraaga, line 41: I think it's worth returning the existing item like `Map` since it's familiar and opens more use cases without extra work. It's not backwards-compatible to change return value of a method unfortunately.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/AttributesMap.java
```diff
@@ -17,84 +17,83 @@
...
-    return super.remove(key);
-  }
-
-  // Added as public to override the newly added methods in java8, so when we do the switch and
-  // start using them we remember to fix them and count the number of attributes added.
-
-  @SuppressWarnings("MissingOverride")
-  public AttributeValue putIfAbsent(String key, AttributeValue value) {
-    throw new UnsupportedOperationException("Do not call methods on the map");
+  void remove(String key) {
```
anuraaga, line 49: Ditto

anuraaga (approved):
Just easy remaining point otherwise LGTM

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/AttributesMap.java
jkwatson: That's fair. I'll make it so.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/AttributesMap.java
jkwatson: ditto

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/AttributesMap.java
jkwatson: although, this is an internal, non-public class, so I don't think it actually matters much. :)

anuraaga: Oops haha. Yeah then no worries at all

PR #1620, opened 2020-09-03
Remove auto-config

anuraaga:
Moved to instrumentation repo in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1120


thisthat (approved):

jkwatson:
As an author of an externally-hosted exporter, how do I get access to the interfaces I need to implement for the SPI to work? Are they going to be published in a separate artifact that I can consume? This change is catching me a little off-guard, TBH.

anuraaga:
@jkwatson Sorry about that! I should have filed an issue here too in addition to the instrumentation repo.

We are publishing a `javaagent-tooling` artifact with this SPI

https://oss.jfrog.org/libs-snapshot/io/opentelemetry/javaagent/opentelemetry-javaagent-tooling/

Though we might rename this a bit to have an artifact with all of our SPI interfaces only.

Admittedly I didn't think enough about the impact on external exporters so needed to communicate that better. Next agent release won't use this module anymore.

PR #1622, opened 2020-09-04
Have Propagators expose only a singleton instance

### LINKED ISSUE #1372: Have Propagators expose only a singleton instance
At this moment we allow users to create `HttpTraceContext`, `JaegerPropagator` instances, but we should switch to have them expose a singleton instance, as they don't keep state. The `B3` propagators already have this behavior.

malafeev:
fix #1372

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- Extensions:
+    - BREAKING CHANGE: Made Propagators to expose only a singleton instance
+
 ## 0.8.0 - 2020-09-01
 
 - Extensions:
--- a/api/src/jmh/java/io/opentelemetry/trace/propagation/HttpTraceContextExtractBenchmark.java
+++ b/api/src/jmh/java/io/opentelemetry/trace/propagation/HttpTraceContextExtractBenchmark.java
@@ -48,7 +48,7 @@ public class HttpTraceContextExtractBenchmark {
           "00-2e7d0ad2390617702e7d0ad239061770-d49582a2de984b86-01",
           "00-905734c59b913b4a905734c59b913b4a-776ff807b787538a-00",
           "00-68ec932c33b3f2ee68ec932c33b3f2ee-68ec932c33b3f2ee-00");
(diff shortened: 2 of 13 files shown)


anuraaga (approved):
Thanks!

Oberon00 (approved):
Nice improvement!

thisthat (approved):

carlosalberto:
Thanks for the PR! Just wondering whether we should also rename the `B3` ones, e.g. `getSingleHeaderPropagator()` to `getSingleHeaderInstance()` or similar ;)

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- Extensions:
+    - BREAKING CHANGE: Made Propagators to expose only a singleton instance
```
jkwatson, line 6: 👍

jkwatson:
I'm ok with exposing a singleton for those who want it, but I'm wondering if there's real value in *enforcing* the singleton nature of these classes. There's definitely no harm in creating extra instances if someone wanted to, and by making the constructors private, anyone who wants to use a DI framework that will automatically call no-arg constructors (I think guice will do this?) is now required to create a provider to return the singleton.

In general, I think that singletons are something that is the concern mostly of the application using them, not the classes themselves. 

What do you thinking about keeping the constructors public for DI users who don't want to have to do extra work to just call no-arg constructors?

malafeev:
@jkwatson some propagators already are singletons with private constructor. If we will make constructor public and provide static `getInstance()` method it will create more questions. 
If we want to keep propagators consistent the we should make all of them singletons and some already or 
make all constructors of all propagators public and get rid of getInstance method.

For spring it doesn't matter because in any case you need to provide method like
```
@Bean
public TextMapPropagator getPropagator() {
    return // singleton or new instance
}
```  
probably for guice is the same?

jkwatson:
@malafeev guice will inject a no-arg constructor class instance with no additional code.

This probably isn't a big deal, but we should keep in mind that forcing singletons on people can make their lives more annoying.

PR #1625, opened 2020-09-08
Add incubator module with utilities for mutating SpanData.

anuraaga:
Sorry for the huge delay on this.

For #1321


sdk_extensions/tracing_incubator/src/main/java/io/opentelemetry/sdk/extensions/incubator/trace/data/SpanData.java
```diff
@@ -0,0 +1,177 @@
...
+ *   .build();
+ * exporter.export(data);
+ *
+ * }</pre>
+ */
+// AutoValue generated hashCode is fine but we need to define equals to accept the base SpanData
+// type.
+@Immutable
+@AutoValue
+public abstract class SpanData implements io.opentelemetry.sdk.trace.data.SpanData {
```
jkwatson, line 51: Can we name this something else, so we don't have the name collision, requiring using FQNs for hte interface?

jkwatson:
Thanks! Just one small question about naming the one with the builder.

sdk_extensions/tracing_incubator/src/main/java/io/opentelemetry/sdk/extensions/incubator/trace/data/SpanData.java
anuraaga: Not sure why I missed this review, sorry! I could have sworn I added a comment in description about it but didn't - I agree, but couldn't come up with a better name :( Any ideas?

jkwatson: BuildableSpanData? SpanDataExtender? BuiltSpanData? 

Maybe just SpanDataBuilder for now, since it's really a temporary move to see if this is the best approach?

jkwatson:
@anuraaga any updates on the naming here?

carlosalberto:
Is this *really* urgent?

jkwatson:
> Is this _really_ urgent?

It's not, but I'd rather not have open PRs lying around, even if they aren't technically needed for GA.

anuraaga:
@jkwatson Yup sorry for the slowness, finally renamed.

jkwatson (approved):

PR #1627, opened 2020-09-08
Add rationale.md to document design decisions that people may be curi…

### LINKED ISSUE #688: Create a FAQ file
Today in our SIG call there was a question regarding why `Span` does not implement the `Closeable` interface, which has a reason but is not so obvious. Because of this, it would be nice to have a FAQ with common questions/issues.

anuraaga:
…ous about later.

Fixes #688

### DIFF
--- a/docs/rationale.md
+++ b/docs/rationale.md
@@ -1,0 +1,26 @@
+# OpenTelemetry Rationale
+
+When creating a library, often times designs and decisions are made that get lost over time. This
+document tries to collect information on design decisions to answer common questions that may come
+up when you explore the SDK.
+
+## Span not `Closeable`
+
+Because a `Span` has a lifecycle, where it is started and MUST be ended, it seems intuitive that a
+`Span` should implement `Closeable` or `AutoCloseable` to allow usage with Java try-with-resources
+construct. However, `Span`s are unique in that they must still be alive when handling exceptions,
+which try-with-resources does not allow. Take this example:
+
+```java
+Span span = tracer.buildSpan("someWork").start();
+try (Scope scope = tracer.scopeManager().activate(span)) {
+    // Do things.
+} catch(Exception ex) {
+    span.recordException(ex);
+} finally {
+    span.finish();
+}
+```
+
+It would not be possible to call `recordException` if `span` was also using try-with-resources.
+Because this is a common usage for spans, we do not support try-with-resources.
--- a/docs/rationale.md
+++ b/docs/rationale.md
@@ -12,13 +12,13 @@ construct. However, `Span`s are unique in that they must still be alive when han
 which try-with-resources does not allow. Take this example:
 
 ```java
-Span span = tracer.buildSpan("someWork").start();
-try (Scope scope = tracer.scopeManager().activate(span)) {
+Span span = tracer.spanBuilder("someWork").startSpan();
+try (Scope scope = tracer.withSpan(span)) {
     // Do things.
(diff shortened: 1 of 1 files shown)


docs/rationale.md
```diff
@@ -0,0 +1,26 @@
...
+    // Do things.
+} catch (Exception ex) {
+    span.recordException(ex);
+} finally {
+    span.finish();
+}
+```
+
+It would not be possible to call `recordException` if `span` was also using try-with-resources.
+Because this is a common usage for spans, we do not support try-with-resources.
```
Oberon00, line 26: In that case, wouldn't the compiler tell you that span is not available anymore? So you could then switch to using a nested try/catch or a try/finally instead of try().

docs/rationale.md
```diff
@@ -0,0 +1,26 @@
...
+## Span not `Closeable`
+
+Because a `Span` has a lifecycle, where it is started and MUST be ended, it seems intuitive that a
+`Span` should implement `Closeable` or `AutoCloseable` to allow usage with Java try-with-resources
+construct. However, `Span`s are unique in that they must still be alive when handling exceptions,
+which try-with-resources does not allow. Take this example:
+
+```java
+Span span = tracer.buildSpan("someWork").start();
+try (Scope scope = tracer.scopeManager().activate(span)) {
```
Oberon00, line 16: I don't think we have a "scopeManager" method on tracer. Should this be:

```suggestion
try (Scope scope = tracer.withSpan(span)) {
```

carlosalberto (approved):

docs/rationale.md
```diff
@@ -0,0 +1,26 @@
...
+    // Do things.
+} catch (Exception ex) {
+    span.recordException(ex);
+} finally {
+    span.finish();
```
jkwatson, line 21:
```suggestion
    span.end();
```

docs/rationale.md
anuraaga: My reading of the attached issue is this is the rationale for the decision (@arminru you originally provided the link to the OT issue so maybe you can confirm) and for this PR I want to capture it as directly as possible, attempts by me to backport additional rationale or change it would be like rewriting history ;) Also, main motivation for the PR itself is to have a location to continue to add even more rationale when things come up (or more concretely, to knock out a `required-for-ga` issue :)).

arminru: I was not part in making that decision since this was merely taken over from OT (without taking over a rationale doc) and only dug up the explanation in the issue linked above when I was wondering myself.

A nested try/catch would work but not having Span implement Closeable at all, makes it even harder to build a structure that makes recording errors impossible. Personally, I think this should be left to the user to decide how they want to use things (maybe they don't have any means of getting a reasonable exception or error status/message to set and with Span.Status being removed there's nothing left at all to do) but I don't have strong opinion here (one could always add a wrapper to make their span closeable).

Either way, this seems to be the original rationale from back when the decision was made (@carlosalberto please confirm), so it would be the correct reason to state in this doc.

jkwatson (approved):

docs/rationale.md
carlosalberto: That's indeed the original rationale - that's also why this is not allowed (worth adding it?):

```java
try (Scope scope = tracer.spanBuilder().startAsCurrentSpan()) {
} catch (Exception e) {
  // Cannot report the error AND there's no handle to `Span` at this point.
}
```

PR #1628, opened 2020-09-08
Move the TraceFlags isSampled boolean into the SpanContext

### LINKED ISSUE #1603: Consider moving TraceFlags into the SpanContext
It's a wrapper for a single `byte`, so it probably can be moved into the SpanContext without losing functionality or developer ergonomics.

jkwatson:
Resolves #1603

### DIFF
--- a/api/src/main/java/io/opentelemetry/trace/SpanContext.java
+++ b/api/src/main/java/io/opentelemetry/trace/SpanContext.java
@@ -37,7 +37,7 @@ public abstract class SpanContext {
       create(
           TraceId.getInvalid(),
           SpanId.getInvalid(),
-          /* isSampled=*/ false,
+          TraceFlags.getDefault(),
           TraceState.getDefault());
 
   /**
@@ -54,24 +54,20 @@ public abstract class SpanContext {
    *
    * @param traceIdHex the trace identifier of the span context.
    * @param spanIdHex the span identifier of the span context.
-   * @param isSampled if the underlying span should be sampled.
+   * @param traceFlags the byte representation of the {@link TraceFlags}
    * @param traceState the trace state for the span context.
    * @return a new {@code SpanContext} with the given identifiers and options.
    * @since 0.1.0
    */
   public static SpanContext create(
-      String traceIdHex, String spanIdHex, boolean isSampled, TraceState traceState) {
-    return create(traceIdHex, spanIdHex, isSampled, traceState, /* remote=*/ false);
+      String traceIdHex, String spanIdHex, byte traceFlags, TraceState traceState) {
+    return create(traceIdHex, spanIdHex, traceFlags, traceState, /* remote=*/ false);
   }
 
   private static SpanContext create(
-      String traceIdHex,
-      String spanIdHex,
-      boolean isSampled,
-      TraceState traceState,
-      boolean remote) {
+      String traceIdHex, String spanIdHex, byte traceFlags, TraceState traceState, boolean remote) {
     return new AutoValue_SpanContext(
-        traceIdHex, spanIdHex, isSampled, traceState, /* remote$=*/ remote);
+        traceIdHex, spanIdHex, traceFlags, traceState, /* remote$=*/ remote);
   }
 
   /**
@@ -80,14 +76,14 @@ public abstract class SpanContext {
    *
    * @param traceIdHex the trace identifier of the span context.
    * @param spanIdHex the span identifier of the span context.
-   * @param isSampled if the underlying span should be sampled.
+   * @param traceFlags the byte representation of the {@link TraceFlags}
    * @param traceState the trace state for the span context.
    * @return a new {@code SpanContext} with the given identifiers and options.
    * @since 0.1.0
    */
   public static SpanContext createFromRemoteParent(
-      String traceIdHex, String spanIdHex, boolean isSampled, TraceState traceState) {
-    return create(traceIdHex, spanIdHex, isSampled, traceState, /* remote=*/ true);
+      String traceIdHex, String spanIdHex, byte traceFlags, TraceState traceState) {
+    return create(traceIdHex, spanIdHex, traceFlags, traceState, /* remote=*/ true);
   }
 
   abstract String getTraceIdHex();
@@ -137,7 +133,12 @@ public abstract class SpanContext {
   }
 
   /** Whether the span in this context is sampled. */
-  public abstract boolean isSampled();
+  public boolean isSampled() {
+    return (getTraceFlags() & 1) == 1;
+  }
+
+  /** The byte-representation of {@link TraceFlags}. */
+  public abstract byte getTraceFlags();
 
   public void copyTraceFlagsHexTo(char[] dest, int destOffset) {
     dest[destOffset] = '0';
--- a/api/src/main/java/io/opentelemetry/trace/SpanContext.java
+++ b/api/src/main/java/io/opentelemetry/trace/SpanContext.java
@@ -23,9 +23,9 @@ import javax.annotation.concurrent.Immutable;
 /**
  * A class that represents a span context. A span context contains the state that must propagate to
  * child {@link Span}s and across process boundaries. It contains the identifiers (a {@link TraceId
- * trace_id} and {@link SpanId span_id}) associated with the {@link Span} and a set of {@link
- * TraceFlags options}, as well as the {@link TraceState traceState} and the {@link boolean remote}
- * flag.
+ * trace_id} and {@link SpanId span_id}) associated with the {@link Span} and a set of options
+ * (currently only whether the context is sampled or not), as well as the {@link TraceState
+ * traceState} and the {@link boolean remote} flag.
  *
  * @since 0.1.0
  */
@@ -37,7 +37,7 @@ public abstract class SpanContext {
       create(
           TraceId.getInvalid(),
           SpanId.getInvalid(),
-          TraceFlags.getDefault(),
+          /* isSampled=*/ false,
           TraceState.getDefault());
 
   /**
@@ -54,24 +54,24 @@ public abstract class SpanContext {
    *
    * @param traceIdHex the trace identifier of the span context.
    * @param spanIdHex the span identifier of the span context.
-   * @param traceFlags the trace options for the span context.
+   * @param isSampled if the underlying span should be sampled.
    * @param traceState the trace state for the span context.
    * @return a new {@code SpanContext} with the given identifiers and options.
    * @since 0.1.0
    */
   public static SpanContext create(
-      String traceIdHex, String spanIdHex, TraceFlags traceFlags, TraceState traceState) {
-    return create(traceIdHex, spanIdHex, traceFlags, traceState, /* remote=*/ false);
+      String traceIdHex, String spanIdHex, boolean isSampled, TraceState traceState) {
+    return create(traceIdHex, spanIdHex, isSampled, traceState, /* remote=*/ false);
   }
 
   private static SpanContext create(
       String traceIdHex,
       String spanIdHex,
-      TraceFlags traceFlags,
+      boolean isSampled,
       TraceState traceState,
       boolean remote) {
     return new AutoValue_SpanContext(
-        traceIdHex, spanIdHex, traceFlags, traceState, /* remote$=*/ remote);
+        traceIdHex, spanIdHex, isSampled, traceState, /* remote$=*/ remote);
   }
 
   /**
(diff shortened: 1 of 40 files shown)


api/src/main/java/io/opentelemetry/trace/TraceFlags.java
```diff
@@ -16,45 +16,32 @@
...
-   * Returns the size in bytes of the {@code TraceFlags}.
+   * Returns the size in Hex of trace flags.
    *
-   * @return the size in bytes of the {@code TraceFlags}.
-   * @since 0.1.0
+   * @since 0.9.0
    */
-  public static int getSize() {
-    return SIZE;
+  public static int getBase16Size() {
```
anuraaga, line 43: Think we're using `getHexLength`

https://github.com/open-telemetry/opentelemetry-java/blob/master/api/src/main/java/io/opentelemetry/trace/TraceId.java

api/src/main/java/io/opentelemetry/trace/SpanContext.java
```diff
@@ -136,13 +136,13 @@ public String getSpanIdAsHexString() {
...
-   * Returns the {@code TraceFlags} associated with this {@code SpanContext}.
-   *
-   * @return the {@code TraceFlags} associated with this {@code SpanContext}.
-   * @since 0.1.0
-   */
-  public abstract TraceFlags getTraceFlags();
+  /** Whether the span in this context is sampled. */
+  public abstract boolean isSampled();
+
+  public void copyTraceFlagsHexTo(char[] dest, int destOffset) {
```
anuraaga, line 142: Do you think we can just keep this in the w3c propagator? We previously had a data object that corresponded with the w3c concept of traceflags, but now if we don't, it seems we don't need this here in the first place. I guess this comment applies to the entire `TraceFlags` class itself too.

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagator.java
```diff
@@ -206,7 +204,7 @@ public static JaegerPropagator getInstance() {
   private static SpanContext buildSpanContext(String traceId, String spanId, String flags) {
     try {
       int flagsInt = Integer.parseInt(flags);
-      TraceFlags traceFlags = ((flagsInt & 1) == 1) ? SAMPLED_FLAGS : NOT_SAMPLED_FLAGS;
+      boolean traceFlags = ((flagsInt & 1) == 1) ? SAMPLED_FLAGS : NOT_SAMPLED_FLAGS;
```
anuraaga, line 207: Nit but we should rename these variables to `sampled` throughout

api/src/main/java/io/opentelemetry/trace/SpanContext.java
jkwatson: I think that seems fair to me, yes.

api/src/main/java/io/opentelemetry/trace/TraceFlags.java
jkwatson: yep. good catch.

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagator.java
jkwatson: agreed. will do.

api/src/main/java/io/opentelemetry/trace/SpanContext.java
jkwatson: Let's consider that as a follow-on change, if you're ok with that. TraceFlags is used in other places that just the w3c propagator right now.

api/src/main/java/io/opentelemetry/trace/TraceFlags.java
jkwatson: done

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/JaegerPropagator.java
jkwatson: done. I think I caught a few of them. let me know if there are others I missed

api/src/main/java/io/opentelemetry/trace/TraceFlags.java
```diff
@@ -63,191 +50,14 @@ public static int getSize() {
...
-    /**
-     * Builds and returns a {@code TraceFlags} with the desired options.
-     *
-     * @return a {@code TraceFlags} with the desired options.
-     * @since 0.1.0
-     */
-    public TraceFlags build() {
-      return fromByte(options);
-    }
+  public static boolean getDefault() {
```
anuraaga, line 53: Just for context, this is a big reason for removing / moving `TraceFlags` to w3c - it's pretty weird for the default `TraceFlags` to just be a boolean instead of `00` :)

anuraaga (approved):

api/src/main/java/io/opentelemetry/trace/TraceFlags.java
jkwatson: Yep. I agree. I'll create an issue for it, as a follow-up.

jkwatson: #1632

jkwatson:
@bogdandrutu @carlosalberto this is an API change. If you're interested, please review!

api/src/main/java/io/opentelemetry/trace/SpanContext.java
```diff
@@ -54,24 +54,24 @@ public static SpanContext getInvalid() {
    *
    * @param traceIdHex the trace identifier of the span context.
    * @param spanIdHex the span identifier of the span context.
-   * @param traceFlags the trace options for the span context.
+   * @param isSampled if the underlying span should be sampled.
    * @param traceState the trace state for the span context.
    * @return a new {@code SpanContext} with the given identifiers and options.
    * @since 0.1.0
    */
   public static SpanContext create(
```
bogdandrutu, line 62: We should keep this as byte because more bits can be used in the future and we don't want to have to deal with adding more overloads to this method (and all public methods). What do you think?

For the getters we can have a `getTraceFlags` as well as `isSampled` (helper method to access properties in the flags).

carlosalberto: Agreed. Not sure more flags will be added anytime soon, but we should stay (relatively) conservative.

carlosalberto (approved):

carlosalberto:
Overall LGTM, thanks for this PR!

api/src/main/java/io/opentelemetry/trace/SpanContext.java
jkwatson: So, you're suggesting changing the creators to take a `byte`, rather than the `boolean` here? And, exposing the raw byte as well?

jkwatson: @bogdandrutu that's an easy change, if that's what you're suggesting.

anuraaga: I don't think we need to expose the raw byte - I'd suggest `SpanContext` only has the semantic meaning of the flags (isSampled, isBlah, etc). I think I would take more overloads, or adding `SpanContext.Builder` if we come to that later over having `SpanContext` concerned with any sort of marshalling (marshalling is a propagation concept, not a `SpanContext` concept) - the fact that traceflags are a byte that these booleans are marshalled from can be left to the w3c propagator which we're planning on refactoring to, right @jkwatson?

jkwatson: I think so, yes. I think the worry is an explosion of booleans on the creation methods, if more things end up in the TraceFlags down the road.

carlosalberto: Same, slightly concerned with a potential explosion of booleans ;)

jkwatson: My proposal, after discussion in the SIG meeting this morning, is that we keep the boolean parameter, and when/if the W3C introduces more flags, that we can create a Builder for SpanContext that exposes all the possible options (or potentially a bitmasked byte, if that makes sense for more than just the W3C propagator.

anuraaga: I think we've had enough history of a sampling bit to have it as a native boolean here. And when we feel that w3c adds too many booleans we can convert to a builder pattern at the time. It's also possible for the w3c header to increment it's version bit and add completely new stuff, and if everything is a constructor parameter it explodes anyways, there's no way to protect us from constructor explosion if matching w3c 1:1. I suggest one of two options

1) Rename `SpanContext` to `W3CTraceParent` or alternatively to match dotnet naming convention `ParentId` - if we want the span context to logically reflect 1:1 the spec of that header, I find the current naming to be unclear

2) Keep with `SpanContext` being independent of marshalling format, it presents semantics. It ends up being a superset of all possible semantics with regards to tracing.

I lean towards 2) since I don't see why a span context should care whether booleans are passed as a list of booleans or a byte, that's what propagation does, but if the intent of `SpanContext` is to model w3c 1:1 I think the naming is currently not intuitive.

anuraaga: And see @jkwatson's comment as I wrote mine but it's exactly what I'm thinking too.

bogdandrutu: @anuraaga @jkwatson the downside of this is that backwards compatibility is hard to achieve. Let's assume w3c define isBlah as the second least significant bit, and one app in the middle runs a Java version that only propagates isSampled bit (in the current design). Users will have a hard time to start using `isBlah` because they have to upgrade all services to propagate the new property, by propagating the whole `byte` we at least ensure that the property is propagate immediately. Does this argument change your vision?

anuraaga: This is a great example. It doesn't change my vision I think since if pass-through propagation is required, I think that's also a feature of propagation layer for the most part. I think we have similar conversation about `tracestate` and haven't really resolved it. But if w3c requires a byte that comes in coming to be copied outgoing, I would expect that to be implemented by the propagator, maybe by stuffing something into context. Does that sound reasonable?

Oberon00: @anuraaga as we discussed on https://github.com/open-telemetry/opentelemetry-specification/pull/875, currently the Context cannot be used for that, as it would be lost if someone called setParent(Span) or setParent(SpanContext), but I agree it should be useable for that. If we implemented https://github.com/open-telemetry/opentelemetry-specification/pull/875 by merging #1611, it would work. 😃

anuraaga: @oberon00 I personally find our current API to be too span-native to be a big fan of setting `Context` as parent of `Span`. For this or similar discussion, I would consider `Context` equal to map of stuff on `SpanContext` (former context-native, latter span-native) - latter would also be fine here I hope :) 

Also to be clear, I don't intend to block 875 since I added my insight and that's all I can really do. FWIW when it's merged we'll of course also merge the Java PR :-D

Oberon00: Sure. I just wanted to point out that "maybe by stuffing something into the context" wouldn't currently work, if we are talking about the Context object returned by the propagator.

bogdandrutu: > But if w3c requires a byte that comes in coming to be copied outgoing, I would expect that to be implemented by the propagator, maybe by stuffing something into context. Does that sound reasonable?

Not sure if this is not overcomplicates things, we have to do this now then in order to help users for future possible additions to the flags. So I think we better just carry the flags as we received (modify them if needed), and not have to deal with extra things propagated in the context for w3c.

jkwatson: I think some of this boils down to the question: Is OpenTelemetry a W3C-trace-context native implementation? If so, then we should definitely keep the W3C byte, and make other propagators deal with converting their "trace flag" representation into the W3C/OTel byte. If, however, OTel is intended to be propagation-agnostic, then keeping the W3C byte around feels like an odd API choice to make other propagators conform to.

I honestly am on the fence with this one. I don't mind being more aggressively W3C-centric, but it feels like that's a bigger decision than a single person should make.

bogdandrutu: This is a great way to look at things. In my personal opinion we are w3c centric, and that is the reason we invested in that standard (again personal opinion).

bogdandrutu: @jkwatson @anuraaga in order to make progress, I think we should use byte for the moment because that's what we had before, and discuss in a followup PR about changing that to boolean:
1. We had byte before so no change there
2. That is a separate change.

jkwatson: I'm fine with that approach. I have this prepped to go, so I'll push it in a bit.

jkwatson: pushed.

anuraaga:
LGTM makes sense to discuss byte vs boolean separately.

thisthat (approved):

bogdandrutu (approved):

PR #1630, opened 2020-09-09
add javadoc to InMemorySpanExporter shutdown() method

### LINKED ISSUE #536: In-memory Exporter shutdown behavior
I was surprised that calling shutdown on the exporter called shutdown on the tracer. On my first pass, I created an exporter per test class and then called shutdown at the end of the test run for cleanup. The objective being to keep each test clean of side effects.

It took hours to figure out the problem. When I ran single classes inside the IDE no problems but running a command line build of the entire project always failed.

If after further analysis, it makes sense to leave the tracer shutdown code, then the logging message needs to be enhanced so other users don't make the same mistake I did.

malafeev:
fix #536


jkwatson (approved):

jkwatson:
Thanks!

carlosalberto (approved):

anuraaga (approved):
Thanks!

Issue #922, opened 2020-02-25
Context change interceptor


bogdandrutu:
This feature was requested initially in https://github.com/open-telemetry/opentelemetry-java/issues/295 and moved here because all other points are addressed.

One of the goal is to support log correlation, but at one point @kbrockhoff mentioned that what OpenCensus implemented https://github.com/census-instrumentation/opencensus-java/tree/master/contrib/log_correlation/log4j2 is not good, can you please clarify?

bogdandrutu:
@tylerbenson you asked for a prototype on how to build a real interceptor for Context changes, see https://github.com/open-telemetry/opentelemetry-java/pull/923

bogdandrutu:
Ping @kbrockhoff

carlosalberto:
Closing due to lack of activity. @kbrockhoff Please re-open if/as needed ;)

trask:
@carlosalberto I think this is important, can we re-open?

The only way to create an OpenTelemetry scope listener currently is by creating a class with a very special name io.grpc.override.ContextStorageOverride. This seems like a very odd public API surface that we are exposing, and what happens if there is already one of these classes on the classpath?

Scope listeners are important for logging MDC integrations (to populate spanId and traceId into logging MDC).

carlosalberto:
Marking this initially as after-ga. @trask let me know if you think we ought to have this for GA instead.

trask:
The main use case we need this for is logging MDC. Since logging is out of scope for GA, I think this is ok, at least from instrumentation repo perspective.

HaloFour:
Please reconsider addressing this pre-GA.  I understand that logging isn't included in GA, but I think it's critically important for adopters of OTel to be able to correlate logs with tracing.

I am currently also looking at the ContextStorageProvider SPI route to see if I can implement this but it seems like a brittle solution.

jkwatson:
@HaloFour "after ga" in this case mostly means that the current maintainers don't think we can get to it before GA, and that it's not a strict requirement for GA. If you're interested in contributing an implementation, we certainly can review and it and work on getting it merged before GA.

HaloFour:
@jkwatson 

Understood.  That backlog never seems to get shorter, does it?  😄 

I'm still toying around but I was able to move my MDC-managing logic to a `ContextStorage` implementation provided by SPI and my unit tests all pass, so that's a good sign.  I'll have to put more thought into how something like this might be generalized into an API that doesn't involve SPI.

[...]

PR #1633, opened 2020-09-10
Proof of concept: ContextChangeListener

mateuszrzeszutek:
Alternative idea to https://github.com/open-telemetry/opentelemetry-java/issues/922 & https://github.com/open-telemetry/opentelemetry-java/pull/923. This is only an idea, please forgive me lack of unit tests & JavaDocs.

I've decided to treat `ContextChangeListener` similarly to `ContextPropagators` so that it can be used both in traces and correlation contexts (and so that it's accessible from the API, where `Scope` implementation is defined).

There are 2 significant downsides to this solution:
* calling `Context#attach()` and `Context#detach()` manually will not run the listener;
* `ContextUtils#withScopedContext(Context)` needs to be deprecated  -- and it's being used in a lot of places in the java agent.

It would be quite easy to implement MDC support using the new listener:
```java
Class<?> threadContextClass = Class.forName("org.slf4j.MDC");
MethodHandles.Lookup lookup = MethodHandles.publicLookup();
MethodHandle putMethod =
    lookup.findStatic(
        threadContextClass,
        "put",
        MethodType.methodType(void.class, String.class, String.class));
MethodHandle removeMethod =
    lookup.findStatic(
        threadContextClass, "remove", MethodType.methodType(void.class, String.class));

OpenTelemetry.getContextChangeListeners()
    .addContextChangeListener(new LoggingSpanContextSetter(putMethod, removeMethod));

class LoggingSpanContextSetter implements ContextChangeListener {
  private final MethodHandle putMethod;
  private final MethodHandle removeMethod;

  LoggingSpanContextSetter(MethodHandle putMethod, MethodHandle removeMethod) {
    this.putMethod = putMethod;
    this.removeMethod = removeMethod;
  }

  @Override
  public void updated(Context oldContext, Context newContext) {
    try {
      SpanContext newSpanContext = TracingContextUtils.getSpan(newContext).getContext();
      if (newSpanContext.isValid()) {
        putMethod.invoke("otel.trace_id", newSpanContext.getTraceIdAsHexString());
        putMethod.invoke("otel.span_id", newSpanContext.getSpanIdAsHexString());
      } else {
        removeMethod.invoke("otel.trace_id");
        removeMethod.invoke("otel.span_id");
      }
    } catch (Error e) {
      throw e;
    } catch (Throwable ignored) {
    }
  }
}
```

### DIFF
--- a/api/src/main/java/io/opentelemetry/OpenTelemetry.java
+++ b/api/src/main/java/io/opentelemetry/OpenTelemetry.java
@@ -16,6 +16,8 @@
 
 package io.opentelemetry;
 
+import io.opentelemetry.context.ContextManager;
+import io.opentelemetry.context.interceptor.ContextChangeInterceptors;
 import io.opentelemetry.context.propagation.ContextPropagators;
 import io.opentelemetry.context.propagation.DefaultContextPropagators;
 import io.opentelemetry.correlationcontext.CorrelationContextManager;
@@ -186,6 +188,10 @@ public final class OpenTelemetry {
   public static void setPropagators(ContextPropagators propagators) {
     Utils.checkNotNull(propagators, "propagators");
     getInstance().propagators = propagators;
+  }
+
+  public static ContextChangeInterceptors getContextChangeInterceptors() {
+    return ContextManager.getInstance().getContextChangeInterceptors();
   }
 
   /** Lazy loads an instance. */
--- a/api/src/main/java/io/opentelemetry/correlationcontext/CorrelationsContextUtils.java
+++ b/api/src/main/java/io/opentelemetry/correlationcontext/CorrelationsContextUtils.java
@@ -91,8 +91,7 @@ public final class CorrelationsContextUtils {
    * @since 0.1.0
    */
   public static Scope currentContextWith(CorrelationContext corrContext) {
-    Context context = withCorrelationContext(corrContext, Context.current());
-    return ContextUtils.withScopedContext(context);
+    return ContextUtils.withScopedContext(withCorrelationContext(corrContext, Context.current()));
   }
 
   private CorrelationsContextUtils() {}
--- a/context_prop/src/main/java/io/opentelemetry/context/ContextManager.java
+++ b/context_prop/src/main/java/io/opentelemetry/context/ContextManager.java
@@ -1,0 +1,53 @@
+package io.opentelemetry.context;
+
+import io.grpc.Context;
+import io.opentelemetry.context.interceptor.ContextChangeInterceptors;
+import io.opentelemetry.context.interceptor.DefaultContextChangeInterceptors;
+
+// TODO: better name
+public final class ContextManager {
+  private static final ContextManager INSTANCE = new ContextManager();
+
+  private final ContextChangeInterceptors contextChangeInterceptors =
+      new DefaultContextChangeInterceptors();
+
+  public static ContextManager getInstance() {
+    return INSTANCE;
+  }
+
+  public ContextChangeInterceptors getContextChangeInterceptors() {
+    return contextChangeInterceptors;
+  }
+
+  public Scope withScopedContext(Context context) {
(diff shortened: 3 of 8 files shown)


jkwatson:
@trask you had some interest in this, I believe!

jkwatson:
@iNikem ping on MDC context-related issue!

iNikem:
@mateuszrzeszutek Have you thought about promoting `ContextInScope` to a higher level component and let it handle a list of listeners? This way we don't need to change `ContextUtils#withScopedContext(Context)` method. Current listener passing  seems suboptimal :( Like different context changing cases may need different listeners, which is misleading.

anuraaga:
Since the listeners only listen but can't update the context, it seems to miss a lot of utility. What's so special about our context that it shouldn't also accept writes like MDC does ;) It may not have enough utility to warrant a dedicated API but we should keep in mind whether this approach of listening is powerful enough.

Also, after writing instrumentation for logback / log4j, bypassing MDC, I'm starting to feel comfortable to implement MDC using bytecode manipulation of the gRPC Storage class - it doesn't have the caveat that `Storage.attach` are missed and with bytecode manipulation we don't have the restriction if gRPC's "SPI". And when library instrumentation is needed, we still have logback / log4j covered.

For context, I added the logging instrumentation since ids are important but wanted to avoid putting too much effort into a new API before GA.

Not saying we don't need a proper API for context interception but I'm leaning towards designing it when another use case comes up that may also, for example, need to mutate the Context.

How does that sound?

iNikem:
@anuraaga do I understand you correctly, that one of your suggestions/options is to auto-instrument gRPC Storage/Context class and to implement context change notification that way?

If yes, then I think this is a bad idea :) ContextChangeListener is a good general purpose functionality. Having it only when using agent is too restrictive.

iNikem:
I am OK with implementing MDC support via library-specific bytecode modification, as @anuraaga did for Logback.

anuraaga:
> do I understand you correctly, that one of your suggestions/options is to auto-instrument gRPC Storage/Context class and to implement context change notification that way?

Yeah only for solving MDC, not as a generic mechanism. There are concerns with a context listener I have

- Not handling Context.attach - seems like a dealbreaker to me
- Is listening enough? I think most similar APIs I've seen allow mutating the context 
- Is this a global concern (attached to `OpenTelemetry`)? Listeners could be attached to contexts themselves to make listeners scoped to, for example, a specific instrumentation module. Maybe no use case just brainstorming
- API vs SDK?

Many questions - my thinking is if MDC is the only use case, we have manual instrumentation support for logback / log4j, agent support for logback and soon log4j, and can easily add agent support for slf4j if it seems helpful. This seems enough to solve the MDC use case for now well enough. Having other use cases would help us answer questions before committing to this API.

iNikem:
> add agent support for slf4j if it seems helpful.

you mean similarly to your logback instrumentation? By injecting wrapping logging events and settings things in their MDC?

Yeah, I start feeling the same way.

mateuszrzeszutek:
> Is listening enough? I think most similar APIs I've seen allow mutating the context

mateuszrzeszutek:
> Is listening enough? I think most similar APIs I've seen allow mutating the context
I've just implemented intercepting `Context` changes with a possibility to modify the context - but this does not solve any other problems (`Context.attach`), and I believe makes the whole POC unnecessarily complicated. And there's no other use case besides MDC currently.

I'm also in favor of instrumenting each logging library, this seems much simpler.

mateuszrzeszutek:
I've added 2 issues for [log4j1](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1197) and [log4j2](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1196). In case of slf4j there's nothing to hook into, so we'll have to depend on the actual logging implementations' instrumentations to set MDC data.

PR #1637, opened 2020-09-11
Allow forward slash in metric name

Noremac201:
Allow metric names to have a forward slash in them.

I couldn't find consistent and explicit naming conventions, but it appears that other languages (like otel-go) are not as strict, and thus this poses problems when defining a single metrics pipeline with different language agents since you have to define metrics to the least common denominator.


jkwatson:
Hi @Noremac201 ! Thanks for doing the work on this, but unfortunately, the spec is pretty clear that metric names can't include slashes. https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#instrument-naming-requirements

If we want to change this for the Java implementation, we'll need to get the spec updated to include slashes.

Issue #1551, opened 2020-08-18
Configuration to disable ResourceProviders


iNikem:
As a follow up for #1548 there should be a configuration option to disable any `ResourceProvider` service provider found on the classpath. E.g. add env variable/system property with a coma-separated list of class names of  `ResourceProvider` that should be ignored during Resource composition.

iNikem:
Probably should be assigned to me.

anuraaga:
@zoercai Sorry for the delay I finally found your GitHub handle :) I think this is a good issue to get familiar with the SDK, would you be able to take it on?

@jkwatson Please reassign if task accepted (I confirmed with @iNikem already).

zoercai:
@anuraaga Yep I'll take a look, thanks!

jkwatson:
Done! Thanks for picking it up, @zoercai !

zoercai:
This can be closed now from #1640

jkwatson:
closed via #1640

PR #1640, opened 2020-09-14
Add config to allow disabling of resource providers

zoercai:
For https://github.com/open-telemetry/opentelemetry-java/issues/1551

### DIFF
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
@@ -61,6 +61,11 @@ public abstract class Resource {
           .build()
           .merge(TELEMETRY_SDK)
           .merge(readResourceFromProviders());
+  private static final ResourcesConfig RESOURCES_CONFIG =
+      ResourcesConfig.getDefault().toBuilder()
+          .readEnvironmentVariables()
+          .readSystemProperties()
+          .build();
 
   @Nullable
   private static String readVersion() {
@@ -78,6 +83,9 @@ public abstract class Resource {
   private static Resource readResourceFromProviders() {
     Resource result = Resource.EMPTY;
     for (ResourceProvider resourceProvider : ServiceLoader.load(ResourceProvider.class)) {
+      if (RESOURCES_CONFIG.getDisabledResourceProviders().contains(resourceProvider.toString())) {
+        continue;
+      }
       result = result.merge(resourceProvider.create());
     }
     return result;
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
@@ -82,7 +82,7 @@ public abstract class Resource {
     for (ResourceProvider resourceProvider : ServiceLoader.load(ResourceProvider.class)) {
       if (RESOURCES_CONFIG
           .getDisabledResourceProviders()
-          .contains(resourceProvider.getClass().getSimpleName())) {
+          .contains(resourceProvider.getClass().getName())) {
         continue;
       }
       result = result.merge(resourceProvider.create());
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
@@ -62,10 +62,7 @@ public abstract class Resource {
           .merge(TELEMETRY_SDK)
           .merge(readResourceFromProviders());
   private static final ResourcesConfig RESOURCES_CONFIG =
-      ResourcesConfig.newBuilder()
-          .readEnvironmentVariables()
-          .readSystemProperties()
-          .build();
+      ResourcesConfig.newBuilder().readEnvironmentVariables().readSystemProperties().build();
 
   @Nullable
   private static String readVersion() {
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
@@ -80,7 +80,9 @@ public abstract class Resource {
   private static Resource readResourceFromProviders() {
     Resource result = Resource.EMPTY;
     for (ResourceProvider resourceProvider : ServiceLoader.load(ResourceProvider.class)) {
-      if (RESOURCES_CONFIG.getDisabledResourceProviders().contains(resourceProvider.getName())) {
+      if (RESOURCES_CONFIG
+          .getDisabledResourceProviders()
+          .contains(resourceProvider.getClass().getSimpleName())) {
         continue;
       }
       result = result.merge(resourceProvider.create());
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
@@ -61,8 +61,6 @@ public abstract class Resource {
           .build()
           .merge(TELEMETRY_SDK)
           .merge(readResourceFromProviders());
-  private static final ResourcesConfig RESOURCES_CONFIG =
-      ResourcesConfig.newBuilder().readEnvironmentVariables().readSystemProperties().build();
 
   @Nullable
   private static String readVersion() {
@@ -78,9 +76,11 @@ public abstract class Resource {
   }
 
   private static Resource readResourceFromProviders() {
+    ResourcesConfig resourcesConfig =
+        ResourcesConfig.newBuilder().readEnvironmentVariables().readSystemProperties().build();
     Resource result = Resource.EMPTY;
     for (ResourceProvider resourceProvider : ServiceLoader.load(ResourceProvider.class)) {
-      if (RESOURCES_CONFIG
+      if (resourcesConfig
           .getDisabledResourceProviders()
           .contains(resourceProvider.getClass().getName())) {
         continue;
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
@@ -62,7 +62,7 @@ public abstract class Resource {
           .merge(TELEMETRY_SDK)
           .merge(readResourceFromProviders());
   private static final ResourcesConfig RESOURCES_CONFIG =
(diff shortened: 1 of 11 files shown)


zoercai:
Hi @anuraaga ! Could you please take a look at the changes here and let me know if this is on the right track? I will add tests if it is. Thanks!

sdk/common/src/main/java/io/opentelemetry/sdk/resources/ResourcesConfig.java
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * <ul>
+ *   <li>{@code otel.config.resources.providers.disable}: to set the ResourceProvider service
+ *       providers found on the classpath to be disabled.
+ * </ul>
+ *
+ * <p>For environment variables, {@link ResourcesConfig} will look for the following names:
+ *
+ * <ul>
+ *   <li>{@code OTEL_CONFIG_RESOURCES_PROVIDERS_DISABLE}: to set the ResourceProvider service
```
thisthat, line 48: I would change this name to `OTEL_JAVA_DISABLE_RESOURCES_PROVIDES`. According to the spec, this is a language-specific feature which requires the convention `OTEL_{LANGUAGE}_{FEATURE}`: [Spec link](https://github.com/open-telemetry/opentelemetry-specification/blob/14b5b6a944e390e368dd2e2ef234d220d8287d19/specification/sdk-environment-variables.md#language-specific-environment-variables)

sdk/common/src/main/java/io/opentelemetry/sdk/resources/ResourcesConfig.java
```diff
@@ -0,0 +1,176 @@
...
+@Immutable
+public abstract class ResourcesConfig {
+  // These values are the default values for all the global parameters.
+  private static final ImmutableSet<String> DEFAULT_RESOURCES_PROVIDERS_DISABLE = ImmutableSet.of();
+
+  /**
+   * Returns the default {@code ResourcesConfig}.
+   *
+   * @return the default {@code ResourcesConfig}.
+   * @since 0.1.0
```
thisthat, line 62:
```suggestion
   * @since 0.9.0
```

sdk/common/src/main/java/io/opentelemetry/sdk/resources/ResourcesConfig.java
```diff
@@ -0,0 +1,176 @@
...
+
+    /**
+     * * Sets the configuration values from system properties for only the available keys.
+     *
+     * @return this.
+     */
+    @Override
+    public Builder readSystemProperties() {
+      return super.readSystemProperties();
+    }
```
thisthat, line 151: You can remove these methods

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -78,6 +83,9 @@ private static String readVersion() {
   private static Resource readResourceFromProviders() {
     Resource result = Resource.EMPTY;
     for (ResourceProvider resourceProvider : ServiceLoader.load(ResourceProvider.class)) {
+      if (RESOURCES_CONFIG.getDisabledResourceProviders().contains(resourceProvider.toString())) {
```
jkwatson, line 86: I'm concerned with this relying on the `toString()` implementation of a given `ResourceProvider`.  In general, `toString` should be reserved for debugging purposes, rather than actual runtime functionality. If we need to have a way to identify resource provider implementations, we should add a method to the interface for this purpose. I recommend something like `getId()` or `getIdentifier()` or `getName()`.

sdk/common/src/main/java/io/opentelemetry/sdk/resources/ResourcesConfig.java
```diff
@@ -0,0 +1,145 @@
...
+ * Class that holds global resources parameters.
+ *
+ * <p>Configuration options for {@link ResourcesConfig} can be read from system properties,
+ * environment variables, or {@link Properties} objects.
+ *
+ * <p>For system properties and {@link Properties} objects, {@link ResourcesConfig} will look for
+ * the following names:
+ *
+ * <ul>
+ *   <li>{@code otel.java.disable.resources.providers}: to set the ResourceProvider service
```
anuraaga, line 39: I think properties are usually nouns, not verbs. So `disabled`?

anuraaga, line 39: `to set the fully qualified class names of {@link ResourceProvider} implementations that are found on the classpath but should be disabled`

sdk/common/src/main/java/io/opentelemetry/sdk/resources/ResourcesConfig.java
```diff
@@ -0,0 +1,145 @@
...
+ * <ul>
+ *   <li>{@code OTEL_JAVA_DISABLE_RESOURCES_PROVIDERS}: to set the ResourceProvider service
+ *       providers found on the classpath to be disabled.
+ * </ul>
+ */
+@AutoValue
+@Immutable
+public abstract class ResourcesConfig {
+  // These values are the default values for all the global parameters.
+  private static final ImmutableSet<String> OTEL_JAVA_DISABLE_RESOURCES_PROVIDERS =
```
anuraaga, line 54: I didn't look at other config classes and suspect we may have a pattern here - I find it hard to read that we have two constants with the same name here. I think it's fine to just inline `ImmutableSet.of()` in `newBuilder()`

sdk/common/src/main/java/io/opentelemetry/sdk/resources/ResourcesConfig.java
```diff
@@ -0,0 +1,142 @@
...
+     */
+    @VisibleForTesting
+    @Override
+    protected Builder fromConfigMap(
+        Map<String, String> configMap, NamingConvention namingConvention) {
+      configMap = namingConvention.normalize(configMap);
+
+      String stringValue = getStringProperty(OTEL_JAVA_DISABLE_RESOURCES_PROVIDERS, configMap);
+      if (stringValue != null) {
+        this.setDisabledResourceProviders(ImmutableSet.copyOf(stringValue.split(",")));
```
anuraaga, line 116: Use a Guava `Splitter` and do `omitEmptyValues` and `trimResults`

https://guava.dev/releases/23.0/api/docs/com/google/common/base/Splitter.html

sdk/common/src/main/java/io/opentelemetry/sdk/resources/ResourcesConfig.java
```diff
@@ -0,0 +1,142 @@
...
+      }
+      return this;
+    }
+
+    /**
+     * @param disabledResourceProviders the ResourceProvider service providers found on the
+     *     classpath to be disabled.
+     * @return this.
+     */
+    public abstract Builder setDisabledResourceProviders(Set<String> disabledResourceProviders);
```
anuraaga, line 126: Interestingly we already are using `disabled` here :-)

sdk_extensions/resources/src/main/java/io/opentelemetry/sdk/extensions/resources/ProcessResource.java
```diff
@@ -79,4 +81,9 @@ protected Attributes getAttributes() {
 
     return attributes.build();
   }
+
+  @Override
+  public String getName() {
```
anuraaga, line 86: Instead of adding `getName` which might have conflicting values and is a bit annoying to implement in resource providers, let's just match `getClass().getName()` vs the property

sdk/common/src/main/java/io/opentelemetry/sdk/resources/ResourcesConfig.java
anuraaga: And I guess I'd probably go with `otel.java.disabled.resource_providers`. I can imagine us also having `otel.java.disabled.tracer_providers` in the future as a similar pattern.

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -78,6 +80,11 @@ private static String readVersion() {
   private static Resource readResourceFromProviders() {
     Resource result = Resource.EMPTY;
     for (ResourceProvider resourceProvider : ServiceLoader.load(ResourceProvider.class)) {
+      if (RESOURCES_CONFIG
+          .getDisabledResourceProviders()
+          .contains(resourceProvider.getClass().getSimpleName())) {
```
anuraaga, line 85:
```suggestion
          .contains(resourceProvider.getClass().getName())) {
```

So it's fully qualified

sdk/common/src/main/java/io/opentelemetry/sdk/resources/ResourcesConfig.java
```diff
@@ -0,0 +1,143 @@
...
+ * <ul>
+ *   <li>{@code otel.java.disabled.resource_providers}: to set the fully qualified class names of
+ *       {@link ResourceProvider} implementations that are found on the classpath but should be
+ *       disabled.
+ * </ul>
+ *
+ * <p>For environment variables, {@link ResourcesConfig} will look for the following names:
+ *
+ * <ul>
+ *   <li>{@code OTEL_JAVA_DISABLED_RESOURCES_PROVIDERS}: to set the ResourceProvider service
```
anuraaga, line 48: Copy over comment from above

anuraaga (approved):
Just a couple small points, but LGTM. Thanks for the help!

sdk/common/src/main/java/io/opentelemetry/sdk/resources/ResourcesConfig.java
```diff
@@ -0,0 +1,146 @@
...
+ * <ul>
+ *   <li>{@code otel.java.disabled.resource_providers}: to set the fully qualified class names of
+ *       {@link ResourceProvider} implementations that are found on the classpath but should be
+ *       disabled.
+ * </ul>
+ *
+ * <p>For environment variables, {@link ResourcesConfig} will look for the following names:
+ *
+ * <ul>
+ *   <li>{@code OTEL_JAVA_DISABLED_RESOURCES_PROVIDERS}: to set the fully qualified class names of
```
thisthat, line 48: Please add this also to `package-info.java`

thisthat (approved):
Just a final small point. Thank you for your contribution @zoercai :)

jkwatson (approved):

PR #1643, opened 2020-09-15
Case-insensitive names for metric instruments

thisthat:
This PR properly addresses the requirements for metric instrument names.
[Spec link](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#instrument-naming-requirements)

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractInstrument.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractInstrument.java
@@ -90,7 +90,7 @@ abstract class AbstractInstrument implements Instrument {
       implements Instrument.Builder {
     /* VisibleForTesting */ static final String ERROR_MESSAGE_INVALID_NAME =
         "Name should be a ASCII string with a length no greater than "
-            + StringUtils.NAME_MAX_LENGTH
+            + StringUtils.METRIC_NAME_MAX_LENGTH
             + " characters.";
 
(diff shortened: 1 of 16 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractInstrument.java
```diff
@@ -88,10 +88,9 @@ public int hashCode() {
 
   abstract static class Builder<B extends AbstractInstrument.Builder<?>>
       implements Instrument.Builder {
-    /* VisibleForTesting */ static final int NAME_MAX_LENGTH = 255;
     /* VisibleForTesting */ static final String ERROR_MESSAGE_INVALID_NAME =
         "Name should be a ASCII string with a length no greater than "
-            + NAME_MAX_LENGTH
+            + StringUtils.NAME_MAX_LENGTH
```
jkwatson, line 93: while we're here, can we rename this to `METRIC_NAME_MAX_LENGTH` ?

jkwatson (approved):
Thanks! Just one small change requested.

anuraaga (approved):
Thanks!

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractInstrument.java
thisthat: Done in 42c17f1

Issue #1649, opened 2020-09-15
Drop Java 7 support for both the API and SDK


jkwatson:
Let's make the minimum JDK requirement java 8.

jkwatson:
@AloisReitbauer @Oberon00 @thisthat If we /you can create a fork for dynatrace ASAP, we could get this done before GA.

Oberon00:
Duplicate of #1543

jkwatson:
> Duplicate of #1543

Disagree. That issue is about making a decision. This issue is about executing that decision. Dynatrace is the only vendor who has interest in Java 7 support at this point, and out-of-band discussion has indicated that you/they would be amenable to maintaining a Java 7 fork of the project.

trask:
Since Dynatrace is the only vendor that requires Java 7 support in OpenTelemetry, it sounds like creating a Java 7 fork (e.g. https://github.com/open-telemetry/opentelemetry-java7) with Dynatrace folks as maintainers would be a good way forward from this impasse.

Btw, other vendors are planning to support the small number of customers that are still on Java 7 by using the prior generation of their java agent/tracer, not sure if this is an option for Dynatrace as well?

danielkhan:
This was a misunderstanding and Dynatrace will not fork, maintain, and support OTel for Java 7 openly.

Oberon00:
>> Duplicate of #1543
> Disagree. That issue is about making a decision.

You have a point, but then why is the issue for making the decision still open while we already haven an issue for executing a decision? 😉

jkwatson:
> @jkwatson
> 
> > > Duplicate of #1543
> > 
> > 
> > Disagree. That issue is about making a decision.
> 
> You have a point, but then why is the issue for making the decision still open while we already have an issue for executing a decision? 😉

Honestly, it was to force the decision to happen, and make sure that it didn't get stalled, which it was looking like it was.

Oberon00:
Why open a new issue though, instead of commenting on the old one?

jkwatson:
> This was a misunderstanding and Dynatrace will not fork, maintain, and support OTel for Java 7 openly.

So, does this mean Dynatrace is ok with us moving to java 8, or not?

Oberon00:
I commented on #1543.

AkshayBabbar:
hi want to take it this up how to proceed any suggestions.

Oberon00:
[...]

PR #1650, opened 2020-09-16
SDK: Remove Javadocs since tags (for pre 1.0 release)

JasonXZLiu:
~This PR adds Javadocs ‘since’ tags to the SDK for public methods (related to this issue: open-telemetry#344). The versions that are used for each tag were found going through each release to see when the method was added.~

This PR removes Javadocs `@since` tags from the SDK for public methods (related to this issue: open-telemetry#344 and see discussion below). The `@since` tags were removed to maintain consistency while the API is still changing.

### DIFF
--- a/sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
+++ b/sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
@@ -38,7 +38,6 @@ public final class OpenTelemetrySdk {
    * Returns a {@link TracerSdkProvider}.
    *
    * @return TracerProvider returned by {@link OpenTelemetry#getTracerProvider()}.
-   * @since 0.1.0
    */
   public static TracerSdkProvider getTracerProvider() {
     return (TracerSdkProvider) ((Obfuscated<?>) OpenTelemetry.getTracerProvider()).unobfuscate();
@@ -48,7 +47,6 @@ public final class OpenTelemetrySdk {
    * Returns a {@link MeterSdkProvider}.
    *
    * @return MeterProvider returned by {@link OpenTelemetry#getMeterProvider()}.
-   * @since 0.1.0
    */
   public static MeterSdkProvider getMeterProvider() {
     return (MeterSdkProvider) OpenTelemetry.getMeterProvider();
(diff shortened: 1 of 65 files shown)


anuraaga:
Hi @JasonXZLiu - really sorry for not chiming in on the issue earlier >< I think instead of adding `@since` tags I would remove all of our existing ones - we continue to churn on the API and see significant changes coming ahead with things like #1649. Basically, there's nothing guaranteed by our API right now and the since tags are just extra maintenance. I want us to release 1.0 with no `@since` tags anywhere, and then after that release we add the tag when adding new APIs. @jkwatson what do you think?

jkwatson:
> Hi @JasonXZLiu - really sorry for not chiming in on the issue earlier >< I think instead of adding `@since` tags I would remove all of our existing ones - we continue to churn on the API and see significant changes coming ahead with things like #1649. Basically, there's nothing guaranteed by our API right now and the since tags are just extra maintenance. I want us to release 1.0 with no `@since` tags anywhere, and then after that release we add the tag when adding new APIs. @jkwatson what do you think?

Interesting thought. I hadn't pondered it much, but it does seem reasonable. Or, set all the `@since` to 1.0 just prior to release of 1.0. either one would be fine.

@JasonXZLiu I don't think we should put generated javadoc in the repo like this. We're already getting our javadoc published via javadoc.io, and I don't think it's worth the extra maintenance effort to also host generated javadoc in the repo itself.

jkwatson:
BTW, I hadn't ever seen those instructions for generating javadoc before. We should, IMO, get rid of that and just rely on javadoc.io for our javadoc.

JasonXZLiu:
@anuraaga @jkwatson Oh, I see. No worries at all. I'll remove the docs from this PR, and following @anuraaga's approach, I can remove all the `@since` tags from the SDK.

jkwatson:
Please don't delete the existing docs in the `doc/` directory.

JasonXZLiu:
@jkwatson Sorry about that! I just noticed that and rebased my branch (to remove my docs changes entirely).

jkwatson:
Thanks! @anuraaga look ok to you?

jkwatson (approved):

JasonXZLiu:
@jkwatson @anuraaga It seems like some integration tests are failing (with `Container startup failed`) even though the only changes are on documentation. I was wondering if the tests can be re-run or if I'm missing something. Thank you!

jkwatson:
> @jkwatson @anuraaga It seems like some integration tests are failing (with `Container startup failed`) even though the only changes are on documentation. I was wondering if the tests can be re-run or if I'm missing something. Thank you!

Looks like a problem pulling docker images. I'm rerunning the job.

anuraaga (approved):
Thanks a lot for bearing with us!

PR #1656, opened 2020-09-16
Reverse the precedence on Labels and Attributes to prefer the last one added, rather than the first.

### LINKED ISSUE #1642: Labels with duplicate key must take the last value
**Describe the bug**
According to the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#labels), labels with a duplicated key must take the last value but in the current implementation, the first one is taken.

**Steps to reproduce**
https://github.com/open-telemetry/opentelemetry-java/blob/a4cd21ff49f7de495c031fac469a367311f7636a/api/src/test/java/io/opentelemetry/common/LabelsTest.java#L77-L86

```java
  @Test
  void deduplication() {
    Labels one =
        Labels.of(
            "key1", "value1",
            "key1", "valueX");
    Labels two = Labels.of("key1", "valueX");

    assertThat(one).isEqualTo(two);
  }
```

jkwatson:
resolves #1642 

Note: this necessitated changing the way Resources were merged, and the way recordException was implemented.
Also, this adds an addAll method to the Attributes.Builder to support efficient addition of attributes from existing attributes.

### DIFF
--- a/exporters/otlp/src/test/java/io/opentelemetry/exporters/otlp/MetricAdapterTest.java
+++ b/exporters/otlp/src/test/java/io/opentelemetry/exporters/otlp/MetricAdapterTest.java
@@ -52,7 +52,7 @@ class MetricAdapterTest {
     assertThat(MetricAdapter.toProtoLabels(Labels.of("k", "v")))
         .containsExactly(StringKeyValue.newBuilder().setKey("k").setValue("v").build());
     assertThat(MetricAdapter.toProtoLabels(Labels.of("k1", "v1", "k2", "v2")))
-        .containsExactlyInAnyOrder(
+        .containsExactly(
             StringKeyValue.newBuilder().setKey("k1").setValue("v1").build(),
             StringKeyValue.newBuilder().setKey("k2").setValue("v2").build());
   }
--- a/exporters/otlp/src/test/java/io/opentelemetry/exporters/otlp/MetricAdapterTest.java
+++ b/exporters/otlp/src/test/java/io/opentelemetry/exporters/otlp/MetricAdapterTest.java
@@ -52,7 +52,7 @@ class MetricAdapterTest {
     assertThat(MetricAdapter.toProtoLabels(Labels.of("k", "v")))
         .containsExactly(StringKeyValue.newBuilder().setKey("k").setValue("v").build());
     assertThat(MetricAdapter.toProtoLabels(Labels.of("k1", "v1", "k2", "v2")))
-        .containsExactly(
+        .containsExactlyInAnyOrder(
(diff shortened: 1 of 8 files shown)


exporters/otlp/src/test/java/io/opentelemetry/exporters/otlp/MetricAdapterTest.java
```diff
@@ -52,7 +52,7 @@ void toProtoLabels() {
     assertThat(MetricAdapter.toProtoLabels(Labels.of("k", "v")))
         .containsExactly(StringKeyValue.newBuilder().setKey("k").setValue("v").build());
     assertThat(MetricAdapter.toProtoLabels(Labels.of("k1", "v1", "k2", "v2")))
-        .containsExactly(
+        .containsExactlyInAnyOrder(
```
anuraaga, line 55: You're the metric expert :) But at a glance it seems like order is supposed to matter here?

sdk/tracing/src/test/java/io/opentelemetry/sdk/trace/SpanBuilderSdkTest.java
```diff
@@ -139,7 +139,7 @@ void truncateLinkAttributes() {
     RecordEventsReadableSpan span = (RecordEventsReadableSpan) spanBuilder.startSpan();
     try {
       assertThat(span.toSpanData().getLinks())
-          .containsExactly(
+          .containsExactlyInAnyOrder(
```
anuraaga, line 142: Not a big deal, but surprised by the change since looks like only one link :)

anuraaga (approved):

thisthat (approved):
Thank you for fixing this :shipit:

exporters/otlp/src/test/java/io/opentelemetry/exporters/otlp/MetricAdapterTest.java
jkwatson: "expert". actually, I think this can be reverted, now that I reverse the list after de-duping. I'll look

sdk/tracing/src/test/java/io/opentelemetry/sdk/trace/SpanBuilderSdkTest.java
jkwatson: isn't needed. this snuck in while I was hacking on stuff, and forgot to revert.

jkwatson: fixed, and thanks for the sharp eye!

PR #1658, opened 2020-09-17
Add an OpenTelemetry context mechanism.

anuraaga:
We have had an open issue, #575, to find an alternative to an external dependency for context propagation. As the spec states that for languages with no built in mechanism, like Go, or no obviously popular library, like JS's cls-hooked, we should create our own. So let's follow the spec and move on.

The API is heavily inspired by gRPC and Armeria. Much code / docs are borrowed too and I kept license headers in tact to reflect that.

- Context is immutable. However, `ContextStorage.rootContext()` is customizable to allow storing a mutable value at the very earliest point possible, effectively making context mutable if that is needed by a user. One reason someone may do this is to store their context in ours
- `attach` returns a try-with-resources `Scope` instead of having a `detach`
- Getting values is from the context, not from the key (personally find more natural)
- Wrappers for pretty much any callback-like type in JDK. If targeting JDK8 could add even more for the completable stuff
- Context storage can be customized using SPI with normal mechanisms, with less of the trickiness of mechanisms like gRPC
  - I'm thinking of adding https://github.com/line/armeria/pull/2723 to allow customizing at app startup without SPI too, but deferring for now (we're deferring our other stuff too I guess)

One key point is that because storage is customizable, this can be used fine with any other context mechanism, it is just the API OpenTelemetry happens to use. If a user prefers a different mechanism, it is not that difficult to delegate our API to them and users can keep on using what they're used to. We can provide extensions implementing using other context propagation libraries anytime.

Give it a first look to see if it generally looks ok, and I'll add tests, etc.

/cc @zoercai


extensions/context/src/main/java/io/opentelemetry/context/Context.java
```diff
@@ -0,0 +1,302 @@
...
+ *   <li>While Context objects are immutable they do not place such a restriction on the state they
+ *       store.
+ *   <li>Context is not intended for passing optional parameters to an API and developers should
+ *       take care to avoid excessive dependence on context when designing an API.
+ * </ul>
+ */
+public final class Context {
+
+  /** Returns the root {@link Context} which all other {@link Context} are derived from. */
+  public static Context root() {
```
tylerbenson, line 77: Nice work.

What is the motivation for exposing the root?  When is this useful?

extensions/context/src/main/java/io/opentelemetry/context/Context.java
```diff
@@ -0,0 +1,302 @@
...
+  /** Returns a new context with the given key value set. */
+  public <V1, V2> Context withValues(ContextKey<V1> k1, V1 v1, ContextKey<V2> k2, V2 v2) {
+    PersistentHashArrayMappedTrie.Node<ContextKey<?>, Object> newEntries =
+        PersistentHashArrayMappedTrie.put(entries, k1, v1);
+    newEntries = PersistentHashArrayMappedTrie.put(newEntries, k2, v2);
+    return new Context(newEntries);
+  }
+
+  /** Returns a new context with the given key value set. */
+  public <V1, V2, V3> Context withValues(
```
jkwatson, line 161: Do we need all the overrides at this point? I'd like to figure out how minimal we can make this API for starters.

jkwatson:
This looks like a great start! Thanks for pulling this together so quickly!

extensions/context/src/main/java/io/opentelemetry/context/Context.java
carlosalberto: Same comment. I'd be happier if we don't expose it *now*, in order to keep things simple.

extensions/context/src/main/java/io/opentelemetry/context/Context.java
```diff
@@ -0,0 +1,302 @@
...
+   * <pre>{@code
+   * Context prevCtx = Context.current();
+   * try (Scope ignored = ctx.attach()) {
+   *   assert Context.current() == ctx;
+   *   ...
+   * }
+   * assert Context.current() == prevCtx;
+   * }</pre>
+   */
+  public Scope attach() {
```
carlosalberto, line 220: +1. Definitely something I've always wanted to have in grpc.Context

carlosalberto:
Hey @anuraaga, thanks for prototyping this out.

> As the spec states that for languages with no built in mechanism, like Go, or no obviously popular library, like JS's cls-hooked, we should create our own. So let's follow the spec and move on.

Historically this was added to cover up Ruby, as there was no existent implementation, as opposed to Java, where there are more than a few bullet-proof implementations (although it got slightly hard, slightly political, which one to settle down with).

> However, ContextStorage.rootContext() is customizable

As commented, is there any specific case for which we truly need this? 

> Wrappers for pretty much any callback-like type in JDK

We originally had similar wrappers in `Tracer` itself (inherited by OC), but decided to move them to its own tracer-propagation package under `extensions/`. We may need to do it later on, or at least have them in a separate class/package in the same artifact (so the core `Context` API stays simple).

> Give it a first look to see if it generally looks ok, and I'll add tests, etc.

I'd advise to not go further till this idea overall has been accepted, or otherwise you may lose cycles if eventually this change is not accepted 😢 

IMHO this is a good way to go, even though we 'duplicate' the `grpc.Context` logic - otherwise fine. But I'm wondering about the typical case that @bogdandrutu used to address for these kind of changes, regarding the double wrapping.

Will include this PR in tomorrow's agenda so we can discuss it properly.

**Disclaimer**: I had a hyper simple prototype long ago, where you had a simple `Context` API, implementing almost nothing, simply falling back to `grpc.Context` or any other implementation, such as Reactor's, which would have been easier to maintain. Of course, the advantage of this very PR is that we provide a full, out-of-the-box implementation.

anuraaga (commented):
> Historically this was added to cover up Ruby, as there was no existent implementation, as opposed to Java, where there are more than a few bullet-proof implementations (although it got slightly hard, slightly political, which one to settle down with).

Quoting the spec again "Languages are expected to use the single, widely used Context implementation if one exists for them. In the cases where an extremely clear, pre-existing option is not available, OpenTelemetry MUST provide its own Context implementation."

There's obviously no extremely clear option, hence the tension - perhaps Ruby was involved in the initial thought process, but it seems like a nice spec statement to avoid the exact political issues you mention. As this issue has been sitting too long, I encourage everyone just taking a fresh take, and the simplest solution is to just do what the spec says, provide our Context implementation.

> As commented, is there any specific case for which we truly need this?

Yeah this is important for interop with other contexts - if the context is immutable, the only way to interop with it from another system is when it's attached, as that's when it's finally in a threadlocal. It may never be attached, or is often too late when it is. So because this has a goal of interop, allowing interception at the moment a "context is created" is very important. @trask mentioned he has a use case for this as well.

> We originally had similar wrappers in Tracer itself (inherited by OC), but decided to move them to its own tracer-propagation package under extensions/. 

I can't find the wrappers in `trace_propagators` extension - maybe we're referring to something different here. These are just the way to propagate context using JDK's mechanisms like `Runnable`. Context propagation is the most important feature of context, I'd say they're first-class APIs. Having them in the same class or not is just about usability of the API - I'd say having them on `Context` directly is the easiest for users, alternative would probably be something using static methods.

FWIW, since this is OpenTelemetry `Context`, I'm interesting in not keeping it too simple but making it easy to use for OTel users. This means merging `TracingContextUtils` in as well in the future - currently it's a bunch of static methods just living in the middle of nowhere. They, *ahem*, lack context making them a hard API to use in my experience in the instrumentation.

> Will include this PR in tomorrow's agenda so we can discuss it properly.

Thanks - hoping to get this issue done with so we can launch!

extensions/context/src/main/java/io/opentelemetry/context/Context.java
tylerbenson: Any reason to call it root?  Perhaps `empty` makes more sense?

carlosalberto:
>  the simplest solution is to just do what the spec says

Sadly no. The simplest solution would be to use the solid, robust, well understood and independent `grpc.Context` implementation (Reactor's context would have been another possibility, but the fact it does not exist as a standalone artifact doesn't make it illegible). 

Overall, I just want to make clear that staying with `grpc.Context` is not the end of the world (in the *bad* scenario of not agreeing on taking this path).

> Context propagation is the most important feature of context, I'd say they're first-class APIs. 

I prefer having them in its own class or package, just as `grpc.Context` does. I think having _all_ of this functionality in the same artifact is enough to make it a first-class component. But that's a detail we can discuss later as well.

> This means merging TracingContextUtils in as well in the future 

This may be hard. We *really* don't want to have any bit of trace(or baggage or any other component) in the Context layer, as discussed in previous prototypes, like one that @trask had  (although I agree that we need to make `TraceContextUtil` cleaner).

Thanks again for the prototype and hope we can come to a decision after discussing it with @bogdandrutu tomorrow!

bogdandrutu:
Just to make sure it is documented, and I can tell everyone "I told you" :)

This will make manual instrumentation a nightmare when it comes to deal with third party libraries instrumented with other Context, not that gRPC was solving that, but we were aiming for unification, now we are encouraging more and more contexts.

extensions/context/src/grpcInOtelTest/java/io/opentelemetry/context/GrpcInOtelTest.java
```diff
@@ -0,0 +1,134 @@
...
+              otelValue.set(Context.current().getValue(ANIMAL));
+            };
+        otherThread.submit(runnable).get();
+        assertThat(grpcValue).hasValue(null);
+        assertThat(otelValue).hasValue(null);
+
+        otherThread.submit(io.grpc.Context.current().wrap(runnable)).get();
+        assertThat(grpcValue).hasValue("japan");
+
+        // Since gRPC context is inside the OTel context, propagating gRPC context does not
```
anuraaga, line 97: When using storage override with immutable contexts, this is where there can be gaps - you pick one context propagation mechanism and go with it.

However, since `Context` is an interface now it would be fairly simple to allow providers to implement that too to close this gap without any breaking changes. We do still need a default context, `DefaultContext` here for users of libraries with no context propagation mechanism.

extensions/context/src/main/java/io/opentelemetry/context/ContextStorage.java
```diff
@@ -0,0 +1,94 @@
...
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.opentelemetry.context;
+
+/**
+ * The storage for storing and retrieving the current {@link Context}.
+ *
+ * <p>If you want to implement your own storage or add some hooks when a {@link Context} is attached
```
anuraaga, line 37: To make this easier without SPI we can add a static method that, when called close to app startup, does similar configuration (probably similar to what `setResource` type of things will end up at in the future). Left it out for now so this currently isn't completely trivial.

extensions/context/src/main/java/io/opentelemetry/context/PersistentHashArrayMappedTrie.java
```diff
@@ -0,0 +1,309 @@
...
+/**
+ * A persistent (copy-on-write) hash tree/trie. Collisions are handled linearly. Delete is not
+ * supported, but replacement is. The implementation favors simplicity and low memory allocation
+ * during insertion. Although the asymptotics are good, it is optimized for small sizes like less
+ * than 20; "unbelievably large" would be 100.
+ *
+ * <p>Inspired by popcnt-based compression seen in Ideal Hash Trees, Phil Bagwell (2000). The rest
+ * of the implementation is ignorant of/ignores the paper.
+ */
+final class PersistentHashArrayMappedTrie {
```
anuraaga, line 45: I'm not convinced this is actually faster than a normal CoW array for our propagation cases. Will benchmark and examine that in the future.

FroMage:
Hi,

Following up on the discussion about MicroProfile-Context-Propagation, I see that with this PR, MP-CP will be able to propagate your context to users of MP-CP, which is great.

On the other hand, the lambda wrappers and your `ContextExecutorService` will not allow you to execute tasks that will have more than your context.

So, for instance, if we use open-telemetry in a Java project that relies on many contexts, as long as we're using MP-CP we will be able to propagate the CDI, JAX-RS, Transaction and OpenTelemetry contexts, which is great. But if OpenTelemetry runs a user task, they will get random contexts for CDI, JAX-RS and Transaction, because you only propagate the OpenTelemetry context.

Now, I don't know enough about open-telemetry to guess if that will matter, because it depends what kinds of tasks you are wrapping. If they're internal tasks to open-telemetry, that don't call user code, that'll be fine. If on the other hand, they do call user code, the integrator will have to go through hoops to pre-wrap the user code with MP-CP to propagate every context (excluding the OpenTelemetry one, I assume, to not have to restore it twice) before passing that lambda to OpenTelemetry.

I do think it's a pity to see OpenTelemetry re-implement parts of MP-CP (the wrapping and executor service) but only for a single context, when a standard solution exists and could be used (from what I see in this PR), but this is probably due to your JDK7 requirement.

kenfinnigan:
I found out on another issue that JDK 8 is now in play, so maybe it's possible to re-evaluate using MP-CP as a solution?

jkwatson:
> I found out on another issue that JDK 8 is now in play, so maybe it's possible to re-evaluate using MP-CP as a solution?

I think the original motivation still precludes this. MP-CP might be super awesome and the new hotness, but it's definitely not "the single, widely used Context implementation" for java at this point.

kenfinnigan:
And that's totally fine, wanted to raise the possibility but I understand the reasoning

extensions/context/src/test/java/io/opentelemetry/context/ContextTest.java
```diff
@@ -0,0 +1,448 @@
...
+
+  @Test
+  void newThreadStartsWithRoot() throws Exception {
+    Context context = Context.current().withValues(ANIMAL, "cat");
+    try (Scope ignored = context.attach()) {
+      assertThat(Context.current().getValue(ANIMAL)).isEqualTo("cat");
+      AtomicReference<Context> current = new AtomicReference<>();
+      Thread thread =
+          new Thread(
+              () -> {
```
jkwatson, line 109: very minor: you can get rid of the brackets here

extensions/context/src/main/java/io/opentelemetry/context/Context.java
```diff
@@ -0,0 +1,214 @@
...
+   */
+  @Nullable
+  <V> V getValue(ContextKey<V> key);
+
+  /**
+   * Returns a new context with the given key value set.
+   *
+   * <pre>{@code
+   * Context withCredential = Context.current().withValue(CRED_KEY, cred);
+   * withCredential.run(new Runnable() {
```
jkwatson, line 92: I think this needs to change to be "wrap"

extensions/context/src/grpcInOtelTest/java/io/grpc/override/ContextStorageOverride.java
```diff
@@ -0,0 +1,71 @@
...
+
+package io.grpc.override;
+
+import io.grpc.Context;
+import io.opentelemetry.context.ContextKey;
+import io.opentelemetry.context.Scope;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class ContextStorageOverride extends Context.Storage {
```
jkwatson, line 25: I might be missing something, but I don't see this being used anywhere.

jkwatson: Neve rmind. I read the grpc docs and see how this works. Let's add an annotation so that IDEA doesn't suggest deleting it. :)

jkwatson:
This looks great. I don't think this module should be in the extensions directory, though, as it will be a top-level construct needed by the API. I think it should just be at the top level.

kenfinnigan:
If this were approved, would it replace the dependency on `opentelemetry-context-prop` in `opentelemetry-api`?

If so, should this work move from the extension to `opentelemetry-context-prop` to replace what's there?

Or is the intention to support both?

jkwatson:
> If this were approved, would it replace the dependency on `opentelemetry-context-prop` in `opentelemetry-api`?
> 
> If so, should this work move from the extension to `opentelemetry-context-prop` to replace what's there?
> 
> Or is the intention to support both?

The plan would be to have `opentelemetry-context-prop` depend on this, rather than on grpc-context. `opentelemetry-api` would continue to depend on the context propagation package, I believe.

kenfinnigan:
Thanks @jkwatson, from taking another look at the code that makes sense.

Overuse of the same term for different pieces was throwing me.

anuraaga (commented):
> I don't think this module should be in the extensions directory, though, as it will be a top-level construct needed by the API. 

Yeah figured we hadn't quite decided a final home for this yet so had it in extensions for incubation, but went ahead and moved to top-level.

While not explored at all, I see a possibility of putting the interfaces in API and `DefaultContext` and friends in SDK even :P

> Overuse of the same term for different pieces was throwing me.

I'm always confused too :) The `ContextUtils` will go away, and I think we can merge the propagators into API. @jkwatson do you know if they need to be a separate artifact?

extensions/context/src/main/java/io/opentelemetry/context/Context.java
```diff
@@ -0,0 +1,302 @@
...
+ * <ul>
+ *   <li>Every {@code attach()} must be followed by a {@code Scope#close()}. Breaking these rules
+ *       may lead to memory leaks.
+ *   <li>While Context objects are immutable they do not place such a restriction on the state they
+ *       store.
+ *   <li>Context is not intended for passing optional parameters to an API and developers should
+ *       take care to avoid excessive dependence on context when designing an API.
+ * </ul>
+ */
+public final class Context {
```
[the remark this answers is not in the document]
jkwatson: thoughts on this, @anuraaga ?

extensions/context/src/grpcInOtelTest/java/io/grpc/override/ContextStorageOverride.java
jkwatson: interesting. IDEA was marking it as unused for me. :hmmm:

jkwatson (approved):
Thanks!

context/src/main/java/io/opentelemetry/context/ContextStorage.java
```diff
@@ -0,0 +1,101 @@
...
+   * Returns the current {@link DefaultContext}. If no {@link DefaultContext} has been attached yet,
+   * this will be the {@linkplain Context#root()} root context}.
+   */
+  Context current();
+
+  /**
+   * Returns a {@link ContextKey} for the given name. This is only useful when integrating with a
+   * separate context propagation mechanism, where
+   */
+  default <T> ContextKey<T> contextKey(String name) {
```
anuraaga, line 98: Added this extension point to allow customizing the implementation of a key to match a storage.

context/src/otelAsGrpcTest/java/io/opentelemetry/context/OtelAsGrpcTest.java
```diff
@@ -0,0 +1,132 @@
...
+              grpcValue.set(COUNTRY.get());
+              otelValue.set(Context.current().getValue(ANIMAL));
+            };
+        otherThread.submit(runnable).get();
+        assertThat(grpcValue).hasValue(null);
+        assertThat(otelValue).hasValue(null);
+
+        otherThread.submit(Context.current().wrap(runnable)).get();
+
+        assertThat(grpcValue).hasValue("japan");
```
anuraaga, line 125: Works

anuraaga (commented):
I added one more extension point and an example of directly using the grpc context from the otel api when that interop is enabled. This allows both gRPC or OTel API to be used for context propagation without chance of data loss.

extensions/context/src/main/java/io/opentelemetry/context/Context.java
jkwatson: @anuraaga ?

anuraaga: @jkwatson I think this is addressed, maybe you found an older commit?

extensions/context/src/main/java/io/opentelemetry/context/Context.java
anuraaga: Sorry I missed this comment - @tylerbenson I'm not sure about `empty` because a context override could change the default context to include something (needed if someone needs to stuff mutable state in there). I could rename to `defaultContext()` (not `default()` because Java), but any thoughts on which sounds best?

anuraaga:
I added a couple of examples of context interop with Brave. Since Brave is a tracing library, the interop will require more, such as a way to map spans to/from OTel and Brave, or at least just allow parent/child between them. But figured it could provide some hints /cc @marcingrzejszczak

jkwatson (approved):
Let's do it!

carlosalberto:
Hey, I will do a late review first thing tomorrow. Sorry for the delay.

context/src/main/java/io/opentelemetry/context/ContextStorageProvider.java
```diff
@@ -0,0 +1,42 @@
...
+ * io.grpc.context.Context}</a>, or <a
+ * href="https://download.eclipse.org/microprofile/microprofile-context-propagation-1.0.2/apidocs/org/eclipse/microprofile/context/ThreadContext.html">{@code
+ * org.eclipse.microprofile.context.ThreadContext}</a>
+ *
+ * <p>if you are already using one of those systems in your application. Then you would not have to
+ * use methods like {@link Context#wrap(Executor)} and can use your current system instead.
+ */
+public interface ContextStorageProvider {
+
+  /** Returns the {@link ContextStorage} to use to store {@link DefaultContext}. */
```
carlosalberto, line 40: Is `DefaultContext` really the expected one? Shouldn't be `Context` instead?

carlosalberto:
Overall LGTM - my only problem is that I think the Brave/Grpc.Context parts should exist in an extension package, as they are NOT really needed (nor will be) by the majority of users. We already have exporters in a separate package, so logically speaking we should do the same ;)

PS - Let's also give @bogdandrutu some time to review.

context/src/main/java/io/opentelemetry/context/ContextStorageProvider.java
anuraaga: Doh thought I caught all these, thanks!

anuraaga:
@carlosalberto The gRPC / brave are both examples and also verification of some of the interop parts, so I put them in tests, they're not in the module. I think it's worth keeping there to be able to verify interop pieces while iterating, if they also make sense to present to users (which they probably will at some point), then I'll move them to the extensions.

Yesterday Bogdan mentioned he's generally ok with this and working with this approach - @bogdandrutu can you confirm you're ok with it? Let me know if you want to review this more :)

carlosalberto:
> The gRPC / brave are both examples and also verification of some of the interop parts, so I put them in tests, they're not in the module.

Oh, my bad. Verified this is the case. LGTM.

carlosalberto (approved):

anuraaga:
@bogdandrutu You asked for one day and got three - going to go ahead and merge but let me know if anything comes up.

PR #1659, opened 2020-09-17
Update env var names

### LINKED ISSUE #1657: Update the exporter configuration environment variable names to match the spec.
See the specs here:  https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-environment-variables.md

thisthat:
Resolves #1657

### DIFF
--- a/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcMetricExporter.java
+++ b/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcMetricExporter.java
@@ -295,7 +295,7 @@ public final class OtlpGrpcMetricExporter implements MetricExporter {
 
       Long value = getLongProperty(KEY_TIMEOUT, configMap);
       if (value == null) {
-        value = getLongProperty(KEY_TIMEOUT.replace("metric", ""), configMap);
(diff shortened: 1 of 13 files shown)


exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcMetricExporter.java
```diff
@@ -282,24 +292,43 @@ private Builder() {}
     protected Builder fromConfigMap(
         Map<String, String> configMap, NamingConvention namingConvention) {
       configMap = namingConvention.normalize(configMap);
-      Long value = getLongProperty(KEY_METRIC_TIMEOUT, configMap);
+
+      Long value = getLongProperty(KEY_TIMEOUT, configMap);
+      if (value == null) {
+        value = getLongProperty(KEY_TIMEOUT.replace("metric", ""), configMap);
```
jkwatson, line 298: Rather than the clever replacing, I think I'd personally prefer just having another constant defined with the non-metric value. To me, that makes the code a little easier to read. Also, the common properties could be put somewhere common and shared between the metric/span exporter config.

thisthat: Done in 34c8c25

jkwatson (approved):
Thanks!

PR #1665, opened 2020-09-18
Switch to java 8

jkwatson:
And Android level 24

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -14,8 +14,8 @@ dependencies {
 
     annotationProcessor libraries.auto_value
 
-    signature "org.codehaus.mojo.signature:java17:1.0@signature"
-    signature "net.sf.androidscents.signature:android-api-level-14:4.0_r4@signature"
+    signature "org.codehaus.mojo.signature:java18:1.0@signature"
+    signature "net.sf.androidscents.signature:android-api-level-24:7.0_r2@signature"
 }
 
 javadoc {
--- a/build.gradle
+++ b/build.gradle
@@ -15,8 +15,6 @@ subprojects {
     group = "io.opentelemetry"
     version = "0.9.0-SNAPSHOT" // CURRENT_OPEN_TELEMETRY_VERSION
 
-
-
     plugins.withId("maven-publish") {
         // Always include the artifactory/bintray plugins to do the deployment.
         pluginManager.apply "com.jfrog.artifactory"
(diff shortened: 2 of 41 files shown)


build.gradle
```diff
@@ -148,8 +146,8 @@ configure(opentelemetryProjects) {
         mavenLocal()
     }
 
-    sourceCompatibility = 1.7
-    targetCompatibility = 1.7
+    sourceCompatibility = 1.8
```
iNikem, line 149: Quick note: https://docs.gradle.org/current/userguide/building_java_projects.html#sec:java_cross_compilation

jkwatson: we can only use the new flag if we're compiling with java9+, though, correct?

bogdandrutu (approved):

build.gradle
jkwatson: I tried switching to the new flag and it build locally with java 14, but failed to build on CI.

trask (approved):

api/build.gradle
```diff
@@ -14,8 +14,8 @@ dependencies {
 
     annotationProcessor libraries.auto_value
 
-    signature "org.codehaus.mojo.signature:java17:1.0@signature"
-    signature "net.sf.androidscents.signature:android-api-level-14:4.0_r4@signature"
+    signature "org.codehaus.mojo.signature:java18:1.0@signature"
+    signature "net.sf.androidscents.signature:android-api-level-24:7.0_r2@signature"
```
anuraaga, line 18: Okhttp uses API 21 signature and once before we had discussed following what they do to get Java 8 support in a way that follows the most used Android library. Is it possible to stick with that?

https://github.com/square/okhttp/blob/master/build.gradle#L153

Though if we just avoid `CompletableFuture` I think any signature there should be fine too.

build.gradle
anuraaga: Yeah we'll want to switch to using Java 11 to build on CI and drop support for building on older to take advantage of the new features. Even 15 would be ok there are usually new compilation features (often javadoc improvements) and still works on older. We would need to change the CI to have a build and separate test jdk configured with Gradle to continue testing with Java 8.

For another issue of course :)

anuraaga (approved):

api/build.gradle
jkwatson: I don't have a strong opinion on this one. If `CompletableFuture` is the only thing we lose between 21 and 24 then I'm cool with that.

anuraaga (commented):
We can revisit the Android API level later let's go ahead and merge first!

PR #1668, opened 2020-09-21
Change built-in SpanProcessor implementations to ignore more than one shutdown call

### LINKED ISSUE #1636: Change built-in SpanProcessor implementations to ignore more than one shutdown call.
See here:  https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#shutdown

malafeev:
fix #1636

### DIFF
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/MultiSpanProcessor.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/MultiSpanProcessor.java
@@ -29,6 +29,7 @@ public final class MultiSpanProcessor implements SpanProcessor {
   private final List<SpanProcessor> spanProcessorsStart;
   private final List<SpanProcessor> spanProcessorsEnd;
   private final List<SpanProcessor> spanProcessorsAll;
+  private volatile boolean isShutdown = false;
 
   /**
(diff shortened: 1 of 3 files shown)


sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/MultiSpanProcessor.java
```diff
@@ -68,11 +69,17 @@ public boolean isEndRequired() {
 
   @Override
   public CompletableResultCode shutdown() {
-    List<CompletableResultCode> results = new ArrayList<>(spanProcessorsAll.size());
-    for (SpanProcessor spanProcessor : spanProcessorsAll) {
-      results.add(spanProcessor.shutdown());
+    synchronized (this) {
```
iNikem, line 72: `synchronized (this)` is always a bad idea, as you open yourself for deadlock if somebody external decides to synchronise on this object instance. In this particular case I think `AtomicBoolean` is better than `volatile+lock`

malafeev: updated

iNikem (approved):

jkwatson (approved):

PR #1671, opened 2020-09-22
support aws eks attributes

wangzlei:
**Description:** Add AWS EKS attrubutes support

Can get AWS EKS attributes in Span.


sdk_extensions/aws_v1_support/build.gradle
```diff
@@ -16,6 +16,11 @@ dependencies {
             libraries.guava,
             platform(boms.guava)
 
+    implementation 'com.fasterxml.jackson.core:jackson-databind:2.11.0',
+            libraries.guava,
+            platform(boms.guava)
+    implementation 'io.kubernetes:client-java:9.0.0'
```
anuraaga, line 22: This is a pretty heavyweight dependency, just one of the transitive dependencies alone is 8MB and I think total it will possibly double the agent size.

Can you try making direct calls to the API? FWIU, token is just a file to read so authentication is very easy for k8s API so it's mostly just parsing, which doesn't seem too bad.

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EksHelper.java
```diff
@@ -0,0 +1,127 @@
...
+      V1ConfigMapList configMapList =
+          k8sApi.listNamespacedConfigMap(
+              NAMESPACE_KUBE_SYSTEM, null, null, null, null, null, null, null, null, null);
+
+      if (configMapList == null) {
+        logger.log(Level.INFO, "Failed to get configmap of namespace %s.", NAMESPACE_KUBE_SYSTEM);
+        return false;
+      }
+
+      for (V1ConfigMap item : configMapList.getItems()) {
```
anuraaga, line 76: Should be able to do one call to get a config map instead of listing

https://docs.openshift.com/container-platform/3.7/rest_api/api/v1.ConfigMap.html#Get-api-v1-namespaces-namespace-configmaps-name

jkwatson:
This seems like a very specific, cloud-vendor-restricted Resource. Does it belong in the main repo, or should this be hosted by AWS? Should this kind of integration be hosted here? Has this been addressed in the specs yet, like other vendor-specific items have been?

anuraaga:
@jkwatson My understanding is cloud-vendor specific is OK for the SDKs, while tracing-vendor specific isn't. You can see this one isn't tracing-vendor specific since it's filling in a normal OTel semantic attribute, the k8s cluster name. From what I can tell, languages that have cloud resource detectors (JS, Go) include them in SDK or SDK contrib.

jkwatson:
> @jkwatson My understanding is cloud-vendor specific is OK for the SDKs, while tracing-vendor specific isn't. You can see this one isn't tracing-vendor specific since it's filling in a normal OTel semantic attribute, the k8s cluster name. From what I can tell, languages that have cloud resource detectors (JS, Go) include them in SDK or SDK contrib.

Heh. Cloud-vendor vs. Tracing-vendor is an interesting slippery slope. Doesn't AWS technically have tracing as well?

anuraaga:
@jkwatson Yeah, my team :) But the thinking is anything that's a semantic attribute already is valuable to any tracing vendor, and we want to make sure AWS customers get a good experience from any of the tracing solutions, not just X-Ray. Other attributes which are X-Ray specific we would include in our distros, not here.

sdk_extensions/aws_v1_support/build.gradle
wangzlei: use HttpURLConnection instead of k8s client

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/EksHelper.java
wangzlei: yeah, seems k8s client sdk does not support accurate query, will change to call restful api directly

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/RestfulClient.java
```diff
@@ -0,0 +1,187 @@
...
+import java.util.Collection;
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.net.ssl.HttpsURLConnection;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLSocketFactory;
+import javax.net.ssl.TrustManagerFactory;
+
+public class RestfulClient {
```
anuraaga, line 47: Nothing particularly restful, let's call it just `HttpClient` or `JdkHttpClient`

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/RestfulClient.java
```diff
@@ -0,0 +1,187 @@
...
+      logger.log(Level.WARNING, "Build SslSocketFactory for K8s restful client exception: %s", e);
+    }
+    return null;
+  }
+
+  String getK8sCredHeader(String tokenFilePath) {
+    try {
+      File file = new File(tokenFilePath);
+      String content = Files.asCharSource(file, Charsets.UTF_8).read();
+      return String.format("Bearer %s", content);
```
anuraaga, line 140:
```suggestion
      return "Bearer " + content;
```


Avoid `String.format` since it's slow performance and rarely that much more readable

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/RestfulClient.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+  }
+
+  private static SSLSocketFactory buildSslSocketFactory(String keystorePath) {
+    try {
+      KeyStore keyStore = getK8sKeystore(keystorePath);
+
+      String tmfAlgorithm = TrustManagerFactory.getDefaultAlgorithm();
+      TrustManagerFactory tmf = TrustManagerFactory.getInstance(tmfAlgorithm);
+      tmf.init(keyStore);
```
anuraaga, line 124: Let's avoid doing anything if `keyStore` is null due to an error. So I guess in the caller,

```java
KeyStore keyStore = getK8sKeystore(keystorePath);
if (keyStore != null) { ((HttpsURLConnection))connection.setSslSocketFactory(buildSslSocketFactory(keyStore));
```

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/RestfulClient.java
```diff
@@ -0,0 +1,187 @@
...
+      } else {
+        connection = (HttpsURLConnection) new URL(urlStr).openConnection();
+        ((HttpsURLConnection) connection).setSSLSocketFactory(buildSslSocketFactory(keyStorePath));
+      }
+
+      connection.setRequestMethod(httpMethod);
+      connection.setConnectTimeout(TIMEOUT_MILLIS);
+      connection.setReadTimeout(TIMEOUT_MILLIS);
+
+      if (requestPropertyMap != null) {
```
anuraaga, line 72: This is an internal class, let's assume it's not null

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/RestfulClient.java
```diff
@@ -0,0 +1,187 @@
...
+
+  String fetchString(
+      String httpMethod,
+      String urlStr,
+      Map<String, String> requestPropertyMap,
+      String keyStorePath) {
+    final HttpURLConnection connection;
+
+    try {
+      if (Strings.isNullOrEmpty(keyStorePath)) {
```
anuraaga, line 61: This is an internal class, let's keep the logic more direct - we can assume not null means not empty, these are passed by our code, not users

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/RestfulClient.java
```diff
@@ -0,0 +1,187 @@
...
+        connection = (HttpURLConnection) new URL(urlStr).openConnection();
+      } else {
+        connection = (HttpsURLConnection) new URL(urlStr).openConnection();
```
anuraaga, line 64:
```suggestion
        connection = new URL(urlStr).openConnection();
```

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/RestfulClient.java
```diff
@@ -0,0 +1,187 @@
...
+  String fetchString(
+      String httpMethod,
+      String urlStr,
+      Map<String, String> requestPropertyMap,
+      String keyStorePath) {
+    final HttpURLConnection connection;
+
+    try {
+      if (Strings.isNullOrEmpty(keyStorePath)) {
+        connection = (HttpURLConnection) new URL(urlStr).openConnection();
```
anuraaga, line 62:
```suggestion
        connection = new URL(urlStr).openConnection();
```

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/RestfulClient.java
```diff
@@ -0,0 +1,187 @@
...
+      } else {
+        connection = (HttpsURLConnection) new URL(urlStr).openConnection();
+        ((HttpsURLConnection) connection).setSSLSocketFactory(buildSslSocketFactory(keyStorePath));
```
anuraaga, line 65: Check the urlStr is `https` before

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/RestfulClient.java
```diff
@@ -0,0 +1,187 @@
...
+    return "";
+  }
+
+  private static KeyStore getK8sKeystore(String keyStorePath) {
+    InputStream certificateFile = null;
+    try {
+      KeyStore trustStore = null;
+      File caFile = new File(keyStorePath);
+
+      if (caFile.exists()) {
```
anuraaga, line 153: Can clean up a bit I think

```java
File caFile = new File(caFilePath); // Oh by the way, let's rename this variable from keystorePath :)
if (!caFile.exists()) {
  return null;
}

try (InputStream certStream = new FileInputStream(caFile)) {
  KeyStore trustStore = KeyStore.getInstance()
  ...

  return trustStore;
} catch ()
// No need for finally :)
```

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/RestfulClient.java
```diff
@@ -0,0 +1,187 @@
...
+        trustStore = KeyStore.getInstance(KeyStore.getDefaultType());
+        trustStore.load(null, null);
+        CertificateFactory certificateFactory = CertificateFactory.getInstance("X.509");
+
+        certificateFile = new FileInputStream(caFile);
+        Collection<? extends Certificate> certificates =
+            certificateFactory.generateCertificates(certificateFile);
+
+        if (certificates.isEmpty()) {
+          throw new IllegalArgumentException("K8s cert file contained no certificates.");
```
anuraaga, line 163:
```suggestion
          return null;
```

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/RestfulClient.java
```diff
@@ -0,0 +1,187 @@
...
+      if (requestPropertyMap != null) {
+        for (Map.Entry<String, String> requestProperty : requestPropertyMap.entrySet()) {
+          connection.setRequestProperty(requestProperty.getKey(), requestProperty.getValue());
+        }
+      }
+
+      int responseCode = connection.getResponseCode();
+      if (responseCode != 200) {
+        logger.log(
+            Level.WARNING,
```
anuraaga, line 81: I switched the logging to `FINE` recently - since resource attributes are still a nice-to-have and not must, we should only log for debug logging and not as warnings.

sdk_extensions/aws_v1_support/src/test/java/io/opentelemetry/sdk/extensions/trace/aws/resource/RestfulClientTest.java
```diff
@@ -0,0 +1,64 @@
...
+import com.google.common.collect.ImmutableMap;
+import java.io.IOException;
+import java.util.Map;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+public class RestfulClientTest {
+
+  @ClassRule
+  public static WireMockClassRule server = new WireMockClassRule(wireMockConfig().dynamicPort());
```
anuraaga, line 40: Offline you mentioned trouble testing HTTPS. Yeah it looks like wiremock only supports Java keystore, but not a certificate file. Armeria's mock server supports certificate files well so we'll probably switch to it, for now can leave a TODO

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/RestfulClient.java
wangzlei: Change to JdkHttpClient

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/RestfulClient.java
wangzlei: ok, will change all of String.format to '+'

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/RestfulClient.java
wangzlei: fix in next commit

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/RestfulClient.java
wangzlei: check the doc, have to use this style:
connection = (HttpsURLConnection) new URL(urlStr).openConnection();
and
connection = (HttpURLConnection) new URL(urlStr).openConnection();

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/RestfulClient.java
wangzlei: fix in next commit

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/RestfulClient.java
wangzlei: ok

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/JdkHttpClient.java
```diff
@@ -0,0 +1,163 @@
...
+  String fetchString(
+      String httpMethod,
+      String urlStr,
+      Map<String, String> requestPropertyMap,
+      String keyStorePath) {
+    final HttpURLConnection connection;
+
+    try {
+      if (urlStr.startsWith("https")) {
+        connection = (HttpsURLConnection) new URL(urlStr).openConnection();
```
anuraaga, line 58: Ah sorry in that case let cast to `HttpUrlConnection` here - the variable is that type so it's weird using a different one when casting.

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/JdkHttpClient.java
```diff
@@ -0,0 +1,163 @@
...
+                + responseCode
+                + ") text "
+                + readResponseString(connection));
+        return "";
+      }
+
+      return readResponseString(connection).trim();
+
+    } catch (IOException e) {
+      logger.log(Level.WARNING, "JdkHttpClient fetch string failed: " + e);
```
anuraaga, line 91: Here also FINE, I think this is the block that's hit when the server doesn't exist at all (matters more when we switch `Ec2Resource` to this new class)

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/JdkHttpClient.java
```diff
@@ -0,0 +1,163 @@
...
+import java.util.Collection;
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.net.ssl.HttpsURLConnection;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLSocketFactory;
+import javax.net.ssl.TrustManagerFactory;
+
+public class JdkHttpClient {
```
anuraaga, line 43: Package private

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/JdkHttpClient.java
```diff
@@ -0,0 +1,163 @@
...
+      context.init(null, tmf.getTrustManagers(), null);
+      return context.getSocketFactory();
+
+    } catch (Exception e) {
+      logger.log(Level.WARNING, "Build SslSocketFactory for K8s restful client exception: " + e);
+    }
+    return null;
+  }
+
+  String getK8sCredHeader(String tokenFilePath) {
```
anuraaga, line 135: Move this into `EksResource` it's k8s specific

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/JdkHttpClient.java
```diff
@@ -0,0 +1,163 @@
...
+      File file = new File(tokenFilePath);
+      String content = Files.asCharSource(file, Charsets.UTF_8).read();
+      return "Bearer " + content;
+    } catch (IOException e) {
+      logger.log(Level.WARNING, "Unable to load K8s client token: " + e);
+    }
+    return "";
+  }
+
+  private static KeyStore getK8sKeystore(String keyStorePath) {
```
anuraaga, line 146: Rename to not be k8s since it's just about client TLS trusted cert.

`getKeystoreForTrustedCert(String certPath)` for example

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/JdkHttpClient.java
wangzlei: yeah, here can also be HttpURLConnection

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/JdkHttpClient.java
wangzlei: update code to generic getKeystoreForTrustedCert

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/JdkHttpClient.java
wangzlei: move method to `EksResource`

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/JdkHttpClient.java
```diff
@@ -0,0 +1,148 @@
...
+  private static KeyStore getKeystoreForTrustedCert(String certPath) {
+    try (FileInputStream fis = new FileInputStream(certPath)) {
+      KeyStore trustStore = KeyStore.getInstance(KeyStore.getDefaultType());
+      trustStore.load(null, null);
+      CertificateFactory certificateFactory = CertificateFactory.getInstance("X.509");
+
+      Collection<? extends Certificate> certificates = certificateFactory.generateCertificates(fis);
+
+      int i = 0;
+      for (Certificate certificate : certificates) {
```
anuraaga, line 138: Ah good catch - if we need index let's just use traditional for loop. `for(int i = 0; I < ...`

anuraaga (approved):
Just small nit, thanks!

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/JdkHttpClient.java
wangzlei: Collection does not support random access, have to convert it to Array. Here I also fell code looks ugly, if we have other way to name the cert in for loop, I would prefer not call Collection.toArray().

anuraaga: Ah was thinking collection has it as opposed to iterable. In that case no worries.

PR #1681, opened 2020-09-22
Fix example code for 0.8.0, non-snapshot dependency

jkwatson:


examples/build.gradle
```diff
@@ -14,7 +14,7 @@ subprojects {
     version = "0.1.0-SNAPSHOT"
 
     ext {
-        opentelemetryVersion = "0.8.0-SNAPSHOT"
+        opentelemetryVersion = "0.9.0-SNAPSHOT"
```
anuraaga, line 17: I think similar to how we don't update README until after a release, we probably should keep this in the burner until the release (and possibly change the version to just `0.8.0` in the meantime) so people aren't confused. What do you think?

jkwatson: Hmm. It'd be a change. We've been depending on snapshots up to now, I think to keep things moving forward. I'm not opposed to it, and it would make things change less often in the example code, for sure. 

I'm ambivalent on this one.

anuraaga: The examples will change frequently too as our API does, it's more just whether we want it to be an example of the next release, or the current release. Generally we would want examples for the current release. Presumably that's why we have a gradle wrapper here instead of just including in the normal build?

jkwatson: It's a fair point. I'm not 100% sure why we decided to have the examples based on snapshot builds in the first place. I'll update this to just point at 0.8.0 and make sure everything is working.

jkwatson: done

thisthat (approved):

anuraaga (approved):
Thanks!

PR #1682, opened 2020-09-22
Make the getKey method on AttributeKey non-nullable

### LINKED ISSUE #1678: The getKey() method in the AttributeKey interface should not be marked as Nullable
I believe this was originally done to help AutoValue do something internally. We should fix it, as it is incorrect and should never be null.

jkwatson:
This was done by tricking AutoValue with a separate field for the storage.

resolves #1678

### DIFF
--- a/api/src/main/java/io/opentelemetry/common/AttributeKeyImpl.java
+++ b/api/src/main/java/io/opentelemetry/common/AttributeKeyImpl.java
@@ -17,14 +17,23 @@
 package io.opentelemetry.common;
 
 import com.google.auto.value.AutoValue;
+import javax.annotation.Nullable;
 
 @SuppressWarnings("rawtypes")
 @AutoValue
 abstract class AttributeKeyImpl<T> implements AttributeKey<T> {
 
   static <T> AttributeKeyImpl<T> create(String key, AttributeType type) {
-    return new AutoValue_AttributeKeyImpl<>(key, type);
+    return new AutoValue_AttributeKeyImpl<>(type, key);
   }
+
+  @Override
+  public String getKey() {
+    return key();
+  }
+
+  @Nullable
+  abstract String key();
 
   //////////////////////////////////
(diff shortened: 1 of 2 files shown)


api/src/main/java/io/opentelemetry/common/AttributeKeyImpl.java
```diff
@@ -17,15 +17,24 @@
...
 @AutoValue
 abstract class AttributeKeyImpl<T> implements AttributeKey<T> {
 
   static <T> AttributeKeyImpl<T> create(String key, AttributeType type) {
-    return new AutoValue_AttributeKeyImpl<>(key, type);
+    return new AutoValue_AttributeKeyImpl<>(type, key);
   }
 
+  @Override
+  public String getKey() {
```
anuraaga, line 31: But can't `key()` return null?

jkwatson: It can, but I really want the public API to have the getKey() be non-nullable. We can't have the API blow up with null keys, but we really want the API to signal that it should never be null. I'm open to other suggestions to solve this one.

anuraaga: Got it - yeah that's a corner case I've ran into before too. I wasn't sure that was the intention but makes sense to me. If we ever add nullness checking (want to!) we'll need some "nullness casts" for these cases, just as a future food for thought.

anuraaga (approved):

api/src/main/java/io/opentelemetry/common/AttributeKeyImpl.java
jkwatson: Right now, we drop null keys, I believe, so this seems like a decent compromise. You can put null in there, but you'll never actually see a null-keyed attribute anywhere, since we throw them away.

PR #870, opened 2020-02-18
Rename MeterRegistry -> MeterProvider

thisthat:
This addresses #794 

Specification: [link](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-metrics-user.md#obtaining-a-meter)

PR #1683, opened 2020-09-22
Rename dangling references to MeterRegistry

breedx-nr:
This is a follow-up from #870 

After `MeterRegistry` was renamed to `MeterProvider`, some of the javadoc examples still pointed to the old name, which could be misleading/confusing for new users.

There was also one remaining variable reference.


anuraaga (approved):
Thanks for the cleanup!

jkwatson (approved):
Thanks!

PR #1685, opened 2020-09-23
[Draft] Batch observer

### LINKED ISSUE #1331: Implement the metric BatchObserver from the spec
See here:  https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#batch-observer

thisthat:
This PR fixes #1331 implementing the [Batch Observer](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#batch-observer).

Before adding tests, I would like to have feedback on the implementation of the API.

### DIFF
--- a/api/src/main/java/io/opentelemetry/metrics/AsynchronousInstrument.java
+++ b/api/src/main/java/io/opentelemetry/metrics/AsynchronousInstrument.java
@@ -22,7 +22,7 @@ import javax.annotation.concurrent.ThreadSafe;
  * @since 0.1.0
  */
 @ThreadSafe
-public interface AsynchronousInstrument<R extends Result> extends Instrument {
+public interface AsynchronousInstrument<R extends Result, T extends Number> extends Instrument {
   /**
    * A {@code Callback} for a {@code AsynchronousInstrument}.
    *
@@ -43,10 +43,12 @@ public interface AsynchronousInstrument<R extends Result> extends Instrument {
    */
   void setCallback(Callback<R> callback);
 
+  Observation observation(T observation);
+
   /** Builder class for {@link AsynchronousInstrument}. */
   interface Builder extends Instrument.Builder {
     @Override
-    AsynchronousInstrument<?> build();
+    AsynchronousInstrument<?, ?> build();
   }
 
   interface Result {}
@@ -59,5 +61,14 @@ public interface AsynchronousInstrument<R extends Result> extends Instrument {
   /** The result for the {@link Callback}. */
   interface DoubleResult extends Result {
     void observe(double value, Labels labels);
+  }
+
+  interface Observation {
+    ObservationType getType();
+  }
+
+  enum ObservationType {
+    LONG_OBSERVATION,
+    DOUBLE_OBSERVATION,
   }
 }
--- a/api/src/main/java/io/opentelemetry/metrics/AsynchronousInstrument.java
+++ b/api/src/main/java/io/opentelemetry/metrics/AsynchronousInstrument.java
@@ -22,7 +22,7 @@ import javax.annotation.concurrent.ThreadSafe;
  * @since 0.1.0
  */
 @ThreadSafe
-public interface AsynchronousInstrument<R extends Result, T extends Number> extends Instrument {
+public interface AsynchronousInstrument<R extends Result> extends Instrument {
   /**
    * A {@code Callback} for a {@code AsynchronousInstrument}.
    *
@@ -43,12 +43,10 @@ public interface AsynchronousInstrument<R extends Result, T extends Number> exte
    */
   void setCallback(Callback<R> callback);
 
-  Observation observation(T observation);
-
   /** Builder class for {@link AsynchronousInstrument}. */
   interface Builder extends Instrument.Builder {
     @Override
-    AsynchronousInstrument<?, ?> build();
+    AsynchronousInstrument<?> build();
   }
 
   interface Result {}
@@ -64,4 +62,12 @@ public interface AsynchronousInstrument<R extends Result, T extends Number> exte
   }
 
   interface Observation {}
+
+  interface LongObservation {
+    Observation observation(long observation);
+  }
+
+  interface DoubleObservation {
+    Observation observation(double observation);
+  }
 }
--- a/api/src/main/java/io/opentelemetry/metrics/AsynchronousInstrument.java
+++ b/api/src/main/java/io/opentelemetry/metrics/AsynchronousInstrument.java
@@ -63,12 +63,5 @@ public interface AsynchronousInstrument<R extends Result, T extends Number> exte
     void observe(double value, Labels labels);
   }
 
-  interface Observation {
-    ObservationType getType();
-  }
-
-  enum ObservationType {
-    LONG_OBSERVATION,
-    DOUBLE_OBSERVATION,
-  }
+  interface Observation {}
 }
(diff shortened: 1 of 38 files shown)


api/src/main/java/io/opentelemetry/metrics/AsynchronousInstrument.java
```diff
@@ -54,10 +54,12 @@
    */
   void setCallback(Callback<R> callback);
 
+  Observation observation(T observation);
```
jkwatson, line 57: One bummer about making this generic is that it means that every number will get boxed with an object wrapper, rather than just using the primitive. Perhaps for Observers this is less of an issue than with synchronous instruments, but I do want to make sure we're ok with the extra allocations this will result in.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractAsynchronousInstrument.java
```diff
@@ -88,6 +92,41 @@ LongResultSdk newResult(ActiveBatcher activeBatcher) {
...
+        return ObservationType.LONG_OBSERVATION;
+      }
+
+      @Override
+      public Aggregator getAggregator() {
+        return this.activeBatcher.getAggregator();
+      }
+
+      @Override
+      public Descriptor getDescription() {
```
jkwatson, line 120: nit: this method should be `getDescriptor()`

api/src/main/java/io/opentelemetry/metrics/AsynchronousInstrument.java
jkwatson: Also, this means that users could pass in nulls, which we would need to account for. Is it possible to just have 2 primitive-only operations? I think we could get rid of the `getType()` method then as well, maybe?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/BatchObserverSdk.java
```diff
@@ -0,0 +1,248 @@
...
+          (labels, observations) -> {
+            batcher.setLabels(labels);
+            for (Observation observation : observations) {
+              ObservationType type = observation.getType();
+              Aggregator aggregator;
+              Descriptor descriptor;
+              if (type == ObservationType.LONG_OBSERVATION) {
+                LongObservation longObservation = (LongObservation) observation;
+                aggregator = longObservation.getAggregator();
+                aggregator.recordLong(longObservation.getValue());
```
jkwatson, line 112: rather than requiring a 2-step process here, could we have the LongObservation just have a method like `recordLong` that returns the Aggregator?

api/src/main/java/io/opentelemetry/metrics/Meter.java
```diff
@@ -202,4 +47,14 @@
    * @since 0.1.0
    */
   BatchRecorder newBatchRecorder(String... keyValuePairs);
```
jkwatson, line 49: why wouldn't this go onto the RecorderInstruments interface? Likewise with the newBatchObserver and the ObserverInstruments interface?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractAsynchronousInstrument.java
```diff
@@ -88,6 +92,41 @@ LongResultSdk newResult(ActiveBatcher activeBatcher) {
       return new LongResultSdk(activeBatcher);
     }
 
+    @Override
+    public Observation observation(Long observation) {
+      return new LongObservationSdk(this.getActiveBatcher(), observation);
+    }
+
+    private static final class LongObservationSdk implements LongObservation {
```
jkwatson, line 100: It bugs me that we need both this, and the LongResult, since they seem so very close to being exactly the same thing, just with the execution deferred in one case and immediate in the other.

Would it be simpler to just have the batch recorder keep a handle to the set of (instrument, value) pairs, then just make all the actual recordings at one time, rather than having to make all the wrapper classes? That is, could the Observation implementation just hold on to that Instrument and value?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/BatchObserverSdk.java
```diff
@@ -0,0 +1,248 @@
...
+        Clock clock) {
+      this.resource = resource;
+      this.instrumentationLibraryInfo = instrumentationLibraryInfo;
+      this.clock = clock;
+      this.aggregatorFactory = aggregatorFactory;
+      this.reportList = new ArrayList<>();
+      startEpochNanos = clock.now();
+    }
+
+    public void setLabels(Labels labels) {
```
jkwatson, line 189: could be private

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/BatchObserverSdk.java
```diff
@@ -0,0 +1,248 @@
...
+
+    @Override
+    public Descriptor getDescriptor() {
+      return null;
+    }
+
+    @Override
+    public final void batch(Labels labelSet, Aggregator aggregator, boolean unmappedAggregator) {}
+
+    public void batch(Descriptor descriptor, Aggregator aggregator) {
```
jkwatson, line 206: this could be private

jkwatson:
Thanks for putting this together! I feel like this is more complex than it needs to be, with every observation requiring something like 3-4 wrapper objects, in addition to boxing of the actual value being observed. Could we build something much simpler to start with?

api/src/main/java/io/opentelemetry/metrics/Meter.java
thisthat: From the example in the spec ([link](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#batch-observer)), I thought that `newBatchRecorder` and `newBatchObserver` can only be created by the `Meter`

jkwatson: The Meter interface, as you've defined it, extends both of those interfaces, though...so it's on the Meter either way.

thisthat: `BatchObserver` extends `ObserverInstruments`. If we move `newBatchObserver` inside `ObserverInstruments`, a `BatchObserver` can generate new `BatchObserver`s. Is this what the metric sdk should provide?

thisthat:
Thank you for the valuable feedback @jkwatson! I've removed the boxing and reduce the number of new allocations.

api/src/main/java/io/opentelemetry/metrics/Meter.java
jkwatson: Ugh. I hadn't read that spec carefully enough to realize that you need to be able to create instruments with the batch observer, and not just associate them. I wonder if this spec has been implemented anywhere else but golang yet.

jkwatson: My opinion is that, since those methods just delegate to the MeterSdk anyway, that we should get rid of that from the spec, and have it be an optional thing for APIs to expose, rather than mandatory. I'll open up a spec issue for it.

jkwatson: https://github.com/open-telemetry/opentelemetry-specification/pull/1030

thisthat:
The PR is ready for another round of review :smile: 
In contrast to the GO implementation, this implementation allows asynchronous instruments created by both, meter and batch observer to report an observation. To allows only the instruments created by the batch observer to report an observation, we would need to create a new set of builders and instruments.

jkwatson:
I'm going to close this; the latest API spec does not have this batch functionality described; we can revisit when/if it does.

PR #1688, opened 2020-09-23
validate version from traceparent according to spec

### LINKED ISSUE #1674: traceparent  with invalid version is not rejected
Next w3c context propagation tests fails:
```
def test_traceparent_version_0xff(self):
		'''
		harness sends an invalid traceparent with version 255 (0xff)
		expects a valid traceparent from the output header, with a newly generated trace_id
		'''
		traceparent, tracestate = self.make_single_request_and_get_tracecontext([
			['traceparent', 'ff-12345678901234567890123456789012-1234567890123456-01'],
		])
		self.assertNotEqual(traceparent.trace_id.hex(), '12345678901234567890123456789012')

def test_traceparent_version_illegal_characters(self):
		'''
		harness sends an invalid traceparent with illegal characters in version
		expects a valid traceparent from the output header, with a newly generated trace_id
		'''
		traceparent, tracestate = self.make_single_request_and_get_tracecontext([
			['traceparent', '.0-12345678901234567890123456789012-1234567890123456-01'],
		])
		self.assertNotEqual(traceparent.trace_id.hex(), '12345678901234567890123456789012')
```

Got in https://github.com/open-telemetry/opentelemetry-java/pull/1653
Can be reproduced by running tests from mentioned PR.

spec: https://www.w3.org/TR/trace-context/#version

malafeev:
fix #1674 

spec: https://www.w3.org/TR/trace-context/#version


Oberon00:
This violates the spec. See point 4 in https://www.w3.org/TR/trace-context/#a-traceparent-is-received:

> If the version number is higher than supported by the tracer, the vendor uses the format defined in this specification (00) to parse trace-id and parent-id. The vendor will only parse the trace-flags values supported by this version of this specification and ignore all other values. If parsing fails, the vendor creates a new traceparent header and deletes the tracestate. Vendors will set all unparsed / unknown trace-flags to 0 on outgoing requests.

Oberon00 (changes requested):
This violates the W3C spec. See point 4 in https://www.w3.org/TR/trace-context/#a-traceparent-is-received:

> If the version number is higher than supported by the tracer, the vendor uses the format defined in this specification (00) to parse trace-id and parent-id. The vendor will only parse the trace-flags values supported by this version of this specification and ignore all other values. If parsing fails, the vendor creates a new traceparent header and deletes the tracestate. Vendors will set all unparsed / unknown trace-flags to 0 on outgoing requests.

malafeev:
@Oberon00 I added only validation that version is valid. I don't want to change parsing of traceparent based on version number in this PR.

Oberon00 (approved):
Sorry, that was entirely my fault. Instead of reading `isValid`, I just assumed it would check for `== "00"`

api/src/main/java/io/opentelemetry/trace/TraceVersion.java
```diff
@@ -0,0 +1,34 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.trace;
+
+public class TraceVersion {
```
Oberon00, line 19:
```suggestion
final class TraceVersion {
```

api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java
```diff
@@ -198,6 +199,11 @@ private static SpanContext extractContextFromTraceParent(String traceparent) {
     }
 
     try {
+      String version = traceparent.substring(0, 2);
```
Oberon00, line 202: Shouldn't we check whether there is a trailing `-` too? What does W3C say?

api/src/main/java/io/opentelemetry/trace/TraceVersion.java
malafeev: it cannot be package private because accessed outside of pakcage

Oberon00: You can move it to the right package though. Or just make it a member function of `HttpTraceContext`

jkwatson, line 19: I'm not sure we need a whole class just for one method. Could this just be a method on the HttpTraceContext itself?

api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java
malafeev: it's already checked above:
```
traceparent.charAt(TRACE_ID_OFFSET - 1) == TRACEPARENT_DELIMITER
```

api/src/main/java/io/opentelemetry/trace/TraceVersion.java
malafeev: because `BigendianEncoding` class is package private and I cannot call it.

malafeev: no, problem that I want to call `BigendianEncoding.isValidBase16String(...)` method, but class `BigendianEncoding` is package private in `io.opentelemetry.trace`.

but `HttpTraceContext` class in package `io.opentelemetry.trace.propagation`.

So I need to make TraceVersion public class in package `io.opentelemetry.trace`.
Or move class  `HttpTraceContext` to package `io.opentelemetry.trace`.

jkwatson: I think I'd prefer a 3rd option, which is to create a public utility class (called maybe something like "WireFormatUtils"?) that will provide these needed methods for any propagator that needs them. This could then delegate to the BigEndianEncoding implementation, as appropriate, or maybe we move methods in there, as appropriate.

anuraaga: It's just two bytes - I think I would just generate a `Set` of all possible versions as a private in `HttpTraceContext` (255 elements only) and use that for validation, seems like it could be faster anyways.

malafeev: renamed TraceVersion to WireFormatUtils with method

api/src/main/java/io/opentelemetry/trace/WireFormatUtils.java
```diff
@@ -0,0 +1,34 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.trace;
+
+public class WireFormatUtils {
```
anuraaga, line 19: This is a sort of strange public API - not clear what wire format or version this is referring to, and pretty narrow use case. I think this was added because of the visibility of `BigEndianEncoding` but how about my suggestion of just precomputing the possible valid values and comparing to avoid needing this?

malafeev: set of 255 strings is not tiny. how I can compute all 255 values to insert into set? I suppose I will need to call `BigEndianEncoding.byteToBase16String(...)` method which is in `BigEndianEncoding` class.

anuraaga: Well since this is on the slow path (static initializer), we don't quite need the performance of `BigEndianEncoding` and could just use `Long.toHexString`.

```
Set<String> VALID_VERSIONS = new HashSet<>();
for (int i = 0; i < 255; i++) {
  String version = Long.toHexString(i);
  if (version.length() < 2)
    version = '0' + version;
  VALID_VERSIONS.add(version);
}
```

It's not so bad I guess. We need to be careful with public API, for this I would probably copy rather than expose if we need to.

jkwatson: I think the suggestion is to write some quick code to generate all the possibilities, and then just put that into a static Set field. Is that what you were thinking, @anuraaga ?

malafeev: done

malafeev: I added initialization to `static` block, can move to method if it's better

api/src/main/java/io/opentelemetry/trace/TraceVersion.java
malafeev: removed

malafeev: done

api/src/main/java/io/opentelemetry/trace/WireFormatUtils.java
jkwatson: The other option would be to pre-generate the set and hard-code it into the code itself, but I think this is fine a few extra microseconds at startup shouldn't be an issue, and we can change it to a hard-coded set later if we want.

jkwatson (approved):

anuraaga (approved):
Thanks!

PR #1689, opened 2020-09-23
Convert the ValueObserver instruments to use the LastValue aggegration.

### LINKED ISSUE #1780: Update ValueObserver to have LastValue as default aggregation
ValueObserver should have LastValue as the default aggregation.

jkwatson:
See this spec update:  https://github.com/open-telemetry/opentelemetry-specification/pull/984

Resolves #1780

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/ViewRegistry.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/ViewRegistry.java
@@ -63,8 +63,8 @@ class ViewRegistry {
       case UP_DOWN_COUNTER:
         return Aggregations.sum();
       case VALUE_RECORDER:
-      case VALUE_OBSERVER:
         return Aggregations.minMaxSumCount();
+      case VALUE_OBSERVER:
       case SUM_OBSERVER:
       case UP_DOWN_SUM_OBSERVER:
         return Aggregations.lastValue();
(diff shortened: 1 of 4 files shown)


bogdandrutu (approved):
As mentioned in the Java maintainers meeting. This is partially fixing the issue because our current MetricData does not reflect the latest changes in OTLP and we should produce Gauge for this. Do we want to go ahead with this? That means we will have another breaking change soon when Gauge will be the correct result.

jkwatson:
I think this is the right change for the current released version of OTLP, yes? I think it's fine to address the OTLP changes as a separate issue, once it gets released.

bogdandrutu:
@jkwatson not really, https://github.com/open-telemetry/opentelemetry-java/blob/master/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/MetricAdapter.java#L117

This will be exported as non-monotonic updown sum I believe which is wrong.

jkwatson:
> @jkwatson not really, https://github.com/open-telemetry/opentelemetry-java/blob/master/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/MetricAdapter.java#L117
> 
> This will be exported as non-monotonic updown sum I believe which is wrong.

hmm. yeah, you're right. :(

anuraaga (approved):

jkwatson:
> > @jkwatson not really, https://github.com/open-telemetry/opentelemetry-java/blob/master/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/MetricAdapter.java#L117
> > This will be exported as non-monotonic updown sum I believe which is wrong.
> 
> hmm. yeah, you're right. :(

https://github.com/open-telemetry/opentelemetry-java/issues/1786 to track getting that fixed.

PR #1690, opened 2020-09-23
Avoid usage of getSpanWithoutDefault when possible.

bogdandrutu:
We check for `isValid` everywhere so this will not be a performance hit, explicitly because the DefaultSpan.getInvalid().getContext().isValid() will be memoized.


jkwatson:
It looks like there's only one usage left after this. Can we get rid of that one, as well, and get rid of the method altogether?

bogdandrutu:
@jkwatson you read my mind, but that one usage left is a bit different. I will followup on that soon.

jkwatson (approved):
Thanks!

anuraaga (approved):
Was looking forward to this, thanks!

PR #1692, opened 2020-09-23
Remove constant labels from instruments and MetricData

bogdandrutu:
This concept was removed from the API and OTLP


jkwatson (approved):
Somehow I missed that constant labels had been removed from the API spec. Thanks for the cleanup!

thisthat (approved):

anuraaga (approved):
Nice

PR #1696, opened 2020-09-24
Fill cloud.provider for AWS resources.

anuraaga:
Need to set in SDK or the span looks like it could be from any cloud.

### DIFF
--- a/sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/BeanstalkResource.java
+++ b/sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/BeanstalkResource.java
@@ -95,7 +95,8 @@ public class BeanstalkResource extends ResourceProvider {
       return Attributes.empty();
     }
 
-    attrBuilders.setAttribute(ResourceAttributes.CLOUD_PROVIDER, "aws");
+    attrBuilders.setAttribute(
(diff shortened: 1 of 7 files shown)


sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/BeanstalkResource.java
```diff
@@ -95,6 +95,8 @@ public Attributes getAttributes() {
       return Attributes.empty();
     }
 
+    attrBuilders.setAttribute(ResourceAttributes.CLOUD_PROVIDER, "aws");
```
thisthat, line 98: Should we use a constant for the name? I do not expect AWS to change its name soon :stuck_out_tongue: but it might saves us some headache if we'll change the capitalization.

thisthat (approved):
Thank you :) I have only a *non-bloccker* question

sdk_extensions/aws_v1_support/src/main/java/io/opentelemetry/sdk/extensions/trace/aws/resource/BeanstalkResource.java
anuraaga: Makes sense, thanks

arminru (approved):

jkwatson (approved):

PR #1698, opened 2020-09-24
Document precedence order is multiple context propagators are used

### LINKED ISSUE #1271: Document precedence order is multiple context propagators are used

malafeev:
fix #1271


Oberon00:
I think this is currently unspecified in the spec, as https://github.com/open-telemetry/opentelemetry-specification/issues/496 is still open. If so, I think we should leave this unspecified in Java too to be able to follow the spec later without breaking documented behavior.

jkwatson:
> I think this is currently unspecified in the spec, as [open-telemetry/opentelemetry-specification#496](https://github.com/open-telemetry/opentelemetry-specification/issues/496) is still open. If so, I think we should leave this unspecified in Java too to be able to follow the spec later without breaking documented behavior.

ping @malafeev

carlosalberto:
I'm tempted to state/document here that the order is undefined for the time being. That way at least we are not tied for now ;) Does that sound right? If YES, @malafeev, you can iterate on this PR ;)

malafeev:
updated

jkwatson (approved):

carlosalberto (approved):

PR #1701, opened 2020-09-24
Update for the more restricted span status codes.

### LINKED ISSUE #1700: Update allowed span status codes
See this recent spec PR:  https://github.com/open-telemetry/opentelemetry-specification/pull/966

jkwatson:
resolves #1700 

Note: I made the numeric codes match the proposed protobuf status codes.

### DIFF
--- a/api/src/main/java/io/opentelemetry/trace/Status.java
+++ b/api/src/main/java/io/opentelemetry/trace/Status.java
@@ -51,10 +51,10 @@ public final class Status {
     OK(0),
 
     /** The default status. */
-    UNSET(17),
+    UNSET(1),
 
     /** The operation contains an error. */
-    ERROR(18);
+    ERROR(2);
 
     private final int value;
 
--- a/api/src/main/java/io/opentelemetry/trace/Status.java
+++ b/api/src/main/java/io/opentelemetry/trace/Status.java
@@ -17,10 +17,9 @@
 package io.opentelemetry.trace;
 
 import io.opentelemetry.internal.Utils;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.TreeMap;
 import javax.annotation.Nullable;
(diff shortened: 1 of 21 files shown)


api/src/main/java/io/opentelemetry/trace/Status.java
```diff
@@ -44,163 +43,18 @@
...
-    /**
-     * The service is currently unavailable. This is a most likely a transient condition and may be
-     * corrected by retrying with a backoff.
-     *
-     * <p>See litmus test above for deciding between FAILED_PRECONDITION, ABORTED, and UNAVAILABLE.
-     *
-     * @since 0.1.0
-     */
-    UNAVAILABLE(14),
+    OK(0),
```
anuraaga, line 51: Do we need the numbers anymore?

jkwatson: I'm honestly not sure. I noticed that in the Jaeger mapping, we directly just copy that code over as a status code. Is that correct? No idea, either. I'd prefer to get rid of the numeric codes, and the accessor for them, but we need to decide what to do with the existing mappings.

extensions/trace_utils/src/main/java/io/opentelemetry/extensions/trace/CurrentSpanUtils.java
```diff
@@ -121,7 +121,7 @@ public V call() throws Exception {
 
   private static void setErrorStatus(Span span, Throwable t) {
     span.setStatus(
-        Status.UNKNOWN.withDescription(
+        Status.ERROR.withDescription(
```
iNikem, line 124: I realised that I still an open question about this use case: https://github.com/open-telemetry/opentelemetry-specification/issues/1008

jkwatson: I'd be happy to revisit this when the issue is clarified. Meanwhile, it does seem like approximately preserving the existing behavior is a good thing for now. And, as I commented on that issue, I think the intent with `recordException` is that it's an error, most commonly.

api/src/main/java/io/opentelemetry/trace/Status.java
anuraaga: Got it - we can change this latter anyways.

anuraaga (approved):

api/src/main/java/io/opentelemetry/trace/Status.java
```diff
@@ -44,163 +43,18 @@
...
 
-    /**
-     * Unrecoverable data loss or corruption.
-     *
-     * @since 0.1.0
-     */
-    DATA_LOSS(15),
+    /** The default status. */
+    UNSET(17),
```
bogdandrutu, line 54: How did we come with this numbers? See https://github.com/open-telemetry/opentelemetry-proto/pull/224/files and maybe we should follow the same numbers so translation to otlp is just direct mapping.

jkwatson: I made them up. I was planning on fixing once the proto change was merged.

jkwatson: I'll update this PR to match what is in the proto PR, now that it looks like there aren't any more changes happening.

anuraaga: I think the numbers are now matching the deprecated code, but we should update to 0, 1, 2 here I believe.

jkwatson: yes. will do...that was my intent. :)

extensions/trace_utils/src/main/java/io/opentelemetry/extensions/trace/CurrentSpanUtils.java
jkwatson: When that gets merged, I'll create an issue to address it here. Meanwhile, keeping the existing behavior seems like the right decision for this PR.

api/src/main/java/io/opentelemetry/trace/Status.java
jkwatson: done

iNikem (approved):

PR #1706, opened 2020-09-25
Make setting null-valued attributes a no-op, and document that their behavior is undefined.

### LINKED ISSUE #1703: Update `null` value handling for span and resource attributes
https://github.com/open-telemetry/opentelemetry-specification/pull/992 changed `null` handling for span and resource attribute values from "delete previous value, otherwise ignore" to "undefined behavior". Handling of `null` within arrays stays unchanged.
We should document this in the API accordingly and also add a `@Nonnull` annotation.

I would go for just making it a no-op if `null` is passed nevertheless. WDYT?

Please also update the https://github.com/open-telemetry/opentelemetry-specification/blob/master/spec-compliance-matrix.md once this is done.

jkwatson:
resolves #1703

### DIFF
--- a/api/src/main/java/io/opentelemetry/common/Attributes.java
+++ b/api/src/main/java/io/opentelemetry/common/Attributes.java
@@ -232,7 +232,7 @@ public abstract class Attributes extends ImmutableKeyValuePairs<AttributeKey, Ob
 
     /** Sets a {@link AttributeKey} with associated value into this. */
     public <T> Builder setAttribute(AttributeKey<T> key, T value) {
-      if (value == null || key == null || key.getKey() == null || key.getKey().length() == 0) {
+      if (key == null || key.getKey() == null || key.getKey().length() == 0 || value == null ) {
         return this;
       }
       data.add(key);
--- a/api/src/main/java/io/opentelemetry/common/Attributes.java
+++ b/api/src/main/java/io/opentelemetry/common/Attributes.java
@@ -29,7 +29,6 @@ import com.google.auto.value.AutoValue;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
-import javax.annotation.Nonnull;
 import javax.annotation.concurrent.Immutable;
 
 /**
@@ -83,7 +82,7 @@ public abstract class Attributes extends ImmutableKeyValuePairs<AttributeKey, Ob
   }
 
   /** Returns a {@link Attributes} instance with a single key-value pair. */
-  public static <T> Attributes of(AttributeKey<T> key, @Nonnull T value) {
+  public static <T> Attributes of(AttributeKey<T> key, T value) {
     return sortAndFilterToAttributes(key, value);
   }
 
@@ -92,7 +91,7 @@ public abstract class Attributes extends ImmutableKeyValuePairs<AttributeKey, Ob
    * preserved. Duplicate keys will be removed.
    */
   public static <T, U> Attributes of(
-      AttributeKey<T> key1, @Nonnull T value1, AttributeKey<U> key2, @Nonnull U value2) {
+      AttributeKey<T> key1, T value1, AttributeKey<U> key2, U value2) {
     return sortAndFilterToAttributes(key1, value1, key2, value2);
   }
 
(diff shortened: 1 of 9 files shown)


api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -79,7 +83,7 @@ public static Attributes empty() {
   }
 
   /** Returns a {@link Attributes} instance with a single key-value pair. */
-  public static <T> Attributes of(AttributeKey<T> key, T value) {
+  public static <T> Attributes of(AttributeKey<T> key, @Nonnull T value) {
```
anuraaga, line 86: I don't think we use `@Nonnull`, we default to non-null. We're missing package-info.java to set ParametersNonNullByDefault but that's a better fix I think

anuraaga (approved):

api/src/main/java/io/opentelemetry/common/Attributes.java
jkwatson: I agree. As I was writing this, I was also wondering why we didn't put the same annotation on the key. :)

Having it set for the package as a whole is *much* better solution. I'll make it so.

api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -164,13 +172,7 @@ public static Builder newBuilder() {
...
   public static Builder newBuilder(ReadableAttributes attributes) {
     final Builder builder = new Builder();
-    attributes.forEach(
-        new AttributeConsumer() {
-          @Override
-          public <T> void consume(AttributeKey<T> key, T value) {
-            builder.setAttribute(key, value);
-          }
-        });
+    attributes.forEach(builder::setAttribute);
```
iNikem, line 175: 👍

api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -199,24 +201,7 @@ public Attributes build() {
...
-            // delete key and value
-            itr.remove();
-            itr.next();
-            itr.remove();
-          } else {
-            // skip the value part
-            itr.next();
-          }
-        }
+      if (value == null || key == null || key.getKey() == null || key.getKey().length() == 0) {
```
thisthat, line 204: Nit: for simmetry with the other checks
```suggestion
      if (key == null || key.getKey() == null || key.getKey().length() == 0 || value == null ) {
```

api/src/test/java/io/opentelemetry/common/AttributesTest.java
```diff
@@ -253,26 +253,4 @@ void toBuilder() {
...
-    attributes.setAttribute("arrayDouble", new Double[] {1.0});
-    attributes.setAttribute("arrayBool", new Boolean[] {true});
-    assertThat(attributes.build().size()).isEqualTo(9);
-    attributes.setAttribute(stringKey("attrValue"), null);
-    attributes.setAttribute("string", (String) null);
-    attributes.setAttribute("arrayString", (String[]) null);
-    attributes.setAttribute("arrayLong", (Long[]) null);
-    attributes.setAttribute("arrayDouble", (Double[]) null);
-    attributes.setAttribute("arrayBool", (Boolean[]) null);
-    assertThat(attributes.build().size()).isEqualTo(3);
```
thisthat, line 276: should we check that we implement a no-op behavior instead of leaving it undefined?

jkwatson: I went back and forth on this. I'm not really sure which we should do, since we're explicitly saying that behavior is undefined. :)

iNikem: If undefined, don't test for it. Some may consider tests as a kind of documentation.

jkwatson: that was my thinking as well.

thisthat: I would argue that the specification says "undefined behavior", i.e. each implementation of the spec can do whatever they prefer, but the SDK has a well-defined behavior. IMHO we should verify that we implement correctly the no-op behavior in every point that touches `null` attributes.

api/src/main/java/io/opentelemetry/common/Attributes.java
jkwatson: done

api/src/test/java/io/opentelemetry/common/AttributesTest.java
jkwatson: I've added a test that validates the current behavior, with a clear message that it's not required behavior and would be fine to be changed.

Oberon00: As a C++ dev, I'd say that would be "implementation-defined" behavior, not undefined behavior 😀

PR #1707, opened 2020-09-25
Change the Sampler interface to allow a SamplingResult to update the TraceState

### LINKED ISSUE #1705: Update SamplingResult to include the Tracestate
See changes here: https://github.com/open-telemetry/opentelemetry-specification/pull/988

This also means that the resulting Tracestate should be used when creating the SpanContext for a new span in the SpanBuilderSdk implementation.

jkwatson:
Resolves #1705

### DIFF
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Sampler.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Sampler.java
@@ -12,6 +12,7 @@ import io.opentelemetry.trace.Span;
 import io.opentelemetry.trace.Span.Kind;
 import io.opentelemetry.trace.SpanContext;
 import io.opentelemetry.trace.TraceId;
+import io.opentelemetry.trace.TraceState;
 import java.util.List;
 import javax.annotation.concurrent.ThreadSafe;
 
@@ -77,5 +78,14 @@ public interface Sampler {
      *     {@link Decision#RECORD_AND_SAMPLE}.
      */
     Attributes getAttributes();
+
+    /**
+     * Return the {@link TraceState} that should be associated with the resulting Span. This may be
+     * the same {@link TraceState} that was provided originally, or an updated one.
+     *
+     * <p>Note: if a null is returned, the TraceState will be reset to {@link
+     * TraceState#getDefault()}.
+     */
+    TraceState getTraceState();
   }
 }
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Sampler.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Sampler.java
@@ -84,8 +84,10 @@ public interface Sampler {
      * Return an optionally-updated {@link TraceState}, based on the parent TraceState. This may
      * return the same {@link TraceState} that was provided originally, or an updated one.
      *
-     * @param parentTraceState The TraceState from any parent span. Might be an empty TraceState, if
-     *     there is no parent.
+     * @param parentTraceState The TraceState from the parent span. Might be an empty TraceState, if
+     *     there is no parent. This will be the same TraceState that was passed in via the {@link
+     *     SpanContext} parameter on the {@link #shouldSample(SpanContext, String, String, Kind,
+     *     ReadableAttributes, List)} call.
      */
     default TraceState getUpdatedTraceState(TraceState parentTraceState) {
       return parentTraceState;
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Sampler.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Sampler.java
@@ -14,6 +14,7 @@ import io.opentelemetry.trace.SpanContext;
 import io.opentelemetry.trace.TraceId;
 import io.opentelemetry.trace.TraceState;
 import java.util.List;
+import java.util.function.Function;
 import javax.annotation.concurrent.ThreadSafe;
 
 /** Sampler is used to make decisions on {@link Span} sampling. */
@@ -80,12 +81,12 @@ public interface Sampler {
     Attributes getAttributes();
 
     /**
-     * Return the {@link TraceState} that should be associated with the resulting Span. This may be
-     * the same {@link TraceState} that was provided originally, or an updated one.
-     *
-     * <p>Note: if a null is returned, the TraceState will be reset to {@link
-     * TraceState#getDefault()}.
+     * Return a function that may update the incoming parent {@link TraceState}. This may return the
+     * same {@link TraceState} that was provided originally, or an updated one.
      */
-    TraceState getTraceState();
+    @SuppressWarnings("NoFunctionalReturnType")
+    default Function<TraceState, TraceState> traceStateUpdate() {
+      return Function.identity();
+    }
   }
 }
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Sampler.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Sampler.java
@@ -14,12 +14,12 @@ import io.opentelemetry.trace.SpanContext;
 import io.opentelemetry.trace.TraceId;
 import io.opentelemetry.trace.TraceState;
 import java.util.List;
-import java.util.function.Function;
 import javax.annotation.concurrent.ThreadSafe;
 
 /** Sampler is used to make decisions on {@link Span} sampling. */
 @ThreadSafe
 public interface Sampler {
+
   /**
    * Called during {@link Span} creation to make a sampling samplingResult.
    *
@@ -81,12 +81,14 @@ public interface Sampler {
     Attributes getAttributes();
 
     /**
-     * Return a function that may update the incoming parent {@link TraceState}. This may return the
-     * same {@link TraceState} that was provided originally, or an updated one.
+     * Return an optionally-updated {@link TraceState}, based on the parent TraceState. This may
+     * return the same {@link TraceState} that was provided originally, or an updated one.
+     *
+     * @param parentTraceState The TraceState from any parent span. Might be an empty TraceState, if
+     *     there is no parent.
      */
-    @SuppressWarnings("NoFunctionalReturnType")
-    default Function<TraceState, TraceState> traceStateUpdate() {
-      return Function.identity();
+    default TraceState getUpdatedTraceState(TraceState parentTraceState) {
+      return parentTraceState;
     }
   }
 }
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
@@ -19,6 +19,7 @@ import io.opentelemetry.trace.Span;
 import io.opentelemetry.trace.Span.Kind;
 import io.opentelemetry.trace.SpanContext;
 import io.opentelemetry.trace.TraceId;
+import io.opentelemetry.trace.TraceState;
 import java.util.List;
 import java.util.Objects;
 import javax.annotation.concurrent.Immutable;
@@ -38,13 +39,6 @@ public final class Samplers {
   // Visible for tests.
   static final AttributeKey<Double> SAMPLING_PROBABILITY = doubleKey("sampling.probability");
 
-  private static final SamplingResult EMPTY_RECORDED_AND_SAMPLED_SAMPLING_RESULT =
-      SamplingResultImpl.createWithoutAttributes(Decision.RECORD_AND_SAMPLE);
-  private static final SamplingResult EMPTY_NOT_SAMPLED_OR_RECORDED_SAMPLING_RESULT =
-      SamplingResultImpl.createWithoutAttributes(Decision.DROP);
-  private static final SamplingResult EMPTY_RECORDED_SAMPLING_RESULT =
-      SamplingResultImpl.createWithoutAttributes(Decision.RECORD_ONLY);
-
   // No instance of this class.
   private Samplers() {}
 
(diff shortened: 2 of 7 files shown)


sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Sampler.java
```diff
@@ -88,5 +89,14 @@ SamplingResult shouldSample(
      *     {@link Decision#RECORD_AND_SAMPLE}.
      */
     Attributes getAttributes();
+
+    /**
+     * Return the {@link TraceState} that should be associated with the resulting Span. This may be
+     * the same {@link TraceState} that was provided originally, or an updated one.
+     *
+     * <p>Note: if a null is returned, the TraceState will be reset to {@link
```
anuraaga, line 97:
```suggestion
     * <p>Note: if {@code null} is returned, the TraceState will be reset to {@link
```

Or I might remove this, this is not a nullable method so we don't want to indicate null is allowed. We just happen to have a default to prevent crashing in case that happens.

anuraaga (approved):

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
```diff
@@ -96,19 +92,20 @@ public static SamplingResult samplingResult(Decision decision, Attributes attrib
...
    * @param decision The decision made on the span.
+   * @param traceState The {@link TraceState} to use.
    * @return A {@link SamplingResult} with empty attributes and the provided {@code decision}.
    */
-  public static SamplingResult emptySamplingResult(Decision decision) {
+  public static SamplingResult emptySamplingResult(Decision decision, TraceState traceState) {
     switch (decision) {
       case RECORD_AND_SAMPLE:
-        return EMPTY_RECORDED_AND_SAMPLED_SAMPLING_RESULT;
+        return SamplingResultImpl.createWithoutAttributes(Decision.RECORD_AND_SAMPLE, traceState);
```
Oberon00, line 104: If we use null (or some static final instance of TraceState) to indicate "use parent tracestate", we could keep the constant objects, conserving allocations.

bogdandrutu (commented):
@jkwatson I see what you mentioned offline about an extra allocation all the time. Can we maybe find a way to avoid that? Some ideas, passing the traceState as an argument to the shouldSample? Maybe we need a "MutableTraceState" to do that? Not sure but we should investigate some options.

jkwatson:
> @jkwatson I see what you mentioned offline about an extra allocation all the time. Can we maybe find a way to avoid that? Some ideas, passing the traceState as an argument to the shouldSample? Maybe we need a "MutableTraceState" to do that? Not sure but we should investigate some options.

My thought was to alter the SamplingResult to include a boolean which would signal if the Sampler was providing an updated TraceState. Then, the simpler samplers can just have that set to always be `false` and pass in a static null as the TraceState. Then, we can preserve the existing static result values and get rid of the need for the allocation in the standard Sampler implementations.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Sampler.java
iNikem: Indeed, should we require non-null return value?

jkwatson: The spec explicitly specifies the behavior for the `null` return, but I wouldn't have a problem making this non-nullable. It's easy enough for samplers to explicitly return `TraceState.getInvalid()` if that's what they want.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
jkwatson: yeah, but the spec describes the behavior for null, which is to invalidate the tracestate. :(

Oberon00: OK, that was unintentional. But still, we could do `private static final TraceState USE_PARENT = TraceState.builder.build()`  and then check if the object identity of the contained tracestate `== USE_PARENT`.

Oberon00: I think the spec doesn't say anything about a `null` tracestate here, only an empty one: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#shouldsample

> * A `Tracestate` that will be associated with the `Span` through the new
>  `SpanContext`.
>  Note: If the sampler returns an empty `Tracestate` here, the `Tracestate` will be cleared,
>  so samplers should normally return the passed-in `Tracestate` if they do not intend
>  to change it.

jkwatson: You're right. I read that wrong. But, I don't think using `null` as a signal to use the parent is a great API. I'd rather have an explicit way for Samplers to signal that, rather than an implicit null.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Sampler.java
jkwatson: Sorry..my comment above is wrong. I mis-read "empty" as "null" in the spec.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
Oberon00: Agreed, `null` isn't a great API here. So maybe the sentinel object I mentioned above?

jkwatson: I'm not sure yet. My main concern is making sure that Sampler authors have a clear obvious way to do it. At the moment, the Sampler interface forces everyone to think about whether to modify the TraceState or not, and I think the most common use-case will be to not modify the TraceState, but just use the existing parent's. 

@Oberon00 do you think it's worth opening up a spec issue to make it easier to do the "right thing" in our SDK, or do you think the current spec is flexible enough to allow default-behavior to be implemented (using a sentinel value, or explicit boolean attribute)?

Oberon00: I no one will complain that you violate the current spec if you make "use parent" the default somehow, but an editorial spec change / recommendation would make sense I think.

jkwatson: I decided to start with an issue:  https://github.com/open-telemetry/opentelemetry-specification/issues/1031

jkwatson:
I came up with (I think) a better option: have the SamplingResult return an optional `Function<TraceState, TraceState>` that will allow updating the TraceState, per spec. I added a default implementation that returns `Function.identity()` so that existing Samplers need not be updated, and so that the most common thing is easy.

@bogdandrutu @anuraaga @Oberon00 What do you think?

anuraaga:
Sorry finally taking a detailed look at this. @jkwatson, can you point me at the allocation that went away from changing to a `Function`? Can't find it in https://github.com/open-telemetry/opentelemetry-java/pull/1707/commits/33ac2b472b0d5f966af47ccd6ada2c66d073f366

Oberon00:
@jkwatson I think using a callback for this is a bit overkill. I would still prefer the sentinel object. We don't even need to expose it to implementers, it would just be the default:

So like this:
```java
class SamplingResultImpl {
    // ...
    private static final TraceState UNCHANGED_TRACESTATE = new TraceState();

    static SamplingResult create(Decision decision, Attributes attributes) {
      return new AutoValue_Samplers_SamplingResultImpl(decision, attributes, UNCHANGED_TRACESTATE);
    }

    static SamplingResult create(Decision decision, Attributes attributes, TraceState traceState) {
      return new AutoValue_Samplers_SamplingResultImpl(decision, attributes, traceState);
    }

    @Override
    public abstract Decision getDecision();

    protected abstract TraceState getTraceState();

    @Override // Needs a change in the base interface too
    public TraceState getUpdatedTraceState(TraceState original) {
        return getTraceState() == UNCHANGED_TRACESTATE ? origininal: getTraceState();
    }
```

jkwatson:
> Sorry finally taking a detailed look at this. @jkwatson, can you point me at the allocation that went away from changing to a `Function`? Can't find it in [33ac2b4](https://github.com/open-telemetry/opentelemetry-java/commit/33ac2b472b0d5f966af47ccd6ada2c66d073f366)

Before the latest change, we had to return a new instance of SamplingResult with every single call to the sampler. Now, we can return constant SamplingResult instances for all the default samplers.

jkwatson:
> @jkwatson I think using a callback for this is a bit overkill. I would still prefer the sentinel object. We don't even need to expose it to implementers, it would just be the default:
> 
> So like this:
> 
> ```java
> class SamplingResultImpl {
>     // ...
>     private static final TraceState UNCHANGED_TRACESTATE = new TraceState();
> 
>     static SamplingResult create(Decision decision, Attributes attributes) {
>       return new AutoValue_Samplers_SamplingResultImpl(decision, attributes, UNCHANGED_TRACESTATE);
>     }
> 
>     static SamplingResult create(Decision decision, Attributes attributes, TraceState traceState) {
>       return new AutoValue_Samplers_SamplingResultImpl(decision, attributes, traceState);
>     }
> 
>     protected abstract TraceState getTraceState();
> 
>     @Override // Needs a change in the base interface too
>     public TraceState getUpdatedTraceState(TraceState original) {
>         return getTraceState() == UNCHANGED_TRACESTATE ? origininal: getTraceState();
>     }
> ```

I'm really not a fan of the proposed "sentinel object" pattern you're suggesting. It's not particularly idiomatic java, and feels like a hack to work around a poorly designed API that requires you to return an instance of *something*, even if you don't want to. With a function in the response, the API is clearly purpose designed to support both use-cases: not changing the TraceState, or changing it, as you desire.

Oberon00:
The sentinel object is not part of the API though. It is hidden away in SamplingResultImpl. The key point is the `TraceState getUpdatedTraceState(TraceState original)` signature that gets passed the original tracestate.

jkwatson:
> The sentinel object is not part of the API though. It is hidden away in SamplingResultImpl. The key point is the `TraceState getUpdatedTraceState(TraceState original)` signature that gets passed the original tracestate. Actually, since TraceState is not nullable, we could use `null` just as well.

Fair enough. It still feels hacky to me; Function::identity seems more natural, rather than your alternative. We could still have your method be implemented via a default that always returns the original TraceState, rather than having the sentinel in there at all, though, correct?

jkwatson:
@Oberon00 I updated it to use your method signature, but have it implemented via a default method on the interface, rather than requiring the sentinel implementation. This feels like a simpler implementation that either of us had originally proposed.

Oberon00:
@jkwatson Yes, that would make sense.

But if you think your approach is cleaner, I can't really object to it -- I think it's mostly a matter of taste. The additional function object might be a tiny bit overhead (not sure if that identity function you return from an interface implementation can be inlined), but probably not much.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Sampler.java
```diff
@@ -77,5 +79,16 @@ SamplingResult shouldSample(
      *     {@link Decision#RECORD_AND_SAMPLE}.
      */
     Attributes getAttributes();
+
+    /**
+     * Return an optionally-updated {@link TraceState}, based on the parent TraceState. This may
+     * return the same {@link TraceState} that was provided originally, or an updated one.
+     *
+     * @param parentTraceState The TraceState from any parent span. Might be an empty TraceState, if
```
Oberon00, line 87: We could document here that this will always be the same instance as the one passed to ShouldSample before.

Oberon00 (approved):

Oberon00:
> @Oberon00 I updated it to use your method signature, but have it implemented via a default method on the interface, rather than requiring the sentinel implementation. This feels like a simpler implementation that either of us had originally proposed.

My proposal was more complicated because it supported the optional TraceState paramter to SamplingResultImpl.create.  But if we don't want to support that (for now) that's totally fine.

jkwatson:
> > @Oberon00 I updated it to use your method signature, but have it implemented via a default method on the interface, rather than requiring the sentinel implementation. This feels like a simpler implementation that either of us had originally proposed.
> 
> My proposal was more complicated because it supported the optional TraceState parameter to SamplingResultImpl.create. But if we don't want to support that (for now) that's totally fine.

True. I'm honestly not sure what any real-life usage of this is going to end up looking like. We can always add enhanced constructors if users who need to modify TraceState don't think it's viable to provide the implementation simply via a method that maps from one TraceState to another.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Sampler.java
jkwatson: good idea. done!

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Sampler.java
```diff
@@ -77,5 +79,18 @@ SamplingResult shouldSample(
...
+    /**
+     * Return an optionally-updated {@link TraceState}, based on the parent TraceState. This may
+     * return the same {@link TraceState} that was provided originally, or an updated one.
+     *
+     * @param parentTraceState The TraceState from the parent span. Might be an empty TraceState, if
+     *     there is no parent. This will be the same TraceState that was passed in via the {@link
+     *     SpanContext} parameter on the {@link #shouldSample(SpanContext, String, String, Kind,
+     *     ReadableAttributes, List)} call.
+     */
+    default TraceState getUpdatedTraceState(TraceState parentTraceState) {
```
anuraaga, line 92: Thanks sorry for missing the allocation issue, got it. TBH this API still feels weird to me - having to call the sampler, and then the caller has to correctly call this method to get the trace state. 

I'm reading the discussion here

https://github.com/open-telemetry/opentelemetry-java/pull/1707#discussion_r496019521

and I'm personally not sure this functional style is that much better than `null` indicating a non-update, in a caller

```
span.traceState = result.getUpdatedTraceState() != null ? result.getUpdatedTraceState() : parentTraceState;
```

seems more direct? It's true that there's no updated trace state, so use the non-updated one.

Otherwise, another approach that is similar to this one but seems a bit nicer is having this sort of method on the `Sampler`, accepting the decision with the parent trace state. "update trace state" is a business logic operation, it's nice if that lives in the Sampler which has the logic, a result feels like it should be mostly constant.

Finally, maybe we can just live with an extra allocation if it seems like a normal `getTraceState()` seems like a more intuitive API. Sampling result is very short lived, it may even be optimized away who knows.

Food for thought, I don't really have a strong preference for any, but a small preference for eating the allocation.

anuraaga (approved):

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Sampler.java
jkwatson: The other reason to avoid "eating the allocation" is that we really want to make it super easy for "normal" samplers to do the right thing (that is, not modify the TraceState). I really dislike using "null" as a meaningful business result, but if y'all prefer that as the API, then I can get on board. What if we had it return `Optional<TraceState>` so the "normal" samplers can just have a constant `Optional.empty()`?

jkwatson: @anuraaga and I discussed offline and decided to stick with what we have right now; if someone actually uses this API and doesn't like how it works, we can definitely change it, since it's SDK-only functionality, and we can be a little more loose with making changes.

PR #1712, opened 2020-09-26
Merge SpanContext into Span and make DefaultSpan private

### LINKED ISSUE #1135: Why is there a DefaultSpan?

### LINKED ISSUE #1704: Update DefaultSpan to match the new spec
This might just be a rename to PropagatedSpan, but please check the changes to see if there are any behavioral changes as well:  https://github.com/open-telemetry/opentelemetry-specification/pull/994

anuraaga:
While https://github.com/open-telemetry/opentelemetry-specification/issues/999 won't be merged, I think we can say that we have `SpanContext`, it just happens to be merged into `Span` as an implementation detail.

The API seems to be vastly simplified by this change. Far less dancing, no creating a `Span` just to have a container for storing in `Context` - we always just work with `Span`s. No need to explain to users what the difference between `SpanContext` and `Context` is, we only have one context. How to link to a span? Add a link to the span. 

I noticed that we use `DefaultSpan` for three purposes currently

- A propagated span, as in https://github.com/open-telemetry/opentelemetry-specification/pull/994. I added `Span.getPropagated`
- An unsampled, no-op span. I added `Span.getUnsampled`, though not sure it's a better name than `Span.getNoop`
- A test span just to use as a holder of context in unit tests. I just used `Span.getPropagated` for most of these though it's not technically precise, but we could just as easily define a test utility for this and I don't think it's a huge deal

Let me know any thoughts. Will finish up with javadoc if seems ok.

Fixes #1704
Fixes #1135


api/src/main/java/io/opentelemetry/trace/Link.java
```diff
@@ -29,13 +29,14 @@
...
-   */
-  SpanContext getContext();
+
+  String getTraceIdAsHexString();
+
+  String getSpanIdAsHexString();
+
+  TraceState getTraceState();
+
+  byte getTraceFlags();
```
Oberon00, line 39: This actually seems worse than before. Could we return a DefaultSpan here instead?

Oberon00:
This PR is already huge. I suggest moving the "make DefaultSpan private" to a separate/follow-up PR". You could select this PR's branch as target branch and link it in the PR description. Or ensure it is separate in the commit history of this PR at least.

api/src/main/java/io/opentelemetry/trace/Link.java
anuraaga: `Link` is strange in being over-coupled to export right now. I don't think export structures should be based on e.g., `Span` so I'll look at it in #1714.

anuraaga:
@Oberon00 I found a lot of code using `SpanContext` and `DefaultSpan` interchangeably. So I think the story of this PR isn't complete without combining both. If the direction is ok, I can split out the `DefaultSpan` visibility change if others also find it changes the reviewability significantly.

bogdandrutu (changes requested):
This needs to go to the specs, because we remove an important concept from the API

anuraaga:
Yeah but since https://github.com/open-telemetry/opentelemetry-specification/issues/999 is marked after-ga it won't be acted on right? Are we stuck with the cruft because of that? I would still say we haven't really removed anything, it's just a matter of presentation within the Java code.

trask:
@bogdandrutu can you give us some options for how we can move forward with this before GA?

This looks like a really good change, especially now with [Propagated Span](https://github.com/open-telemetry/opentelemetry-specification/pull/994/files), `SpanContext` doesn't seem to be pulling any weight.

I also think this is a very good point from @anuraaga:

> When users come to tracing for the first time, I think they understand the concept of a span. And they notice that a span is supposed to have a trace id and span id, they see this in tracing UIs. The presence of `SpanContext` seems to just add indirection that complicates the model for new users. It's especially confusing with a similarly named but very different `Context`.

trask:
@anuraaga maybe go ahead and send spec PR for https://github.com/open-telemetry/opentelemetry-specification/issues/999? IIRC after-ga means not required for ga, it doesn't preclude a spec PR from being accepted prior to GA, I would think especially a spec PR that doesn't require other SDKs to do any new work

Oberon00:
Spec is already frozen for non-editorial changes. Maybe you can get an exception for that one though? As it does not increase (required) work for SIGs.

trask:
I added this topic to the maintainers meeting agenda for tomorrow to explore our options

anuraaga:
@trask (I think) it's just a one line change so doesn't hurt to make a PR indeed :) https://github.com/open-telemetry/opentelemetry-specification/pull/1022

iNikem:
Does current spec prevent us from having `SpanContext` as interface, btw?

iNikem:
Does current spec prevent us from having `SpanContext` as interface, btw?

anuraaga:
@iNikem Nope that was changed in https://github.com/open-telemetry/opentelemetry-specification/pull/969. So for issues of bridging, I think an interface would be fine too if this can't be accepted. But I realized the API could be made much simpler here while working on that PR :)

iNikem:
If we make `SpanContext` an interface, and make `Span` implement it, maybe we can still simplify Java API significantly without direct contradiction to spec?

anuraaga:
@iNikem Didn't think of that idea - seems to be an ok option if `SpanContext` must survive but a bit of an unnecessary interface. Guess I can start with it to get this out of draft

opentracing_shim/src/main/java/io/opentelemetry/opentracingshim/ScopeManagerShim.java
```diff
@@ -30,11 +29,9 @@ public ScopeManagerShim(TelemetryInfo telemetryInfo) {
   @Override
   @SuppressWarnings("ReturnMissingNullable")
   public Span activeSpan() {
-    // As OpenTracing simply returns null when no active instance is available,
-    // we need to do an explicit check against DefaultSpan,
-    // which is used in OpenTelemetry for this very case.
+    // OpenTracing simply returns null when the current span is not valid.
     io.opentelemetry.trace.Span span = tracer().getCurrentSpan();
-    if (DefaultSpan.getInvalid().equals(span)) {
+    if (!span.isValid()) {
```
carlosalberto, line 34: Slightly incorrect, as in OT it's possible to set an invalid Span (one with zero-ed ids) as the active instance. We should check against `Span.getInvalid()`

anuraaga: Do you think you can give me a unit test that fails with the current code? OTel has many invalid checks, for example in instrumentation - I'd be surprised if we could even support a zero'd but live span in OTel. I'm guessing OT would propagate such a span, but OTel won't - so maybe we can just unify this behavior with the OTel semantics? Anyways a unit test would probably make things very clear, I'm not changing it back if we can't get the current code to fail first :)

anuraaga: Another way to phrase my point of confusion - I would think returning a singleton for an invalid span is a performance optimization but wouldn't expect any behavior change. So if the use of a singleton is relied on here, it's unexpected and seems fragile. If it's required I need to add some more detailed docs.

Oberon00: Not supporting valid spans with all zero trace *or* span ID is something that irks me about OTel actually https://github.com/open-telemetry/opentelemetry-specification/issues/753

anuraaga:
Separated out #1791 so closing this

PR #1713, opened 2020-09-26
Add overrides to allow using a long key with an int value.

### LINKED ISSUE #1711: Is there any way to avoid (long) cast when setting NET_PEER_PORT?
I've been updating to the new attribute keys, and they are very nice!

The only (very small) issue is having to cast to long (nearly) every time we call:

```
span.setAttribute(SemanticAttributes.NET_PEER_PORT, (long) port);
```

Not sure if it's worth adding `AttributeKey<Integer>`, or if spec even supports doing that, or maybe there's some other clever solution?

If no good solution, feel free to close this.

anuraaga:
Fixes #1711

### DIFF
--- a/api/src/test/java/io/opentelemetry/common/AttributesTest.java
+++ b/api/src/test/java/io/opentelemetry/common/AttributesTest.java
@@ -138,6 +138,7 @@ class AttributesTest {
         Attributes.newBuilder()
             .setAttribute("string", "value1")
             .setAttribute("long", 100)
+            .setAttribute(longKey("long2"), 10)
             .setAttribute("double", 33.44)
             .setAttribute("boolean", "duplicateShouldBeRemoved")
             .setAttribute("boolean", false)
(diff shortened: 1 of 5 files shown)


api/src/test/java/io/opentelemetry/common/AttributesTest.java
```diff
@@ -149,6 +150,8 @@ void builder() {
             "value1",
             longKey("long"),
             100L,
+            longKey("long2"),
+            10L,
```
jkwatson, line 154: hm. a shame it doesn't fix this usage as well. :(

jkwatson (approved):

api/src/test/java/io/opentelemetry/common/AttributesTest.java
trask: I did run into 2 of these in java-instrumentation, but just realized they probably should have been long in those particular cases anyways:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1266/commits/a83da9ed4fd0a26294391b45b34fc1404a2b4421

trask (approved):

PR #1716, opened 2020-09-26
Remove Descriptor from metrics data to be consistent with proto

bogdandrutu:


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/MetricData.java
```diff
@@ -54,22 +69,54 @@
    */
   public abstract InstrumentationLibraryInfo getInstrumentationLibraryInfo();
 
+  /**
+   * Returns the metric descriptor name.
```
jkwatson, line 73: These javadocs need to be updated, since there is no descriptor any more.

bogdandrutu: done.

jkwatson (approved):

PR #1723, opened 2020-09-28
Expose a new TracerSdkManagement interface on the OpenTelemetrySdk class.

### LINKED ISSUE #1635: Consider not using private type TracerSdk in public API TracerSdkProvider::get
``TracerSdkProvider::get`` is currently defined like this:

    @Override
    public TracerSdk get(String instrumentationName) {
      return tracerSdkComponentRegistry.get(instrumentationName);
    }

It uses covariant return types to return a package private type. 

When I use the API like this:

    OpenTelemetrySdk.getTracerProvider().get("foo");

javac will put the private class name ``TracerSdk`` into my own class. 

Renaming this private class would end up breaking binary compatibility. 

Having the covariant return type here also makes it impossible to instrument ``TracerSdkProvider::get`` with a Java agent in order to return an alternative tracer object when ``TracerSdkProvider::get`` is called in application code.

I therefore suggest to not leak internal details into method signatures of public APIs and instead change the method to

    @Override
    public Tracer get(String instrumentationName) {
      return tracerSdkComponentRegistry.get(instrumentationName);
    }

jkwatson:
Rather than expose the TracerSdkProvider directly, this introduces a new interface for managing the Tracing SDK. This will make it more obvious that the OpenTelemetrySdk access methods are not to be used for instrumentation purposes, but for operating the SDK itself.

This is a draft of how to resolve #1635 . 

Note: this only covers the Tracing side of the house. Metrics and Baggage probably want the same treatment, but I wanted to keep this relatively limited to make sure we're ok with moving in this direction.

### DIFF
--- a/exporters/inmemory/src/main/java/io/opentelemetry/exporters/inmemory/InMemoryTracing.java
+++ b/exporters/inmemory/src/main/java/io/opentelemetry/exporters/inmemory/InMemoryTracing.java
@@ -36,7 +36,7 @@ import javax.annotation.concurrent.Immutable;
  *   public void testCondition() {
  *     TracerSdkProvider tracerSdkProvider = TracerSdkProvider.builder().build()
  *     InMemoryTracing tracing =
- *         InMemoryTracing.builder().setTracerSdkProvider(tracerSdkProvider).build();
+ *         InMemoryTracing.builder().setTracerSdkManagement(tracerSdkProvider).build();
  *     Tracer tracer = tracerSdkProvider.getTracer("MyTestClass");
  *     tracer.spanBuilder("span").startSpan().end();
  *
--- a/exporters/inmemory/src/main/java/io/opentelemetry/exporters/inmemory/InMemoryTracing.java
+++ b/exporters/inmemory/src/main/java/io/opentelemetry/exporters/inmemory/InMemoryTracing.java
@@ -20,15 +20,14 @@ import com.google.auto.value.AutoValue;
 import io.opentelemetry.OpenTelemetry;
 import io.opentelemetry.context.propagation.DefaultContextPropagators;
 import io.opentelemetry.sdk.trace.TracerSdkManagement;
-import io.opentelemetry.sdk.trace.TracerSdkProvider;
 import io.opentelemetry.sdk.trace.export.SimpleSpanProcessor;
 import io.opentelemetry.trace.propagation.HttpTraceContext;
 import javax.annotation.concurrent.Immutable;
 
 /**
  * InMemoryTracing is an utility class that helps installing the {@link SimpleSpanProcessor} and an
- * instance of the {@link InMemorySpanExporter} to a given {@link TracerSdkProvider}. Can be used to
- * test OpenTelemetry integration.
+ * instance of the {@link InMemorySpanExporter} to a given {@link TracerSdkManagement}. Can be used
+ * to test OpenTelemetry integration.
  *
  * <p>Example usage:
  *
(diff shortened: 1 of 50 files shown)


sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/TracerSdkProvider.java
```diff
@@ -36,7 +36,7 @@
  * io.opentelemetry.OpenTelemetry}. However, if you need a custom implementation of the factory, you
  * can create one as needed.
  */
-public class TracerSdkProvider implements TracerProvider {
+public class TracerSdkProvider implements TracerProvider, TracerSdkManagement {
```
anuraaga, line 39: Does the new interface allow this to be private? And then maybe we don't need to change the return types

exporters/inmemory/src/main/java/io/opentelemetry/exporters/inmemory/InMemoryTracing.java
```diff
@@ -57,7 +58,7 @@
    * @return the {@code TracerSdkProvider} passed during construction.
    * @since 0.1.0
    */
-  abstract TracerSdkProvider getTracerProvider();
+  abstract TracerSdkManagement getTracerProvider();
```
iNikem, line 61: Methods names should be changed too, shouldn't they?

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/TracerSdkProvider.java
jkwatson: what do you want to be private?

exporters/inmemory/src/main/java/io/opentelemetry/exporters/inmemory/InMemoryTracing.java
jkwatson: yep. will fix.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/TracerSdkProvider.java
jkwatson: If you want to make TracerSdkProvider package-private, we would need to move the SPI implementation into this package: `TracerProviderFactorySdk`. That seems possible, if we're ok with moving it.

jkwatson: although it's used in a lot of tests as well right now, so we'd need to do quite a bit of test refactoring.

anuraaga: I feel like moving the SPI implementation into this package is a win, less API is nice - I would say SPI package usually defines an SPI interface, but doesn't have to or even should define an SPI implementation. Doesn't have to be in this PR though.

anuraaga (approved):

exporters/inmemory/src/main/java/io/opentelemetry/exporters/inmemory/InMemoryTracing.java
jkwatson: fixed.

jkwatson:
If I move this to a real PR, are we ok with metrics and baggage coming as a separate set of PRs?

trask:
This is a great change from auto-instrumentation perspective, as it cleanly separates TracerProvider (which we instrument) from the sdk management (which we do not instrument) 👍

anuraaga (approved):

jkwatson:
do we want this change for 0.9.0, or should we wait until we can do the same with metrics and baggage before releasing in 0.10.0?

PR #1727, opened 2020-09-29
check traceparent for extra trailing characters for version 00

### LINKED ISSUE #1675: invalid traceparent with extra trailing characters is not rejected
Next w3c context propagation tests fails:

```
def test_traceparent_version_0x00(self):
		'''
		harness sends an invalid traceparent with extra trailing characters
		expects a valid traceparent from the output header, with a newly generated trace_id
		'''
		traceparent, tracestate = self.make_single_request_and_get_tracecontext([
			['traceparent', '00-12345678901234567890123456789012-1234567890123456-01.'],
		])
		self.assertNotEqual(traceparent.trace_id.hex(), '12345678901234567890123456789012')

		traceparent, tracestate = self.make_single_request_and_get_tracecontext([
			['traceparent', '00-12345678901234567890123456789012-1234567890123456-01-what-the-future-will-be-like'],
		])
		self.assertNotEqual(traceparent.trace_id.hex(), '12345678901234567890123456789012')


```

Got in #1653
Can be reproduced by running tests from mentioned PR

malafeev:
fix #1675
based on spec https://www.w3.org/TR/trace-context-1/#traceparent-header

### DIFF
--- a/api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java
+++ b/api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java
@@ -211,6 +211,9 @@ public class HttpTraceContext implements TextMapPropagator {
       if (!VALID_VERSIONS.contains(version)) {
         return SpanContext.getInvalid();
       }
+      if (version.equals("00") && traceparent.length() > TRACEPARENT_HEADER_SIZE) {
+        return SpanContext.getInvalid();
+      }
 
       String traceId =
           traceparent.substring(TRACE_ID_OFFSET, TRACE_ID_OFFSET + TraceId.getHexLength());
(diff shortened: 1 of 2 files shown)


api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java
```diff
@@ -211,6 +211,9 @@ private static SpanContext extractContextFromTraceParent(String traceparent) {
       if (!VALID_VERSIONS.contains(version)) {
         return SpanContext.getInvalid();
       }
+      if (version.equals("00") && traceparent.length() > TRACEPARENT_HEADER_SIZE) {
```
jkwatson, line 214: What do you think about making "00" a constant in the class, rather than hard-coding it down here?

anuraaga (commented):
Can we add a unit test for the change? It's not clear to me if the integration test covers this case, but even if it does a unit test is fast and nice.

api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java
malafeev: done

malafeev:
@anuraaga added unit tests

anuraaga (approved):
Thanks!

jkwatson (approved):

PR #1728, opened 2020-09-30
Add workflow for Update Gradle Wrapper Action.

cristiangreco:
Hey there 👋, first of all thanks for your work on OpenTelemetry!

I've got a suggested change: would you be willing to use [this GitHub Action](https://github.com/gradle-update/update-gradle-wrapper-action) to automatically keep Gradle Wrapper updated to latest release?

**What does "Update Gradle Wrapper Action" do?** It can be configured to run at scheduled intervals (e.g. daily or weekly) and will check whether the Wrapper script in the repo is up-to-date to the latest Gradle release: in case a new Gradle version is available, it will create a PR to update the Wrapper. And that's it!

**Why is that a good thing?** Well, first of all it alleviates the chore of manually updating the Wrapper, as you got a task that keeps track of new Gradle releases for you! More importantly, it boosts security around the Wrapper update and usage processes: this actions verifies that the `gradle-wrapper.jar` file has not been tampered with (uses checksum comparison), and it sets the `distributionSha256Sum` property so that the new Gradle binary itself will be verified locally upon download.

**Where can I find more about?** The [README](https://github.com/gradle-update/update-gradle-wrapper-action/blob/master/README.md) contains quite detailed information!

In this PR I propose adding a new workflow which runs the action every day at midnight (but feel free to adjust the frequency as you prefer). I've verified it works correctly in my fork of the repo, and you can [see here](https://github.com/gradle-update/opentelemetry-java/pull/1) how the generated PR looks like, with multiple commits for each subproject in your repo.

The action is under active development, you can have a look at the [list of inputs](https://github.com/gradle-update/update-gradle-wrapper-action#action-inputs) currently supported. There's new features coming up soon and if you'd like to request any particular change just let me know!

I'd love to see the action used by OpenTelemetry and I genuinely hope you can find this useful. Would love your feedback! ❤️

### DIFF
--- a/.github/workflows/update-gradle-wrapper.yml
+++ b/.github/workflows/update-gradle-wrapper.yml
@@ -1,0 +1,19 @@
+name: Update Gradle Wrapper
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  update-gradle-wrapper:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Update Gradle Wrapper
+        uses: gradle-update/update-gradle-wrapper-action@74a035c
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - uses: gradle/wrapper-validation-action@v1


anuraaga (approved):
Thanks! This seems like a good idea to me.

jkwatson:
I'm personally nervous about accepting random GH action PRs. This seems like a possible security issue vector, since we don't know much about the source of the action.

anuraaga:
@jkwatson Thanks for the diligence - yeah didn't think of that but I agree. If the action was hosted in the gradle/ org I'd be comfortable with it (it's our primary tool) but the separate org indicates less scrutiny.

cristiangreco:
Hey @anuraaga @jkwatson thanks for your replies! I think I understand your security concern around introducing a new automated workflow. The source code for this action is licensed under Apache 2.0 and available [here](https://github.com/gradle-update/update-gradle-wrapper-action) for review. It's super small and should be easy to read through.

Regarding the fact that the code is hosted outside of the Gradle organisation, it's a bit of a chicken and egg problem. Unless more projects start to find this action useful I doubt it could be accepted for inclusion within the Gradle org. For this reason, would personally love if a project as important and well known as OpenTelemetry would consider using the action. And you're not alone, as projects like [Anki](https://github.com/ankidroid/Anki-Android/blob/master/.github/workflows/update-gradle-wrapper.yml) and [Element](https://github.com/vector-im/element-android/blob/develop/.github/workflows/update-gradle-wrapper.yml) have already started using it.

Another consideration around the security aspect: the whole purpose of this action is to make sure the Wrapper is updated in a reproducible way and also **verified**, that's why it additionally computes the checksum of the updated jar before creating the PR (and fails in case of mismatch). I see you already use the [Gradle Wrapper Validation](https://github.com/open-telemetry/opentelemetry-java/blob/master/.github/workflows/gradle-wrapper-validation.yml) action to verify the checksum of the wrapper jar on each code change. As an additional security measure, you can potentially concatenate the two actions like this:

```yml
      - name: Update Gradle Wrapper
        uses: gradle-update/update-gradle-wrapper-action@v1
        with:
          repo-token: ${{ secrets.GITHUB_TOKEN }}

      - uses: gradle/wrapper-validation-action@v1
```

with the net effect of getting the proposed change validated twice (second time by the official WVA). I'm happy to apply this change for you if that would make your more confortable using the action. 😊

Happy to answer any other question or comment you have, and thanks for taking the time!

Oberon00:
Since we seem to use a versioned action here `uses: gradle-update/update-gradle-wrapper-action@v1`, if this v1 is immutable and someone takes the time to shortly look over the code, it could be verified to be OK.

cristiangreco:
> Since we seem to use a versioned action here `uses: gradle-update/update-gradle-wrapper-action@v1`, if this v1 is immutable and someone takes the time to shortly look over the code, it could be verified to be OK.

Hey @Oberon00 👋, just a note about the `v1` tag: it is not immutable actually, it is moved to point to the latest patch version of the 1.x release. This follows the [GitHub](https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/about-actions#using-tags-for-release-management) [guidelines](https://github.com/actions/toolkit/blob/master/docs/action-versioning.md) around versioning actions, so that you - as a user - can write your workflow once and continue to benefit from non-breaking (minor and patch) releases of the action over time. You could decide to stick to a particular version (e.g. `v1.0.8`) but I wouldn't recommend as I'm continuing to introduce valuable changes (e.g. more descriptive PR content).

Oberon00:
If we have a trust problem (not saying we have), just reviewing the code is not enough, if the release is mutable. Nothing prevents a malicious maintainer, or someone who hacks the maintainer's account, from pushing a 1.0.1 with malicious changes.

anuraaga:
> Regarding the fact that the code is hosted outside of the Gradle organisation, it's a bit of a chicken and egg problem. Unless more projects start to find this action useful I doubt it could be accepted for inclusion within the Gradle org. 

Thanks for phrasing it as this - I do want to help the ecosystem so I think we can be receptive here. @Oberon00 is correct that with the mutable release, just checking the code is unfortunately not enough to address security concerns - can you change it to reference the sha hash instead? I think we can be pretty comfortable with that.

cristiangreco:
> > Regarding the fact that the code is hosted outside of the Gradle organisation, it's a bit of a chicken and egg problem. Unless more projects start to find this action useful I doubt it could be accepted for inclusion within the Gradle org.
> 
> Thanks for phrasing it as this - I do want to help the ecosystem so I think we can be receptive here. @Oberon00 is correct that with the mutable release, just checking the code is unfortunately not enough to address security concerns - can you change it to reference the sha hash instead? I think we can be pretty comfortable with that.

Cool, that sounds good to me! Will update this PR with the sha of the latest version.

cristiangreco:
@anuraaga @Oberon00 and everyone involved, I've updated the workflow file to point to release v1.0.9 (74a035c) and have added a new step that runs `wrapper-validation-action`. Let me know if everything looks ok to you.

anuraaga:
Thanks for the change - it will take me a bit to read through the code for that hash. After making sure the code looks ok I don't see any other reason to block.

jkwatson:
> Thanks for the change - it will take me a bit to read through the code for that hash. After making sure the code looks ok I don't see any other reason to block.

@anuraaga Have you had a chance to look at this yet?

jkwatson:
> > Thanks for the change - it will take me a bit to read through the code for that hash. After making sure the code looks ok I don't see any other reason to block.
> 
> @anuraaga Have you had a chance to look at this yet?

@anuraaga ?

anuraaga:
So sorry for the delay on this @jkwatson @cristiangreco. I have finally reviewed the source code, and can confirm that the source code is safe. That being said, one improvement would be using the normal github-actions name / email here

https://github.com/gradle-update/update-gradle-wrapper-action/blob/master/src/index.ts#L76

This would follow the GitHub documentation at https://github.com/actions/checkout#push-a-commit-using-the-built-in-token

But on a related note, @cristiangreco do you know of a way to have the automated PR pass our CLA check?

PR #1729, opened 2020-09-30
Use awaitility / higher timeout to reduce flakiness.

anuraaga:
Noticed these sometimes fail randomly. Also went ahead and made awaitility automatic test dependency since it's otherwise harmless yet often useful since we have many async tests.

### DIFF
--- a/sdk/common/src/test/java/io/opentelemetry/sdk/common/CompletableResultCodeTest.java
+++ b/sdk/common/src/test/java/io/opentelemetry/sdk/common/CompletableResultCodeTest.java
@@ -240,7 +240,7 @@ class CompletableResultCodeTest {
               result.succeed();
             })
         .start();
-    assertThat(result.join(500, TimeUnit.MILLISECONDS).isSuccess()).isTrue();
+    assertThat(result.join(10, TimeUnit.SECONDS).isSuccess()).isTrue();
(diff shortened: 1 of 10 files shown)


sdk/common/src/test/java/io/opentelemetry/sdk/common/CompletableResultCodeTest.java
```diff
@@ -240,7 +240,7 @@ void join() {
               result.succeed();
             })
         .start();
-    assertThat(result.join(500, TimeUnit.MILLISECONDS).isSuccess()).isTrue();
+    assertThat(result.join(10, TimeUnit.SECONDS).isSuccess()).isTrue();
```
anuraaga, line 243: 10 seconds tends to be a "common test timeout" so just went with that here.

jkwatson (approved):
Thanks! Was this failing locally for you? Maybe time for a new laptop?  :)

anuraaga:
@jkwatson Was failing on GitHub actions Mac executor. Though I do want a new laptop :)

jkwatson:
> @jkwatson Was failing on GitHub actions Mac executor. Though I do want a new laptop :)

ah, yes. I haven't ever seen this fail locally on Mac OS. I wonder what's up in GH.

PR #1730, opened 2020-09-30
Add GitHub Actions continuous build to replace CircleCI.

anuraaga:
For #1381

We've made a lot of good progress on migrating instrumentation repo to actions. This repo is much easier, so we may as well do it. It let's us easily test multiple OS's, to avoid an e.g. Windows contributor have the build fail for them. Supports testcontainers well without having to fallback to a machine executor. And I think we can use a similar config for release builds later, which would allow starting a release from a button on the UI without any local, manual steps at all.

I haven't added snapshot publishing to this yet since presumably that should go together with removing the CircleCI master build. I do upload to codecov since I think uploading twice is harmless. And I just run `:check` - the circleci config separates out several tasks, but from what I can tell `:check` runs all of them anyways.

### DIFF
--- a/.github/workflows/continuous-build.yml
+++ b/.github/workflows/continuous-build.yml
@@ -1,0 +1,44 @@
+name: "Continuous Build"
+
+on:
+  pull_request:
+    branches:
+      - master
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+    name: Build
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os:
+          - macos-latest
+          - ubuntu-latest
+          - windows-latest
+        java:
+          # Test all OS's with Java 11
+          - 11
+        include:
+          - os: ubuntu-latest
+            java: 11
+            coverage: true
+          - os: ubuntu-latest
+            java: 8
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+      - uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+      - uses: burrunan/gradle-cache-action@v1.5
+        with:
+          job-id: jdk${{ matrix.java }}
+          arguments: check --stacktrace ${{ matrix.coverage && ':opentelemetry-all:jacocoTestReport' || '' }}
+          properties: |
+            enable.docker.tests=${{ matrix.os == 'ubuntu-latest' }}
+      - uses: codecov/codecov-action@v1
+        if: ${{ matrix.coverage }}


.github/workflows/continuous-build.yml
```diff
@@ -0,0 +1,44 @@
...
+      - master
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+    name: Build
+    runs-on: ${{ matrix.os }}
+    strategy:
```
iNikem, line 15: You may want to use `fail-fast: false`

iNikem (approved):

.github/workflows/continuous-build.yml
anuraaga: Yeah I thought of it - I think on instrumentation repo we have pretty different test profile for each environment (e.g. `maxVersionForTest`. And some flakiness :) Here I'm expecting fail-fast to generally help rather than hurt since I guess it's more common that one failure == all failure.

jkwatson: I think this is correct. When that's not true, it's usually a flaky test, but we don't tend to have very many of those, so failing fast seems like the right strategy. We can definitely revisit if we start getting more flaky tests.

jkwatson (approved):
let's give it a whirl! I assume that the circle jobs will continue to run for now as well.

jkwatson:
@anuraaga hrm.  https://github.com/open-telemetry/opentelemetry-java/actions/runs/280669708 maybe we shouldn't build on MacOS for CI?

jkwatson:
hm. there's an actual failing test in this one:  https://github.com/open-telemetry/opentelemetry-java/runs/1188561431

And, digging deeper on the previous one I mentioned, there's a failing test there, as well.  :(

jkwatson:
FYI, these builds are passing on circle, so it's something specific to the GH actions.

anuraaga:
@jkwatson Thanks for these links - interesting the span export problem wasn't just a timing issue. I'll keep an eye on these, in the meantime let's make sure it's not a required check (btw, I realized I don't have access to the Settings tab of this repo in case that's not intentional).

jkwatson:
> @jkwatson Thanks for these links - interesting the span export problem wasn't just a timing issue. I'll keep an eye on these, in the meantime let's make sure it's not a required check (btw, I realized I don't have access to the Settings tab of this repo in case that's not intentional).

I also have very limited access to the Settings tab. I think only @carlosalberto and @bogdandrutu have full access as TC members.

anuraaga:
The maintainers have it in the instrumentation repo FWIW.

When creating a release build flow on GitHub I'd need to be able to add secrets. I'm ok with either doing it directly or just bugging people when needed :)

jkwatson:
another failure: https://github.com/jkwatson/opentelemetry-java/runs/1194229575

PR #1732, opened 2020-10-01
Remove addLink that accepts Link interface

bogdandrutu:
Currently the Link interface cannot easily be removed, used in the Sampler interface,
will do a separate PR for that.

Based on the specs we don't have to offer a "lazy" formatted Link API (was removed some time ago).


jkwatson (approved):
very happy to make the API surface smaller, pre release.
Thanks!

api/src/main/java/io/opentelemetry/trace/Span.java
```diff
@@ -444,10 +444,13 @@ default void setAttribute(AttributeKey<Long> key, int value) {
     /**
      * Adds a {@link Link} to the newly created {@code Span}.
      *
+     * <p>Links are used to link {@link Span}s in different traces. Used (for example) in batching
```
anuraaga, line 447: Let's copy this to the other overload too.

anuraaga (approved):
Thanks!

api/src/main/java/io/opentelemetry/trace/Span.java
bogdandrutu: Done.

PR #1733, opened 2020-10-01
Remove Event interface, and it's usage

bogdandrutu:
Based on the specification we are not required to expose this interface or have APIs that accept this interface.


sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/data/SpanData.java
```diff
@@ -237,7 +237,23 @@ public static Link create(
     Link() {}
   }
 
-  interface Event extends io.opentelemetry.trace.Event {
```
jkwatson, line 240: 🦸

jkwatson (approved):
yes yes yes. Thanks!

anuraaga (approved):

PR #1736, opened 2020-10-02
Increase deadline for deadline test.

anuraaga:
If the thread sleeps for 500ms or so between the calls of `withDeadlineAfter` and `export`, the test will fail. A full GC or similar, during CI, can get to that range and we see this happen more often on GitHub Actions which might have more thread contention that CircleCI. Increasing the deadline should reduce the chance of this, and I think for the original intent of the regression test, we can get seconds of runtime back by only exporting once and sleeping for a small amount more than the deadline, which guarantees triggering the previously broken behavior.


jkwatson (approved):
let's try it!

PR #1737, opened 2020-10-02
Make OpenTelemetry API an instance so it can be injected.

### LINKED ISSUE #747: Registries created through builders should be returned by OpenTelemetry static methods
TracerSdkRegistry, MeterSdkRegistry and DistributedContextManagerSdk objects created programmatically using the builders and intended to be global registries should be returned by the static methods of the OpenTelemetry class. Currently new registry objects with default configuration get created using the Java ServiceLoader mechanism. This should be the fallback if they are not set explicitly.

### LINKED ISSUE #1109: Tracing multiple applications in application servers
In an application server that supports multiple applications, we are using OpenTracing API in JAX-RS filters to automatic instrument applications. Each application has its own instance of `Tracer`. Implementations such as `BraveTracer` or `JaegerTracer` can be instantiated and configured with their own service names.  Multiple tracer implementations can export spans to a single OpenTracing server. 

In OpenTelemetry, created 2 instances of `Tracer` using

```
Tracer tracer1 = OpenTelemetry.getTracerProvider().get("app1");
Tracer tracer2 = OpenTelemetry.getTracerProvider().get("app2");
```

Created a `SpanProcessor` and  `SpanExporter` using
```
ManagedChannel jaegerChannel = ManagedChannelBuilder.forAddress(ip, port).usePlaintext().build();
JaegerGrpcSpanExporter jaegerExporter = JaegerGrpcSpanExporter.newBuilder().setServiceName("app1")
                .setChannel(jaegerChannel).setDeadlineMs(30000).build();
OpenTelemetrySdk.getTracerProvider().addSpanProcessor(SimpleSpansProcessor.newBuilder(jaegerExporter).build());
```

In this case, spans from `tracer1` and `tracer2` got reported as "app1" in the backend because of the serviceName set in the `SpanExporter`.  Creating another instance of `SpanProcessor` for "app2" to the same backend will result of duplication of span IDs in the backend.

### LINKED ISSUE #1149: Add setup method to configure the OpenTelemetry API
According to the [specifications](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-configuration.md#programmatic), the OpenTelemetry SDK must be programmatically configurable. The current initialization looks for implementations of the OpenTelemetry API via Service Provider and if no instance is found, it defaults to a no-op implementation. In the SIG call, we agreed that the default behavior should be no-op until a configuration is provided. 

#1058 provides a first implementation that allows to set up individually tracer provider, metric provider, and context manager. We should iterate over this proposal and making it friendly to the auto-instrumentation.

### LINKED ISSUE #1454: Being able to specify Resource manually

anuraaga:
This makes `OpenTelemetry` not a bag of static methods but an instantiable interface itself. We've heard feedback from Spring that they prefer to inject telemetry APIs rather than rely on statics - this is a relatively reasonable pattern for Java. So we can imagine a spring configuration look like

```java
@Configuration
class OpenTelemetryConfiguration {

  @Bean
  TracerProvider tracerProvider(Optional<IdsGenerator> idsGenerator, Optional<Resource> resource) {
    TracerSdkProvider.Builder builder = TracerSdkProvider.builder();
    idsGenerator.ifPresent(builder::setIdsGenerator);
    resource.ifPresent(builder::setResource);
    return builder.build();
  }

  @Bean
  MeterProvider meterProvider(MeterRegistry meterRegistry) {
   return new MicrometerMeterProvider(meterRegistry);
  }

  @Bean
  OpenTelemetry openTelemetry(TracerProvider tracerProvider, MeterProvider meterProvider, OpenTelemetryProperties properties) {
    OpenTelemetrySdk.Builder builder OpenTelemetrySdk.builder()
      .setMeterProvider(meterProvider);
    if (!properties.getPropagators().isEmpty()) {
     builder.setPropagators(propagators(properties.getPropagators());
    }
    return builder.build();
  }
```

And the `OpenTelemetry` instance can be injected as needed, and different `OpenTelemetry` can be used in different scopes as desired. Inspired by Brave, the first built SDK also gets automatically registered as the global, statically accessible API - both patterns are supported which I think is user friendly.

We have several concerns that are cross-cutting, and currently on each provider, but By allowing `OpenTelemetry` to have state, we can manage them in the same location

- `Resource`
- `Clock`
- Context propagation

In a future PR, I would want to remove `Resource` / `Clock` from the subproviders and have them referenced through the `OpenTelemetry` itself I think. Also, I would make all of the classes immutable - the `OpenTelemetry` API itself does not need to be mutable, it's only our static accessors of a "global OpenTelemetry" that does. @bogdandrutu mentioned some concern about performance overhead of volatile lookups, and while we currently have a lot of volatile state in the tracer provider, it would be simple to have all such state be updated by reseting the global OpenTelemetry with another immutable one, instead of individual volatile mutable state. And users of direct OpenTelemetry injection will have no volatile lookups anywhere.

The downside is this approach removes support for this part of the spec

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#tracer-creation

"The TracerProvider MAY provide methods to update the configuration" - instead the `TracerProvider` would be immutable and the expectation is it is configured up front when being built before any tracers are used. This seems pretty conventional to me so not sure if this is that bad, interested in any thoughts.

We could also consider removing the individual SPIs for the providers, I think the OpenTelemetry SPI should satisfy it OK and it could make things simpler to just have one SPI.

If this approach seems ok at all, I will clean up javadocs, add some tests, and finish this up.

Fixes #1109 (each app can have its own OpenTelemetry for different Resource which seems to be the crux of the issue)
Fixes #1149
Fixes #1454
Fixes #747


jkwatson:
I really like this change, thanks very much for putting it together.

So, I'd like to validate that I understand how we would be recommending people should use this change for 2 use-cases.

1) Library instrumentation: library instrumentation code should always take an OpenTelemetry instance (or TracerProvider, if they only need tracing) as a DI-able attribute, rather than accessing the global instance. Maybe fall back to the global instance if one isn't injected?
2) Manual instrumentation without the agent: Create your implementation, or use the global, and inject it wherever you need to use it, including any library instrumentation you're using.
3) Manual instrumentation with the agent: Do whatever you want, the agent will end up putting its own implementation in the right place at runtime?

Does that line up with your understanding?

anuraaga:
@jkwatson Yup, basically. However I think we will need to accept `OpenTelemetry` as the configuration for libraries even if they only do tracing, not `TracerProvider`. I realized there is actually a great use case for per-library `OpenTelemetry` - configuring the propagators. While it's simple and safe to have a multi propagator used everywhere in `OpenTelemetry`, advanced users should be allowed to set propagators for backends when they know only one is supported - for the foreseeable future, it's far more efficient for the AWS SDK to only support x-ray propagation for example, so since propagators are set on `OpenTelemetry`, we end up with having that being the one-and-true configuration point.

trask:
> While it's simple and safe to have a multi propagator used everywhere in `OpenTelemetry`, advanced users should be allowed to set propagators for backends when they know only one is supported - for the foreseeable future, it's far more efficient for the AWS SDK to only support x-ray propagation for example

even without this use case, I think we want to inject propagators into libraries, just to further reduce static lookups?

api/src/main/java/io/opentelemetry/OpenTelemetry.java
```diff
@@ -5,264 +5,253 @@
...
-   * @since 0.3.0
+   * @param instrumentationName The name of the instrumentation library, not the name of the
+   *     instrument*ed* library (e.g., "io.opentelemetry.contrib.mongodb"). Must not be null.
+   * @return a tracer instance.
    */
-  public static void setPropagators(ContextPropagators propagators) {
-    Objects.requireNonNull(propagators, "propagators");
-    getInstance().propagators = propagators;
+  default Tracer getTracer(String instrumentationName) {
+    return getTracerProvider().get(instrumentationName);
```
Oberon00, line 178: I'm a bit worried about this. The docs say it is just a shortcut, but it is overridable.

anuraaga: Great point, removed the wording about shortcuts for the interface methods since not a big deal.

Oberon00: I would not be sure if this is not a big deal. This now allows creation of a tracer that does not go through a TracerProvider. I think we must at least add javadoc that this method should not be overridden, or if it is, it must only create a tracer through a TracerProvider.

jkwatson: If someone wants to muck about with the API in non-spec-compliant ways, isn't that really their problem? Not sure why this matters all that much in the real world.

jkwatson:
After pondering this for a while, I am currently deeply ambivalent about this change. 

On the plus side, I think that the ability to inject an OpenTelemetry instance into instrumentation is a boon for instrumentation authors, and for end-users who don't desire (or are unable) to use SPI as the means of configuring an SDK implementation.

On the minus side, I do think this complicates the story of how to use the OpenTelemetry API significantly. As a new user, attempting to write instrumentation, or add manual instrumentation to my application, I think I would be confused by the various options that were exposed on the OpenTelemetry interface. I fear that if we adopt this model, we will have a significantly greater support burden onboarding new users, both end-users, and instrumentation authors.

If we end up with general consensus on moving forward with this change, we should do a very large overhaul of our 'getting started' documentation, with very clear sections broken out by user persona : library author, application author, operator. This is probably a good idea, either way, but this change would make that documentation rework very urgent, to avoid issues going forward.

Also, we probably would want to put in PRs to update existing library instrumentation, [such as this in apache camel](https://github.com/apache/camel/blob/master/components/camel-opentelemetry/src/main/java/org/apache/camel/opentelemetry/OpenTelemetrySpanAdapter.java#L53-L62).

ebullient:
This would certainly make my attempt easier: 
https://github.com/micrometer-metrics/micrometer/pull/2282

@anuraaga .. on the call last Wednesday, I shared the _rough_ beginnings of an attempt at an OTel Micrometer Registry (rather than a MeterProvider.. ). What I'm attempting with the former might morph into the latter, we'll see. 

If the whole of this isn't palatable, minimally being able to remove existing instances (e.g. not just get but also remove) would be nice.

trask:
@jkwatson how would you feel about this PR with only one static method on OpenTelemetry e.g. `getInstance()` which returns the global instance, and then end-users would use:

* OpenTelemetry.getInstance().getTracerProvider()
* OpenTelemetry.getInstance().getTracer()
* OpenTelemetry.getInstance().getMeterProvider()
* OpenTelemetry.getInstance().getMeter()

jkwatson:
> @jkwatson how would you feel about this PR with only one static method on OpenTelemetry e.g. `getInstance()` which returns the global instance, and then end-users would use:
> 
> * OpenTelemetry.getInstance().getTracerProvider()
> * OpenTelemetry.getInstance().getTracer()
> * OpenTelemetry.getInstance().getMeterProvider()
> * OpenTelemetry.getInstance().getMeter()

By "end users" here are you referring to application developers who want to do manual instrumentation? Or application users who want to configure their library instrumentation, when the agent isn't present? Both, I suppose.

So, that "getInstance()" would return a "classic" singleton instance, but just the "default" instance, that's either been manually installed, or installed via SPI?

I wonder if that name might be misleading, but I do think this might simplify explaining to people what's going on.

Again, I want to make it clear: I love this idea, and I want to help make it happen...I just want to make sure we're all prepared to update all the docs and help with the transition to the new model. :)

trask:
Does `get()` instead of `getInstance()` help it look like less of a classic singleton? (it's shorter 😄)

* OpenTelemetry.get().getTracerProvider()
* OpenTelemetry.get().getTracer()
* OpenTelemetry.get().getMeterProvider()
* OpenTelemetry.get().getMeter()

or maybe `getDefault()`?

anuraaga:
@jkwatson Thanks! Yeah the many global APIs causing confusion was definitely a concern but I think @trask's suggestion to just have `.get` helps a lot let me try it out.

We definitely will need to reflect the builders in the docs. Creating `OpenTelemetry` seems natural compared to reconfiguring it statically so this will hopefully make the docs cleaner. I'm happy to add updates to the docs in this PR to see what it looks like.

@trask By the way I'm thinking about the agent case, but do we still need the API bridge if we can set the implementation of the API to one created by the agent?

trask:
> @trask By the way I'm thinking about the agent case, but do we still need the API bridge if we can set the implementation of the API to one created by the agent?

I might be missing your question, it will definitely help reduce our instrumentation points, but most of the bridge is the implementation of the API using the API in the agent.

anuraaga:
I've done some cleanup and added a README update (which I will unadd and stash if things look ok to apply it after next release). Let me know if things make sense. It is the end-user application hat, instrumentation hat docs we'd probably add to the instrumentation repo.

> but most of the bridge is the implementation of the API using the API in the agent.

@trask Yeah was just wondering, is it possible to instead do something like `OpenTelemetry.set(sdkShadedAndCreatedByAgent)` without instrumenting each API? I think the forwarding was mostly to allow the application's OpenTelemetry to use the SDK we setup in the agent. But I'm probably forgetting something else too :)

trask:
> @trask Yeah was just wondering, is it possible to instead do something like `OpenTelemetry.set(sdkShadedAndCreatedByAgent)` without instrumenting each API?

👍 OpenTelemetryApiInstrumentation will get simpler with this.

Most of the other instrumentation (`TracingContextUtilsInstrumentation`, `ContextUtilsInstrumentation`, `GrpcContextInstrumentation`, `BaggageUtilsInstrumentation`) is hacky gRPC Context instrumentation.

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
```diff
@@ -5,50 +5,351 @@
...
-   * @return MeterProvider returned by {@link OpenTelemetry#getMeterProvider()}.
+   * @param providerClass a provider class
+   * @param <T> provider type
+   * @return a provider or null if not found
+   * @throws IllegalStateException if a specified provider is not found
    */
-  public static MeterSdkProvider getMeterProvider() {
-    return (MeterSdkProvider) OpenTelemetry.getMeterProvider();
+  @Nullable
+  private static <T> T loadSpi(Class<T> providerClass) {
```
jkwatson, line 300: It bums me out that this method is an exact copy of the one in DefaultOpenTelemetry. It seems like having a common SpiUtilities somewhere in the API would be useful so we don't have to repeat ourselves on this boilerplate.

kenfinnigan: Maybe add an abstract `AbstractOpenTelemetry` class to the api that has this method, and maybe others, that `DefaultOpenTelemetry` and `OpenTelemetrySdk` can both extend?

jkwatson (approved):
I approve. Since this includes a significant change to the API surface (the OpenTelemetry global), at least one of @carlosalberto or @bogdandrutu should also give it their approval.

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
anuraaga: Yeah it's not ideal but since it's private and still relatively small, I'm hoping to address this in a followup instead of adding more public API right away for this.

QUICKSTART.md
```diff
@@ -31,10 +32,34 @@ configuration must be provided by **Applications** which should also depend on t
...
+
+To get started, setup the OpenTelemetry SDK - note that if you are using a different implementation 
+of the OpenTelemetry API, you should refer to the configuration instructions for that SDK. Also, if
+you are using the Auto-Instrumentation Agent, the SDK is configured by the agent itself and you
+should not setup your own.
+
+You can setup a basic configuration that exports traces to a logging stream like this:
+
+```java
+OpenTelemetry openTelemetry = OpenTelemetrySdk.newBuilder()
```
carlosalberto, line 45: Is there any reason to not use the auto-detected path as the default? i.e. I'd prefer we mention first the 54-56 lines - and then as the more detailed version (which is something that in general more advanced users will use, such as the Auto-Instrumentation).

api/src/main/java/io/opentelemetry/OpenTelemetry.java
```diff
@@ -5,264 +5,160 @@
...
+   */
+  static void set(OpenTelemetry openTelemetry) {
+    DefaultOpenTelemetry.setGlobalOpenTelemetry(openTelemetry);
+  }
 
-  private volatile ContextPropagators propagators = DefaultContextPropagators.builder().build();
+  /**
+   * Returns the globally registered {@link ContextPropagators} for remote propagation of a context.
+   */
+  static ContextPropagators getGlobalPropagators() {
```
carlosalberto, line 69: I personally prefer the old `getPropagators()` as it's shorter. Should be fine as long as we document it as the globally registered instance, we should be fine.

api/src/main/java/io/opentelemetry/OpenTelemetry.java
```diff
@@ -5,264 +5,160 @@
...
+   * io.opentelemetry.spi.OpenTelemetryFactory}, {@link
+   * io.opentelemetry.trace.spi.TracerProviderFactory}, {@link
+   * io.opentelemetry.metrics.spi.MeterProviderFactory}, or {@link
+   * io.opentelemetry.baggage.spi.BaggageManagerFactory} found on the classpath, or otherwise * will
+   * be default, with no-op behavior.
+   *
+   * @throws IllegalStateException if a provider has been specified by system property using the
+   *     interface FQCN but the specified provider cannot be found.
+   */
+  static OpenTelemetry get() {
```
carlosalberto, line 53: Although this is nice, I don't _really_ want to be writing this every time I use `OpenTelemetry` ;)

```java
OpenTelemetry.get().getTracerProvider().etc()
OpenTelemetry.get().getMeterProvider().etc()
```

I'd rather provide this `get()` method **plus** shortcuts (I know we are not fan of them, but hey, they are in the same class):

```java
OpenTelemetry.getTracerProvider().etc()
OpenTelemetry.getMeterProvider().etc()
```

carlosalberto:
Left a few comments - overall it looks great (this is a much needed improvement). Mostly I'd like to keep the previous access as shortcuts (which you guys already discussed):

```java
OpenTelemetry.getTracerProvider() //global
OpenTelemetry.getMeterProvider() // global
OpenTelemetry.get() // Useful in case you want to capture it all.
```

As said in the comment, not sure writing `get()` all the time is fun/needed. The majority of users will simply use the global instance, and this would make this easier (and of course, you can always fetch the entire global instance through `get()` if you want).

anuraaga:
> The majority of users will simply use the global instance, and this would make this easier

I'm not sure about this - currently the SDK isn't really functional without configuring a span processor at least anyways, so directing them to build the SDK instead of using the global SDK and mutating it a very small delta. Since DI is so common, especially with the vast usage of Spring, my hypothesis is this would actually be more desired as it follows the pattern of Micrometer and Brave.

Agent users will need to use the static accessors indeed, but I think they do far less instrumentation in general, and will often use `@WithSpan` and if they want to add attributes, I'm still going to propose the idea that `Span.current().setAttribute(...)` is just a lot simpler than going through tracer.

Also expecting SDK to be built up front instead of mutating the static singletons will fit better with my followup goal here to improve performance a bit:
 
> Also, I would make all of the classes immutable - the OpenTelemetry API itself does not need to be mutable, it's only our static accessors of a "global OpenTelemetry" that does. @bogdandrutu mentioned some concern about performance overhead of volatile lookups, and while we currently have a lot of volatile state in the tracer provider, it would be simple to have all such state be updated by reseting the global OpenTelemetry with another immutable one, instead of individual volatile mutable state. And users of direct OpenTelemetry injection will have no volatile lookups anywhere.

I'm ok with having the shortcuts though, but I would have a preference for including the word `global` to differentiate between a different configured instance. I'm also ok with including the shortcuts with the current shorter names for now and reconsider as other aspects may change. Let me know what you think after this extra info.

bogdandrutu (commented):
I would prefer to split this PR into two parts (even if after the first one we will not solve the Spring problem):
* Make Otel an immutable object/interface (not necessary an interface, we can discuss about that)
* Change the global and initialization logic.

anuraaga:
@bogdandrutu Sure will split out to make it easier to review. At least for me it tends to be easier to see the big picture with the full change so if any high level comments on the general approach probably good to have here.

api/src/main/java/io/opentelemetry/OpenTelemetry.java
anuraaga: Realized we can't have static and non-static with the same name anyways, ow

jkwatson:
Spent some more time digging through this today. One thing that seems like we could do is to have the OpenTelemetrySdk builder just build an instance of DefaultOpenTelemetry, rather than having it re-implement the interface. That might simplify things a bit, maybe? Or maybe it needs to extend it, to provide the additional "management"/operational methods? 

That might just end up confusing things, though, so just consider it an idea, not a firm suggestion. :)

jkwatson:
can this be closed with the other 2 PRs being merged now?

anuraaga:
Yeah will continue in other PRs (the task itself is not done yet :) )

PR #1738, opened 2020-10-02
Rename EventImpl to ImmutableEvent

bogdandrutu:


jkwatson (approved):
I agree this is a better name.

PR #1739, opened 2020-10-02
Remove Link from API, use SpanData.Link in Sampler

bogdandrutu:


jkwatson:
What do you think about pulling `Link` up out of the SpanData class, since it's now used for something else aside from Export?

bogdandrutu:
Not sure what is the benefit. Also I want to keep the dependency order the same. The SDK can depend on data but not the other way around

bogdandrutu:
Also I would do later, this one is important to cleanup the API first.

jkwatson:
> Not sure what is the benefit. Also I want to keep the dependency order the same. The SDK can depend on data but not the other way around

I was thinking of keeping it in the data package, just hoisting it up out of the SpanData. The benefit is that it's now used outside the context of SpanData, and so shouldn't be embedded as an auxiliary class in it.

jkwatson (approved):
API cleanup++
Thanks!

Issue #1438, opened 2020-07-20
Create a stress test that shows trace-SDK behavior under high load


jkwatson:
I'm particularly interested in making sure that we don't grow our heap in an unbounded fashion under high load. And, if we make changes to the export pipeline, that we don't inadvertently make a change that results in unbounded heap growth.

Note: this does not necessarily need to be a part of the unit test suite, because it might take a few minutes to run. However, having a test harness that can be run to verify this behavior when changes are made would be extremely valuable.

huntc:
In advance, I’d say that we probably are growing the heap in a unbounded manner presently. We should look at the use case where an exporter blocks on a network connection and goes through its retries. The Telemetry should pile up given the unbounded queues used by the schedulers within the sdk.

jkwatson:
The BatchSpanProcessor is built to drop spans beyond the configured limit. If that is not happening, then we definitely need to know about it!

huntc:
The `SimpleSpanProcessor` would be a good one to try. The simple one calls `export` when a span ends. Just keep ending spans...

jkwatson:
Yes, I think that the SSP probably wouldn't do well at all with a slow exporter and high throughput! I'm more worried personally about the BSP, since it actually intends to have dealt with this issue. :)

jkwatson:
Ideally, this stress test would additional include sending the data to the collector via OTLP, and have the ability to introduce network issues in between the running JVM and the collector. Maybe with something like toxiproxy?

parallelstream:
Hey @jkwatson for our internal needs at Atlassian we created a tool which can define OTel data as a simple JSON structure and replay it against running collector - https://github.com/parallelstream/open-telemetry-data-simulator. We can possibly modify it to add toxiproxy. Is this smth which can be useful as general purpose testing tool?

jkwatson:
that sounds very useful, yes. the stress test that I wrote could definitely use additional beefing up. Could your tool run the data through the SDK and then through to the collector?

jkwatson:
I'm going to close this particular issue for now. We have a basic stress test available and I have validated that the SDK behaves acceptably under adverse network conditions. @parallelstream If you are able to contribute more testing tools, that would be fantastic.

PR #1740, opened 2020-10-02
Simple throughput tester tool, with toxiproxy to make things difficult.

jkwatson:
This is a very very simple little stress-testing setup that I'm starting to play around with. It's the first step on the way to doing something useful around #1438 . 

Currently, it sends spans to the collector and outputs metrics on the console, in order to be able to see how many spans are being dropped when things get difficult.

It uses toxiproxy in docker-compose to sit in between the SDK and the collector, to enable making life hard for the either the collector, or the SDK or both. 

The load it is currently generating is single-threaded, almost certainly biased in every bad way possible.


perf_harness/src/main/java/io/opentelemetry/perf/OtlpPipelineDriver.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    IntervalMetricReader intervalMetricReader = setupSdk();
+
+    Tracer tracer = OpenTelemetry.getTracer("io.opentelemetry.perf");
+
+    for (int i = 0; i < 10000; i++) {
+      Span exampleSpan = tracer.spanBuilder("exampleSpan").startSpan();
+      try (Scope scope = tracer.withSpan(exampleSpan)) {
+        exampleSpan.setAttribute("good", "true");
```
iNikem, line 60: Maybe start using setup described in https://github.com/open-telemetry/opentelemetry-specification/pull/748 ?

jkwatson: Yeah, I hadn't even begun to think about the shape of the data being pushed through here. My main goal with this PR was just to get something up and running, with toxiproxy in between the collector and the SDK, so we can start making it more interesting and realistic over time.

perf_harness/docker/README.md
```diff
@@ -0,0 +1,25 @@
+# OpenTelemetry Collector Demo
```
anuraaga, line 1: Should we delete this file? It seems copied for a demo that doesn't relate to our config at all

perf_harness/docker/otel-collector-config-perf.yaml
```diff
@@ -0,0 +1,36 @@
...
+        grpc:
+
+exporters:
+  logging:
+
+# Alternatively, use jaeger_thrift_http with the settings below. In this case
+# update the list of exporters on the traces pipeline.
+#
+#  jaeger_thrift_http:
+#    url: http://jaeger-all-in-one:14268/api/traces
```
anuraaga, line 13:
```suggestion
```

perf_harness/src/main/java/io/opentelemetry/perf/OtlpPipelineDriver.java
```diff
@@ -0,0 +1,185 @@
...
+import java.io.IOException;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
+/** A testing tool. */
+public class OtlpPipelineDriver {
```
anuraaga, line 34: As a followup, testcontainers would make it easier to run the benchmark than docker-compose

anuraaga (approved):

perf_harness/docker/README.md
jkwatson: heh. yes. :)

perf_harness/src/main/java/io/opentelemetry/perf/OtlpPipelineDriver.java
jkwatson: yeah, i was thinking about that as well. it would definitely be a good follow-up.

perf_harness/docker/README.md
jkwatson: done, along with the prometheus config

PR #1741, opened 2020-10-03
Remove Status object from API, keep StatusCanonicalCode

bogdandrutu:


sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/data/ImmutableStatus.java
```diff
@@ -0,0 +1,68 @@
...
+  /**
+   * Creates a derived instance of {@code Status} with the given description.
+   *
+   * @param description the new description of the {@code Status}.
+   * @return The newly created {@code Status} with the given description.
+   * @since 0.1.0
+   */
+  public static SpanData.Status create(
+      StatusCanonicalCode canonicalCode, @Nullable String description) {
+    return new AutoValue_ImmutableStatus(canonicalCode, description);
```
jkwatson, line 46: this could just delegate to `createInternal`, so we don't have to keep two autovalue constructor calls in synch.

api/src/main/java/io/opentelemetry/trace/StatusCanonicalCode.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @since 0.1.0
+ */
+public enum StatusCanonicalCode {
+
+  /**
+   * The operation has been validated by an Application developers or Operator to have completed
+   * successfully.
+   */
+  OK(0),
```
anuraaga, line 20: The spec doesn't seem to define numeric values for the codes, do we need these?

https://www.github.com/open-telemetry/opentelemetry-specification/tree/master/specification%2Ftrace%2Fapi.md

api/src/main/java/io/opentelemetry/trace/DefaultSpan.java
```diff
@@ -77,7 +78,7 @@ public void addEvent(String name, Attributes attributes) {}
   public void addEvent(String name, Attributes attributes, long timestamp) {}
 
   @Override
-  public void setStatus(Status status) {}
+  public void setStatus(StatusCanonicalCode canonicalCode, @Nullable String description) {}
```
anuraaga, line 81: I would use two methods instead of requiring null when no description, similar to how we handle exception / link attributes.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/data/ImmutableStatus.java
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Defines the status of a {@link Span} by providing a standard {@link StatusCanonicalCode} in
+ * conjunction with an optional descriptive message. Instances of {@code Status} are created by
+ * starting with the template for the appropriate {@link StatusCanonicalCode} and supplementing it
+ * with additional information: {@code Status.NOT_FOUND.withDescription("Could not find
+ * 'important_file.txt'");}
+ */
+@AutoValue
+@Immutable
+public abstract class ImmutableStatus implements SpanData.Status {
```
anuraaga, line 24: Since we are on Java 8 now we shouldn't have to expose this class. Can you add `ok()`, etc static methods on the interface which return the values here that can be made package private?

api/src/main/java/io/opentelemetry/trace/DefaultSpan.java
bogdandrutu: I left a TODO, will do in the next PR https://github.com/open-telemetry/opentelemetry-java/pull/1741/files#diff-cc14ea515ee096dbd8ed1e630e45745fR217

api/src/main/java/io/opentelemetry/trace/StatusCanonicalCode.java
bogdandrutu: Filed https://github.com/open-telemetry/opentelemetry-java/issues/1744 to track this.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/data/ImmutableStatus.java
bogdandrutu: Done.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/data/ImmutableStatus.java
bogdandrutu: We may be able to do that, but I would do the same for ImmutableLink and ImmutableEvent then. Filed https://github.com/open-telemetry/opentelemetry-java/issues/1745 to track this.

jkwatson (approved):

anuraaga (approved):
Thanks!

Issue #1710, opened 2020-09-26
Rename AttributesKeys to AttributeKeys


trask:
Or probably you are planning to move those creation methods directly into the AttributeKey interface now with Java 8 👍

jkwatson:
oh wow. that's a major typo! Shame on me (and everyone who reviewed it and didn't catch it!) And, yes, I was also thinking of moving all of those methods to the AttributeKey now that we can. I prefer that.

imavroukakis:
Hello 😃 Will this do the trick? https://github.com/open-telemetry/opentelemetry-java/pull/1743

PR #1743, opened 2020-10-03
subsume methods from AttributesKeys to AttributeKey

imavroukakis:
Closes https://github.com/open-telemetry/opentelemetry-java/issues/1710


jkwatson (approved):
Thanks! I had intended to do this, but it slipped my mind, so much appreciated for making it happen.

imavroukakis:
> Thanks! I had intended to do this, but it slipped my mind, so much appreciated for making it happen.

You're very welcome!

api/src/main/java/io/opentelemetry/common/AttributeKey.java
```diff
@@ -25,4 +26,44 @@
   /** Returns the type of attribute for this key. Useful for building switch statements. */
   @Nonnull
   AttributeType getType();
+
+  /** Create a new AttributeKey for String valued attributes. */
```
anuraaga, line 30: Let's change these to `Returns a` while here.

anuraaga (approved):
Thanks!

api/src/main/java/io/opentelemetry/common/AttributeKey.java
imavroukakis: Done

jkwatson:
I'm going to close and open this to hopefully kick the build

imavroukakis:
> I'm going to close and open this to hopefully kick the build

If it's not too much hassle, could you do the same for https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1307 ?

jkwatson:
> > I'm going to close and open this to hopefully kick the build
> 
> If it's not too much hassle, could you do the same for [open-telemetry/opentelemetry-java-instrumentation#1307](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1307) ?

looks like that might be something different. @trask mentioned something about jfrog retention on that one.

imavroukakis:
> > > I'm going to close and open this to hopefully kick the build
> > 
> > 
> > If it's not too much hassle, could you do the same for [open-telemetry/opentelemetry-java-instrumentation#1307](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1307) ?
> 
> looks like that might be something different. @trask mentioned something about jfrog retention on that one.

Ah just saw it, thanks!

PR #1749, opened 2020-10-05
Add an override for Span.setStatus without description.

anuraaga:
@bogdandrutu I know you mentioned you would do this in a subsequent PR but I went ahead and did it now. As the Java SDK PRs have gotten faster, we need to keep up the snapshot dependency in the java-instrumentation repo more frequently, so I was trying to update. I realized that losing the ability to set a code without a description in a one-arg method is a significant regression so want to get it fixed separately / sooner.

### DIFF
--- a/api/src/test/java/io/opentelemetry/trace/DefaultSpanTest.java
+++ b/api/src/test/java/io/opentelemetry/trace/DefaultSpanTest.java
@@ -45,6 +45,7 @@ class DefaultSpanTest {
     span.addEvent("event", 0);
     span.addEvent("event", Attributes.of(booleanKey("MyBooleanAttributeKey"), true));
     span.addEvent("event", Attributes.of(booleanKey("MyBooleanAttributeKey"), true), 0);
+    span.setStatus(StatusCanonicalCode.OK);
     span.setStatus(StatusCanonicalCode.OK, null);
     span.recordException(new IllegalStateException());
     span.recordException(new IllegalStateException(), Attributes.empty());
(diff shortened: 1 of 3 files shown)


api/src/test/java/io/opentelemetry/trace/DefaultSpanTest.java
```diff
@@ -45,6 +45,7 @@ void doNotCrash() {
     span.addEvent("event", 0);
     span.addEvent("event", Attributes.of(booleanKey("MyBooleanAttributeKey"), true));
     span.addEvent("event", Attributes.of(booleanKey("MyBooleanAttributeKey"), true), 0);
+    span.setStatus(StatusCanonicalCode.OK);
```
jkwatson, line 48: should we also have one with a description?

jkwatson (approved):

api/src/test/java/io/opentelemetry/trace/DefaultSpanTest.java
anuraaga: Yeah probably - many cases of null, e.g null code, seem to be missing from this test too, so not sure what it's meant to test TBH. I'm relatively happy with the javadoc changes here but am fine with merging either this or #1754 to get a fix for the signature in!

bogdandrutu:
@anuraaga so sorry, I missed this :( and I also merge that PR. I am very sorry. Please rebase and if more changes are needed we will merge this.

jkwatson:
I think this is basically a duplicate of the one already merged, so I think we can close it.

anuraaga:
@bogdandrutu No worries - yeah will close this, and I'm sure we'll have a more broad javadoc scrub at some point :)

PR #1750, opened 2020-10-05
Add a prepareRelease task that can automatically sub the versions in the README

anuraaga:
I think we should get to a point where there is one button in a GitHub actions workflow which we press, it asks us for a version to release, and it then automatically builds, updates README, tags, and publishes. I don't want to disrupt the next release with such a change so for now just added a task that can update the README when the version is the current release version.


thisthat (approved):
Thank you for taking care of this :rocket: 
I would also add the execution of this task inside RELEASING.md so we won't forget about it :wink:

anuraaga:
I don't want to affect the next release too much - after it, I'll update RELEASING.md :)

anuraaga:
Going to send a large update to RELEASING.md soon, let me merge this in for now.

PR #1751, opened 2020-10-05
Migrate usage of gRPC context to OTel context.

### LINKED ISSUE #575: remove grpc-context dependency from the api package

anuraaga:
Merging should be held off until next release to give this code some time to bake in snapshots but sending it out now.

Fixes #575


QUICKSTART.md
```diff
@@ -167,10 +167,6 @@ For more details how to read context from remote processes, see
 
 ### Context Propagation
 
-In-process propagation leverages [gRPC Context](https://grpc.github.io/grpc-java/javadoc/io/grpc/Context.html),
-a well established context propagation library, contained in a small artifact, which is non-dependent on the
-entire gRPC engine.
-
```
jkwatson, line 173: Is it worth replacing this with a comment about our own context implementation, rather than just deleting it?

anuraaga: I wasn't sure how to phrase it. I read this sentence as justifying the external dependency because it is "a small artifact, which is non-dependent on the entire gRPC engine." for people that might worry about that, rather than anything particularly interesting. Let me know if you have any suggestion on how to replace this, I'll put it in.

jkwatson: it's fine for now. I was just wondering if people might be interested in how our in-process propagation was implemented.

jkwatson (approved):
:shipit:

tylerbenson (approved):

jkwatson:
Merging! 🎉

PR #1758, opened 2020-10-05
Fixes for BOM publishing

anuraaga:
Don't set packaging type explicitly since Gradle determines it automatically, set description for BOM project, and remove outdated bom publication configuration


anuraaga:
Confirmed published POM for the BOM has `<packaging>pom</packaging` and a description now.

jkwatson (approved):
🤞

PR #1760, opened 2020-10-05
Remove assertion on interrupted which is not reliable.

anuraaga:
Not totally sure but seems reasonable to require some time to propagate interruption to another thread and can get unflipped by the time of assertion.


anuraaga:
For https://github.com/open-telemetry/opentelemetry-java/runs/1211175353?check_suite_focus=true

jkwatson (approved):
Thanks! Let's hope your hunch is right!

PR #1761, opened 2020-10-06
Remove SpanData implementation classes from public API.

### LINKED ISSUE #1745: Should sdk/trace/data/ImmutableXYZ be package protected?
From @anuraaga 
> Since we are on Java 8 now we shouldn't have to expose this class. Can you add ok(), etc static methods on the interface which return the values here that can be made package private?

anuraaga:
Fixes #1745


jkwatson (approved):
I think this is a good change. @bogdandrutu what do you think?

bogdandrutu (approved):

Issue #1530, opened 2020-08-11
CorrelationsContext does not work from the client


rvowles:
**Is your feature request related to a problem? Please describe.**
I have come from OpenTracing and I would like to use OpenTelemetry in a complete replacement. I am advocating it for use in passing feature flag state from clients to servers so they propagate through the infrastructure (which we do for OpenTracing). However, with the Java implementation (instrumentation) this simply doesn't work. 

This for example with the Jaeger libraries and OpenTracing works fine, baggage appears:

curl -v -H "jaeger-debug-id: 11" -H "uberctx-fhub.FEATURE_TITLE_TO_UPPERCASE: true" -H 'content-type: application/json' http://localhost:8099/todo/list

With OpenTelemetry replacing it, this doesn't work:

curl -v -H 'traceparent: 00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01' -H 'otcorrelations: FEATURE_TITLE_TO_UPPERCASE=true'  -H 'content-type: application/json' http://localhost:8099/todo/list

I can see the span and parent and so forth all turning up, but the CorrelationContext is always empty. 

OpenTelemetry.getCorrelationContextManager().getCurrentContext().getEntries()

**Describe the solution you'd like**
Either this to work or some indication as to why it isn't working or I have it completely wrong and it will never work. 

**Describe alternatives you've considered**
I'm not aware apart from sticking with OpenTracing what other options there are.

**Additional context**

jkwatson:
Hi @rvowles . This definitely hasn't been implemented yet. I think it's currently held up by this spec issue:  https://github.com/open-telemetry/opentelemetry-specification/issues/536

It definitely needs to be addressed before GA, so I will tag it as such.

Thanks for highlighting this! I had forgotten this was still not done.

rvowles:
Thanks - the implication in that issue is they have settled on the header "Baggage" and are just waiting on the process to move its inexorable way through the w3c process.

jkwatson:
> Thanks - the implication in that issue is they have settled on the header "Baggage" and are just waiting on the process to move its inexorable way through the w3c process.

I think that's the case, but I haven't really been following it all that closely, TBH.

anuraaga:
[...]

PR #1764, opened 2020-10-06
Turn the Baggage metadata into a simple wrapper for a String.

jkwatson:
This is the first step toward a W3C Baggage Propagator implementation.

Amends #1530

### DIFF
--- a/api/src/main/java/io/opentelemetry/baggage/Entry.java
+++ b/api/src/main/java/io/opentelemetry/baggage/Entry.java
@@ -6,9 +6,9 @@
 package io.opentelemetry.baggage;
 
 import com.google.auto.value.AutoValue;
-import io.opentelemetry.baggage.EntryMetadata.EntryTtl;
 import io.opentelemetry.internal.StringUtils;
 import io.opentelemetry.internal.Utils;
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
 /**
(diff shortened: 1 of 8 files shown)


api/src/main/java/io/opentelemetry/baggage/Entry.java
```diff
@@ -71,11 +66,12 @@ public static Entry create(String key, String value, EntryMetadata entryMetadata
   public abstract String getValue();
 
   /**
-   * Returns the {@link EntryMetadata} associated with this {@link Entry}.
+   * Returns the (optional) {@link EntryMetadata} associated with this {@link Entry}.
```
anuraaga, line 69:
```suggestion
   * Returns the {@link EntryMetadata} associated with this {@link Entry}, or {@code null} if there is no metadata.
```

anuraaga (approved):

api/src/main/java/io/opentelemetry/baggage/Entry.java
jkwatson: I've been preferring `EntryMetadata.EMPTY` rather than null. Did I miss a spot?

anuraaga: Oh that's fine too, in that case can remove `@Nullable` I think - I just noticed the annotation but didn't look at the code

jkwatson: oh yeah. I should remove that. will do!

jkwatson: done

anuraaga (approved):

jkwatson:
Merging, since #1767 which stacks on this PR has 3 approvals. :)

PR #1765, opened 2020-10-06
Remove baggage parent from the API

jkwatson:
The spec mentions nothing about Baggage having an explicit parent attribute, so this removes it. This PR retains the implicit parent that gets amended if the Context contains Baggage when new Baggage is created.

This is a follow-on to #1764 and continues work toward having the Baggage implementation match the specification.

Amends #1530 

The next step after this will be to move the Baggage implementation up to the API, removing the no-op baggage implementation.

### DIFF
--- a/api/src/main/java/io/opentelemetry/baggage/Baggage.java
+++ b/api/src/main/java/io/opentelemetry/baggage/Baggage.java
@@ -5,7 +5,6 @@
 
 package io.opentelemetry.baggage;
 
-import io.grpc.Context;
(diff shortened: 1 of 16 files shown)


bogdandrutu:
@jkwatson not sure about this change, we should discuss this in the meeting tomorrow

jkwatson:
I'm closing this in favor of a smaller step to just remove parenting via Baggage, keeping parenting via Context like the Span equivalent.

PR #1767, opened 2020-10-07
Remove the method to parent via Baggage. 

jkwatson:
Parenting is still allowed via the Context.
Also, renamed many old variables and method to use the new baggage name.

Builds on changes in #1764

Amends #1530

### DIFF
--- a/api/src/main/java/io/opentelemetry/baggage/BaggageManager.java
+++ b/api/src/main/java/io/opentelemetry/baggage/BaggageManager.java
@@ -16,7 +16,7 @@ import javax.annotation.concurrent.ThreadSafe;
  *
  * <p>Implementations may have different constraints and are free to convert entry contexts to their
  * own subtypes. This means callers cannot assume the {@link #getCurrentBaggage() current context}
- * is the same instance as the one {@link #withContext(Baggage) placed into scope}.
+ * is the same instance as the one {@link #withBaggage(Baggage) placed into scope}.
  *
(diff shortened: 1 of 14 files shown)


api/src/main/java/io/opentelemetry/baggage/BaggageManager.java
```diff
@@ -16,7 +16,7 @@
  *
  * <p>Implementations may have different constraints and are free to convert entry contexts to their
  * own subtypes. This means callers cannot assume the {@link #getCurrentBaggage() current context}
- * is the same instance as the one {@link #withContext(Baggage) placed into scope}.
+ * is the same instance as the one {@link #withBaggage(Baggage) placed into scope}.
```
anuraaga, line 19:
```suggestion
 * is the same instance as the one {@linkplain #withBaggage(Baggage) placed into scope}.
```

anuraaga (commented):
Is this PR merged with another one?

jkwatson:
> Is this PR merged with another one?

yes, as commented in the description. :)

api/src/main/java/io/opentelemetry/baggage/BaggageManager.java
jkwatson: 👍

anuraaga (approved):
Oops missed that

bogdandrutu (approved):

carlosalberto (approved):

PR #1769, opened 2020-10-07
Update wrapper declaration in build.gradle

Oberon00:
Follow-up for #1529.

Since we also have a "Validate Gradlew Wrapper" task, I was somehow expecting this to be checked already. Since it was outdated for quite some time, maybe we can just remove the `wrapper{}` block from build.gradle? No idea what the implications would be.


jkwatson (approved):
I'm not sure how to keep this in synch automatically. It is useful if you want to regenerate the wrapper, but it isn't used aside from that.

PR #1772, opened 2020-10-08
Add package-info.java to all packages and ParamatersAreNonNullByDefault.

anuraaga:
My primary objective is adding the annotation. I tried to make relevant documentation where I could, but I'm hoping we continue to improve them as followups :)


jkwatson (approved):
❤️

PR #1773, opened 2020-10-08
Fix warnings found by idea IDE in the SDK dir.

bogdandrutu:
* Use lambdas or method references where possible
* Remove dead code
* Make members final where possible
* Use Objects.equals instead of manually implementing the same functionality

### DIFF
--- a/sdk/all/src/jmh/java/io/opentelemetry/sdk/metrics/MetricsTestOperationBuilder.java
+++ b/sdk/all/src/jmh/java/io/opentelemetry/sdk/metrics/MetricsTestOperationBuilder.java
@@ -23,102 +23,90 @@ import io.opentelemetry.metrics.Meter;
  */
 public enum MetricsTestOperationBuilder {
   LongCounterAdd(
-      new OperationBuilder() {
-        @Override
-        public Operation build(final Meter meter) {
-          return new Operation() {
-            LongCounter metric = meter.longCounterBuilder("long_counter").build();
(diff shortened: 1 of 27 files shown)


carlosalberto (approved):

jkwatson (approved):
Thanks!

jkwatson:
Looks like this needs a rebase.

sdk/all/src/jmh/java/io/opentelemetry/sdk/metrics/MetricsTestOperationBuilder.java
```diff
@@ -23,102 +23,90 @@
...
-            }
-
-            @Override
-            public void performBound() {
-              boundMetric.add(5L);
-            }
-          };
-        }
+      meter -> {
+        return new Operation() {
```
anuraaga, line 27: While we're addressing IDE warnings, what about the IDE suggestion of removing `return` too? Up to you though

anuraaga (approved):

sdk/all/src/jmh/java/io/opentelemetry/sdk/metrics/MetricsTestOperationBuilder.java
bogdandrutu: Did not find that, but I will follow-up

thisthat (approved):

anuraaga:
Thanks!

PR #1775, opened 2020-10-08
changed B3Propagator to default to inject/extra single header

### LINKED ISSUE #1684: Create a combined B3 Propagator that handle both single and multi-header cases
See the recently updated spec:  https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md#b3-requirements

jarebudev:
resolves #1684 

Wasn't sure what you had in mind for making the ```B3Propagator``` configurable so I added a builder that can configure creation of it.

Default (inject single header)
```
B3Propagator.getInstance();
```

Using builder
```
    B3Propagator.builder()
      .injectFromMultipleHeaders()
      .build();
```

### DIFF
--- a/extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/B3Propagator.java
+++ b/extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/B3Propagator.java
@@ -39,7 +39,7 @@ import javax.annotation.concurrent.Immutable;
  * OpenTelemetry.setPropagators(
  *   DefaultContextPropagators
  *     .builder()
- *     .addTextMapPropagator(B3Propagator.builder().injectFromMultipleHeaders().build())
+ *     .addTextMapPropagator(B3Propagator.builder().injectMultipleHeaders().build())
  *     .build());
  * }</pre>
  */
@@ -89,7 +89,7 @@ public class B3Propagator implements TextMapPropagator {
       injectSingleHeader = true;
     }
 
-    public Builder injectFromMultipleHeaders() {
+    public Builder injectMultipleHeaders() {
       this.injectSingleHeader = false;
       return this;
     }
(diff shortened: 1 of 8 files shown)


extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/B3PropagatorExtractor.java
```diff
@@ -7,10 +7,12 @@
 
 import io.grpc.Context;
 import io.opentelemetry.context.propagation.TextMapPropagator;
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
 @Immutable
 interface B3PropagatorExtractor {
 
+  @Nullable
```
jkwatson, line 16: Any thoughts about using `Optional<Context>` as the return type, now that we have java 8?

jarebudev: Nice! I'm glad Java 8 is supported now :)
Yeh returning an ```Optional``` would be better than null.  I'll update it.

jkwatson (approved):
Thanks!

thisthat (approved):
Thank you :)

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/B3Propagator.java
```diff
@@ -31,44 +58,54 @@
...
-  public static B3Propagator getMultipleHeaderPropagator() {
-    return MULTI_HEADER;
+  public static class Builder {
+    private boolean injectSingleHeader;
+
+    private Builder() {
+      injectSingleHeader = true;
+    }
+
+    public Builder injectFromMultipleHeaders() {
```
anuraaga, line 92: Inject happens to headers, not from. But how about just `injectMultipleHeaders`?

anuraaga (approved):
Just nit thanks!

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/B3Propagator.java
jarebudev: sounds good, i'll tidy it up

PR #1776, opened 2020-10-09
Add a few missing package-info.java

anuraaga:
Noticed I missed a few in the API artifact. A quick Gradle search didn't come up with any plugins for automatically validating there is a package-info.java, but may or may not add some tooling for that eventually.


jkwatson (approved):
Thanks!

PR #1781, opened 2020-10-12
Add GitHub workflows to automate releases

anuraaga:
This makes three major changes

- Use nebula-release-plugin for determining version number of project. Mainly to allow SNAPSHOT version to automatically be set to the minor version after the previous release tag on master. This is same pattern as instrumentation repo.
- Add a release workflow which builds artifacts for a version specified by the releaser, updates README, tags master with the released version
- Add a patch release workflow that first creates a release branch if none exists yet, cherrypicks provided commits, and builds artifacts for provided version, tags release branch, updates README in master

Because SNAPSHOT versions are based on the previous tag on master, there is a chicken/egg problem where our current tag is on a release branch and not picked up. So until the next release, I explicitly set the snapshot version as well.

@iNikem will remember noticing it's hard to "skip a patch version" in the instrumentation repo release process. Here I have the releaser specify a version manually, I think this control is good and can allow RC releases, etc flexibility and is not really any harder than picking between "minor" and "patch" scope.

Hope this makes future releases easier :)

Sample runs

- Normal release - https://github.com/anuraaga/opentelemetry-java/actions/runs/301530974
- First patch release, creates release branch too - https://github.com/anuraaga/opentelemetry-java/actions/runs/301554274
- Second patch release, uses existing release branch - https://github.com/anuraaga/opentelemetry-java/actions/runs/301579635

### DIFF
--- a/.github/workflows/patch-release-build.yml
+++ b/.github/workflows/patch-release-build.yml
@@ -1,3 +1,5 @@
+# Releases a patch by cherrypicking commits into a release branch based on the previous
+# release tag.
 name: Patch Release Build
 on:
   workflow_dispatch:
@@ -76,9 +78,8 @@ jobs:
           remote-build-cache-proxy-enabled: false
           arguments: final --stacktrace -Prelease.version=${{ github.event.inputs.version }}
           properties: |
-            enable.docker.tests=trued
+            enable.docker.tests=true
         env:
-          CI: true
           BINTRAY_USER: ${{ secrets.BINTRAY_USER }}
           BINTRAY_KEY: ${{ secrets.BINTRAY_KEY }}
           GRGIT_USER: ${{ github.actor }}
--- a/.github/workflows/patch-release-build.yml
+++ b/.github/workflows/patch-release-build.yml
@@ -1,0 +1,99 @@
+name: Patch Release Build
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: The version to tag the release with, e.g., 1.2.1, 1.2.2
+        required: true
+      commits:
+        description: Comma separated list of commit shas to cherrypick
+        required: true
+
+jobs:
+  prepare-release-branch:
+    runs-on: ubuntu-latest
+    outputs:
+      release-branch-name: ${{ steps.parse-release-branch.outputs.release-branch-name }}
+    steps:
+      - id: parse-release-branch
+        name: Parse release branch name
+        run: |
+          # Sets the release-branch-name output to the version number with the last non-period element replaced with an 'x' and preprended with v.
+          echo "::set-output name=release-branch-name::$(echo '${{ github.event.inputs.version }}' | sed -E 's/([^.]+)\.([^.]+)\.([^.]+)/v\1.\2.x/')"
+          # Sets the release-tag-name output to the version number with the last non-period element replace with a '0' and prepended with v
+          echo "::set-output name=release-tag-name::$(echo '${{ github.event.inputs.version }}' | sed -E 's/([^.]+)\.([^.]+)\.([^.]+)/v\1.\2.0/')"
+      - id: checkout-release-branch
+        name: Check out release branch
+        continue-on-error: true
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ steps.parse-release-branch.outputs.release-branch-name }}
+      - id: checkout-release-tag
+        name: Check out release tag
+        if: ${{ steps.checkout-release-branch.outcome == 'failure' }}
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ steps.parse-release-branch.outputs.release-tag-name }}
+      - name: Create release branch
+        if: ${{ steps.checkout-release-tag.outcome == 'success' }}
+        run: |
+          git checkout -b ${{ steps.parse-release-branch.outputs.release-branch-name }}
+          git push --set-upstream origin ${{ steps.parse-release-branch.outputs.release-branch-name }}
+  build:
+    runs-on: ubuntu-latest
+    needs: prepare-release-branch
+    steps:
+      - name: Checkout release branch
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ needs.prepare-release-branch.outputs.release-branch-name }}
+          submodules: true
+      - uses: actions/setup-java@v1
+        with:
+          java-version: 11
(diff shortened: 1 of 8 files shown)


.github/workflows/patch-release-build.yml
```diff
@@ -0,0 +1,99 @@
...
+          properties: |
+            enable.docker.tests=true
+      - name: Publish artifacts
+        uses: burrunan/gradle-cache-action@v1.5
+        with:
+          job-id: jdk11
+          remote-build-cache-proxy-enabled: false
+          arguments: final --stacktrace -Prelease.version=${{ github.event.inputs.version }}
+          properties: |
+            enable.docker.tests=trued
```
iNikem, line 79:
```suggestion
            enable.docker.tests=true
```

.github/workflows/patch-release-build.yml
```diff
@@ -0,0 +1,99 @@
...
+        env:
+          CI: true
```
iNikem, line 81: Isn't this set by GHA?

.github/workflows/patch-release-build.yml
```diff
@@ -0,0 +1,99 @@
...
+      - run: git push
+      - name: Checkout master
+        uses: actions/checkout@v2
+        with:
+           submodules: true
+      - uses: burrunan/gradle-cache-action@v1.5
+        with:
+          job-id: jdk11
+          remote-build-cache-proxy-enabled: false
+          arguments: prepareRelease --stacktrace -Prelease.version=${{ github.event.inputs.version }}
```
iNikem, line 95: What does this do after `final` task above?

.github/workflows/release-build.yml
```diff
@@ -0,0 +1,42 @@
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+      - uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - uses: burrunan/gradle-cache-action@v1.5
+        with:
+          job-id: jdk8
```
iNikem, line 22:
```suggestion
          job-id: jdk11
```

.github/workflows/release-build.yml
```diff
@@ -0,0 +1,42 @@
+name: Release Build
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: The version to tag the release with, e.g., 1.2.0, 1.2.1-alpha.1
```
iNikem, line 6: This releases current HEAD of the main branch, correct?

Makefile
```diff
@@ -24,7 +24,8 @@ verify-format:
 .PHONY: publish-snapshots
 publish-snapshots:
 ifeq ($(CIRCLE_BRANCH),master)
-	./gradlew artifactoryPublish
+  # TODO(anuraaga): Remove version specificer after next release creates a tag on master.
+	./gradlew artifactoryPublish -Prelease.version=0.10.0-SNAPSHOT
```
iNikem, line 28: Cannot you just create a correct tag on the main branch for the current released version?

build.gradle
```diff
@@ -11,15 +14,35 @@ ext {
...
-    version = "0.10.0-SNAPSHOT" // CURRENT_OPEN_TELEMETRY_VERSION
 
     plugins.withId("maven-publish") {
         // Always include the artifactory/bintray plugins to do the deployment.
         pluginManager.apply "com.jfrog.artifactory"
         pluginManager.apply "com.jfrog.bintray"
 
+        releaseTask.configure {
+            if (version.toString().endsWith('-SNAPSHOT')) {
+                finalizedBy(tasks.named('artifactoryPublish'))
```
iNikem, line 40: This will run them even if `releaseTask` fails, are you sure?

.github/workflows/patch-release-build.yml
anuraaga: Yeah I should just rename to `updateReadme` instead of `prepareRelease`, I didn't notice this use case of needing to do it on a different branch from the release at the time. Should I do it in this PR?

.github/workflows/release-build.yml
anuraaga: Yup, added a comment.

Makefile
anuraaga: Well, we already have a tag referring to a release branch which is what we built from. While moving it to the main branch could be effectively the same, I think it's some rewriting history that I want to avoid, this workaround isn't so bad I think.

build.gradle
anuraaga: `releaseTask` is just a placeholder, so it can't fail. I think it's ok - it's the simplest way to have publishing guaranteed to happen after the dependencies of `releaseTask` instead of wiring dependencies if needed.

.github/workflows/patch-release-build.yml
iNikem: I think so

iNikem (approved):

.github/workflows/patch-release-build.yml
```diff
@@ -0,0 +1,100 @@
...
+# release tag.
+name: Patch Release Build
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: The version to tag the release with, e.g., 1.2.1, 1.2.2
+        required: true
+      commits:
+        description: Comma separated list of commit shas to cherrypick
```
jkwatson, line 11: In case we need a patch that can't be done with a simple cherrypick, do we have a description of what that process would look like? An example would be a bugfix change that was done against the main branch, but where the main branch APIs have changed from what was on the release branch.

jkwatson:
This seems like an enormous improvement over the current release process. @anuraaga I think, once this is merged, that we should do an 0.9.2 release to test that it works. What do you think?

.github/workflows/patch-release-build.yml
anuraaga: Great callout - I added docs for this case.

anuraaga:
@jkwatson I think it makes sense - though I guess we could only verify the patch release? 😄

jkwatson:
> @jkwatson I think it makes sense - though I guess we could only verify the patch release? 😄

Hrm. Yeah, well, I guess it's only 3 weeks until 0.10.0 is ready to go, so we won't have to wait too long to find all the bugs. 😬

jkwatson (approved):
🚀 🎸

PR #1782, opened 2020-10-12
Exclude jaeger proto classes from coverage.

anuraaga:
I noticed reports with low coverage had these protos included, which have almost no coverage. I'm not sure why sometimes they're included in the report and sometimes not, but I could reliably see reports locally have lots of red for this folder before the change and it's gone after.


thisthat (approved):
thank you :)

jkwatson (approved):

PR #1783, opened 2020-10-12
Add constants for message operations which are defined by spec.

anuraaga:


anuraaga:
I think this PR demonstrates that we need to split `SemanticAttributes` into multiple classes? Reasons

- attributes.setAttribute(MESSAGING_OPERATION, MESSAGING_OPERATION_SEND) is stuttery
- I don't see any way to automatically generate the current class based on the yaml files in a nice way

iNikem (approved):

thisthat (approved):

thisthat:
> * I don't see any way to automatically generate the current class based on the yaml files in a nice way

I'll open a draft PR to show how I would address the code generation. My idea is to provide an overload that accepts also enumerations as attribute values.

jkwatson:
@anuraaga Yes, I'd like to have the *values* moved into their own class. Maybe SemanticConventionValues or some such.

jkwatson (approved):

PR #1789, opened 2020-10-12
Make the RERS return isRecording=false after it has ended.

### LINKED ISSUE #1784: Make a Span return isRecording=false after it has ended
See this recent spec change: https://github.com/open-telemetry/opentelemetry-specification/pull/1011

jkwatson:
Resolves #1784

### DIFF
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
@@ -461,7 +461,9 @@ final class RecordEventsReadableSpan implements ReadWriteSpan {
 
   @Override
   public boolean isRecording() {
-    return true;
+    synchronized (lock) {
+      return !hasEnded;
+    }
   }
 
   @GuardedBy("lock")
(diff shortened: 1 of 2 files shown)


sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -461,7 +461,9 @@ public SpanContext getContext() {
 
   @Override
   public boolean isRecording() {
-    return true;
+    synchronized (lock) {
+      return !hasEnded;
```
anuraaga, line 465: What do you think about switching this to volatile? Sort of nice that something seemingly trivial like checking recording state doesn't have to lock.

anuraaga (approved):

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
jkwatson: I'm easy either way. I think it's worth looking into the entire thread-safety management of this class, but let's save that for another PR?

PR #1790, opened 2020-10-13
Rename newBuilder to builder for consistency.

anuraaga:
We currently are not consistent with whether we use `newBuilder` or `builder` for a builder factory. This consolidates on `builder`, similar to Guava and slightly recommended by AutoValue.


jkwatson (approved):
Thanks! Consistency is important in the API to reduce confusion.

carlosalberto (approved):

PR #1791, opened 2020-10-13
Remove DefaultSpan from public API.

### LINKED ISSUE #1704: Update DefaultSpan to match the new spec

anuraaga:
We use `DefaultSpan`, a no-op wrapper of a spancontext, for three use cases

- propagated span
- unsampled span
- disabling implicit tracing in a scope

So it actually didn't make sense to use the propagated span terminology to replace it. I used a `wrapAsNoop` factory instead since that's literally what we're doing. ~~To give the propagated term some love, I renamed the relevant `SpanContext` factory to use the word.~~ Realized I was confusing propagated span with remote span but they are different.

Would be much easier to reason about without the `SpanContext` I guess 😋 

Fixes #1704

### DIFF
--- a/api/src/main/java/io/opentelemetry/trace/DefaultTracer.java
+++ b/api/src/main/java/io/opentelemetry/trace/DefaultTracer.java
@@ -64,9 +64,7 @@ public final class DefaultTracer implements Tracer {
         spanContext = TracingContextUtils.getCurrentSpan().getContext();
       }
 
-      return spanContext != null && !SpanContext.getInvalid().equals(spanContext)
-          ? new DefaultSpan(spanContext)
-          : DefaultSpan.getInvalid();
+      return Span.getPropagated(spanContext);
     }
 
     @Override
--- a/api/src/main/java/io/opentelemetry/trace/DefaultTracer.java
+++ b/api/src/main/java/io/opentelemetry/trace/DefaultTracer.java
@@ -59,7 +59,7 @@ public final class DefaultTracer implements Tracer {
         spanContext = TracingContextUtils.getCurrentSpan().getContext();
       }
 
-      return Span.getPropagated(spanContext);
+      return Span.wrap(spanContext);
     }
 
     @Override
--- a/api/src/main/java/io/opentelemetry/trace/Span.java
+++ b/api/src/main/java/io/opentelemetry/trace/Span.java
@@ -24,6 +24,26 @@ import javax.annotation.concurrent.ThreadSafe;
 public interface Span {
 
   /**
+   * Returns an invalid {@link Span}. An invalid {@link Span} is used when tracing is disabled,
+   * usually because there is no OpenTelemetry SDK installed.
+   */
+  static Span getInvalid() {
+    return PropagatedSpan.INVALID;
+  }
+
+  /**
+   * Returns a non-recording {@link Span} that holds the provided {@link SpanContext} but has no
+   * functionality. It will not be exported and all tracing operations are no-op, but it can be used
+   * to propagate a valid {@link SpanContext} downstream.
+   */
+  static Span getPropagated(SpanContext spanContext) {
+    if (spanContext == null || !spanContext.isValid()) {
+      return getInvalid();
+    }
+    return new PropagatedSpan(spanContext);
+  }
+
+  /**
    * Type of span. Can be used to specify additional relationships between spans in addition to a
    * parent/child relationship.
    *
@@ -303,6 +323,15 @@ public interface Span {
    * @since 0.1.0
    */
   boolean isRecording();
+
+  /**
+   * Returns whether this {@link Span} is valid.
+   *
+   * @see Span#getInvalid()
+   */
+  default boolean isValid() {
+    return getContext().isValid();
+  }
 
   /**
    * {@link Builder} is used to construct {@link Span} instances which define arbitrary scopes of
--- a/api/src/main/java/io/opentelemetry/trace/Span.java
+++ b/api/src/main/java/io/opentelemetry/trace/Span.java
@@ -34,7 +34,7 @@ public interface Span {
    * functionality. It will not be exported and all tracing operations are no-op, but it can be used
    * to propagate a valid {@link SpanContext} downstream.
    */
-  static Span getPropagated(SpanContext spanContext) {
+  static Span wrap(SpanContext spanContext) {
     if (spanContext == null || !spanContext.isValid()) {
       return getInvalid();
     }
@@ -285,15 +285,6 @@ public interface Span {
    * @return {@code true} if this {@code Span} records tracing events.
    */
   boolean isRecording();
-
-  /**
-   * Returns whether this {@link Span} is valid.
-   *
-   * @see Span#getInvalid()
-   */
-  default boolean isValid() {
-    return getContext().isValid();
-  }
 
   /**
    * {@link Builder} is used to construct {@link Span} instances which define arbitrary scopes of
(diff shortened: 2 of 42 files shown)


api/src/main/java/io/opentelemetry/trace/Span.java
```diff
@@ -23,6 +23,26 @@
...
+  static Span getInvalid() {
+    return DefaultSpan.INVALID;
+  }
+
+  /**
+   * Returns a {@link Span} that only wraps the provided {@link SpanContext} but has no
+   * functionality. It will not be exported and all operations are no-op, but it can be used to
+   * propagate the {@link SpanContext} downstream.
+   */
+  static Span wrapWithNoop(SpanContext spanContext) {
```
anuraaga, line 39: @Oberon00 Based on the conversation in https://github.com/open-telemetry/opentelemetry-specification/issues/1086 this should be something like `wrapForPropagation`, we're wrapping this `SpanContext` into a `PropagatedSpan`. This seems like odd naming to me though, as there is nothing about propagation happening here really, doubly-so because propagation is still more commonly used for remote propagation than in-process propagation I think. That's not great here since we don't do remote propagation of the invalid span.

Oberon00: >  as there is nothing about propagation happening here really

If one considers making a span a local parent "propagation" (which I think one should), then it makes sense. Because what other uses would this returned object have?

anuraaga: > Because what other uses would this returned object have?

I dunno maybe someone will have something creative :) While it's indirect usage of this API, one case where it will be used is when extracting the `SpanContext` from a link - the result of this method will be taken out of the context, the link will be made, and the result will be dropped without ever being propagated. So this is where the disconnect with the naming comes from - since it's completely up to the caller on whether it's propagated or not, so I find it pretty confusing. There's the other point where there's no such thing as a non-propagated span - all spans can be propagated, so why do we have the propagated span called out in particular? I'm sure someone is going to call `Span.getPropagated(tracer.startSpan().getContext())` assuming that's how you're supposed to propagate your span ;) `Span.getNoop(tracer.startSpan().getContext())`, `Span.getNonRecording(...)` seems more obviously wrong.

jkwatson: I don't love this name, but I 100% agree with the sentiment. Making it super clear what the capabilities (and lack thereof) will be of the returned Span seems like a very important signal to the user.

api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java
```diff
@@ -144,44 +144,45 @@ public static HttpTraceContext getInstance() {
...
+    if (!span.isValid()) {
       return context;
     }
 
-    return TracingContextUtils.withSpan(DefaultSpan.create(spanContext), context);
+    return TracingContextUtils.withSpan(span, context);
   }
 
-  private static <C> SpanContext extractImpl(C carrier, Getter<C> getter) {
+  private static <C> Span extractImpl(C carrier, Getter<C> getter) {
```
jkwatson, line 155: This change seems a bit orthogonal to the PR. Does it need to be included?

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/AwsXRayPropagator.java
```diff
@@ -123,18 +122,18 @@ public static AwsXRayPropagator getInstance() {
     Objects.requireNonNull(carrier, "carrier");
     Objects.requireNonNull(getter, "getter");
 
-    SpanContext spanContext = getSpanContextFromHeader(carrier, getter);
-    if (!spanContext.isValid()) {
+    Span span = extractSpan(carrier, getter);
```
jkwatson, line 125: I'd like to separate out this changes into a follow-on PR, just to keep this one a little more tight in its scope.

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/B3PropagatorExtractorMultipleHeaders.java
```diff
@@ -29,33 +28,33 @@
...
+    if (!span.isValid()) {
       return Optional.empty();
     }
 
-    return Optional.of(TracingContextUtils.withSpan(DefaultSpan.create(spanContext), context));
+    return Optional.of(TracingContextUtils.withSpan(span, context));
   }
 
-  private static <C> SpanContext getSpanContextFromMultipleHeaders(
+  private static <C> Span extractFromMultipleHeaders(
```
jkwatson, line 39: ditto on reducing scope for this PR

anuraaga:
@jkwatson Good call - I also went ahead and named the method `getPropagated` to stick with the spec for now. Hoping will be able to rename it later, it's just 10secs in IntelliJ to rename :)

anuraaga:
Had a scary unexpected force push but branch seems to be ok, phew

api/src/main/java/io/opentelemetry/trace/DefaultTracer.java
```diff
@@ -64,9 +64,7 @@ public Span startSpan() {
         spanContext = TracingContextUtils.getCurrentSpan().getContext();
       }
 
-      return spanContext != null && !SpanContext.getInvalid().equals(spanContext)
-          ? new DefaultSpan(spanContext)
-          : DefaultSpan.getInvalid();
+      return Span.getPropagated(spanContext);
```
jkwatson, line 67: this is a nice cleanup. much tighter and clearer.

api/src/main/java/io/opentelemetry/trace/Span.java
```diff
@@ -23,6 +23,26 @@
...
+  static Span getInvalid() {
+    return PropagatedSpan.INVALID;
+  }
+
+  /**
+   * Returns a non-recording {@link Span} that holds the provided {@link SpanContext} but has no
+   * functionality. It will not be exported and all tracing operations are no-op, but it can be used
+   * to propagate a valid {@link SpanContext} downstream.
+   */
+  static Span getPropagated(SpanContext spanContext) {
```
bogdandrutu, line 39: s/getPropagated/toPropagated?

api/src/main/java/io/opentelemetry/trace/Span.java
```diff
@@ -304,6 +324,15 @@ default void setAttribute(AttributeKey<Long> key, int value) {
...
   boolean isRecording();
 
+  /**
+   * Returns whether this {@link Span} is valid.
+   *
+   * @see Span#getInvalid()
+   */
+  default boolean isValid() {
+    return getContext().isValid();
+  }
```
bogdandrutu, line 334: I think this change is unrelated with the PR description and purpose, consider to not add "secondary" changes to a PR.

bogdandrutu (changes requested):
I like the overall change of "PropagationSpan" but I don't like that we add other changes to the API like the "isValid" in the same PR. That change is independent and can be added in a separate PR.

api/src/main/java/io/opentelemetry/trace/Span.java
Oberon00: I think this should be called `Span.wrap(SpanContext)`. Then we are also safe against renamings of propagated/noop/etc

bogdandrutu: I don't think we are safe against behavior change, this needs to always propagate the SC. But +1 for the name

api/src/main/java/io/opentelemetry/trace/Span.java
anuraaga: Sure makes sense

bogdandrutu (approved):

PR #1792, opened 2020-10-13
Add Context arg to SpanProcessor.OnStart.

### LINKED ISSUE #1785: Update SpanProcessor.OnStart to receive parent Context.
See spec at https://github.com/open-telemetry/opentelemetry-specification/blob/7e0c10260178b947777cf9c11d7f7ac4d4700725/specification/trace/sdk.md#onstart

> parentContext - the parent Context of the span that the SDK determined (the explicitly passed Context, the current Context or an empty Context if that was explicitly requested).

This was introduced in https://github.com/open-telemetry/opentelemetry-specification/pull/875 but was not handled in #1611.

Current signature (at time of issue creation): https://github.com/open-telemetry/opentelemetry-java/blob/2d7f7f05d9b3ac41a4169ec7bc277a9729eaf45d/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/SpanProcessor.java#L25

Oberon00:
Fixes #1785.

### DIFF
--- a/sdk_extensions/async_processor/src/main/java/io/opentelemetry/sdk/extensions/trace/export/DisruptorEventQueue.java
+++ b/sdk_extensions/async_processor/src/main/java/io/opentelemetry/sdk/extensions/trace/export/DisruptorEventQueue.java
@@ -12,11 +12,14 @@ import com.lmax.disruptor.RingBuffer;
 import com.lmax.disruptor.WaitStrategy;
 import com.lmax.disruptor.dsl.Disruptor;
 import com.lmax.disruptor.dsl.ProducerType;
+import io.opentelemetry.context.Context;
 import io.opentelemetry.sdk.common.CompletableResultCode;
 import io.opentelemetry.sdk.common.DaemonThreadFactory;
 import io.opentelemetry.sdk.trace.ReadWriteSpan;
 import io.opentelemetry.sdk.trace.ReadableSpan;
 import io.opentelemetry.sdk.trace.SpanProcessor;
+import java.util.AbstractMap;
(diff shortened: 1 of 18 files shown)


sdk_extensions/async_processor/src/main/java/io/opentelemetry/sdk/extensions/trace/export/DisruptorEventQueue.java
```diff
@@ -84,14 +87,14 @@ public DisruptorEvent newInstance() {
...
-  void enqueueStartEvent(ReadWriteSpan span) {
+  void enqueueStartEvent(ReadWriteSpan span, Context parentContext) {
     if (isShutdown) {
       if (!loggedShutdownMessage.getAndSet(true)) {
         logger.info("Attempted to enqueue start event after Disruptor shutdown.");
       }
       return;
     }
-    enqueue(EventType.ON_START, span, null);
+    enqueue(EventType.ON_START, new AbstractMap.SimpleImmutableEntry<>(span, parentContext), null);
```
Oberon00, line 97: This seems a bit dirty, so if someone can suggest a better solution that would be good.

jkwatson: I think it's fine for now, although agree it's ugly. I would like to revisit the way the disruptor messages are built at some point, since I ran into difficulties when trying to work with it a few months ago, as well. I think the best answer might be to create a Tuple to hold the message data, or maybe split the message types so they can be more easily customized.

anuraaga (approved):
Thanks!

PR #1795, opened 2020-10-14
Move static access of current span to Span.

anuraaga:
I was thinking of talking about this after changes like #1753 and #1737 (after which I'm thinking of moving context interaction to `OpenTelemetry`) but figured it doesn't hurt to split this out as an idea.

I think getting the current span is a very common operation, especially for javaagent users. They will rarely create spans, or deal with tracers. But adding attributes to a span is very common for anyone.

```java
Response myRequestHandler(Request request) {
  Span.current().setAttribute("userid", request.getUserId());
}
```

We are also toying with the idea of having static methods on `Tracer`
```java
Response myRequestHandler(Request request) {
  Tracer.currentSpan().setAttribute("userid", request.getUserId());
}
```

For simple usage, which is common for accessing the current span since when you have a tracer, you usually also will have created a span from it to track yourself, the former feels more intuitive and looks like better code from a user perspective to me. Tracer is a very important concept for instrumentation authors, but not so important for end users.

### DIFF
--- a/api/src/main/java/io/opentelemetry/trace/DefaultTracer.java
+++ b/api/src/main/java/io/opentelemetry/trace/DefaultTracer.java
@@ -35,7 +35,7 @@ public final class DefaultTracer implements Tracer {
 
   @Override
   public Span getCurrentSpan() {
-    return TracingContextUtils.getCurrentSpan();
+    return Span.current();
   }
 
   @Override
@@ -61,7 +61,7 @@ public final class DefaultTracer implements Tracer {
     @Override
     public Span startSpan() {
       if (spanContext == null) {
-        spanContext = TracingContextUtils.getCurrentSpan().getContext();
+        spanContext = Span.current().getContext();
       }
 
       return spanContext != null && !SpanContext.getInvalid().equals(spanContext)
--- a/api/src/main/java/io/opentelemetry/trace/TracingContextUtils.java
+++ b/api/src/main/java/io/opentelemetry/trace/TracingContextUtils.java
@@ -33,14 +33,7 @@ public final class TracingContextUtils {
     return context.withValues(CONTEXT_SPAN_KEY, span);
   }
 
-  /**
-   * Returns the {@link Span} from the current {@code Context}, falling back to a default, no-op
-   * {@link Span}.
-   *
-   * @return the {@link Span} from the current {@code Context}.
-   * @since 0.3.0
-   */
-  public static Span getCurrentSpan() {
+  static Span getCurrentSpan() {
     return getSpan(io.opentelemetry.context.Context.current());
   }
 
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/TracerSdk.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/TracerSdk.java
@@ -24,7 +24,7 @@ final class TracerSdk implements Tracer {
 
   @Override
   public Span getCurrentSpan() {
-    return TracingContextUtils.getCurrentSpan();
+    return Span.current();
   }
 
   @Override
--- a/api/src/main/java/io/opentelemetry/trace/Span.java
+++ b/api/src/main/java/io/opentelemetry/trace/Span.java
@@ -24,6 +24,17 @@ import javax.annotation.concurrent.ThreadSafe;
 public interface Span {
 
   /**
+   * Returns the {@link Span} from the current {@code Context}, falling back to a default, no-op
+   * {@link Span} if there is no current span in the context. This is generally used to access a
+   * {@link Span} that you did not create, for example one created by instrumentation of a request
+   * handler. There is some overhead in accessing the current span statically, so if you have
+   * created the {@link Span} yourself it is recommended to maintain a reference to it instead.
+   */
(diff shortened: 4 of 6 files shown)


iNikem:
It seems to contradict [this spec issue](https://github.com/open-telemetry/opentelemetry-specification/issues/1019)

> Remove all current public helpers from Tracer and Trace Context Utilities that interact with "current Context".
Things like Get the currently active span and Set the currently active span should be removed.

Oberon00:
I think that is a spec wording issue and should be fixed there. IMHO the spec just wants these methods to not require a Span/Tracer instance as parameters (e.g. implicit `this` parameter)

api/src/main/java/io/opentelemetry/trace/TracingContextUtils.java
```diff
@@ -33,14 +33,7 @@ public static Context withSpan(Span span, Context context) {
...
-  /**
-   * Returns the {@link Span} from the current {@code Context}, falling back to a default, no-op
-   * {@link Span}.
-   *
-   * @return the {@link Span} from the current {@code Context}.
-   * @since 0.3.0
-   */
-  public static Span getCurrentSpan() {
+  static Span getCurrentSpan() {
     return getSpan(io.opentelemetry.context.Context.current());
```
jkwatson, line 37: odd fully-qualified class here. might cleaning it up while you're here?

carlosalberto:
> I think that is a spec wording issue and should be fixed there

No, we also want to reduce the places where we expose this functionality.

That being said, whether we keep this functionality (`Tracer` or `TracingContextUtils`), I'd *personally* be fine with `Span.current()` as a **single** shortcut (that is, no more shortcuts, at least for now). Else, maybe a reason to have `Tracer.getCurrentSpan()` as a static member.

jkwatson (approved):
I'm definitely not a fan of TCU, so this seems like a good change to me. Would love @carlosalberto to comment, as well.

Oberon00:
> No, we also want to reduce the places where we expose this functionality.

But this is not adding more places?

carlosalberto:
> But this is not adding more places?

Yes! And that's why I'd be fine with a *single* shortcut (or, well, none). The idea is that Spec change is to not have *all* duplicated (at this moment we have *many* operations all over the place).

jkwatson:
We had some discussion about this during the SIG meeting this morning. The request from @bogdandrutu , which I think everyone agreed with, is possibly an end-state-describing issue that can track all the work on this cleanup, so we can have a good handle on where you/we are headed. For instance, we all like having `Span.current()`, but many of us also think that if we add it, we should remove at least one of the other public options for accessing the current span.

So, the ask is for a little bit of design work to describe where you see the end-state of the API (this might tie into the instance/global work, additionally, if you feel they are related).

anuraaga:
Thanks, summarized in #1807

api/src/main/java/io/opentelemetry/trace/DefaultTracer.java
```diff
@@ -35,7 +35,7 @@ public static Tracer getInstance() {
 
   @Override
   public Span getCurrentSpan() {
-    return TracingContextUtils.getCurrentSpan();
+    return Span.current();
```
carlosalberto, line 38: Given that `TracingContextUtils` _still_ is the root, I'd be happier with using that one instead (users and instrumentation are free to use whatever they want of course ;) )

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/TracerSdk.java
```diff
@@ -24,7 +24,7 @@
 
   @Override
   public Span getCurrentSpan() {
-    return TracingContextUtils.getCurrentSpan();
```
carlosalberto, line 27: Likewise.

api/src/main/java/io/opentelemetry/trace/DefaultTracer.java
jkwatson: this is gone from the Tracer now, so probably not relevant. :)

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/TracerSdk.java
jkwatson: same

api/src/main/java/io/opentelemetry/trace/DefaultTracer.java
carlosalberto: Disagree - we want to use whatever the root is. Otherwise, let's simply move the logic to `Span` ;) (although the Spec may not allow that).

jkwatson: I'm just saying that this method is no longer on the Tracer on the main branch, so this comment can't apply without adding it back to the Tracer.

bogdandrutu:
Proposing for the moment to close this PR and update or create a new PR with the latest discussion from https://github.com/open-telemetry/opentelemetry-java/issues/1807

carlosalberto:
@bogdandrutu Fine by me to close this PR while we discuss our options.

PR #1797, opened 2020-10-14
Publish -context package

anuraaga:
Forgot to add the line of magic, noticed when trying to depend on the snapshots.


jkwatson (approved):
Thanks!

PR #1798, opened 2020-10-14
Add .gitattributes.

Oberon00:
This fixes `spotlessCheck` when checking out on windows with git `core.eol=crlf` and `core.autocrlf=true`.

Note that for git itself, the eol=lf next to the text=auto changes nothing, but spotless tries to
enforce the core.eol line endings anyway.

Use https://github.com/open-telemetry/opentelemetry-java/pull/1798/files?w=1 (note the ?w=1) to verify that only whitespace (line endings) is changed in the batch files.


anuraaga (commented):
Yeah I think we need a gitattributes file I always have issues on windows without it.

@Oberon00 What do you think about a simpler one like we use in instrumentation repo?

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/.gitattributes

trask:
+1 to having the same `.gitattributes` file across both repos (if there is something you think we are missing in instrumentation repo we can update there)

Oberon00:
The simpler one should also be fine.

Oberon00:
That said, I think at least the `-text` should be kept for safety.

anuraaga (approved):
Thanks!

jkwatson:
closing and reopening for the CLA bot

PR #1800, opened 2020-10-14
Update Jaeger exporter to map Resources to process tags

### LINKED ISSUE #1609: Update Jaeger exporter to map Resources to process tags
See https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk_exporters/jaeger.md#resource

malafeev:
fix #1609

### DIFF
--- a/exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/Adapter.java
+++ b/exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/Adapter.java
@@ -16,17 +16,14 @@ import io.opentelemetry.common.AttributeKey;
 import io.opentelemetry.common.ReadableAttributes;
 import io.opentelemetry.exporters.jaeger.proto.api_v2.Model;
 import io.opentelemetry.sdk.extensions.otproto.TraceProtoUtils;
-import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.sdk.trace.data.SpanData;
 import io.opentelemetry.sdk.trace.data.SpanData.Event;
 import io.opentelemetry.sdk.trace.data.SpanData.Link;
 import io.opentelemetry.trace.SpanId;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
-import java.util.Map;
 import javax.annotation.concurrent.ThreadSafe;
 
 /** Adapts OpenTelemetry objects to Jaeger objects. */
@@ -41,22 +38,6 @@ final class Adapter {
   static final String KEY_INSTRUMENTATION_LIBRARY_VERSION = "otel.library.version";
 
   private Adapter() {}
-
-  /**
-   * Groups {@link SpanData}'s by {@link Resource}.
-   *
-   * @param spans the list of spans to be grouped
-   * @return the map of grouped spans
-   */
-  static Map<Resource, List<SpanData>> groupByResource(Collection<SpanData> spans) {
-    Map<Resource, List<SpanData>> result = new HashMap<>();
-    for (SpanData spanData : spans) {
-      Resource resource = spanData.getResource();
-      List<SpanData> spanDataList = result.computeIfAbsent(resource, k -> new ArrayList<>());
-      spanDataList.add(spanData);
-    }
-    return result;
-  }
 
   /**
    * Converts a list of {@link SpanData} into a collection of Jaeger's {@link Model.Span}.
--- a/exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/Adapter.java
+++ b/exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/Adapter.java
@@ -75,10 +75,6 @@ final class Adapter {
     target.addAllLogs(toJaegerLogs(span.getEvents()));
     target.addAllReferences(toSpanRefs(span.getLinks()));
 
-    if (span.getResource() != null && span.getResource().getAttributes() != null) {
-      target.addAllTags(toKeyValues(span.getResource().getAttributes()));
-    }
-
     // add the parent span
     if (SpanId.isValid(span.getParentSpanId())) {
       target.addReferences(
--- a/exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/Adapter.java
+++ b/exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/Adapter.java
@@ -16,14 +16,17 @@ import io.opentelemetry.common.AttributeKey;
 import io.opentelemetry.common.ReadableAttributes;
 import io.opentelemetry.exporters.jaeger.proto.api_v2.Model;
 import io.opentelemetry.sdk.extensions.otproto.TraceProtoUtils;
+import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.sdk.trace.data.SpanData;
 import io.opentelemetry.sdk.trace.data.SpanData.Event;
 import io.opentelemetry.sdk.trace.data.SpanData.Link;
 import io.opentelemetry.trace.SpanId;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 import javax.annotation.concurrent.ThreadSafe;
 
 /** Adapts OpenTelemetry objects to Jaeger objects. */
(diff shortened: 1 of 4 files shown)


exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/Adapter.java
```diff
@@ -192,7 +196,10 @@ private Adapter() {}
   @VisibleForTesting
   static <T> Model.KeyValue toKeyValue(AttributeKey<T> key, T value) {
     Model.KeyValue.Builder builder = Model.KeyValue.newBuilder();
-    builder.setKey(key.getKey());
+    builder.setKey(
+        key.getKey()
+            .replace("InstrumentationLibrary.name", "otel.library.name")
+            .replace("InstrumentationLibrary.version", "otel.library.version"));
```
bogdandrutu, line 202: Why is this needed?

anuraaga: Thanks @malafeev - looks like there was some confusion on the issue, yeah you shouldn't need to do any sort of name translation logic.

exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/Adapter.java
```diff
@@ -75,6 +75,10 @@ private Adapter() {}
     target.addAllLogs(toJaegerLogs(span.getEvents()));
     target.addAllReferences(toSpanRefs(span.getLinks()));
 
+    if (span.getResource() != null && span.getResource().getAttributes() != null) {
+      target.addAllTags(toKeyValues(span.getResource().getAttributes()));
```
anuraaga, line 79: Resource corresponds to process, so we want to add these to the process's tags. Can you move this logic to the process here?

https://github.com/open-telemetry/opentelemetry-java/blob/master/exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java#L89

malafeev: @anuraaga there I don't have `SpanData` to get `getResource().getAttributes()`.
should I get resource attributes using `Resource.getDefault().getAttributes()` ?

anuraaga: Oh interesting - in that case I think this does need to be a bigger refactoring to move the process struct construction into the export flow. We should be able to follow a logic flow very similar to OTLP which starts by grouping by resource

https://www.github.com/open-telemetry/opentelemetry-java/tree/master/exporters%2Fotlp%2Fsrc%2Fmain%2Fjava%2Fio%2Fopentelemetry%2Fexporters%2Fotlp%2FSpanAdapter.java

malafeev: `Collector.PostSpansRequest` accepts only one `Model.Batch`. But `Model.Batch` is per `Model.Process`. 
https://github.com/open-telemetry/opentelemetry-java/blob/master/exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java#L107
If there are several different Resources then I need to create several batches and send several requests.

Should I change [collector.proto](https://github.com/open-telemetry/opentelemetry-java/blob/master/exporters/jaeger/src/main/proto/jaeger/api_v2/collector.proto#L14) to accept multiple batches?
My concern is that jaeger backend will not accept such payload.

jkwatson: Not an expert in this, but my guess is that we'll need to change the exporter to potentially send more than one batch, one per Resource on the span data. The case where you get multiple Resources per batch of spans should be very rare, so I wouldn't worry about the fact that we're sending multiple requests per export too much.

malafeev: refactored to group spans by resource and then set process tags per resource

exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/Adapter.java
malafeev: removed

malafeev: I was confused by https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk_exporters/jaeger.md#resource
I decided that `InstrumentationLibrary.name` should be translated into `otel.library.name` etc

exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/Adapter.java
```diff
@@ -39,6 +42,22 @@
...
+   * @return the map of grouped spans
+   */
+  static Map<Resource, List<SpanData>> groupByResource(Collection<SpanData> spans) {
+    Map<Resource, List<SpanData>> result = new HashMap<>();
+    for (SpanData spanData : spans) {
+      Resource resource = spanData.getResource();
+      List<SpanData> spanDataList = result.computeIfAbsent(resource, k -> new ArrayList<>());
+      spanDataList.add(spanData);
+    }
+    return result;
```
jkwatson, line 58: Did you consider using a java 7 Stream with a `groupingBy` collector to do this work? It might read a little more clearly. It's a one-liner in this case:

`    return spans.stream().collect(Collectors.groupingBy(SpanData::getResource));`

jkwatson: (and, given that it's a one-liner...you can probably just one-liner it in the calling code, rather than adding a function here.)

exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java
```diff
@@ -104,26 +109,42 @@ private JaegerGrpcSpanExporter(String serviceName, ManagedChannel channel, long
...
+      }
+
+      Collector.PostSpansRequest request =
+          Collector.PostSpansRequest.newBuilder()
+              .setBatch(
+                  Model.Batch.newBuilder()
+                      .addAllSpans(Adapter.toJaeger(entry.getValue()))
+                      .setProcess(builder.build())
+                      .build())
+              .build();
```
jkwatson, line 132: I'd extract lines 120-132 into a separate method, and then you could use a java 8 Stream/map to collapse this logic significantly.

jkwatson: Also, please add a unit test to the `JaegerGrpcSpanExporterTest` to cover this change.

malafeev: done

exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/Adapter.java
malafeev: done

exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java
jkwatson: I think it would be good to have a test that uses multiple resources, so we can verify that multiple batches are sent, and we don't just use the first Resource we find. I didn't see this case in the test code. did I miss it?

exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java
```diff
@@ -137,6 +144,21 @@ public void onFailure(Throwable t) {
     return result;
   }
 
+  private Collector.PostSpansRequest buildRequest(Resource resource, List<SpanData> spans) {
+    Process.Builder builder = this.processBuilder.clone();
+    if (resource.getAttributes() != null) {
```
anuraaga, line 149: Don't need null check for attributes

anuraaga (commented):
Thanks for the help! Couple of small points only

exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/Adapter.java
```diff
@@ -75,6 +75,10 @@ private Adapter() {}
     target.addAllLogs(toJaegerLogs(span.getEvents()));
     target.addAllReferences(toSpanRefs(span.getLinks()));
 
+    if (span.getResource() != null && span.getResource().getAttributes() != null) {
```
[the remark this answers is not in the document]
malafeev: removed

exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java
malafeev: removed

exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/JaegerGrpcSpanExporter.java
malafeev: added test

jkwatson (approved):

anuraaga (approved):
Thanks @malafeev!

PR #1803, opened 2020-10-14
Move the stress test code to being a TestContainers-based test.

jkwatson:
Also: 

* Upgrade testcontainers version to work with the latest docker-for-mac.
* Add metrics in the OTLP exporter, to record success/failure of export.
* Use the TestContainers annotation to disable those tests when docker isn't present.

The only part I'm leery about here, is adding metrics to the exporter. I did find them super useful to gain insight into what the exporter thought had been exported, vs what the collector actually received, during network flakiness tests. I'd be happy to revert that piece for now, but I definitely think it could be a very useful addition that could help when debugging real-world issues.


integration_tests/src/test/java/io/opentelemetry/JaegerExporterIntegrationTest.java
```diff
@@ -58,7 +59,7 @@
   @SuppressWarnings("rawtypes")
   @Container
   public static GenericContainer jaegerExampleAppContainer =
-      new GenericContainer("adoptopenjdk/openjdk8")
+      new GenericContainer(DockerImageName.parse("adoptopenjdk/openjdk8"))
```
anuraaga, line 62: Due to a bug in the RC of testcontainers, you need to add `:latest` here or notice the broken CI (it downloads every tag for the image ever right now...) Can you add `:lastest` to all the testcontainers untagged images?

jkwatson: oh blah. ok, I'll fix it

jkwatson: done. 🤞

anuraaga (approved):

anuraaga:
EasyCLA seems to be dead :(

jkwatson:
> EasyCLA seems to be dead :(

While we wait, any thoughts on adding the  metrics to the OTLP exporter?

anuraaga:
Yeah the metrics seem important for exactly the reason you found :) I think I saw a spec issue about defining them but we could rename later as needed I think.

jkwatson:
closing and reopening for the CLA bot

PR #1805, opened 2020-10-15
Extract an interface for the OpenTelemetry API to later allow an SDK …

anuraaga:
…to implement it.

Extracted from #1737 see there for the larger picture


api/src/main/java/io/opentelemetry/OpenTelemetry.java
```diff
@@ -5,255 +5,194 @@
...
+   *     interface FQCN but the specified provider cannot be found.
    */
-  public static TracerProvider getTracerProvider() {
-    return getInstance().tracerProvider;
+  static OpenTelemetry get() {
+    return DefaultOpenTelemetry.getGlobalOpenTelemetry();
+  }
+
+  /** Returns the globally registered {@link TracerProvider}. */
+  static TracerProvider getGlobalTracerProvider() {
```
anuraaga, line 47: Unfortunately the statics can't be named the same as the method name @carlosalberto. So added them with the "global" prefix for now.

api/src/main/java/io/opentelemetry/OpenTelemetry.java
```diff
@@ -5,255 +5,194 @@
...
    *
-   * @return registered TracerProvider or default via {@link DefaultTracerProvider#getInstance()}.
-   * @throws IllegalStateException if a specified TracerProvider (via system properties) could not
-   *     be found.
+   * @throws IllegalStateException if a provider has been specified by system property using the
+   *     interface FQCN but the specified provider cannot be found.
    */
-  public static TracerProvider getTracerProvider() {
-    return getInstance().tracerProvider;
+  static OpenTelemetry get() {
```
jkwatson, line 42: Do we need to expose this on the public interface at this point? Could the callers here just make the call to `DefaultOpenTelemetry` directly?

api/src/main/java/io/opentelemetry/DefaultOpenTelemetry.java
```diff
@@ -0,0 +1,239 @@
...
+    }
+    return null;
+  }
+
+  // for testing
+  static void reset() {
+    globalOpenTelemetry = null;
+  }
+
+  public Builder toBuilder() {
```
jkwatson, line 134: The builder methods can be package access at this point, correct?

sdk_extensions/logging/src/main/java/io/opentelemetry/sdk/logging/export/BatchLogProcessor.java
```diff
@@ -70,7 +70,7 @@ public CompletableResultCode forceFlush() {
 
   private static class Worker implements Runnable {
     static {
-      Meter meter = OpenTelemetry.getMeter("io.opentelemetry.sdk.logging");
+      Meter meter = OpenTelemetry.getGlobalMeter("io.opentelemetry.sdk.logging");
```
jkwatson, line 73: this will probably be something we want to inject, rather than grabbing it globally. follow-on PRs are fine.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -133,7 +133,7 @@ public CompletableResultCode forceFlush() {
   private static final class Worker implements Runnable {
 
     static {
-      Meter meter = OpenTelemetry.getMeter("io.opentelemetry.sdk.trace");
+      Meter meter = OpenTelemetry.getGlobalMeter("io.opentelemetry.sdk.trace");
```
jkwatson, line 136: another case where we'll want to inject a MeterProvider, probably?

jkwatson:
1. Needs a rebase.
2. Thanks!!!
3. This looks good; I added a couple comments about places where we will (I think) want to not grab the global instances, but follow-on PRs are fine for that.
4. Thanks!

api/src/main/java/io/opentelemetry/OpenTelemetry.java
anuraaga: Oops, thanks for finding where I forgot to remove publicness, cleaned up

jkwatson (approved):

jkwatson:
@bogdandrutu @carlosalberto please comment ASAP if you'd like changes to this, as I think we'd like to get it into the main line so the work can continue.

api/src/main/java/io/opentelemetry/DefaultOpenTelemetry.java
```diff
@@ -0,0 +1,235 @@
...
+ * and {@link BaggageManager}.
+ *
+ * <p>The telemetry objects are lazy-loaded singletons resolved via {@link ServiceLoader} mechanism.
+ *
+ * @see TracerProvider
+ * @see MeterProviderFactory
+ * @see BaggageManagerFactory
+ */
+@ThreadSafe
+class DefaultOpenTelemetry implements OpenTelemetry {
```
bogdandrutu, line 39: final?

api/src/main/java/io/opentelemetry/OpenTelemetry.java
bogdandrutu: If I did this change, I would split the renaming in a separate PR, because that is a trivial change. Hard now to look at all changes and determine what is trivial vs what is behavior change.

anuraaga: Do you mean making the instance methods something like getInstanceTracerProvider and swapping the names later? That seems like intentionally merging a bad name for a minor improvement in code review - the changes are only to this interface and the class. It's not something I'd generally recommend a contributor to have to do.

anuraaga: Went ahead and renamed the instance methods, still don't think it's a good idea to merge this.

api/src/main/java/io/opentelemetry/DefaultOpenTelemetry.java
```diff
@@ -0,0 +1,235 @@
...
+  @Nullable private static volatile OpenTelemetry globalOpenTelemetry;
+
+  private final TracerProvider tracerProvider;
+  private final MeterProvider meterProvider;
+  private final BaggageManager baggageManager;
+
+  private final ContextPropagators propagators;
+
+  @Override
+  public TracerProvider getMyTracerProvider() {
```
jkwatson, line 70: if we do merge this as-is, we need to get these renamed ASAP, for sure.

jkwatson: can you stack a 2nd PR on top of this one with the renames, so we can move fast in getting it fixed?

anuraaga: I'd set up another PR before merging this after it's green-lighted by @bogdandrutu

api/src/main/java/io/opentelemetry/OpenTelemetry.java
```diff
@@ -5,255 +5,185 @@
...
   }
 
   /**
-   * Gets or creates a named and versioned tracer instance.
+   * Gets or creates a named and versioned tracer instance from the globally registered {@link
+   * TracerProvider}.
    *
-   * <p>This is a shortcut method for <code>
-   * getTracerProvider().get(instrumentationName, instrumentationVersion)</code>.
+   * <p>This is a shortcut method for @codegetTracerProvider().get(instrumentationName,
```
carlosalberto, line 54: {@code getTracreProvider()...} ?

carlosalberto:
Looks good to me, with the only caveat of the naming that @jkwatson mentioned (which seems we will fix ASAP after this is merged, correct?)

api/src/main/java/io/opentelemetry/OpenTelemetry.java
jkwatson: @anuraaga fix this up and we're ready go merge, IMO

anuraaga: Ouch - thanks fixed

jkwatson:
I think this needs a rebase? Baggage issues

anuraaga:
@carlosalberto Yup, filed #1842 to prepare for the rename

carlosalberto (approved):

Issue #1807, opened 2020-10-15
Context usability improvements


anuraaga:
Context is an object that is a global concern in `OpenTelemetry`, while the "current span" is a concept of tracing specifically. I'm hoping we can model this ownership correctly, while also provide users an intuitive way of accessing these objects when they need to.

## Accessing Context

The `Context` is outside of `OpenTelemetry` and something we use statically to insert telemetry information. So we only expect static access like this

```
class Context
  static Context current() {
    return contextImplementation.current();
  }
```

## Accessing Span

The active span is only available during tracing. The parallel to the above for a span is 

```
class Span {
  static Span current() {
    return Context.current().getValue(CURRENT_SPAN);
  }
}
```

This is very important, since many users don't do any tracing at all, they're just using the agent, or added some library instrumentation. But adding attributes is common - `Span.current` allows them to enrich spans without worry about the concept of tracing itself. 

## Making context / span current

Because the context / span are by definition available if a user needs to make them current, instance methods are convenient `try (Scope ignored = context.makeCurrent())`, `try (Scope ignored = span.makeCurrent())`. 

## Adding span to context

The final operation is adding a span to a existing context, which instrumentation authors need. Because we only really expect instrumentation authors to use this, `tracer.addToContext(Span span, Context context)` or `span.addToContext(Context context)` both seem OK.

Oberon00:
> Context is an object that is a global concern in OpenTelemetry


```
class Context
  static Context current() {
    return OpenTelemetry.get().currentContext();
  }
```

It's the other way round: Context is not in itself related to telemetry, but telemetry uses Context. So I think the code snippet above violates the layering prescribed by the spec.

anuraaga:
I guess that makes sense - in that case context storage will always be separate from `OpenTelemetry` and initialized / accessed statically I think. In that case, I think the parallel to `Context.current()` of `Span.current()` is important, and we may not even bother with `tracer.currentSpan()`?

Oberon00:
+1 on removing current span access from tracer.

iNikem:
`tracer.getCurrentSpan()` should die. It really makes an impression, that spans are somehow scoped/related to tracers and you have to have _the correct tracer_ to get a span. I think this is a huge source of confusion.

bogdandrutu:
[...]

PR #1809, opened 2020-10-15
Remove getCurrentContext and withSpan from Tracer

bogdandrutu:
Updates #1807 

### DIFF
--- a/api/src/main/java/io/opentelemetry/trace/TracingContextUtils.java
+++ b/api/src/main/java/io/opentelemetry/trace/TracingContextUtils.java
@@ -5,6 +5,7 @@
 
 package io.opentelemetry.trace;
 
+import com.google.errorprone.annotations.MustBeClosed;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.ContextKey;
 import io.opentelemetry.context.Scope;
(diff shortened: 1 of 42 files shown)


api/src/main/java/io/opentelemetry/trace/TracingContextUtils.java
```diff
@@ -66,9 +67,46 @@ public static Span getSpanWithoutDefault(Context context) {
...
+   *   Span span = tracer.spanBuilder("my span").startSpan();
+   *   try (Scope ws = TracingContextUtils.currentContextWith(span)) {
+   *     TracingContextUtils.getCurrentSpan().addEvent("my event");
+   *     doSomeOtherWork();  // Here "span" is the current Span.
+   *   }
+   *   span.end();
+   * }
+   * }</pre>
+   *
+   * <p>Prior to Java SE 7, you can use a finally block to ensure that a resource is closed
```
bogdandrutu, line 85: Need to remove this, no longer supported :)

bogdandrutu: Done :)

carlosalberto (approved):

carlosalberto:
LGTM. Now that this will be the official way, I feel like we should do a follow up to rename `TracingContextUtils` to something more friendly ;)

Oberon00:
Shouldn't we just move TracingContextUtils to static Span interface methods? Or even default Span interface methods where it makes sense.

jkwatson:
> Shouldn't we just move TracingContextUtils to static Span interface methods? Or even default Span interface methods where it makes sense.

I think this is a topic of conversation in #1807

jkwatson:
I hate this change, by the way. It makes the Tracer into something silly and nearly useless. It should be called SpanFactory. :(

PR #1815, opened 2020-10-15
Revert "Remove getCurrentContext and withSpan from Tracer"

bogdandrutu:
Reverts open-telemetry/opentelemetry-java#1809

Per comments from @jkwatson let's revert this and go back to the drawing board.


jkwatson:
I don't know if we need to revert. But, we do need to put things in a better place than TracingContextUtils. And, I'd like to understand if a Tracer will ever be anything but a SpanFactory, which is what we're currently left with.

jkwatson:
Moving forward with discussion about what to do with the Tracer in #1807 . So let's close this and just move forward with making the whole thing coherent ASAP.

Issue #1794, opened 2020-10-13
Feature Request: debug/troubleshooting mode for the Context implementation


jkwatson:
When writing instrumentation that is concerned with context propagation, especially of async frameworks, it would be very useful to be able to enable a "test" or "debug" mode for the Context that will allow throwing an exception, or failing loudly in some way, if not all the Scopes have been closed during the execution of the tests, or if mismatched Scopes are resolved.

In addition, it would be helpful to have a system property (perhaps) that you could enable in a production system that would raise the level error reporting about mis-matched scope/context resolution from the default FINE level.

jkwatson:
This is done via the StrictContextStorage and the ability to turn it on via `-Dio.opentelemetry.context.enableStrictContext=true`

PR #1821, opened 2020-10-16
Log Context's Scope.close misuse SEVERE (not FINE).

Oberon00:
We could argue about WARNING, but FINE (being disabled by default) is IMHO too low for this.

### DIFF
--- a/context/src/main/java/io/opentelemetry/context/ThreadLocalContextStorage.java
+++ b/context/src/main/java/io/opentelemetry/context/ThreadLocalContextStorage.java
@@ -36,7 +36,7 @@ enum ThreadLocalContextStorage implements ContextStorage {
     return () -> {
       if (current() != toAttach) {
         logger.log(
-            Level.FINE,
+            Level.SEVERE,
             "Context in storage not the expected context, Scope.close was not called correctly");
       }
       THREAD_LOCAL_STORAGE.set(beforeAttach);


carlosalberto (approved):

carlosalberto:
Makes sense. Also, this is what grpc.Context does as well.

anuraaga:
We'll provide a proper debug mode when solving #1794. I don't think the sdk should ever have a log statement that can spam every request potentially multiple times.

Oberon00:
This is not really in the SDK but in the context implementation, which is in theory usable without OpenTelemetry. And as @carlosalberto said, the old context implementation did the same. You can always disable the warning manually.

We also can't really record a metric for this, which we could do elsewhere in the SDK.

jkwatson (changes requested):
Please don't merge this until we can discuss.

anuraaga:
By SDK I meant library a user adds to get telemetry - spamming logs can bring down apps and a library shouldn't do that I think. Google of course doesn't have such a problem.

carlosalberto:
> spamming logs can bring down apps and a library

Although this is totally true, closing a Context's Scope in the incorrect order ir a rather fatal scenario (in the grpc.Context docs it even says that this is the safest alternative to throwing an actual Exception). Hence I'd be up for merging this PR.

anuraaga:
I think it's only fatal to tracing, but not to an app. I'd recommend our sdk not take any risk of taking down an app by log spamming in an environment not ready for that since tracing is still a secondary function. I would encourage it to be opt in instead of default.

carlosalberto:
> I think it's only fatal to tracing, but not to an app.

If we ever want this to become an implementation that others can use (and, I think, that's the dream), instead of having Reactor, grpc and other have each its own implementation, then this change makes perfect sense, as it's an invalid, almost fatal option in general.

Even if we don't ever share it, the Specification's ideal has long be to have this as an independent layer (hence its separation as a package in the first place), that can be used by anybody who wants to integrate their cross-cutting concern with OpenTelemetry. In that regard, such users will definitely want to know when things are going wrong.

If we truly think we will only use it for Tracing, then I'd be fine with FINE, but I doubt that will be the case.

anuraaga:
@carlosalberto We can always change the log level later if it seems that is a useful change right? Let's not speculate so much for now and use a more conservative default. As I mentioned, we will be adding better tools for debugging scope issues and maybe we'll find them to be enough anyways - or maybe we'll add a metric. A log message when detaching tends to not be so informative since context propagation already broke earlier some where - what's more useful is a log message when attaching while a different local root is already attached. But that can only be done in the instrumentation layer, not here.

jkwatson:
@johnbley Do you have any thoughts here?

carlosalberto:
> We can always change the log level later if it seems that is a useful change right?

Oh definitely. But if that's the case, why don't we start with grpc.Context's default, and see how it goes? Also, I'd argue such value IS the conservative value, as it's something that has been used for long time in grpc.Context, and I don't really think it caused a big problem there :)

anuraaga:
> Also, I'd argue such value IS the conservative value, as it's something that has been used for long time in grpc.Context

That's not a conservative value, that's just cargo culting from gRPC :P If we copy everything from gRPC context, we would throw NPE on null keys which isn't allowed for OTel - in this implementation, I call `identityHashCode` instead of `key.hashCode` to avoid it, and I chuckled a bit remembering people calling grpc context a robust implementation. But it's true that server frameworks often do validate and throw exceptions - the expectations of error handling vs instrumentation is just different and we can't just blindly copy from gRPC.

Think of the team running their service on bare metal with logs stored to disk. If one API is updated to use a cool async pattern but doesn't close scope properly, it will spam the logs at least once per request. This can easily fill up the disk and cause the entire service to go down, despite nothing being wrong for the users. The SRE team will yell at the engineers again for adding a spammy log statement when they should have used metrics to have alerts without the potential of an outage.

OpenTelemetry must not have a default behavior that can cause an outage, period. And our context implementation needs to behave for this lowest common denominator, even if it does get used in cases outside of telemetry. By then we may have metrics complete and better handling than the logging anyways, so let's not speculate about this distant use case now and focus on our needs to not cause outages.

trask:
Is there any interest in logging at SEVERE/WARNING at most once a minute (and including in the message that it will only be logged at most once a minute)?

Or logging at SEVERE/WARNING only the first time it occurs (again, including in the message that it will only be logged once)?

This could be in addition to always logging at FINE.

anuraaga:
I think if we log once per call site, it would be fine at severe (sorry couldn't help it). `ConcurrentHashMap<StackTraceElement, Boolean>` could do it?

Oberon00:
> I think if we log once per call site

That would be nice, but overkill. Maybe for now we could use a static boolean to log only the first (or a static int to log only the first 20 or so)? A more featureful throttled logger would be nice of course, but maybe too much for a lightweight context package.

anuraaga:
@Oberon00 Log just once seems fine to me too, indeed it's a lot simpler than call-site tracking.

jkwatson:
My opinion is in the "do no harm" camp. Any logging that is on by default MUST be non-disruptive. So, either opt-in to get logging, or be very restrictive in how many times we log out of the box.

carlosalberto:
> either opt-in to get logging, or be very restrictive in how many times we log out of the box.

What about allowing users to configure this? In that case we can use FINE by default, but allow users to trigger a SEVERE if the `Context` is configured through some env var or system property (and no, having the user intercept this logging does not sound very nice IMHO).

jkwatson:
> > either opt-in to get logging, or be very restrictive in how many times we log out of the box.
> 
> What about allowing users to configure this? In that case we can use FINE by default, but allow users to trigger a SEVERE if the `Context` is configured through some env var or system property (and no, having the user intercept this logging does not sound very nice IMHO).

That's this issue here, I believe:  https://github.com/open-telemetry/opentelemetry-java/issues/1794

carlosalberto:
@jkwatson Oh, I totally had forgotten about that. Yes, in that case IMHO we should follow up the SEVERE case/check as part of that.

Oberon00:
OK then should I close this PR and this will be done as part of #1794?

anuraaga:
Really sorry @Oberon00 but that's my recommendation. I have some ideas to play with for #1794 and have wanted to after getting the instrumentation on the new context implementation in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1384 but it's been a rabbit hole of random bytecode manipulation issues 🐇

Oberon00:
@anuraaga No problem, it's not like this PR is the result of hours of hard work 😄

PR #1822, opened 2020-10-16
Make the baggage API fully functional in the API

jkwatson:
and get rid of the BaggageManager that was needed previously.

This moves the Baggage implementation to the API, per https://github.com/open-telemetry/opentelemetry-specification/pull/1103, which is now merged.

I could split this up, if desired, into 2 parts:
1) Move the implementation to the API, remove it from the SDK, but keep the BaggageManager in place.
2) Remove the BaggageManager entirely.


api/src/main/java/io/opentelemetry/OpenTelemetry.java
```diff
@@ -126,17 +125,6 @@ public static Meter getMeter(String instrumentationName, String instrumentationV
...
   }
 
-  /**
-   * Returns a singleton {@link BaggageManager}.
-   *
-   * @return registered manager or default via {@link DefaultBaggageManager#getInstance()}.
-   * @throws IllegalStateException if a specified manager (via system properties) could not be
-   *     found.
-   */
-  public static BaggageManager getBaggageManager() {
```
anuraaga, line 136: 🎉

api/src/main/java/io/opentelemetry/baggage/ImmutableBaggage.java
```diff
@@ -22,24 +17,29 @@
 @Immutable
 // TODO: Migrate to AutoValue
 // @AutoValue
-class BaggageSdk implements Baggage {
+public class ImmutableBaggage implements Baggage {
```
anuraaga, line 20: Does this need to be public?

anuraaga (approved):
Mostly good, just one point about visibility - presume since we have an interface, we don't need to make the implementation of baggage public

api/src/main/java/io/opentelemetry/baggage/ImmutableBaggage.java
jkwatson: hmm. good point. I have to adjust a silly test, but that's easy enough to do.

jkwatson: fixed

anuraaga:
Thanks! Can you rebase to get the build fix?

carlosalberto (approved):

carlosalberto:
Late review, thanks for this!

Issue #1819, opened 2020-10-16
If there is only one ContextStorageProvider, the system property is ignored without warning


Oberon00:
**Describe the bug**
(Note: This bug report comes only from looking at the code, I did not try to actually reproduce it)

The following conditions must be fulfilled:
1. The `io.opentelemetry.context.contextStorageProvider` system property is set
2. There is exactly a single provider found by `ServiceLoader.load(ContextStorageProvider.class)`
3. The found provider is not the one requested by the system property.

Relevant code: https://github.com/open-telemetry/opentelemetry-java/blob/fa1e876df20910ca5ec3d43a788f2bcf35013d68/context/src/main/java/io/opentelemetry/context/LazyStorage.java#L51-L54

**Steps to reproduce**
Obvious from description.

**What did you expect to see?**
1. A warning message that the requested provider could not be found.
2. The default implementation should be used instead of the one of the provider. (This point is debatable.)

**What did you see instead?**
1. No message at all.
2. The single, found provider was used.

dengliming:
> The default implementation should be used instead of the one of the provider. (This point is debatable.)

@jkwatson Can you confirmed this?

jkwatson:
> > The default implementation should be used instead of the one of the provider. (This point is debatable.)
> 
> @jkwatson Can you confirm this?

I'm not sure what the right behavior is. I'll defer to @anuraaga on this one.

anuraaga:
@dengliming Thanks for the help! I think using the default after logging makes sense.

PR #1823, opened 2020-10-17
Fix using default context storage implemenation in LazyStorage

dengliming:
fixes [https://github.com/open-telemetry/opentelemetry-java/issues/1819](https://github.com/open-telemetry/opentelemetry-java/issues/1819)

### DIFF
--- a/context/src/main/java/io/opentelemetry/context/LazyStorage.java
+++ b/context/src/main/java/io/opentelemetry/context/LazyStorage.java
@@ -36,7 +36,7 @@ final class LazyStorage {
     }
   }
 
-  private static ContextStorage createStorage(AtomicReference<Throwable> deferredStorageFailure) {
+  static ContextStorage createStorage(AtomicReference<Throwable> deferredStorageFailure) {
     String providerClassName = System.getProperty(CONTEXT_STORAGE_PROVIDER_PROPERTY, "");
 
     List<ContextStorageProvider> providers = new ArrayList<>();
--- a/context/src/main/java/io/opentelemetry/context/LazyStorage.java
+++ b/context/src/main/java/io/opentelemetry/context/LazyStorage.java
@@ -49,9 +49,13 @@ final class LazyStorage {
     }
 
     if (providerClassName.isEmpty()) {
+      if (providers.size() == 1) {
+        return providers.get(0).get();
+      }
+
       deferredStorageFailure.set(
           new IllegalStateException(
-              "Found ContextStorageProvider. Set the "
+              "Found multiple ContextStorageProvider. Set the "
                   + "io.opentelemetry.context.ContextStorageProvider property to the fully "
                   + "qualified class name of the provider to use. Falling back to default "
                   + "ContextStorage. Found providers: "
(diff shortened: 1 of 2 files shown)


anuraaga (commented):
Thanks a lot for the PR! I think we do want to be able to use a context provider implicitly if it's the only one since it's fairly unambiguous - the issue was that the property may also be set and to a different provider so we should handle that case specially. Does that make sense?

dengliming:
@anuraaga So The `providerClassName` is not required, is that right? I'm still a little confused 

> I think we do want to be able to use a context provider implicitly if it's the only one since it's fairly unambiguous

Do you mean that the `io.opentelemetry.context.contextStorageProvider ` system property is not set and there is exactly a single provider found by `ServiceLoader.load(ContextStorageProvider.class)`?

anuraaga:
Right if it's unset we can use the single implementation. We want to prevent a user explicitly wanting an implementation by setting the property but getting a different one.

jkwatson (commented):
Is it possible to have unit tests to cover this change?

dengliming:
@jkwatson yeah. Seems the unit tests only cover `ContextStorage.get()`. We don't have unit tests for `LazyStorage` now. I will add one later if that make senses.

context/src/main/java/io/opentelemetry/context/LazyStorage.java
```diff
@@ -48,17 +48,11 @@ private static ContextStorage createStorage(AtomicReference<Throwable> deferredS
...
     }
 
-    if (providers.size() == 1) {
-      ContextStorageProvider provider = providers.get(0);
-      try {
-        return provider.get();
-      } catch (Throwable t) {
-        deferredStorageFailure.set(t);
-        return DefaultContext.threadLocalStorage();
+    if (providerClassName.isEmpty()) {
```
anuraaga, line 51: @dengliming Sorry if my suggestions weren't clear, thanks for bearing with me. We have these success cases

- 1 provider, providerClassName empty
- 1 provider, providerClassName matches the provider <- currently missing
- multiple providers, providerClassName matches one of them

Hope this makes it clear

anuraaga (commented):
Yeah since this is SPI I think tests are more complicated by needing a testset per case. I'm ok with a followup. Thanks for the help @dengliming

context/src/main/java/io/opentelemetry/context/LazyStorage.java
dengliming: yeah. a little confused. Can you also add the right return for each case？

> 1 provider, providerClassName matches the provider <- currently missing

don't know what is missing. The following code should cover this case. is that correct?
```
for (ContextStorageProvider provider : providers) {
      if (provider.getClass().getName().equals(providerClassName)) {
        return provider.get();
      }
    }
```

anuraaga: Sure updated my comment, in every other case we log and return default storage.

Yeah starting with that, and than one more branch for if there's a single provider and empty providerClassName seems fine to me.

dengliming: I think the current changes can satisfy all the cases. Just need log in some case. WDYT?

anuraaga: Sorry got it! Yeah seems good, really needed to expand the diff

anuraaga (approved):

thisthat (approved):

PR #1827, opened 2020-10-19
Add a ContextValue interface to allow Context.withValue(value) withou…

anuraaga:
…t a key.

Proposal to demonstrate https://github.com/open-telemetry/opentelemetry-java/issues/1807#issuecomment-711544626

### DIFF
--- a/context/src/main/java/io/opentelemetry/context/ContextValue.java
+++ b/context/src/main/java/io/opentelemetry/context/ContextValue.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.context;
+
+/**
+ * A value that can be stored inside {@link Context}. Types will generally use this interface to
+ * allow storing themselves in {@link Context} without exposing a {@link ContextKey}.
+ */
+public interface ContextValue {
+
+  /**
+   * Returns a new {@link Context} created by setting {@code this} into the provided {@link
+   * Context}. It is generally recommended to call {@link Context#withValues(ContextValue)} instead
+   * of this method. The following are equivalent.
+   *
+   * <ul>
+   *   <li>{@code context.withValues(myContextValue)}
+   *   <li>{@code myContextValue.storeInContext(context)}
+   * </ul>
+   */
+  Context storeInContext(Context context);
+}
--- a/extensions/trace_utils/src/test/java/io/opentelemetry/extensions/trace/CurrentSpanUtilsTest.java
+++ b/extensions/trace_utils/src/test/java/io/opentelemetry/extensions/trace/CurrentSpanUtilsTest.java
@@ -6,8 +6,9 @@
 package io.opentelemetry.extensions.trace;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.verifyNoInteractions;
+import static org.mockito.Mockito.when;
 
 import io.opentelemetry.trace.Span;
 import io.opentelemetry.trace.StatusCode;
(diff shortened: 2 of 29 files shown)


extensions/trace_utils/src/test/java/io/opentelemetry/extensions/trace/CurrentSpanUtilsTest.java
```diff
@@ -60,7 +61,6 @@ void withSpanRunnable() {
           assertThat(getCurrentSpan()).isSameAs(span);
         };
     CurrentSpanUtils.withSpan(span, false, runnable).run();
-    verifyNoInteractions(span);
```
Oberon00, line 63: I think in this case we should use verifyNoFurtherInteractions and assert on the call to storeInContext to not lose the original assertion's intent.

Oberon00 (approved):
Cool, I like this idea!

extensions/trace_utils/src/test/java/io/opentelemetry/extensions/trace/CurrentSpanUtilsTest.java
anuraaga: Thanks - I was actually a bit confused on the original intent but took another look. I went with explicitly checking end isn't called (as specified by the `false` above) which is symmetric with the other tests, I think this would have been clearer to me from the start. Hope this looks OK.

context/src/main/java/io/opentelemetry/context/ContextValue.java
```diff
@@ -0,0 +1,25 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.context;
+
+/**
+ * A value that can be stored inside {@link Context}. Types will generally use this interface to
+ * allow storing themselves in {@link Context} without exposing a {@link ContextKey}.
+ */
+public interface ContextValue {
```
jkwatson, line 12: Just thinking out loud about naming. 
What about 
"Propagatable"
"ContextAware"
"ContextPropagatable"
"Contextual"
"PropagationAware"
"PropagationSupport"
"ContextSupport"

jkwatson:
I like this idea a lot. Not 100% sure about the interface name, but that's a small detail. Thanks for putting this together!

context/src/main/java/io/opentelemetry/context/ContextValue.java
Oberon00: AutoContextStorable
ContextKeyAssociated
ImplicitlyContextKeyed
WithImplicitContextKey

anuraaga: I like `ContextStorable`, not auto since a user still has to call `withValue` so not so automatic, and `store` is similar naming to the method name. `ImplicitContextKeyed` seems nice too for explaining the why if not necessarily the how. I don't know how to pick among any of the names here, what should I do? :)

trask: +1 for `ImplicitContextKeyed`

anuraaga:
Went ahead and renamed to `ImplicitContextKeyed` for now, it's easy to rename to anything though so still open to changes if anyone feels strongly.

Also went ahead and matched the method name with #1838, and `context.with(span)` seems to look swell.

context/src/main/java/io/opentelemetry/context/ContextValue.java
jkwatson: How about just `ContextKeyed` ?

jkwatson (approved):

carlosalberto (approved):

carlosalberto:
> Not 100% sure about the interface name, but that's a small detail. 

Same feeling here, although bot a big deal. I also happen to like `ContextKeyed` more ;)

trask:
fwiw, `Implicit` is the part of the name that I like, because anything can be a context value, but some context values have implicit keys, which explains why they can be passed to the context without a (explicit) key

PR #1828, opened 2020-10-19
Restore third party copyright headers in context package.

### LINKED ISSUE #1820: Original license headers are missing for context package.
In the PR description of #1658:

> The API is heavily inspired by gRPC and Armeria. Much code / docs are borrowed too and I kept license headers in tact to reflect that.

which is sensible and actually required by the OTel/CNCF policy:

> According to CNCF recommendations if you are contributing third-party code you will need to retain the original copyright notice.

-- https://github.com/open-telemetry/community/blob/master/CONTRIBUTING.md#copyright-notices

However, the statement from the PR description is unfortunately, wrong there is only the OTel license. Maybe the original license header got killed by spotlessApply.

CC @anuraaga

anuraaga:
Got lost in merge.

Fixes #1820

/cc @iNikem

### DIFF
--- a/context/src/main/java/io/opentelemetry/context/PersistentHashArrayMappedTrie.java
+++ b/context/src/main/java/io/opentelemetry/context/PersistentHashArrayMappedTrie.java
@@ -3,6 +3,23 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+// Includes work from:
+/*
+ * Copyright 2015 The gRPC Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package io.opentelemetry.context;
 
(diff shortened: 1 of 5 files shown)


iNikem (approved):

context/src/main/java/io/opentelemetry/context/PersistentHashArrayMappedTrie.java
```diff
@@ -3,6 +3,23 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+// Includes work from:
```
Oberon00, line 6:
```suggestion
// The code in this file was developed by:
```

Or something else to indicate that this file includes *solely* (mostly) work from gRPC might be better.

Oberon00 (approved):

context/src/main/java/io/opentelemetry/context/PersistentHashArrayMappedTrie.java
iNikem: What will be the criterium for "mostly"? When should we use one variant or another?

anuraaga: Yeah I think it's a bit hard to judge, and we may modify the file later. In my experience, I've seen dual-license applied to files that are completely copied, not sure but probably for these reasons.

jkwatson (approved):
Thanks!

context/src/main/java/io/opentelemetry/context/PersistentHashArrayMappedTrie.java
jkwatson: I think just keeping the "Includes work from" is a good all-encompassing bit of verbiage.

PR #1829, opened 2020-10-19
Upgrade to Gradle 6.7 and use toolchains feature to build on 11 and t…

anuraaga:
…est on 11/8.

Gradle 6.7 has a great feature to automatically download toolchains when it can't find them on the system. This can allow easily testing multiple Java versions without build system tweaks. As long as the user has any version of Java installed, Gradle will run, and it will automatically downloads additional needed toolchains, including the primary one (e.g. Java 8 user can run our build and it will use Java 11)

This tweaks the build so now

- Build always uses Java 11. In the future, I would probably bump this to 15 - while runtime tweaks are often backported to older Java, compiler tweaks rarely are. I'd generally recommend using latest Java for building to get the latest compiler tweaks (which are in practice mostly improvements to Javadoc), when targeted to older Java versions and running tests on them there's basically no downside
  - `release` flag is used instead of source/target compatibility to allow Java to enforce classpath version

- Tests default to using Java 11. With the `-PtestAdditionalJavaVersions`, additionally `testJava8` task runs the tests using Java 8. CI enables this flag on ubuntu test runner in GitHub and CircleCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,8 @@ init_task: &init_task
 build_task: &build_task
   name: Build
   command: make test-with-docker
+  # For slow build if cache is cleared
+  no_output_timeout: 60m
 compile_benchmark_task: &compile_benchmark_task
   name: Compile JMH
   command: make benchmark
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,8 +28,9 @@ jobs:
       # Configure the JVM and Gradle to avoid OOM errors
       _JAVA_OPTIONS: "-Xmx1g"
       GRADLE_OPTS: "-Dorg.gradle.daemon=false -Dorg.gradle.workers.max=2"
+      ORG_GRADLE_PROJECT_testAdditionalJavaVersions: true
     docker:
-      - image: circleci/openjdk:8-jdk
+      - image: circleci/openjdk:11.0.6-jdk-buster
     steps:
       - checkout
       - restore_cache:
@@ -56,32 +57,6 @@ jobs:
           paths:
             - ~/.gradle
           key: gradle-{{ checksum "build.gradle" }}
-  java11:
-    environment:
-      # Configure the JVM and Gradle to avoid OOM errors
-      _JAVA_OPTIONS: "-Xmx3g"
-      GRADLE_OPTS: "-Dorg.gradle.daemon=false -Dorg.gradle.workers.max=2"
-    docker:
-      - image: circleci/openjdk:11.0.6-jdk-buster
-    steps:
-      - checkout
-      - restore_cache:
-          keys:
-            - java11-gradle-{{ checksum "build.gradle" }}
-#     make in missing in openjdk:9-jdk image
-      - run: sudo apt-get update && sudo apt-get install -y make
-      - run:
-          <<: *init_task
-      - run:
-          <<: *verify_task
-      - run:
-          <<: *build_task
-      - run:
-          <<: *compile_benchmark_task
-      - save_cache:
-          paths:
-            - ~/.gradle
-          key: java11-gradle-{{ checksum "build.gradle" }}
 # Publish the released artifacts using the `build` job env.
   release_job:
     environment:
@@ -113,4 +88,3 @@ workflows:
   build_and_test:
     jobs:
       - build
-      - java11
--- a/.github/workflows/continuous-build.yml
+++ b/.github/workflows/continuous-build.yml
@@ -18,28 +18,32 @@ jobs:
           - macos-latest
           - ubuntu-latest
           - windows-latest
-        java:
-          # Test all OS's with Java 11
-          - 11
         include:
           - os: ubuntu-latest
-            java: 11
+            testAdditionalJavaVersions: true
             coverage: true
-          - os: ubuntu-latest
-            java: 8
     steps:
       - uses: actions/checkout@v2
         with:
           submodules: true
-      - uses: actions/setup-java@v1
+      - id: setup-java-8
+        name: Setup Java 8
+        uses: actions/setup-java@v1
         with:
-          java-version: ${{ matrix.java }}
+          java-version: 8
+          java-package: jre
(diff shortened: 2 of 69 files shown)


anuraaga:
While debugging CircleCI I noticed we've never built on Java 11 using it due to `executorType: machine` XD Murphy's law of course prevails as now that I thought we can finally delete it, a test flaked on GitHub

anuraaga:
Absolutely no clue why the Windows build seems to have kicked in with this change @inikem @vlsi and it's finding a lot of flakes - Murphy's all over this one

anuraaga:
Green :)

.github/workflows/continuous-build.yml
```diff
@@ -18,28 +18,32 @@ jobs:
...
+            testAdditionalJavaVersions: true
             coverage: true
-          - os: ubuntu-latest
-            java: 8
     steps:
       - uses: actions/checkout@v2
         with:
           submodules: true
-      - uses: actions/setup-java@v1
+      - id: setup-java-8
```
iNikem, line 29: If gradle is supposed to downloaded needed toolchain, why do we provide it manually?

vlsi: Java is required to launch Gradle itself.

examples/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.6-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.7-all.zip
```
vlsi, line 3: Do you really need `all`?
Can you use `bin` distribution?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.6-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.7-all.zip
```
vlsi, line 3: Do you really need `all`?
Can you use `bin` distribution?

.github/workflows/continuous-build.yml
iNikem: For that we have java 11 next to this 8

anuraaga: Yeah but setup-java doesn't download Java it just extracts from a local VM disk very quickly. Figure it's worth saving some downloading.

examples/gradle/wrapper/gradle-wrapper.properties
anuraaga: Back in the day -all was required for intellij completion from what I understand. Now I do see gradle compile stubs on first launch with -bin. Does that mean bin is enough for projects, but not plugin authors? If so it mostly makes sense - but I personally found the time it takes for -bin to generate stubs to be way longer (several minutes on a macbook pro) than just downloading -all so not sure it's a savings in practice. What do you think?

vlsi: Oh, I thought `-all` was generating stubs as well. I did not compare the contents, however, I thought the difference is just source code (*.java files are present in `-all`, and they are not there in `-bin`).

I think `-all` was required earlier for the IDE to navigate to Gradle sources, however, it is not needed now.

vlsi: https://twitter.com/CedricChampeau/status/1113871652809531392

If you are here, then there's Madrid JUG happening right NOW: https://www.meetup.com/madrid-gug/events/273721823

Modernizing the Groovy build by Cédric Champeau, so you can join and ask `-bin` vs `-all` :)

anuraaga: I _think_ I noticed slowness with -bin with the generations of stubs that wasn't there with -all but am not sure, I'll need to try a controlled experiments to confirm.

And as much as I [love](https://www.github.com/curioswitch/curiostack/tree/master/tools%2Fgradle-plugins%2Fgradle-curiostack-plugin) Gradle I was supposed to pass out an hour ago and will promptly do so for tomorrow :) Sounds like a fun event for sure though.

jkwatson:
@anuraaga do you think the builds are stable enough to go with this?

anuraaga:
Seems not on Windows yet - let me temporarily remove it from the config since it wasn't activated before anyways for whatever reason and I'll keep on playing with it.

anuraaga:
Switched to -bin

jkwatson (approved):
Thanks!

.circleci/config.yml
```diff
@@ -28,8 +30,9 @@ jobs:
       # Configure the JVM and Gradle to avoid OOM errors
       _JAVA_OPTIONS: "-Xmx1g"
       GRADLE_OPTS: "-Dorg.gradle.daemon=false -Dorg.gradle.workers.max=2"
+      ORG_GRADLE_PROJECT_testAdditionalJavaVersions: true
     docker:
-      - image: circleci/openjdk:8-jdk
+      - image: circleci/openjdk:11.0.6-jdk-buster
```
pavolloffay, line 35: Are we sure that this image is being used? The line 25 defines executor type as machine which is VM and I think as it is applied at root all jobs will use it.

anuraaga: Right I found out it's not being used. I'm going to be deleting CircleCI build anyways :)

build.gradle
```diff
@@ -169,15 +169,32 @@ configure(opentelemetryProjects) {
         mavenLocal()
     }
 
-    sourceCompatibility = 1.8
-    targetCompatibility = 1.8
```
tylerbenson, line 173: @anuraaga I think by removing this, Intellij doesn't configure the language level correctly.

PR #1831, opened 2020-10-19
Synchronize access in TestLogExporter to make sure memory updates are…

anuraaga:
… synced.

This test flaked in https://github.com/open-telemetry/opentelemetry-java/actions/runs/314900370

I believe since `BatchLogProcessor` uses a worker thread, we need to synchronize accesses in the exporter to ensure updates are visible across threads.


iNikem (approved):

jkwatson (approved):
🤞

PR #1835, opened 2020-10-19
Fix non-null carrier not working in extract.

Oberon00:
The added unit tests for HttpTraceContext demonstrates one case where that is useful. A getter that handles null gracefully would be another case.

Note that inject already allowed null carriers, but extract (where these are actually more likely to happen) did not. For example, if I have a getter that handles null like an empty headerlist, I still get a NPE when using HttpTraceContext. This PR fixes that.

### DIFF
--- a/api/src/test/java/io/opentelemetry/trace/propagation/HttpTraceContextTest.java
+++ b/api/src/test/java/io/opentelemetry/trace/propagation/HttpTraceContextTest.java
@@ -26,7 +26,7 @@ import java.util.LinkedHashMap;
 import java.util.Map;
 import org.junit.jupiter.api.Test;
 
-/** Unit tests for {@link io.opentelemetry.trace.propagation.HttpTraceContext}. */
+/** Unit tests for {@link HttpTraceContext}. */
 class HttpTraceContextTest {
 
   private static final TraceState TRACE_STATE_DEFAULT = TraceState.builder().build();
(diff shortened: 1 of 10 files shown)


extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/OtTracerPropagator.java
```diff
@@ -64,8 +65,9 @@ public static OtTracerPropagator getInstance() {
   }
 
   @Override
-  public <C> Context extract(Context context, C carrier, Getter<C> getter) {
+  public <C> Context extract(Context context, @Nullable C carrier, Getter<C> getter) {
     if (context == null || getter == null) {
+      // TODO Other propagators throw exceptions here
```
Oberon00, line 70: This should probably fixed -- either remove the check and let an NPE occur "naturally" or use Objects.requireNonNull. Or we could change all propagators to do something like here (best with logging though).

In any case, this should be a follow-up.

api/src/test/java/io/opentelemetry/trace/propagation/HttpTraceContextTest.java
```diff
@@ -166,6 +166,18 @@ void extract_SampledContext() {
                 TRACE_ID_BASE16, SPAN_ID_BASE16, SAMPLED_TRACE_OPTIONS, TRACE_STATE_DEFAULT));
   }
 
+  @Test
+  void extract_NullCarrier() {
+    Map<String, String> carrier = new LinkedHashMap<>();
+    carrier.put(TRACE_PARENT, TRACEPARENT_HEADER_SAMPLED);
+    assertThat(
+            getSpanContext(
+                httpTraceContext.extract(Context.current(), null, (c, k) -> carrier.get(k))))
```
Oberon00, line 175: This test here demonstrates one case where a null carrier might be sensibly used.

Oberon00: *sensibly is relative here but at least it demonstrates how checking for null could be surprising

arminru (approved):

jkwatson (approved):
Thanks!

PR #1837, opened 2020-10-20
Fix Logger name

dengliming:


anuraaga (commented):
Thanks! Can you run `./gradlew spotlessApply`?

dengliming:
@anuraaga sure. I will do it later. Thanks for so quick reply.

iNikem (approved):

jkwatson (approved):

PR #1838, opened 2020-10-20
Context.with/get, AttributesBuilder/LabelsBuilder.set

### LINKED ISSUE #1825: Rename Context.withValues to Context.with
I mostly copied the name `withValues` from grpc context, but a Context can't be with anything but a value. How about just `with`?

I think this could be applied to others too, `set` instead of `setAttribute` for example and we should if we think this is in general a good idea.

anuraaga:
A context can only have values, attributesbuilder can only have attributes, labelsbuilder can only have labels, so this reduces some repetition in user code.

Fixes #1825


anuraaga:
Interesting `Attributes.get` was not `Attributes.getAttribute` so seems to add symmetry too.

api/src/main/java/io/opentelemetry/common/Attributes.java
```diff
@@ -231,105 +231,105 @@ public Attributes build() {
...
     }
 
     /**
-     * Add all the provided attributes to this Builder.
+     * Sets all the provided attributes into this Builder.
      *
      * @return this Builder
      */
-    public Builder addAll(Attributes attributes) {
+    public Builder setAll(Attributes attributes) {
```
anuraaga, line 332: @jkwatson Any reason to use `addAll` before? Couldn't think of one so made this symmetrical with above

iNikem: But this does not `set` as in "replace current value", it actually `adds` given attributes to already existing ones. No?

jkwatson: last-in wins, so this will replace current values. `setAll` is a better name. 👍

jkwatson (approved):
Thanks! This is a good cleanup, pre-GA. @carlosalberto and @bogdandrutu please take a look.

jkwatson:
@anuraaga looks like it needs a rebase.

anuraaga:
Thanks, updated

Oberon00:
Shouldn't we use put instead of set, to match the Java collections?

trask:
> Shouldn't we use put instead of set, to match the Java collections?

fwiw, I stumbled when I read `setAll`. I wondered, does it set the entire underlying collection? (meaning wipes out existing values even if they had different keys than the new ones)

I wouldn't have had the same question about `putAll`.

anuraaga:
SGTM, thanks

Issue #1812, opened 2020-10-15
Update Jaeger exporter to conform to updated specification


jkwatson:
Recent clarification on mapping events to Jaeger logs: https://github.com/open-telemetry/opentelemetry-specification/pull/1042

PR #1839, opened 2020-10-20
Update Jaeger exporter to conform to updated specification

dengliming:
resolves [https://github.com/open-telemetry/opentelemetry-java/issues/1812](https://github.com/open-telemetry/opentelemetry-java/issues/1812)

### DIFF
--- a/exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/Adapter.java
+++ b/exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/Adapter.java
@@ -159,7 +159,7 @@ final class Adapter {
     builder.addFields(
         Model.KeyValue.newBuilder().setKey(KEY_LOG_EVENT).setVStr(event.getName()).build());
 
-    int droppedAttributesCount = event.getTotalAttributeCount() - event.getAttributes().size();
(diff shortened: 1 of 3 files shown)


exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/Adapter.java
```diff
@@ -156,7 +157,16 @@ private Adapter() {}
 
     // name is a top-level property in OpenTelemetry
     builder.addFields(
-        Model.KeyValue.newBuilder().setKey(KEY_LOG_MESSAGE).setVStr(event.getName()).build());
+        Model.KeyValue.newBuilder().setKey(KEY_LOG_EVENT).setVStr(event.getName()).build());
+
+    int droppedAttributesCount = event.getTotalAttributeCount() - event.getAttributes().size();
```
jkwatson, line 162: I'd be in favor of adding a method to the Event which does this work for you.

dengliming: 👍  What about adding a default method？
```
default boolean getDroppedAttributesCount() {
      return getTotalAttributeCount() == getAttributes().size();
    }
```

jkwatson (approved):
Thanks!

PR #1840, opened 2020-10-20
Change the sampling API to take a parent Context, rather than SpanContext

jkwatson:
See this recent spec update: https://github.com/open-telemetry/opentelemetry-specification/pull/881


anuraaga (approved):
Dunno why this looks like dejavu

lzchen:
@jkwatson 
You guys are so fast!

PR #1843, opened 2020-10-21
Move Span context accessors to Span static methods.

### LINKED ISSUE #1807: Context usability improvements

anuraaga:
I think there's general consensus in #1807 on having these static methods on `Span`. The one that was open is whether `Span.current()` or `Span.fromCurrentContext()` - I went with the former since personally like it a bit better but happy to rename, it's just a few seconds with IntelliJ :)

The only remaining `TracingContextUtils` method is `currentContextWith`, the vast usage of which is in tests. In a separate PR I'll just remove it and we could provide a helper in a `testing` package if it seems like a useful enough shortcut there.

Fixes #1807


api/src/main/java/io/opentelemetry/trace/DefaultTracer.java
```diff
@@ -54,7 +54,7 @@ public Span startSpan() {
     @Override
     public NoopSpanBuilder setParent(Context context) {
       Objects.requireNonNull(context, "context");
-      spanContext = TracingContextUtils.getSpan(context).getContext();
+      spanContext = Span.fromContext(context).getContext();
```
anuraaga, line 57: Since it's like `SpanContext` won't actually be renamed, `getContext()` desparately needs to be renamed to `getSpanContext()`.

jkwatson (approved):
Thanks! The more TCU we can eliminate the better.l

iNikem (approved):

Oberon00 (approved):
What is left to remove TracingContextUtil completely?

anuraaga:
I will remove `currentContextWith` in a separate PR and then it'll be gone.

api/src/main/java/io/opentelemetry/trace/Span.java
```diff
@@ -22,6 +23,31 @@
 @ThreadSafe
 public interface Span extends ImplicitContextKeyed {
 
+  /**
+   * Returns the {@link Span} from the current {@link Context}, falling back to a default, no-op
+   * {@link Span} if there is no span in the current context.
+   */
+  static Span current() {
```
jkwatson, line 30: Question from the SIG meeting today: would it make sense to have a method on `ImplicitContextKeyed` for retrieving the item from the Context as well? Seems like having both saving and retrieval would be a win.

anuraaga: It came to mind, but these need to be static so can't :)

jkwatson: hmmm. Yeah, no static methods on an interface. This is what happens when I try to reason about things the morning after the 10pm meeting the night before. ;)

jkwatson:
oh noes. needs a rebase, again.

PR #1844, opened 2020-10-21
Fix build breakage due to conflicting merges.

anuraaga:
Sometimes happens


jkwatson (approved):
you beat me to it. Although, I think my formatting was a little nicer.  😉

PR #1846, opened 2020-10-21
Automatically generate Semantic Convention attributes

### LINKED ISSUE #1841: Semantic constants for db.system
Does it make sense have semantic constants somewhere for `db.system`? based on the spec table https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md

| Value  | Description |
|---|---|
| `other_sql` | Some other SQL database. Fallback only. See notes. |
| `mssql` | Microsoft SQL Server |
| `mysql` | MySQL |
| `oracle` | Oracle Database |
| `db2` | IBM Db2 |
| `postgresql` | PostgreSQL |
| `redshift` | Amazon Redshift |
| `hive` | Apache Hive |
| `cloudscape` | Cloudscape |
| `hsqlsb` | HyperSQL DataBase |
| `progress` | Progress Database |
| `maxdb` | SAP MaxDB |
| `hanadb` | SAP HANA |
| `ingres` | Ingres |
| `firstsql` | FirstSQL |
| `edb` | EnterpriseDB |
| `cache` | InterSystems Caché |
| `adabas` | Adabas (Adaptable Database System) |
| `firebird` | Firebird |
| `derby` | Apache Derby |
| `filemaker` | FileMaker |
| `informix` | Informix |
| `instantdb` | InstantDB |
| `interbase` | InterBase |
| `mariadb` | MariaDB |
| `netezza` | Netezza |
| `pervasive` | Pervasive PSQL |
| `pointbase` | PointBase |
| `sqlite` | SQLite |
| `sybase` | Sybase |
| `teradata` | Teradata |
| `vertica` | Vertica |
| `h2` | H2 |
| `coldfusion` | ColdFusion IMQ |
| `cassandra` | Apache Cassandra |
| `hbase` | Apache HBase |
| `mongodb` | MongoDB |
| `redis` | Redis |
| `couchbase` | Couchbase |
| `couchdb` | CouchDB |
| `cosmosdb` | Microsoft Azure Cosmos DB |
| `dynamodb` | Amazon DynamoDB |
| `neo4j` | Neo4j |

thisthat:
Fixes #1841. I am suffixing this class with `V2` because `EXCEPTION_EVENT_NAME` is not part of the YAML definitions. I will follow up by refactoring the code and moving that constant somewhere.

Does this solve your issue @trask? Or do you prefer having a new class for just the enum values?
For your example, the enum is available here: https://github.com/open-telemetry/opentelemetry-java/pull/1846/files#diff-50fb78fbdb13d122ada839fd124b2beffec95fe51107c23cffc9a5e48ca3e4ceR714

As future work, I am leaving the definition of the proper workflow to follow. ATM, I have the following in mind:
- Manually run the `generate.sh` script provided within this PR
- Use git submodules as we do for the protobuf definitions and define a gradle task
- Create a GitHub action that opens a PR in otel-java with the newly generated java class(es) every time `master` in otel-spec updates a YAML file


Oberon00:
> EXCEPTION_EVENT_NAME is not part of the YAML definitions. I will follow up by refactoring the code and moving that constant somewhere.

That's fine for now, but in the mid-term we should add semantic conventions for events to the generator. Are the event attributes implemented as span attributes in the YAML definitions currently?

api/src/main/java/io/opentelemetry/trace/attributes/SemanticAttributesV2.java
```diff
@@ -0,0 +1,1006 @@
...
+
+package io.opentelemetry.trace.attributes;
+
+import static io.opentelemetry.common.AttributeKey.booleanKey;
+import static io.opentelemetry.common.AttributeKey.longKey;
+import static io.opentelemetry.common.AttributeKey.stringKey;
+
+import io.opentelemetry.common.AttributeKey;
+
+// DO NOT EDIT, this is an Auto-generated file from /templates/SemanticAttributesV2.java.j2
```
Oberon00, line 14: I think it should be relatively straightforward to generate this in gradle using e.g. <https://docs.gradle.org/current/dsl/org.gradle.api.tasks.Exec.html>. Then we would not have to check in this file.

On the minus side, we would then require contributors to have a recent Python 3 installed (actually Docker currently which makes it un-buildable on Windows 🤔).

If we don't auto-generate the file, we should check during build that the file matches the generated output (does the tool support that for any generator currently or only markdown?)

thisthat:
There is no event in YAML right now. The only event is defined by [Exception SemConv](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/exceptions.md#recording-an-exception) and it is just a text line. If it would have been defined in the YAML defs it would appear in the generated class!

semantic_convention/generate.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+rm -rf opentelemetry-specification || true
+git clone https://github.com/open-telemetry/opentelemetry-specification.git --depth 1
```
Oberon00, line 6: We should pin this to a certain commit, or check out a release branch to have a latest 0.x version at least.

Oberon00:
@thisthat 
> There is no event in YAML right now.

Right, that's what I meant here:

> That's fine for now, but in the mid-term we should add semantic conventions for events to the YAML model/generator.

By that I meant we need to extend the *model*, i.e. the syntax for the YAML to support that.

thisthat:
I think this is outside the scope of this PR. I have opened https://github.com/open-telemetry/build-tools/issues/17 for tracking this.

Oberon00:
Of course this is outside the scope of this PR 😃  But since you mentioned that you will work around this in the description, I brought it up.

jkwatson:
1) I really don't like the V2 suffix. It makes it sound like we're on version 2 of the semantic conventions, which we definitely are not. I'd rather this PR rename the existing class, if we intend the yaml-generated ones to the be true source of truth.
2) I'd prefer checking in this file, rather than generating it at build time. Yet another build step that can go wrong doesn't seem like an improvement.
3) I'd prefer not to introduce another git submodule, if we can possibly help it. I'd rather that the code that generates the file pulled down a specific version from GH if you wanted to regenerate the file, rather than using a submodule for it.

trask:
Thanks @thisthat!

Do you think enum is useful for "db.system" values (over string constants), since we can't enforce enum usage since it's [a non-exhaustive list](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md#notes-and-well-known-identifiers-for-dbsystem)?

thisthat:
@trask I opted for an enum for simplicity of the template. The strings can be retrieved via `getValue()`. Would you prefer for open enums to have public constants? If so, I can adapt the template and differentiate between the two cases :) 

@jkwatson I tend to agree with you. I'll refactor the code and remove the V2 as part of this PR then :)

trask:
Would you prefer for open enums to have public constants? If so, I can adapt the template and differentiate between the two cases :)

Ya, the enum makes it look like an exhaustive list, so I'd probably opt not to use enum in this case 👍

Oberon00:
BTW, regarding EXCEPTION_EVENT_NAME: We could write that literally in the template for now.

semantic_convention/templates/SemanticAttributesV2.java.j2
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+
+{%- macro to_java_type(type) -%}
+  {%- if type == "string" -%}
+    String
+  {%- elif type == "boolean" -%}
+    boolean
+  {%- elif type == "number" -%}
+    long
```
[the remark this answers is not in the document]
thisthat: Please see https://github.com/open-telemetry/build-tools/issues/13

jkwatson:
> BTW, regarding EXCEPTION_EVENT_NAME: We could write that literally in the template for now.

yeah, adding a couple of custom pre-specced items to the template itself seems like a fine approach to me.

jkwatson (approved):
Thanks!

anuraaga (approved):
This is great, thanks

PR #1852, opened 2020-10-22
Rename getContext to getSpanContext.

anuraaga:
To prevent the extremely confusing `Span.fromContext(context).getContext()`. Context has a different meaning in our codebase so any `getContext` should have a return type of `Context`.


jkwatson (approved):
Thanks!

PR #1853, opened 2020-10-22
Remove CircleCI build.

anuraaga:
GitHub Actions build is working great, seems to be faster and has a great ecosystem where we can expect even better things to come.


jkwatson (approved):
Thanks!  ;)

PR #1855, opened 2020-10-22
Replace TracingContextUtils.currentContextWith with ImplicitContextKeyed.makeCurrent

anuraaga:
I will remove the `TracingContextUtils` class itself in a followup (last in this series of PRs :D)


api/src/jmh/java/io/opentelemetry/trace/DefaultTracerBenchmarks.java
```diff
@@ -34,7 +35,7 @@
   @Warmup(iterations = 5, time = 1)
   public void measureFullSpanLifecycle() {
     span = tracer.spanBuilder("span").startSpan();
-    try (io.opentelemetry.context.Scope ignored = TracingContextUtils.currentContextWith(span)) {
+    try (io.opentelemetry.context.Scope ignored = Context.current().with(span).makeCurrent()) {
```
jkwatson, line 38: would it be possible to make this also a method on the span instance?

I'm struggling with what the name of that method would be right now, though.
`span.activate()` ?

anuraaga: Oh yeah did have that idea - not sure if it's worth it. But I would just use the same name as `Context`, so `span.makeCurrent()`. Does it seem useful to have?

jkwatson: or maybe `Context.inCurrentScope(span)` ?

anuraaga: I believe that looks like

```java
interface Context
  static Scope inCurrentScope(ImplicitContextKeyed value) {
    return Context.current().with(value).makeCurrent();
  }
```

For symmetry, I think we would need a version that also accepts `ContextKey`. I think I like this a bit less than adding to `Span` itself (rather, I could add it to `ImplicitContextKeyed`).

anuraaga:
Went ahead and added `makeCurrent` let me know how you like it

jkwatson (approved):
Thanks! I like this a lot.

HaloFour:
I have a lot of unit tests where I mock a `Span` instance and put it into the current `Context`.  The issue with the new Context APIs in the 0.10.0 snapshot is that putting a `Span` into a `Context` requires calling an instance method on `Span`, and there are no alternative public APIs.  This is making it very difficult to write unit tests that test logic that uses the `Span` from the current `Context`.  Is it possible to have an API to put a `Span` into a `Context` without it calling `Span#storeInContext`?

jkwatson:
> I have a lot of unit tests where I mock a `Span` instance and put it into the current `Context`. The issue with the new Context APIs in the 0.10.0 snapshot is that putting a `Span` into a `Context` requires calling an instance method on `Span`, and there are no alternative public APIs. This is making it very difficult to write unit tests that test logic that uses the `Span` from the current `Context`. Is it possible to have an API to put a `Span` into a `Context` without it calling `Span#storeInContext`?

Can you share what these unit tests look like? We're definitely revamping the context-related APIs in 0.10.0 and we're hoping these will be close the final APIs, so if there's something we can do to facilitate testing, we'll do what we can.

HaloFour:
@jkwatson 

>

Here's one of them:

https://github.com/Comcast/money/blob/caa8a00a79afb0cbebec3c76edad43a61ce4510b/money-spring/src/test/java/com/comcast/money/spring/MoneyClientHttpInterceptorSpec.java

We have a number of others that rely on obtaining the current span via `Tracer#getCurrentSpan` or setting the current context via `Tracer#withSpan` which have both been removed by OpenTelemetry 0.10.0 but I've kept in this branch to reduce the amount of breaks I have to deal with at this time:

https://github.com/Comcast/money/blob/caa8a00a79afb0cbebec3c76edad43a61ce4510b/money-core/src/test/scala/com/comcast/money/core/logging/MethodTracerSpec.scala

To give you some context about the project I'm trying to take an older tracing library that is used by at least several dozen existing services at our company (and in various stages of maintenance) and make it an OpenTelemetry-compatible tracer.  The goal is to make it as easy as revving the version number and recompiling to make these services able to participate in the larger OpenTelemetry ecosystem as well as offering a way to either opt-in to portions of the OpenTelemetry ecosystem or to switch wholesale to another OpenTelemetry-compatible implementation.  I don't know if that makes us a unicorn project in that regard, especially since I've taken the approach of blending the APIs and replacing bits of our implementation with the OpenTelemetry API and SDK implementations.

anuraaga:
Hi @HaloFour - does it work if you call `when(span.storeInContext(any()).thenReturnRealMethod()`? That should allow the normal logic to run to let context be stored.

It's sort of strange that mockito doesn't do this by default since a mock implementation should act like any other implementation of an interface, which calls default methods unless told otherwise. If this works, I can see if there's any way to make this easier on the mockito side.

> but I've kept in this branch to reduce the amount of breaks I have to deal with at this time:

Sorry for causing so many breaks 😞 But we're still not at 1.0 and this is the only time we can make changes so you may be seeing some more (hopefully we're almost there though). This is so that after 1.0 we don't have to break anything :)

anuraaga:
Found that mockito took the approach of using `spy` instead of `mock` since that's always how they handled abstract classes

https://github.com/mockito/mockito/issues/680

I guess in the Java 8 world mock is far less useful than spy. @HaloFour does it seem reasonable to replace mock with spy in your tests?

HaloFour:
@anuraaga 

> That should allow the normal logic to run to let context be stored.

Ah yes, you can call through to a default method on a mocked interface with Mockito 2.  Unfortunately this project is still using Mockito 1, so I guess I'll have some migrating to do.  😉 

> Sorry for causing so many breaks 😞 But we're still not at 1.0 and this is the only time we can make changes so you may be seeing some more (hopefully we're almost there though). This is so that after 1.0 we don't have to break anything :)

No worries, I know what I signed up for. 😁 

Thanks!

PR #1856, opened 2020-10-22
Increase Jaeger container timeout

anuraaga:
This flaky test I actually saw locally too, not just GitHub :) For now a random attempt of increasing the timeout to see what happens.


jkwatson (approved):
2 minutes!

anuraaga:
I'm suspecting docker pull can be slow - might need to look at caching `.docker` or something like that

Issue #1834, opened 2020-10-19
Using the tracing SDK module on its own has some usability issues


Oberon00:
The trace SDK needs to be configured using the `TracerSdkManagement` interface (e.g. to add a span processor). However, this is only possible with the `OpenTelemetrySdk` class https://github.com/open-telemetry/opentelemetry-java/blob/fbfba341349379e423affdc965e79071b41c3372/sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java#L31-L32 which is in opentelemetry-sdk-all and depends on all other SDK componentes. One cannot even cast the return value of `OpenTelemetry.getTracerProvider()` to `TracerSdkManagement` due to the internal `Obfuscated` class (which is good in it itself).

Oberon00:
IMHO this would best be fixed by removing the OpenTelemetrySdk and splitting it up into one separate class per module.

anuraaga:
@Oberon00 What do you think about continuing to depend on the SDK but excluding the dependencies you don't want with gradle / maven exclusions? I think that should already work, though I haven't tried it.

I think having `OpenTelemetrySdk` should be fine still to manage shared SDK concerns like `Resource`, but we can probably also add individual accessors in each module if that seems like better UX. But I guess when wanting to trim dependencies, dependency exclusion is relatively idiomatic? Not sure just food for thought :)

Oberon00:
> excluding the dependencies you don't want with gradle / maven exclusions

That cannot work. This is not merely a compile-time dependency declaration problem. If I load the OpenTelemetrySdk, class it *will* reference the baggage classes so a NoClassDefFound error will be thrown I think.

anuraaga:
@Oberon00 Java is suprisingly lazy :) Classes shouldn't be loaded until they're used. Definitely adding examples for partial SDK usage is on my list of follow up PRs with #1857 - my plan is for this to work fine even with dependencies excluded. If I find something wrong will fix it :) 

`OpenTelemetrySdk.builder().setTracerProvider(TracerSdkProvider.builder().build()).setMeterProvider(DefaultMeterProvider.getInstance()).build()`

Oberon00:
OK this still is dirty. If I accidentally call OpenTelemetrySdk.getMeterProvider, the compiler will not warn me but it will fail at runtime, right? Besides, what is the benefit of having a shared class here over one class per module?

anuraaga:
[...]

PR #1857, opened 2020-10-22
Make OpenTelemetrySdk implement the OpenTelemetry interface and have SPI find it.

### LINKED ISSUE #747: Registries created through builders should be returned by OpenTelemetry static methods

### LINKED ISSUE #1109: Tracing multiple applications in application servers

### LINKED ISSUE #1149: Add setup method to configure the OpenTelemetry API

### LINKED ISSUE #1454: Being able to specify Resource manually

anuraaga:
See #1737 for the discussion.

I don't want to merge a bad PR and a followup renaming like last time since it's confusing and a bit too tedious. I have two commits here, https://github.com/open-telemetry/opentelemetry-java/commit/949bfec50307d4da4a8cd028da3bf617049b8224 is the significant one to check out, the other just applies renames.

Fixes #1109 (each app can have its own OpenTelemetry for different Resource which seems to be the crux of the issue)
Fixes #1149
Fixes #1454
Fixes #747

### DIFF
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -500,4 +500,3 @@ public class IgnoreExportErrorsFilter implements Filter {
 ## Registering the custom filter on the BatchSpanProcessor
 io.opentelemetry.sdk.trace.export.BatchSpanProcessor = io.opentelemetry.extensions.logging.IgnoreExportErrorsFilter
 ```
-
--- a/api/src/main/java/io/opentelemetry/OpenTelemetry.java
+++ b/api/src/main/java/io/opentelemetry/OpenTelemetry.java
@@ -16,15 +16,47 @@ import io.opentelemetry.trace.TracerProvider;
 /**
  * The entrypoint to telemetry functionality for tracing, metrics and baggage.
  *
- * <p>The default for the OpenTelemetry API will include any {@link
+ * <p>A global singleton can be retrieved by {@link #get()}. The default for the returned {@link
+ * OpenTelemetry}, if none has been set via {@link #set(OpenTelemetry)}, will be created with any
+ * {@link io.opentelemetry.spi.OpenTelemetryFactory}, {@link
  * io.opentelemetry.trace.spi.TracerProviderFactory} or {@link
  * io.opentelemetry.metrics.spi.MeterProviderFactory} found on the classpath, or otherwise will be
  * default, with no-op behavior.
+ *
+ * <p>If using the OpenTelemetry SDK, you may want to instantiate the {@link OpenTelemetry} to
+ * provide configuration, for example of {@code Resource} or {@code Sampler}. See {@code
+ * OpenTelemetrySdk} and {@code OpenTelemetrySdk.builder} for information on how to construct the
+ * SDK {@link OpenTelemetry}.
  *
  * @see TracerProvider
  * @see MeterProvider
  */
 public interface OpenTelemetry {
+
+  /**
+   * Returns the registered global {@link OpenTelemetry}. If no call to {@link #set(OpenTelemetry)}
+   * has been made so far, a default {@link OpenTelemetry} composed of functionality any {@link
+   * io.opentelemetry.spi.OpenTelemetryFactory}, {@link
+   * io.opentelemetry.trace.spi.TracerProviderFactory} or{@link
+   * io.opentelemetry.metrics.spi.MeterProviderFactory}, found on the classpath, or otherwise will
+   * be default, with no-op behavior.
+   *
+   * @throws IllegalStateException if a provider has been specified by system property using the
+   *     interface FQCN but the specified provider cannot be found.
+   */
+  static OpenTelemetry get() {
+    return DefaultOpenTelemetry.getGlobalOpenTelemetry();
+  }
+
+  /**
+   * Sets the {@link OpenTelemetry} that should be the global instance. Future calls to {@link
+   * #get()} will return the provided {@link OpenTelemetry} instance. This should be called once as
+   * early as possible in your application initialization logic, often in a {@code static} block in
+   * your main class.
+   */
+  static void set(OpenTelemetry openTelemetry) {
+    DefaultOpenTelemetry.setGlobalOpenTelemetry(openTelemetry);
+  }
 
   /** Returns the globally registered {@link TracerProvider}. */
   static TracerProvider getGlobalTracerProvider() {
@@ -109,9 +141,7 @@ public interface OpenTelemetry {
    */
   static void setGlobalPropagators(ContextPropagators propagators) {
     requireNonNull(propagators, "propagators");
-    DefaultOpenTelemetry.setGlobalOpenTelemetry(
-        ((DefaultOpenTelemetry) DefaultOpenTelemetry.getGlobalOpenTelemetry())
-            .toBuilder().setPropagators(propagators).build());
+    set(get().toBuilder().setPropagators(propagators).build());
   }
 
   /** Returns the {@link TracerProvider} for this {@link OpenTelemetry}. */
@@ -173,4 +203,26 @@ public interface OpenTelemetry {
 
   /** Returns the {@link ContextPropagators} for this {@link OpenTelemetry}. */
   ContextPropagators getPropagators();
+
+  /** Returns a new {@link Builder} with the configuration of this {@link OpenTelemetry}. */
+  Builder<?> toBuilder();
+
+  /**
+   * A builder of an implementation of the OpenTelemetry API. Generally used to reconfigure SDK
+   * implementations.
+   */
+  interface Builder<T extends Builder<T>> {
+
+    /** Sets the {@link TracerProvider} to use. */
+    T setTracerProvider(TracerProvider tracerProvider);
+
+    /** Sets the {@link MeterProvider} to use. */
+    T setMeterProvider(MeterProvider meterProvider);
+
+    /** Sets the {@link ContextPropagators} to use. */
+    T setPropagators(ContextPropagators propagators);
+
+    /** Returns a new {@link OpenTelemetry} based on the configuration in this {@link Builder}. */
+    OpenTelemetry build();
+  }
 }
--- a/api/src/main/java/io/opentelemetry/spi/OpenTelemetryFactory.java
+++ b/api/src/main/java/io/opentelemetry/spi/OpenTelemetryFactory.java
@@ -1,0 +1,22 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.spi;
+
+import io.opentelemetry.OpenTelemetry;
+
+/**
+ * A service provider interface (SPI) for a {@link OpenTelemetry}. Fully qualified class name of the
+ * implementation should be registered in a resource file {@code
+ * META-INF/services/io.opentelemetry.spi.OpenTelemetryFactory}.
+ *
+ * <p>A specific implementation can be selected by setting the system property {@code
+ * io.opentelemetry.spi.OpenTelemetryFactory} with the value of the fully qualified class name.
+ */
+public interface OpenTelemetryFactory {
+
+  /** Returns a new {@link OpenTelemetry} instance. */
+  OpenTelemetry create();
+}
--- a/api/src/main/java/io/opentelemetry/DefaultOpenTelemetry.java
+++ b/api/src/main/java/io/opentelemetry/DefaultOpenTelemetry.java
@@ -7,16 +7,13 @@ package io.opentelemetry;
 
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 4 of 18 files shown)


anuraaga:
John has suggested using `DefaultOpenTelemetry` within the SDK, or other, implementations of the API. I think it's a good idea but requires making it public - I'd like to address that in a separate PR.

Oberon00:
Have you considered #1834? How would we solve this after this PR?

anuraaga:
Added some comments in the issue - I am thinking that regardless we do want to have a single entry point into the SDK. It's the very common use case so presenting the single entry point simplifies the experience. But we may be able to improve the ux for custom sdk users by adding similar entry points into the SDK components. I don't believe that this change would hinder such improvements and want to make those with the understanding that we do want to optimize for the common case of full SDK if we ever make a trade-off.

QUICKSTART.md
```diff
@@ -31,10 +32,34 @@ configuration must be provided by **Applications** which should also depend on t
 libraries will obtain a real implementation only if the user application is configured for it. For
 more details, check out the [Library Guidelines].
 
+## Setting up the SDK
+
+To get started, setup the OpenTelemetry SDK - note that if you are using a different implementation 
+of the OpenTelemetry API, you should refer to the configuration instructions for that SDK. Also, if
+you are using the Auto-Instrumentation Agent, the SDK is configured by the agent itself and you
+should not setup your own.
```
jkwatson, line 40:
```suggestion
should not set up your own.
```

QUICKSTART.md
```diff
@@ -31,10 +32,34 @@ configuration must be provided by **Applications** which should also depend on t
 libraries will obtain a real implementation only if the user application is configured for it. For
 more details, check out the [Library Guidelines].
 
+## Setting up the SDK
+
+To get started, setup the OpenTelemetry SDK - note that if you are using a different implementation
```
jkwatson, line 37:
```suggestion
To get started, set up the OpenTelemetry SDK - note that if you are using a different implementation 
```

QUICKSTART.md
```diff
@@ -31,10 +32,34 @@ configuration must be provided by **Applications** which should also depend on t
...
 more details, check out the [Library Guidelines].
 
+## Setting up the SDK
+
+To get started, setup the OpenTelemetry SDK - note that if you are using a different implementation 
+of the OpenTelemetry API, you should refer to the configuration instructions for that SDK. Also, if
+you are using the Auto-Instrumentation Agent, the SDK is configured by the agent itself and you
+should not setup your own.
+
+You can setup a basic configuration that exports traces to a logging stream like this:
```
jkwatson, line 42:
```suggestion
You can set up a basic configuration that exports traces to a logging stream like this:
```

QUICKSTART.md
```diff
@@ -31,10 +32,34 @@ configuration must be provided by **Applications** which should also depend on t
...
+        TracerSdkProvider.builder()
+            .setSpanProcessors(
+                SimpleSpanProcessor.builder(new LoggingSpanExporter()).build())
+            .build())
+    .build();
+```
+
+Note that a preconfigured `OpenTelemetry` can be retrieved by calling `OpenTelemetry.get()`. If you
+are using the Auto-Instrumentation Agent, it will be the `OpenTelemetry` that the agent configured.
+Otherwise, it will enable propagation of context to remote servers but will not export any spans.
```
jkwatson, line 56: I'd make it clear that this is only if you haven't manually installed an SDK.

QUICKSTART.md
```diff
@@ -43,15 +68,15 @@ monitored. More information is available in the specification chapter [Obtaining
...
+    openTelemetry.getTracer("instrumentation-library-name","semver:1.0.0");
 ```
 
 ### Create basic Span
 To create a basic span, you only need to specify the name of the span.
 The start and end time of the span is automatically set by the OpenTelemetry SDK.
 ```java
 Span span = tracer.spanBuilder("my span").startSpan();
-try (Scope scope = TracingContextUtils.currentContextWith(span)) {
+try (Scope scope = tracer.withSpan(span)) {
```
jkwatson, line 79: `tracer.withSpan` is gone. This should be `Context.with(span).makeCurrent()`, right?

QUICKSTART.md
```diff
@@ -421,17 +434,23 @@ builder.readEnvironmentVariables()
...
 |----------------------------------|----------------------------------|-----------------------------------------------------------------------------------------------------|       
 | otel.config.sampler.probability  | OTEL_CONFIG_SAMPLER_PROBABILITY  | Sampler which is used when constructing a new span (default: 1)                                     |                        
-| otel.span.attribute.count.limit  | OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT  | Max number of attributes per span, extra will be dropped (default: 1000)                              |                        
-| otel.span.event.count.limit      | OTEL_SPAN_EVENT_COUNT_LIMIT      | Max number of Events per span, extra will be dropped (default: 1000)                                 |                        
-| otel.span.link.count.limit       | OTEL_SPAN_LINK_COUNT_LIMIT       | Max number of Link entries per span, extra will be dropped (default: 1000)                            |
+| otel.config.max.attrs            | OTEL_CONFIG_MAX_ATTRS            | Max number of attributes per span, extra will be dropped (default: 32)                              |
```
jkwatson, line 451: was this supposed to sneak into this PR?

api/src/main/java/io/opentelemetry/OpenTelemetry.java
```diff
@@ -17,18 +16,47 @@
 /**
  * The entrypoint to telemetry functionality for tracing, metrics and baggage.
  *
- * <p>The default for the OpenTelemetry API will include any {@link
- * io.opentelemetry.trace.spi.TracerProviderFactory}, {@link
- * io.opentelemetry.metrics.spi.MeterProviderFactory}, or {@link
- * io.opentelemetry.baggage.spi.BaggageManagerFactory} found on the classpath, or otherwise will be
+ * <p>A global singleton can be retrieved by {@link #get()}. The default for the returned {@link
+ * OpenTelemetry}, if none has been set via {@link #set(OpenTelemetry)} will be created with any
```
jkwatson, line 20:
```suggestion
 * OpenTelemetry}, if none has been set via {@link #set(OpenTelemetry)}, will be created with any
```

api/src/main/java/io/opentelemetry/OpenTelemetry.java
```diff
@@ -17,18 +16,47 @@
...
+   * @throws IllegalStateException if a provider has been specified by system property using the
+   *     interface FQCN but the specified provider cannot be found.
+   */
+  static OpenTelemetry get() {
+    return DefaultOpenTelemetry.getGlobalOpenTelemetry();
+  }
+
+  /**
+   * Sets the {@link OpenTelemetry} that should be the global instance. Future calls to {@link
+   * #get()} will return the {@link OpenTelemetry}. This should be called once as early as possible
```
jkwatson, line 53:
```suggestion
   * #get()} will return the provided {@link OpenTelemetry} instance. This should be called once as early as possible
```

api/src/main/java/io/opentelemetry/spi/OpenTelemetryFactory.java
```diff
@@ -0,0 +1,22 @@
...
+
+import io.opentelemetry.OpenTelemetry;
+
+/**
+ * A service provider interface (SPI) for a {@link OpenTelemetry}. Fully qualified class name of the
+ * implementation should be registered in a resource file {@code
+ * META-INF/services/io.opentelemetry.spi.OpenTelemetryFactory}.
+ *
+ * <p>A specific implementation can be selected by setting the system property {@code
+ * io.opentelemetry.spi.OpenTelemetryFactor} with the value of the fully qualified class name.
```
jkwatson, line 16:
```suggestion
 * io.opentelemetry.spi.OpenTelemetryFactory} with the value of the fully qualified class name.
```

api/src/main/java/io/opentelemetry/spi/OpenTelemetryFactory.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+public interface OpenTelemetryFactory {
+
+  /** Returns a new {@link OpenTelemetry}. */
```
jkwatson, line 20:
```suggestion
  /** Returns a new {@link OpenTelemetry} instance. */
```

examples/http/src/main/java/io/opentelemetry/example/http/HttpClient.java
```diff
@@ -37,7 +37,7 @@
 
   private static void initTracerSdk() {
     // Get the tracer management instance.
-    TracerSdkManagement tracerManagement = OpenTelemetrySdk.getTracerManagement();
+    TracerSdkManagement tracerManagement = $1erManagement();
```
jkwatson, line 40: ???

examples/grpc/src/main/java/io/opentelemetry/example/HelloWorldServer.java
```diff
@@ -91,7 +91,7 @@ private void stop() {
 
   private void initTracer() {
     // Get the tracer management instance
-    TracerSdkManagement tracerManagement = OpenTelemetrySdk.getTracerManagement();
+    TracerSdkManagement tracerManagement = OpenTelemetrySdk.getGlobalTracerManagement();
```
jkwatson, line 94: we shouldn't update examples until we do a release, correct? They should be pointing at the latest release, not a snapshot.

jkwatson:
Thanks! This needs a few cleanups, but it's looking pretty solid to me.

QUICKSTART.md
anuraaga: And something else soon - I'm taking this change out of this PR anyways since we'd want to do it with the next release.

anuraaga (commented):
Reverted QUICKSTART change since we'll do that with the release - copied it to https://gist.github.com/anuraaga/e53f5c226624b6efd83ceb276438ebd4

jkwatson (approved):
Looks good. Thanks!

PR #1859, opened 2020-10-22
Add a factory method for creating PropagatedSpan to allow auto-instru…

anuraaga:
…mentation to intercept the method.

Due to a corner case, auto instrumentation currently needs this factory method.


trask (approved):

jkwatson (approved):
Thanks!

PR #1863, opened 2020-10-22
Add OpenCensus shim

zoercai:
The [OpenTelemetry merger announcement](https://opensource.googleblog.com/2019/05/opentelemetry-merger-of-opencensus-and.html) promised straightforward backwards compatibility with OpenCensus.

This PR adds an OpenCensus shim, which allows applications and libraries that are instrumented with OpenTelemetry, but depend on other libraries instrumented with OpenCensus,to export trace spans from both OpenTelemetry and OpenCensus
with the correct parent-child relationship.

The shim maps all OpenCensus spans to OpenTelemetry spans, which are then exported by any configured OpenTelemetry exporter.

For OpenCensus spans to be exported with the correct parent-child relationship with OpenTelemetry spans, this shim needs to be added as a dependency. No code change is necessary, as the OpenCensus behaviour is overridden automatically via reflection.

cc @nilebox @james-bebbington


opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/SpanConverter.java
```diff
@@ -0,0 +1,246 @@
...
+  static io.opentelemetry.trace.Span toOtelSpan(Span span) {
+    if (span == null) {
+      return null;
+    }
+    SpanData spanData = ((RecordEventsSpanImpl) span).toSpanData();
+    io.opentelemetry.trace.Span.Builder builder =
+        TRACER
+            .spanBuilder(spanData.getName())
+            .setStartTimestamp(
+                spanData.getStartTimestamp().getSeconds() * NANOS_PER_SECOND
```
jkwatson, line 73: What about using `TimeUnit.SECONDS.toNanos()` here?

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryCtx.java
```diff
@@ -0,0 +1,36 @@
...
+package io.opentelemetry.opencensusshim;
+
+import io.opencensus.trace.ContextHandle;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.Scope;
+
+public class OpenTelemetryCtx implements ContextHandle {
+
+  private final Context context;
+  private Scope scope;
```
jkwatson, line 15: I'm somewhat nervous about holding on to a Scope reference like this. Can we guarantee that attach and detach are always called on the same thread, and no other detach calls will be made on that thread in between the attach/detach pairs? I don't know enough (or anything) about Census to answer this question, but it makes me nervous.

jkwatson:
Thanks for this work, @zoercai . Does census have some sort of stress test that we could apply to the shim to make sure that the context wrapping/unwrapping isn't going to end up being an issue?

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryCtx.java
nilebox: FYI the attach/detach methods are an abstraction for gRPC context used in OpenCensus by default:
- https://grpc.github.io/grpc-java/javadoc/io/grpc/Context.html#attach--
- https://grpc.github.io/grpc-java/javadoc/io/grpc/Context.html#detach-io.grpc.Context-

> This context should be the same context that was previously attached. The provided replacement should be what was returned by the same attach() call. If an attach() and a detach() meet above requirements, they match.

> It is expected that between any pair of matching attach() and detach(), all attach()es and detach()es are called in matching pairs. If this method finds that this context is not current, either you or some code in-between are not detaching correctly, and a SEVERE message will be logged but the context to attach will still be bound. Never use Context.current().detach(), as this will compromise this error-detecting mechanism.

Are there any safeguards we can introduce to make sure we have the same semantics in this shim implementation?

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryCtx.java
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  Context getContext() {
+    return context;
+  }
+
+  @Override
+  public ContextHandle attach() {
+    scope = context.makeCurrent();
+    return this;
```
nilebox, line 28: From https://grpc.github.io/grpc-java/javadoc/io/grpc/Context.html#attach--

> The previously current context is returned.

We may need to create a "child" context here somehow?

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryCtx.java
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  @Override
+  public void detach(ContextHandle ctx) {
+    OpenTelemetryCtx impl = (OpenTelemetryCtx) ctx;
```
nilebox, line 33: Note that we are closing the scope of a context passed as an argument here, not the scope of `this.scope`.
Which seems to imply that we should implement a "nesting" mechanism (using a Stack data structure probably), e.g.

```
create context
-- attach - creates root scope 1
   -- attach - creates child scope 2
       -- attach creates a sub-child scope 3
       -- detach closes scope 3 and restores scope 2 as current
   -- detach closes scope 2 and restores scope 1 as current
-- detach closes scope 1 and there is no current scope anymore
```

jkwatson: The OpenTelemetry context has the same semantics as the grpc context (the implementation was heavily inspired by it), so I think this should all "just work" as long as everything is kept in alignment the same way.

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryTraceComponentImpl.java
```diff
@@ -0,0 +1,62 @@
...
+  private final Clock clock;
+  private final TraceConfig traceConfig = new TraceConfigImpl();
+  private final Tracer tracer;
+
+  /** Public constructor to be used with reflection loading. */
+  public OpenTelemetryTraceComponentImpl() {
+    clock = MillisClock.getInstance();
+    RandomHandler randomHandler = new ThreadLocalRandomHandler();
+    StartEndHandler startEndHandler = new OpenTelemetryStartEndHandler();
+    tracer = new TracerImpl(randomHandler, startEndHandler, clock, traceConfig);
```
anuraaga, line 34: Instead of using opencensus `TracerImpl`, can you use a shim implementation that delegates to a opentelemetry tracer? This will remove the need for the `SpanCache` which looks like it can affect performance a lot, the custom OC Tracer would return OC Spans that just wrap the OTel spans. FWIU, the opentracing shim takes such an approach so might be a good reference.

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryContextManager.java
```diff
@@ -0,0 +1,46 @@
...
+
+package io.opentelemetry.opencensusshim;
+
+import io.opencensus.trace.ContextHandle;
+import io.opencensus.trace.ContextManager;
+import io.opencensus.trace.Span;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.trace.TracingContextUtils;
+
+public class OpenTelemetryContextManager implements ContextManager {
```
anuraaga, line 14: Have you considered the opposite approach, having OTel context use the OpenCensus context? Scope is more flexible than returning previousCtx, so the bridge can be a bit cleaner than the other way around. I suspect this code could be adapted to opencensus context trivially.

https://github.com/open-telemetry/opentelemetry-java/blob/master/context/src/otelAsGrpcTest/java/io/opentelemetry/context/GrpcContextStorageProvider.java

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryCtx.java
```diff
@@ -0,0 +1,36 @@
...
+    impl.scope.close();
```
anuraaga, line 34: Null check for impl.scope - if the user did incorrectly use scope and introduce a threading problem, it could be null here.

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryContextManager.java
zoercai: Yep, decided on the current approach as we're trying to migrate everyone away from OpenCensus. Having OTel context use the OpenCensus context would mean that applications and libraries would have to configure OpenCensus rather than OpenTelemetry, which is something we want to avoid. The goal is to remove this once everyone moves to OpenTelemetry.

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryCtx.java
zoercai: I think we're okay 🙂  I've just added a [test](https://github.com/open-telemetry/opentelemetry-java/pull/1863/files#diff-c1ae03736614e7e31b5afcbb427c2eb673c78085cee664a1fd61ededab33a519R44) that tests parent-child relationships for:

- OpenTelemetry span (grandparent)
    - OpenCensus span (parent)
       - OpenTelemetry span (child)

nilebox: Thanks! Could you also add a test where all 3 spans are coming from OpenCensus? This is to make sure they don't overwrite each other.

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryContextManager.java
nilebox: My understanding is that with @anuraaga 's suggestion we would still use the OpenTelemetry pipeline at a high level, though it would be storing context in OpenCensus.

I still do agree this would be more confusing than migrating everything to OpenTelemetry like it's done in this PR.

Side effects of this shim only affect code instrumented with OpenCensus, not touching parts instrumented with OpenTelemetry at all.

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryCtx.java
anuraaga: I think the use case is very small, but it's conceivable that user code relies on the return value being the previous value (they get values out of it expecting old values). So something like this could be more correct if I'm not mistaken.

```java
Context current = Context.current();
current.scope = this.context.makeCurrent();
return current;
```

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryContextManager.java
anuraaga: I guess this shim by nature assumes OC is there - so I think having the OTel context implementation also assume that should be ok? Once OC is removed from the classpath it'd just be normal OTel with no change in the app.

This approach seems fine too though, so just wanted to point it out, the main advantage would just be that the `.scope` stuff below could be a bit simpler.

opencensus_shim/src/test/java/io/opentelemetry/opencensusshim/InteroperabilityTest.java
```diff
@@ -0,0 +1,116 @@
...
+import io.opentelemetry.trace.TracingContextUtils;
+import java.util.Collection;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
+import org.mockito.MockitoAnnotations;
+import org.mockito.Spy;
+
+public class InteroperabilityTest {
```
anuraaga, line 30: Do you mind writing a similar test as this one?

https://github.com/open-telemetry/opentelemetry-java/blob/master/context/src/otelAsGrpcTest/java/io/opentelemetry/context/OtelAsGrpcTest.java

Want to make sure both sides of the propagation are working.

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryContextManager.java
anuraaga: Crosstalk :) But yeah I think either approach is fine if they can satisfy bidirectional interop like in 

https://github.com/open-telemetry/opentelemetry-java/blob/master/context/src/otelAsGrpcTest/java/io/opentelemetry/context/OtelAsGrpcTest.java

I know the OTelAsOC approach would work, but not sure about the reverse until we try :)

nilebox: > I guess this shim by nature assumes OC is there

Actually no - this shim will work fine for libraries instrumented with OpenCensus API (e.g. Google Cloud Spanner client) even if OpenCensus is not configured in the application.

i.e. the assumption is that OpenCensus "Implementation" pipeline has been replaced with OpenTelemetry but there are certain cases where OpenCensus "API" is still used, so shim helps with implicitly exporting those into OpenTelemetry as well.

anuraaga: Ah got it - thanks for the explanation

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryCtx.java
zoercai: @nilebox yep, added

@anuraaga Did you mean something like this?
```
    // Get the current context, soon to be replaced
    Context prev = Context.current();
    // Replace the scope with the scope of `this` context
    scope = context.makeCurrent();
    // Return the previous context and the current scope
    context = prev;
    return this;
```
This would make the `context` inconsistent, as it would replace `this` context with the previous context every time `attach()` is called, whereas before `attach()` is called the context would be `this` context.

anuraaga: Nah think my code is what I was thinking. The scope (used to close later) is attached to the previous context (which is what is normally returned by attach). Is there anything that would break by doing that?

zoercai: The `attach()` doesn't return a `Context`, it returns a `ContextHandle`, so to return the previous context I'd either have to do the above or create a new `ContextHandle` with the previous handle and the current scope — is this what you're suggesting?

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryTraceComponentImpl.java
zoercai: Yeah I [tried a solution](https://github.com/census-instrumentation/opencensus-java/compare/master...zoercai:migration-create-new-span-in-otel-context) where I changed OpenCensus so that the shim would turn on another path in span creation, where an OTel tracer would take over to create the span and handle the rest. We didn't end up going with this approach as we didn't want OpenCensus to depend on OpenTelemetry, especially since we're not really maintaining OpenCensus anymore and don't want to do any more releases on OpenCensus unless there is some critical bug.

zoercai: I haven't looked into opentracing much, but my understanding is that the opentracing shim is able to do so as opentracing is mainly interfaces that can be overridden, whereas with OpenCensus we'd actually have to change the implementation inside as well

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryCtx.java
anuraaga: Ah yeah I think that's what I'm suggesting. I think it's required to match the definition of that API (return the previous context) but it could be reasonable to tell interop users this isn't supported because it's a very small use case for their to actually be an effect.

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryTraceComponentImpl.java
anuraaga: Apologies if this is basic since I didn't dig into enough but from my understanding here we have `Tracer` which is an interface but we are returning `TracerImpl` from OC. Doesn't returning an implementation backed by OTel reduce the OC in the app instead of increasing it? 

IIUC the goal of the project is to allow OC users to gracefully migrate to OTel. So I guess the first step is to implement the OC API using OTel - then OC API usage can be migrated to OTel API usage gradually with no practical effect since the implementation is still OTel, the OC API dependency is finally removed and user has fully migrated. If this isn't the case, I think I'm missing a big point so let me know :)

nilebox: 1. Tracer is an [abstract class not an interface](https://github.com/census-instrumentation/opencensus-java/blob/660e8f375bb483a3eb817940b3aa8534f86da314/api/src/main/java/io/opencensus/trace/Tracer.java#L72)
2. Span is an [abstract class not an interface](https://github.com/census-instrumentation/opencensus-java/blob/660e8f375bb483a3eb817940b3aa8534f86da314/api/src/main/java/io/opencensus/trace/Span.java#L37)
3. In addition, I believe there are some static methods which may rely on certain implementation details which can't be easily swapped out with custom code using OT.

Given that OpenCensus was never designed to support custom SDK implementation (unlike OpenTracing and OpenTelemetry), it makes it pretty hard to completely replace the implementation.

So what we did instead is replaced the low-level `Context` management to use OT under the hood and also introduced a `StartEndHandler` hook which allows us to mirror all changes on OC side to OT. The rest of the code behaves as if there is a standard OC implementation.

anuraaga: Thanks for the context, makes sense especially given the Span not being an interface

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/SpanConverter.java
```diff
@@ -0,0 +1,252 @@
...
+import io.opentelemetry.api.trace.TraceState;
+import io.opentelemetry.api.trace.Tracer;
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+public class SpanConverter {
+
+  public static class FakeSpan extends Span {
```
anuraaga, line 37: Let's document what this fake span does, it's job seems to be to record events and only that.

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/SpanConverter.java
```diff
@@ -0,0 +1,252 @@
...
+    return tracestateBuilder.build();
+  }
+
+  static void mapAndAddTimedEvents(
+      io.opentelemetry.api.trace.Span span, List<TimedEvent<MessageEvent>> events) {
+    for (TimedEvent<MessageEvent> event : events) {
+      span.addEvent(
+          String.valueOf(event.getEvent().getMessageId()),
+          Attributes.of(
+              AttributeKey.stringKey("message.event.type"),
```
anuraaga, line 154: Store the attribute keys as `private static final`

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/ThreadLocalRandomHandler.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+package io.opentelemetry.opencensusshim;
+
+import io.opencensus.implcore.trace.internal.RandomHandler;
+import java.util.Random;
+import java.util.concurrent.ThreadLocalRandom;
+import javax.annotation.concurrent.ThreadSafe;
+
+/** Implementation of the {@link RandomHandler} using {@link ThreadLocalRandom}. */
```
anuraaga, line 13: Let's add a comment that the OpenCensus one is in the implementation which we don't want to use here.

opencensus_shim/src/test/java/io/opentelemetry/opencensusshim/InteroperabilityTest.java
```diff
@@ -0,0 +1,190 @@
...
+
+  private static final String NULL_SPAN_ID = "0000000000000000";
+
+  @Captor private ArgumentCaptor<Collection<SpanData>> spanDataCaptor;
+
+  @Spy private SpanExporter spanExporter;
+
+  @Before
```
anuraaga, line 43: Since this is a new test can you use junit5 / mockitoextension? e.g. https://github.com/open-telemetry/opentelemetry-java/blob/7ebbe7979d6505f89b6b848e0ede18ffc11d2562/extensions/trace_utils/src/test/java/io/opentelemetry/extensions/trace/CurrentSpanUtilsTest.java#L20

opencensus_shim/src/test/java/io/opentelemetry/opencensusshim/SpanConverterTest.java
```diff
@@ -0,0 +1,200 @@
...
+class SpanConverterTest {
+  private static final IdsGenerator RANDOM_IDS_GENERATOR = IdsGenerator.random();
+
+  @Test
+  void testToOtelSpan() {
+    RecordEventsSpanImpl span = createOpenCensusSpan();
+
+    Span otelSpan = SpanConverter.toOtelSpan(span);
+
+    assertEquals(0x1, otelSpan.getSpanContext().getTraceFlags());
```
anuraaga, line 51: Prefer assertj assertions

anuraaga (commented):
Thanks for the work!

api/src/main/java/io/opentelemetry/api/trace/TracingContextUtils.java
```diff
@@ -23,7 +23,7 @@
    * @param context the parent {@code Context}.
    * @return a new context with the given value set.
    */
-  static Context withSpan(Span span, Context context) {
+  public static Context withSpan(Span span, Context context) {
```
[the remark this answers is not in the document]
anuraaga: GitHub ordered this above the important explanatory note in https://github.com/open-telemetry/opentelemetry-java/pull/1863#discussion_r513976532

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryTraceComponentImpl.java
```diff
@@ -0,0 +1,62 @@
...
+  }
```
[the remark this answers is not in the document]
anuraaga: By the way if it sounds good, happy for it to be a followup PR, I know the rebases must be hurting (sorry ><). I think the rest of the changes are mechanical.

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/SpanConverter.java
```diff
@@ -0,0 +1,252 @@
...
+      return null;
+    }
+    SpanData spanData = ((RecordEventsSpanImpl) span).toSpanData();
+    io.opentelemetry.api.trace.Span.Builder builder =
+        TRACER
+            .spanBuilder(spanData.getName())
+            .setStartTimestamp(
+                TimeUnit.SECONDS.toNanos(spanData.getStartTimestamp().getSeconds())
+                    + spanData.getStartTimestamp().getNanos());
+    if (spanData.getKind() != null) {
```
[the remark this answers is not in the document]
zoercai: `.getKind()` has a `@Nullable` annotation and `kind` isn't required to be set when an OpenCensus span is created, so I think we do need it here?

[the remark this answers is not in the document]
anuraaga: Sorry I think I read this as OTel SpanData. Can you rename the variable to `ocSpanData`? :-)

opencensus_shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryTraceComponentImpl.java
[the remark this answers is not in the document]
zoercai: Yep I've made a ticket, will come back to this later!

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/SpanConverter.java
```diff
@@ -0,0 +1,261 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+class SpanConverter {
+
+  /**
+   * FakeSpan is used to represent OpenTelemetry spans in the OpenCensus context. Only the trace ID,
+   * span ID, trace options, and trace state are mapped, so that the correct context information can
+   * be picked up by the child spans. FakeSpan does not record events or links.
```
anuraaga, line 40: Doesn't it record events because we pass the RECORD_EVENTS option?

anuraaga (approved):
Thanks a lot for bearing with this review and the rebases!

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/SpanConverter.java
zoercai: What I'm trying to say with the comment is that the FakeSpan doesn't record events since `.addAnnotation()` is never called on it, and that it doesn't do anything even if `.addAnnotation()` was called on it, but it is the OpenCensus representation of the OpenTelemetry span which does record events. It has the record events option, which according to the `RECORD_EVENTS` documentation means "This option is set if the Span is part of a sampled distributed trace", so having this option just means it's sampled.

zoercai: Let me know if that doesn't make sense, happy to modify the comment :)

anuraaga: Ah got it - since we have a word event with a specific meaning it was confusing. "Does not record" by itself seems fine to me.

jkwatson:
I think it would be very useful to have @bogdandrutu give this a review, as our resident OC expert.

bogdandrutu:
@jkwatson had an offline discussion with @james-bebbington and @nilebox. We should merge this as the first attempt, and I will comment in an issue an aternative.

nilebox:
Thanks @bogdandrutu !
I created the issue #2012 for tracking this work.
Please add comments if anything I described there is inaccurate.

PR #1867, opened 2020-10-22
Simplify a call so that we don't have cycles in the TCU->Span call chain

jkwatson:
Just a tiny bit of cleanup.


anuraaga (approved):

carlosalberto:
Thanks!

PR #1868, opened 2020-10-22
Add a toBuilder method to the Baggage

jkwatson:


anuraaga (approved):

api/src/main/java/io/opentelemetry/baggage/ImmutableBaggage.java
```diff
@@ -67,6 +68,15 @@ public String getEntryValue(String entryKey) {
     }
   }
 
+  @Override
+  public Baggage.Builder toBuilder() {
+    Builder builder = new Builder();
+    builder.entries.putAll(entries);
+    builder.parent = parent;
+    builder.noImplicitParent = true;
```
jkwatson, line 76: this was the only bit that gave me pause. I think it ends with the simplest outcome, but does have a consequence that you're forced to manually parent if you use the toBuilder() method and you want to use Context.current() as the parent.

anuraaga: Yeah this felt natural to me so I didn't even think of it. But that's an interesting callout - seems ok to me with this behavior though.

anuraaga: Though I wonder if this is a reason to consider making Baggage mutable? Spans are mutable, so there would be symmetry with that.

jkwatson: I'm not going to push for that, personally, but I'm also not an expert, so I'm going to leave it as-is for now.

Issue #1864, opened 2020-10-22
Span.setAttribute should return `this` to enable chain calls.


Oberon00:
**Describe the solution you'd like**
I want to be able to use `mySpan.setAttribute(MY_KEY, "foo").setAttribute(OTHER_KEY, 42).end()` instead of `mySpan.setAttribute(MY_KEY, "foo"); mySpan.setAttribute(OTHER_KEY, 42); mySpan.end()`.

More often than not, you set multiple attributes at once. SpanBuilder already allows this.

dengliming:
👍 I am willing to take this.

PR #1869, opened 2020-10-23
Span return `this` to enable chain calls

dengliming:
resolves [https://github.com/open-telemetry/opentelemetry-java/issues/1864](https://github.com/open-telemetry/opentelemetry-java/issues/1864)

### DIFF
--- a/api/src/main/java/io/opentelemetry/trace/PropagatedSpan.java
+++ b/api/src/main/java/io/opentelemetry/trace/PropagatedSpan.java
@@ -101,14 +101,10 @@ final class PropagatedSpan implements Span {
   }
 
   @Override
-  public Span end() {
-    return this;
-  }
+  public void end() {}
 
   @Override
-  public Span end(EndSpanOptions endOptions) {
-    return this;
-  }
+  public void end(EndSpanOptions endOptions) {}
 
   @Override
   public SpanContext getSpanContext() {
--- a/api/src/main/java/io/opentelemetry/trace/PropagatedSpan.java
+++ b/api/src/main/java/io/opentelemetry/trace/PropagatedSpan.java
@@ -31,52 +31,84 @@ final class PropagatedSpan implements Span {
   }
 
   @Override
-  public void setAttribute(String key, String value) {}
+  public Span setAttribute(String key, String value) {
+    return this;
+  }
 
   @Override
-  public void setAttribute(String key, long value) {}
+  public Span setAttribute(String key, long value) {
+    return this;
+  }
 
   @Override
-  public void setAttribute(String key, double value) {}
+  public Span setAttribute(String key, double value) {
+    return this;
+  }
 
   @Override
-  public void setAttribute(String key, boolean value) {}
+  public Span setAttribute(String key, boolean value) {
+    return this;
+  }
 
   @Override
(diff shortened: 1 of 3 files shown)


anuraaga (approved):
Thanks a lot, great change!

thisthat (approved):

Oberon00 (approved):
That was fast!

As a follow-up after this PR we could even have both span and spanbuilder implement an interface SpanAttributeSettable which has these setAttribute overloads.

api/src/main/java/io/opentelemetry/trace/Span.java
```diff
@@ -274,16 +289,19 @@ default void setAttribute(AttributeKey<Long> key, int value) {
...
   /**
    * Marks the end of {@code Span} execution.
    *
    * <p>Only the timing of the first end call for a given {@code Span} will be recorded, and
    * implementations are free to ignore all further calls.
+   *
+   * @return this.
    */
-  void end();
+  Span end();
```
carlosalberto, line 304: Maybe I'm kind of obsessive - but I don't want to encourage users to do something with a `Span` after `end()` ;)

carlosalberto (approved):

api/src/main/java/io/opentelemetry/trace/Span.java
Oberon00: Oh, that's true! Please keep the `void` here, @dengliming

api/src/main/java/io/opentelemetry/trace/Span.java
```diff
@@ -295,8 +313,9 @@ default void setAttribute(AttributeKey<Long> key, int value) {
    * explicit values are required, use {@link #end()}.
    *
    * @param endOptions the explicit {@link EndSpanOptions} for this {@code Span}.
+   * @return this.
    */
-  void end(EndSpanOptions endOptions);
+  Span end(EndSpanOptions endOptions);
```
Oberon00, line 318: Same here (https://github.com/open-telemetry/opentelemetry-java/pull/1869/files#r510874028)

arminru (approved):
LGTM once `End` "returns" `void` again 🙂 Thanks!

api/src/main/java/io/opentelemetry/trace/PropagatedSpan.java
```diff
@@ -31,52 +31,84 @@ private PropagatedSpan(SpanContext spanContext) {
...
 
   @Override
-  public void updateName(String name) {}
+  public Span updateName(String name) {
+    return this;
+  }
 
   @Override
-  public void end() {}
+  public Span end() {
```
jkwatson, line 104: blocking the PR until the `end` methods return void again.

jkwatson (changes requested):

api/src/main/java/io/opentelemetry/trace/PropagatedSpan.java
Oberon00: both of them 😃

jkwatson: yes. didn't feel the need to beat the dead horse any more. :)

jkwatson (approved):
Thanks!

PR #1870, opened 2020-10-23
Split off PR build to separate workflow

anuraaga:
There is a small disruption to contributor experience with the single file because GitHub does not collapse the checks if any are skipped. It seems like something they should fix - but these files shouldn't have much churn anyways so even with some blatant duplication I think this is better.


thisthat (approved):

jkwatson (approved):
Thanks! Being a yaml programmer is the best!

PR #1872, opened 2020-10-23
Remove Makefile

anuraaga:
Makefile isn't idiomatic for Java, we have a perfectly good task runner in Gradle already and it will run on any OS without problem. This will also bring consistency with instrumentation and java-contrib repos - we used to use it from CircleCI but not anymore.


CONTRIBUTING.md
```diff
@@ -97,17 +97,16 @@ you can enable the docker tests by setting a gradle property of
...
+   `git submodule init`
+   `git submodule update`   
 
-    ##### Windows
+3. Run the following commands to build, run tests and most static analysis, and
+check formatting:
 
-    `gradlew.bat clean assemble check verGJF`
\ No newline at end of file
+    `./gradlew check`
```
anuraaga, line 112: I didn't separate out a Windows command since I generally don't see that in READMEs, I think users are used to switching to the bat version for these sort of commands.

iNikem (approved):

thisthat (approved):

Oberon00:
@thisthat Will you add a gradle task for #1846 ?

CONTRIBUTING.md
```diff
@@ -97,17 +97,16 @@ you can enable the docker tests by setting a gradle property of
 ``"enable.docker.tests"`` to true. See the gradle.properties file in the root of the project
 for more details.
 
-2. From the root project directory, initialize repository dependencies
+2. Clone the repository recursively
 
-   `make init-git-submodules`
+    `git clone git@github.com:<username>/opentelemetry-java.git --recursive`
```
Oberon00, line 102: I think recommending SSH clones is asking for trouble. Usually HTTPS is easier.

Oberon00: Also, either explicitly say that a fork should be used, or just use the full URL `https://github.com/open-telemetry/opentelemetry-java.git`

thisthat:
> @thisthat Will you add a gradle task for #1846 ?

I don't think so based on https://github.com/open-telemetry/opentelemetry-java/pull/1846#issuecomment-713753902

Oberon00:
I meant a manually runnable convenience gradle task that invokes the shell script (or docker directly). Since @anuraaga called gradle our task runner. So that I can run `./gradlew updateSemConv` to update but that task would never be ran automatically.

anuraaga:
Yeah I think we should add a gradle task for it and any task we need to add for our build :) It doesn't have to be wired up to the build just a defined task.

jkwatson (approved):
Thanks!

arminru (approved):

CONTRIBUTING.md
anuraaga, line 102:
```suggestion
    `git clone https://github.com/open-telemetry/opentelemetry-java.git --recursive`
```

iNikem: @anuraaga are you making code change suggestions to yourselves? :)

anuraaga: Yeah I think it's an easy way to get a oneliner in :)

PR #1876, opened 2020-10-25
Replace mock with spy in CurrentSpanUtilsTest.

anuraaga:
I'll probably replace usages of mock(Span) with spy globally later, but figure here it's an important example of using it to avoid doCallRealMethod, which is tedious for interfaces IMO (but a decision mockito has made and seems final).


jkwatson:
The mockito docs still recommend using spy sparingly. I think we should stick with just using it for interface mocks that need default methods in place, rather than replacing the usage wholesale.

jkwatson (approved):

PR #1878, opened 2020-10-25
Add more details for library authors to quickstart

MariusVolkhart:
Add additional links to the glossary where it is made clear that the instrumenting library may be the same as the instrumented library, in which case the library author is to provider their own library's name as the instrumenting library.


anuraaga (approved):
Thanks, links are always nice to have

jkwatson (approved):

PR #1880, opened 2020-10-26
Add a JUnit5 extension for setting up OpenTelemetry in a test.

anuraaga:
It's good if we can provide some test helpers for OpenTelemetry adopters. Started with a junit5 extension, would add a junit4 rule and move TestSpanData in in followups if this makes sense.

I just replaced one test for now though I'd probably remove `InMemoryTracing` from the exporter package in favor of these testing utilities instead.


iNikem (approved):

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/junit5/OpenTelemetryExtension.java
```diff
@@ -0,0 +1,110 @@
...
+  private OpenTelemetry previousGlobalOpenTelemetry;
+
+  private OpenTelemetryExtension(
+      OpenTelemetrySdk openTelemetry, InMemorySpanExporter spanExporter) {
+    this.openTelemetry = openTelemetry;
+    this.spanExporter = spanExporter;
+  }
+
+  /** Returns the {@link OpenTelemetrySdk} created by this extension. */
+  public OpenTelemetrySdk getOpenTelemetry() {
```
jkwatson, line 78: Should we return `OpenTelemetry`, rather than SDK, so people don't create tracers using the SDK directly and expect that to be the normal way to operate? We could have a separate method that returns the TracerManagement interface, perhaps?

anuraaga: It's a good point - I wasn't sure how important the distinction is in tests so made it this way. I originally had this in the top level package but moved it into the sdk package / artifact name after realizing it's only useful with the SDK. Instead of exposing API and management separately I figured this is simpler but open to alternatives.

jkwatson: My only concern is having people develop good habits wrt. using the API vs. using the SDK, so I think I'd prefer having the 2 methods exposed separately, even if it's just for this test utility. We've had enough people try to use the SDK for writing instrumentation that anything we can do to get them to not do that would be good. :)

anuraaga: Makes sense, thanks!

jkwatson (approved):
Thanks! I think this will be very useful!

PR #1881, opened 2020-10-26
Update zPages readme

MariusVolkhart:
- Add a badge to indicate the latest version. This is more reliable than manually updating the readme with releases.
- Adjust the Maven and Gradle snippets to not include a version number, since it quickly becomes out of date.
- Remove dependencies that are pulled in transitively from the Maven and Gradle snippets. They already don't include all of the transitive dependencies, so no need to only include some.
- Add a link to the blog post introducing zPages, to give more information.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -480,19 +480,25 @@ wrapper {
     gradleVersion = '6.7'
 }
 
-tasks.register("updateVersionInDocs") {
-    doLast {
-        def versionParts = version.toString().split('\\.')
-        def minorVersionNumber = Integer.parseInt(versionParts[1])
-        def nextSnapshot = "${versionParts[0]}.${minorVersionNumber + 1}.0-SNAPSHOT"
+allprojects {
+    tasks.register("updateVersionInDocs") {
+        group("documentation")
+        doLast {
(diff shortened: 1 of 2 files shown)


anuraaga:
Thanks! We have a task to automatically update version strings when releasing here

https://github.com/open-telemetry/opentelemetry-java/blob/master/build.gradle#L483

These files aren't included but do you think we can add them instead of removing entirely? I think generally docs do need to have the version for easy consumption by end users. Where we haven't maintained well we need to do better.

MariusVolkhart:
@anuraaga Agreed, using that task is preferable. I've updated the PR to:
- undo the README changes I made removing the version number
- Modify the Gradle task to apply to all projects.

When I ran the task locally, the zPages and the root READMEs got updated.

MariusVolkhart:
@anuraaga Agreed, using that task is preferable. I've updated the PR to:
- undo the README changes I made removing the version number
- Modify the Gradle task to apply to all projects.

When I ran the task locally, the zPages and the root READMEs got updated.

jkwatson (approved):
Thanks!

anuraaga (approved):
Thanks a lot!

PR #1882, opened 2020-10-26
Make GRPC an API dependency of Jaeger exporter

MariusVolkhart:
The `JaegerGrpcSpanExporter.Builder` has a public method that takes a `io.grpc.ManagedChannel`. Unless the Jaeger exporter declares an api dependency on GRPC API in Gradle, Gradle consumers will not find the GRPC API on their compile path.


anuraaga (approved):
Thanks!

bogdandrutu (approved):

jkwatson (approved):

PR #1883, opened 2020-10-26
Disable inner class imports to avoid errorprone BadImport error

bogdandrutu:
Signed-off-by: Bogdan Drutu <bogdandrutu@gmail.com>

PR #1884, opened 2020-10-26
Move ContextKey as inner class to Context. Rename to just Key.

bogdandrutu:


anuraaga (changes requested):
The only thing an inner class of Key provides is for IntelliJ to automatically rename references from Context.Key to Key. The same problem affects our builder classes but they are dependent on (Google's) maintainer-oriented priorities.

If this actually gets cargo culted in from grpc I'm never touching this repo again.

Oberon00:
@bogdandrutu Why this (breaking) change?

bogdandrutu:
> The only thing an inner class of Key provides is for IntelliJ to automatically rename references from Context.Key to Key.

IntelliJ does what you configure it to do, and that was a wrong config in `.editorconfig`, see https://github.com/open-telemetry/opentelemetry-java/pull/1883 for the fix.

> The same problem affects our builder classes but they are dependent on (Google's) maintainer-oriented priorities.

I am not sure where is this coming from, and where the "Google" part comes from, see explanation why an inner class describes better the relationship between classes than a "name" hierarchy.

> If this actually gets cargo culted in from grpc I'm never touching this repo again.

Not going to answer to offensive things and radical decisions, I would be happy to discuss reasons in technical way.

> Why this (breaking) change?

@Oberon00 this may be a personal preference, but I think using the name of a class to describe relationship between classes such as `ContextKey`, which for me means this is a `Key` in the `Context` it is better described by an inner class which clearly describes the relationship.

jkwatson:
My general approach to deciding on the "inner class vs. not" debate is to ask the question: is this class/interface useful outside of the context (no pun intended) of how it's being used. In this case, the ContextKey is only useful for setting things into the Context, so it does seem like a good usage of an inner interface. I think it also makes it a bit more consistent with the rest of our APIs.

@anuraaga Is there a technical reason why you oppose this change? I don't feel super strongly one way or another, myself.

trask:
I like modeling using inner classes also (e.g. `X.Key` and `Y.Builder`), but @anuraaga [has brought up the point previously](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1387#discussion_r505091054) about Intellij dropping the `X.` and `Y.`, which is annoying.

I'm definitely interested if #1883 resolves this issue, at least for us, though I think @anuraaga's point is more about users of our library, and they would unlikely be aware of this setting.

bogdandrutu:
@trask I did the renaming using #1883 and did not have to do the re-adding the "x.". The reason I investigated the problem was because when I first did the change in intellij it changed everywhere the "x.", so errorprone was complaining, because of that I found the problem and fixed so in this PR I did not have to manually add back "x." :)

anuraaga:
> I am not sure where is this coming from, and where the "Google" part comes from

Came from AutoValue forcing this :)

> I'm definitely interested if #1883 resolves this issue, at least for us, though I think @anuraaga's point is more about users of our library, and they would unlikely be aware of this setting.

Yeah, I don't think .editorconfig here would help with the end users. I think an inner class with a generic name like Key has too much risk of damaging user code.

bogdandrutu:
> Came from AutoValue forcing this :)

I think you have to blame Joshua Bloch for this, as mentioned in my first comment

anuraaga:
@trustin gets the credit for my comments actually :P Encourage giving these a read they're quite nice

https://assets.ctfassets.net/oxjq45e8ilak/22C0GxEW5qThqGE4q8kGXT/921f99129699591342880424d6f1541e/Trustin_Lee_Writing_a_Java_library_with_better_experience_2020_06_28_10_49_34.pdf

bogdandrutu:
It is funny because I tried to apply almost all of them to this project:
* Using final everywhere is possible;
* Avoid exposing API that don't need to be public (people herd me always asking if that can be package-protected or private);
* All our API and SDK had pre-conditions which were removed in the past 3 months because a telemetry library should "not kill" the application (which personally I disagree, any API that is misused should fail to avoid unexpected behaviors).
* Consistency - I always ask to do things consistent with the other APIs that we provide;
* We use "@Nullable";

The only part which I disagree from the presentation is the inner class suggestion, because I feel that they express much better the scope of a class (like a Builder, e.g. `Foo.Builder` is a `Builder` for class `Foo`), I can see how users that don't use one of the style guides that forbids the import of inner classes may get confusing code. If the user does not setup any style guide, do they care about this?

anuraaga:
> If the user does not setup any style guide, do they care about this?

Instrumentation repo cares about style but IntelliJ refactoring almost got us a `Builder` import in. It's indeed cool if we can add an editorconfig to avoid this problem, but I think this is a pretty advanced setting that most users wouldn't know about. My general rule of thumb is to not use IntelliJ refactoring when it involves an inner class since it'll do a bad thing commonly. If our code avoided this intrinsicly, that means no chance of that issue.

I don't see the presence of a dot really changing the scope semantics of `Foo.Builder` and `FooBuilder`. And we have `AttributeKey` which looks very similar to `ContextKey` so seems consistent?

jkwatson:
I think we're suffering a bit from the fact that we don't have a common set of design guidelines for our APIs. There's very little in API design that's categorically "right" or "wrong". It all depends on what the audience is used to using and consuming. All of us have been building and designing APIs for a long time and I don't think anyone is actively trying to design harmful APIs. 

If you come from a googley ecosystem, with guava, grpc, etc, you have one set of expectations for the shape of your API. If you come from a Spring ecosystem, you're used to another one. And so on and so on.

So, what criteria can we use to make a decision, that isn't just biased by what we're most familiar with?

anuraaga:
BDFL decides the design guidelines and hopefully documents it somewhere so we just point at it? Is that you @jkwatson? :)

trask:
It's super helpful to have well thought out (existing) documents that we can align around.

I would not mind at all aligning to @trustin's document from the link above, and documenting that decision in both repos.

jkwatson:
If I'm BDFL don't I get to assign that work to my minions?

carlosalberto (approved):

carlosalberto:
I'm ok with this change, maybe because it makes it similar to the original `io.grpc` API ;)

> If this actually gets cargo culted in from grpc I'm never touching this repo again.

We have all compromised at some point in the past, e.g. I was deeply against 'forking' `io.grpc.Context` for many reasons. But finding a compromise is part of the OSS workflow, even if it may become annoying at times ;)

> Instrumentation repo cares about style but IntelliJ refactoring

Don't forget some of us don't use IDEs (yeah, I know, but I'm still *too fond* of Vim ;) )

bogdandrutu:
> Instrumentation repo cares about style but IntelliJ refactoring almost got us a Builder import in.

That was funny, see https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1489

trask:
> So, what criteria can we use to make a decision, that isn't just biased by what we're most familiar with?

This would be a good topic for tomorrow's [Java Special Topics meeting](https://calendar.google.com/event?action=TEMPLATE&tmeid=Nm1vZXBtNmxxMm44dW9sOWRmZWwwdm5icnRfMjAyMDEwMjhUMDUwMDAwWiBnb29nbGUuY29tX2I3OWUzZTkwajdiYnNhMm4ycDVhbjVsZjYwQGc&tmsrc=google.com_b79e3e90j7bbsa2n2p5an5lf60%40group.calendar.google.com)

bogdandrutu:
@trask just helped me find a good example in JDK `Map.Entry` :)

trask:
But is that a reason? `Map.Entry` imports annoy me constantly 😂

PR #1886, opened 2020-10-27
Set plugin versions in only one place

bogdandrutu:
Also update spotless plugin.


anuraaga (approved):

anuraaga:
Thanks, was always a bit weird

PR #1887, opened 2020-10-27
Remove support to override create key

bogdandrutu:
Indeed this removes one of the testing case otelAsGrpc, but we should not expose an API just for this purpose until we know for sure we need it (applying the rule of minimal API).

Also the createKey it is a strange method on the Storage (which should encapsulate only methods related to storage, and the creation of the key is a Context property).
In the future if needed this should be a method on the Context or other ContextKeyFactory, not on the Storage.


anuraaga (commented):
The `otelAsGrpc` case is the true interop case and most important. It's definitely going to be important as I start to add context interop in the instrumentation (took a week to update the SDK dependency in instrumentation due to several Java 8 bytecode issues but finally ready!).

The naming is definitely an issue, how about renaming `ContextStorage` to something more generic, maybe `ContextManager`?

bogdandrutu:
> The otelAsGrpc case is the true interop case and most important. It's definitely going to be important as I start to add context interop in the instrumentation (took a week to update the SDK dependency in instrumentation due to several Java 8 bytecode issues but finally ready!).

Not sure, I think the `otelInGrpc` is more performant (less wrappers), and it is still interop, context will get propagated based on the grpc propagation.

> The naming is definitely an issue, how about renaming ContextStorage to something more generic, maybe ContextManager, ContextProvider?

I think we need 2 classes `ContextStorage` for the storage/propagation part (which uses a Context object), and another class `ContextKeyManager` for the Key part. Then you can put them into one `ContextProvider` but I would not mix these 2 different functionalities.

anuraaga:
> context will get propagated based on the grpc propagation.

But I think if instrumentation propagates the context, which it often does, changes will get out of sync

https://github.com/open-telemetry/opentelemetry-java/blob/master/context/src/otelInGrpcTest/java/io/opentelemetry/context/OtelInGrpcTest.java#L114

We had a user report about it

https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1313#issuecomment-703951027

So I'd like to continue to try with the key customization and removing the method would block that.

>  Then you can put them into one ContextProvider but I would not mix these 2 different functionalities.

Yeah as long as it's one SPI, that sounds nice.

bogdandrutu:
I can send an issue for that bug without the key. Let me send it first to see that it is possible without key

bogdandrutu:
@anuraaga see https://github.com/open-telemetry/opentelemetry-java/pull/1889

anuraaga (approved):
Seems this isn't needed, thanks!

PR #1888, opened 2020-10-27
Add test for otel in brave

bogdandrutu:

### DIFF
--- a/context/src/otelInBraveTest/java/io/opentelemetry/context/BraveContextStorageProvider.java
+++ b/context/src/otelInBraveTest/java/io/opentelemetry/context/BraveContextStorageProvider.java
@@ -26,16 +26,17 @@ public class BraveContextStorageProvider implements ContextStorageProvider {
     public Scope attach(Context toAttach) {
       CurrentTraceContext currentTraceContext = Tracing.current().currentTraceContext();
       TraceContext currentBraveContext = currentTraceContext.get();
-      Context currentContext = ContextWrapper.fromBraveContext(currentBraveContext);
+
+      Context currentContext = fromBraveContext(currentBraveContext);
       if (currentContext == toAttach) {
         return Scope.noop();
       }
 
       TraceContext newBraveContext;
-      if (toAttach instanceof ContextWrapper) {
-        newBraveContext = ((ContextWrapper) toAttach).toBraveContext();
+      if (toAttach instanceof BraveContextWrapper) {
+        newBraveContext = ((BraveContextWrapper) toAttach).toBraveContext();
       } else {
-        newBraveContext = currentBraveContext.toBuilder().addExtra(toAttach).build();
+        newBraveContext = toBraveContext(currentBraveContext, toAttach);
       }
 
       if (currentBraveContext == newBraveContext) {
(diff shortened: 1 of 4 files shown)


context/src/otelInBraveTest/java/io/opentelemetry/context/BraveContextStorageProvider.java
```diff
@@ -0,0 +1,86 @@
...
+      for (int i = extra.size() - 1; i >= 0; i--) {
+        Object nextExtra = extra.get(i);
+        if (nextExtra.getClass() == ContextWrapper.class) return (ContextWrapper) nextExtra;
+      }
+      return null;
+    }
+  }
+
+  // Need to wrap the Context because brave findExtra searches for perfect match of the class.
+  static final class ContextWrapper implements Context {
```
bogdandrutu, line 67: Instead of this, I may be able to check for `DefaultContext` but that may not work for everyone, so to show how to do it I wrapped it.

bogdandrutu:
This needs the same trick as #1889 to propagate the base brave context as well

context/src/otelInBraveTest/java/io/opentelemetry/context/OtelInBraveTest.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicReference;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class OtelInBraveTest {
+
+  private static final ContextKey<String> ANIMAL = ContextKey.named("animal");
+  private static final Context CONTEXT_WITH_ANIMAL = Context.root().with(ANIMAL, "japan");
```
jkwatson, line 23: an unusual animal. :)

bogdandrutu: @anuraaga is guilty for this :)) and me that I copied paste

jkwatson:
This looks ok to me. I'll let @anuraaga give it a once over, though, as he knows Brave and our context much much better than I do.

context/src/main/java/io/opentelemetry/context/DefaultContext.java
```diff
@@ -61,22 +61,22 @@ private DefaultContext(PersistentHashArrayMappedTrie.Node<ContextKey<?>, Object>
   }
 
   @Override
-  public <V> Context with(ContextKey<V> k1, V v1) {
+  public <V> DefaultContext with(ContextKey<V> k1, V v1) {
```
anuraaga, line 64: Don't think it hurts but is there something we get with covariant types here?

context/src/otelInBraveTest/java/io/opentelemetry/context/BraveContextStorageProvider.java
```diff
@@ -0,0 +1,108 @@
...
+    @Override
+    public <V> Context with(ContextKey<V> k1, V v1) {
+      return new ContextWrapper(baseBraveContext, context.with(k1, v1));
+    }
+
+    private static DefaultContext fromBraveContext(TraceContext braveContext) {
+      List<Object> extra = braveContext.extra();
+      for (int i = extra.size() - 1; i >= 0; i--) {
+        Object nextExtra = extra.get(i);
+        if (nextExtra.getClass() == DefaultContext.class) {
```
anuraaga, line 101: Does it work to check class != ContextWrapper? This test is lucky to have package private access but a real provider wouldn't.

anuraaga (approved):

context/src/main/java/io/opentelemetry/context/DefaultContext.java
anuraaga: Ah realized it does hurt a bit since bytecode instrumentation doesn't work anymore (root() being the important one)

context/src/otelInBraveTest/java/io/opentelemetry/context/BraveContextStorageProvider.java
bogdandrutu: We cannot add `ContextWrapper` because it will create a circular reference. So if that is the case I need another ContextWrapper that wraps only the otel Context.

jkwatson:
Needs a rebase, and it looks like one outstanding comment from @anuraaga on this. Ping @bogdandrutu

context/src/main/java/io/opentelemetry/context/DefaultContext.java
jkwatson: You approved this PR. Is this ok, @anuraaga ?

bogdandrutu: Can you explain what is the problem?

anuraaga: Yeah - root() is intercepted by auto instrumentation to return an agent managed context. We need the return value to be the public type.

anuraaga (changes requested):
Sorry approved before realizing the effect on auto instrumentation

carlosalberto (approved):

bogdandrutu:
@anuraaga @trask we are getting back to "auto-instrumentation" requires this, but not documented. Please make sure that this requirement is documented, not clear why this cannot happen, seems like a correct change.

trask:
yup, agree that documenting these weird things is important: #1977

context/src/main/java/io/opentelemetry/context/DefaultContext.java
bogdandrutu: No more changes.

bogdandrutu:
@anuraaga no more changes to DefaultContext.

anuraaga (approved):

PR #1889, opened 2020-10-27
Fix propagating Otel context to propagate gRPC context

bogdandrutu:

### DIFF
--- a/context/src/otelInGrpcTest/java/io/opentelemetry/context/GrpcContextStorageProvider.java
+++ b/context/src/otelInGrpcTest/java/io/opentelemetry/context/GrpcContextStorageProvider.java
@@ -6,6 +6,8 @@
 package io.opentelemetry.context;
 
 public class GrpcContextStorageProvider implements ContextStorageProvider {
+  private static final io.grpc.Context.Key<Context> OTEL_CONTEXT =
+      io.grpc.Context.keyWithDefault("otel-context", Context.root());
 
   @Override
   public ContextStorage get() {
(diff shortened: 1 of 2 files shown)


context/src/otelInGrpcTest/java/io/opentelemetry/context/GrpcContextStorageProvider.java
```diff
@@ -27,15 +26,58 @@ public Scope attach(Context toAttach) {
...
 
+      io.grpc.Context toRestore = newGrpcContext.attach();
       return () -> newGrpcContext.detach(toRestore);
     }
 
     @Override
     public Context current() {
-      return OTEL_CONTEXT.get();
+      // We return an object that embeds both the
+      io.grpc.Context grpcContext = io.grpc.Context.current();
```
anuraaga, line 44: From what I understand, we essentially resync gRPC with OTel context any time `current` is called. And since we naturally can't have implicit context propagation without calling `current` anyways this covers all use cases. Does that sound right?

anuraaga (approved):
Nice, thanks! If it's ok, let me try this pattern out in the opentelemetry agent bridge to check it there too but seems like it should be fine.

context/src/otelInGrpcTest/java/io/opentelemetry/context/GrpcContextStorageProvider.java
bogdandrutu: Yes. This is similar to otelAsGRPC but instead of getting values from gRPC context using the wrapped keys, you first get our context from there, then use the key to get the value from our context.

We can simplify the code to always call into the gRPC via the extra call I mentioned but that is a bit more inefficient.

anuraaga:
Confirmed this works well in the agent too https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1493 (though not quite what I was aiming for since that's not a pure interop story 😅 )

PR #1891, opened 2020-10-27
Add JMH benchmark for context.

anuraaga:
I'm interested in how much performance we get from the trie implementation for context. For now, this just adds some benchmarks, and open to any suggestions.

Results with current implementation https://gist.github.com/anuraaga/b67d32c6a08ad9d2a38d39fd357efb00

Results with a CoW array implementation https://gist.github.com/anuraaga/2f2187cf1886197b5d65c2018001750d

It's pretty interesting. I think for the array, the worst case for read is `readLast`. Even for that, performance of the array implementation seems better up to 10 items in context. 20 / 40 there is a sharp drop as we'd expect. Because we have the caveat that "Context should not be treated like a general purpose map with a large number of keys and values" perhaps this is an acceptable level? For our expected usage, I think we generally have 5 in our code - span, local root, client root, baggage, miscellaneous. Arguably local root, client root, miscellaneous should just be grouped into one struct anyways.

For write, the first and second write are a lot faster for the trie, but third and up become a lot faster for the array. For a use case of 5 elements, it seems like a wash. For 10, I think array is faster. Writes should be less common than reads though so it might not matter either way.

Let me know if this benchmark and numbers make sense and if they seem motivating enough for a shift to a simple array implementation.


context/src/jmh/java/io/opentelemetry/context/ContextBenchmark.java
```diff
@@ -0,0 +1,83 @@
...
+      keys.add(key);
+    }
+  }
+
+  /** Read value for first key. */
+  @Benchmark
+  public String readFirst() {
+    return context.get(keys.get(0));
+  }
+
```
thisthat, line 53: Would it a get to the middle element make sense?

bogdandrutu:
Personally if what you say is true (that we have ~5) I would special case this (similar to what Reactor does) then fallback to trie or Array.

Also the current trie is not optimal, I would do 2 more changes before making a decision:
1. Making Node in the trie implement the Context. Avoid one extra wrapping.
2. Which may be good for both cases if the value that we replace is same reference return same object.
2.

context/src/jmh/java/io/opentelemetry/context/ContextBenchmark.java
bogdandrutu: For trie getting any element should have almost the same impact.

bogdandrutu: So yes a test for the middle should be acceptable

iNikem (approved):

jkwatson (approved):
Thanks! Would love a screenshot/paste of the output from a run, just for some flavor of what the numbers are looking like.

anuraaga:
@jkwatson Do you mean something else besides the gist links I posted? Not sure what else output there is :)

context/src/jmh/java/io/opentelemetry/context/ContextBenchmark.java
anuraaga: Seems interesting to have, will try to run it later. Thanks for the suggestion!

anuraaga:
@bogdandrutu Thanks for the optimization suggestions, will try them out

jkwatson:
> @jkwatson Do you mean something else besides the gist links I posted? Not sure what else output there is :)

No, it would help if I read every word in the description and noticed that there were links in there.  🤕

PR #1892, opened 2020-10-27
Move API packages to .api. package

### LINKED ISSUE #1830: Move opentelemetry-api to io.opentelemetry.api package
Right now the OTel API is housed in `io.opentelemetry` package. We could move it to `io.opentelemetry.api`. This

- Makes it a bit more explicit in the package name that this is the OpenTelemetry API, as opposed to some OpenTelemetry helper
- Makes auto instrumentation much less fragile by having a single Java package prefix to handle specially with regards to shading. Currently, all classes in `io.opentelemetry` have to be handled specially since the `io.opentelemetry` prefix contains the world, not just the OpenTelemetry API
- Seems to have no real downside (other than a ton of work for me when updating the agent since our tests use Groovy 😿 )

anuraaga:
Fixes #1830


anuraaga:
I realized our `moduleName` was already set to this anyways.

thisthat (approved):
Thank you :)

api/src/test/java/io/opentelemetry/api/OpenTelemetryTest.java
```diff
@@ -59,11 +60,13 @@ void after() {
 
   @Test
   void testDefault() {
-    assertThat(OpenTelemetry.getGlobalTracerProvider()).isInstanceOf(DefaultTracerProvider.class);
-    assertThat(OpenTelemetry.getGlobalTracerProvider())
+    Assertions.assertThat(OpenTelemetry.getGlobalTracerProvider())
```
jkwatson, line 63: odd that this is not being statically imported any more. I won't hold up the PR for it, since it's minor, but odd that it snuck in here.

jkwatson (approved):
Thanks! 
One thing I'm not 100% sure about is whether the SPI interfaces should be in the API package, or whether they should be in a separate `spi` package which stands alone.

bogdandrutu:
@jkwatson consider to file an issue for the spi.

bogdandrutu (approved):

api/src/test/java/io/opentelemetry/api/OpenTelemetryTest.java
bogdandrutu: haha :) this is because I disabled imports of inner classes/methods. Do you want to allow imports for the assertions I expect.

carlosalberto (approved):

api/src/test/java/io/opentelemetry/api/OpenTelemetryTest.java
bogdandrutu: Did the refactoring myself, this change did not happen. so it needs to be reverted

jkwatson:
> @jkwatson consider to file an issue for the spi.

https://github.com/open-telemetry/opentelemetry-java/issues/1897

api/src/test/java/io/opentelemetry/api/OpenTelemetryTest.java
anuraaga: Thanks for finding! I'm used to IntelliJ refactoring causing havoc with static imports, just one file being hit was relatively tame

jkwatson:
needs a rebase, looks like. We need to get this thing merged ASAP, otherwise someone is going to be rebasing over and over.

Issue #1885, opened 2020-10-26
Traceparent header constant


mariusioana:
Hi,

A very simple question :)

Is there any way to get from the open telemetry java sdk the "traceparent" header name as a constant? Traceparent header is not among the common headers list inside a framework that we use and we need to whitelist it so that it goes further to downstream services. Of course, we can whitelist it by just using the "traceparent" string but wondered if there is any way to abstract it and get it from the SDK. I see HttpTraceContext having it as TRACE_PARENT but with package access level. We do not want to extract and do extra processing for now, we just want to let it pass through.

Thanks!

jkwatson:
Hi @mariusioana . the "traceparent" header is part of the W3C trace context specification, so I would guess that it would be a good thing to put onto the allow-list generally. I'd also be fine if we wanted to make that constant public in the `HttpTraceContext` class, if that would be helpful.

mariusioana:
Thanks @jkwatson ! Making it public in HttpTraceContext would be a small thing but would be helpful at least for us.

Oberon00:
Don't forget the `tracestate` header that should always be kept together with `traceparent`.

dengliming:
@mariusioana Seems `fields() `  is good for you. Can we close this?

mariusioana:
Sure thing! Thanks!

PR #1894, opened 2020-10-27
Make TRACE_PARENT、TRACE_STATE public in the HttpTraceContext

dengliming:
closes [https://github.com/open-telemetry/opentelemetry-java/issues/1885](https://github.com/open-telemetry/opentelemetry-java/issues/1885)

### DIFF
--- a/api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java
+++ b/api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java
@@ -36,8 +36,8 @@ public class HttpTraceContext implements TextMapPropagator {
   private static final Logger logger = Logger.getLogger(HttpTraceContext.class.getName());
 
   private static final TraceState TRACE_STATE_DEFAULT = TraceState.builder().build();
-  static final String TRACE_PARENT = "traceparent";
-  static final String TRACE_STATE = "tracestate";
+  public static final String TRACE_PARENT = "traceparent";
+  public static final String TRACE_STATE = "tracestate";
   private static final List<String> FIELDS =
       Collections.unmodifiableList(Arrays.asList(TRACE_PARENT, TRACE_STATE));
 


api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java
```diff
@@ -36,8 +36,8 @@
   private static final Logger logger = Logger.getLogger(HttpTraceContext.class.getName());
 
   private static final TraceState TRACE_STATE_DEFAULT = TraceState.builder().build();
-  static final String TRACE_PARENT = "traceparent";
-  static final String TRACE_STATE = "tracestate";
+  public static final String TRACE_PARENT = "traceparent";
+  public static final String TRACE_STATE = "tracestate";
```
jkwatson, line 40: if we make these public, I think we should rename them to `TRACE_PARENT_HEADER_NAME` and `TRACE_STATE_HEADER_NAME` so it's more clear what they represent.

Oberon00:
For the use case mentioned in the issue, would using fields() be possible.

api/src/main/java/io/opentelemetry/trace/propagation/HttpTraceContext.java
anuraaga: @mariusioana Just wondering, does the `fields` method work well for your use case without making the header names themselves public? They are all the fields for any given propagator, which may solve your case more generically?

I'm ok with making these public, with the renaming, since they're very stable constants but just wanted to check.

mariusioana: @anuraaga @Oberon00 I think something like `HttpTraceContext.getInstance().fields().get(0)` does it, it's more cryptic,  but something that we can use. I would still prefer to reference like suggested by @jkwatson using constants like `TRACE_PARENT_HEADER_NAME`.

Oberon00: @mariusioana why `.get(0)`? If you want to add them to the allowed-headers-list of a component, you need to add all the fields in the returned value.

mariusioana: @Oberon00 true that! I was still thinking only about `Traceparent`. Btw, `Tracestate` is optional right? Reading from https://www.w3.org/TR/trace-context/#tracestate-header

Oberon00: Depending on what you mean by optional. Many tracing systems will work fine without tracestate, but some might require it. Also, according to W3C, "Vendors receiving a tracestate request header MUST send it to outgoing requests.", so you may not drop it. See https://www.w3.org/TR/trace-context/#mutating-the-tracestate-field

anuraaga: @mariusioana If `fields()` sounds good for you, how about we go ahead and close this PR for now? If we find another use case for these constants we can revisit, but it's easier to make private public than other way around.

mariusioana: @anuraaga Sure, let's close this PR. Thanks!

dengliming: Closing it. glad to help. : )

PR #1896, opened 2020-10-27
Use LongAdder and DoubleAdder instead of atomic variants

tylerbenson:
Now that we're based on Java 8, we can use these instead.


bogdandrutu:
Please do the same for Double

jkwatson (approved):
Thanks! As @bogdandrutu mentioned, the double version would also be nice (could be a separate PR)

anuraaga (approved):
Sweet

jkwatson:
@tylerbenson needs an unused-import scrub, looks like.

PR #1899, opened 2020-10-27
Make the ParentBased sampler (and builder) public so SDK users can create their own combinations

### LINKED ISSUE #1898: Samplers.ParentBased.Builder should be made public, so SDK users can create their own ParentBased samplers
**Describe the bug**
Currently both the `Samplers.ParentBased` class and its internal `Samplers.ParentBased.Builder` class are package-private. They are used internally by the `Samplers` class to, for example, create a `parentBased(sampler)` that explicitly sets the _remoteParentSampled_, _remoteParentNotSampled_, _localParentSampled_, and _localParentNotSampled_ using the Builder's public methods, but it's impossible for an SDK end-user to do the same.

**Steps to reproduce**

```java
Sampler s = new Samplers.ParentBased.Builder().setRemoteParentNotSampled(Samplers.alwaysOn()).build()
```

should compile for a user importing `opentelemetry-sdk`, but does not.

Expected to be able to create my own `ParentBased` sampler with my own choices for the 4 different sub-samplers (remote, local, sampled, not sampled).

**What did you see instead?**

Unable to compile because `Samplers.ParentBased` is not public.

Artifacts: `opentelemetry-sdk`
Version: `v0.8.0`, `v0.9.0`

**Environment**
Compiler: AdoptOpenJDK 8
OS: MacOS

**Additional context**

I don't think hiding these was intentional, but I could understand if perhaps it was in order to limit the initial scope of the GA release.

jkwatson:
resolves #1898

### DIFF
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
@@ -217,8 +217,13 @@ public final class Samplers {
     }
   }
 
+  /**
+   * A Sampler that uses the sampled flag of the parent Span, if present. If the span has no parent,
+   * this Sampler will use the "root" sampler that it is built with. See documentation on the {@link
+   * Builder} methods for the details on the various configurable options.
+   */
   @Immutable
-  static class ParentBased implements Sampler {
+  public static class ParentBased implements Sampler {
 
     private final Sampler root;
     private final Sampler remoteParentSampled;
@@ -288,7 +293,8 @@ public final class Samplers {
       return getDescription();
     }
 
-    static class Builder {
+    /** A builder for creating ParentBased sampler instances. */
+    public static class Builder {
 
       private final Sampler root;
       private Sampler remoteParentSampled;
@@ -345,7 +351,7 @@ public final class Samplers {
        *
        * @return the ParentBased sampler.
        */
-      public ParentBased build() {
+      public Sampler build() {
         return new ParentBased(
             this.root,
             this.remoteParentSampled,


evantorrie (approved):
LGTM

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
```diff
@@ -217,8 +217,13 @@ public String toString() {
...
   }
 
+  /**
+   * A Sampler that uses the sampled flag of the parent Span, if present. If the span has no parent,
+   * this Sampler will use the "root" sampler that it is built with. See documentation on the {@link
+   * Builder} methods for the details on the various configurable options.
+   */
   @Immutable
-  static class ParentBased implements Sampler {
+  public static class ParentBased implements Sampler {
```
anuraaga, line 226: Do we need to make the class public?

jkwatson: we do, if you want to use the builder, otherwise you can't import the inner class builder

anuraaga: Ah got it

anuraaga (approved):

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/Samplers.java
jkwatson: (as an aside, Samplers.ParentBased.Builder : 🤮 )

PR #1900, opened 2020-10-28
Remove unused package-info from context_prop io.otel.context

MariusVolkhart:
The presence of package-info.java in the io.opentelemetry.context package of the io.opentelemetry.context.propagation JPMS module results in a split package, due to the io.opentelemetry.context module declaring the same package. This causes big problems for JPMS consumers.

Since the package-info.java is unused, removing it should not cause any problems.


anuraaga:
I thought JPMS is concerned with package boundaries, not "package prefix" - aren't `.context` and `.context.propagation` different packages? Just want to ask since I am not that familiar with it TBH.

This package-info is mostly for the `@ParametersNonNullByDefault` annotation which I think is important. I think context_prop artifact can / should be merged (in-process and remote context propagation are basically same concept) into context and have been thinking of doing it but will definitely do it sooner if the split is causing some breakage. @jkwatson what do you think?

anuraaga:
Oh doh! Now I get it :)

anuraaga (approved):
Thanks!

jkwatson (approved):
good catch!

PR #1901, opened 2020-10-28
Expose factory for random IDs generator instead of type.

anuraaga:
I didn't keep the `ThreadLocalRandom` piece from the previous javadoc but am a bit on the fence. While it's an implementation detail that we may even change later, so usually wouldn't put it in a javadoc, it's an important implementation detail that users may need to know.


anuraaga:
Well added a commit that tries to just say exactly that maybe it's ok

jkwatson:
I'd also like to rename `IdsGenerator` to `IdGenerator`, but that's a separate bit of work.

jkwatson:
Also, the ThreadLocalRandom bit is important to users to know, since it's totally broken on older android versions.

jkwatson (approved):
Thanks!

PR #1902, opened 2020-10-28
Revert package rename in examples.

anuraaga:
Not sure why IntelliJ did this for a refactoring... should be fixed


jkwatson (approved):
Thanks!

PR #1903, opened 2020-10-28
Move samplers into samplers package instead of outer class.

anuraaga:
I noticed `Samplers` is a huge bag of classes - discussed with @jkwatson and it seems easier to organize as a package. Though making `Samplers` package private, moving up `SamplingResult` and just moving the factories to `Sampler` / `SamplingResult` would be a good enough win too, I'm happy to change to that if it seems better. Even in that case, I would probably still leave the tests split to keep them better contained.


sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/SpanBuilderSdk.java
```diff
@@ -258,4 +258,15 @@ private static Clock getClock(Span parent, Clock clock) {
       return MonotonicClock.create(clock);
     }
   }
+
+  @VisibleForTesting
+  static boolean isRecording(SamplingResult.Decision decision) {
```
anuraaga, line 263: These were only used in this class so moved them here

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/samplers/ImmutableSamplingResult.java
```diff
@@ -0,0 +1,57 @@
...
+import static io.opentelemetry.api.common.AttributeKey.doubleKey;
+
+import com.google.auto.value.AutoValue;
+import io.opentelemetry.api.common.AttributeKey;
+import io.opentelemetry.api.common.Attributes;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+@AutoValue
+abstract class ImmutableSamplingResult implements SamplingResult {
```
anuraaga, line 17: I think we have a pattern of naming an AutoValue class `ImmutableBlah` so used that but not sure how cosistent are on it currently. It's a private class anyways

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/samplers/ParentBasedSampler.java
```diff
@@ -0,0 +1,202 @@
...
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * A Sampler that uses the sampled flag of the parent Span, if present. If the span has no parent,
+ * this Sampler will use the "root" sampler that it is built with. See documentation on the {@link
+ * Builder} methods for the details on the various configurable options.
+ */
+@Immutable
+public class ParentBasedSampler implements Sampler {
```
anuraaga, line 24: This is top level now instead of `class Samplers.ParentBased`.

Actually if `ParentBasedSamplerBuilder` is made top-level, this class could be made private.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/samplers/SamplingResult.java
```diff
@@ -0,0 +1,104 @@
...
+import io.opentelemetry.api.trace.SpanContext;
+import io.opentelemetry.api.trace.TraceState;
+import io.opentelemetry.context.Context;
+import java.util.List;
+
+/**
+ * Sampling result returned by {@link Sampler#shouldSample(Context, String, String, Span.Kind,
+ * ReadableAttributes, List)}.
+ */
+public interface SamplingResult {
```
anuraaga, line 22: This is top level now, `Samplers.SamplingResult` had some stuttering of "sampl"

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/samplers/SamplingResult.java
```diff
@@ -0,0 +1,104 @@
...
+   * returning {@code decision}.
+   *
+   * <p>This is meant for use by custom {@link Sampler} implementations.
+   *
+   * <p>Use {@link #create(Decision, Attributes)} if you need attributes.
+   *
+   * @param decision The decision made on the span.
+   * @return A {@link SamplingResult} with empty attributes and the provided {@code decision}.
+   */
+  static SamplingResult create(Decision decision) {
```
anuraaga, line 35: Factories for `SamplingResult` moved to the `SamplingResult`

anuraaga: Also I think using a method overload (same name) seems consistent with our normal style instead of the previous name `empty`

jkwatson (approved):
I think these are good changes. Thanks!

carlosalberto (approved):

PR #1904, opened 2020-10-28
Hide Default* implementations behind interface factory methods.

anuraaga:


api/src/main/java/io/opentelemetry/api/metrics/Meter.java
```diff
@@ -25,6 +25,14 @@
 @ThreadSafe
 public interface Meter {
 
+  /**
+   * Returns a {@link Meter} that only creates no-op {@link Instrument}s that neither record nor are
+   * emitted.
+   */
+  static Meter noop() {
+    return DefaultMeter.getInstance();
```
bogdandrutu, line 33: The name of the method is misleading, for metrics it is right now a noop but we may add small functionality. Not sure about a better name but wanted to raise this issue

api/src/main/java/io/opentelemetry/api/metrics/DefaultMeter.java
```diff
@@ -14,23 +14,18 @@
 
 /** No-op implementations of {@link Meter}. */
 @ThreadSafe
-public final class DefaultMeter implements Meter {
+final class DefaultMeter implements Meter {
 
-  private static final DefaultMeter INSTANCE = new DefaultMeter();
+  private static final DefaultMeter instance = new DefaultMeter();
```
jkwatson, line 19: why lowercase? Isn't the DefaultMeter immutable?

jkwatson:
I definitely like making the `Default*` classes hidden. Do we need to expose them on the API at all? Can they just be an implementation detail? This is a case where we really need "friend" classes, or "nestmate" classes that aren't exposed on the public API at all.

kenfinnigan:
I haven't looked at the SDK code to see the implications, but instead of hiding behind a factory could the "Default" classes just be `package` and not `public` classes?

jkwatson:
> I haven't looked at the SDK code to see the implications, but instead of hiding behind a factory could the "Default" classes just be `package` and not `public` classes?

These are API classes, unfortunately. And, they're used by the `DefaultOpenTelemetry` implementation, which is in a different package.

jkwatson:
> > I haven't looked at the SDK code to see the implications, but instead of hiding behind a factory could the "Default" classes just be `package` and not `public` classes?
> 
> These are API classes, unfortunately. And, they're used by the `DefaultOpenTelemetry` implementation, which is in a different package.

urgh. and the OT shim, as well.

kenfinnigan:
Shame, if all the usage was within the same package that would make things easier

carlosalberto:
Java should have implemented by now something like .Net's Friendly Assemblies 😃

jkwatson:
> Java should have implemented by now something like .Net's Friendly Assemblies 😃

https://openjdk.java.net/jeps/181

api/src/main/java/io/opentelemetry/api/metrics/DefaultMeter.java
anuraaga: It seemed slightly more cool than a boring constant but that's true :)

api/src/main/java/io/opentelemetry/api/metrics/Meter.java
anuraaga: Hmm - how about just `getDefault` then? An unfortunately generic name but it does give us flexibility to change so it seems like a better name.

anuraaga:
> Java should have implemented by now something like .Net's Friendly Assemblies 😃

I wish we could use them earlier than 20 years after they're released given normal Java uptake patterns :D

jkwatson:
> > Java should have implemented by now something like .Net's Friendly Assemblies 😃
> 
> I wish we could use them earlier than 20 years after they're released given normal Java uptake patterns :D

Yeah, and I don't even think there's really language-level support for making friendly packages with nestmates even now. If we were all in on java 9 JPMS, we could make it happen, of course. :)

api/src/main/java/io/opentelemetry/api/metrics/Meter.java
jkwatson: yeah, I think `getDefault` might be better, even though I much prefer saying "NOOP" loudly whenever I read the code. ;)

anuraaga:
Renamed to getDefault

api/src/test/java/io/opentelemetry/api/OpenTelemetryTest.java
```diff
@@ -59,10 +57,12 @@ void after() {
 
   @Test
   void testDefault() {
-    assertThat(OpenTelemetry.getGlobalTracerProvider()).isInstanceOf(DefaultTracerProvider.class);
+    assertThat(OpenTelemetry.getGlobalTracerProvider().getClass().getSimpleName())
+        .isEqualTo("DefaultTracerProvider");
```
jkwatson, line 61: I wonder how much value these tests are really providing.

jkwatson (approved):
Thanks!

api/src/test/java/io/opentelemetry/api/OpenTelemetryTest.java
anuraaga: I think they make sure we don't actually add a dependency on the SDK from the API and don't know if any of our other tests do. So that's probably good value.

PR #1906, opened 2020-10-28
Remove Context.with overloads.

anuraaga:
I noticed after integrating Context into the instrumentation we don't really need the `with` overloads. Also on a second thought, they seem to conflict with the recommendation in the javadoc to "combine multiple related items together into a single key instead of separating them" - it seems extremely rare you'd have key / value available to store at the same time even when they're not related (in our potential use case, it's basically when one of them is `ImplicitContextKeyed` in which case we don't have a method for combining into a single write anyways). So figured it's good to keep it small for now.

Also added some javadoc drift fixes.

### DIFF
--- a/context/src/main/java/io/opentelemetry/context/Context.java
+++ b/context/src/main/java/io/opentelemetry/context/Context.java
@@ -141,8 +141,8 @@ public interface Context {
   }
 
   /**
-   * Returns a {@link Runnable} that makes this the {@linkplain Context#current() current context} and
-   * then invokes the input {@link Runnable}.
+   * Returns a {@link Runnable} that makes this the {@linkplain Context#current() current context}
+   * and then invokes the input {@link Runnable}.
    */
   default Runnable wrap(Runnable runnable) {
     return () -> {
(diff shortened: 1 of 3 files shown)


bogdandrutu (commented):
Should we consider different names for the two `with` methods that we have now?

jkwatson (approved):
Reducing API surface is a good thing. Thanks!

jkwatson:
> Should we consider different names for the two `with` methods that we have now?

like maybe "with" and "withKeyed"?

jkwatson:
> > Should we consider different names for the two `with` methods that we have now?
> 
> like maybe "with" and "withKeyed"?

although, it sure is nice to just say `context.with(span)`

anuraaga:
Yeah I'd rather rename the explicit vs implicit one - `withKeyValue`? It's a bit long, not sure if it's better than just letting both be named `with` since they have identical semantics.

jkwatson:
> Yeah I'd rather rename the explicit vs implicit one - `withKeyValue`? It's a bit long, not sure if it's better than just letting both be named `with` since they have identical semantics.

I agree that it's super nice to have `with` for both of them..nice and concise. I'd vote to keep it as-is.

carlosalberto (approved):

jkwatson (approved):

PR #1907, opened 2020-10-28
Remove integration tests from coverage report.

anuraaga:
These don't have any coverage since the code is run in docker. A more general rule is to only have coverage of what we publish to maven - which I'll probably refactor this list into being generated programatically at some point.


jkwatson (approved):
Thanks!

PR #1913, opened 2020-10-28
Add a method to the Baggage interface to return the current baggage in the context

jkwatson:
This will make it easy for a user to modify existing Baggage, something like this:

```
Baggage updated = Baggage.current().toBuilder().put("newKey", "newValue").build();
try (Scope s = Context.with(update).makeCurrent()) {
  //do your thing
}
```

note: I had to update some other tests that were assuming that `Context.current()` was always clean and empty of Baggages.


carlosalberto (approved):

api/src/test/java/io/opentelemetry/api/baggage/BaggageTest.java
```diff
@@ -0,0 +1,27 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.opentelemetry.context.Context;
+import org.junit.jupiter.api.Test;
+
+class BaggageTest {
+  @Test
+  void current_empty() {
+    Context.root().makeCurrent();
```
anuraaga, line 16: try/resources the scopes - guessing you had to add the root.makeCurrent to the other tests because of missing closing here but we should have tests not interfere with each other :)

anuraaga: Though I don't think I'd make the root current here anyways, can leave this out to test "the default".

api/src/test/java/io/opentelemetry/api/baggage/BaggageUtilsTest.java
```diff
@@ -9,10 +9,16 @@
...
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.Scope;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 class BaggageUtilsTest {
 
+  @BeforeEach
+  void setUp() {
+    Context.root().makeCurrent();
```
anuraaga, line 19: Ditto, don't make the root current, it's the default.

api/src/test/java/io/opentelemetry/api/baggage/ScopedBaggageTest.java
```diff
@@ -29,6 +31,11 @@
       EntryMetadata.create("unlimited");
   private static final EntryMetadata METADATA_NO_PROPAGATION = EntryMetadata.create("noprop");
 
+  @BeforeEach
+  void setUp() {
```
anuraaga, line 35: Ditto

api/src/test/java/io/opentelemetry/api/baggage/BaggageUtilsTest.java
jkwatson: yeah, this was my tired brain doing something silly. Interestingly, it did fix up all the tests!

api/src/test/java/io/opentelemetry/api/baggage/BaggageTest.java
jkwatson: > Though I don't think I'd make the root current here anyways, can leave this out to test "the default".

assuming everyone else has properly closed their scopes! unlike me, apparently.

jkwatson:
@anuraaga ok...fixed up. sorry for the brain farts.

api/src/test/java/io/opentelemetry/api/baggage/ScopedBaggageTest.java
```diff
@@ -31,15 +30,12 @@
...
-  @BeforeEach
-  void setUp() {
-    Context.root().makeCurrent();
-  }
-
   @Test
   void emptyBaggage() {
-    Baggage defaultBaggage = BaggageUtils.getCurrentBaggage();
-    assertThat(defaultBaggage.getEntries()).isEmpty();
+    try (Scope s = Context.root().makeCurrent()) {
```
anuraaga, line 35: Let's remove the scope - just to make sure to test what happens when makeCurrent hasn't ever been called

anuraaga (approved):

PR #1914, opened 2020-10-29
Add MustBeClosed to Context.makeCurrent

anuraaga:
False positives happen only in interop cases where scopes are wrapped, but in vast amount of user code it should be precise. Agent will have to disable it though since we open scope in `onMethodEnter` and close in `onMethodExit`


jkwatson (approved):
Thanks!

PR #1915, opened 2020-10-29
Add workflow to copy docker hub images to bintray and use in integration tests.

anuraaga:
Docker Hub pull limit coming up, so let's copy our images to hopefully reduce potential flakiness. Working on hopefully getting a GHCR repo which would be less backdoor (tied to GitHub permissions, etc) than bintray.

### DIFF
--- a/.github/workflows/docker-test-containers.yml
+++ b/.github/workflows/docker-test-containers.yml
@@ -21,6 +21,7 @@ jobs:
       - name: Docker login
         uses: docker/login-action@v1
         with:
+          registry: open-telemetry-docker-dev.bintray.io
           username: ${{ secrets.BINTRAY_USER }}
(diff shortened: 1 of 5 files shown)


iNikem:
Can you explain your sentiment about ghcr vs bintray?

iNikem (approved):

anuraaga:
With GHCR, the permissions for the images can be managed similar to repos, with GitHub teams. With Bintray, it needs to be separate credentials and yet another team management. It's not a huge deal for Java since we are stuck managinng Bintray anyways, but I figure other languages may need to do similar due to the pull limits so GHCR should be easier for use generally in open-telemetry. I'd also expect it to be at least as fast when accessed from GitHub Actions, though that's less of a deal.

jkwatson (approved):
Thanks!!!

PR #1916, opened 2020-10-29
Have Baggage context interaction API mirror Span.

anuraaga:


carlosalberto (approved):

api/src/main/java/io/opentelemetry/api/baggage/BaggageUtils.java
```diff
@@ -5,18 +5,16 @@
...
-import io.opentelemetry.context.Scope;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
 /** Utility methods for accessing the {@link Baggage} contained in the {@link Context}. */
 @Immutable
 public final class BaggageUtils {
-  private static final ContextKey<Baggage> CORR_CONTEXT_KEY =
-      ContextKey.named("opentelemetry-corr-context-key");
+  private static final ContextKey<Baggage> BAGGAGE_KEY =
```
jkwatson, line 16: oh, good catch! good thing it's just a name. :)

api/src/main/java/io/opentelemetry/api/baggage/BaggageUtils.java
```diff
@@ -5,18 +5,16 @@
...
-import com.google.errorprone.annotations.MustBeClosed;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.ContextKey;
-import io.opentelemetry.context.Scope;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
 /** Utility methods for accessing the {@link Baggage} contained in the {@link Context}. */
 @Immutable
 public final class BaggageUtils {
```
jkwatson, line 15: do we still need the class to be public?

jkwatson (approved):
Just make that BaggageUtils non-public and it's good to go.  Thanks!

bogdandrutu:
Could be done in a separate PR.

api/src/main/java/io/opentelemetry/api/baggage/Baggage.java
```diff
@@ -39,6 +39,23 @@ static Baggage current() {
...
+  static Baggage fromContext(Context context) {
+    return BaggageUtils.getBaggage(context);
+  }
+
+  /**
+   * Returns the {@link Baggage} from the specified {@link Context}, or {@code null} if there is no
+   * baggage in the context.
+   */
+  @Nullable
+  static Baggage fromContextOrNull(Context context) {
```
bogdandrutu, line 55: Do we need this? People can compare with empty or can check for size of entries.

bogdandrutu: May be in a separate PR.

bogdandrutu (approved):

api/src/main/java/io/opentelemetry/api/baggage/BaggageUtils.java
carlosalberto: As long as `TracingContextUtils`, yes, I suggest we keep it public (and `TracingContextUtils` should probably stay public to be Specification compliant).

api/src/main/java/io/opentelemetry/api/baggage/Baggage.java
carlosalberto: +1

api/src/main/java/io/opentelemetry/api/baggage/BaggageUtils.java
jkwatson: even if none of the methods are public?

bogdandrutu: It does not need to be public, the specification says that the context interaction functionality can be added in few places, and avoid duplicate. So that needs to not be public.

carlosalberto: You guys can do whatever you want 😉  But if I may give my opinion, it looks really weird to be we have `TracingContextUtils` and no equivalent for `BaggageUtils` (why didn't we have public methods there, btw?)

Also, the specification says:

> they MAY be exposed as static methods on the trace module, as static methods on a class inside the trace module (it MAY be named TracingContextUtilities), or on the Tracer class.

Sure, it only mentions the trace portion, but I would have hoped it would be _similar_ for Baggage ;)

PR #1918, opened 2020-10-29
Rename getHasRemoteParent and getHasEnded to remove get.

### LINKED ISSUE #1910: Rename SpanData.getHasRemoteParent() to hasRemoteParent()
and also SpanData.getHasEnded() to SpanData.hasEnded()

anuraaga:
It's either all `get` or no `get` so this seems the way to allow a mix.

Alternatively, `isEnded` and `isParentRemote` would be allowed without hacking, but the latter sounds bad to me (but still better than `isRemoteParent` which is incorrect)

Fixes #1910


carlosalberto (approved):

jkwatson (approved):
Excellent! Autovalue will be brought into submission!

Issue #1808, opened 2020-10-15
gradle: Centralize animalsniffer signature dependencies


Oberon00:
Currently, every subproject seems to have these two signature lines:

```gradle
dependencies {
    signature "org.codehaus.mojo.signature:java18:1.0@signature"
    signature "net.sf.androidscents.signature:android-api-level-24:7.0_r2@signature"
}
```

It would be better to centralize that in the top-level build.gradle.

PR #1920, opened 2020-10-29
Restore source/targetCompatibility declarations for intellij to set version correctly and remo…

anuraaga:
…ve animalsniffer for java 8 since we set the `release` option now.

Thanks @tylerbenson for the callout.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -169,6 +169,9 @@ configure(opentelemetryProjects) {
         mavenLocal()
     }
 
+    sourceCompatibility = 1.8
+    targetCompatibility = 1.8
+
     java {
         toolchain {
             languageVersion = JavaLanguageVersion.of(11)
--- a/build.gradle
+++ b/build.gradle
@@ -169,10 +169,10 @@ configure(opentelemetryProjects) {
         mavenLocal()
     }
 
-    sourceCompatibility = 1.8
-    targetCompatibility = 1.8
-
     java {
+        sourceCompatibility = JavaVersion.VERSION_1_8
+        targetCompatibility = JavaVersion.VERSION_1_8
+
         toolchain {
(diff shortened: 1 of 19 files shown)


sdk_extensions/jaeger_remote_sampler/build.gradle
```diff
@@ -25,7 +25,6 @@ dependencies {
 
     testRuntime "io.grpc:grpc-netty-shaded:${grpcVersion}"
 
-    signature "org.codehaus.mojo.signature:java18:1.0@signature"
     signature "net.sf.androidscents.signature:android-api-level-24:7.0_r2@signature"
```
tylerbenson, line 29: Is it worth trying to put this string in a common place?

tylerbenson (approved):

build.gradle
```diff
@@ -169,6 +169,9 @@ configure(opentelemetryProjects) {
         mavenLocal()
     }
 
+    sourceCompatibility = 1.8
```
Oberon00, line 172: Could we use the new-style declarations as documented there? https://docs.gradle.org/current/userguide/java_plugin.html#sec:java-extension ?

```java
java {
    sourceCompatibility = JavaVersion.VERSION_1_8
    targetCompatibility = JavaVersion.VERSION_1_8
}
```

anuraaga: Sure

sdk_extensions/jaeger_remote_sampler/build.gradle
anuraaga: I think we have issue #1808 for that, let me look at it separately from this PR.

Oberon00 (approved):

jkwatson (approved):
Thanks!

Issue #1912, opened 2020-10-28
Update our OTLP protobuf dependency to the v0.6.0 release


jkwatson:
https://github.com/open-telemetry/opentelemetry-proto/releases/tag/v0.6.0

PR #1921, opened 2020-10-29
Update opentelemetry-proto to latest version.

dengliming:
closes [https://github.com/open-telemetry/opentelemetry-java/issues/1912](https://github.com/open-telemetry/opentelemetry-java/issues/1912)

### DIFF
--- a/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/SpanAdapter.java
+++ b/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/SpanAdapter.java
@@ -172,7 +172,7 @@ final class SpanAdapter {
     // todo: Update this when the proto definitions are updated to include UNSET and ERROR
     Status.StatusCode protoStatusCode = Status.StatusCode.STATUS_CODE_OK;
     if (status.getCanonicalCode() == StatusCode.ERROR) {
-      protoStatusCode = Status.StatusCode.STATUS_CODE_UNKNOWN_ERROR;
+      protoStatusCode = Status.StatusCode.STATUS_CODE_ERROR;
     }
     Status.Builder builder = Status.newBuilder().setCode(protoStatusCode);
     if (status.getDescription() != null) {
--- a/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/SpanAdapter.java
+++ b/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/SpanAdapter.java
@@ -178,7 +178,7 @@ final class SpanAdapter {
       deprecatedStatusCode = DEPRECATED_STATUS_CODE_UNKNOWN_ERROR;
     }
 
-    @SuppressWarnings("deprecation")
+    @SuppressWarnings("deprecation") // setDeprecatedCode is deprecated.
     Status.Builder builder =
         Status.newBuilder().setCode(protoStatusCode).setDeprecatedCode(deprecatedStatusCode);
     if (status.getDescription() != null) {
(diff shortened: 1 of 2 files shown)


exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/SpanAdapter.java
```diff
@@ -172,7 +172,7 @@ static Status toStatusProto(SpanData.Status status) {
     // todo: Update this when the proto definitions are updated to include UNSET and ERROR
     Status.StatusCode protoStatusCode = Status.StatusCode.STATUS_CODE_OK;
     if (status.getCanonicalCode() == StatusCode.ERROR) {
-      protoStatusCode = Status.StatusCode.STATUS_CODE_UNKNOWN_ERROR;
+      protoStatusCode = Status.StatusCode.STATUS_CODE_ERROR;
```
jkwatson, line 175: I think there's a little more we need to do here. See the comments in the proto:  https://github.com/open-telemetry/opentelemetry-proto/blob/master/opentelemetry/proto/trace/v1/trace.proto#L239-L249

dengliming: When `status.getCanonicalCode()` is `StatusCode.UNSET`, the `protoStatusCode ` should be set to `Status.StatusCode.STATUS_CODE_OK`. Is that right?

jkwatson: I think we're also supposed to be setting the `deprecated_code` as well, if I read those comments correctly.

dengliming: Can you help to confirm again? Thanks.

exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/SpanAdapter.java
```diff
@@ -169,12 +171,16 @@ static Span toProtoSpan(SpanData spanData) {
   }
 
   static Status toStatusProto(SpanData.Status status) {
-    // todo: Update this when the proto definitions are updated to include UNSET and ERROR
     Status.StatusCode protoStatusCode = Status.StatusCode.STATUS_CODE_OK;
```
Oberon00, line 174: I think there is a mistake here: For UNSET, we should use UNSET, not OK.

Oberon00 (changes requested):

exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/SpanAdapter.java
dengliming: Thanks. Can you review again?

Oberon00 (approved):

exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/SpanAdapter.java
Oberon00: Thanks, looks good now!

exporters/otlp/src/test/java/io/opentelemetry/exporters/otlp/SpanAdapterTest.java
```diff
@@ -109,17 +112,35 @@ void toProtoSpanKind() {
   }
 
   @Test
+  @SuppressWarnings("deprecation") // setDeprecatedCode is deprecated.
```
jkwatson, line 115: this line of code makes me laugh, with the comment.  😆

jkwatson (approved):
Thanks! Looks good.

exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/SpanAdapter.java
jkwatson: looks good!

PR #1922, opened 2020-10-29
Add a creation method to the TraceShim that takes an OTel instance.

jkwatson:
Also, update the TraceContext integration test to use an OTel instance.


anuraaga (approved):
Nice

PR #1925, opened 2020-10-29
Initial list of best practices to follow in our project

bogdandrutu:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,6 +43,13 @@ which uses [google-java-format](https://github.com/google/google-java-format) li
 
 `./gradlew spotless`
 
+### Best practices that we follow
+
+* Avoid exposing publicly any class/method/variable that don't need to be public;
+* By default, all arguments/members are treated as non-null. Every argument/member that can be `null` must be annotated with `@Nullable`;
+* The project aims to provide a consistent experience across all the public APIs. It is important to ensure consistency (same look and feel) across different public packages;
+* Use `final` everywhere is possible, this ensures that classes cannot be extended when the API does not intend to offer that functionality;
+
 ### Pre-commit hook
 To completely delegate code style formatting to the machine, 
 you can add [git pre-commit hook](https://git-scm.com/docs/githooks).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,7 +48,7 @@ which uses [google-java-format](https://github.com/google/google-java-format) li
 * Avoid exposing publicly any class/method/variable that don't need to be public;
 * By default, all arguments/members are treated as non-null. Every argument/member that can be `null` must be annotated with `@Nullable`;
 * The project aims to provide a consistent experience across all the public APIs. It is important to ensure consistency (same look and feel) across different public packages;
-* Use `final` everywhere is possible, this ensures that classes cannot be extended when the API does not intend to offer that functionality;
+* Use `final` classes everywhere it is possible, this ensures that these classes cannot be extended when the API does not intend to offer that functionality;
 
 ### Pre-commit hook
 To completely delegate code style formatting to the machine, 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,10 +45,12 @@ which uses [google-java-format](https://github.com/google/google-java-format) li
 
 ### Best practices that we follow
 
-* Avoid exposing publicly any class/method/variable that don't need to be public;
-* By default, all arguments/members are treated as non-null. Every argument/member that can be `null` must be annotated with `@Nullable`;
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -43,6 +43,13 @@ which uses [google-java-format](https://github.com/google/google-java-format) li
 
 `./gradlew spotless`
 
+### Best practices that we follow
+
+* Avoid exposing publicly any class/method/variable that don't need to be public;
+* By default, all arguments/members are treated as non-null. Every argument/member that can be `null` must be annotated with `@Nullable`;
+* The project aims to provide a consistent experience across all the public APIs. It is important to ensure consistency (same look and feel) across different public packages;
+* Use `final` everywhere is possible, this ensures that classes cannot be extended when the API does not intend to offer that functionality;
```
jkwatson, line 51:
```suggestion
* Use `final` classes everywhere it is possible, this ensures that these classes cannot be extended when the API does not intend to offer that functionality;
```

jkwatson (approved):
good start!

CONTRIBUTING.md
```diff
@@ -43,6 +43,13 @@ which uses [google-java-format](https://github.com/google/google-java-format) li
 
 `./gradlew spotless`
 
+### Best practices that we follow
+
+* Avoid exposing publicly any class/method/variable that don't need to be public;
+* By default, all arguments/members are treated as non-null. Every argument/member that can be `null` must be annotated with `@Nullable`;
+* The project aims to provide a consistent experience across all the public APIs. It is important to ensure consistency (same look and feel) across different public packages;
+* Use `final` classes everywhere it is possible, this ensures that these classes cannot be extended when the API does not intend to offer that functionality;
```
anuraaga, line 51:
```suggestion
* Avoid exposing publicly any class/method/variable that don't need to be public.
* By default, all arguments/members are treated as non-null. Every argument/member that can be `null` must be annotated with `@Nullable`.
* The project aims to provide a consistent experience across all the public APIs. It is important to ensure consistency (same look and feel) across different public packages. If you notice any practice being applied in the project that isn't listed here, it would be great to add it.
* Use `final` classes everywhere it is possible, this ensures that these classes cannot be extended when the API does not intend to offer that functionality.
```

anuraaga: Just some formatting and a note about adding to these when something's found.

For `final` classes do you mean public? I tend to not nit for package-private classes though we could be more strict about that if the consistency helps.

anuraaga (approved):

CONTRIBUTING.md
jkwatson: I agree. It's most important for public classes on the API. Let's not go overboard if we don't have a good reason to.

anuraaga: Cool - edited my suggestion to add "for public" to final

iNikem: > If you notice any practice being applied in the project that isn't listed here, it would be great to add it

I don't understand this sentence in the context of consistent experience. Also "would be great to" is not particularly actionable :)

anuraaga: I thought there may be things noticed that are consistent but not documented here so wanted to encourage people adding them. Is there a better wording, or should I remove it if it's too unclear?

iNikem: Encouragement may be good, but I think it should be as a separate sentence then. Not part of consistency section. Because that section does not actually list anything, so I got confused: "listed here" - where?

anuraaga: Ah I see yeah - tried separating it out

bogdandrutu: I would prefer for all the classes to use final (for consistency and showing clearly the intent), but definitely it is very important for everything that is public.

bogdandrutu: Also @jkwatson made that correction for only classes, but methods as well in case the class is not final for whatever reason.

carlosalberto (approved):

tylerbenson (approved):

CONTRIBUTING.md
```diff
@@ -43,6 +43,15 @@ which uses [google-java-format](https://github.com/google/google-java-format) li
 
 `./gradlew spotless`
 
+### Best practices that we follow
+
+* Avoid exposing publicly any class/method/variable that don't need to be public.
+* By default, all arguments/members are treated as non-null. Every argument/member that can be `null` must be annotated with `@Nullable`.
+* The project aims to provide a consistent experience across all the public APIs. It is important to ensure consistency (same look and feel) across different public packages.
+* Use `final` for public classes everywhere it is possible, this ensures that these classes cannot be extended when the API does not intend to offer that functionality.
```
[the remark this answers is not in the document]
anuraaga: Mockito can mock final classes now adays and it's awesome :)

PR #1926, opened 2020-10-29
Add some simplistic fuzz-based tests for the 2 built-in W3C propagators.

### LINKED ISSUE #1770: Verify the safety of the propagators with a fuzzing utility
It is very important that our propagator implementations are safe. That is, that they will not crash or cause problems with running programs in the face of arbitrary inputs. In the case of propagators, that data is coming off the wire from potentially hostile sources.

I suggest we implement some fuzz-based testing of the extract methods of the propagators, to make sure that they will fail-safe and not throw exceptions in the face of bad inputs.

jkwatson:
Resolves #1770

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -2,7 +2,6 @@ plugins {
     id "java"
     id "maven-publish"
 
-    id "org.unbroken-dome.test-sets"
     id "me.champeau.gradle.jmh"
     id "ru.vyarus.animalsniffer"
 }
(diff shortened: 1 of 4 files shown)


jkwatson:
I think we should probably, if we have time, create tests like this for all parts of the API to make sure that the API is 100% safe to use.

anuraaga (approved):
Cool

PR #1929, opened 2020-10-30
Remove BaggageManager for real.

anuraaga:
We removed the usage but file was still leftover.


jkwatson (approved):
I could have sworn that we had already done this last week. Thanks for the cleanup!

PR #1931, opened 2020-10-30
Remove BaggageUtils

anuraaga:
Since it's private, it's relatively harmless to have, but does cause the JVM optimizer a bit more work to inline more methods and somewhat abstracts the fact that BaggageUtils is really just a holder of a context key so seems nice to remove. Another option is to rename to `BaggageContextKey` with the lone field inside, and still removing the methods. Or just not do anything is fine too but lean towards removing this.


jkwatson (approved):
If for some reason we feel like this is necessary in the future, it would be easy enough to add it back. Removing unneeded code is a win, for sure. Thanks!

carlosalberto:
Although this change is OK, it feels _very_ weird to have the key in a class called `ImmutableBaggage`. Isn't this key used for **all** `Baggage` instances?

api/src/test/java/io/opentelemetry/api/baggage/BaggageContextTest.java
```diff
@@ -24,7 +24,7 @@ void testGetCurrentBaggage_Default() {
   @Test
   void testGetCurrentBaggage_SetCorrContext() {
```
jkwatson, line 25: this test method name is pretty out of date..still referring to it as "correlationContext"

jkwatson:
> Although this change is OK, it feels _very_ weird to have the key in a class called `ImmutableBaggage`. Isn't this key used for **all** `Baggage` instances?

I'm sympathetic to this point of view. Maybe we want a simple package-private "BaggageKey" class that holds on to the key?

PR #1933, opened 2020-10-30
Add test cases for array attributes with null inside.

anuraaga:
Don't think I saw tests for this but they're worth special care.


jkwatson (approved):
Thanks!

PR #1934, opened 2020-10-30
Replace EndSpanOptions.Builder with factory.

anuraaga:
We don't have to merge this, this is just an idea. Do we have any short, mid-term plans / ideas for adding end span options? If not, we can expose a simple factory instead of a builder for it. It's compatible, and pretty common I guess, to add a builder in the future when there are too many options, but right now we only have one :)


api/src/main/java/io/opentelemetry/api/trace/EndSpanOptions.java
```diff
@@ -15,22 +15,21 @@
...
 @AutoValue
 public abstract class EndSpanOptions {
-  private static final EndSpanOptions DEFAULT = builder().build();
+  private static final EndSpanOptions DEFAULT = createWithEndTimestamp(0L);
+
+  /**
+   * Returns a {@link EndSpanOptions} indicating the span ended at the given {@code endTimestamp},
+   * in nanoseconds.
+   */
+  public static EndSpanOptions createWithEndTimestamp(long endTimestamp) {
```
jkwatson, line 24: maybe just `create(long endTimestamp)` ?

jkwatson:
I agree. A builder for a single field seems like silly overkill, and, as you mention, we can add one later if we find we need more options.

bogdandrutu (commented):
The reason to do this way was mainly because this is a very rare case where end timestamp will be provided. So we decided to choose the most flexible and proven pattern for future changes.

If we believe that we will not have a need to extend this, an alternative is to overload the end span to accept an endTimestamp and completely remove this class.

anuraaga:
I would be good with removing this class too since I can't imagine more options. But if we can foresee any additional options, a class seems ok too - but the builder seems like overkill for now which we could add later.

api/src/main/java/io/opentelemetry/api/trace/EndSpanOptions.java
anuraaga: Renamed - I think either this or removing the class and inlining into `Span` are both good options.

carlosalberto:
> would be good with removing this class too since I can't imagine more options. 

We can always add it 'back' if needed, so I'm +1 on removing it.

anuraaga:
Went ahead and removed `EndSpanOptions`. Will change to accept `TimeUnit` after merging this in #1969

api/src/main/java/io/opentelemetry/api/trace/Span.java
```diff
@@ -304,17 +304,18 @@ default Span setAttribute(AttributeKey<Long> key, int value) {
...
    *
    * <p>Only the timing of the first end call for a given {@code Span} will be recorded, and
    * implementations are free to ignore all further calls.
    *
    * <p>Use this method for specifying explicit end options, such as end {@code Timestamp}. When no
    * explicit values are required, use {@link #end()}.
    *
-   * @param endOptions the explicit {@link EndSpanOptions} for this {@code Span}.
+   * @param timestamp the explicit timestamp, as nanos from the epoch, for this {@code Span}. {@code
+   *     0} indicates current time should be used.
```
jkwatson, line 316: As usual, I'm not a fan of using `0` as a magic number here. Would prefer something that was obviously not a valid timestamp, Long.MIN_VALUE or something.

trask: does this method need any magic for current time, since `end()` supports that?

anuraaga: I don't think so either but that's the same logic as we have right now - can we update the semantics separately? Want to get the API locked in

anuraaga: I wonder if no one noticed if I hadn't updated javadoc :-)

jkwatson: I did see the zeros floating around in tests and such and it was starting to make me grumpy, even before I saw the javadoc. :)

jkwatson: Before this PR, zero was an undocumented magic number. I guess it's better to have it documented, if we're going to use it! But, I'd rather not have the magic '0' at all, personally. I mean, maybe you want a span to end at unix epoch! (ok, not really)

jkwatson (approved):
Approved, to at least lock down the API, even if we decide to tweak the default value in the future, pre GA

carlosalberto (approved):

PR #1935, opened 2020-10-30
Split SpanContext into interface / impl

anuraaga:
As enabled by https://github.com/open-telemetry/opentelemetry-specification/pull/969


Oberon00:
This seems to make the code more complicated and arguably does not implement the "SHOULD" in "The API MUST implement methods to create a SpanContext. These methods SHOULD be the only way to create a SpanContext." since you can now create a SpanContext by deriving from the interface and instantiate the derived class.
So can you please state the reasons for these changes here?

api/src/main/java/io/opentelemetry/api/trace/ImmutableSpanContext.java
```diff
@@ -0,0 +1,51 @@
...
+
+package io.opentelemetry.api.trace;
+
+import com.google.auto.value.AutoValue;
+import com.google.auto.value.extension.memoized.Memoized;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+@AutoValue
+abstract class ImmutableSpanContext implements SpanContext {
```
Oberon00, line 14: Since the interface is also annotated `@Immutable`  and according to the spec any SpanContext must be Immutable, this is maybe not the best name to distinguish it from possible other implementations which would also be Immutable. Maybe DefaultSpanContext?

anuraaga:
@Oberon00 Weren't all the reasons in the long discussion in https://github.com/open-telemetry/opentelemetry-specification/pull/969 and https://github.com/open-telemetry/opentelemetry-specification/issues/970 ? :-) The short answer is for agent users, it will always be the agent's version of the API, not the user's version of the API, that instantiates a `SpanContext`. I thought we agreed this is ok in that issue.

Also one point of note is that we don't lose any protection with this PR since it's already an abstract class. Of course, we could stop using AutoValue to prevent that, just an observation. With #1937 I am thinking of a general question on whether we should just avoid classes at all in the API - it's an API after all. If a custom SDK (which our auto instrumentation agent effectively is) finds a reason to override the implementation and satisfies a use case with it, why block it?

api/src/main/java/io/opentelemetry/api/trace/ImmutableSpanContext.java
anuraaga: Yeah I had the same impression - we have `Immutable*` as a pattern in many AutoValue implementations of interfaces currently. Renaming them all to `Default` seems fine to me, or even `*Impl` so the implementation orders next to the interface in the filetree.

Oberon00:
I suspected as much 😃 But I thought it was better to document it on the PR that does the change.

So the interface makes it even easier for the auto-instrumentation than the previous abstract class?

anuraaga:
@Oberon00 Yeah - one class can implement both the agent API interface and application API interface, but with abstract classes we wouldn't be able to do that.

api/src/main/java/io/opentelemetry/api/trace/ImmutableSpanContext.java
jkwatson: No strong opinion. I'd keep this name for this PR, and we can always rename later, since this is an internal detail.

jkwatson (approved):
If it helps auto-instrumentation and doesn't hurt the end-user experience, I'm all in favor.

tylerbenson (approved):

api/src/main/java/io/opentelemetry/api/trace/SpanContext.java
```diff
@@ -17,23 +15,15 @@
  * traceState} and the {@link boolean remote} flag.
  */
 @Immutable
```
tylerbenson, line 17: Does this still make sense on an interface?

tylerbenson:
Can this please be included in the `0.10.0` release?

api/src/main/java/io/opentelemetry/api/trace/ImmutableSpanContext.java
bogdandrutu, line 14: Not sure why this class is needed, if you just move the "@AutoValue" annotation on the SpanContext you achieve the same thing, and just call directly the ctor of the AutoValue, am I missing something?

bogdandrutu (dismissed):
Blocked until https://github.com/open-telemetry/opentelemetry-java/issues/1946

jkwatson:
After further thinking, I'd prefer SpanContext stay not as an interface, but as a final class (or AutoValue'd class). It's a core piece of data for the API, and since it's data-only, having it be non-interface is preferable to me.

api/src/main/java/io/opentelemetry/api/trace/ImmutableSpanContext.java
anuraaga: As far as I know, `AutoValue` doesn't support interfaces still.

bogdandrutu:
@anuraaga @trask we are getting back to "auto-instrumentation" requires this, but not documented. Please make sure that this requirement is documented, not clear why this cannot happen, seems like a correct change.

jkwatson (approved):
Thanks!

trask (approved):

trask:
@bogdandrutu can you unblock this PR, or comment on #1946 which you blocked this on?

jkwatson:
> After further thinking, I'd prefer SpanContext stay not as an interface, but as a final class (or AutoValue'd class). It's a core piece of data for the API, and since it's data-only, having it be non-interface is preferable to me.

Note: although I usually favor data-as-classes, I'm on board with this change.

api/src/main/java/io/opentelemetry/api/trace/SpanContext.java
```diff
@@ -128,22 +104,21 @@ public void copyTraceFlagsHexTo(char[] dest, int destOffset) {
...
    * Returns {@code true} if this {@code SpanContext} is valid.
    *
    * @return {@code true} if this {@code SpanContext} is valid.
    */
-  @Memoized
-  public boolean isValid() {
-    return TraceId.isValid(getTraceIdHex()) && SpanId.isValid(getSpanIdHex());
+  default boolean isValid() {
+    return TraceId.isValid(getTraceIdAsHexString()) && SpanId.isValid(getSpanIdAsHexString());
   }
```
bogdandrutu, line 116: this will become more expensive after this change, we should calculate this only once since it is used couple of times per instance created.

api/src/main/java/io/opentelemetry/api/trace/SpanContext.java
```diff
@@ -66,60 +51,51 @@ private static SpanContext create(
...
   /**
    * Returns the byte[] representation of the span identifier associated with this {@link
    * SpanContext}.
    */
-  @Memoized
-  public byte[] getSpanIdBytes() {
-    return SpanId.bytesFromHex(getSpanIdHex(), 0);
+  default byte[] getSpanIdBytes() {
+    return SpanId.bytesFromHex(getSpanIdAsHexString(), 0);
   }
```
bogdandrutu, line 88: Why removing the ability to memoized this call? It may get expensive. I would suggest to not default this and any method that was memoized before.

bogdandrutu:
My concern still stays on a general level. If there was no auto-instrumentation project majority of the engineers have implemented this as a final class, so I am concerned that we do designed choices based on an external project need that may influence the overall usability and performance of the core product. This designed choices may affect us in the future in an unknown way.

I do understand the needs coming from auto-instrumentation, but you need to keep in mind that right now we need to be prepared to deal with cases where users will implement this in a "unexpected" way like having mutable objects inside the SpanContext, etc. this may cause problems for us, and we may have to deal with this.

api/src/main/java/io/opentelemetry/api/trace/SpanContext.java
jkwatson: This is still memoized by the autovalue implementation. See ImmutableSpanContext above.

api/src/main/java/io/opentelemetry/api/trace/SpanContext.java
jkwatson: Also, see above...this is still memoized by the implementation.

anuraaga:
Thanks! Added a warning

jkwatson:
@marcingrzejszczak this is potentially a breaking change

PR #1936, opened 2020-10-30
Reduce max workers since Gradle cheats anyways

anuraaga:
I just noticed that this repo also has the same problem as instrumentation repo, which I filed an issue about in https://github.com/gradle/gradle/issues/14224

The actual max workers is 8, which is high enough to cause a reasonable probability of two testcontainers at the same time which usually fails. One alternative is to combine all of our testcontainer tests into one project, integration-tests, so they don't run in parallel. This seems like an OK approach to me. In the meantime, let's match the value with the instrumentation repo, which I think can help with our testcontainer woes.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 org.gradle.parallel=true
-org.gradle.workers.max=4
+org.gradle.workers.max=2
 org.gradle.jvmargs=-Xmx4096m -XX:MaxMetaspaceSize=512m
 
 ### Override this property to 'true' to enable the end-to-end tests that use docker.


iNikem:
> two testcontainers at the same time which usually fails

This is not OK. Testcontainers SHOULD run just fine in parallel. If you suspect otherwise in this repo, then this problem must be solved. And not worked around by re-arranging our tests.

Unless you suspect that they fail due to insufficient resources in GHA. This may be the only "valid reason to fail" :)

anuraaga:
I looked at a few slack threads mentioning our error message in testcontainers slack and the response was generally reduce your parallelism. I don't think two simultaneous testcontainers is guaranteed to fail, but due to the resource situation has a much higher chance of it.

anuraaga:
@iNikem Anyways while I am hopeful this setting can reduce testcontainers flakiness, regardless, since the current settings result in 8 workers, maxing out a 4-core machine, I guess they're a bit too agressive and was the reason we went with `2` in the instrumentation repo, so sounds ok to match up here?

jkwatson:
I'm going to re-run the build jobs a few times to see if TC fails with this configuration.

jkwatson:
4-5 runs and everything looks good. I'm going to approve and merge, even if it's not a perfect solution.

jkwatson (approved):
🤞 this fixes most of the transient failures.

PR #1937, opened 2020-10-30
Replace AutoValue usage in baggage with final concrete classes.

anuraaga:
Related to #1925. I think making public classes in the API final is a good idea, but we currently have many public abstract classes for the use of AutoValue. I think we should be consistent for public API in that all extensible classes are interface, and all non-extensible ones are final. If that means implementing (e.g., hitting some keys in IntelliJ) these methods, I think it's better than the current inconsistency. 

Here I went with making baggage classes concrete as an example - alternatively splitting interfaces and package-private AutoValue classes would make the extensibility consistent with other interfaces.

### DIFF
--- a/api/src/main/java/io/opentelemetry/api/baggage/EntryMetadata.java
+++ b/api/src/main/java/io/opentelemetry/api/baggage/EntryMetadata.java
@@ -5,7 +5,6 @@
 
 package io.opentelemetry.api.baggage;
 
-import com.google.auto.value.AutoValue;
 import javax.annotation.concurrent.Immutable;
 
 /**
@@ -13,20 +12,26 @@ import javax.annotation.concurrent.Immutable;
  * wrapper for a String metadata value.
  */
 @Immutable
-@AutoValue
-public abstract class EntryMetadata {
-  public static final EntryMetadata EMPTY = create("");
+public final class EntryMetadata {
+  private static final EntryMetadata EMPTY = create("");
 
-  EntryMetadata() {}
+  /** Returns an {@link EntryMetadata} with no value. */
+  public static EntryMetadata empty() {
+    return EMPTY;
+  }
 
-  /**
-   * Creates an {@link EntryMetadata} with the given value.
-   *
-   * @param metadata TTL of an {@code Entry}.
-   * @return an {@code EntryMetadata}.
-   */
+  /** Returns an {@link EntryMetadata} with the given value. */
   public static EntryMetadata create(String metadata) {
-    return new AutoValue_EntryMetadata(metadata);
+    if (metadata == null) {
+      return empty();
+    }
+    return new EntryMetadata(metadata);
+  }
+
+  private final String metadata;
+
+  private EntryMetadata(String metadata) {
+    this.metadata = metadata;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


api/src/main/java/io/opentelemetry/api/baggage/EntryMetadata.java
```diff
@@ -5,34 +5,63 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    return metadata.hashCode();
+  }
+
+  @Override
+  public String toString() {
+    return "'" + metadata + "'";
```
anuraaga, line 65: Changed this a bit, I think the previous autovalue implementation would have been `EntryMetadata{metadata=foo}`

anuraaga:
@trask Maybe can leave a note too, I guess any final class we introduce will make the bridging in the agent harder so maybe just Interface for all! is a better approach for our instrumentation.

bogdandrutu:
Ok, I think you misinterpreted the section about final in the guideline. The idea is that we should not let user extend the class because we don't want to deal with unexpected behavior, but in this class this cannot be extended because the ctor is package protected.

Also I would not make the API decisions based on auto-instrumentation. Making every class an interface adds unnecessary overhead for users that do not use auto instrumentation (yes every function call needs a jump to the vtable to resolve the actual method address)

bogdandrutu:
So in this case, by making Baggage an interface forces us to deal with cases where a third-party implementation of the interface is passed to our APIs, and that may complicate things unnecessary and may disallow future improvements that we can make.

api/src/main/java/io/opentelemetry/api/baggage/EntryMetadata.java
jkwatson: I think the old version was a bit better. Any time we make a `toString()` look too usable by an end-user for something other than debugging, I think we open ourselves up to people mis-using it, which would lock us into a particular implementation.

anuraaga:
Filed #1946 to discuss interfaces vs classes.

bogdandrutu:
> I've heard this comment a lot, but I've never heard of an actual example of this happening in the real world. Can you show a case where you've been actually impacted by this in the past? Java 8 default methods let us evolve interfaces in a similar way that we've been able to evolve abstract classes, pre Java-8.

FYI: This is not about evolving the API, but about dealing with third-party implementation of the interface. I am advocating to not allow other implementations for classes that we don't need to offer that capability.

Yes, we had the problem in OpenCensus. We had an interface that represented the Tags (a.k.a baggage) that did not expose publicly a get method, and have our own implementation that exposed (package protected) a get method. Now because that was an interface and everyone can extend the interface it was not trivial to convert to the internal implementation because we had a path where we have to deal with another implementation that does not have our functionality.

You may argue that all functionality must be public, but that is not necessary true if you don't want user to have access to that or abuse your API. By restricting who can implement an interface/abstract class you can rely on package protected functionality to optimize performance etc.

jkwatson:
> Yes, we had the problem in OpenCensus. We had an interface that represented the Tags (a.k.a baggage) that did not expose publicly a get method, and have our own implementation that exposed (package protected) a get method. Now because that was an interface and everyone can extend the interface it was not trivial to convert to the internal implementation because we had a path where we have to deal with another implementation that does not have our functionality.
> 
> You may argue that all functionality must be public, but that is not necessary true if you don't want user to have access to that or abuse your API. By restricting who can implement an interface/abstract class you can rely on package protected functionality to optimize performance etc.

It sounds like you were casting to your implementation below the API surface? Yes, that does sound like it would get you in trouble, for sure. I don't think that's a good reason not to use interfaces in your API, though. You just need to make sure you don't rely on special implementation-only functionality if you do expose your data as an interface.

In general, I think if what you are exposing is *pure data*, then having a final class is fine. If you're exposing functionality, I will always prefer to have an interface for that functionality.

In the particular case of Baggage...  We have pure, immutable data, so I'm totally 100% good with having a final class at the API layer for people to use.  

Note: I'll also update the linked issue around interfaces with these comments. :)

anuraaga:
I don't think this PR is necessary since indeed the private constructor, which I forgot about, does make it effectively final. But still continuing the discussion on interfaces!

PR #1938, opened 2020-10-30
Add a StrictContextStorage which keeps track of scope open / close to…

anuraaga:
… throw errors when it's incorrect for tests.

If this looks good and merged, I'd followup with a junit extension which when applied, automatically checks scopes without additional code. JUnit extensions can also be applied globally via properties file, applying it to builds in an easy way.

Mostly copied from https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/StrictScopeDecorator.java

Part of #1794

### DIFF
--- a/context/src/main/java/io/opentelemetry/context/LazyStorage.java
+++ b/context/src/main/java/io/opentelemetry/context/LazyStorage.java
@@ -56,16 +56,12 @@ final class LazyStorage {
     return storage;
   }
 
-  static void set(ContextStorage storage) {
-    LazyStorage.storage = storage;
-  }
-
   private static final String CONTEXT_STORAGE_PROVIDER_PROPERTY =
       "io.opentelemetry.context.contextStorageProvider";
 
   private static final Logger logger = Logger.getLogger(LazyStorage.class.getName());
 
-  private static ContextStorage storage;
+  private static final ContextStorage storage;
 
   static {
     AtomicReference<Throwable> deferredStorageFailure = new AtomicReference<>();
@@ -83,11 +79,18 @@ final class LazyStorage {
 
     List<ContextStorageProvider> providers = new ArrayList<>();
     for (ContextStorageProvider provider : ServiceLoader.load(ContextStorageProvider.class)) {
+      if (provider
+          .getClass()
+          .getName()
+          .equals("io.opentelemetry.sdk.testing.context.SettableContextStorageProvider")) {
+        // Always use our testing helper context storage provider if it is on the classpath.
+        return provider.get();
+      }
       providers.add(provider);
     }
 
     if (providers.isEmpty()) {
-      return DefaultContext.threadLocalStorage();
+      return ContextStorage.defaultStorage();
     }
 
     if (providerClassName.isEmpty()) {
--- a/context/src/main/java/io/opentelemetry/context/LazyStorage.java
+++ b/context/src/main/java/io/opentelemetry/context/LazyStorage.java
@@ -56,12 +56,16 @@ final class LazyStorage {
     return storage;
   }
 
+  static void set(ContextStorage storage) {
+    LazyStorage.storage = storage;
+  }
+
   private static final String CONTEXT_STORAGE_PROVIDER_PROPERTY =
       "io.opentelemetry.context.contextStorageProvider";
 
   private static final Logger logger = Logger.getLogger(LazyStorage.class.getName());
 
-  private static final ContextStorage storage;
+  private static ContextStorage storage;
 
   static {
     AtomicReference<Throwable> deferredStorageFailure = new AtomicReference<>();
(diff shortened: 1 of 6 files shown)


context/src/main/java/io/opentelemetry/context/LazyStorage.java
```diff
@@ -56,12 +56,16 @@ static ContextStorage get() {
...
+    LazyStorage.storage = storage;
+  }
+
   private static final String CONTEXT_STORAGE_PROVIDER_PROPERTY =
       "io.opentelemetry.context.contextStorageProvider";
 
   private static final Logger logger = Logger.getLogger(LazyStorage.class.getName());
 
-  private static final ContextStorage storage;
+  private static ContextStorage storage;
```
anuraaga, line 68: Intentionally not volatile for now - maybe it's too optimistic to hope `set` is only ever called by a user at the very beginning of the map but unlike `OpenTelemetry` itself this is a very advanced API.

anuraaga: An idea that comes to mind is having ResettableContextStorageProvider in the testing package. I would need to add a slightly hacky check for that provider in `LazyStorage.createStorage()` to use it + another normally SPI-loaded or default contextstorage when it's present but it doesn't seem so bad actually. Let me play with it.

anuraaga: Was able to remove `set(ContextStorage)` in the main artifact - there is some crosstalk where the prod artifact is aware of the classname of the test helper, and the test helper duplicates some logic from the prod logic, but it's all contained in implementation and seems safe to me. Added `defaultStorage` API which was needed anyways to support custom storage implementations delegating to the standard.

context/src/main/java/io/opentelemetry/context/LazyStorage.java
```diff
@@ -79,11 +79,18 @@ static ContextStorage createStorage(AtomicReference<Throwable> deferredStorageFa
 
     List<ContextStorageProvider> providers = new ArrayList<>();
     for (ContextStorageProvider provider : ServiceLoader.load(ContextStorageProvider.class)) {
+      if (provider
+          .getClass()
+          .getName()
+          .equals("io.opentelemetry.sdk.testing.context.SettableContextStorageProvider")) {
```
jkwatson, line 85: Could we create an interface for this, rather than relying on the classname explicitly?

sdk/testing/src/test/java/io/opentelemetry/sdk/testing/context/StrictContextStorageTest.java
```diff
@@ -0,0 +1,181 @@
...
+  void decorator_close_afterCorrectUsage() {
+    try (Scope ws = Context.current().with(ANIMAL, "cat").makeCurrent()) {
+      try (Scope ws2 = Context.current().with(ANIMAL, "dog").makeCurrent()) {}
+    }
+
+    strictStorage.ensureAllClosed(); // doesn't error
+  }
+
+  @Test
+  public void doesntDecorateNoop() {
```
jkwatson, line 89: Wouldn't we want to decorate the no-op scope, to make sure that even those are being properly closed? I can imagine use-cases where someone would accidentally end up creating a no-op scope in a test, but in the real world it would end up being a real scope, and then we wouldn't flag it as a problem.

context/src/main/java/io/opentelemetry/context/LazyStorage.java
anuraaga: That would add a public API for this testing use case. I'd rather keep this implementation detail for now and if another use case comes up can add an interface for a "forced context storage provider" - wdyt?

sdk/testing/src/test/java/io/opentelemetry/sdk/testing/context/StrictContextStorageTest.java
anuraaga: Yeah that makes sense, removed the noop handling

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/context/StrictContextStorage.java
```diff
@@ -0,0 +1,193 @@
...
+ * <pre>{@code
+ * > class MyInstrumentationTest {
+ * >   private static ContextStorage previousStorage;
+ * >   private static StrictContextStorage strictStorage;
+ * >
+ * >   @BeforeAll
+ * >   static void setUp() {
+ * >     previousStorage = ContextStorage.get()
+ * >     strictStorage = StrictContextStorage.create(previousStorage);
+ * >     ContextStorage.set(previousStorage);
```
iNikem, line 49:
```suggestion
 * >     ContextStorage.set(strictStorage);
```

iNikem (commented):
Should this live here or in instrumentation repo?

context/src/main/java/io/opentelemetry/context/LazyStorage.java
jkwatson: Agreed for now. I think it could be a confusing addition to the public API. I'm fine with this for now...especially as we can start trying it out and seeing how well it works.

jkwatson (approved):
I don't love the string-based coupling, but let's give this a try and improve as we go.

PR #1940, opened 2020-10-30
Allow subclassing SpanContext by other packages

tylerbenson:
Having abstract methods with default visibility makes the class not able to be subclassed by other packages.
Changing to protected to maintain the intended restrictions from #1594, but allowing it to be subclassed.


jkwatson:
Why do you need to create a non-AutoValue subclass of SpanContext? I think this is probably not a good idea, since we do not want the SpanContext to end up mutable. If we allow subclassing, we cannot guarantee that.

Is there another way to solve your use-case that doesn't open up the possibility of breaking things?

Oberon00:
The spec says that the SpanContext functionality SHOULD NOT be overridable, so I don't think we should do that.

tylerbenson:
Closing in favor of #1935.

PR #1941, opened 2020-10-30
Prevent SDK from throwing on null instrumentation name

### LINKED ISSUE #1879: OpenTelemetry.getTracer(...) differs from API spec
**Describe the bug**
The `OpenTelemetry.getTracer(...)` [family of methods](https://www.javadoc.io/static/io.opentelemetry/opentelemetry-api/0.9.1/io/opentelemetry/OpenTelemetry.html#getTracer-java.lang.String-) claim to reject a `null` library name. [The spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#get-a-tracer) for obtaining a tracer indicates that invalid names such as `null` and empty string should lead to a default value.

This behavior (returning a default) is found when only the API is on the classpath, but the SDK rejects `null` and throws an exception.

**Steps to reproduce**
```java
OpenTelemetry.getGlobalTracer(null);
OpenTelemetrySdk.get().getTracer(null);
```

I don't know what the right answer is, but it's odd to me that the API silently allows `null` (but documents it as invalid), while the SDK throws on `null`, and the spec suggests `null` should be silently handled.

Artifacts: `opentelemetry-api`, `opentelemetry-sdk`
Version: `v0.9.1`, `0.10.0-SNAPSHOT`

MariusVolkhart:
Fixes #1879 

The first commit has the fix. The other 2 commits are more of a code cleanup based on changes made in the first commit, but are not strictly necessary to fix the issue.

### DIFF
--- a/api/src/main/java/io/opentelemetry/api/metrics/MeterProvider.java
+++ b/api/src/main/java/io/opentelemetry/api/metrics/MeterProvider.java
@@ -6,7 +6,6 @@
 package io.opentelemetry.api.metrics;
 
 import io.opentelemetry.api.OpenTelemetry;
-import javax.annotation.Nullable;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
@@ -44,5 +43,5 @@ public interface MeterProvider {
    * @param instrumentationVersion The version of the instrumentation library.
    * @return a tracer instance.
    */
-  Meter get(String instrumentationName, @Nullable String instrumentationVersion);
+  Meter get(String instrumentationName, String instrumentationVersion);
 }
--- a/api/src/main/java/io/opentelemetry/api/metrics/MeterProvider.java
+++ b/api/src/main/java/io/opentelemetry/api/metrics/MeterProvider.java
@@ -6,6 +6,7 @@
 package io.opentelemetry.api.metrics;
 
 import io.opentelemetry.api.OpenTelemetry;
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
@@ -43,5 +44,5 @@ public interface MeterProvider {
    * @param instrumentationVersion The version of the instrumentation library.
    * @return a tracer instance.
    */
-  Meter get(String instrumentationName, String instrumentationVersion);
+  Meter get(String instrumentationName, @Nullable String instrumentationVersion);
 }
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/internal/ComponentRegistry.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/internal/ComponentRegistry.java
@@ -23,18 +23,6 @@ public abstract class ComponentRegistry<V> {
   private final ConcurrentMap<InstrumentationLibraryInfo, V> registry = new ConcurrentHashMap<>();
 
   /**
-   * Returns the registered value associated with this name and {@code null} version if any,
-   * otherwise creates a new instance and associates it with the given name and {@code null}
-   * version.
-   *
-   * @param instrumentationName the name of the instrumentation library.
-   * @return the registered value associated with this name and {@code null} version.
-   */
-  public final V get(String instrumentationName) {
-    return get(instrumentationName, null);
-  }
-
-  /**
    * Returns the registered value associated with this name and version if any, otherwise creates a
    * new instance and associates it with the given name and version.
    *
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/internal/ComponentRegistry.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/internal/ComponentRegistry.java
@@ -23,6 +23,18 @@ public abstract class ComponentRegistry<V> {
   private final ConcurrentMap<InstrumentationLibraryInfo, V> registry = new ConcurrentHashMap<>();
 
   /**
+   * Returns the registered value associated with this name and {@code null} version if any,
+   * otherwise creates a new instance and associates it with the given name and {@code null}
+   * version.
+   *
+   * @param instrumentationName the name of the instrumentation library.
+   * @return the registered value associated with this name and {@code null} version.
+   */
+  public final V get(String instrumentationName) {
+    return get(instrumentationName, null);
+  }
+
+  /**
    * Returns the registered value associated with this name and version if any, otherwise creates a
    * new instance and associates it with the given name and version.
    *
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/MeterSdkProvider.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/MeterSdkProvider.java
@@ -19,7 +19,9 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import java.util.logging.Logger;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
 /**
  * {@code Meter} provider implementation for {@link MeterProvider}.
@@ -29,6 +31,8 @@ import javax.annotation.Nonnull;
  */
 public final class MeterSdkProvider implements MeterProvider {
 
+  private static final Logger logger = Logger.getLogger(MeterSdkProvider.class.getName());
+  private static final String defaultMeterName = "unknown";
   private final MeterSdkComponentRegistry registry;
   private final MetricProducer metricProducer;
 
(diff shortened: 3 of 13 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/MeterSdkProvider.java
```diff
@@ -29,6 +31,8 @@
  */
 public final class MeterSdkProvider implements MeterProvider {
 
+  private static final Logger logger = Logger.getLogger(MeterSdkProvider.class.getName());
+  private static final String defaultMeterName = "unknown";
```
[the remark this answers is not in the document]
bogdandrutu: We follow the Google's style guide, so we should enforce not prefer :)

sdk/common/src/main/java/io/opentelemetry/sdk/internal/ComponentRegistry.java
```diff
@@ -22,18 +22,6 @@
...
 
-  /**
-   * Returns the registered value associated with this name and {@code null} version if any,
-   * otherwise creates a new instance and associates it with the given name and {@code null}
-   * version.
-   *
-   * @param instrumentationName the name of the instrumentation library.
-   * @return the registered value associated with this name and {@code null} version.
-   */
-  public final V get(String instrumentationName) {
```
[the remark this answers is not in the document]
MariusVolkhart: The method was no longer used by code within this repository after I changed how `MeterSdkProvider` and `TracerSdkProvider` worked. Since it is in the `internal` package, I figured it wasn't intended for public consumption, and rather than having the unused method in the repo, I removed it.

If I should leave it, please let me know and I'll happily undo the change to remove it :)

MariusVolkhart:
@jkwatson Thanks for the review! I rebased, and made the changes you requested. The last 2 commits are new, the first 3 are unchanged.

sdk/common/src/main/java/io/opentelemetry/sdk/internal/ComponentRegistry.java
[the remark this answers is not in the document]
carlosalberto: @MariusVolkhart I suggest undoing this change and removing it in a follow-up PR ;)

carlosalberto (approved):

sdk/common/src/main/java/io/opentelemetry/sdk/internal/ComponentRegistry.java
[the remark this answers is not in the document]
MariusVolkhart: I've reverted my changes, so the `ComponentRegistry` is untouched. Thanks @carlosalberto

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/MeterSdkProvider.java
```diff
@@ -41,11 +45,16 @@ private MeterSdkProvider(Clock clock, Resource resource) {
...
+    return get(instrumentationName, null);
   }
 
   @Override
-  public MeterSdk get(String instrumentationName, String instrumentationVersion) {
+  public MeterSdk get(String instrumentationName, @Nullable String instrumentationVersion) {
+    // Per the spec, both null and empty are "invalid" and a "default" should be used.
+    if (instrumentationName == null || instrumentationName.isEmpty()) {
+      LOGGER.fine("Meter requested without instrumentation name.");
+      instrumentationName = DEFAULT_METER_NAME;
```
bogdandrutu, line 56: I think this will be a common case, I would suggest to actually resolve this as a static variable and return the instance.

MariusVolkhart: I'm not sure I understand. Are you suggesting a static `MeterSdk` to represent the default, rather than loading the cached instance from the registry? Are we OK with ignoring the potentially valid version number provided when returning the static instance?

I don't think we should use a static here. Firstly, we have the cache, and secondly, optimizing the code for what is supposed to be a "don't do this" use case seems wrong. If this will be a common case, then it feels like there should be [a stronger push on the spec](https://github.com/open-telemetry/opentelemetry-specification/issues/586) to make `null` a valid value.

bogdandrutu: Just have a "private static Meter DEFAULT_METER = registry.get(DEFAULT_NAME, null)" and return that if no name instead of always asking the registry to return the value.

bogdandrutu: I understand that this is a no in the specs, but I do believe users will do this a lot :)). We can do that later

bogdandrutu: If we don't throw exception, user will use this most of the time, that is my theory

Oberon00: Should we keep throwing an exception then? The instrumentation name should not be a dynamically generated string that can accidentally be null at runtime anyway.

MariusVolkhart: I drafted #2036 which goes the route of not only throwing the exception in the SDK, but also in the API. By doing so, 3rd-party library authors who consume just the API are less likely to create a library that works until the application developer drops in the SDK.

jkwatson: Unfortunately, we explicitly can't do what you put into #2036. We cannot break applications that have poorly written instrumentation in them.

Oberon00: OK. But can we log a SEVERE at least? The logspam should be enough pressure to fix an empty instrumentation name hopefully.

jkwatson: We definitely should not spam an end-application's logs, as this can crash the app (run out of disk space, etc). And, if the end-user's application is using a library with bad instrumentation built in to it, they may not even have the option of fixing it (and they may not even be using OpenTelemetry at all!)

MariusVolkhart: My original motivation for opening this issue and submitting a PR was that I was instrumenting an internal library that another team consumes. It was unclear to me as a library author what value I was supposed to provide as the `instrumentationName`, so I tried `null`. The API accepted it without crashing! It wasn't until later, when I did some testing and added the SDK, that it crashed. That would have been the consuming team who publish the app.

As the author of a library, I _wish_ my code had failed when just the API artifact was on the classpath, or at least complained loudly about it to me. But, since the spec declares that `null` should be treated as "use a default", I made this PR to modify the SDK to be lenient (thereby matching the API).

___
I think this PR meets the requirements of the current spec. More importantly, it makes the SDK and API artifacts behave consistently on invalid inputs, so there are no surprises when the SDK is used with instrumented code. It does _not_ optimize for the case that the spec says is invalid, and it doesn't help developers adding instrumentation to "do it right". 

I'll happily try to improve documentation or samples to give both library developers and application developers a better idea of what a meaningful `instrumentationName` is for their use case, and why it is important. My "using Otel for the first time" experience is still relatively fresh in my memory.

MariusVolkhart:
Rebased to resolve conflicts.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/MeterSdkProvider.java
jkwatson: I'm 100% ok with the SDK not blowing up when a null instrumentation name is used. But, we should not change the API parameter annotations to make it something that is "approved", nor should we make the API throw exceptions if a null name is passed in.

api/src/main/java/io/opentelemetry/api/metrics/MeterProvider.java
```diff
@@ -43,5 +44,5 @@ static MeterProvider getDefault() {
    * @param instrumentationVersion The version of the instrumentation library.
    * @return a tracer instance.
    */
-  Meter get(String instrumentationName, String instrumentationVersion);
+  Meter get(String instrumentationName, @Nullable String instrumentationVersion);
```
anuraaga, line 47: @MariusVolkhart Thanks for this change - the implementation looks good, but I agree that we don't need to add the `@Nullable` annotation to parameters that, we defensively check, but still expect to not be `null`. Do you think you can remove the annotations?

MariusVolkhart: Done! :)

jkwatson (approved):
I think this is a fine step forward. Thanks!

MariusVolkhart:
Rebased onto latest

PR #1945, opened 2020-10-30
Inline the TracingContextUtils methods into the Span

jkwatson:
This also moves the Context-key into a new private SpanContextKey class, and does the same with the Baggage key.

This brings the Context interaction for Span & Baggage into parity.


anuraaga (approved):
Yay!

bogdandrutu (approved):

PR #1956, opened 2020-11-01
Fix api link in README.md

dengliming:
Now API packages already moved to `.api.package`


anuraaga (approved):
Regex didn't catch this - thanks!

bogdandrutu (approved):

PR #1958, opened 2020-11-01
Move builders to top level.

anuraaga:
As discussed before and at sigs, it could be more user-friendly to make sure our class names are named with top-level builders to reduce the chance of introducing underspecified imports into user code where they don't have tooling to prevent that. I started with API artifact, and non-metrics packages and can apply elsewhere if this is accepted.


bogdandrutu:
Consider to split the TraceStateEntry in a separate PR, and everything that is not builder. I think for that we agree.

anuraaga:
@bogdandrutu Sure

bogdandrutu:
My understanding is that we agreed on everything besides builders, i cannot see a PR for that. That should be trivial to merge

anuraaga:
For API, #1979 deals with the only non-builder

jkwatson (approved):
I think this is a good change, but I wonder if we should save it for 0.11.0, since it still seems to be somewhat controversial.

anuraaga:
0.10.0 is out so thinking of merging this in soon, last call for comments.

carlosalberto:
To me this sounds more like a personal style change than an improvement, but I'm fine with it otherwise.

I'd suggest waiting for @bogdandrutu for a little bit before merging, as he raised initial concern.

trask (approved):
fwiw, I've come around to thinking the pros of XBuilder (simplifying imports) outweigh the pros of X.Builder (nesting as namespacing / following Effective Java and Guava style)

anuraaga:
Been a few days - going to go ahead and merge this in. Thanks all

PR #1961, opened 2020-11-01
WIP: Adds listener method to register callback on context change

HaloFour:
Prototype API for registering a callback which is triggered when the current Context for the current thread is changing.

Exposes the current `ContextStorage` from the method `Context#storage`.

Adds a new method `ContextStorage#onAttach(Consumer<Context>)` to register a callback interface which is invoked when the current context is going to be changed.

I've not touched Javadocs or tests yet, wanted to first get feedback on the API surface.

Example of usage:

```java
ContextStorage.get().onAttach(toAttach -> {

    Span currentSpan = Span.fromContextOrNull(toAttach);
    if (currentSpan != null) {
        SpanContext spanContext = currentSpan.getSpanContext();
        MDC.put("trace_id", spanContext.getTraceIdAsHexString());
        MDC.put("span_id", spanContext.getSpanIdAsHexString());
    } else {
        MDC.remove("trace_id");
        MDC.remove("span_id");
    }
});
```

### DIFF
--- a/context/src/main/java/io/opentelemetry/context/ContextStorage.java
+++ b/context/src/main/java/io/opentelemetry/context/ContextStorage.java
@@ -22,6 +22,8 @@
 
 package io.opentelemetry.context;
 
+import java.util.function.Consumer;
+
 /**
  * The storage for storing and retrieving the current {@link Context}.
  *
@@ -81,4 +83,6 @@ public interface ContextStorage {
    * this will be the {@linkplain Context#root()} root context}.
    */
   Context current();
+
+  void onAttach(Consumer<Context> contextConsumer);
 }
--- a/context/src/main/java/io/opentelemetry/context/ThreadLocalContextStorage.java
+++ b/context/src/main/java/io/opentelemetry/context/ThreadLocalContextStorage.java
@@ -15,13 +15,10 @@ enum ThreadLocalContextStorage implements ContextStorage {
 
   private static final Logger logger = Logger.getLogger(ThreadLocalContextStorage.class.getName());
 
-  private static final ThreadLocal<Context> THREAD_LOCAL_STORAGE = new ThreadLocal<>();
+  private static final ThreadLocal<Context> THREAD_LOCAL_STORAGE = ThreadLocal.withInitial(
+      Context::root);
 
   private static final AtomicReference<Consumer<Context>> onAttachConsumer = new AtomicReference<>();
-
-  static {
-    THREAD_LOCAL_STORAGE.set(Context.root());
-  }
 
   @Override
   public Scope attach(Context toAttach) {
--- a/context/src/main/java/io/opentelemetry/context/ThreadLocalContextStorage.java
+++ b/context/src/main/java/io/opentelemetry/context/ThreadLocalContextStorage.java
@@ -46,13 +46,6 @@ enum ThreadLocalContextStorage implements ContextStorage {
     };
   }
 
-  private void onAttach(Context toAttach) {
-    Consumer<Context> consumer = onAttachConsumer.get();
-    if (consumer != null) {
-      consumer.accept(toAttach);
-    }
-  }
-
   @Override
   public Context current() {
     return THREAD_LOCAL_STORAGE.get();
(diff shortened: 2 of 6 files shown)


context/src/main/java/io/opentelemetry/context/ThreadLocalContextStorage.java
```diff
@@ -15,9 +17,8 @@
 
   private static final ThreadLocal<Context> THREAD_LOCAL_STORAGE = new ThreadLocal<>();
 
-  static {
-    THREAD_LOCAL_STORAGE.set(Context.root());
```
HaloFour, line 19: I removed this since it only sets the value on whichever thread happens to run the static constructor.  Every other thread would remain uninitialized and will return `null`.  Overriding `initialValue()` or calling the static `withInitialValue` method can provide a default value but the latter seemingly doesn't exist in Android and it's not really necessary since read access to the `ThreadLocal` is always through `current()` anyway.

context/src/main/java/io/opentelemetry/context/ContextStorage.java
```diff
@@ -81,4 +83,6 @@ static ContextStorage get() {
    * this will be the {@linkplain Context#root()} root context}.
    */
   Context current();
+
+  void onAttach(Consumer<Context> contextConsumer);
```
anuraaga, line 87: I think I would take a approach that allows registering wrappers - as long as they are registered early i the app before storage is initialized, it allows customizing without thread sync issues.

```java

public interface ContextStorage {
  default void wrap(Function<ContextStorage, ContextStorage> wrapper) {
    ContextStorageWrappers.add(wrapper);
  }
}

LazyStorage {
  ContextStorage contextStorage = createStorage();
  for (Function wrapper : ContextStorageWrappers.get()) {
    contextStorage = wrapper.apply(contextStorage);
  }
}
```

What do you think?

HaloFour: Registering decorators for the `ContextStorage` would be an interesting approach.  How would you get to the `ContextStorage` to call `wrap` without triggering `LazyStorage` to create the storage, though?

anuraaga: `LazyStorage` will only be loaded when `get` is called so I think it should be ok :-D

HaloFour: Got it, that makes sense now.  My only concern with this approach is that you only have that window of time during which you can register these decorators.  Perhaps `ContextStorage#wrap` should warn or throw if the storage has already been created?  Should there also be an SPI for ContextStorageWrappers?

HaloFour: Maybe it's sufficient to just make it easy to wrap `ThreadLocalContextStorage` via SPI.

The Javadocs on `ContextStorage` imply that you can access the instance of `ThreadLocalContextStorage` via `Context#threadLocalStorage` but that method doesn't exist and I don't see another way to access that instance.  Without the ability to reference that instance rolling your own ContextStorage would involve rewriting thread-local storage which is where I think the opportunities for error mostly lie.

HaloFour: I've opened #1972 as an alternative to this PR.  All it does is expose the `ThreadLocalContextStorage` instance to make it easier to roll your own basic `ContextStorage` implementation.

anuraaga: Exposing ThreadLocalContextStorage was definitely required (it was javadoc'd, just the method got lost during some refactoring 😅 ).

I think `wrap` can help where SPI is too tedious, or to allow wrapping the non-threadlocal storage, but if our current code satisfies your use case well, we can consider that later. I think we could add it now though if it helps at all (it was the very first ask from users when we implemented storage SPI in Armeria so I do expect it to happen)

HaloFour: The SPI approach scratches my itch for now, and now that ThreadLocalContextStorage is exposed again it makes my implementation that much simpler.   As for establishing an API here I think it worthwhile.  I'd be curious what the other use cases might be.

anuraaga: Here's some reference of what led to the API in Armeria

https://github.com/line/armeria/issues/2634

In particular it seems, from my interpretation, to be about having a similar API for hooking as reactor for users of both.

anuraaga: @HaloFour FYI sent out https://github.com/open-telemetry/opentelemetry-java/pull/2044

HaloFour: Looks great, thanks!

HaloFour:
Closing in favor of #2044

PR #1962, opened 2020-11-01
Add github action to build examples when changes are identified in example directory

bogdandrutu:
Some fixes:
1. Update grpc version to match the core version.
2. Fix build after grpc examples were moved to `.grpc` package, but proto package was not updated.

### DIFF
--- a/.github/workflows/pr-examples-build.yml
+++ b/.github/workflows/pr-examples-build.yml
@@ -1,0 +1,42 @@
+name: "PR Examples Build"
+
+on:
+  pull_request:
+    branches:
+      - master
+    paths:
+      - 'examples/**'
+jobs:
+  build:
+    name: Build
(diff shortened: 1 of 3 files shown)


.github/workflows/pr-examples-build.yml
```diff
@@ -0,0 +1,42 @@
...
+    name: Build
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os:
+          - macos-latest
+          - ubuntu-latest
+        include:
+          - os: ubuntu-latest
+            testAdditionalJavaVersions: true
```
anuraaga, line 20: We don't currently have this set up for examples, but maybe we should doesn't hurt to leave it

anuraaga (approved):
Thanks!

Issue #1960, opened 2020-11-01
Remove Baggage Entry?


anuraaga:
It seems like we can remove `Entry` from baggage package if we change `getEntries` to `forEach` pattern like `Attributes` which would also make them a bit more consistent. However if doing that, it makes me wonder again if `EntryMetadata` is really pulling weight it becomes quite lonely.

jkwatson:
How do you forEach with the metadata, then?

anuraaga:
Trying with a tri consumer, does it work?

https://github.com/open-telemetry/opentelemetry-java/pull/1982/files#diff-3dc5218e425f937361a39f3d3d2992cae6ab821451fcf15e54f8fee691278540R12

jkwatson:
> Trying with a tri consumer, does it work?
> 
> https://github.com/open-telemetry/opentelemetry-java/pull/1982/files#diff-3dc5218e425f937361a39f3d3d2992cae6ab821451fcf15e54f8fee691278540R12

I guess that works ok. A little weird given the current state of metadata, but I guess it's an ok way to do it.

anuraaga:
@jkwatson Agree with the current state of metadata being weird. While making the PR, I sort of felt like we could potentially punt on metadata until there are more use cases. e.g., it's not even clear if a String is appropriate since it's supposed to be split on semicolon.

jkwatson:
Mostly, we need to make sure we propagate it. I don't know that being able to access it in-process is all that important right now. But, propagation is pretty important, IMO.

PR #1964, opened 2020-11-02
Don't throw exceptions for invalid baggage like other APIs

anuraaga:
I noticed that our baggage APIs validate parameters with exceptions, which isn't what we do in other APIs. This makes `Entry` creation package private (#1960 is a bit related) and ignores bad input in baggage construction.


jkwatson (approved):
Thanks!

api/src/main/java/io/opentelemetry/api/baggage/ImmutableBaggage.java
```diff
@@ -135,6 +136,9 @@ public String toString() {
 
     @Override
     public Baggage.Builder put(String key, String value, EntryMetadata entryMetadata) {
+      if (!isKeyValid(key) || !isValueValid(value)) {
```
carlosalberto, line 139: Maybe we should do a fine-level logging here (can be discussed in/as a follow-up).

carlosalberto (approved):

api/src/main/java/io/opentelemetry/api/baggage/EntryMetadata.java
```diff
@@ -26,6 +26,9 @@
    * @return an {@code EntryMetadata}.
    */
   public static EntryMetadata create(String metadata) {
```
bogdandrutu, line 28: Mark metadata as `Nullable`

api/src/main/java/io/opentelemetry/api/baggage/ImmutableBaggage.java
bogdandrutu: why not throwing exception?

api/src/main/java/io/opentelemetry/api/baggage/ImmutableBaggage.java
```diff
@@ -146,14 +150,14 @@ public String toString() {
...
-      return this;
+      return put(key, value, EntryMetadata.EMPTY);
     }
 
     @Override
     public Baggage.Builder remove(String key) {
+      if (key == null) {
+        return this;
+      }
       entries.remove(Objects.requireNonNull(key, "key"));
```
bogdandrutu, line 161: This is wrong, either not allow and check for null or throw exception.

api/src/main/java/io/opentelemetry/api/baggage/ImmutableBaggage.java
```diff
@@ -135,6 +136,9 @@ public String toString() {
 
     @Override
     public Baggage.Builder put(String key, String value, EntryMetadata entryMetadata) {
+      if (!isKeyValid(key) || !isValueValid(value)) {
+        return this;
+      }
       entries.put(
           Objects.requireNonNull(key, "key"),
```
bogdandrutu, line 143: Either throw exception or allow `Nullable`

bogdandrutu (changes requested):

api/src/main/java/io/opentelemetry/api/baggage/EntryMetadata.java
anuraaga: We don't use Nullable on APIs which shouldn't be passed null but we have defensive behavior anyways. IIRC this is the answer to all your comments?

api/src/main/java/io/opentelemetry/api/baggage/ImmutableBaggage.java
anuraaga: We currently don't log for parameter checks in most places I believe. Maybe we should, but I think we should do it consistently separate from removing the exceptions.

@bogdandrutu To throw exceptions, I believe we'd need a spec change https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/error-handling.md#basic-error-handling-principles

api/src/main/java/io/opentelemetry/api/baggage/ImmutableBaggage.java
anuraaga: Oops, removed the redundant checks

jkwatson:
@anuraaga this needs a rebase. thanks!

anuraaga:
@jkwatson Sorry for delay, finally rebased

Issue #1946, opened 2020-10-31
Only expose interfaces in API


anuraaga:
We have a use case for completely reimplementing the API - auto instrumentation which does so too provide safety to an app in the face of version skew. Migrating span context in #1935 is helpful, but there are still classes like attributes which can only be bridged with copies. How about we go ahead and expose only interfaces to support this sort of use case better? It does allow a user to implement an interface themselves, but if this actually solves a use case for them, I don't see why we need to proactively block this, and it will be rare.

bogdandrutu:
From one of the PR that generated this discussion:

@jkwatson:
> I've heard this comment a lot, but I've never heard of an actual example of this happening in the real world. Can you show a case where you've been actually impacted by this in the past? Java 8 default methods let us evolve interfaces in a similar way that we've been able to evolve abstract classes, pre Java-8.

@bogdandrutu:

FYI: This is not about evolving the API, but about dealing with third-party implementation of the interface. I am advocating to not allow other implementations for classes that we don't need to offer that capability.

Yes, we had the problem in OpenCensus. We had an interface that represented the Tags (a.k.a baggage) that did not expose publicly a get method, and have our own implementation that exposed (package protected) a get method. Now because that was an interface and everyone can extend the interface it was not trivial to convert to the internal implementation because we had a path where we have to deal with another implementation that does not have our functionality.

You may argue that all functionality must be public, but that is not necessary true if you don't want user to have access to that or abuse your API. By restricting who can implement an interface/abstract class you can rely on package protected functionality to optimize performance etc.

So I am fine with using final classes or auto-value classes with package protected ctor, because it means we control the implementation.

bogdandrutu:
Also to use @jkwatson expression :), I've heard this comment a lot that instrumentation needs to re-implement all the classes, but a clear documentation and design document (why is this really required) was never documented in the repository. I saw some documents in google docs (please move that to the repo so others can see and propose better alternatives if possible) which did not convince me that this is a real requirement (mostly like I am missing stuff).

So I suggest that initially we prove that the hypothesis is clear and understood, before taking actions.

jkwatson:
My opinion on this:  

If what you are exposing is pure data, then having a final class is fine. If you're exposing functionality, I will always prefer to have an interface for that functionality.

For example, Baggage is pure, immutable data. Having a final class for that is great.  A Span, however, is a combination of data and function, so we should use an interface for it (which we do!).

bogdandrutu:
[...]

PR #1965, opened 2020-11-02
Add point to contributor document about preserving semantic versioning.

anuraaga:
As came up in #1946

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,6 +45,9 @@ which uses [google-java-format](https://github.com/google/google-java-format) li
 
 ### Best practices that we follow
 
+* This project uses semantic versioning. Except for major versions, a user should be able to update
+their dependency version on this project and have nothing break. This means we do not make breaking
+changes to the API (e.g., remove a public method) or to the ABI (e.g., change return type from void to non-void).
 * Avoid exposing publicly any class/method/variable that don't need to be public.
 * By default, all arguments/members are treated as non-null. Every argument/member that can be `null` must be annotated with `@Nullable`.
 * The project aims to provide a consistent experience across all the public APIs. It is important to ensure consistency (same look and feel) across different public packages.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,7 +45,7 @@ which uses [google-java-format](https://github.com/google/google-java-format) li
 
 ### Best practices that we follow
 
-* This project uses semantic versioning. Except for major versions, a user should be able to update
+* This project uses [semantic versioning](https://semver.org/). Except for major versions, a user should be able to update
 their dependency version on this project and have nothing break. This means we do not make breaking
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -45,6 +45,9 @@ which uses [google-java-format](https://github.com/google/google-java-format) li
 
 ### Best practices that we follow
 
+* This project uses [semantic versioning](https://semver.org/). Except for major versions, a user should be able to update
+their dependency version on this project and have nothing break. This means we do not make breaking
+changes to the API (e.g., remove a public method) or to the ABI (e.g., change return type from void to non-void).
```
bogdandrutu, line 50: Is the example for ABI really breaking? Because if that is breaking I think adding a new method is also breaking correct?

Also changing an argument from String -> CharSequence I thought it is fine, I know it breaks the sdk implementation but does it break an artifact that depends on small API version + new API + new SDK?

Oberon00: So what you are saying is that we make a special exception for ABI compatibility to allow changes that would break SDK implementations but from which we hope that they do not break typical uses of the API package for instrumentation? We need to document that then.

iNikem: Adding new method is not breaking. Changing return type is breaking because `invokevirtual/invokeinterface` bytecoe instruction contain method descriptor to call. That descriptor contains return value. When method with such return value is not found when called, the `AbstractMethodError` is thrown.

Oberon00: Note that *technically* implementing the API is just a special case of "using the API".

bogdandrutu: I found a very good and detailed explication of the ABI compatibility https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html

Indeed changing the return type, or the type of an argument even if it is source compatible as in my example with the String is considered backwards incompatible.

bogdandrutu: @Oberon00 have not read the entire document, but at first it seems that adding a new method to an interface (even without default) is considered backwards compatible, but that may break SDK implementation. I would not aim to fix that problem personally.

bogdandrutu: Implementing the API is a special case, but that special case is not supposed to be included in other libraries so does not require ABI compatibility if I understand correctly

anuraaga: Enforcing version coupling for API + SDK, including custom SDKs is ok I think though especially for the case of custom SDKs we need to declare that clearly (hopefully separate from this PR).

Still, ideally we can avoid it, and default interface methods help with that.

jkwatson:
This is a good addition, but I'm a little hesitant to merge this without a caveat around pre-1.0 releases. This will absolutely be true after 1.0, but before then things are more muddy.

jkwatson (approved):
Thanks!

PR #1966, opened 2020-11-02
Use fine logging in w3c/b3 propagator instead of info.

### LINKED ISSUE #1963: Use log level FINE (or lower) for header extraction errors (which are spammable by bad actor)
I don't know if there's a good standard here(?)

I like to use INFO for startup/shutdown messages, and run at INFO (nice to always see startup/shutdown messages), and pick either WARN+ or DEBUG- (FINE-) for all other messages.

anuraaga:
It looks like w3c was the only one logging at info

Fixes #1963

### DIFF
--- a/extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/Common.java
+++ b/extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/Common.java
@@ -44,7 +44,7 @@ final class Common {
           traceFlags,
           TraceState.getDefault());
     } catch (Exception e) {
-      logger.log(Level.INFO, "Error parsing header. Returning INVALID span context.", e);
+      logger.log(Level.FINE, "Error parsing header. Returning INVALID span context.", e);
       return SpanContext.getInvalid();
     }
   }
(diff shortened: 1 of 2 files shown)


trask:
I think also here https://github.com/open-telemetry/opentelemetry-java/blob/master/extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/Common.java#L47

anuraaga:
@trask Ah thanks for finding!

iNikem (approved):

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/Common.java
```diff
@@ -44,7 +44,7 @@ static SpanContext buildSpanContext(String traceId, String spanId, String sample
           traceFlags,
           TraceState.getDefault());
     } catch (Exception e) {
-      logger.log(Level.INFO, "Error parsing header. Returning INVALID span context.", e);
+      logger.log(Level.FINE, "Error parsing header. Returning INVALID span context.", e);
```
bogdandrutu, line 47: This is mostly inconsistent, but we should probably use logger.info everywhere instead of logger.log(info,...)

bogdandrutu (approved):

extensions/trace_propagators/src/main/java/io/opentelemetry/extensions/trace/propagation/Common.java
bogdandrutu: This is a bit, so no need to really change here, but maybe good to have consistent.

carlosalberto (approved):

PR #1969, opened 2020-11-02
Add unit to start, end and event timestamp accepting methods in Span.

### LINKED ISSUE #1957: Have methods that accept a timestamp except a value and TimeUnit
Currently we have some methods that accept a nanoseconds timestamp with documentation on the fact that it's nanos. If we accept `TimeUnit` instead, the API usage becomes self-documenting, both for writers and readers of end-user code, while also providing some flexibility when the time provided is not available as nanos.

An alternative is to rename the methods to e.g., `setStartTimestampNanos` to still be self-documenting, but without the added flexibility.

anuraaga:
...and accept Instant.

Fixes #1957 

Changelog:

- BREAKING CHANGE: Span.Builder.setStartTimestamp, Span.end, and Span.addEvent methods which accept a timestamp now accept a timestamp with a TimeUnit instead of requiring a nanos timestamp

- Versions of Span.Builder.setStartTimestamp, Span.end, and Span.addEvent added which accept Instant timestamps


Oberon00 (commented):
Nice! I think we should wait for  #1934 though and do `end` in the same PR, to avoid the temporary inconsistency.

api/src/main/java/io/opentelemetry/api/trace/PropagatedSpan.java
```diff
@@ -61,7 +62,7 @@ public Span addEvent(String name) {
   }
 
   @Override
-  public Span addEvent(String name, long timestamp) {
+  public Span addEvent(String name, long timestamp, TimeUnit unit) {
```
bogdandrutu, line 65: Why not using Java.time.Instant? We are java8 now.

bogdandrutu: I think I found myself the reason, is android version, the `java.time.Instant` was added in Android API level 26, and we try to be API level 24 compatible.

jkwatson (approved):
Great improvement in the API! Thanks!

api/src/main/java/io/opentelemetry/api/trace/PropagatedSpan.java
anuraaga: Adding overloads with Instant sounds good too just not for now.

anuraaga:
Updated end timestamp too

jkwatson (approved):
Thanks!

jkwatson:
From the SIG meeting: would be preferable to use `Instant` for this API. Can we hold off on merging this until we figure out to get `Instant` with the current android requirements?

anuraaga:
@jkwatson I'd generally recommend having versions that accept both `TimeUnit` and `Instant` since it's pretty common to have one or the other available and it allows accepting them both conveniently and without overhead.

jkwatson:
> @jkwatson I'd generally recommend having versions that accept both `TimeUnit` and `Instant` since it's pretty common to have one or the other available and it allows accepting them both conveniently and without overhead.

I agree. I think we're willing to take both of those changes in 0.11.0, even though they are both breaking.

anuraaga:
Updated to add the `Instant` overloads

PR #1972, opened 2020-11-02
Expose ThreadLocalContextStorage publicly

HaloFour:
Exposes the `ThreadLocalContextStorage` enum publicly to enable writing cimple `ContextStorage` implementations without having to rewrite the storage.  Updates the Javadocs for `ContextStorage` to replace the non-existent `Context#threadLocalStorage` method with `ThreadLocalContextStorage.INSTANCE`.

### DIFF
--- a/context/src/main/java/io/opentelemetry/context/ContextStorage.java
+++ b/context/src/main/java/io/opentelemetry/context/ContextStorage.java
@@ -34,7 +34,7 @@ package io.opentelemetry.context;
  * >
  * >   @Override
  * >   public ContextStorage get() {
- * >     ContextStorage threadLocalStorage = Context.threadLocalStorage();
+ * >     ContextStorage threadLocalStorage = ContextStorage.threadLocalStorage();
  * >     return new RequestContextStorage() {
  * >       @Override
  * >       public Scope T attach(Context toAttach) {
@@ -67,6 +67,15 @@ public interface ContextStorage {
    */
   static ContextStorage get() {
     return LazyStorage.get();
+  }
+
+  /**
+   * Returns the default {@link ContextStorage} used to attach {@link Context}s to scopes of
+   * execution. Should only be used when defining your own {@link ContextStorage} in case you want
+   * to delegate functionality to the default implementation.
+   */
+  static ContextStorage threadLocalStorage() {
+    return ThreadLocalContextStorage.INSTANCE;
   }
 
   /**
--- a/context/src/main/java/io/opentelemetry/context/ContextStorage.java
+++ b/context/src/main/java/io/opentelemetry/context/ContextStorage.java
@@ -34,7 +34,7 @@ package io.opentelemetry.context;
  * >
  * >   @Override
  * >   public ContextStorage get() {
- * >     ContextStorage threadLocalStorage = ContextStorage.threadLocalStorage();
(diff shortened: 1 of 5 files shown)


context/src/main/java/io/opentelemetry/context/ContextStorage.java
```diff
@@ -69,6 +69,15 @@ static ContextStorage get() {
     return LazyStorage.get();
   }
 
+  /**
+   * Returns the default {@link ContextStorage} used to attach {@link Context}s to scopes of
+   * execution. Should only be used when defining your own {@link ContextStorage} in case you want
+   * to delegate functionality to the default implementation.
+   */
+  static ContextStorage threadLocalStorage() {
```
bogdandrutu, line 77: This makes the method public. What is the reason to do it?

HaloFour: To make it easier to write a delegating `ContextStorage` implementation that uses thread-local storage.  This is already suggested by the Javadocs on this type which reference the method `Context#threadLocalStorage` which does not exist.

context/src/main/java/io/opentelemetry/context/ThreadLocalContextStorage.java
```diff
@@ -45,7 +41,8 @@ public Scope attach(Context toAttach) {
 
   @Override
   public Context current() {
-    return THREAD_LOCAL_STORAGE.get();
+    Context threadContext = THREAD_LOCAL_STORAGE.get();
+    return threadContext != null ? threadContext : Context.root();
```
bogdandrutu, line 45: Changes in this file seem independent of the main issue, consider to split them.

context/src/main/java/io/opentelemetry/context/ContextStorage.java
bogdandrutu: What about then we make the implementation thread local public enum? I feel that exposing this in the Context ties too much the threadlocal storage to the context.

context/src/main/java/io/opentelemetry/context/ThreadLocalContextStorage.java
HaloFour: Done and done, #1974

context/src/main/java/io/opentelemetry/context/ContextStorage.java
HaloFour: That'd be fine by me.  I was going with the existing convention suggested in the Javadocs on `ContextStorage` as well as by `LazyStorage` (which references a similar helper method on `DefaultContext`).  Should we get others to weigh in before I change this PR?

bogdandrutu: Yes, I would like to hear others opinions @anuraaga @jkwatson @carlosalberto

HaloFour:
I've changed the PR so that instead of exposing the `ThreadLocalContextStorage` via a static helper method on `ContextStorage` that the `ThreadLocalContextStorage` enum is exposed publicly and can be used directly.

anuraaga:
I think this PR already happened as a part of https://github.com/open-telemetry/opentelemetry-java/pull/1938 @HaloFour what do you think?

HaloFour:
Yes, that PR does address the same concern.  Thanks!

Issue #1527, opened 2020-08-10
Use microsecond precision on Java 9+


anuraaga:
A small amount of version detecting would let us have better precision on timestamps (I don't think I see this anywhere here yet), a couple of examples

https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/internal/Platform.java#L166
https://github.com/line/armeria/blob/0bb591d252884ec471af00e2757032ae57196d5e/core/src/main/java9/com/linecorp/armeria/internal/common/Java9VersionSpecific.java#L40

PR #1973, opened 2020-11-02
Use Instant to get the Clock now. This helps with precision where the OS offers that (java9 +)

bogdandrutu:

### DIFF
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/internal/InstantClock.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/internal/InstantClock.java
@@ -6,29 +6,31 @@
 package io.opentelemetry.sdk.internal;
 
 import io.opentelemetry.sdk.common.Clock;
+import java.time.Instant;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.concurrent.ThreadSafe;
 
 /** A {@link Clock} that uses {@link System#currentTimeMillis()} and {@link System#nanoTime()}. */
 @ThreadSafe
-public final class MillisClock implements Clock {
+public final class InstantClock implements Clock {
 
-  private static final MillisClock INSTANCE = new MillisClock();
+  private static final InstantClock INSTANCE = new InstantClock();
 
-  private MillisClock() {}
+  private InstantClock() {}
 
   /**
    * Returns a {@code MillisClock}.
    *
    * @return a {@code MillisClock}.
    */
-  public static MillisClock getInstance() {
+  public static InstantClock getInstance() {
     return INSTANCE;
   }
 
   @Override
   public long now() {
-    return TimeUnit.MILLISECONDS.toNanos(System.currentTimeMillis());
+    Instant now = Instant.now();
+    return TimeUnit.SECONDS.toNanos(now.getEpochSecond()) + now.getNano();
   }
 
   @Override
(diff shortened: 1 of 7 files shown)


carlosalberto (approved):

carlosalberto:
Niceee.

bogdandrutu:
I think we will have to:
1. Increase android API
2. OR load with reflection if method is available

sdk/common/src/main/java/io/opentelemetry/sdk/internal/InstantClock.java
```diff
@@ -6,29 +6,31 @@
...
 
 import io.opentelemetry.sdk.common.Clock;
+import java.time.Instant;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.concurrent.ThreadSafe;
 
 /** A {@link Clock} that uses {@link System#currentTimeMillis()} and {@link System#nanoTime()}. */
 @ThreadSafe
-public final class MillisClock implements Clock {
+public final class InstantClock implements Clock {
```
Oberon00, line 15: To avoid renaming this class in the future whenever the implementation changes (this is a public class!), I strongly suggest choosing a more neutral name like SystemClock or UnixTimeClock.

Oberon00 (changes requested):

jkwatson:
> I think we will have to:
> 
> 1. Increase android API
> 2. OR load with reflection if method is available

I'd vote #2 unless we can get some Android folks to comment and/or provide some guidance.

bogdandrutu:
@jkwatson not sure if I can make animalsniffer pass with the option 2 (I think it checks all usages, I will try more), another alternative is to have the SystemClock use SPI and provide the implementation externally.

jkwatson:
> @jkwatson not sure if I can make animalsniffer pass with the option 2 (I think it checks all usages, I will try more), another alternative is to have the SystemClock use SPI and provide the implementation externally.

Even without SPI, the end-user can currently manually build an SDK with their own Clock implementation at the moment. I don't know enough about how Android builds to work to know if this would be something that an Android dev could adjust for themselves already.

anuraaga:
Instant only has better precision on Java 9+ from what I understand so on Java 8 it will just be some extra overhead.

This and the Android issue may be solved in this case by using an mrjar?

anuraaga:
This will fix #1527

anuraaga:
Ah btw recent Android can desugar most (not all) Java 8 APIs like Instant on any Android API level. Don't think a new signature has been released that models this.

bogdandrutu:
@anuraaga link?

anuraaga:
https://developer.android.com/studio/write/java8-support#library-desugaring

https://developer.android.com/studio/write/java8-support-table

anuraaga:
By the way, here's an example of the mrjar approach - it's my recommendation to ensure optimal performance on Java 8

https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/internal/common/JavaVersionSpecific.java
https://github.com/line/armeria/blob/master/core/src/main/java9/com/linecorp/armeria/internal/common/Java9VersionSpecific.java
https://github.com/line/armeria/blob/master/core/build.gradle#L12

anuraaga:
After a lot of searching, I managed to find a signature! #2015

jkwatson:
I think with the animal-sniffer gummibears update in place, this will be viable, after a rebase.

anuraaga:
Filed #2154 to use mrjar approach

jkwatson:
> Filed #2154 to use mrjar approach

Should we close this, pending that?

PR #1974, opened 2020-11-02
Documents that ContextStorage.current can return null

HaloFour:
~~Currently the `ThreadLocalContextStorage` sets the `ThreadLocal` to `Context.root()` within the static constructor.  However, this only sets the value of the `ThreadLocal` for the thread that happens to execute that constructor.  For all other threads the `ThreadLocal` remains unset and `THREAD_LOCAL_STORAGE.get()` will return `null` which is likely unexpected by the consumer.~~

~~There are two options here:~~

1. ~~Have `THREAD_LOCAL_STORAGE` be a class derived from `ThreadLocal<Context>` and override the `initialValue` method to return an initial value for when the thread local has not been set for the current thread.~~
2. ~~Have `current()` check if `THREAD_LOCAL_STORAGE.get()` returns `null` and return `Context.root()` instead.~~

~~Since all reads to the `ThreadLocal` go through the `current()` method I went with the second option as I find it the easiest to understand.~~

Updates the docs and annotations on `ContextStorage#current` to make it clear that the method may return `null` if no context has been attached to the current thread.  Also removes the static constructor in `ThreadLocalContextStorage` that sets the `ThreadLocal<Context>` to `Context.root()` for the thread that invokes the constructor.

### DIFF
--- a/context/src/main/java/io/opentelemetry/context/ThreadLocalContextStorage.java
+++ b/context/src/main/java/io/opentelemetry/context/ThreadLocalContextStorage.java
@@ -15,10 +15,6 @@ enum ThreadLocalContextStorage implements ContextStorage {
 
   private static final ThreadLocal<Context> THREAD_LOCAL_STORAGE = new ThreadLocal<>();
 
-  static {
-    THREAD_LOCAL_STORAGE.set(Context.root());
-  }
-
   @Override
   public Scope attach(Context toAttach) {
     if (toAttach == null) {
@@ -45,7 +41,8 @@ enum ThreadLocalContextStorage implements ContextStorage {
 
   @Override
   public Context current() {
-    return THREAD_LOCAL_STORAGE.get();
+    Context threadContext = THREAD_LOCAL_STORAGE.get();
+    return threadContext != null ? threadContext : Context.root();
   }
 
   enum NoopScope implements Scope {
--- a/context/src/main/java/io/opentelemetry/context/ThreadLocalContextStorage.java
+++ b/context/src/main/java/io/opentelemetry/context/ThreadLocalContextStorage.java
@@ -7,6 +7,7 @@ package io.opentelemetry.context;
 
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import javax.annotation.Nullable;
 
 enum ThreadLocalContextStorage implements ContextStorage {
   INSTANCE;
@@ -40,9 +41,9 @@ enum ThreadLocalContextStorage implements ContextStorage {
   }
 
   @Override
+  @Nullable
   public Context current() {
-    Context threadContext = THREAD_LOCAL_STORAGE.get();
-    return threadContext != null ? threadContext : Context.root();
+    return THREAD_LOCAL_STORAGE.get();
   }
 
   enum NoopScope implements Scope {
--- a/context/src/main/java/io/opentelemetry/context/ContextStorage.java
+++ b/context/src/main/java/io/opentelemetry/context/ContextStorage.java
@@ -22,6 +22,8 @@
 
 package io.opentelemetry.context;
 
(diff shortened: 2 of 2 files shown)


context/src/main/java/io/opentelemetry/context/ThreadLocalContextStorage.java
```diff
@@ -45,7 +41,8 @@ public Scope attach(Context toAttach) {
 
   @Override
   public Context current() {
-    return THREAD_LOCAL_STORAGE.get();
+    Context threadContext = THREAD_LOCAL_STORAGE.get();
+    return threadContext != null ? threadContext : Context.root();
```
bogdandrutu, line 45: I think we can simply remove this line, we check for null in the Context and this is not necessarily.

bogdandrutu: And add to the javadoc this behavior.

HaloFour: Ok, if it's expected that `ContextStorage#current` can return `null` it feels like I should also add the `@Nullable` annotation so that developers can expect this.

HaloFour: Are you suggesting to change this Javadoc on `ContextStorage#current`?  It implies that this method will never return `null`:

```
  /**
   * Returns the current {@link DefaultContext}. If no {@link DefaultContext} has been attached yet,
   * this will be the {@linkplain Context#root()} root context}.
   */
```

Also feels weird that it calls out `DefaultContext` given that's a package-private class.

Oberon00: See here: https://github.com/HaloFour/opentelemetry-java/blob/1512fa1d754644b0c2de3f30a6039c848bfc1c03/context/src/main/java/io/opentelemetry/context/Context.java#L72

HaloFour: That's why I'm asking for clarity, because if it's expected that `ContextStorage#current` can return `null` then the Javadocs on that method should be updated to reflect this, and the method should be marked as `@Nullable`.

Oberon00: What is more

>  ThreadLocalContextStorage sets the ThreadLocal to Context.root() within the static constructor

this seems to be useless then and could be removed.

CC @anuraaga

HaloFour: I've committed the suggested changes, which more or less reverses what the PR was intended to do so I've changed the description of the PR (if that's ok).

bogdandrutu (approved):

iNikem:
Am I the only one who thinks that this PR was rushed through? 5h from creation to merge without giving @anuraaga a chance to comment on this? I think the author of the current context implementation should have a say in this.

@bogdandrutu @anuraaga @jkwatson

context/src/main/java/io/opentelemetry/context/ThreadLocalContextStorage.java
```diff
@@ -15,10 +16,6 @@
 
   private static final ThreadLocal<Context> THREAD_LOCAL_STORAGE = new ThreadLocal<>();
 
-  static {
-    THREAD_LOCAL_STORAGE.set(Context.root());
```
anuraaga, line 19: Thanks for fixing my silliness :)

anuraaga (commented):
Thanks @iNikem - would have appreciated a chance to look, but if there were two maintainer approvals that would be fine too. Does seem to have been quick.

Change looks great though, thanks a lot @HaloFour!

PR #1977, opened 2020-11-02
Add doc explaining the restriction on return type

trask:
per @bogdandrutu request https://github.com/open-telemetry/opentelemetry-java/pull/1888#issuecomment-720712989


bogdandrutu (approved):

anuraaga (commented):
We have a similar comment in three places, should we add to all of them?

anuraaga (approved):

Issue #1976, opened 2020-11-02
Prevent empty OTLP's Int/DoubleSum from being sent


carlosalberto:
Currently, if a given counter has no reported values in a given interval (that is, `MetricData.getPoints().isEmpty == true`), it will be sent anyway (was OTLP's `IntSum`/`DoubleSum` without any data point at all). 

Is this needed at all? Maybe we can simply ignore them in such a case? Or is there any reason why we keep on sending them, even if empty?

jkwatson:
This sounds like a bug, but I'm not sure what the collector expects in these cases. @bogdandrutu do you know?

jrcamp:
@carlosalberto I'm not aware of anything in the collector that would make use of metricdata if there are zero datapoints. However I was wondering why is it generating empty metricdata instead of having a datapoint with value of zero?

jkwatson:
> @carlosalberto I'm not aware of anything in the collector that would make use of metricdata if there are zero datapoints. However I was wondering why is it generating empty metricdata instead of having a datapoint with value of zero?

Currently our Sum aggregations are all cumulative, so a zero report would mean that the counter had been reset, which would not be correct.

jrcamp:
If it's a cumulative shouldn't a datapoint value always be reported each interval then? Even if the value is unchanged it should still be sent as datapoints can be dropped at the collector or further upstream.

keitwb:
Yes, it is generally more robust to report something on a regular interval, even if it hasn't changed.  It might not be as optimal but it will be much more robust for the backends that ultimately consume this data.

If there are any delta values where it reports only the difference since the last report, it would be good to emit 0s as well.

jmacd:
I see that we need to add some specification to clarify this matter. In the OTel-Go SDK the Processor component has two distinct behaviors:

1. It can keep state needed for DELTA to CUMULATIVE translation
2. it can keep state needed purely for memory

Both of these mechanisms are needed for a Prometheus exporter that doesn't otherwise manage its own state. When we discuss what the behavior should be for OTLP export, both (1) and (2) are valid behaviors. 

https://github.com/open-telemetry/opentelemetry-specification/issues/731 states that (1) should be on by default, but we haven't specified what to do about (2). I'm afraid this has to be tackled in the specification, but's tangential to the question here.

[...]

PR #1978, opened 2020-11-03
Do not export counters with no points for OTLP.

carlosalberto:
Potential fix for #1976 

It shows that nothing seems to break with this simple change.


anuraaga (commented):
LGTM - this seems like a reasonable interpretation of the proto to me

jkwatson:
My only worry is there might be a reason to send an empty payload, just as a heartbeat. Other than that, this looks fine to me.

carlosalberto:
@bogdandrutu Any take on this? 😉

carlosalberto:
Marking this as Ready to Review, in order to get more eyes ;)

anuraaga (approved):

jkwatson (approved):

jmacd:
I agree with this as a short-term fix that will make the Java OTLP Metrics export not send what to me is a meaningless point. It's uncovered that we should probably specify in the protocol that a Metric with empty point list may be regarded as equivalent to an absent Metric. If the specification says we can silently drop these Metrics with no data points, then we can assert that this is a correct change.

I would like to bring the Java Metric SDK into more conceptual alignment with the OTel-Go SDK. There we have a `basic` Processor component with two distinct kinds of support that are present in order to implement Prometheus and/or Cumulative metric export pipelines.

1. First, the Processor is the component that is responsible for converting DELTA inputs (from Counter instruments) into CUMULATIVE outputs (i.e., OTLP Metric Sums with AggregationTemporality=CUMULATIVE)
2. Second, the Processor has a "Memory" option that is responsible for maintaining state about every combination of metric instrument and label set that was ever encountered.

Both of these behaviors combine are needed for a Prometheus export pipeline. Only the first of these behaviors is needed for an OTLP/CUMULATIVE export pipeline. None of these behaviors is needed for an OTLP/DELTA export pipeline. Since I'm not familiar with the internals, I don't know how much restructuring is needed to accomplish this goal, but when it is accomplished, there will be a basic Processor component with two optional behaviors. I approve this PR with a TODO mentioning the final state and the need for more SDK specification.

Whether using a CUMULATIVE or DELTA strategy with OTLP, there is no reason to use empty data points. We should either not report a Metric ("no observation(s)") or we should report a Metric with more than zero points ("some observations").

There are independent questions raised here that should not block the resolution of the immediate issue. For example: if a Processor is configured for CUMULATIVE export and not MEMORY behavior, shall we omit those points because of a lack of observations? If the instrument is a SumObserver and the point that was previously observed is later not observed and then subsequently observed again, without changing the start timestamp: what should we make of this? I don't think it's too important what we say about this situation, but in any case we should not be sending Metric values with empty points, right?

@open-telemetry/specs-metrics-approvers  ^^^ some questions for the Metrics specification.

@jkwatson Re: "to send an empty payload, just as a heartbeat". I would advocate for using the memory behavior described here if you need this kind of information. See also https://github.com/open-telemetry/opentelemetry-specification/pull/1102.

jmacd (approved):

jmacd:
@bogdandrutu I'd like your opinion specifically. See also https://github.com/open-telemetry/opentelemetry-java/issues/1976#issuecomment-722585324

carlosalberto:
@bogdandrutu @jkwatson @jmacd I think we should merge this change, to keep things 'conservative'. Then we can iterate and make this selectable (in order to support Prometheus, for example). Let me know what you think.

bogdandrutu (approved):
As @jmacd pointed this is a good short term fix.

PR #1979, opened 2020-11-03
Expose foreach consumer instead of entries for trace state.

anuraaga:
I noticed instead of exposing `Entry`, we can follow the same pattern as we do in `Attributes`. We could remove Entry completely in the future if wanted (presumably users will prefer us to implement `Map` in the future anyways)

### DIFF
--- a/api/src/main/java/io/opentelemetry/api/trace/TraceState.java
+++ b/api/src/main/java/io/opentelemetry/api/trace/TraceState.java
@@ -11,6 +11,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import java.util.function.BiConsumer;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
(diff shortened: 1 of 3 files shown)


api/src/main/java/io/opentelemetry/api/trace/TraceState.java
```diff
@@ -170,15 +188,16 @@ public TraceState build() {
   /** Immutable key-value pair for {@code TraceState}. */
   @Immutable
   @AutoValue
-  public abstract static class Entry {
+  abstract static class Entry {
```
bogdandrutu, line 191: You can file a follow-up to remove this and keep two vectors.

anuraaga: Sure #1981

jkwatson:
This doesn't actually remove the `getEntries()` method from the public API, so doesn't making `Entry` non-public break things?

anuraaga:
Oops, thanks

jkwatson (approved):
Thanks!

bogdandrutu (approved):

carlosalberto (approved):

Issue #1944, opened 2020-10-30
Consider dropping guava from core SDK dependencies


Oberon00:
Guava is a rather big dependency (size-wise) and the usages in the SDK are rather trivial. At Dynatrace we managed to replace Guava with core Java 7 standard library with little effort in our internal fork of opentelemetry-java. With Java 8, it should be even easier to achieve this.

The advantage would be that the deployment size shrinks considerably (for the core SDK + depenendencies probably by at least 80%). This could be important for size-sensitive deployment environments such as AWS lambda.

anuraaga:
Luckily it's not in the API so we can take some time on this, but agree (was going to bring it up at some point myself).

jkwatson:
closing this. we no longer depend on guava transitively in the SDK. Please re-open if you see further work to be done.

PR #1982, opened 2020-11-03
Expose foreach consumer instead of Entry for baggage.

### LINKED ISSUE #1960: Remove Baggage Entry?

anuraaga:
Since sent the tracestate version, figured I'd just go ahead and send this out.

Fixes #1960

### DIFF
--- a/api/src/main/java/io/opentelemetry/api/baggage/Baggage.java
+++ b/api/src/main/java/io/opentelemetry/api/baggage/Baggage.java
@@ -7,7 +7,6 @@ package io.opentelemetry.api.baggage;
 
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.ImplicitContextKeyed;
-import java.util.Collection;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
@@ -62,16 +61,19 @@ public interface Baggage extends ImplicitContextKeyed {
     return context.with(BaggageContextKey.KEY, this);
   }
 
-  /**
-   * Returns an immutable collection of the entries in this {@code Baggage}. Order of entries is not
-   * guaranteed.
-   *
-   * @return an immutable collection of the entries in this {@code Baggage}.
-   */
-  Collection<Entry> getEntries();
+  /** Returns the number of entries in this {@link Baggage}. */
+  int size();
+
+  /** Returns whether this {@link Baggage} is empty, containing no entries. */
+  default boolean isEmpty() {
+    return size() == 0;
+  }
+
+  /** Iterates over all the entries in this {@link Baggage}. */
+  void forEach(BaggageConsumer consumer);
 
   /**
-   * Returns the {@code String} associated with the given key.
+   * Returns the {@code String} value associated with the given key, without metadata.
    *
    * @param entryKey entry key to return the value for.
    * @return the value associated with the given key, or {@code null} if no {@code Entry} with the
--- a/api/src/main/java/io/opentelemetry/api/baggage/BaggageConsumer.java
+++ b/api/src/main/java/io/opentelemetry/api/baggage/BaggageConsumer.java
@@ -1,0 +1,13 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.api.baggage;
+
+/** A consumer of entries in {@link Baggage}. */
+@FunctionalInterface
+public interface BaggageConsumer {
+  /** Consumes an entry of a {@link Baggage}. */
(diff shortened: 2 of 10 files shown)


carlosalberto:
Any reason to not keep the `entries()` method, other than the potential removal of `Entry`? 

Although this is very Java-ish, sometimes it feels like an overkill. Also, using `entries()` under Android would also feel better ;)

anuraaga:
I think either implementing `Map` directly, or having `asMap` will be user friendly since they're very familiar interfaces. But they're bigger changes, and until then, I think it's good to avoid entries, at least as currently implemented, as at least it seems better long term as `entrySet` to fit with normal maps. How does that sound?

carlosalberto:
I think it's fine either way, as long as we expose a simple way to return the values (besides the new `forEach()` functionality).

Also, the Specification at this moment requires us to return either a collection or an iterator over the values, so I'd vow for not breaking it. Because of this, I'd prefer if `getEntries()` stay and we discuss the removal of `Entry` a little more.

anuraaga:
Do you think we can add it back after 0.10.0? We want it to be mostly stable for the API, and also want to get it out this week (from what I understand) so I'm focusing on getting semver-breaking API changes in for that. Adding methods afterward shouldn't be a problem though, but I don't want to commit to e.g getEntries vs entrySet vs asMap for now if it's possible as those are names that effect semver.

Oberon00:
What is the advantage of thesis foreach stuff over itertors/collections? From an user-perspective it seems far clunkier.

jkwatson:
I'm not a fan of losing the ability to do Java 8 `Stream` operations on the contents of the `Baggage`. Agree with @Oberon00 that this is a clunkier API. Is it worth introducing that clunkiness just for the sake of consistency with Labels and Attributes?

Oberon00:
Maybe we should actually change labels and attributes to provide some sort of iterator?

jkwatson:
> Maybe we should actually change labels and attributes to provide some sort of iterator?

We could consider adding an iterator, but that would require wrapping each key-value pair in an `Entry` style class, leading to more allocation pressure. I wouldn't want to get rid of the lighter-weight `forEach` for use-cases that are memory sensitive.

anuraaga:
I don't love the foreach either but I do think we should be consistent with the others - is there a difference in memory pressure for baggage / tracestate vs attributes when they're used?

I figured since forEach is consistent with Map, it could provide an easier path to exposing a Map interface in the future. If losing the iterable now and adding later is not palatable, then I'd like to go ahead and expose `entrySet` now instead of getEntries. Does it make sense?

carlosalberto:
> I wouldn't want to get rid of the lighter-weight forEach for use-cases that are memory sensitive.

+1 on that. Although it's not user-friendly, it is supposed to be *only* consumed by SDKs 😉

carlosalberto:
>  then I'd like to go ahead and expose entrySet now instead of getEntries. Does it make sense?

Not a strong opinion, but fine by me.

jkwatson:
After some more reflection, I'm ok with having all the "open telemetry iterable" interfaces work the same way, even if it's a little clunky for the first release. We can always add more options later.

jkwatson (approved):
Thanks!

anuraaga:
Since #1979 is merged is this ok to merge too? My plan is to then have the tracestate / baggage become arrays, then add a mechanism for them to all expose something like `asMap` or directly implement it.

jkwatson (approved):
Yes, I think this is the right route for consistency. Being able to model all the various and sundry types as flat lists/arrays internally seems like a win for memory usage overall.

jkwatson:
Discussion at the SIG meeting this morning was to go ahead with this. @carlosalberto was going to log an issue to track being able to access the entries (and assumedly any other `forEach`ables in the project) more directly.

Oberon00:
>  Being able to model all the various and sundry types as flat lists/arrays internally seems like a win

A Map interface can be backed by an array. It would just be a bit of boilerplate code if we would go with that optimization to implement Map on top of a sorted array, but since we only need read-only it should be OK. And before #1944, I think Guava would have that implemented (how does ImmutableMap work?)

jkwatson:
> Oh, it's just a `Collection`? Then `Arrays.asList` would do the job. Collection gives us plenty flexibility already. If we want the option of using a data structure that doesn't know it's size, Iterable would also still be better than the forEach callback.

It's not that simple because the underlying list has alternating entries of keys and values, to reduce allocations.

Oberon00:
OK, then a bit of boilerplate is required, but implementing the Collection interface is no rocket science.

PR #1983, opened 2020-11-03
Make Attribute/LabelConsumer consistent with Java 8 BiConsumer

anuraaga:


api/src/test/java/io/opentelemetry/api/common/AttributesTest.java
```diff
@@ -61,7 +61,7 @@ void forEach_empty() {
     emptyAttributes.forEach(
         new AttributeConsumer() {
```
anuraaga, line 62: @jkwatson Did we realize that AttributeConsumer can't be used with lambdas due to the type parameter?

carlosalberto (approved):

api/src/test/java/io/opentelemetry/api/common/AttributesTest.java
jkwatson: yes. This is unfortunate, but true.

jkwatson (approved):
Thanks!

PR #1984, opened 2020-11-03
Rename exporter artifacts to singular

### LINKED ISSUE #1967: Rename exporter artifacts from opentelemetry-exporters-* to opentelemetry-exporter-*?
Currently we have exporter artifacts plural but extensions are singular (opentelemetry-extension-*). We should be consistent for them. Singular seem ok?

anuraaga:
Fixes #1967


carlosalberto (approved):

carlosalberto:
Not related to this change, but as we have introduced many breaking changes for 0.10.0, it feels like we should enforce in the future a CHANGELOG entry, as we do in the Specification (which helps A LOT people wanting to see what has changed).

jkwatson:
> Not related to this change, but as we have introduced many breaking changes for 0.10.0, it feels like we should enforce in the future a CHANGELOG entry, as we do in the Specification (which helps A LOT people wanting to see what has changed).

Agreed 100%.  I've been also keeping Draft release notes up-to-date, so at least maintainers can see what's what. That's not as good as pre-signaling to end-users as well what we're planning on releasing. So, 👍 to mandating CHANGELOG entries with PRs, that end up impacting end-users.

jkwatson (approved):
Singular artifact names make sense to me. Thanks!

bogdandrutu (approved):

PR #1990, opened 2020-11-03
Remove unused class InMemoryTracing, junit5 helper is enough

bogdandrutu:


anuraaga:
I think we should add a junit4 rule (and switch current junit5 dependency to compileOnly) before removing this, I can get on it.

jkwatson (approved):
Thanks! Cleanup FTW

carlosalberto (approved):

bogdandrutu:
@anuraaga is it ok if we merge this and you provide later the extension?

anuraaga:
I'm hoping we don't make a release with no junit4 support. Will be working on that today, if it seems likely to merge into a release then I have no problem with merging this in advance.

anuraaga:
Sent #1996

PR #1992, opened 2020-11-03
Rename package name for exporters to singular exporter

bogdandrutu:
Fix module name, and ensure that jaeger thrift is not in the same package as jaeger.


carlosalberto (approved):

exporters/jaeger-thrift/src/main/java/io/opentelemetry/exporter/jaeger_thrift/Adapter.java
```diff
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package io.opentelemetry.exporters.jaeger;
+package io.opentelemetry.exporter.jaeger_thrift;
```
jkwatson, line 6: Hm. I don't like underscores in package names. maybe `io.opentelemetry.exporter.jaeger.thrift`?

bogdandrutu: Cannot do that I think, because of the java9 modules, the grpc is in jaeger so I think we cause a module conflict.

bogdandrutu: I can move the grpc one to `jaeger.grpc` and this to `jaeger.thrift` that should be safe

bogdandrutu: Also fyi the checkstyle failed :)) cannot have "_" in the name :)). I used for the moment `jaeger.thrift` but can someone confirm if that is a problem or not with java9 modules?

jkwatson: I believe that JPMS does not restrict sub-packages from being in a separate module, as long as each package is only exported by a single module. There's really no such thing as a "sub package" in java. There shouldn't be any issues. But, also, having `.grpc` for the other one sounds totally fine to me, as well.

anuraaga (approved):
Oops thanks!

PR #1996, opened 2020-11-04
Add a junit4 rule for OTel testing.

anuraaga:
Also added a unit test for the junit5 extension

### DIFF
--- a/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/junit4/OpenTelemetryRule.java
+++ b/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/junit4/OpenTelemetryRule.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.testing.junit4;
+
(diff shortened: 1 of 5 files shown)


sdk/testing/src/main/java/io/opentelemetry/sdk/testing/junit4/OpenTelemetryRule.java
```diff
@@ -0,0 +1,112 @@
...
+            .setTracerProvider(tracerProvider)
+            .build();
+
+    return new OpenTelemetryRule(openTelemetry, spanExporter);
+  }
+
+  private final OpenTelemetrySdk openTelemetry;
+  private final InMemorySpanExporter spanExporter;
+
+  private OpenTelemetry previousGlobalOpenTelemetry;
```
jkwatson, line 71: does this need to be `volatile` ? I'm never 100% clear on the threading model of junit.

jkwatson (approved):

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/junit4/OpenTelemetryRule.java
anuraaga: Good point - I'm not sure either and it doesn't hurt so added

PR #1999, opened 2020-11-04
Remove testbed artifact, add testbed tests to trace sdk

bogdandrutu:


jkwatson (approved):
This seems like a good cleanup. I was never sure why these tests were in their own module in the first place.

bogdandrutu:
@jkwatson not sure who thought this way 🥇  :))

PR #2004, opened 2020-11-04
Keep Span Attributes in order, as the spec strongly suggests.

jkwatson:
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/common/common.md#attributes

This only changes attributes attached to Spans. Other attributes are sorted in lexigraphical order, by key, at the moment.

I don't know why the spec says that we should keep them in order...that feels like an odd expectation, especially for Java users who would probably think of them like a `Map`, which wouldn't provide ordering guarantees, normally.

### DIFF
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/AttributesMap.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/AttributesMap.java
@@ -7,9 +7,9 @@ package io.opentelemetry.sdk.trace;
 
 import io.opentelemetry.api.common.AttributeConsumer;
 import io.opentelemetry.api.common.AttributeKey;
-import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.common.ReadableAttributes;
-import java.util.HashMap;
+import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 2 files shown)


carlosalberto (approved):

carlosalberto:
Thanks!

anuraaga (approved):
Just wondering is consistency with `Attributes` important?

jkwatson:
> Just wondering is consistency with `Attributes` important?

No clue, but it can be done in a later PR. I'm not sure how exactly we would do it, though, without losing much of the advantage of the current array-based implementation.

Oberon00:
> I don't know why the spec says that we should keep them in order...that feels like an odd expectation, especially for Java users who would probably think of them like a Map, which wouldn't provide ordering guarantees, normally.

👍 
I think we should remove that wording from the spec, or otherwise move it to the general attributes spec.

jkwatson:
> > I don't know why the spec says that we should keep them in order...that feels like an odd expectation, especially for Java users who would probably think of them like a Map, which wouldn't provide ordering guarantees, normally.
> 
> 👍
> I think we should remove that wording from the spec, or otherwise move it to the general attributes spec.

Unfortunately, it is already in the general attributes spec: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/common/common.md#attributes

Oberon00:
Well, that is better than if it was only in the span spec IMHO. But we should still discuss removing it.

PR #2007, opened 2020-11-04
Make the BatchRecorderSdk actually batch the recordings

### LINKED ISSUE #1621: Finish implementing the BatchRecorder for batched metric recordings.
Currently, the BatchRecorder just delegates all calls directly to the underlying instruments, rather than saving up the recordings until the `record` method is invoked. The implementation needs to be finished.

jkwatson:
Rather than just forwarding them immediately.

Resolves #1621

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/BatchRecorderSdk.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/BatchRecorderSdk.java
@@ -10,18 +10,24 @@ import io.opentelemetry.api.metrics.BatchRecorder;
 import io.opentelemetry.api.metrics.DoubleCounter;
 import io.opentelemetry.api.metrics.DoubleUpDownCounter;
 import io.opentelemetry.api.metrics.DoubleValueRecorder;
+import io.opentelemetry.api.metrics.Instrument;
 import io.opentelemetry.api.metrics.LongCounter;
 import io.opentelemetry.api.metrics.LongUpDownCounter;
 import io.opentelemetry.api.metrics.LongValueRecorder;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 /**
  * Minimal implementation of the {@link BatchRecorder} that simply redirects the calls to the
  * instruments.
- *
- * <p>TODO: Add an async queue processing to process batch records.
  */
 final class BatchRecorderSdk implements BatchRecorder {
   private final Labels labelSet;
+
+  private volatile Map<Instrument, List<Number>> recordings = new HashMap<>();
+  private volatile Map<Instrument, List<Number>> previousRecordings = new HashMap<>();
 
   BatchRecorderSdk(String... keyValuePairs) {
     this.labelSet = Labels.of(keyValuePairs);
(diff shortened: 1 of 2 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/BatchRecorderSdk.java
```diff
@@ -10,61 +10,105 @@
...
-    // No-op in this minimal implementation.
+    previousRecordings.clear();
+
+    Map<Instrument, List<Number>> temp = recordings;
+    recordings = previousRecordings;
+    previousRecordings = temp;
+
+    previousRecordings.forEach(
+        (instrument, numbers) -> {
+          if (instrument instanceof DoubleUpDownCounter) {
```
anuraaga, line 82: If I'm reading this right, I think there's less code duplication if we have the loop on the outside of the instanceof checks

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/BatchRecorderSdk.java
```diff
@@ -10,61 +10,105 @@
...
+  private volatile Map<Instrument, List<Number>> previousRecordings = new HashMap<>();
+
   BatchRecorderSdk(String... keyValuePairs) {
     this.labelSet = Labels.of(keyValuePairs);
   }
 
   @Override
   public BatchRecorder put(LongValueRecorder valueRecorder, long value) {
-    valueRecorder.record(value, labelSet);
+    recordings.computeIfAbsent(valueRecorder, instrument -> new ArrayList<>()).add(value);
```
anuraaga, line 38: I'm guessing this can be called concurrently - so the maps need to be `ConcurrentHashMap` and the value something synchronized, probably `BlockingQueue`.

anuraaga: Though given the primitive types have to be boxed anyways, I think wrapping in our own type would have very low overhead. So instead of two maps, I guess we can use a single queue? Would be easier to reason about.

```java
private static class LongRecording implements Recording {
  private final Instrument instrument;
  private final long value;
}

BlockingQueue<Recoding> recordings;

record() {
  recordings.drainTo(buffer);
  for (Recording recording : buffer) {
    ...
  }
```

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/BatchRecorderSdk.java
jkwatson: another thing I went back and forth on a couple times. I doubt it makes much difference performance-wise, so if you think it's clearer the other way, happy to change it.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/BatchRecorderSdk.java
jkwatson: yeah, I went back and forth about this. I'd be happy to change it to something like that. And, yes, you're totally right about the concurrent hashmap bit.

jkwatson: @anuraaga take a look now.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/BatchRecorderSdk.java
jkwatson: done

anuraaga (approved):
Looks much better, thanks!

PR #2015, opened 2020-11-05
Switch animal sniffer signature to gummy bears.

### LINKED ISSUE #1808: gradle: Centralize animalsniffer signature dependencies

anuraaga:
As far as I can tell, the androidscents we use and is commonly documented is basically unmaintained - I can't find a project page and the last release was Android API 27 some years ago. So naturally they also don't have any signature that models the desugaring mechanism we are expecting Android users to do.

Gummy bears is a relatively new project but seems to be filling this void excellently

https://github.com/open-toast/gummy-bears

I confirmed that the new signature accepts `Instant` with the `coreLib` classifier version and fails without it

Fixes #1808

Changelog:
- Replace animalsniffer Android API check with gummy bears to recognize modern Android desugaring

### DIFF
--- a/exporters/jaeger/build.gradle
+++ b/exporters/jaeger/build.gradle
@@ -30,7 +30,7 @@ dependencies {
     testRuntime "io.grpc:grpc-netty-shaded:${grpcVersion}"
 
     signature "org.codehaus.mojo.signature:java18:1.0@signature"
-    signature "net.sf.androidscents.signature:android-api-level-24:7.0_r2@signature"
+    signature libraries.android_signature
 }
 
 animalsniffer {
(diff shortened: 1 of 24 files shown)


exporters/jaeger/build.gradle
```diff
@@ -30,7 +30,7 @@ dependencies {
     testRuntime "io.grpc:grpc-netty-shaded:${grpcVersion}"
 
     signature "org.codehaus.mojo.signature:java18:1.0@signature"
```
anuraaga, line 32: I'll remove these unnecessary java 8 ones in a separate PR

jkwatson (approved):
The title of this PR is simply ridiculous. Thanks!

PR #2016, opened 2020-11-05
Implement baggage using array of key/value pairs instead of Map.

anuraaga:
This just reuses `ImmutableKeyValuePairs` for most heavy-lifting.

I did this now mainly since I realized `equals` is buggy (doesn't handle parenting correctly) and this fixes it.

Changelog:

- Baggage implementation changed from a HashMap with parenting for data reuse to array.

### DIFF
--- a/api/src/main/java/io/opentelemetry/api/baggage/ImmutableBaggage.java
+++ b/api/src/main/java/io/opentelemetry/api/baggage/ImmutableBaggage.java
@@ -5,37 +5,33 @@
 
 package io.opentelemetry.api.baggage;
 
+import static java.util.Objects.requireNonNull;
+
+import com.google.auto.value.AutoValue;
+import io.opentelemetry.api.internal.ImmutableKeyValuePairs;
 import io.opentelemetry.context.Context;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
+import java.util.ArrayList;
+import java.util.List;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
 @Immutable
-// TODO: Migrate to AutoValue
-// @AutoValue
-class ImmutableBaggage implements Baggage {
+class ImmutableBaggage extends ImmutableKeyValuePairs<String, Entry> implements Baggage {
 
-  static final Baggage EMPTY = new ImmutableBaggage.Builder().build();
+  static final Baggage EMPTY = new ImmutableBaggage.Builder().setNoParent().build();
 
-  // The types of the EntryKey and Entry must match for each entry.
-  private final Map<String, Entry> entries;
-  @Nullable private final Baggage parent;
+  @AutoValue
+  @Immutable
+  abstract static class ArrayBackedBaggage extends ImmutableBaggage {
+    ArrayBackedBaggage() {}
 
-  /**
-   * Creates a new {@link ImmutableBaggage} with the given entries.
-   *
-   * @param entries the initial entries for this {@code BaggageSdk}.
-   * @param parent providing a default set of entries
-   */
-  private ImmutableBaggage(Map<String, ? extends Entry> entries, Baggage parent) {
-    this.entries =
-        Collections.unmodifiableMap(new HashMap<>(Objects.requireNonNull(entries, "entries")));
-    this.parent = parent;
+    @Override
+    public abstract List<Object> data();
+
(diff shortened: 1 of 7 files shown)


api/src/main/java/io/opentelemetry/api/baggage/ImmutableBaggage.java
```diff
@@ -160,38 +97,48 @@ public String toString() {
...
+
+      List<Object> data = this.data;
+      if (parent != null && !parent.isEmpty()) {
+        List<Object> merged = new ArrayList<>(parent.size() * 2 + data.size());
+        parent.forEach(
+            (key, value, metadata) -> {
+              merged.add(key);
+              merged.add(Entry.create(key, value, metadata));
+            });
+        merged.addAll(data);
```
jkwatson, line 138: I believe the original implementation intentionally was built not to make copies of all the data, but rely on traversing up the stack. I've been told by @bogdandrutu that this was a very large CPU/memory savings at google. Something to consider before we remove that implementation detail.

api/src/main/java/io/opentelemetry/api/internal/ImmutableKeyValuePairs.java
```diff
@@ -3,34 +3,36 @@
...
  * @see Labels
  * @see Attributes
  */
 @SuppressWarnings("rawtypes")
 @Immutable
-abstract class ImmutableKeyValuePairs<K, V> {
+public abstract class ImmutableKeyValuePairs<K, V> {
 
-  List<Object> data() {
+  public List<Object> data() {
```
jkwatson, line 35: this does bum me out quite a bit. By making this public, we've now exposed implementation details on all of Attributes, Labels and Baggage that could easily be abused. Can we make this protected, so the subclasses get access, but the public does not?

jkwatson: Just tried it locally, and yes, we can make this protected. I'll feel much better about things then.

api/src/main/java/io/opentelemetry/api/common/Attributes.java
```diff
@@ -175,7 +176,8 @@ private static Attributes sortAndFilterToAttributes(Object... data) {
         data[i] = null;
       }
     }
-    return new AutoValue_Attributes_ArrayBackedAttributes(sortAndFilter(data));
+    return new AutoValue_Attributes_ArrayBackedAttributes(
+        sortAndFilter(data, /* filterNullValues= */ false));
```
jkwatson, line 180: could we send in `true` here, and get rid of lines 175-177?

api/src/main/java/io/opentelemetry/api/internal/ImmutableKeyValuePairs.java
jkwatson: and also all the subclasses can likewise have their overrides made protected.

anuraaga: Sweet thanks!

api/src/main/java/io/opentelemetry/api/baggage/ImmutableBaggage.java
anuraaga: Yeah but I don't know if we can make it efficient. Previously we were actually copying everything when calling `getEntries`. Now with `forEach`, there's still the questionable `seenKeys` stuff. And I can't think of a way to fix the bug with `equals` without similar copying with the parent approach - actually it's exactly what Sleuth is currently doing to work around our equals being broken :P So the up-front copy seems ok to me.

anuraaga: By the way, noticed that opencensus API doesn't actually allow end users to access tags (baggage)

https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/tags/TagContext.java#L42

It appears the default implementation does seem to not use parenting and expose the tags directly, presumably requiring a downcast though not sure.

https://github.com/census-instrumentation/opencensus-java/blob/master/impl_core/src/main/java/io/opencensus/implcore/tags/TagMapImpl.java

It seems this latter one is closer to our spec and is what we want to implement here. I think a custom baggage propagator + baggage implementation should be possible still for those that want to prioritize marshalling at the expense of access speed, but I don't think it should be our default.

jkwatson (approved):
Thanks! Optimizations can be made later, if we find we need them. 👍

PR #2017, opened 2020-11-05
Add paths-ignore for PR Build

dengliming:
It's not necessary to run this build workflow when only  `*.md` files are modified. Please consider if that make sense.

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -4,6 +4,8 @@ on:
   pull_request:
     branches:
       - master
+    paths-ignore:
+      - "**.md"
 
 jobs:
   build:
--- a/.github/workflows/pr-examples-build.yml
+++ b/.github/workflows/pr-examples-build.yml
@@ -6,6 +6,7 @@ on:
       - master
     paths:
       - 'examples/**'
+      - '!**.md'
 jobs:
   build:
     name: Build


iNikem (approved):

anuraaga (commented):
In instrumentation repo we have formatting of MD files in the build, and I presume we will add that here eventually (do you want to help? :D) to have consistent tooling between the two repos. So I think we shouldn't filter them out here in anticipation of that.

Oberon00:
> we have formatting of MD files in the build, and I presume we will add that here eventually

That could go to a separate workflow though, which would be much faster.

anuraaga (approved):
Ah that's a cool idea. If anyone can help implement that on instrumentation, we would save enormous builds for md-only files :D

dengliming:
@anuraaga Thanks. I can try to take a look if I have some spare time. Can you help to file a issue to track this case in `instrumentation ` repo?

carlosalberto (approved):

.github/workflows/pr-build.yml
```diff
@@ -4,6 +4,8 @@ on:
   pull_request:
     branches:
       - master
+    paths-ignore:
+      - "**.md"
```
bogdandrutu, line 8: Add same line to examples build?

bogdandrutu (approved):

.github/workflows/pr-build.yml
dengliming: sure. will do it later.

jkwatson (approved):
Thanks!

Issue #1997, opened 2020-11-04
Provide default implementations for SpanProcessor.shutdown and forceFlush


iNikem:
I think the majority of non-exporting span processors are going to just augment span attributes in `onStart`. For them these two methods don't make any sense and every author will be forced to copy-paste `return CompletableResultCode.ofSuccess()`. Can this be the default implementation?

Oberon00:
Makes sense. The default implementation of shutdown would call forceFlush and the default implementation of forceFlush would do nothing.

PR #2025, opened 2020-11-05
Provide default implementations for SpanProcessor.shutdown and forceFlush

dengliming:
closes [https://github.com/open-telemetry/opentelemetry-java/issues/1997](https://github.com/open-telemetry/opentelemetry-java/issues/1997)

### DIFF
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/SpanProcessor.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/SpanProcessor.java
@@ -57,7 +57,7 @@ public interface SpanProcessor {
    * @return a {@link CompletableResultCode} which completes when shutdown is finished.
    */
   default CompletableResultCode shutdown() {
-    return CompletableResultCode.ofSuccess();
+    return forceFlush();
   }
 
   /**
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/SpanProcessor.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/SpanProcessor.java
@@ -56,7 +56,9 @@ public interface SpanProcessor {
    *
    * @return a {@link CompletableResultCode} which completes when shutdown is finished.
    */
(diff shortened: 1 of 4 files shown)


anuraaga (approved):
Awesome, thanks!

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/SpanProcessor.java
```diff
@@ -56,13 +56,17 @@
    *
    * @return a {@link CompletableResultCode} which completes when shutdown is finished.
    */
-  CompletableResultCode shutdown();
+  default CompletableResultCode shutdown() {
+    return CompletableResultCode.ofSuccess();
```
Oberon00, line 60:
```suggestion
    return forceFlush();
```

Oberon00 (changes requested):

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/SpanProcessor.java
dengliming: Can you explain why do we need this? : ）

Oberon00: Because shutdown should normally ensure that everything is flushed. See https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#shutdown-1

> Shutdown MUST include the effects of ForceFlush.

Oberon00 (approved):
Thanks!

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/SpanProcessor.java
dengliming: 👍 Thanks. fixed.

jkwatson (approved):
Thanks!

carlosalberto (approved):

Issue #2021, opened 2020-11-05
Latest proto version needed in 0.14.0 release


gcacace:
**Is your feature request related to a problem? Please describe.**
The change in https://github.com/open-telemetry/opentelemetry-java/commit/711efc3ed497df4d796421d004b389c4ff8e972e is needed for our use case and we would like to see this shipped in the next version of this package

**Describe the solution you'd like**
Bump-up to 0.14.0 and release to Maven Central

**Describe alternatives you've considered**
None. For our use case we need the new version in Maven Central

**Additional context**
This is required for a CloudWatch use case

jkwatson:
Hi @gcacace . We have just released 0.10.0. thanks!

gcacace:
Reopening as the new version is not in Maven Centra: https://mvnrepository.com/artifact/io.opentelemetry/opentelemetry-proto

bogdandrutu:
And does not have the proto change you need :)

gcacace:
Can we then reopen this issue please?

jkwatson:
We need to wait for a protobuf release before we will depend on it in the project. Otherwise we risk releases with non-released versions of the protobuf that won't be supported by anything.

shalk:
the commit open-telemetry/opentelemetry-proto@8ab21e9 which release in proto  v0.6.0   have released in 0.12.0 or later. ([see proto/build.gradle]( https://github.com/open-telemetry/opentelemetry-java/blob/v0.12.0/proto/build.gradle#L47)).

gcacace:
@bogdandrutu made the change to consume the latest proto package (see: https://github.com/open-telemetry/opentelemetry-java/commit/711efc3ed497df4d796421d004b389c4ff8e972e). Can someone release the new version of this Java package in Maven Central please?

jkwatson:
> @bogdandrutu made the change to consume the latest proto package (see: [711efc3](https://github.com/open-telemetry/opentelemetry-java/commit/711efc3ed497df4d796421d004b389c4ff8e972e)). Can someone release the new version of this Java package in Maven Central please?

This release has been incorporated into the main branch of the java project. It is currently available in snapshot builds. It'll get released on the normal release schedule, within a week or so.

@gcacace I will repeat what I've said before: if you have urgent dependencies on having java protobuf bindings in a more timely manner than what happens to be included in the java SDK, you should maintain your own build/publishing pipeline to manage this, and not rely on the larger java project artifacts.

[...]

PR #2027, opened 2020-11-05
Update proto to latest proto commit

bogdandrutu:
Related to https://github.com/open-telemetry/opentelemetry-java/issues/2021 


carlosalberto (approved):

arminru (approved):

jkwatson:
Wait! Is this a released version of the proto?

jkwatson (changes requested):
We should not use a non-released version of the proto

bogdandrutu:
@jkwatson do we?

jkwatson:
> @jkwatson do we?

I assert yes, we should only use released versions of the protobuf.

gcacace:
Can we please then do a release of proto and then another one for the java package?

jkwatson:
> Can we please then do a release of proto and then another one for the java package?

Can you explain why this is needed so urgently? We are on a 2 week release cycle now, and do not generally intend to release patch releases unless there is an urgent bugfix.

gcacace:
Because we need the new `DoubleSummary` data structure as per https://github.com/open-telemetry/opentelemetry-specification/issues/1146 and https://github.com/open-telemetry/opentelemetry-specification/issues/1145

jkwatson:
> Because we need the new `DoubleSummary` data structure as per [open-telemetry/opentelemetry-specification#1146](https://github.com/open-telemetry/opentelemetry-specification/issues/1146) and [open-telemetry/opentelemetry-specification#1145](https://github.com/open-telemetry/opentelemetry-specification/issues/1145)

What explains what you need, but not why it is so urgent that we would be forced to do a patch release.

jkwatson:
Also, just updating the protobuf won't mean that we're actually using the protobuf structures for anything. How is just updating the protobuf definitions going to help you at all?

gcacace:
Because we are using opentelemetry-proto as transport layer for our use case.

jkwatson:
> Because we are using opentelemetry-proto as transport layer for our use case.

So...let me understand. You're using the java repo's published generated protobuf classes for something other than use in the SDK? That does not sound like a reason for us to do an emergency patch release. If you need these protobuf classes for something outside the use in the SDK, I recommend setting up a project that will build and publish those for you, rather than depending on this project with just happens to publish them as a part of the release process.

jkwatson:
IMO, the correct solution to this problem is to have the protobuf repository publish various language bindings. Then, the opentelemetry-java project could depend on them like a normal library (rather than as a git submodule), and usages by other projects would be supported as well.

gcacace:
> IMO, the correct solution to this problem is to have the protobuf repository publish various language bindings. Then, the opentelemetry-java project could depend on them like a normal library (rather than as a git submodule), and usages by other projects would be supported as well.

Agree, since for our use case we only need the OT Protobuf per-language bindings.

anuraaga:
I agree that publishing stubs from the opentelemetry-proto repo would be ideal for contributor experience and version management.

I have sent https://github.com/open-telemetry/opentelemetry-proto/pull/231 as a proposal to start to make that happen.

For reference, I've done something similar on a previous project and it works pretty well https://github.com/infostellarinc/stellarstation-api

gcacace:
Would it be possible to create a new release of the proto with `DoubleSummary` support?

jkwatson:
> Would it be possible to create a new release of the proto with `DoubleSummary` support?

Please post that question in the proto repo, since it's not something the java folks control.

jkwatson:
closing this. when the proto project cuts a release, I'll be happy to update our dependency.

PR #2029, opened 2020-11-05
Require the OpenTelemetrySdk to have an instance of TracerSdkManagement configured

### LINKED ISSUE #2018: Setting custom TracerProvider breaks access to TracerSdkManagement
Adding the following test to `OpenTelemetrySdkTest` results in test failure:

```
  @Test
  void testTracerProviderAccess() {
    OpenTelemetrySdk openTelemetry = OpenTelemetrySdk.builder()
        .build();
    assertThat(openTelemetry.getTracerProvider()).isEqualTo(tracerProvider);
    assertThat(openTelemetry.getTracerManagement()).isNotNull();
  }
```

```
class io.opentelemetry.api.trace.TracerProvider$MockitoMock$1536679414 cannot be cast to class io.opentelemetry.sdk.OpenTelemetrySdk$ObfuscatedTracerProvider (io.opentelemetry.api.trace.TracerProvider$MockitoMock$1536679414 and io.opentelemetry.sdk.OpenTelemetrySdk$ObfuscatedTracerProvider are in unnamed module of loader 'app')
java.lang.ClassCastException: class io.opentelemetry.api.trace.TracerProvider$MockitoMock$1536679414 cannot be cast to class io.opentelemetry.sdk.OpenTelemetrySdk$ObfuscatedTracerProvider (io.opentelemetry.api.trace.TracerProvider$MockitoMock$1536679414 and io.opentelemetry.sdk.OpenTelemetrySdk$ObfuscatedTracerProvider are in unnamed module of loader 'app')
	at io.opentelemetry.sdk.OpenTelemetrySdk.getTracerManagement(OpenTelemetrySdk.java:116)
	at io.opentelemetry.sdk.OpenTelemetrySdkTest.testTracerProviderAccess(OpenTelemetrySdkTest.java:93)
```

It also breaks `testGlobalDefault` from the same class as it happens to run after the one above:

```
Trying to access global TracerSdkManagement but global TracerProvider is not an instance created by this SDK.
java.lang.IllegalStateException: Trying to access global TracerSdkManagement but global TracerProvider is not an instance created by this SDK.
	at io.opentelemetry.sdk.OpenTelemetrySdk.getGlobalTracerManagement(OpenTelemetrySdk.java:52)
	at io.opentelemetry.sdk.OpenTelemetrySdkTest.testGlobalDefault(OpenTelemetrySdkTest.java:36)
```

jkwatson:
And add an option to the builder to set one explicitly.

This is one possible way to resolve #2018 .

### DIFF
--- a/sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
+++ b/sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
@@ -74,7 +74,7 @@ public final class OpenTelemetrySdk implements OpenTelemetry {
   private static final AtomicBoolean INITIALIZED_GLOBAL = new AtomicBoolean();
 
   private final TracerProvider tracerProvider;
-  @Nullable private final TracerSdkManagement tracerSdkManagement;
+  private final TracerSdkManagement tracerSdkManagement;
   private final MeterProvider meterProvider;
   private final ContextPropagators contextPropagators;
 
@@ -83,7 +83,7 @@ public final class OpenTelemetrySdk implements OpenTelemetry {
 
   private OpenTelemetrySdk(
       TracerProvider tracerProvider,
-      @Nullable TracerSdkManagement tracerSdkManagement,
+      TracerSdkManagement tracerSdkManagement,
       MeterProvider meterProvider,
       ContextPropagators contextPropagators,
       Clock clock,
@@ -123,10 +123,7 @@ public final class OpenTelemetrySdk implements OpenTelemetry {
 
   /** Returns the {@link TracerSdkManagement} for this {@link OpenTelemetrySdk}. */
   public TracerSdkManagement getTracerManagement() {
-    if (tracerSdkManagement != null) {
-      return tracerSdkManagement;
-    }
-    return (TracerSdkProvider) ((ObfuscatedTracerProvider) tracerProvider).unobfuscate();
+    return tracerSdkManagement;
   }
 
   /** Returns a new {@link Builder} initialized with the values of this {@link OpenTelemetrySdk}. */
--- a/sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
+++ b/sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
@@ -47,13 +47,20 @@ public final class OpenTelemetrySdk implements OpenTelemetry {
 
   /** Returns the global {@link TracerSdkManagement}. */
   public static TracerSdkManagement getGlobalTracerManagement() {
-    TracerProvider tracerProvider = OpenTelemetry.get().getTracerProvider();
+    OpenTelemetry openTelemetry = OpenTelemetry.get();
+    if (openTelemetry instanceof OpenTelemetrySdk) {
+      return ((OpenTelemetrySdk) openTelemetry).tracerSdkManagement;
+    }
+    TracerProvider tracerProvider = openTelemetry.getTracerProvider();
+    if (tracerProvider instanceof TracerSdkManagement) {
+      return (TracerSdkManagement) tracerProvider;
+    }
     if (!(tracerProvider instanceof ObfuscatedTracerProvider)) {
       throw new IllegalStateException(
           "Trying to access global TracerSdkManagement but global TracerProvider is not an "
               + "instance created by this SDK.");
     }
-    return (TracerSdkProvider) ((ObfuscatedTracerProvider) tracerProvider).unobfuscate();
+    return (TracerSdkManagement) ((ObfuscatedTracerProvider) tracerProvider).unobfuscate();
   }
 
   /** Returns the global {@link MeterSdkProvider}. */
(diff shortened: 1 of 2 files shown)


sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
```diff
@@ -47,13 +47,20 @@ public static OpenTelemetrySdk get() {
...
+    OpenTelemetry openTelemetry = OpenTelemetry.get();
+    if (openTelemetry instanceof OpenTelemetrySdk) {
+      return ((OpenTelemetrySdk) openTelemetry).tracerSdkManagement;
+    }
+    TracerProvider tracerProvider = openTelemetry.getTracerProvider();
+    if (tracerProvider instanceof TracerSdkManagement) {
+      return (TracerSdkManagement) tracerProvider;
+    }
     if (!(tracerProvider instanceof ObfuscatedTracerProvider)) {
       throw new IllegalStateException(
```
anuraaga, line 59: Should we just return a no-op `TracerSdkManagement` instead of throwing an exception, perhaps with a debug log?

jkwatson: I'm not sure. Would that possibly hide problems and make it more confusing for people to debug if/when things weren't working?

sdk/all/src/test/java/io/opentelemetry/sdk/OpenTelemetrySdkTest.java
```diff
@@ -82,4 +85,15 @@ void testReconfigure() {
     assertThat(openTelemetry.getResource()).isEqualTo(resource);
     assertThat(openTelemetry.getClock()).isEqualTo(clock);
   }
+
+  @Test
+  void testTracerProviderAccess() {
+    OpenTelemetrySdk openTelemetry =
+        OpenTelemetrySdk.builder()
+            .setTracerProvider(tracerProvider)
+            .setTracerSdkManagement(tracerSdkManagement)
```
iNikem, line 94: What if I don't set `tracerSdkManagement` manually? Will this test fail?

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
anuraaga: Setting a tracerprovider that isn't an SDK one seems pretty explicit so Just logging and retuning a no-op seems like a reasonable fallback for this incorrect usage I guess (incorrect usage being it's not allowed to call `getTracerManagement` if the registered provider is not an SDK tracer, either because it's default or custom).

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
```diff
@@ -135,17 +145,39 @@ public Builder toBuilder() {
 
     private TracerProvider tracerProvider;
     private MeterProvider meterProvider;
+    private TracerSdkManagement tracerSdkManagement;
 
     /**
      * Sets the {@link TracerProvider} to use. This can be used to configure tracing settings by
-     * returning the instance created by a {@link TracerSdkProvider.Builder}.
+     * returning the instance created by a {@link TracerSdkProvider.Builder}. If the TracerProvider
```
iNikem, line 152: I think this is very confusing now. Why may `TracerProvider` want to implement `TracerSdkManagement`? The fact that `TracerSdkProvider` implements both of these interfaces is accidental(?) implementation detail.

And this still forces any user who wants to implement their own `TraceProvider` to think about `TracerSdkManagement`, which is not obvious. Btw, do we have any documentation/manual/example for implementing `TraceProvider`?

sdk/all/src/test/java/io/opentelemetry/sdk/OpenTelemetrySdkTest.java
jkwatson: it will, because you haven't fully built a valid SDK.

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
jkwatson: I disagree, actually. What this requires is this: if you want to build a valid *`OpenTelemetrySdk`* (not `OpenTelemetry` in general), you have to provide something that implements the management interface. 

Every implementation of `OpenTelemetry` will have its own specific needs and requirements. The default SDK requires a management interface to be provided.

iNikem: Ok, but then if I want to change one of the `TracerProvider`/`TracerSdkManagement` I actually have to change both of them, correct? I cannot just call `setTracerProvider`, I have to call `setTracerSdkManagement`. So probably we should not have them both as separate methods? Instead have one method to set both.

jkwatson: If your TracerProvider happens to implement both, you can set them both with just the one method, but I do agree things are pretty messy right now. Other solutions are welcome!

jkwatson:
closing this in favor of #2069 as a better solution to this problem.

PR #2031, opened 2020-11-05
Revert "Add paths-ignore for PR Build (#2017)"

jkwatson:
This reverts commit 72c88b61

Unfortunately, this made it so doc-only PRs can never go green, because we require the builds to pass for PRs.


anuraaga (approved):
Oops, thanks

tylerbenson (approved):

PR #2034, opened 2020-11-06
Add coverage of exceptional cases (and some handling) to opentracer s…

anuraaga:
…him.

I think ocshim has brought our coverage threshold down below our target. Figured I'd hit some low-hanging fruit in otshim but a more stable approach will probably be to define lower thresholds for certain projects.


carlosalberto (approved):

opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanBuilderShim.java
```diff
@@ -136,6 +136,9 @@ public SpanBuilder withTag(String key, boolean value) {
 
   @Override
   public SpanBuilder withTag(String key, Number value) {
+    if (value == null) {
```
jkwatson, line 139: good catch!

jkwatson (approved):
Thanks!

PR #2035, opened 2020-11-06
Add AssertJ custom assertions for SpanData and ReadableAttributes.

anuraaga:
Still plenty that we can add, but here's some assertions to start with. 

Changelog
- Adds a testing helper with type-specific assertions for SpanData and ReadableAttributes based on AssertJ


sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/OpenTelemetryAssertions.java
```diff
@@ -0,0 +1,28 @@
...
+import io.opentelemetry.api.common.ReadableAttributes;
+import io.opentelemetry.sdk.trace.data.SpanData;
+import org.assertj.core.api.Assertions;
+
+/**
+ * Entry point for assertion methods for OpenTelemetry types. To use type-specific assertions,
+ * static import any {@code assertThat} method in this class instead of {@code
+ * Assertions.assertThat}.
+ */
+public class OpenTelemetryAssertions extends Assertions {
```
Oberon00, line 17: Private constructor?

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/SpanDataAssert.java
```diff
@@ -0,0 +1,370 @@
...
+      failWithMessage("Expected span to have remote parent but did not");
+    }
+    return this;
+  }
+
+  /** Asserts the span has ended. */
+  public SpanDataAssert hasEnded() {
+    isNotNull();
+    if (!actual.hasEnded()) {
+      failWithMessage("Expected span to have ended but did has not");
```
Oberon00, line 311:
```suggestion
      failWithMessage("Expected span to have ended but did not");
```

Oberon00, line 311: Maybe add the span name to the message?

jkwatson:
cool!

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/OpenTelemetryAssertions.java
anuraaga: Made it protector - a user might extend this into their own custom assertions class, doesn't hurt since just a bag of statics.

Oberon00: Extending would also work with a private ctor I think. You just can't instantiate the derived class. But protected is OK for me here.

anuraaga: Yeah thought it might work but when I tried it, it didn't :)

jkwatson (approved):

PR #2036, opened 2020-11-06
Reject null instrumentationName for Meters and Tracers

### LINKED ISSUE #1879: OpenTelemetry.getTracer(...) differs from API spec

MariusVolkhart:
The parameters were already marked as being @Nonnull. However, the SDK implementation failed when null was passed, while the API allowed null. The result is that one may publish a library that compiles and tests fine, but fails when the SDK is applied by an application developer.

This change adds test to verify the non-null behavior is consistently enforced, and adds checks to the API to enforce non-null. These checks will prevent library authors from creating libraries that crash at runtime.

Fixes #1879

### DIFF
--- a/api/src/main/java/io/opentelemetry/api/trace/DefaultTracerProvider.java
+++ b/api/src/main/java/io/opentelemetry/api/trace/DefaultTracerProvider.java
@@ -6,6 +6,7 @@
 package io.opentelemetry.api.trace;
 
 import javax.annotation.concurrent.ThreadSafe;
+import java.util.Objects;
 
 @ThreadSafe
 class DefaultTracerProvider implements TracerProvider {
@@ -23,6 +24,7 @@ class DefaultTracerProvider implements TracerProvider {
 
   @Override
   public Tracer get(String instrumentationName, String instrumentationVersion) {
+    Objects.requireNonNull(instrumentationName);
     return Tracer.getDefault();
   }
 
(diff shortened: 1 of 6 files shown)


api/src/main/java/io/opentelemetry/api/trace/DefaultTracerProvider.java
```diff
@@ -23,6 +24,7 @@ public Tracer get(String instrumentationName) {
 
   @Override
   public Tracer get(String instrumentationName, String instrumentationVersion) {
+    Objects.requireNonNull(instrumentationName);
```
jkwatson, line 27: we can't add these requirements. The API must not throw exceptions in the face of bad inputs.

jkwatson:
The API is required to not throw exceptions in the face of bad inputs, so as to not adversely impact running programs that use badly written library instrumentation (or manual instrumentation for that matter).

Issue #1818, opened 2020-10-16
Add a way to enforce default context storage implemenation


Oberon00:
**Is your feature request related to a problem? Please describe.**
For example in an auto-instrumenting agent, we do not want the Context implementation to scan for user-provided SPIs that may or may not work as expected with our agents. But there is not way to enforce the default thread local storage (for example, there is no provider class for the default storage).

Relevant code is in this function: https://github.com/open-telemetry/opentelemetry-java/blob/fa1e876df20910ca5ec3d43a788f2bcf35013d68/context/src/main/java/io/opentelemetry/context/LazyStorage.java#L39

**Describe the solution you'd like**
The simplest solution would be to add a special name like `default` that is accepted for the `io.opentelemetry.context.contextStorageProvider` system property.

**Describe alternatives you've considered**
One could add a Provider implementation returning the default context manually and specify the name of that. However, this is quite a bit of boilerplate and non-obvious.

Oberon00:
Related: #1819 "If there is only one ContextStorageProvider, the system property is ignored without warning"

dengliming:
What if we provide another System property name like( `io.opentelemetry.context.contextStorageProvider.spi.enable` or `io.opentelemetry.context.contextStorageProvider.enforceDefault`)  to enforce it (not `io.opentelemetry.context.contextStorageProvider`) using default context storage implemenation?  Is that more obvious?

anuraaga:
I think one variable with `default` sounds fine to me. It prevents the use of `default.java` as a provider, but that's a very small limitation. Instead we get just one property we have to explain.

dengliming:
@anuraaga Hey~ So you mean we can just accept `default `  for the io.opentelemetry.context.contextStorageProvider system property?

PR #2039, opened 2020-11-07
Add a way to enforce default context storage implementation

dengliming:
closes [https://github.com/open-telemetry/opentelemetry-java/issues/1818](https://github.com/open-telemetry/opentelemetry-java/issues/1818)

### DIFF
--- a/context/src/main/java/io/opentelemetry/context/LazyStorage.java
+++ b/context/src/main/java/io/opentelemetry/context/LazyStorage.java
@@ -87,7 +87,7 @@ final class LazyStorage {
     String providerClassName = System.getProperty(CONTEXT_STORAGE_PROVIDER_PROPERTY, "");
     // Allow user to enforce default ThreadLocalContextStorage
     if (ENFORCE_DEFAULT_VALUE.equals(providerClassName)) {
-      return DefaultContext.threadLocalStorage();
+      return ContextStorage.defaultStorage();
     }
 
     List<ContextStorageProvider> providers = new ArrayList<>();
@@ -118,7 +118,7 @@ final class LazyStorage {
                   + "qualified class name of the provider to use. Falling back to default "
                   + "ContextStorage. Found providers: "
                   + providers));
-      return DefaultContext.threadLocalStorage();
+      return ContextStorage.defaultStorage();
     }
 
     for (ContextStorageProvider provider : providers) {
@@ -134,7 +134,7 @@ final class LazyStorage {
                 + providerClassName
                 + " but found providers: "
                 + providers));
-    return DefaultContext.threadLocalStorage();
+    return ContextStorage.defaultStorage();
   }
 
   private LazyStorage() {}
--- a/context/src/main/java/io/opentelemetry/context/LazyStorage.java
+++ b/context/src/main/java/io/opentelemetry/context/LazyStorage.java
@@ -65,8 +65,7 @@ final class LazyStorage {
 
   private static final String CONTEXT_STORAGE_PROVIDER_PROPERTY =
       "io.opentelemetry.context.contextStorageProvider";
-  private static final String CONTEXT_STORAGE_PROVIDER_ENFORCE_DEFAULT_PROPERTY =
-      "io.opentelemetry.context.contextStorageProvider.enforceDefault";
+  private static final String ENFORCE_DEFAULT_VALUE = "default";
 
   private static final Logger logger = Logger.getLogger(LazyStorage.class.getName());
 
@@ -85,8 +84,8 @@ final class LazyStorage {
 
   static ContextStorage createStorage(AtomicReference<Throwable> deferredStorageFailure) {
     String providerClassName = System.getProperty(CONTEXT_STORAGE_PROVIDER_PROPERTY, "");
-    boolean enforceDefault = Boolean.getBoolean(CONTEXT_STORAGE_PROVIDER_ENFORCE_DEFAULT_PROPERTY);
-    if (enforceDefault) {
+    // enforceDefault
+    if (ENFORCE_DEFAULT_VALUE.equals(providerClassName)) {
       return DefaultContext.threadLocalStorage();
     }
 
--- a/context/src/main/java/io/opentelemetry/context/LazyStorage.java
+++ b/context/src/main/java/io/opentelemetry/context/LazyStorage.java
@@ -65,7 +65,7 @@ final class LazyStorage {
 
   private static final String CONTEXT_STORAGE_PROVIDER_PROPERTY =
       "io.opentelemetry.context.contextStorageProvider";
-  private static final String ENFORCE_DEFAULT_VALUE = "default";
+  private static final String ENFORCE_DEFAULT_STORAGE_VALUE = "default";
 
   private static final Logger logger = Logger.getLogger(LazyStorage.class.getName());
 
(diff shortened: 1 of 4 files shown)


context/src/main/java/io/opentelemetry/context/LazyStorage.java
```diff
@@ -83,6 +85,10 @@ static ContextStorage get() {
 
   static ContextStorage createStorage(AtomicReference<Throwable> deferredStorageFailure) {
     String providerClassName = System.getProperty(CONTEXT_STORAGE_PROVIDER_PROPERTY, "");
+    boolean enforceDefault = Boolean.getBoolean(CONTEXT_STORAGE_PROVIDER_ENFORCE_DEFAULT_PROPERTY);
```
Oberon00, line 88: Why a second system property? As discussed on the issue, a single one would be simpler.

dengliming: My bad. It's my misunderstand your point. Can you review again? thanks.

context/src/test/java/io/opentelemetry/context/LazyStorageTest.java
```diff
@@ -76,6 +76,13 @@ public void set_storage_provider_property_matches_one_provides() throws Exceptio
     }
   }
 
+  @Test
+  public void enforce_default_thread_local_storage() {
```
anuraaga, line 80: I don't think we can test this sort of behavior with other ones since the system property is only loaded once. Can you extract another testset for this case? I think `context` will be full of many test sets because of his global initialization logic :)

anuraaga (approved):

context/src/test/java/io/opentelemetry/context/LazyStorageTest.java
anuraaga: Actually since we're not testing the public method it looks ok. But we should still make sure the system properties are cleaned up. I realize this already affects the existing tests, but can you use junit pioneer?

https://github.com/open-telemetry/opentelemetry-java/blob/71ba8e1b87bfd16f2ecd21643bf55b88d1ecfcaa/sdk/common/src/test/java/io/opentelemetry/sdk/common/export/ConfigBuilderTest.java#L238

dengliming: I already clearup the properties in aftereach. It is not enough?
```
@AfterEach
  public void after() {
    System.clearProperty(CONTEXT_STORAGE_PROVIDER_PROPERTY);
  }
```

anuraaga: Using the annotation is more self-documenting and I guess less error-prone. It's not so clear which tests depend on that `AfterEach` existing, so the annotation is nice to make it very self-documenting / clear. It's not a blocking change but I think it's a good idea.

dengliming: You are right. Thanks for help. That makes it more simpler and clearer. : )

context/src/test/java/io/opentelemetry/context/LazyStorageTest.java
```diff
@@ -13,37 +13,36 @@
...
 
-  @AfterEach
-  public void after() {
-    System.clearProperty(CONTEXT_STORAGE_PROVIDER_PROPERTY);
-  }
-
   @Test
-  public void empty_provides() {
+  @ClearSystemProperty(key = CONTEXT_STORAGE_PROVIDER_PROPERTY)
+  public void empty_providers() {
```
anuraaga, line 31: Just one last nit, if you can change the methods to be package-private, since junit5, that would be swell, not a required change though.

anuraaga (approved):
Thanks!

context/src/test/java/io/opentelemetry/context/LazyStorageTest.java
dengliming: Thanks. That was done.

context/src/main/java/io/opentelemetry/context/LazyStorage.java
```diff
@@ -83,6 +84,10 @@ static ContextStorage get() {
 
   static ContextStorage createStorage(AtomicReference<Throwable> deferredStorageFailure) {
     String providerClassName = System.getProperty(CONTEXT_STORAGE_PROVIDER_PROPERTY, "");
```
jkwatson, line 86: shouldn't we change the default value here to ENFORCE_DEFAULT_VALUE, rather than an empty String?

jkwatson: oh, probably not, since we want the SPI to load before we fall back to this value. hmm. This code has ended up pretty confusing. Maybe some additional comments about the priorities here would be helpful.

context/src/main/java/io/opentelemetry/context/LazyStorage.java
```diff
@@ -83,6 +84,10 @@ static ContextStorage get() {
...
+    // enforceDefault
+    if (ENFORCE_DEFAULT_VALUE.equals(providerClassName)) {
+      return DefaultContext.threadLocalStorage();
```
jkwatson, line 89: How are we choosing between calling this method, vs. `ContextStorage.defaultStorage()` which return the same thing?

dengliming: yes. they return the same thing. what about calling `ContextStorage.defaultStorage()` consistently?

dengliming: @anuraaga I would really appreciate it If you can also share your opinion here.  : )

anuraaga: Yeah let's remove `DefaultContext.threadLocalStorage`, that is a leftover from refactoring during the initial implementation. I was thinking of doing it after this PR if you weren't to do so :P

dengliming: @jkwatson Should we remove `DefaultContext.threadLocalStorage`  in this PR?

jkwatson: If we don't, please create an issue for it, so we don't lose track.

dengliming: Let's remove it now. : )

jkwatson (approved):
Thanks!

context/src/main/java/io/opentelemetry/context/LazyStorage.java
```diff
@@ -65,6 +65,7 @@ static ContextStorage get() {
 
   private static final String CONTEXT_STORAGE_PROVIDER_PROPERTY =
       "io.opentelemetry.context.contextStorageProvider";
+  private static final String ENFORCE_DEFAULT_VALUE = "default";
```
Oberon00, line 68:
```suggestion
  private static final String ENFORCE_DEFAULT_STORAGE_VALUE = "default";
```

context/src/main/java/io/opentelemetry/context/LazyStorage.java
```diff
@@ -82,7 +83,12 @@ static ContextStorage get() {
   }
 
   static ContextStorage createStorage(AtomicReference<Throwable> deferredStorageFailure) {
+    // Get the specified SPI implementation first here
     String providerClassName = System.getProperty(CONTEXT_STORAGE_PROVIDER_PROPERTY, "");
```
Oberon00, line 87:
```suggestion
    final String providerClassName = System.getProperty(CONTEXT_STORAGE_PROVIDER_PROPERTY, "");
```

Oberon00 (approved):
Thank you!

PR #2044, opened 2020-11-09
Allow registering ContextStorage wrappers with code to avoid requirin…

### LINKED ISSUE #922: Context change interceptor

anuraaga:
…g SPI for listening to events.

@sfriberg reached out in wanting to be able to listen `ContextStorage` events without overriding the detected storage mechanism, and I think being able to register listeners with code is important in any case. It was the first feature request we got in Armeria after adding the `ContextStorage` SPI since SPI is too tedious to just do something like pushing MDC.

Fixes #922

### DIFF
--- a/context/src/main/java/io/opentelemetry/context/ContextStorageWrappers.java
+++ b/context/src/main/java/io/opentelemetry/context/ContextStorageWrappers.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.context;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Function;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Holder of functions to wrap the used {@link ContextStorage}. Separate class from {@link
+ * LazyStorage} to allow registering wrappers before initializing storage.
+ */
+final class ContextStorageWrappers {
+
+  private static final Logger log = Logger.getLogger(ContextStorageWrappers.class.getName());
+
+  private static boolean storageInitialized;
+
+  private static final List<Function<? super ContextStorage, ? extends ContextStorage>> wrappers =
+      new ArrayList<>();
+
+  static synchronized void addWrapper(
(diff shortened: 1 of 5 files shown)


context/src/main/java/io/opentelemetry/context/ContextStorageWrappers.java
```diff
@@ -0,0 +1,49 @@
...
+final class ContextStorageWrappers {
+
+  private static final Logger log = Logger.getLogger(ContextStorageWrappers.class.getName());
+
+  private static boolean storageInitialized;
+
+  private static final List<Function<? super ContextStorage, ? extends ContextStorage>> wrappers =
+      new ArrayList<>();
+
+  static synchronized void addWrapper(
```
jkwatson, line 27: It probably doesn't matter too much, since this class isn't public, but exposing synchronized methods is generally a bad practice, as it could allow accidental deadlocking if someone synchronizes on this class.

anuraaga: Yeah figured it's ok due to the privateness but went with consistency.

jkwatson (approved):
Looks good to me.

sfriberg:
Thanks for doing this @anuraaga I think this looks like something that would be easy to use.

Would it make sense to have to be able to register a SPI similar to what is used for the contextstorage and thus instead of manually coding things the ContextStorage init code would scan the class/module path for any classes implementing the ContextStorageWrapper and add those?

jkwatson:
> Thanks for doing this @anuraaga I think this looks like something that would be easy to use.
> 
> Would it make sense to have to be able to register a SPI similar to what is used for the contextstorage and thus instead of manually coding things the ContextStorage init code would scan the class/module path for any classes implementing the ContextStorageWrapper and add those?

@sfriberg can you log an issue for this, if you think it's a good idea yourself? Thanks!

PR #2045, opened 2020-11-09
Link to the JARs on Mvn Central

holyjak:
because I could not find them easily

### DIFF
--- a/opentracing-shim/README.md
+++ b/opentracing-shim/README.md
@@ -1,7 +1,9 @@
 [![Javadocs][javadoc-image]][javadoc-url]
-# OpenTelemetry - OpenTracing Shim
+# OpenTelemetry - OpenTracing Shim (Bridge)
 The OpenTracing shim is a bridge layer from OpenTelemetry to the OpenTracing API.
 It takes OpenTelemetry Tracer and exposes it as an implementation of an OpenTracing Tracer.
+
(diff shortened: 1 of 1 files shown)


opentracing-shim/README.md
```diff
@@ -1,8 +1,10 @@
 [![Javadocs][javadoc-image]][javadoc-url]
-# OpenTelemetry - OpenTracing Shim
+# OpenTelemetry - OpenTracing Shim (Bridge)
```
jkwatson, line 2: I'm not totally opposed to adding the parenthetical, but it's not the naming that we've been using on the project. Do you think this makes things more clear or discoverable?

anuraaga: Yeah agree with @jkwatson - Shim is already a bridge so it seems like stuttering, I'd leave it out.

jkwatson, line 2:
```suggestion
# OpenTelemetry - OpenTracing Shim
```

jkwatson:
@holyjak thanks for the contribution. Looks like there are a couple of comments to take a look at, and you'll need to sign the CLA before we can get it merged. Thanks!

opentracing-shim/README.md
breedx-splk: Also worth pointing out that "bridge" has a different meaning in the instrumentation/agent project, so I'd avoid that here.

jkwatson:
closing this, because we cannot merge without the CLA being signed. I created an issue to update our docs before release: #2393

PR #2049, opened 2020-11-09
Switch the key/value pair sorting to mergesort, because it is stable

### LINKED ISSUE #2043: Setting a duplicate attribute doesn't always overwrite the previous value
In particular, the test below fails when added to `AttributesTest`:

```
  @Test
  void deduplication() {
    Attributes one =
        Attributes.builder()
            .put(stringKey("key2"), "valueX")
            .put(stringKey("key2"), "value2")
            .put(stringKey("key1"), "value1")
            .build();
    Attributes two =
        Attributes.builder()
            .put(stringKey("key2"), "value2")
            .put(stringKey("key1"), "value1")
            .build();

    assertThat(one).isEqualTo(two);
  }
```

jkwatson:
Turns out quicksort isn't stable (equal-comparing keys don't stay in source-order).

Resolves #2043

### DIFF
--- a/api/src/main/java/io/opentelemetry/api/internal/ImmutableKeyValuePairs.java
+++ b/api/src/main/java/io/opentelemetry/api/internal/ImmutableKeyValuePairs.java
@@ -65,34 +65,72 @@ public abstract class ImmutableKeyValuePairs<K, V> {
     checkArgument(
         data.length % 2 == 0, "You must provide an even number of key/value pair arguments.");
 
-    quickSort(data, 0, data.length - 2);
+    mergeSort(data);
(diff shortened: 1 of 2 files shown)


anuraaga (approved):
Hoping we don't have to update the sorting algorithm every month :) But this is important since it fixes a bug

jkwatson:
> Hoping we don't have to update the sorting algorithm every month :) But this is important since it fixes a bug

Yes, I certainly wouldn't have touched this code if there weren't a bug to fix. I tried some other hacks on the input side, but eventually decided it was simpler just to have a proper, stable sort algorithm.

Issue #2047, opened 2020-11-09
Update the zipkin exporter Span Status handling


jkwatson:
There was a recent tweak to the spec, and I believe we have a bit of updating to do:  https://github.com/open-telemetry/opentelemetry-specification/pull/1186

PR #2050, opened 2020-11-10
Update the zipkin exporter Span Status handling

dengliming:
closes [https://github.com/open-telemetry/opentelemetry-java/issues/2047](https://github.com/open-telemetry/opentelemetry-java/issues/2047)

### DIFF
--- a/exporters/zipkin/src/main/java/io/opentelemetry/exporter/zipkin/ZipkinSpanExporter.java
+++ b/exporters/zipkin/src/main/java/io/opentelemetry/exporter/zipkin/ZipkinSpanExporter.java
@@ -14,7 +14,6 @@ import io.opentelemetry.api.common.AttributeType;
 import io.opentelemetry.api.common.ReadableAttributes;
 import io.opentelemetry.api.trace.Span.Kind;
 import io.opentelemetry.api.trace.SpanId;
-import io.opentelemetry.api.trace.attributes.SemanticAttributes;
 import io.opentelemetry.sdk.common.CompletableResultCode;
 import io.opentelemetry.sdk.common.InstrumentationLibraryInfo;
 import io.opentelemetry.sdk.common.export.ConfigBuilder;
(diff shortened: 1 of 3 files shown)


exporters/zipkin/src/main/java/io/opentelemetry/exporter/zipkin/ZipkinSpanExporter.java
```diff
@@ -141,7 +141,9 @@ static Span generateSpan(SpanData spanData, Endpoint localEndpoint) {
         });
     SpanData.Status status = spanData.getStatus();
     // for GRPC spans, include status code & description.
-    if (status != null && spanAttributes.get(SemanticAttributes.RPC_SERVICE) != null) {
+    if (status != null
```
anuraaga, line 144: Don't believe status can actually be null, if it can we should fix it to return unset in such cases in a separate PR.

exporters/zipkin/src/main/java/io/opentelemetry/exporter/zipkin/ZipkinSpanExporter.java
```diff
@@ -141,7 +141,9 @@ static Span generateSpan(SpanData spanData, Endpoint localEndpoint) {
...
+        && !status.isUnset()
+        && spanAttributes.get(SemanticAttributes.RPC_SERVICE) != null) {
```
anuraaga, line 146: I'm reading the attached issue https://github.com/open-telemetry/opentelemetry-specification/pull/1186/files and can't find anything about this check, can you point me at it? IIUC we're supposed to set status even if not rpc.

dengliming: > can't find anything about this check

So do I 
Can you help to confirm this? @Oberon00

jkwatson: This was definitely a holdover from when the statuses were so closely aligned with the GRPC status codes. Should be fine, I would think, to remove it. 👍

exporters/zipkin/src/main/java/io/opentelemetry/exporter/zipkin/ZipkinSpanExporter.java
```diff
@@ -140,8 +139,8 @@ static Span generateSpan(SpanData spanData, Endpoint localEndpoint) {
           }
         });
     SpanData.Status status = spanData.getStatus();
-    // for GRPC spans, include status code & description.
-    if (status != null && spanAttributes.get(SemanticAttributes.RPC_SERVICE) != null) {
+    // include status code & description.
+    if (status != null && !status.isUnset()) {
```
jkwatson, line 143: yes, status cannot be null here, so this extra check can be removed.

exporters/zipkin/src/main/java/io/opentelemetry/exporter/zipkin/ZipkinSpanExporter.java
dengliming: Thanks. It was removed.

jkwatson (approved):
Thanks!

PR #2051, opened 2020-11-10
Add assertj helpers for asserting traces and expose from the junit5 h…

anuraaga:
…elper.

I think this makes it very nice to assert on traces in tests. There's some future work though that requiring ordering doesn't seem very helpful here, especially for `hasTracesSatisfying`. Leaving it for now to not complicate this change too much.

### DIFF
--- a/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/TraceAssert.java
+++ b/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/TraceAssert.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.testing.assertj;
+
+import io.opentelemetry.sdk.trace.data.SpanData;
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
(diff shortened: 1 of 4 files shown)


sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/TraceAssert.java
```diff
@@ -0,0 +1,64 @@
...
+    return this;
+  }
+
+  /**
+   * Asserts that the trace under assertion has the same number of spans as provided {@code
+   * assertions} and executes each {@link SpanDataAssert} in {@code assertions} in order with the
+   * corresponding span.
+   */
+  @SafeVarargs
+  @SuppressWarnings("varargs")
```
anuraaga, line 46: Since it's a testing helper, I think we can assume user's won't try to trick the compiler with a bad array.

jkwatson: I think you mean: Users will certainly try to trick the compiler with a bad array.  😆

jkwatson (approved):
Seems useful. We should start converting our own tests to use these assertions.

jkwatson:
@anuraaga I see you're still amending this PR. Feel free to merge it, if/when you think it's ready to go.

anuraaga:
Yup was missing some test cases but thing I got the non-defensive oens

Issue #2013, opened 2020-11-05
Alternative OpenCensus shim implementation proposal


bogdandrutu:
The current OpenCensus shim is implemented here:
https://github.com/open-telemetry/opentelemetry-java/tree/master/opencensus-shim

This implementation has some disadvantages regarding the requirement to use a cache of Spans, spans need to be ended in a certain amount of time, etc. Right now the current approach is to rely on the opencensus-impl to do the work of accumulating events into Spans, export, etc.

**First Proposal**
The proposal is to make the shim rely on the opentelemetry-sdk to do the heavy work of accumulating events into Spans, export them, etc.

This way you will benefit of the fact that this project is currently very active and maintained, and also I believe this will simplify the implementation, here are some thoughts:
1. The OpenCensus.Tracer will be a simple wrapper of the OpenTelemetry.Tracer with a name "opencensus".
1. The `PropagationComponent` can be implemented by wrapping the propagators from OpenTelemetry (w3c, b3). One small problem is that you will need to re-implement the `BinaryFormat` (this is most likely a duplicate code).
1. The clock does not need to be re-implemented. just return the Millis clock.
1. The `ExporterComponent` can be implemented by wrapping the BatchSpanProcessor, the OpenCensus.SpanExporter.Handler is equivalent with the OpenTelemetry.SpanExporter
1. The `TraceConfig` is a simple wrapper of the config management in the OpenTelemetry.

**Second Proposal**
Take a dependency on otel api in the opencensus API and redirect all calls. This may not be possible until GA.

bogdandrutu:
Also the ContextManager will redirect into the OpenTelemetry.Context which we will need to configure to be using as otelInGRPC, because OpenCensus users delegate the context propagation to gRPC Context.

bogdandrutu:
Alternative to for example wrapping the ExporterComponent is to tell users to configure an OTel exporter. This can simplify the implementation.

nilebox:
> Alternative to for example wrapping the ExporterComponent is to tell users to configure an OTel exporter. 

This is what we'd prefer I think, as this helps to avoid duplicate OC and OT spans being exported in case both OC and OT pipelines are configured in the application.

Current shim implementation configures OC Tracer with no-op exporter, so we can probably do the same.

bogdandrutu:
We may still want to offer user an alternative to the RunningSpanStore option which can be extracted from the current OTel zpages and offer as a SpanProcessor.

PR #2060, opened 2020-11-11
Alternative OpenCensus shim implementation

zoercai:
This implements the alternative shim approach suggested at https://github.com/open-telemetry/opentelemetry-java/issues/2013:

✅  The OpenCensus.Tracer will be a simple wrapper of the OpenTelemetry.Tracer with a name "opencensus".
✅  The PropagationComponent can be implemented by wrapping the propagators from OpenTelemetry (w3c, b3). One small problem is that you will need to re-implement the BinaryFormat (this is most likely a duplicate code).
✅  The clock does not need to be re-implemented. just return the Millis clock.
☑️  The ExporterComponent has not been implemented as we are dropping all OpenCensus spans and only exporting OpenTelemetry spans.
☑️ The TraceConfig seems unlikely to be used, so leaving it unmapped here unless someone actually asks for it.


zoercai:
I can't assign reviewers so tagging you here:
@bogdandrutu @nilebox @james-bebbington @anuraaga @jkwatson

anuraaga:
I remember there being concerns about using the OTel span directly, were they addressed?

https://github.com/open-telemetry/opentelemetry-java/pull/1863#discussion_r513833590

nilebox:
> I remember there being concerns about using the OTel span directly, were they addressed?

If I understand correctly, they were addressed by copy-pasting some parts of the original implementation to this PR.
If you're fine with maintaining this "forked" code, that should be okay.

Also, some OpenCensus plugins may not work correctly if they rely on the OC implementation details, but that is also probably fine since with this shim we only care about supporting "libraries instrumented using OpenCensus API" and not "pipelines using OpenCensus to export data", as we switch to OpenTelemetry pipeline here.

In other words, there might be side effects of using this shim due to difference in `OC impl` vs `OC shim + OT impl`, but that is by design.

anuraaga:
Thanks for clarifying - yeah I'm a fan of the approach. Let me know if you'd like a detailed look at the code now or later :)

nilebox:
@anuraaga Another clever workaround to note in this PR is the OC/OT span hybrid which **extends** OC Span while **implementing** OT Span:
```java
public class OpenTelemetrySpanImpl extends Span implements io.opentelemetry.api.trace.Span
```
In https://github.com/open-telemetry/opentelemetry-java/pull/1863#discussion_r513833590 we discussed the possibility of implementing the OC Span interface which wasn't an option but since the OT Span is an interface that works too.

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryNoRecordEventsSpanImpl.java
```diff
@@ -0,0 +1,68 @@
+package io.opentelemetry.opencensusshim;
```
nilebox, line 1: This file is missing a licence header, and I'm actually not sure what should we put here, since parts of the code were copied from OpenCensus IIUC?

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryNoRecordEventsSpanImpl.java
```diff
@@ -0,0 +1,68 @@
...
+
+  @Override
+  public void addAnnotation(Annotation annotation) {
+    Preconditions.checkNotNull(annotation, "annotation");
+  }
+
+  @Override
+  public void putAttribute(String key, AttributeValue value) {
+    Preconditions.checkNotNull(key, "key");
+    Preconditions.checkNotNull(value, "value");
```
nilebox, line 37: nit: Do we even need to perform these checks if the implementation is a no-op?

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryNoRecordEventsSpanImpl.java
anuraaga: We have this pattern for including two licenses, yeah in any file with copies let's use it

https://github.com/open-telemetry/opentelemetry-java/blob/master/context/src/main/java/io/opentelemetry/context/Context.java#L6

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryNoRecordEventsSpanImpl.java
jkwatson: This seems like less of a nit, and more of a problem. If it throws exceptions on bad inputs, it's not a no-op, I think.

zoercai: I actually just copied all of this from https://github.com/census-instrumentation/opencensus-java/blob/f0e78ba5a10165cce2b63809bfdfe3e34b1eb456/impl_core/src/main/java/io/opencensus/implcore/trace/NoRecordEventsSpanImpl.java
Not sure why there are checks in the OC implementation, happy to remove the checks here though if it makes more sense

jkwatson: If OC users are fine with that behavior, then it's cool. We've taken a different approach in OTel, but this is fine if it's what users are used to.

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanImpl.java
```diff
@@ -0,0 +1,286 @@
...
+   * @param context the context associated with this {@code Span}.
+   * @param options the options associated with this {@code Span}. If {@code null} then default
+   *     options will be set.
+   * @param timestampConverter time stamp converter
+   * @throws NullPointerException if context is {@code null}.
+   * @throws IllegalArgumentException if the {@code SpanContext} is sampled but no RECORD_EVENTS
+   *     options.
+   * @since 0.5
+   */
+  protected OpenTelemetrySpanImpl(
```
bogdandrutu, line 80: Why protected?

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanBuilderImpl.java
```diff
@@ -0,0 +1,267 @@
...
+import io.opencensus.trace.TraceOptions;
+import io.opencensus.trace.Tracestate;
+import io.opencensus.trace.config.TraceConfig;
+import io.opencensus.trace.config.TraceParams;
+import java.util.Collections;
+import java.util.List;
+import java.util.Random;
+import javax.annotation.Nullable;
+
+public class OpenTelemetrySpanBuilderImpl extends SpanBuilder {
```
bogdandrutu, line 48: Maybe I miss something, but I would see this simpler than this:

public class OpenTelemetrySpanBuilderImpl extends SpanBuilder {
  private final otel.Builder builder;

  ... // All methods redirect to the otel builder
}

public class OpenCensusSpanWrapper extends oc.Span {
  private final otel.ReadableSpan;
  private final oc.SpanContext;  // probably worth constructing this once not when getContext is called.

  ... // All methods redirect to the 
}


Can you tell me if this is possible?

bogdandrutu: Context Interaction:
* a wrapped.Span is added to the OC context, we add the "wrapped span" to the Otel Context
* a blank oc Span is added to the OC context, we add a "blank span" to the Otel Context
* a otel.Span is added to the Otel Context, we create a wrapper and add it to OC Context.

zoercai: @bogdandrutu yep I've updated everything now, could you please take another look?

bogdandrutu: I am not sure this is what I meant. I was trying to suggest to use composition when implementing the OC required types and delegate everything to otel.

zoercai: I'm quite confused. I'm delegating everything to oTel after creating the oTel span here: https://github.com/open-telemetry/opentelemetry-java/pull/2060/files#diff-4b913d64466c48a3f3149a7e7c9345886e02931d69396c8223fc855818f3e261R166-R167

Could you explain what you were expecting a bit more?

zoercai: I've moved some of the logic from `startSpan()` to the methods overriding OC methods, such as `addParentLinks()` and `setSpanKind()` to set Otel fields directly. The rest depend on the sampler, which does all of https://github.com/open-telemetry/opentelemetry-java/pull/2060/files#diff-4b913d64466c48a3f3149a7e7c9345886e02931d69396c8223fc855818f3e261R109-R146 to determine the kind of span that should be created. Since the sampler takes in OC parameters (parent, hasRemoteParent, traceId, spanId, name, parentLinks), I don't think I can map and delegate this nor its parameters to Otel, especially since the sampler could be implemented by the user.

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryB3FormatImpl.java
```diff
@@ -0,0 +1,58 @@
...
+  @Override
+  public <C> SpanContext extract(C carrier, Getter<C> getter) {
+    Context context =
+        OTEL_B3_PROPAGATOR.extract(
+            Context.current(),
+            carrier,
+            new TextMapPropagator.Getter<C>() {
+              @Override
+              public Iterable<String> keys(C carrier) {
+                return new ArrayList<>();
```
anuraaga, line 46:
```suggestion
                // OC Getter cannot return keys for an object, but users should not need it either.
                return Collections.emptyList();
```

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryB3FormatImpl.java
```diff
@@ -0,0 +1,58 @@
...
+import io.opencensus.trace.propagation.TextFormat;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.propagation.TextMapPropagator;
+import io.opentelemetry.extension.trace.propagation.B3Propagator;
+import java.util.ArrayList;
+import java.util.List;
+import javax.annotation.Nullable;
+
+public class OpenTelemetryB3FormatImpl extends TextFormat {
```
anuraaga, line 20:
```suggestion
class OpenTelemetryB3FormatImpl extends TextFormat {
```

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryBinaryFormatImpl.java
```diff
@@ -0,0 +1,98 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import io.opencensus.internal.Utils;
+import io.opencensus.trace.SpanContext;
+import io.opencensus.trace.propagation.BinaryFormat;
+import io.opentelemetry.api.trace.SpanId;
+import io.opentelemetry.api.trace.TraceFlags;
+import io.opentelemetry.api.trace.TraceId;
+import io.opentelemetry.api.trace.TraceState;
+
+public class OpenTelemetryBinaryFormatImpl extends BinaryFormat {
```
anuraaga, line 17:
```suggestion
class OpenTelemetryBinaryFormatImpl extends BinaryFormat {
```

anuraaga: Add a comment something like `OpenCensus's code for BinaryFormat is in the implementation, not API. Since we don't want to depend on the implementation for shim users, we reimplement it here.`

Also, let's just copy https://github.com/census-instrumentation/opencensus-java/blob/0a2f5799a2809b84e4c0da4b2cb743b798ca0fc8/impl_core/src/main/java/io/opencensus/implcore/trace/propagation/BinaryFormatImpl.java 100% - there seems to be no benefit in mapping into the OTel types here

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryB3FormatImpl.java
anuraaga: Also I think this should be `OpenTelemetryTextFormat` and it can have a constructor to accept any `TextMapPropagator` instead of being hardcoded to b3. Presumably we want to add w3c, or basically any propagator a user wants to configure.

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryContextManager.java
```diff
@@ -21,11 +21,8 @@
 
   private static final Logger LOGGER =
       Logger.getLogger(OpenTelemetryContextManager.class.getName());
-  private final SpanCache spanCache;
 
-  public OpenTelemetryContextManager() {
-    this.spanCache = SpanCache.getInstance();
```
anuraaga, line 27: 🎉

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryNoRecordEventsSpanImpl.java
```diff
@@ -0,0 +1,183 @@
...
+import io.opencensus.trace.Status;
+import io.opentelemetry.api.common.AttributeKey;
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.trace.StatusCode;
+import java.util.EnumSet;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nonnull;
+
+public class OpenTelemetryNoRecordEventsSpanImpl extends Span
```
anuraaga, line 42:
```suggestion
class OpenTelemetryNoRecordEventsSpanImpl extends Span
```

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryNoRecordEventsSpanImpl.java
```diff
@@ -0,0 +1,183 @@
...
+  public void end(EndSpanOptions options) {
+    Preconditions.checkNotNull(options, "options");
+  }
+
+  @Override
+  public void end(long timestamp, TimeUnit unit) {
+    // do nothing
+  }
+
+  private OpenTelemetryNoRecordEventsSpanImpl(SpanContext context) {
```
anuraaga, line 108: Let's go ahead and put the constructor above the methods

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryNoRecordEventsSpanImpl.java
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  @Override
+  public io.opentelemetry.api.trace.Span setStatus(StatusCode canonicalCode) {
+    return null;
+  }
+
+  @Override
+  public io.opentelemetry.api.trace.Span setStatus(StatusCode canonicalCode, String description) {
+    return null;
```
anuraaga, line 95: `return this` instead of `return null` throughout

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryNoRecordEventsSpanImpl.java
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  @Override
+  public io.opentelemetry.api.trace.Span updateName(String name) {
+    return null;
+  }
+
+  @Override
+  public io.opentelemetry.api.trace.SpanContext getSpanContext() {
+    return io.opentelemetry.api.trace.SpanContext.getInvalid();
```
anuraaga, line 176: Shouldn't it be the mapped context, not invalid? Is `SpanContext` guaranteed to be invalid in the constructor? I'm guessing no since in that case we wouldn't even need a constructor parameter for it.

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryPropagationComponentImpl.java
```diff
@@ -0,0 +1,32 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.opencensusshim;
+
+import io.opencensus.trace.propagation.BinaryFormat;
+import io.opencensus.trace.propagation.PropagationComponent;
+import io.opencensus.trace.propagation.TextFormat;
+
+public class OpenTelemetryPropagationComponentImpl extends PropagationComponent {
```
anuraaga, line 12:
```suggestion 
class OpenTelemetryPropagationComponentImpl extends PropagationComponent {
```

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanBuilderImpl.java
```diff
@@ -0,0 +1,243 @@
...
+import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.context.Context;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Random;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
+
+public class OpenTelemetrySpanBuilderImpl extends SpanBuilder {
```
anuraaga, line 53:
```suggestion
class OpenTelemetrySpanBuilderImpl extends SpanBuilder {
```

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanBuilderImpl.java
```diff
@@ -0,0 +1,243 @@
...
+  private static final Tracer TRACER =
+      OpenTelemetry.getGlobalTracer("io.opencensus.opentelemetry.migration");
```
anuraaga, line 55:
```suggestion
      OpenTelemetry.getGlobalTracer("io.opentelemetry.opencensusshim");
```

Unless a specific reason for the other name, we want this name to be the name of the instrumentation library, not the instrumented one, so I think it's the shim in this case.

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanBuilderImpl.java
```diff
@@ -0,0 +1,243 @@
...
+  private static final Tracestate TRACESTATE_DEFAULT = Tracestate.builder().build();
```
anuraaga, line 56: I think we need to prefix almost every variable in this file with either `oc` or `otel`, quite hard to read right now

anuraaga: Or at least otel for all otel (e.g., `OTEL_TRACER`)

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanBuilderImpl.java
anuraaga: @bogdandrutu If I'm not mistaken, we can't eagerly delegate to OTel because it doesn't have `setRecordEvents` or `setSampler`. Having `setSampler` on OTel's `SpanBuilder` seems useful to me but guess it's a spec change. `setRecordEvents` we shouldn't need since pretty sure that can just be implemented as `setSampler(ALWAYS_OFF)`. If we could do it, then all the logic in this class goes away I think (would need `mapSampler` to shim the sampler)

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanBuilderImpl.java
```diff
@@ -0,0 +1,243 @@
...
+  @Override
+  public Span startSpan() {
+    // To determine whether to sample this span
+    TraceParams activeTraceParams = options.traceConfig.getActiveTraceParams();
+    Random random = options.randomHandler.current();
+    TraceId traceId;
+    SpanId spanId = SpanId.generateRandomId(random);
+    Tracestate tracestate = TRACESTATE_DEFAULT;
+    SpanContext parentContext = null;
+    Boolean hasRemoteParent = null;
```
anuraaga, line 116:
```suggestion
    boolean hasRemoteParent = false;
```

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanBuilderImpl.java
```diff
@@ -0,0 +1,243 @@
...
+    if (remoteParentSpanContext != null && remoteParentSpanContext.isValid()) {
+      parentContext = remoteParentSpanContext;
+      hasRemoteParent = Boolean.TRUE;
```
anuraaga, line 119:
```suggestion
      hasRemoteParent = true;
```

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanBuilderImpl.java
```diff
@@ -0,0 +1,243 @@
...
+      traceId = parentContext.getTraceId();
+      tracestate = parentContext.getTracestate();
+    } else if (parent != null && parent.getContext().isValid()) {
+      parentContext = parent.getContext();
+      hasRemoteParent = Boolean.FALSE;
```
anuraaga, line 124:
```suggestion
```

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanBuilderImpl.java
anuraaga: I think

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanImpl.java
```diff
@@ -0,0 +1,238 @@
...
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.common.AttributesBuilder;
+import io.opentelemetry.api.trace.StatusCode;
+import java.util.EnumSet;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Logger;
+import javax.annotation.Nonnull;
+
+public class OpenTelemetrySpanImpl extends Span implements io.opentelemetry.api.trace.Span {
```
anuraaga, line 56:
```suggestion
class OpenTelemetrySpanImpl extends Span implements io.opentelemetry.api.trace.Span {
```

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanImpl.java
```diff
@@ -0,0 +1,238 @@
...
+  private static final Logger LOGGER = Logger.getLogger(OpenTelemetrySpanImpl.class.getName());
+  private static final EnumSet<Span.Options> RECORD_EVENTS_SPAN_OPTIONS =
+      EnumSet.of(Span.Options.RECORD_EVENTS);
+
+  private final io.opentelemetry.api.trace.Span otSpan;
```
anuraaga, line 61:
```suggestion
  private final io.opentelemetry.api.trace.Span otelSpan;
```

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanImpl.java
```diff
@@ -0,0 +1,238 @@
...
+
+  OpenTelemetrySpanImpl(io.opentelemetry.api.trace.Span otSpan) {
```
anuraaga, line 63:
```suggestion
  OpenTelemetrySpanImpl(io.opentelemetry.api.trace.Span otelSpan) {
```

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanImpl.java
```diff
@@ -0,0 +1,238 @@
...
+    super(
+        SpanContext.create(
```
anuraaga, line 65: Think you an use `mapSpanContext`

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanImpl.java
```diff
@@ -0,0 +1,238 @@
...
+
+  @Override
+  public void setStatus(Status status) {
+    Preconditions.checkNotNull(status, "status");
+    otSpan.setStatus(status.isOk() ? StatusCode.OK : StatusCode.ERROR);
+  }
+
+  @Override
+  public io.opentelemetry.api.trace.Span setStatus(StatusCode canonicalCode, String description) {
+    return null;
```
anuraaga, line 150:
```suggestion
    otSpan.setStatus(canonicalCode, description);
    return this;
```

Don't we need to delegate to the OTel span for all of these methods too since they might be called from OTel instrumentation?

Also make sure to `return this` everywhere

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryTraceContextFormatImpl.java
```diff
@@ -0,0 +1,57 @@
...
+import io.opencensus.trace.propagation.TextFormat;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.propagation.W3CTraceContextPropagator;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.propagation.TextMapPropagator;
+import java.util.ArrayList;
+import java.util.List;
+import javax.annotation.Nullable;
+
+public class OpenTelemetryTraceContextFormatImpl extends TextFormat {
```
anuraaga, line 20: Per my above comment, you should be able to have just one class for both b3 and w3c

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryTracerImpl.java
```diff
@@ -0,0 +1,54 @@
...
+import io.opencensus.implcore.trace.internal.RandomHandler;
+import io.opencensus.trace.Span;
+import io.opencensus.trace.SpanBuilder;
+import io.opencensus.trace.SpanContext;
+import io.opencensus.trace.Tracer;
+import io.opencensus.trace.config.TraceConfig;
+import io.opentelemetry.opencensusshim.OpenTelemetrySpanBuilderImpl.Options;
+import javax.annotation.Nullable;
+
+public class OpenTelemetryTracerImpl extends Tracer {
```
anuraaga, line 35:
```suggestion
class OpenTelemetryTracerImpl extends Tracer {
```

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/SpanConverter.java
```diff
@@ -50,6 +41,12 @@ protected FakeSpan(SpanContext context) {
       super(context, RECORD_EVENTS_SPAN_OPTIONS);
     }
 
+    @Override
```
anuraaga, line 44: Do we still need FakeSpan? Can't we use one of the new Span implementations where this is being used?

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/SpanConverter.java
```diff
@@ -69,64 +66,12 @@ public void end(EndSpanOptions options) {}
...
-                        arg -> null));
-        builder.addLink(
-            io.opentelemetry.api.trace.SpanContext.create(
-                TraceId.bytesToHex(link.getTraceId().getBytes()),
-                SpanId.bytesToHex(link.getSpanId().getBytes()),
-                TraceFlags.getDefault(),
-                TraceState.getDefault()),
-            attributesBuilder.build());
-      }
+  static Kind mapKind(@Nullable Span.Kind kind) {
```
anuraaga, line 71: Let's use fullly qualified class names, it's really tricky knowing which Kind is which

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/SpanConverter.java
```diff
@@ -69,64 +66,12 @@ public void end(EndSpanOptions options) {}
...
+    if (kind == null) {
+      return null;
```
anuraaga, line 73: An OTel kind should never be null - INTERNAL?

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/SpanConverter.java
```diff
@@ -142,117 +87,90 @@ static Span fromOtelSpan(io.opentelemetry.api.trace.Span otSpan) {
...
-  private static Tracestate mapTracestate(TraceState traceState) {
-    Tracestate.Builder tracestateBuilder = Tracestate.builder();
-    traceState.forEach(tracestateBuilder::set);
-    return tracestateBuilder.build();
+  static SpanContext mapSpanContext(io.opentelemetry.api.trace.SpanContext otelSpanContext) {
+    return SpanContext.create(
+        TraceId.fromLowerBase16(otelSpanContext.getTraceIdAsHexString()),
+        SpanId.fromLowerBase16(otelSpanContext.getSpanIdAsHexString()),
+        TraceOptions.builder().setIsSampled(otelSpanContext.isSampled()).build(),
+        Tracestate.builder().build());
```
anuraaga, line 102:
```suggestion
        mapTracestate(otelSpanContext.getTraceState()));
```

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/SpanConverter.java
```diff
@@ -142,117 +87,90 @@ static Span fromOtelSpan(io.opentelemetry.api.trace.Span otSpan) {
...
-          TimeUnit.NANOSECONDS);
-    }
+  static io.opentelemetry.api.trace.SpanContext mapSpanContext(SpanContext ocSpanContext) {
+    return io.opentelemetry.api.trace.SpanContext.create(
+        ocSpanContext.getTraceId().toLowerBase16(),
+        ocSpanContext.getSpanId().toLowerBase16(),
+        ocSpanContext.getTraceOptions().isSampled()
+            ? TraceFlags.getSampled()
+            : TraceFlags.getDefault(),
+        TraceState.getDefault());
```
anuraaga, line 112: Should we map this trace state?

opencensus-shim/src/test/java/io/opentelemetry/opencensusshim/OpenTelemetryB3FormatImplTest.java
```diff
@@ -0,0 +1,207 @@
...
+import io.opencensus.trace.Tracestate;
+import io.opencensus.trace.propagation.BinaryFormat;
+import io.opencensus.trace.propagation.SpanContextParseException;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Tests for {@link OpenTelemetryBinaryFormatImpl}. */
+@RunWith(JUnit4.class)
+public class OpenTelemetryB3FormatImplTest {
```
anuraaga, line 24: By making a generic `OpenTelemetryTextFormat` we won't need all this testing, we'd be able to test a mock `TextMapPropagator`

anuraaga: Oh wait - this is the binary format test XD. Rename the class. I guess we need tests for the textmappropagator wrapper too.

opencensus-shim/src/test/java/io/opentelemetry/opencensusshim/OpenTelemetryB3FormatImplTest.java
```diff
@@ -0,0 +1,207 @@
...
+import io.opencensus.trace.TraceOptions;
+import io.opencensus.trace.Tracestate;
+import io.opencensus.trace.propagation.BinaryFormat;
+import io.opencensus.trace.propagation.SpanContextParseException;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Tests for {@link OpenTelemetryBinaryFormatImpl}. */
+@RunWith(JUnit4.class)
```
anuraaga, line 23: Not using any junit4 features here, can you switch to jupiter?

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanBuilderImpl.java
zoercai: I think the null is necessary here since the sampler's comment says

> @param hasRemoteParent {@code true} if the parent {@code Span} is remote. {@code null} if this is a root span.

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanImpl.java
zoercai: I think only OpenCensus instrumentation will interact with these spans, unless for some reason a user decided to get an OC parent/child span from a library and change its status. I've updated the Otel methods to delegate to the otelSpan just in case though.

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanBuilderImpl.java
```diff
@@ -0,0 +1,229 @@
...
+    if (!ocTraceOptions.isSampled() && !Boolean.TRUE.equals(recordEvents)) {
+      return OpenTelemetryNoRecordEventsSpanImpl.create(
+          SpanContext.create(ocTraceId, ocSpanId, ocTraceOptions, ocTracestate));
+    }
+
+    // If sampled
+    io.opentelemetry.api.trace.SpanBuilder otelSpanBuilder =
+        OTEL_TRACER
+            .spanBuilder(name)
+            .setStartTimestamp(options.clock.nowNanos(), TimeUnit.NANOSECONDS);
```
anuraaga, line 143: Do you think we need to support the OC clock? I'd recommend removing this line if it seems reasonable.

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanImpl.java
```diff
@@ -0,0 +1,231 @@
...
+        arg -> otelSpan.setAttribute(key, arg),
+        arg -> null);
+  }
+
+  @Override
+  public void putAttributes(Map<String, AttributeValue> attributes) {
+    Preconditions.checkNotNull(attributes, "attributes");
+    for (Map.Entry<String, AttributeValue> attribute : attributes.entrySet()) {
+      putAttribute(attribute.getKey(), attribute.getValue());
+    }
```
anuraaga, line 82:
```suggestion
    attributes.forEach(this::putAttribute);
```

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanImpl.java
```diff
@@ -0,0 +1,231 @@
...
+                setLongAttribute(attributesBuilder, s),
+                setDoubleAttribute(attributesBuilder, s),
+                arg -> null));
+    otelSpan.addEvent(description, attributesBuilder.build());
+  }
+
+  @Override
+  public void addAnnotation(Annotation annotation) {
+    AttributesBuilder attributesBuilder = Attributes.builder();
+    annotation
```
anuraaga, line 102: Looks like you can extract a helper for converting oc attributes to otel attributes

opencensus-shim/src/test/java/io/opentelemetry/opencensusshim/OpenTelemetryBinaryFormatImplTest.java
```diff
@@ -0,0 +1,187 @@
...
+
+  private void testSpanContextConversion(SpanContext spanContext) throws SpanContextParseException {
+    SpanContext propagatedBinarySpanContext =
+        binaryFormat.fromByteArray(binaryFormat.toByteArray(spanContext));
+
+    assertThat(propagatedBinarySpanContext).isEqualTo(spanContext);
+  }
+
+  @Test
+  public void propagate_SpanContextTracingEnabled() throws SpanContextParseException {
```
anuraaga, line 46: Can remove public from test methods

opencensus-shim/src/test/java/io/opentelemetry/opencensusshim/OpenTelemetryTextFormatImplTest.java
```diff
@@ -0,0 +1,135 @@
...
+  private static final Random RANDOM = new Random();
+  private static final SpanContext SPAN_CONTEXT =
+      SpanContext.create(
+          TraceId.generateRandomId(RANDOM),
+          SpanId.generateRandomId(RANDOM),
+          TraceOptions.builder().setIsSampled(true).build(),
+          Tracestate.builder().set("key", "value").build());
+
+  @Test
+  public void testInject() {
```
anuraaga, line 59: Can remove public from all the methods

anuraaga (commented):
Thanks a lot for the work, this mostly looks good!

Do you think it's possible to improve the code coverage?
https://codecov.io/gh/open-telemetry/opentelemetry-java/pull/2060/tree/opencensus-shim/src/main/java/io/opentelemetry/opencensusshim

In particular, given the `return null` we fixed to `return this`, I think we're missing interop tests where OTel instrumentation is included into an app using OC.

Can wait for adding more tests until @bogdandrutu gives a thumbs up too.

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanBuilderImpl.java
zoercai: The clock is created at https://github.com/zoercai/opentelemetry-java/blob/4e7d8b25cb24466f5a678cfd67dd5769fe599d4e/opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryTraceComponentImpl.java#L33
and is publicly available via `Tracing.java`, so it seems like I should use it here just in case. Let me know if you think it's unnecessary though!

anuraaga: Customizing the OC clock seems like a very small use case - it seems unlikely to do any better than what OTel will provide (e.g., we updated the Clock to use Java 9+ for improved precision) so it could be considered a good thing to not respect the OC clock. But no personal opinion food for thought on what could be better here.

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanBuilderImpl.java
bogdandrutu: I see, maybe we can stick with what we have here for the moment, and look into ways to improve this.

anuraaga (commented):
Thanks a lot for all the work and the test coverage! We've gotten much better. We're still a bit lower than the current HEAD though

HEAD - https://codecov.io/gh/open-telemetry/opentelemetry-java 78%
PR - https://codecov.io/gh/open-telemetry/opentelemetry-java/tree/283d3659c75ce0bfe526fe15397f4bb5db44e039 - 75%

I'm open to being flexible if absolutely must, but it would be great to at least match the old number so we're not dropping coverage overall (well, there's more code now so maybe it's not a real drop but the % is what we generally are looking at for ease of reasoning). Would it be possible to do just a bit more testing work? #2242 will help you. And it looks like the remaining major areas of coverage are

- Links
- Events with specific time
- Certain attribute types

I think we these we could get to the previous number.

anuraaga (approved):
@zoercai Thanks a lot for the hard work!

zoercai:
@anuraaga Is this mergeable despite the codecov check failing? The current master branch has less than 90% coverage, and the diff for this PR has 89.24%, I don't think I can increase the whole project coverage to more than 90% with this PR.
And is there anything else I need to do for this to be merged?

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanBuilderImpl.java
zoercai: I've removed the clock

anuraaga:
Hey @zoercai - for me this looks good hence the approval, and we generally merge with two approvals for large changes. I'm hoping @bogdandrutu can approve this since he's most familiar with census, otherwise I think@jkwatson will soon.

anuraaga:
@bogdandrutu @jkwatson Do one of you mind merging this today? We can't let it sit around any longer

anuraaga:
@bogdandrutu @jkwatson Last call for getting a review for this before it ends up getting merged with my single green check

anuraaga:
@zoercai Thanks again for the extreme patience!

@bogdandrutu @jkwatson I can help address any comments you may have in a followup.

PR #2062, opened 2020-11-11
Some cleanup in span land.

jkwatson:
* Clean up the SpanBuilder javadoc to match the current APIs
* Scrub mentions of canonical status code
* Deprecate the getCanonicalCode method on SpanBuilder and replace with getStatusCode

### DIFF
--- a/api/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java
+++ b/api/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java
@@ -45,7 +45,7 @@ import javax.annotation.Nonnull;
  *
  * <pre>{@code
  * class MyRpcServerInterceptorListener implements RpcServerInterceptor.Listener {
- *   private static final Tracer tracer = OpenTelemetry.get().getTracer("com.anyco.rpc");
+ *   private static final Tracer tracer = OpenTelemetry.get().getTracer("com.example.rpc");
  *   private Span mySpan;
  *
  *   public MyRpcInterceptor() {}
--- a/api/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java
+++ b/api/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java
@@ -85,7 +85,7 @@ import javax.annotation.Nonnull;
  *
  * <pre>{@code
  * class MyClass {
- *   private static final Tracer tracer = OpenTelemetry.get().getTracer("com.anyco.rpc");
+ *   private static final Tracer tracer = OpenTelemetry.get().getTracer("com.example.rpc");
  *
  *   void doWork(Span parent) {
  *     Span childSpan = tracer.spanBuilder("MyChildSpan")
--- a/api/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java
+++ b/api/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java
@@ -292,7 +292,7 @@ public interface SpanBuilder {
    *
    * <pre>{@code
    * class MyClass {
-   *   private static final Tracer tracer = OpenTelemetry.get().getTracer("com.anyco.rpc");
+   *   private static final Tracer tracer = OpenTelemetry.get().getTracer("com.example.rpc");
    *
    *   void doWork(Span parent) {
    *     Span childSpan = tracer.spanBuilder("MyChildSpan")
--- a/api/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java
+++ b/api/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java
@@ -24,12 +24,13 @@ import javax.annotation.Nonnull;
  *
  * <pre>{@code
  * class MyClass {
- *   private static final Tracer tracer = OpenTelemetry.getTracer();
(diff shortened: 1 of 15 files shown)


api/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java
```diff
@@ -44,20 +45,20 @@
  *
  * <pre>{@code
  * class MyRpcServerInterceptorListener implements RpcServerInterceptor.Listener {
- *   private static final Tracer tracer = OpenTelemetry.getTracer();
+ *   private static final Tracer tracer = OpenTelemetry.get().getTracer("com.anyco.rpc");
```
arminru, line 48:
```suggestion
 *   private static final Tracer tracer = OpenTelemetry.get().getTracer("com.example.rpc");
```
This would be the reserved domain by IANA for such purposes.

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/data/SpanData.java
```diff
@@ -340,12 +340,19 @@ static Status create(StatusCode code, @Nullable String description) {
...
     /**
-     * Returns the canonical status code.
+     * Returns the status code.
      *
-     * @return the canonical status code.
+     * @deprecated Please use {@link #getStatusCode()}. This method will be removed in the next
+     *     release.
      */
-    StatusCode getCanonicalCode();
+    @Deprecated
```
Oberon00, line 352: I think we currently still have enough other breaking changes that it is OK to completely remove this. It's just a renaming, right?

sdk/tracing/src/test/java/io/opentelemetry/sdk/trace/testbed/errorreporting/ErrorReportingTest.java
```diff
@@ -50,7 +50,7 @@ void testSimpleError() {
 
     List<SpanData> spans = otelTesting.getSpans();
     assertThat(spans).hasSize(1);
-    assertThat(spans.get(0).getStatus().getCanonicalCode()).isEqualTo(StatusCode.ERROR);
+    assertThat(spans.get(0).getStatus().getStatusCode()).isEqualTo(StatusCode.ERROR);
```
arminru, line 53:
```suggestion
    assertThat(spans.get(0).getStatus().getStatusCode()).isEqualTo(StatusCode.ERROR);
    assertThat(spans.get(0).getStatus().getCanonicalCode()).isEqualTo(StatusCode.ERROR);
```
I'd also add tests assuring that the deprecated method still works and doesn't throw until it's removed.

api/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java
```diff
@@ -84,10 +85,12 @@
  *
  * <pre>{@code
  * class MyClass {
- *   private static final Tracer tracer = OpenTelemetry.getTracer();
- *   void DoWork(Span parent) {
+ *   private static final Tracer tracer = OpenTelemetry.get().getTracer("com.anyco.rpc");
```
arminru, line 88:
```suggestion
 *   private static final Tracer tracer = OpenTelemetry.get().getTracer("com.example.rpc");
```

api/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java
```diff
@@ -290,10 +292,11 @@ default SpanBuilder setStartTimestamp(Instant startTimestamp) {
    *
    * <pre>{@code
    * class MyClass {
-   *   private static final Tracer tracer = OpenTelemetry.getTracer();
-   *   void DoWork(Span parent) {
+   *   private static final Tracer tracer = OpenTelemetry.get().getTracer("com.anyco.rpc");
```
arminru, line 295:
```suggestion
   *   private static final Tracer tracer = OpenTelemetry.get().getTracer("com.example.rpc");
```

arminru (approved):

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/data/SpanData.java
jkwatson: I've started thinking that since we're moving to a 2-week release cycle, that it would be more friendly to end users to introduce deprecations that are then deleted in the following release. We discussed this at the SIG meeting this morning, and it was generally agreed to.  This seemed like a easy way to ease into the practice. :)

sdk/tracing/src/test/java/io/opentelemetry/sdk/trace/testbed/errorreporting/ErrorReportingTest.java
jkwatson: fair enough. will add.

carlosalberto (approved):

sdk/tracing/src/test/java/io/opentelemetry/sdk/trace/testbed/errorreporting/ErrorReportingTest.java
jkwatson: done!

anuraaga (approved):
Thanks for scrubbing!

PR #2037, opened 2020-11-06
Very basic Aggregation-configuration API in the SDK.

jkwatson:
Note: related OTEP: open-telemetry/oteps#126

This first-cut supports the following:

select instruments to customize by instrument type and name.
specify the aggregation and the temporality of the aggregation to be used.

An example of how you could use it:

```java
// get a handle to the MeterSdkProvider
 MeterSdkProvider meterProvider = OpenTelemetrySdk.getMeterProvider();

 // create a selector to select which instruments to customize:
 InstrumentSelector instrumentSelector = InstrumentSelector.newBuilder()
  .instrumentType(InstrumentType.COUNTER)
  .build();

 // create a specification of how you want the metrics aggregated:
 ViewSpecification viewSpecification = 
      ViewSpecification.create(Aggregations.minMaxSumCount(), Temporality.DELTA);

 //register the view with the MeterSdkProvider
 meterProvider.registerView(instrumentSelector, viewSpecification);
```

PR #2063, opened 2020-11-11
Add DDSketch as an option for Metric Aggregation.

tylerbenson:
Will require #2037 (or something like it) in order to actually be usable.

Current iteration also doesn't map to any protobuf format.

### DIFF
--- a/sdk/all/src/jmh/java/io/opentelemetry/sdk/metrics/aggregator/DDSketchAggregatorBenchmark.java
+++ b/sdk/all/src/jmh/java/io/opentelemetry/sdk/metrics/aggregator/DDSketchAggregatorBenchmark.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.metrics.aggregator;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Threads;
+import org.openjdk.jmh.annotations.Warmup;
+
+import java.util.concurrent.TimeUnit;
+
+@State(Scope.Benchmark)
+public class DDSketchAggregatorBenchmark {
+
+  private Aggregator aggregator;
+
+  @Setup(Level.Trial)
+  public final void setup() {
+    aggregator = DDSketchAggregator.getBalancedFactory().getAggregator();
+  }
+
+  @Benchmark
+  @Fork(1)
+  @Warmup(iterations = 5, time = 1)
+  @Measurement(iterations = 10, time = 1)
+  @OutputTimeUnit(TimeUnit.MILLISECONDS)
+  @Threads(value = 10)
+  public void aggregate_10Threads() {
+    aggregator.recordDouble(100.0056);
+  }
+
+  @Benchmark
+  @Fork(1)
+  @Warmup(iterations = 5, time = 1)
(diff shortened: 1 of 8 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DDSketchAggregator.java
```diff
@@ -0,0 +1,124 @@
...
+        drain();
+        current.accept(value);
+        sum += value;
+      }
+    }
+  }
+
+  /** Must be called under lock. */
+  private void drain() {
+    assert Thread.holdsLock(lock);
```
jkwatson, line 115: We don't generally use `assert` in this codebase.  Can you use the `@GuardedBy("lock")` annotation on the method instead?

sdk/all/src/jmh/java/io/opentelemetry/sdk/metrics/aggregator/DDSketchAggregatorBenchmark.java
```diff
@@ -0,0 +1,59 @@
...
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Threads;
+import org.openjdk.jmh.annotations.Warmup;
+
+@State(Scope.Benchmark)
+public class DDSketchAggregatorBenchmark {
```
jkwatson, line 21: Can you publish some benchmarks? What kind of throughput can this implementation handle, with the lock on every recording?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DDSketchAggregator.java
```diff
@@ -0,0 +1,124 @@
...
+          ? null
+          : SummaryPoint.create(
+              startEpochNanos,
+              epochNanos,
+              labels,
+              count,
+              sum,
+              Arrays.asList(
+                  ValueAtPercentile.create(0.0, current.getMinValue()),
+                  ValueAtPercentile.create(100.0, current.getMaxValue())));
```
jkwatson, line 95: I would have thought we would get more percentiles out of the sketch. Is that for a later PR?

jkwatson:
I'm pretty worried about requiring a lock on every metric recording that goes through this aggregator. Could we drop recordings if we can't keep up, rather than adding another lock in that path?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DDSketchAggregator.java
```diff
@@ -0,0 +1,124 @@
...
+                  new SignedDDSketch(
+                      new CubicallyInterpolatedMapping(PRECISION), UnboundedSizeDenseStore::new));
+
+  public static AggregatorFactory getBalancedFactory() {
+    return AGGREGATOR_FACTORY;
+  }
+
+  // Since DDSketch is not thread safe, this queue is used to buffer calls to record, reducing
+  // the need to take a lock for every call. (With the downside of having to box the double.)
+  private final ArrayBlockingQueue<Double> pendingValues = new ArrayBlockingQueue<>(64);
```
anuraaga, line 39: Doesn't a blocking queue also imply a taking a lock on every call? (it's a blocking queue :) ). 

https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/concurrent/ArrayBlockingQueue.java#L344

Is the idea that it would be less time under lock than if always calling `current.accept`?

jkwatson:
Had a discussion in the metrics SIG today, and everyone pretty much agrees we should revert to just doing a simple synchronized lock around the recordings and ditch the queue in front of it. We can benchmark that and see how it performs before we do something fancier.

jkwatson:
@tylerbenson I'm going to close this for now, unless you're going to have time to update it for the latest changes to the metrics SDK. Please re-open if appropriate.

tylerbenson:
Good call.  Sorry for leaving this open for so long.

PR #2066, opened 2020-11-12
Add kotlin extension for interaction with Context in coroutines.

anuraaga:
While working on improvements to interaction between kotlin and context in instrumentation repo, I realized it's not really specific to instrumentation and something any API user will need help with if using kotlin coroutines. So here's a kotlin extension

https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1618


extensions/kotlin/src/main/java/io/opentelemetry/extension/kotlin/ContextElement.java
```diff
@@ -0,0 +1,72 @@
...
+import kotlin.jvm.functions.Function2;
+import kotlinx.coroutines.ThreadContextElement;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * {@link ThreadContextElement} for synchronizing a {@link Context} across coroutine suspension and
+ * resumption. Implemented in Java instead of Kotlin to allow usage in auto-instrumentation where
+ * there is an outstanding Kotlin bug preventing it https://youtrack.jetbrains.com/issue/KT-20869.
+ */
+class ContextElement implements ThreadContextElement<Scope> {
```
jkwatson, line 20: I wonder if we should prefix this with `Kotlin` or `Coroutine` to make reading code usages super clear about which context we're referring to.

extensions/kotlin/src/main/java/io/opentelemetry/extension/kotlin/ContextElement.java
```diff
@@ -0,0 +1,72 @@
...
+
+  static final CoroutineContext.Key<ContextElement> KEY =
+      new CoroutineContext.Key<ContextElement>() {};
+
+  private final Context context;
```
jkwatson, line 25: similarly, I might call this field `otelContext` or `openTelemetryContext` for extra clarity.

jkwatson:
This seems great, but someone with more kotlin expertise than I should give it a deeper review. And, I think we should make the naming of things very clear, because it's super easy to confuse which context is being referred to. :)

trask:
> someone with more kotlin expertise than I should give it a deeper review

@johnbley reviewed it here: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1618#pullrequestreview-529219361

extensions/kotlin/src/main/java/io/opentelemetry/extension/kotlin/ContextElement.java
anuraaga: Yup good call

context/src/main/java/io/opentelemetry/context/ImplicitContextKeyed.java
```diff
@@ -21,6 +21,12 @@
    * automatically.
    *
    * <p>This method is equivalent to {@code Context.current().with(value).makeCurrent()}.
+   *
+   * <p>The default implementation of this method will store the {@link ImplicitContextKeyed} in a
+   * {@link ThreadLocal}. Kotlin coroutine users SHOULD NOT use this method as the {@link
```
jkwatson, line 26: I wonder if we should also mention that many async framework users won't want to use this either. Even if we don't have a drop-in solution for them.

jkwatson (approved):
I'll trust that @johnbley has reviewed this and approved. It looks ok to me.

PR #2068, opened 2020-11-12
Implement BatchObserver

kenfinnigan:
This PR is an alternative fix for #1331 implementing [Batch Observer](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#batch-observer).

Appreciate any comments and feedback on the API

### DIFF
--- a/api/src/main/java/io/opentelemetry/api/metrics/batch/BatchObserverContext.java
+++ b/api/src/main/java/io/opentelemetry/api/metrics/batch/BatchObserverContext.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.api.metrics.batch;
+
+import io.opentelemetry.api.common.Labels;
+import io.opentelemetry.api.metrics.Instrument;
+import io.opentelemetry.api.metrics.batch.AsynchronousBatchInstrument.Observation;
+import java.util.function.Consumer;
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * Util class for creating multiple instruments that can be atomically observed in a single
+ * callback.
+ *
+ * <p>This class is equivalent to individually calling {@code observe} on every {@link Instrument},
+ * but has the advantage of all recordings occurring atomically.
+ *
+ * <p>Example:
+ *
+ * <pre>{@code
+ * meter.newBatchObserver(
+ *     context -> {
+ *       DoubleSumBatchObserver timeObserver =
+ *           context.doubleSumObserverBuilder("time").setUnit("s").build();
+ *       LongSumBatchObserver distanceObserver =
+ *           context.longSumObserverBuilder("distance").setUnit("m").build();
+ *
+ *       context.registerCallback(
+ *           result ->
+ *               MyMeasure measure = MyMeasure.getdata();
+ *               result.observe(
+ *                   Labels.of("myKey", "myValue"),
+ *                   distanceObserver.observation(measure.getDistance()),
+ *                   timeObserver.observation(measure.getTime())));
+ *     });
+ * }</pre>
+ */
+@ThreadSafe
+public interface BatchObserverContext extends Instrument {
+
+  /**
+   * Returns a new builder for a {@link DoubleValueBatchObserver}.
+   *
+   * @param name the name of the instrument. Should be a ASCII string with a length no greater than
+   *     255 characters.
+   * @return a new builder for a {@code DoubleValueBatchObserver}.
+   * @throws NullPointerException if {@code name} is null.
+   * @throws IllegalArgumentException if different metric with the same name already registered.
+   * @throws IllegalArgumentException if the {@code name} does not match the requirements.
+   */
+  DoubleValueBatchObserver.Builder doubleValueObserverBuilder(String name);
+
+  /**
+   * Returns a new builder for a {@link LongValueBatchObserver}.
+   *
+   * @param name the name of the instrument. Should be a ASCII string with a length no greater than
+   *     255 characters.
+   * @return a new builder for a {@code LongValueBatchObserver}.
+   * @throws NullPointerException if {@code name} is null.
+   * @throws IllegalArgumentException if different metric with the same name already registered.
+   * @throws IllegalArgumentException if the {@code name} does not match the requirements.
+   */
+  LongValueBatchObserver.Builder longValueObserverBuilder(String name);
+
+  /**
+   * Set a {@link Consumer} function that accepts a {@link BatchObserverContext}. This function will
+   * call {@code observe} to execute the observation of multiple {@link Instrument}s.
+   *
+   * @param resultConsumer Function to consume a {@link BatchObserverContext}.
+   */
+  void registerCallback(Consumer<BatchObserverResult> resultConsumer);
+
+  /** Builder class for {@link BatchObserverContext}. */
+  interface Builder extends Instrument.Builder {
+    @Override
+    BatchObserverContext build();
+  }
+
+  @FunctionalInterface
+  interface BatchObserverResult {
+    void observe(Labels labels, Observation... observations);
+  }
+}
--- a/api/src/main/java/io/opentelemetry/api/metrics/DefaultMeter.java
+++ b/api/src/main/java/io/opentelemetry/api/metrics/DefaultMeter.java
@@ -8,7 +8,11 @@ package io.opentelemetry.api.metrics;
 import io.opentelemetry.api.common.Labels;
 import io.opentelemetry.api.internal.StringUtils;
 import io.opentelemetry.api.internal.Utils;
+import io.opentelemetry.api.metrics.batch.BatchObserverContext;
+import io.opentelemetry.api.metrics.batch.DoubleValueBatchObserver;
+import io.opentelemetry.api.metrics.batch.LongValueBatchObserver;
 import java.util.Objects;
+import java.util.function.Consumer;
 import javax.annotation.concurrent.Immutable;
 import javax.annotation.concurrent.ThreadSafe;
 
@@ -116,6 +120,12 @@ final class DefaultMeter implements Meter {
   public BatchRecorder newBatchRecorder(String... keyValuePairs) {
     Utils.validateLabelPairs(keyValuePairs);
     return NoopBatchRecorder.INSTANCE;
+  }
+
+  @Override
+  public void newBatchObserver(Consumer<BatchObserverContext> context) {
+    Objects.requireNonNull(context, "context");
+    context.accept(NoopBatchObserverContext.INSTANCE);
   }
 
   private DefaultMeter() {}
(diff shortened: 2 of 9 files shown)


api/src/main/java/io/opentelemetry/api/metrics/batch/BatchObserverContext.java
```diff
@@ -0,0 +1,86 @@
...
+   * Returns a new builder for a {@link DoubleValueBatchObserver}.
+   *
+   * @param name the name of the instrument. Should be a ASCII string with a length no greater than
+   *     255 characters.
+   * @return a new builder for a {@code DoubleValueBatchObserver}.
+   * @throws NullPointerException if {@code name} is null.
+   * @throws IllegalArgumentException if different metric with the same name already registered.
+   * @throws IllegalArgumentException if the {@code name} does not match the requirements.
+   */
+  DoubleValueBatchObserver.Builder doubleValueObserverBuilder(String name);
```
thisthat, line 54: AFAIR, we agreed to do not double the instruments before a better specification for the batch observer is done. I cannot find a link that documents this decision, so maybe it was just a conversation between me and @jkwatson

kenfinnigan: Would be good to have that clarified.

The reason I went with separate instruments was to prevent the ability for developers to call `setCallback()` on the `Instrument` as it could lead to confusion.

jkwatson: I'm on both sides of the fence. I'd rather not duplicate the instruments, but I'd also like to make the API really clear. Is there a way we can do both? Suggestions welcome!

jkwatson: Could we use the same instrument, but have the batch one have a no-op for the callback addition? Or could we make the non-batch one extend the interface for the batch one and add the callback method?

kenfinnigan: I'd also welcome suggestions as I couldn't come up with a "clean" way that didn't result in tradeoffs in one place or another.

We could use the same instrument and make `setCallback()` a no-op. My concern would be developers use it and don't realize it's a problem until they wonder why the metric isn't being observed. Maybe there's a way to mitigate that in the actual SDK implementation.

We could also have a non-batch extend a batch interface. A possible concern is that `observation()` is then available on the non-batch Instrument. As with the previous no-op suggestion, maybe there's a way to handle that gracefully.

At the moment there are options, I'm not certain which way we should jump

jkwatson: ok, here are my thoughts:
1) If I have to choose between making things clear and easy for end users and making our lives easier as maintainers, I'll choose the end user. This is an argument in favor of duplicating the instruments.
2) If we like this pattern, we could use it for the non-batch observer instruments as well. In that case, could we convert both types over to the new ones and get rid of the old ones?

kenfinnigan: Wholeheartedly agree with 1).

2) I'm a little unsure of exactly what you mean. Can you clarify what the "pattern" is that you refer to?

It might be possible to use a single instrument for both cases, would need to experiment a bit. Especially as there is a difference between `observe(value, labels)` and `observation(value)`.

api/src/main/java/io/opentelemetry/api/metrics/batch/BatchObserverContext.java
```diff
@@ -0,0 +1,86 @@
...
+ *       LongSumBatchObserver distanceObserver =
+ *           context.longSumObserverBuilder("distance").setUnit("m").build();
+ *
+ *       context.registerCallback(
+ *           result ->
+ *               MyMeasure measure = MyMeasure.getdata();
+ *               result.observe(
+ *                   Labels.of("myKey", "myValue"),
+ *                   distanceObserver.observation(measure.getDistance()),
+ *                   timeObserver.observation(measure.getTime())));
```
jkwatson, line 37: @jmacd What do you think about this API shape for BatchObserver registration? @kenfinnigan and I are thinking about proposing this as a clarification in the specs for this API.

api/src/main/java/io/opentelemetry/api/metrics/batch/BatchObserverContext.java
jkwatson: @kenfinnigan check out https://github.com/open-telemetry/opentelemetry-java/pull/2120 where I've deprecated the `setCallback` method on the Async instruments. I think this might mean we could re-use the same instruments, yes?

thisthat: I think that PR allows for the reuse of the instruments, yes!

api/src/main/java/io/opentelemetry/api/metrics/batch/BatchObserverContext.java
jmacd: LGTM. If another `registerCallback`  happens later, it replaces the format callback, I imagine?

api/src/main/java/io/opentelemetry/api/metrics/batch/BatchObserverContext.java
kenfinnigan: Will take a look, thanks @jkwatson

api/src/main/java/io/opentelemetry/api/metrics/batch/BatchObserverContext.java
Oberon00: I think it would be ideal if there was no `registerCallback` function and the callback is only passed during instrument creation. One design possibility would be

>  An async instrument creation function accepts a callback but returns nothing. Then the callback is invoked with the instrument as parameter. This prevents the API user from accidentally calling the observe() function outside the interval.

-- https://github.com/open-telemetry/opentelemetry-specification/pull/1030#issuecomment-707668179

kenfinnigan: @Oberon00 I like the idea of preventing unintentional calling of `observe()`, but I wasn't able to come up with how this could work given the callback is taking multiple instruments?

Maybe I need more coffee this morning. Can you provide a pseudo-code example?

Oberon00: Actually I don't think this would work as-is for the BatchObserver, only for the other instruments. But maybe we can still make it so that there is no registerCallback method?

kenfinnigan: Maybe creation of the BatchObserver can accept a set of Labels and an array of instruments?

Something like:

```
meter.newBatchObserver(
    Labels.of("myKey", "myValue"),
    meter.doubleSumObserverBuilder("time").setUnit("s").build(),
    meter.longSumObserverBuilder("distance").setUnit("m").build()
)
```

Haven't confirmed this would "work" for the observe side, just a quick stab at what you're thinking

api/src/main/java/io/opentelemetry/api/metrics/batch/BatchObserverContext.java
kenfinnigan: @jkwatson I haven't moved things around, but I think you're right that removing `setCallback` from `AsynchronousInstrument` makes it possible to re-use the existing instrument builders.

Would that also mean `observation()` would be added to the builder?

jkwatson: not sure how that would work, since it needs to return an `Observation`. Maybe I'm missing what you mean by the question.

kenfinnigan: @jkwatson True, was trying to think it through without spending time reworking all the code. ;-)

I think I was getting confused, so ignore that comment for now.

What's the outlook on #2120 being merged?

jkwatson: Seems very good. I'm letting it rest a bit, since it's an API change, and it's a holiday week. I would assume at this point that it will happen, unless someone comes up with a very good reason that it shouldn't.

api/src/main/java/io/opentelemetry/api/metrics/batch/BatchObserverContext.java
```diff
@@ -0,0 +1,86 @@
...
+
+  /** Builder class for {@link BatchObserverContext}. */
+  interface Builder extends Instrument.Builder {
+    @Override
+    BatchObserverContext build();
+  }
+
+  @FunctionalInterface
+  interface BatchObserverResult {
+    void observe(Labels labels, Observation... observations);
```
bogdandrutu, line 84: Unnecessary array allocation for the 99% of cases where only one value per observation.

kenfinnigan: Why would it be one value?

The goal of this is to support atomic capture of multiple instrument values, so there should be at least two.

jmacd: I think one would not choose a BatchObserver in the case where there is more than one observation for the same label set?

kenfinnigan: Maybe I'm confused, I thought the goal of [BatchObserver](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#batch-observer) was to capture observations from multiple instruments within a single atomic observation with an identical label set.

Am I misinterpreting something?

jkwatson: I think @bogdandrutu is suggesting that maybe we only need one API for both..you could just use the batch API for a single instrument.

jkwatson:
Another thought that @bogdandrutu had (discussed offline) was, rather than having an *explicit* batch API, we could devise an implicit one. For instance, if you had a method on the async instrument builder like : 

`<E> Builder setCallback(Supplier<E> expensiveThingToGet, BiConsumer<E, LongResult> callback);`

And two instruments were created with the same `expensiveThingToGet` instance, that that SDK could detect that fact and implicitly batch the two instruments together.

bogdandrutu:
There are few other advantages to this approach:
* SDK can partially update the batch (call the consumer on some of the instruments, in case the others are disabled or configured to be exported at different rate).
* The building step will look very similar in both cases.
* Only one extra object created the result of the supplier, everything else is the same as in the case of no batching.
* Some implementation may decide to not do batching at all

jkwatson:
@kenfinnigan I'm going to close this for now, as there is no longer a definition of a batch observer in the API. We can reopen this work when/if it reappears.

kenfinnigan:
No problem, thanks

PR #2069, opened 2020-11-13
Make the OpenTelemetrySdk a subclass of DefaultOpenTelemetry

### LINKED ISSUE #2018: Setting custom TracerProvider breaks access to TracerSdkManagement

jkwatson:
We discussed this possibility several times, so I thought I'd give it a try.

Feedback welcome. This resolves #2018 as well.

### DIFF
--- a/sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
+++ b/sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
@@ -104,6 +104,15 @@ public final class OpenTelemetrySdk extends DefaultOpenTelemetry {
     private Clock clock = MillisClock.getInstance();
     private Resource resource = Resource.getDefault();
 
+    /**
+     * Sets the {@link TracerSdkProvider} to use. This can be used to configure tracing settings by
+     * returning the instance created by a {@link TracerSdkProvider.Builder}.
+     *
+     * <p>Note: the parameter passed in here must be a {@link TracerSdkProvider} instance.
+     *
+     * @see TracerSdkProvider#builder()
+     * @param tracerProvider A {@link TracerSdkProvider} to use with this instance.
+     */
     @Override
     public Builder setTracerProvider(TracerProvider tracerProvider) {
       if (!(tracerProvider instanceof TracerSdkProvider)) {
@@ -114,12 +123,19 @@ public final class OpenTelemetrySdk extends DefaultOpenTelemetry {
       return this;
     }
 
+    /**
+     * Sets the {@link MeterProvider} to use.. This can be used to configure tracing settings by
+     * returning the instance created by a {@link MeterSdkProvider.Builder}.
+     *
+     * @see MeterSdkProvider#builder()
+     */
     @Override
     public Builder setMeterProvider(MeterProvider meterProvider) {
       super.setMeterProvider(meterProvider);
       return this;
     }
 
+    /** Sets the {@link ContextPropagators} to use. */
     @Override
     public Builder setPropagators(ContextPropagators propagators) {
       super.setPropagators(propagators);
@@ -155,19 +171,14 @@ public final class OpenTelemetrySdk extends DefaultOpenTelemetry {
      */
     @Override
     public OpenTelemetrySdk build() {
-      MeterProvider meterProvider = super.buildMeterProvider();
-      TracerProvider tracerProvider = super.buildTracerProvider();
-      if (!(tracerProvider instanceof TracerSdkProvider)) {
-        throw new IllegalStateException(
-            "The OpenTelemetrySdk can only be configured with a TracerSdkProvider");
-      }
-      ContextPropagators propagators = super.buildContextPropagators();
+      MeterProvider meterProvider = buildMeterProvider();
+      TracerProvider tracerProvider = buildTracerProvider();
 
       OpenTelemetrySdk sdk =
           new OpenTelemetrySdk(
               new ObfuscatedTracerProvider(tracerProvider),
               meterProvider,
-              propagators,
+              super.propagators,
               clock,
               resource);
       // Automatically initialize global OpenTelemetry with the first SDK we build.
@@ -175,6 +186,39 @@ public final class OpenTelemetrySdk extends DefaultOpenTelemetry {
         OpenTelemetry.set(sdk);
       }
       return sdk;
+    }
+
+    private TracerProvider buildTracerProvider() {
+      TracerProvider tracerProvider = super.tracerProvider;
+      if (tracerProvider != null) {
+        if (!(tracerProvider instanceof TracerSdkProvider)) {
+          throw new IllegalStateException(
+              "The OpenTelemetrySdk can only be configured with a TracerSdkProvider");
+        }
+        return tracerProvider;
+      }
+      TracerSdkProvider.Builder tracerProviderBuilder = TracerSdkProvider.builder();
+      if (clock != null) {
+        tracerProviderBuilder.setClock(clock);
+      }
+      if (resource != null) {
+        tracerProviderBuilder.setResource(resource);
+      }
+      return tracerProviderBuilder.build();
+    }
+
+    private MeterProvider buildMeterProvider() {
+      if (super.meterProvider != null) {
+        return super.meterProvider;
+      }
+      MeterSdkProvider.Builder meterProviderBuilder = MeterSdkProvider.builder();
+      if (clock != null) {
+        meterProviderBuilder.setClock(clock);
+      }
+      if (resource != null) {
+        meterProviderBuilder.setResource(resource);
+      }
+      return meterProviderBuilder.build();
     }
   }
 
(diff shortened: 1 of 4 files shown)


anuraaga (commented):
We discussed a bit offline but sending these out :)

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
```diff
@@ -114,68 +85,44 @@ public Clock getClock() {
...
-    private ContextPropagators propagators = DefaultContextPropagators.builder().build();
-
-    private TracerProvider tracerProvider;
-    private MeterProvider meterProvider;
 
-    /**
-     * Sets the {@link TracerProvider} to use. This can be used to configure tracing settings by
-     * returning the instance created by a {@link TracerSdkProvider.Builder}.
-     *
-     * @see TracerSdkProvider#builder()
```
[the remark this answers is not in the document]
jkwatson: for sure. will restore.

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
```diff
@@ -114,68 +85,44 @@ public Clock getClock() {
...
 
-    /**
-     * Sets the {@link MeterProvider} to use.. This can be used to configure tracing settings by
-     * returning the instance created by a {@link MeterSdkProvider.Builder}.
-     *
-     * @see MeterSdkProvider#builder()
-     */
     @Override
     public Builder setMeterProvider(MeterProvider meterProvider) {
-      requireNonNull(meterProvider, "meterProvider");
```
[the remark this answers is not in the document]
jkwatson: the super call has this already, so I don't think we need to duplicate it.

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
```diff
@@ -208,34 +155,21 @@ public Builder setResource(Resource resource) {
...
-          tracerProvider = new ObfuscatedTracerProvider(tracerProviderFactory.create());
-        } else if (HAS_TRACING_SDK) {
-          tracerProvider =
-              new ObfuscatedTracerProvider(
-                  TracerSdkProvider.builder().setClock(clock).setResource(resource).build());
-        } else {
-          tracerProvider = TracerProvider.getDefault();
-        }
+      MeterProvider meterProvider = super.buildMeterProvider();
+      TracerProvider tracerProvider = super.buildTracerProvider();
```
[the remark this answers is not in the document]
jkwatson: I added a really terrible unit test for this.

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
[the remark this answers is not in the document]
jkwatson: done

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
```diff
@@ -158,24 +131,14 @@ public Builder setTracerProvider(TracerProvider tracerProvider) {
      */
     @Override
     public Builder setMeterProvider(MeterProvider meterProvider) {
-      requireNonNull(meterProvider, "meterProvider");
-      this.meterProvider = meterProvider;
+      super.setMeterProvider(meterProvider);
```
anuraaga, line 134: Should we go ahead and be consistent and only accept `MeterSdkProvider` here since our goal is for this class to only configure the SDK?

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
```diff
@@ -114,39 +85,41 @@ public Clock getClock() {
...
+        .setMeterProvider(getMeterProvider())
         .setPropagators(getPropagators())
         .setClock(clock)
         .setResource(resource);
   }
 
   /** A builder for configuring an {@link OpenTelemetrySdk}. */
-  public static class Builder implements OpenTelemetryBuilder<Builder> {
+  public static class Builder extends DefaultOpenTelemetry.Builder {
     private Clock clock = MillisClock.getInstance();
```
anuraaga, line 104: For the below `null` checks to work I think these need to default to `null`

sdk/all/src/test/java/io/opentelemetry/sdk/OpenTelemetrySdkTest.java
```diff
@@ -76,10 +76,38 @@ void testReconfigure() {
...
+  @Test
+  void testReconfigure_justClockAndResource() {
+    Resource resource = Resource.create(Attributes.builder().put("cat", "meow").build());
+    OpenTelemetrySdk openTelemetry =
+        OpenTelemetrySdk.builder().setClock(clock).setResource(resource).build();
+    TracerProvider unobfuscatedTracerProvider =
+        ((ObfuscatedTracerProvider) openTelemetry.getTracerProvider()).unobfuscate();
+
+    assertThat(unobfuscatedTracerProvider).isInstanceOf(TracerSdkProvider.class);
+    // this is not a great way to test this...
```
anuraaga, line 96:
```suggestion
    // Since TracerProvider is in a different package, the only alternative to this reflective
    // approach would be to make the fields public for testing which is worse than this.
```

sdk/all/src/test/java/io/opentelemetry/sdk/OpenTelemetrySdkTest.java
```diff
@@ -76,10 +76,38 @@ void testReconfigure() {
...
+    assertThat(unobfuscatedTracerProvider)
+        .extracting("sharedState")
+        .hasFieldOrPropertyWithValue("clock", clock)
+        .hasFieldOrPropertyWithValue("resource", resource);
+
+    assertThat(openTelemetry.getMeterProvider()).isInstanceOf(MeterSdkProvider.class);
+    // this is not awesome
```
anuraaga, line 103:
```suggestion
    // Since MeterProvider is in a different package, the only alternative to this reflective
    // approach would be to make the fields public for testing which is worse than this.
```

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
jkwatson: very good point! will fix.

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
jkwatson: Yeah, that's probably a good idea. Better to enforce it now even if we don't need to, rather than have to break things later when we do.

anuraaga (approved):
Think there's just one point left about rejecting non-MeterSdkProvider but that doesn't have to happen now anyways. Thanks!

jkwatson:
> Think there's just one point left about rejecting non-MeterSdkProvider but that doesn't have to happen now anyways. Thanks!

oh shoot. I just forgot to do this. I'll add that real quick so we can get this merged and move forward with the SDK configuration story.

jkwatson:
> > Think there's just one point left about rejecting non-MeterSdkProvider but that doesn't have to happen now anyways. Thanks!
> 
> oh shoot. I just forgot to do this. I'll add that real quick so we can get this merged and move forward with the SDK configuration story.

done!

anuraaga (approved):
Nice

PR #2071, opened 2020-11-13
Fetch full repo in continuous build to allow computing nebula version.

anuraaga:
I think in the instrumentation repo where we use the same plugin, it was ok on CircleCI since it defaults to fetching the full repo. And I noticed we don't actually have a GH version of that build there right now.


anuraaga:
```
> Configure project :
Inferred project: opentelemetry-java, version: 0.11.0-SNAPSHOT
```

jkwatson (approved):
Thanks!

PR #2072, opened 2020-11-13
Explicitly bump CI to Java 8.0.275.

anuraaga:
GitHub seems to default to 272 still - while we will want to revert this eventually to make sure it continues to be autobumped in the future, we're currently seeing core dumps that may be fixed by temporarily manual bumping.

http://openjdk.5641.n7.nabble.com/Bug-3812-New-IcedTea7-Backport-JDK-8250861-quot-Crash-in-MinINode-Ideal-PhaseGVN-bool-td431647.html

### DIFF
--- a/.github/workflows/master-build.yml
+++ b/.github/workflows/master-build.yml
@@ -27,7 +27,6 @@ jobs:
         uses: actions/setup-java@v1
         with:
           java-version: 8.0.275
-          java-package: jre
       - id: setup-java-11
         name: Setup Java 11
         uses: actions/setup-java@v1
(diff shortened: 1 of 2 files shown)


anuraaga:
Hmm, 

```
Error: No valid download found for version 8.0.275 and package jre. Check https://static.azul.com/zulu/bin/ for a list of valid versions or download your own jdk file and add the jdkFile argument
```

Even though I can find zulu8.50.0.51-ca-jre8.0.275-macosx_x64.zip in that URL. Maybe we just live with the core dumps for a little longer since presumably it'll get autobumped soon...

anuraaga:
Filed https://github.com/actions/setup-java/issues/111 as there seems to be a misdetection of the JRE macos package, but not JDK. The difference in build time is minor, so I've gone ahead and switched to using JDK for Java 8 as well.

jkwatson (approved):

PR #2074, opened 2020-11-13
Disable javadoc task for kotlin project since there isn't any. And ha…

anuraaga:
…ve CI build instead of check.

Hadn't realized check isn't enough since it doesn't check javadoc.


jkwatson:
wait.. `build` does more than `check`? Shows what I know about gradle!

jkwatson (approved):
Thanks!

PR #2075, opened 2020-11-13
Fix memory pool metrics

devinrsmith:


devinrsmith:
Also - the two comments of the form

```java
    // TODO: Decide if max is needed or not. May be derived with some approximation from max(used).
```

are incorrect. Max is *not* a high water mark - it's the maximum *possible*. Ie, the values that are set when the system starts up via `-Xmx`.

jkwatson:
Thanks for this fix. Would you mind writing a unit test for this? ( I know we don't have one at the moment, but it seems like it would be a good thing to have ).

devinrsmith:
@jkwatson - it would be relatively easy for me to change the structure of these implementations to better support unit testing against `MemoryUsage` instances, and create those unit tests, if that is acceptable. If the idea is to keep the implementations exactly the same though, and mock the full `ManagementFactory` setup, that would take a lot more work and something I wouldn't be able to do at this time.

jkwatson:
Sounds great! Thanks!

jkwatson (approved):
Looks great! Thanks!

Issue #2080, opened 2020-11-16
Unable to build the opentelemetry-java project


ChandrewR:
I'm getting this error while building(executing gradlew.bat) - Could not find method release() for arguments [build_dfiyoporiz4savkbdijrduymk$_run_closure2@4f71c941] on root project 'opentelemetry-java' of type org.gradle.api.Project

**Steps to reproduce**
Download the java project and execute the gradlew.bat to reproduce the issue

**What did you expect to see?**
Build successful

**What did you see instead?**
Error
![Gradle Build Error](https://user-images.githubusercontent.com/26222838/99220708-270c5800-2805-11eb-8ea0-3f1b2105622c.PNG)

**Environment**
Compiler: JAVA 1.8.0_241
OS: Windows 10

Oberon00:
You need at least JDK 11 to build. 8 won't work.

anuraaga:
@ChandrewR Thanks for the report - interesting side effect of the nebula plugin that it is requiring a .git directory. For now, if you can clone from github and install Java 11 it should work ok, will look at fixing that.

jkwatson:
I think this has been fixed with #2082 and #2081. Can you verify, @ChandrewR ?

ChandrewR:
@anuraaga So which components uses JAVA 11?. I see all JAVA 8 in README.md files...Can you pls confirm?

anuraaga:
Hi @ChandrewR - building this repository requires Java 11 as listed here

https://github.com/open-telemetry/opentelemetry-java/blob/master/CONTRIBUTING.md#requirements

The actual built artifacts can be used with a Java 8 app.

I'm trying to find Java 8 in README but can't - can you point me at it? If it's in the context of building this repository, it needs to be updated.

ChandrewR:
Can individual artifacts like api, extensions etc be build with JAVA 8? It says JAVA 8 in README
https://github.com/open-telemetry/opentelemetry-java/tree/master/api
https://github.com/open-telemetry/opentelemetry-java/tree/master/extensions/trace-propagators

anuraaga:
Thanks @ChandrewR - it's indeed confusing to have mention of Java 8 all over the place. I have removed them in #2124 and instead clarify that published artifacts require Java 8+, and building requires looking at CONTRIBUTING.md, in the top-level README. Hopefully this prevents confusion going forward.

jkwatson:
[...]

PR #2081, opened 2020-11-16
Add an error message to better explain Java 11+ build requirement

anuraaga:


build.gradle
```diff
@@ -10,6 +10,13 @@ plugins {
     id "io.morethan.jmhreport" apply false
 }
 
+if (!JavaVersion.current().isJava11Compatible()) {
+    throw new GradleException("Java 11 or higher is required to build. " +
```
Oberon00, line 14:
```suggestion
    throw new GradleException("JDK 11 or higher is required to build. " +
```

build.gradle
```diff
@@ -10,6 +10,13 @@ plugins {
...
+            "Please download it from https://adoptopenjdk.net/. If you believe you already " +
+            "have it, please check that the JAVA_HOME environment variable is pointing at the " +
+            "Java 11 installation.")
```
Oberon00, line 17:
```suggestion
            "JDK 11 installation.")
```

Oberon00:
Partial "fix" for #2080

Oberon00 (approved):

build.gradle
```diff
@@ -10,6 +10,13 @@ plugins {
     id "io.morethan.jmhreport" apply false
 }
 
+if (!JavaVersion.current().isJava11Compatible()) {
+    throw new GradleException("Java 11 or higher is required to build. " +
+            "Please download it from https://adoptopenjdk.net/. If you believe you already " +
```
iNikem, line 15: I am against this endorsement. Albeit all the great work that AdoptOpenJDK did, I cannot recommend their distribution. At least as they are not [TCK-compliant](https://adoptopenjdk.net/quality.html#jck).

anuraaga: I mainly picked it because Gradle defaults to downloading it, and GitHub Actions has a plan to as well https://github.com/actions/setup-java/issues/98. I don't know why, but I'm personally OK with any URL here as long as it's stable, which official OpenJDK ones aren't.

iNikem: My go-to java distribution is [Liberica](https://bell-sw.com). Another option is [Zulu](https://www.azul.com/downloads/zulu-community/) (which is default for setup-java as of now)

Oberon00: I would prefer Adopt, but what about not including any concrete suggestion? Since this seems to be a point of conflict.

anuraaga: I checked openjdk.java.net again for Java 11 but it seems to be as hidden as ever. So the most standard option doesn't seem great.

I have a personal preference for zulu, mainly because of the TCK thing but see the ecosystem consolidating on adopt (GitHub planning to switch from zulu to adopt does mean something), which is why I went with it here. Admittedly the TCK issue seems theoretical given they are in the end effectively the same codebase, and both have wide adoption.

I think a link is far more user friendly than no link. Given our Gradle build already automatically downloads adopt if Java 11 or 8 isn't installed, I am leaning strongly for not making a change to this url. I also wouldn't revert the change to using Gradle's toolchain feature to automatically download adopt because automatically downloading a missing JDK is way more user friendly than not.

What do you think?

anuraaga: Also somewhat random but if someone can help with https://github.com/diffplug/spotless/issues/724 than any Java would work with the build. I want to do it someday (cloned spotless repo :) but probably still will take some time. After that we can remove this message - if anyone has strong issues with the message I encourage them to help with the issue since this is OSS :D

jkwatson: How about mentioning several options, so it doesn't look like we're endorsing a single JDK distro?

trask: I think it's ok to include a link to AdoptOpenJDK. It's great to have a vendor neutral option. And this is only a recommendation for building the project (not for production usage where TCK compliance would be more of an influence).

To make clear that we don't specifically require AdoptOpenJDK:

`Please download it from https://adoptopenjdk.net/.`

==>

`One option is to download it from https://adoptopenjdk.net/.`

trask: thought I had seen news on the AdoptOpenJDK TCK front, found it:

https://blog.adoptopenjdk.net/2020/06/adoptopenjdk-to-join-the-eclipse-foundation/

https://projects.eclipse.org/projects/adoptium/governance:

> Scope: 
Eclipse Adoptium provides runtime binaries that are high performance, enterprise-caliber, cross-platform, open-source licensed, Java SE TCK-tested and compliant for general use across the Java ecosystem

jkwatson: the real facepalm is the name "Adoptium"

carlosalberto (approved):

carlosalberto:
Not a strong opinion on the default recommendation - the actual error message will be super helpful though.

jkwatson (approved):
Thanks!

build.gradle
FadelIsmail: > ```suggestion
>             "JDK 11 installation.")
> ```

PR #2082, opened 2020-11-16
Stub out nebula configuration if .git doesn't exist.

anuraaga:


build.gradle
```diff
@@ -14,17 +14,24 @@ ext {
     opentelemetryProjects = subprojects - project(":opentelemetry-bom")
 }
 
-release {
-    defaultVersionStrategy = Strategies.getSNAPSHOT()
-}
+// Nebula plugin will not configure if .git doesn't exist, let's allow building on it by stubbing it
+// out.
+def releaseTask
+if (file('.git').exists()) {
```
Oberon00, line 20: This looks a bit brittle. I think something like `project.hasProperty('release')` would be better.

iNikem:
When this may be needed? How did you manage to get this repo without `.git` folder? :)

build.gradle
anuraaga: I think we'll always have .git - on the flip side, I used this check because I want to explicitly handle it being missing, without accidentally handling a case where `.git` is actually there and `release` isn't for whatever reason which isn't ever expected.

Oberon00:
I think this is about https://github.com/open-telemetry/opentelemetry-java/issues/2080
You can do that by pressing the download button at GitHub for example (which is also shown with every release, e.g. see the "Assets" section at the end of https://github.com/open-telemetry/opentelemetry-java/releases/tag/v0.10.0

anuraaga (commented):
@iNikem See #2080 - it can happen when downloading the zip archive from GitHub which we may as well support.

build.gradle
Oberon00: I just checked and it is at least there in a `git worktree`. I guess it is also there when checked out as `git submodule` so probably you are right. But there might be cases I haven't thought of :smiley:

jkwatson (approved):
thanks!

build.gradle
```diff
@@ -14,17 +14,24 @@ ext {
     opentelemetryProjects = subprojects - project(":opentelemetry-bom")
 }
 
-release {
-    defaultVersionStrategy = Strategies.getSNAPSHOT()
-}
+// Nebula plugin will not configure if .git doesn't exist, let's allow building on it by stubbing it
+// out.
```
trask, line 18:
```suggestion
// Nebula plugin will not configure if .git doesn't exist, let's allow building on it by stubbing it
// out. This supports building from the zip archive downloaded from GitHub.
```

anuraaga: Nice!

PR #2084, opened 2020-11-17
Ignore invalid usage of TraceState instead of throwing.

anuraaga:
Similar to #1964 and because of https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/error-handling.md#basic-error-handling-principles

### DIFF
--- a/api/src/main/java/io/opentelemetry/api/trace/TraceState.java
+++ b/api/src/main/java/io/opentelemetry/api/trace/TraceState.java
@@ -137,5 +137,4 @@ public abstract class TraceState {
 
     Entry() {}
   }
-
 }
--- a/api/src/main/java/io/opentelemetry/api/trace/TraceState.java
+++ b/api/src/main/java/io/opentelemetry/api/trace/TraceState.java
@@ -6,10 +6,8 @@
 package io.opentelemetry.api.trace;
 
 import com.google.auto.value.AutoValue;
-import io.opentelemetry.api.internal.Utils;
 import java.util.Collections;
 import java.util.List;
-import java.util.Objects;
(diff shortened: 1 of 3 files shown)


jkwatson (approved):
Thanks!

Oberon00:
I don't like how we completely ignore that, and not even log anything. That might lead to difficult to debug problems. I guess it's better than always throwing though.

jkwatson:
> I don't like how we completely ignore that, and not even log anything. That might lead to difficult to debug problems. I guess it's better than always throwing though.

I'd be ok with logging at FINE or FINEST, but I'd hate to introduce a logging-based DOS attack vector via someone sending bad trace state headers.

Oberon00:
Yes, I agree this is something that should not be logged by default.

anuraaga:
I think we need to do the same for many different parameters we don't validate, baggage, span, etc. I don't think it's a simple as adding `Logger` into each class, it's better to have a central class for logging validation messages I think - it will make it easier for users to replace them with exceptions or raise their level by themselves, etc. So can we do it separately in a consistent way?

jkwatson:
> I think we need to do the same for many different parameters we don't validate, baggage, span, etc. I don't think it's a simple as adding `Logger` into each class, it's better to have a central class for logging validation messages I think - it will make it easier for users to replace them with exceptions or raise their level by themselves, etc. So can we do it separately in a consistent way?

Having one named logger for validation messages seems like a good idea. And, yes, it definitely doesn't need to be in this PR.

jkwatson:
Logged #2093 to track the logging of API mis-use.

PR #2085, opened 2020-11-17
OpenCensus to OpenTelemetry metric exporter

zoercai:
The [OpenTelemetry merger announcement](https://opensource.googleblog.com/2019/05/opentelemetry-merger-of-opencensus-and.html) promised straightforward backwards compatibility with OpenCensus.

This PR adds an OpenCensus bridge for metrics, which allows applications and libraries that are instrumented with OpenTelemetry, but depend on other libraries instrumented with OpenCensus, to export metrics from both OpenTelemetry and OpenCensus.

The bridge maps all OpenCensus metric data to OpenTelemetry metric data, which are then exported by any configured OpenTelemetry exporter.

For OpenCensus spans to be exported as OpenTelemetry spans, this bridge needs to be added as a dependency. An extra line of configuration is also necessary in addition to configuring the OpenTelemetry exporter. For example

```
// Normal OpenTelemetry exporter configuration
LoggingMetricExporter loggingMetricExporter = new LoggingMetricExporter();
// Extra line, creating the bridge exporter, with the OpenTelemetry exporter
OpenTelemetryMetricsExporter exporter = OpenTelemetryMetricsExporter.createAndRegister(loggingMetricExporter);
```

cc @nilebox @james-bebbington


anuraaga:
Just curious is there anything special about metrics that let's us handle it at the export layer instead of instrumentation layer like we did for traces?

zoercai:
@anuraaga mainly that metrics don't have parent-child relationships, which was the main reason for not handling traces at the export layer. Label propagation hasn't been implemented in OpenTelemetry either (https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#baggage-into-metric-labels) so it seems that it would be fine (and much easier) to export the metric data output rather than modifying the context.

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryMetricsExporter.java
```diff
@@ -0,0 +1,198 @@
...
+          }
+        }
+        MetricData.Type metricDataType = mapType(type);
+        if (metricDataType != null) {
+          // TODO(@zoercai): resource, attributes, instrumentationLibraryInfo need to be
+          // added/mapped properly
+          metricData.add(
+              MetricData.create(
+                  Resource.getDefault(),
+                  InstrumentationLibraryInfo.create("das", "1"),
```
anuraaga, line 154: Maybe can use `io.opentelemetry.opencensusshim` as name at least for now. Let's go ahead and extract a constant for the InstrumentationLibraryInfo too

anuraaga (commented):
Thanks

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryMetricsExporter.java
```diff
@@ -0,0 +1,206 @@
...
+                          .getValue()
+                          .match(
+                              arg -> null,
+                              arg -> null,
+                              arg -> null,
+                              OpenTelemetryMetricsExporter::mapPercentiles,
+                              arg -> null)));
+              break;
+            default:
+              LOGGER.warning(type + " not supported by OpenCensus to OpenTelemetry migrator.");
```
nilebox, line 151: This message will be printed for every data point - maybe we should aggregate detected unsupported types and print one message per export?

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryMetricsExporter.java
```diff
@@ -0,0 +1,212 @@
...
+                .build());
+    intervalMetricReader =
+        IntervalMetricReader.create(
+            this, reader, options.setExportInterval(exportInterval).build());
+  }
+
+  @Override
+  public void export(Collection<Metric> metrics) {
+    ArrayList<MetricData> metricData = new ArrayList<>();
+    ArrayList<MetricDescriptor.Type> unsupportedTypes = new ArrayList<>();
```
nilebox, line 75: I think this should be a `HashSet`, otherwise it may contain duplicate types from every data point.
```suggestion
    Set<MetricDescriptor.Type> unsupportedTypes = new HashSet<>();
```

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryMetricsExporter.java
```diff
@@ -0,0 +1,212 @@
...
+                .setSpanName(EXPORTER_NAME)
+                .build());
+    intervalMetricReader =
+        IntervalMetricReader.create(
+            this, reader, options.setExportInterval(exportInterval).build());
+  }
+
+  @Override
+  public void export(Collection<Metric> metrics) {
+    ArrayList<MetricData> metricData = new ArrayList<>();
```
nilebox, line 74: nit: here and in other places: `List<>` (interface) is more idiomatic for variable type.
```suggestion
    List<MetricData> metricData = new ArrayList<>();
```

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryMetricsExporter.java
```diff
@@ -0,0 +1,212 @@
...
+    ArrayList<MetricDescriptor.Type> unsupportedTypes = new ArrayList<>();
+    for (Metric metric : metrics) {
+      for (TimeSeries timeSeries : metric.getTimeSeriesList()) {
+        LabelsBuilder labelsBuilder = Labels.builder();
+        for (int i = 0; i < metric.getMetricDescriptor().getLabelKeys().size(); i++) {
+          labelsBuilder.put(
+              metric.getMetricDescriptor().getLabelKeys().get(i).getKey(),
+              Objects.requireNonNull(timeSeries.getLabelValues().get(i).getValue()));
```
nilebox, line 82: Are you sure it's not possible to have a "null" label value in OpenCensus? Maybe it has to be an empty string in that case (then this code is fine)?
Maybe not, just double checking.

zoercai: Ah good catch!

> Returns the value of this {@link LabelValue}. Returns {@code null} if the value is unset and supposed to be ignored.

I've updated the code now to only add a label if it's not null.

anuraaga (approved):
LGTM

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryMetricsExporter.java
```diff
@@ -0,0 +1,210 @@
...
+            labelsBuilder.put(
+                metric.getMetricDescriptor().getLabelKeys().get(i).getKey(),
+                timeSeries.getLabelValues().get(i).getValue());
+          }
+        }
+        Labels labels = labelsBuilder.build();
+        List<MetricData.Point> points = new ArrayList<>();
+        MetricDescriptor.Type type = null;
+        for (Point point : timeSeries.getPoints()) {
+          long timestampNanos =
```
jkwatson, line 92: I'd like to see the contents of the loop extracted into a method, to make this code a little easier to understand.

jkwatson (commented):
Just one small refactoring needed; otherwise looks good to go.

jkwatson (approved):
Thanks!

PR #2089, opened 2020-11-18
Hide default context propagators implementation behind interface.

anuraaga:
This seems to be more consistent with our other default implementations like DefaultTracer.


jkwatson:
> I realized if we removed `ContexrPropagators` and only used `TextMapPropagator` with a multi-implementation it becomes consistent with span processor/exporter. If we plan to extend the interface with more methods, we should keep it but does anyone know if there are currently any ideas for that?

I know there are plans for BinaryPropagators, but I haven't been following when that might land.

Oberon00:
The spec issue seems rather dormant: https://github.com/open-telemetry/opentelemetry-specification/issues/437

context/src/main/java/io/opentelemetry/context/propagation/ContextPropagatorsBuilder.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package io.opentelemetry.context.propagation;
+
+/**
+ * A builder of a default {@link ContextPropagators} which can have multiple {@link
+ * TextMapPropagator} added, and extraction and injection will execute every {@link
+ * TextMapPropagator} in order.
+ */
+public interface ContextPropagatorsBuilder {
```
jkwatson, line 13: Having an interface for a builder, which is just a list of propagators and a constructor call feels like overkill to me. Is the rationale here to make the agent more easily able to bridge the builder? 

Could we ditch the builder entirely and use a pattern like in #2091 where we just have a static method on the ContextPropagators that takes the list of TextMapPropagators?

anuraaga: Yup - it came to mind after sending the PR but before also having the idea of killing `ContextPropagators` completely :) Will do so since now I see the binary propagation idea is probably the main reason for this interface.

anuraaga:
Thanks - I made some bigger changes, curious what people think.

`TextMapPropagator.composite` instead of doing so on `ContextPropagators` - while it does make registering multiple propagators a bit more annoying, it seems to reflect the actual semantics better. In particular, I can imagine us adding `ContextPropagators.create(TextMapPropagator, BinaryMapPropagator)` in the future which is very easy with this API. What do you think?

context/src/main/java/io/opentelemetry/context/propagation/ContextPropagators.java
```diff
@@ -71,25 +73,29 @@
...
-   * ContextPropagators propagators = DefaultContextPropagators.builder()
-   *     .addTextMapPropagator(new HttpTraceContext())
-   *     .addTextMapPropagator(new HttpBaggage())
-   *     .addTextMapPropagator(new MyCustomContextPropagator())
-   *     .build();
+   * ContextPropagators propagators = ContextPropagators.create(
+   *   TextMapPropagator.composite(
+   *     HttpTraceContext.getInstance(),
+   *     W3CBaggagePropagator.getInstance(),
+   *     new MyCustomContextPropagator()));
```
jkwatson, line 86: this usage looks really good to me.  👍

jkwatson (approved):
Thanks!

context/src/main/java/io/opentelemetry/context/propagation/ContextPropagators.java
carlosalberto: I preferred the `Builder` but hiding the default instance looks great. On that note, `TraceMultiPropagator` _may_ need to be lifted too, i.e. no `Builder`.

anuraaga: Thanks for calling out - let me followup with consistency in `TraceMultiPropagator` in a bit

PR #2091, opened 2020-11-18
Hide implementation of MultiSpanExporter/Processor behind interface.

anuraaga:
`forwarding` also came to mind instead of `delegating`. Either way, seems nice to hide the implementation.

### DIFF
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/SpanProcessor.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/SpanProcessor.java
@@ -36,6 +36,12 @@ public interface SpanProcessor {
     for (SpanProcessor processor : processors) {
       processorsList.add(processor);
     }
+    if (processorsList.isEmpty()) {
+      return NoopSpanProcessor.getInstance();
+    }
+    if (processorsList.size() == 1) {
+      return processorsList.get(0);
+    }
     return MultiSpanProcessor.create(processorsList);
   }
 
--- a/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/SpanProcessor.java
+++ b/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/SpanProcessor.java
@@ -22,8 +22,8 @@ public interface SpanProcessor {
    * Returns a {@link SpanProcessor} which simply delegates all processing to the {@code processors}
    * in order.
    */
-  static SpanProcessor delegating(SpanProcessor... processors) {
-    return delegating(Arrays.asList(processors));
+  static SpanProcessor composite(SpanProcessor... processors) {
+    return composite(Arrays.asList(processors));
   }
 
   /**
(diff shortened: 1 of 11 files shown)


Oberon00:
This increases coupling (previously MultiSpanExporter referenced SpanExporter but not the other way round) and makes a class `@Deprecated`. I wouldn't do that nearly breaking change just because it "seems nice".

anuraaga:
@Oberon00 I don't think any of the SDK classes in particular are considered released currently.

> This increases coupling (previously MultiSpanExporter referenced SpanExporter but not the other way round)

The coupling seems fine since we're not exposing `MultiSpanExporter` publicly anymore - how it's implemented, whether it's referencing `SpanExporter`, uses autovalue or becomes just some helper library or something becomes hidden from the user completely, this is good for reducing the API surface right?

anuraaga:
Also in `CONTRIBUTING.md` we do explicitly have a point of not exposing public classes when not needed

https://github.com/open-telemetry/opentelemetry-java/blob/master/CONTRIBUTING.md#best-practices-that-we-follow

jkwatson:
@Oberon00 we're going to be doing quite a bit of re-work in the SDK to clean up the configuration story, which is very scattered and inconsistent. Until GA, SDK changes are going to happen, and we're hoping to signal the changes by deprecating in one release then deleting in the next, where possible.

@anuraaga I don't love the name "delegating" because we're doing more than just delegating here. But, I also can't think of anything particularly better, either. :)

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/SpanProcessor.java
```diff
@@ -8,12 +8,37 @@
...
+  static SpanProcessor delegating(SpanProcessor... processors) {
+    return delegating(Arrays.asList(processors));
+  }
+
+  /**
+   * Returns a {@link SpanProcessor} which simply delegates all processing to the {@code processors}
+   * in order.
+   */
+  @SuppressWarnings("deprecation")
+  static SpanProcessor delegating(Iterable<SpanProcessor> processors) {
```
jkwatson, line 34: `allOf()` ?

jkwatson: or `ofAll()` ?  😆

jkwatson (approved):

sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/SpanProcessor.java
Oberon00: `composite` or `multi`, `exportEach`

jkwatson: I like `composite` or `composedOf`

carlosalberto: `composite` or `multi` 😄

anuraaga: composite has votes from everyone including me so there we go :)

anuraaga:
I added a largish change to add private `NoopSpanExporter` and have the `composite` methods return an optimal thing based on the number of inputs.

carlosalberto (approved):

jkwatson (approved):
Thanks!

PR #2095, opened 2020-11-19
Remove usages of beta Guava APIs.

anuraaga:
Long term we have #1944 but in the meantime, we definitely can't be using beta APIs so removed them and added a checker.

### DIFF
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/common/DaemonThreadFactory.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/common/DaemonThreadFactory.java
@@ -5,7 +5,7 @@
 
 package io.opentelemetry.sdk.common;
 
-import com.google.common.util.concurrent.MoreExecutors;
+import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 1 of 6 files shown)


sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -373,7 +372,9 @@ private void addTimedEvent(Event timedEvent) {
         logger.log(Level.FINE, "Calling addEvent() on an ended Span.");
         return;
       }
-      events.add(timedEvent);
+      if (events.size() < traceConfig.getMaxNumberOfEvents()) {
```
anuraaga, line 375: Spec just says discard after reaching the limit, which is closer to what this code does than the previous FIFO queue (though personally I think either is fine)

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#limits-on-span-collections

sdk/common/src/main/java/io/opentelemetry/sdk/common/DaemonThreadFactory.java
```diff
@@ -23,7 +23,7 @@ public DaemonThreadFactory(String namePrefix) {
 
   @Override
   public Thread newThread(Runnable runnable) {
-    Thread t = MoreExecutors.platformThreadFactory().newThread(runnable);
+    Thread t = Executors.defaultThreadFactory().newThread(runnable);
```
anuraaga, line 26: Previous code has some checks for Google App Engine - no clue what the delta is but this seems reasonable enough.

jkwatson: I peeked at this a few weeks ago. It looks like it's old app-engine stuff, so I'm not too worried about it.

jkwatson (approved):
Thanks!

PR #2096, opened 2020-11-19
Fix toBuilder not unobfuscating tracer.

anuraaga:
Goes to show how useful tobuilder was without a proper test for it...

@jkwatson I know it's deprecated, but we may want to cherrypick this since `OpenTelemetry.setGlobalPropagators` fails after installing the SDK because of this right now.


jkwatson:
If someone is broken by this, it'll be fixed by not using the deprecated method, yes? That doesn't seem like it's urgent to fix, in that case.

jkwatson:
oh, I see what you're saying...before we deprecated it, it was broken, and since it was used internally, it was broken for users. Hm. Yes, definitely an annoying bug. If we get someone logging it as a bug without a workaround, we can definitely think about a patch release.

bogdandrutu:
Closing this since `toBuilder` was removed

PR #2100, opened 2020-11-19
force the version number until we can get the tagging fixed.

jkwatson:
should temporarily resolve #2098

PR #2101, opened 2020-11-19
Get the 0.11.0 tag properly in the history

### LINKED ISSUE #2098: 0.12.0 SNAPSHOT builds not working
It looks like they're still publishing `0.11.0-SNAPSHOT`, rather than the latest.

See the details in this build: https://github.com/open-telemetry/opentelemetry-java/actions/runs/373048710

jkwatson:
It also happens to fix a doc bug that snuck in previously.

Hopefully this resolves #2098 in a better way than #2100 does.


jkwatson:
never mind...this doesn't work

PR #2104, opened 2020-11-20
Add factory methods instead of builder for TraceMultiPropagator

anuraaga:
Also adds a `TextMapPropagator.noop` to use from them - if it's useful for our own extension, it should be useful for other custom propagators.


jkwatson:
needs a rebase to get the fixes from the merge race-condition

extensions/trace-propagators/src/test/java/io/opentelemetry/extension/trace/propagation/TraceMultiPropagatorTest.java
```diff
@@ -142,7 +132,7 @@ void inject_allFormats() {
 
   @Test
   void extract_noPropagators() {
-    TextMapPropagator prop = TraceMultiPropagator.builder().build();
+    TextMapPropagator prop = TraceMultiPropagator.create();
```
jkwatson, line 135: this is an odd use. Do we think maybe this is one case where we could throw an exception? An empty multi-propagator is a very strange beast.

carlosalberto: Agreed. Now that the Builder pattern is not used, we should prevent this from happening.

anuraaga: This is a bit tricky. I agree that this usage `.create()` has no use case. But I think these do

```java
TraceMultiPropagator.create(available.stream().filter(this::isEnabled).toArray(TextMapPropagator[]::new));
TraceMultiPropagator.create(available.stream().filter(this::isEnabled).collect(toList()));
```

and I generally would not require the result of such an expression to be non-empty. What do you think?

jkwatson: It's definitely not harmful to have an empty multi-propagator. I was pondering if users might accidentally create one with nothing in it and be confused. There is a decent use-case for it, though, as you point out. I'm ok with it; but let's keep an eye out for confused users.

jkwatson (approved):

jkwatson:
needs a rebase to pick up a fix

jkwatson:
> needs a rebase to pick up a fix

ping @anuraaga

anuraaga:
@jkwatson Sorry have been heads down in agent-land, finally rebased

PR #2113, opened 2020-11-20
Change the default port that the OTLP exporters point to.

### LINKED ISSUE #2110: Update the default port for OTLP exporters according to spec
See this change to the spec: https://github.com/open-telemetry/opentelemetry-specification/pull/1221

jkwatson:
Resolves #2110

### DIFF
--- a/exporters/otlp/src/main/java/io/opentelemetry/exporter/otlp/OtlpGrpcSpanExporter.java
+++ b/exporters/otlp/src/main/java/io/opentelemetry/exporter/otlp/OtlpGrpcSpanExporter.java
@@ -70,11 +70,12 @@ import javax.annotation.concurrent.ThreadSafe;
 @ThreadSafe
 public final class OtlpGrpcSpanExporter implements SpanExporter {
 
-  public static final String DEFAULT_ENDPOINT = "localhost:55680";
+  public static final String DEFAULT_ENDPOINT = "localhost:4317";
   public static final long DEFAULT_DEADLINE_MS = TimeUnit.SECONDS.toMillis(1);
 
   private static final Logger logger = Logger.getLogger(OtlpGrpcSpanExporter.class.getName());
   private static final boolean DEFAULT_USE_TLS = false;
+  private static final String EXPORTER_NAME = OtlpGrpcSpanExporter.class.getSimpleName();
 
   private final TraceServiceFutureStub traceService;
   private final ManagedChannel managedChannel;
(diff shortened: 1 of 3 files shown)


exporters/otlp/src/main/java/io/opentelemetry/exporter/otlp/OtlpGrpcSpanExporter.java
```diff
@@ -111,8 +112,7 @@ private OtlpGrpcSpanExporter(ManagedChannel channel, long deadlineMs) {
    */
   @Override
   public CompletableResultCode export(Collection<SpanData> spans) {
-    spansSeen.add(
-        spans.size(), Labels.of("exporter", OtlpGrpcMetricExporter.class.getSimpleName()));
+    spansSeen.add(spans.size(), Labels.of("exporter", OtlpGrpcSpanExporter.class.getSimpleName()));
```
anuraaga, line 115: Thanks for the fixes! While we're here, should we extract constants for constant labels?

anuraaga: Though as an aside, after seeing some metrics code recently, I sort of wish constant labels could be specified when creating a meter instead of only when observing to save the tedium of always extracting constants to save allocation

anuraaga (approved):

exporters/otlp/src/main/java/io/opentelemetry/exporter/otlp/OtlpGrpcSpanExporter.java
jkwatson: This was a feature in the API that we decided to remove from the API for now. It may very well come back later, but we wanted to limit the features pre-GA, I think.

jkwatson: yeah, I guess this could just be a constant for the labels instance, eh? Very happy to make that happen, either as a follow-up, or part of this PR.

anuraaga: Either's fine with me this is approved :)

jkwatson: easy enough to do, so I just amended this PR with the constant label extraction

PR #2114, opened 2020-11-20
add a couple of context tests around keys with colliding hashcodes

jkwatson:


anuraaga (approved):
Good tests - just curious was there any trigger for them? For background I think in grpc-java they'd actually conflict, I changed from `Object.hashCode` to `System.identityHashCode` noticing it seemed weird.

jkwatson:
> Good tests - just curious was there any trigger for them? For background I think in grpc-java they'd actually conflict, I changed from `Object.hashCode` to `System.identityHashCode` noticing it seemed weird.

I was looking at code coverage and saw that we were missing a bunch of coverage on colliding hashcodes in the Context impl. But, I also realized that this doesn't actually exercise that, since we're using identityHashCode. I did ponder writing a test that found 2 colliding identityHashCodes, but it generally took a minute or so to fine two objects that collided, so I thought it might not be a good thing to have in the CI test suite. 👍

anuraaga (commented):
Ah yup I couldn't add those tests because of the implementation, which I guess is safer than allowing overridable hashes.

PR #2115, opened 2020-11-21
Update dependencies

anuraaga:
Most changes are related to

- Deprecated methods
- Error prone not allowing `atMost(blah, seconds)`
- Strange behavior in protobuf plugin. It caused me to notice that guava_testlib is applied globally which seems a bit heavyweight, I tried to narrow it down. The best result of that is we don't continue to have a global dependency on `junit4` (we only really need it know for `ZipkinRule` and wiremock

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,16 +1,16 @@
 pluginManagement {
     plugins {
-        id "com.diffplug.spotless" version "5.7.0"
-        id "com.github.ben-manes.versions" version "0.20.0"
+        id "com.diffplug.spotless" version "5.8.2"
+        id "com.github.ben-manes.versions" version "0.36.0"
         id "com.github.kt3k.coveralls" version "2.0.1"
-        id "com.google.protobuf" version "0.8.8"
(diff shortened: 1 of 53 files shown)


settings.gradle
```diff
@@ -1,16 +1,16 @@
 pluginManagement {
     plugins {
-        id "com.diffplug.spotless" version "5.7.0"
-        id "com.github.ben-manes.versions" version "0.20.0"
+        id "com.diffplug.spotless" version "5.8.2"
+        id "com.github.ben-manes.versions" version "0.36.0"
         id "com.github.kt3k.coveralls" version "2.0.1"
-        id "com.google.protobuf" version "0.8.8"
-        id "com.jfrog.artifactory" version "4.13.0"
-        id "com.jfrog.bintray" version "1.8.4"  // Need *specific* version.
```
anuraaga, line 8: I don't know what the comment is referring to, I figure this is a safe upgrade.

jkwatson: let's hope!

jkwatson (approved):
Thanks for the maintenance!

PR #2116, opened 2020-11-23
Rename HttpTraceContext to W3CTraceContextPropagator.

Oberon00:
Rename HttpTraceContext to W3CTraceContextPropagator, implement interface only in inner class, leave deprecated HttpTraceContext behind.


api/src/main/java/io/opentelemetry/api/trace/propagation/W3CTraceContextPropagator.java
```diff
@@ -0,0 +1,239 @@
...
+  private static final int TRACEPARENT_HEADER_SIZE = TRACE_OPTION_OFFSET + TRACE_OPTION_HEX_SIZE;
+  private static final int TRACESTATE_MAX_SIZE = 512;
+  private static final int TRACESTATE_MAX_MEMBERS = 32;
+  private static final char TRACESTATE_KEY_VALUE_DELIMITER = '=';
+  private static final char TRACESTATE_ENTRY_DELIMITER = ',';
+  private static final Pattern TRACESTATE_ENTRY_DELIMITER_SPLIT_PATTERN =
+      Pattern.compile("[ \t]*" + TRACESTATE_ENTRY_DELIMITER + "[ \t]*");
+  private static final Set<String> VALID_VERSIONS;
+  private static final String VERSION_00 = "00";
+  private static final TextMapPropagator INSTANCE =
```
Oberon00, line 69: Moving these tho the inner class, removing the interface implemenation and declaring getInstance as returning only `TextMapPropagator` instead of `W3CTraceContextPropagator` where the only changes here. Unfortunately it seems GitHub does not show that well.

carlosalberto:
Actually I'm wondering if we should instead rename `W3CBaggagePropagator` to `BaggagePropagator`, which is how it is called in Go/Python.

Oberon00:
The most important thing for me is that it does not end in `Context` 😃

api/src/main/java/io/opentelemetry/api/trace/propagation/HttpTraceContext.java
```diff
@@ -5,78 +5,17 @@
...
-import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
-/**
- * Implementation of the TraceContext propagation protocol. See <a
- * href=https://github.com/w3c/distributed-tracing>w3c/distributed-tracing</a>.
- */
 @Immutable
+@Deprecated
```
jkwatson, line 15: If you could add some javadoc with a pointer over to what the user should be using, that would be more user-friendly. Also on the `getInstance()` method, I think.

extensions/trace-propagators/src/test/java/io/opentelemetry/extension/trace/propagation/B3PropagatorTest.java
```diff
@@ -25,7 +26,7 @@
 import javax.annotation.Nullable;
 import org.junit.jupiter.api.Test;
 
-/** Unit tests for {@link io.opentelemetry.api.trace.propagation.HttpTraceContext}. */
+/** Unit tests for {@link W3CTraceContextPropagator}. */
```
jkwatson, line 29:
```suggestion
```

jkwatson (approved):
Thanks! I think this is a good naming change. I won't merge, though, until some others have had a chance to chime in.

anuraaga (approved):
Yeah was thinking of one of the renames, this seems good - thanks!

jkwatson:
@carlosalberto any objection going this route?

carlosalberto:
@jkwatson Let's merge. I will open one for _maybe_ removing the `W3C` prefix ;)

carlosalberto (approved):

PR #2118, opened 2020-11-23
Add a static builder method to the OpenTelemetry interface.

### LINKED ISSUE #1834: Using the tracing SDK module on its own has some usability issues

jkwatson:
Perhaps resolves #1834 

/cc @Oberon00 Does this close the loop on using the tracing SDK alone without the other pieces?

### DIFF
--- a/api/src/main/java/io/opentelemetry/api/OpenTelemetry.java
+++ b/api/src/main/java/io/opentelemetry/api/OpenTelemetry.java
@@ -216,4 +216,9 @@ public interface OpenTelemetry {
    */
   @Deprecated
   OpenTelemetryBuilder<?> toBuilder();
+
+  /** Returns a new {@link OpenTelemetryBuilder}. */
+  static OpenTelemetryBuilder<?> builder() {
+    return new DefaultOpenTelemetry.Builder();
+  }
 }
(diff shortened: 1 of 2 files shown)


Oberon00:
>Does this close the loop on using the tracing SDK alone without the other pieces?

I think so. 👍 

But why add this to the interface, when DefaultOpenTelemetry is already public? It creates the impression that there is only one way to build an OpenTelemetry object, when in fact this class is derived from in the sdk-all package.

jkwatson:
> > Does this close the loop on using the tracing SDK alone without the other pieces?
> 
> I think so. 👍
> 
> But why add this to the interface, when DefaultOpenTelemetry is already public? It creates the impression that there is only one way to build an OpenTelemetry object, when in fact this class is derived from in the sdk-all package.

I look at it the other way around. By attaching the method to the interface, we free ourselves in the future from having `DefaultOpenTelemetry` be the canonical implementation, and substitute with a different one. Having this method here doesn't stop us from making the builder on the `DOT` class public eventually as well, also.

anuraaga (approved):
Thanks, seems consistent with our other APIs too

PR #2120, opened 2020-11-23
Move the AsynchronousInstrument callbacks to the builders.

jkwatson:
Deprecate the existing setCallback method.
Update the code in the SDK to use the new API.

This will make it much more obvious that the callback can only be set once. I think it's more in line with what @kenfinnigan has been working on with #2068 .


thisthat (approved):
Thank you :) 
This also makes the code spec-compliant: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/sdk.md#asynchronous-sdk-instrument

api/src/main/java/io/opentelemetry/api/metrics/AsynchronousInstrument.java
```diff
@@ -34,13 +34,26 @@
...
+  interface Builder<R extends Result> extends Instrument.Builder {
+    /**
+     * Sets a callback that gets executed every collection interval.
+     *
+     * <p>Evaluation is deferred until needed, if this {@code AsynchronousInstrument} metric is not
+     * exported then it will never be called.
+     *
+     * @param callback the callback to be executed before export.
+     */
+    Builder<R> setCallback(Callback<R> callback);
```
anuraaga, line 53: Moving to Builder's seems like a great change. Just confirming, though not for this PR since it's the existing behavior, is `setCallback` good enough and not `addCallback` (more common in callback-oriented APIs I think)?

extensions/runtime-metrics/src/main/java/io/opentelemetry/extension/metrics/runtime/MemoryPools.java
```diff
@@ -70,30 +68,20 @@ public MemoryPools() {
 
   /** Export only the "area" metric. */
   public void exportMemoryAreaMetric() {
-    final LongUpDownSumObserver areaMetric =
-        this.meter
-            .longUpDownSumObserverBuilder("runtime.jvm.memory.area")
-            .setDescription("Bytes of a given JVM memory area.")
-            .setUnit("By")
-            .build();
-    areaMetric.setCallback(
```
anuraaga, line 79: Awesome!

anuraaga (approved):

api/src/main/java/io/opentelemetry/api/metrics/AsynchronousInstrument.java
jkwatson: In this API, you can only associate a single callback, so I think it makes sense to call it `set`.

extensions/runtime-metrics/src/main/java/io/opentelemetry/extension/metrics/runtime/MemoryPools.java
jkwatson: obviously, we'll want to port this change over to the instrumentation repo, since we're deprecating this code. Mostly did this to verify that it was a viable usage pattern. :)

bogdandrutu (approved):

bogdandrutu:
FYI this model may have troubles to support batch callback. Maybe add support for that before merging to prove that it doesn't affect?

jkwatson:
> FYI this model may have troubles to support batch callback. Maybe add support for that before merging to prove that it doesn't affect?

@kenfinnigan and I think this will make it easier to re-use the existing async instruments in the batch observer implementation. See #2068  for discussion.

kenfinnigan:
+1, this looks good to me

Issue #2033, opened 2020-11-06
Rewrite the Metrics SDK


jkwatson:
Our current Metrics SDK is fully functional and works. It is not, unfortunately, in alignment with the Metrics SDK specification, which is based on the golang SDK. Because of the divergence from the specification, it is difficult to understand the state of our implementation, and how it aligns (or does not) with the specification.

I propose we rewrite the metrics SDK to conform more closely to the specification.

kenfinnigan:
Do you have a specific example of where it differs?

jkwatson:
I actually have no example of where it doesn't differ. The implementations are completely different, as far as I can tell.

jkwatson:
Note: I've been roughing out a cleanroom implementation of the SDK based on the current specification here:  https://github.com/jkwatson/opentelemetry-java/tree/metrics_v2_1

/cc @carlosalberto

jkwatson:
Note: @bogdandrutu is doing some work to coerce our current implementation into something closer to what is in the SDK spec. Thanks, @bogdandrutu !

bogdandrutu:
@jkwatson what is the definition of done for this issue?

jkwatson:
Excellent question! I'd like to have a third-party (i.e. not you or me) take a pass through our implementation and make sure they can explain how it works, and how/if it does or doesn't map to the SDK specification.

jkwatson:
closing for now, as the metrics specifications are getting a re-work.

PR #2121, opened 2020-11-23
Metrics v2 first steel thread with a LongCounter

jkwatson:
If there is appetite for #2033, here is the very first "steel thread" that implements a LongCounter from end-to-end.

There are obviously tons of todos here, and lots of work to be done before this could be considered ready to use.


carlosalberto:
Nice!

thisthat (commented):
Thank you for taking care of this! I also think the metric SDK needs to be rewritten to be better aligned with the specification. I wonder if we should have this as an extension rather than an SDK package at the moment. Or maybe adding "beta" or other labels to the module name could also discourage users to think this is fully working and a replacement of the current metric SDK.

sdk/metrics-v2/build.gradle
```diff
@@ -0,0 +1,50 @@
...
+
+    testAnnotationProcessor libraries.auto_value
+    testCompileOnly libraries.auto_value_annotation
+
+    testCompile project(path: ':opentelemetry-sdk-common', configuration: 'testClasses')
+
+//    testImplementation project(':opentelemetry-sdk-testing')
+    testImplementation libraries.junit_pioneer
+
+    signature "net.sf.androidscents.signature:android-api-level-24:7.0_r2@signature"
```
anuraaga, line 27: gummy bears plz

anuraaga (commented):
@jkwatson Thanks for working on this - any recommendation on an ordering to review this PR in for better general context? Or just a line-by-line java review is enough?

jkwatson:
> @jkwatson Thanks for working on this - any recommendation on an ordering to review this PR in for better general context? Or just a line-by-line java review is enough?

I think I'd start by looking at the in-process SDK spec, and looking at the general structure. The line-by-line at this point is less important than the general architecture of the solution matching what is being specified. I would guess that most of the lines will end up being replaced at some point, as we evolve the solution towards being production ready.

jkwatson:
> Thank you for taking care of this! I also think the metric SDK needs to be rewritten to be better aligned with the specification. I wonder if we should have this as an extension rather than an SDK package at the moment. Or maybe adding "beta" or other labels to the module name could also discourage users to think this is fully working and a replacement of the current metric SDK.

Any and all suggestions welcome. At the moment, this isn't wired into the default SDK, so it would only end up being used if someone specifically added this module as a dependency to their application setup. Maybe we could not publish it anywhere until its ready, so no one could depend on it?

sdk/metrics-v2/build.gradle
jkwatson: done

jkwatson:
> > @jkwatson Thanks for working on this - any recommendation on an ordering to review this PR in for better general context? Or just a line-by-line java review is enough?
> 
> I think I'd start by looking at the work-in-progress SDK spec, and looking at the general structure. The line-by-line at this point is less important than the general architecture of the solution matching what is being specified. I would guess that most of the lines will end up being replaced at some point, as we evolve the solution towards being production ready.

jkwatson:
I'm going to close this for now, pending work on morphing our existing SDK to look more like the spec.

PR #2123, opened 2020-11-24
Update to Gradle 6.7.1

anuraaga:


jkwatson (approved):
any cool new features we should know about?

anuraaga:
Nah just a patch release, bugfixes with toolchain on OS X I think. We'll see in 6.8 :)

PR #2124, opened 2020-11-24
Don't repeat Java 8 support among all the subproject READMEs and at t…

anuraaga:
…op level, clarify that published artifacts are Java 8, but building requires additional instructions.

We used to have different compatibility among artifacts, but not anymore. We can expect some modules (like JFR) to have higher requirements but we don't need to write Java 8 support everywhere.

From #2080

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,11 @@ This project contains the following top level components:
 
 We would love to hear from the larger community: please provide feedback proactively.
 
+## Requirements
+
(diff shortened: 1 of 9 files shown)


Oberon00:
Do we ever publish these READMEs separately?

anuraaga:
@Oberon00 Don't think so, can't imagine ever doing so either.

Oberon00 (approved):

jkwatson (approved):
Thanks!

PR #2130, opened 2020-11-25
Some more test coverage for OpenTelemetrySDK.

anuraaga:


sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
```diff
@@ -219,10 +219,8 @@ public OpenTelemetrySdk build() {
     private TracerSdkProvider buildTracerProvider() {
       TracerProvider tracerProvider = super.tracerProvider;
       if (tracerProvider != null) {
-        if (!(tracerProvider instanceof TracerSdkProvider)) {
-          throw new IllegalStateException(
-              "The OpenTelemetrySdk can only be configured with a TracerSdkProvider");
-        }
+        // setTracerProvider checks this.
+        assert tracerProvider instanceof TracerSdkProvider;
```
jkwatson, line 223: do we even need this call? won't the cast do the same thing, if someone figures out how to somehow jam the wrong thing in here?

anuraaga: Makes sense yeah

jkwatson (approved):
Thanks!

PR #2132, opened 2020-11-25
Add Generated annotation to SemanticAttributes.

anuraaga:
This will hopefully exclude it from code coverage.


anuraaga:
Oops I misread https://github.com/jacoco/jacoco/issues/831 this doesn't exclude from coverage. Don't think this helps anything so closing.

PR #2133, opened 2020-11-25
handle case insensitive logic for keys on context extraction in OpenTracing TracerShim

### LINKED ISSUE #2122: OpenTracing TracerShim doesn't handle case-insenstitive logic for keys on context extraction
Trace context can be injected using case insensitive logic for keys.

Then extraction will not be able to extract context because extraction is case-sensitive:
https://github.com/open-telemetry/opentelemetry-java/blob/master/opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/Propagation.java#L76

Proposed fix for `TextMapGetter`:
```
@Override
public String get(Map<String, String> carrier, String key) {
  for (Map.Entry<String, String> entry : carrier.entrySet()) {
    if (key.equalsIgnoreCase(entry.getKey())) {
      return entry.getValue();
    }
  }
  return null;
}
```

malafeev:
fix #2122

### DIFF
--- a/opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/Propagation.java
+++ b/opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/Propagation.java
@@ -73,7 +73,12 @@ final class Propagation extends BaseShimObject {
 
     @Override
     public String get(Map<String, String> carrier, String key) {
-      return carrier.get(key);
+      for (Map.Entry<String, String> entry : carrier.entrySet()) {
+        if (key.equalsIgnoreCase(entry.getKey())) {
+          return entry.getValue();
+        }
+      }
+      return null;
     }
   }
 }
--- a/opentracing-shim/src/test/java/io/opentelemetry/opentracingshim/testbed/clientserver/TestClientServerTest.java
+++ b/opentracing-shim/src/test/java/io/opentelemetry/opentracingshim/testbed/clientserver/TestClientServerTest.java
@@ -47,8 +47,18 @@ class TestClientServerTest {
   @Test
   void test() throws Exception {
     Client client = new Client(queue, tracer);
-    client.send();
+    client.send(false);
+    verify();
+  }
 
+  @Test
+  public void testUpperCaseKeys() throws Exception {
+    Client client = new Client(queue, tracer);
+    client.send(true);
+    verify();
+  }
+
+  private void verify() {
     await().atMost(Duration.ofSeconds(15)).until(finishedSpansSize(otelTesting), equalTo(2));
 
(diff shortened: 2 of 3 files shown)


opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/Propagation.java
```diff
@@ -73,7 +73,12 @@ private TextMapGetter() {}
 
     @Override
     public String get(Map<String, String> carrier, String key) {
-      return carrier.get(key);
+      for (Map.Entry<String, String> entry : carrier.entrySet()) {
```
carlosalberto, line 76: If we plan to iterate over all the values per `get()` call, we should get rid of the intermediate temporary `Map`.

malafeev: I didn't remove temporary map because we need to implement 
```
 @Override
public Iterable<String> keys(Map<String, String> carrier) {
      return carrier.keySet();
}
```

if carrier will be `TextMapExtract` then to get keys I will need to create something temporal...

opentracing-shim/src/test/java/io/opentelemetry/opentracingshim/testbed/clientserver/TestClientServerTest.java
```diff
@@ -47,17 +48,33 @@ void after() throws InterruptedException {
...
 
     List<SpanData> finished = otelTesting.getSpans();
     assertThat(finished).hasSize(2);
 
     finished = sortByStartTime(finished);
     assertThat(finished.get(1).getTraceId()).isEqualTo(finished.get(0).getTraceId());
-    assertThat(finished.get(0).getKind()).isEqualTo(Kind.CLIENT);
-    assertThat(finished.get(1).getKind()).isEqualTo(Kind.SERVER);
+    Kind firstSpanKind = finished.get(0).getKind();
+    if (firstSpanKind == Kind.CLIENT) {
```
jkwatson, line 71: do we need to `sortByStartTime` any more if we're not going to rely on the order? Also, why did this change make the times change?

malafeev: I removed sorting. I guess that client and server spans are created too fast and we have kind of floating of nano seconds. If I put `sleep(100)` in the client then sorted order is `client->server`, I preder to avoid adding `sleep`.

jkwatson (approved):
Thanks!

anuraaga (approved):

PR #2134, opened 2020-11-25
Convert Attributes to an interface

jkwatson:


api/src/main/java/io/opentelemetry/api/common/Attributes.java
```diff
@@ -20,16 +23,25 @@
...
 @Immutable
-public abstract class Attributes extends ImmutableKeyValuePairs<AttributeKey, Object>
-    implements ReadableAttributes {
-  private static final Attributes EMPTY = Attributes.builder().build();
+public interface Attributes extends ReadableAttributes {
 
   @AutoValue
   @Immutable
-  abstract static class ArrayBackedAttributes extends Attributes {
+  abstract class ArrayBackedAttributes extends ImmutableKeyValuePairs<AttributeKey, Object>
```
anuraaga, line 40: This class is public - can you move it to a top level package private class?

jkwatson: ooh. good point. yes, will do so.

api/src/main/java/io/opentelemetry/api/common/Attributes.java
```diff
@@ -20,16 +23,25 @@
  * <p>Null keys will be silently dropped.
  *
  * <p>Note: The behavior of null-valued attributes is undefined, and hence strongly discouraged.
+ *
+ * <p>Implementations of this interface *must* be immutable and have well-defined value-based
```
anuraaga, line 27: By the way - must they be immutable? Actually one of the, completely unsupported but wild, use cases that came up for me for this interface is mutable attributes in resource. Docker cgroups come to mind, but I wouldn't be surprised if there are some good use cases for process-related information that is updatable at runtime.

jkwatson: I'd much rather start by strongly suggesting immutability, then maybe relax it if we find it's ok for them to be mutable.

api/src/main/java/io/opentelemetry/api/common/Attributes.java
```diff
@@ -39,47 +51,65 @@
     public AttributesBuilder toBuilder() {
```
trask, line 51: in order for the javaagent to provide a single implementation that satisfies both the shaded and unshaded interfaces, we need the return value here to be an interface, so that its implementation use a covariant return type that satisfies both, similar to the `TraceState.getTraceState()` example in https://github.com/open-telemetry/opentelemetry-java/issues/1946#issuecomment-727531701

jkwatson:
@marcingrzejszczak FYI this is potentially a breaking change

api/src/main/java/io/opentelemetry/api/common/Attributes.java
jkwatson: done. please take a look

anuraaga (approved):

api/src/main/java/io/opentelemetry/api/common/Attributes.java
jkwatson: done

PR #2137, opened 2020-11-25
Convert Labels to be an interface.

jkwatson:


api/src/main/java/io/opentelemetry/api/common/Labels.java
```diff
@@ -34,31 +58,41 @@ public void forEach(BiConsumer<String, String> consumer) {
...
       }
     }
+
+    static Labels sortAndFilterToLabels(Object... data) {
+      return new AutoValue_Labels_ArrayBackedLabels(
+          sortAndFilter(data, /* filterNullValues= */ false));
+    }
+
+    @Override
+    public LabelsBuilder toBuilder() {
```
trask, line 68: same request as https://github.com/open-telemetry/opentelemetry-java/pull/2134/files#r530758547, would you mind making `LabelsBuilder` an interface?

jkwatson:
@marcingrzejszczak FYI this potentially a breaking change

api/src/main/java/io/opentelemetry/api/common/Labels.java
jkwatson: done. please take a look

anuraaga (approved):
Thanks!

PR #2138, opened 2020-11-25
Convert TraceState to be an interface

jkwatson:
This also gets rid of the un-needed Entry objects.

I wonder if we should do some renaming to prefix with `W3C` : `W3CTraceState` and `W3CTraceStateBuilder`.


api/src/main/java/io/opentelemetry/api/trace/TraceState.java
```diff
@@ -24,21 +21,40 @@
...
+ * <p>For these reasons, it is strongly suggested that you use the implementation that is provided
+ * here via the {@link TraceStateBuilder}.
  */
 @Immutable
-@AutoValue
-public abstract class TraceState {
-  private static final TraceState DEFAULT = TraceState.builder().build();
+public interface TraceState {
+
+  TraceState DEFAULT = new TraceStateBuilder().build();
```
anuraaga, line 38: Move this to a package private field in `ArrayBasedTraceState` so it's not public

api/src/main/java/io/opentelemetry/api/trace/TraceState.java
```diff
@@ -24,21 +21,40 @@
...
-  public static TraceState getDefault() {
+  static TraceState getDefault() {
     return DEFAULT;
   }
 
+  /**
+   * Returns a {@code Builder} based on an empty {@code TraceState}.
+   *
+   * @return a {@code Builder} based on an empty {@code TraceState}.
+   */
```
anuraaga, line 53:
```suggestion
  /**
   * Returns an empty {@code TraceStateBuilder}.
   */
```

api/src/main/java/io/opentelemetry/api/trace/TraceState.java
```diff
@@ -24,21 +21,40 @@
...
+  static TraceStateBuilder builder() {
```
anuraaga, line 54: I pondered a bit on whether this needs to return an interface too - but since it's static guess not

api/src/main/java/io/opentelemetry/api/trace/TraceState.java
```diff
@@ -24,21 +21,40 @@
...
+ *
+ * <p>Implementations of this interface *must* be immutable and have well-defined value-based
+ * equals/hashCode implementations. If an implementation does not strictly conform to these
+ * requirements, behavior of the OpenTelemetry APIs and default SDK cannot be guaranteed.
+ *
+ * <p>Implementations of this interface that do not conform to the W3C specification risk
+ * incompatibility with W3C-compatible implementations.
+ *
+ * <p>For these reasons, it is strongly suggested that you use the implementation that is provided
+ * here via the {@link TraceStateBuilder}.
```
anuraaga, line 33:
```suggestion
 * here via the {@link TraceState#builder}.
```

api/src/main/java/io/opentelemetry/api/trace/TraceStateBuilder.java
```diff
@@ -10,27 +10,31 @@
 import java.util.Objects;
 import javax.annotation.Nullable;
 
-/** A builder of {@link TraceState}. */
+/**
+ * A builder of {@link TraceState}. This implementation does full validation of the keys and values
```
anuraaga, line 14: Thinking about your question about w3c prefix, it's a bit weird. The w3c'ness is only in this builder, not the TraceState, not sure how best to deal with that. I wonder if we should remove the validation from this class and do it in the propagator?

api/src/main/java/io/opentelemetry/api/trace/TraceStateBuilder.java
```diff
@@ -50,20 +54,21 @@ public TraceStateBuilder set(String key, String value) {
...
-    TraceState.Entry entry = TraceState.Entry.create(key, value);
     if (entries == null) {
       // Copy entries from the parent.
       entries = new ArrayList<>(parent.getEntries());
     }
-    for (int i = 0; i < entries.size(); i++) {
-      if (entries.get(i).getKey().equals(entry.getKey())) {
+    for (int i = 0; i < entries.size(); i += 2) {
+      if (entries.get(i).equals(key)) {
+        entries.remove(i);
```
anuraaga, line 63: This loop has a lot of shadiness :) Relies on` entries.size()` being invoked on every loop iteration (true in every language, but not something people expect to matter I think), and two calls to the same `entries.remove(i)` ouch. Let's use `ListIterator.remove` instead.

api/src/main/java/io/opentelemetry/api/trace/TraceStateBuilder.java
```diff
@@ -50,20 +54,21 @@ public TraceStateBuilder set(String key, String value) {
...
         entries.remove(i);
         // Exit now because the entries list cannot contain duplicates.
         break;
       }
     }
     // Inserts the element at the front of this list.
-    entries.add(0, entry);
+    entries.add(0, key);
```
anuraaga, line 70: Inserting at the head of ArrayList looks like lots of overhead for the standard case of having no dupes, etc. Can we instead build up in opposite order and `Collections.reverse` in `build`? We'd need to reverse when copying from `parent.getEntries()` but I guess that's just a couple more lines of code.

api/src/main/java/io/opentelemetry/api/trace/TraceStateBuilder.java
```diff
@@ -81,8 +86,9 @@ public TraceStateBuilder remove(String key) {
       // Copy entries from the parent.
       entries = new ArrayList<>(parent.getEntries());
     }
-    for (int i = 0; i < entries.size(); i++) {
-      if (entries.get(i).getKey().equals(key)) {
+    for (int i = 0; i < entries.size(); i += 2) {
```
anuraaga, line 89: Looks like we can extract a private `doRemove` type of method that's used in both set and remove

api/src/test/java/io/opentelemetry/api/trace/TraceStateTest.java
```diff
@@ -217,11 +217,11 @@ void addAndUpdateEntry() {
...
                 .set(FIRST_KEY, SECOND_VALUE) // update the existing entry
                 .set(SECOND_KEY, FIRST_VALUE) // add a new entry
-                .build()
-                .getEntries())
-        .containsExactly(
-            TraceState.Entry.create(SECOND_KEY, FIRST_VALUE),
-            TraceState.Entry.create(FIRST_KEY, SECOND_VALUE));
+                .build())
+        .asInstanceOf(type(ArrayBasedTraceState.class))
+        .extracting(ArrayBasedTraceState::getEntries)
```
anuraaga, line 222: Ok ok, let me get `asMap` in soon :P

api/src/main/java/io/opentelemetry/api/trace/TraceState.java
anuraaga: @trask Actually can you check this - will we be able to do the interface interop if this doesn't return an interface?

api/src/main/java/io/opentelemetry/api/trace/TraceState.java
jkwatson: I tried this, but it ended up breaking with a class loading issue like the other one. :(

api/src/main/java/io/opentelemetry/api/trace/TraceStateBuilder.java
jkwatson: yeah...it's a weird thing to be first-class in the APIs, and not constrained to be part of the propagator, TBH. I was pondering only making this class prefixed with W3C, and leaving the TraceState itself to be generic.

api/src/main/java/io/opentelemetry/api/trace/TraceStateBuilder.java
jkwatson: we could also consider just waiting until the final build to de-dupe. There are options here, for sure. Luckily, it's all internal, so we can make it better whenever.

api/src/test/java/io/opentelemetry/api/trace/TraceStateTest.java
jkwatson: hah!

api/src/main/java/io/opentelemetry/api/trace/TraceState.java
anuraaga: What if it's in the builder? I think one of the two is supposed to work ok (or otherwise I guess another package private class)

api/src/main/java/io/opentelemetry/api/trace/TraceStateBuilder.java
anuraaga: I think I'd leave it as is then - the builder of TraceState being TraceStateBuilder seems clearer even with the w3c caveats.

api/src/main/java/io/opentelemetry/api/trace/TraceStateBuilder.java
anuraaga: Thought a bit more and realized the suggestion doesn't make much sense. Hmm - let's at least add comments if nothing comes to mind to clean up in this PR.

anuraaga (approved):
Agree we can keep on iterating separately

api/src/main/java/io/opentelemetry/api/trace/TraceState.java
trask: it's the `toBuilder()` method that hurts: https://github.com/open-telemetry/opentelemetry-java/pull/2134/files#r530758547

@jkwatson sorry to ask, would you mind turning these builders into interfaces?

Oberon00:
What about W3CSpanContext? 😉

jkwatson:
@marcingrzejszczak FYI this is potentially a breaking change

api/src/main/java/io/opentelemetry/api/trace/TraceState.java
jkwatson: If I were to put this into a separate class, what would I call it? TraceStateDefaultHolder?

jkwatson: I called it `DefaultTraceState`. Seems like an obvious enough name. :)

api/src/main/java/io/opentelemetry/api/trace/TraceState.java
jkwatson: done. please take a look

api/src/main/java/io/opentelemetry/api/trace/TraceStateBuilder.java
jkwatson: It doesn't actually rely on entries.size() being called each time, since we break as soon as we remove one. I'll extract out the size and add some comments.

api/src/main/java/io/opentelemetry/api/trace/TraceStateBuilder.java
jkwatson: we could put the working data into a `LinkedList` for easy insert efficiency, as well. there are definitely options here. Not sure how much I want to take on this PR, though.

anuraaga (approved):
Thanks!

Issue #2141, opened 2020-11-26
Add support for AWS Lambda resource attributes


mateuszrzeszutek:
**Is your feature request related to a problem? Please describe.**
Add the following resource attributes to the sdk-extensions/aws module:
* `cloud.provider` with value `"aws"`
* `cloud.region` with value extracted from the `AWS_REGION` env var
* `faas.name` with value extracted from the `AWS_LAMBDA_FUNCTION_NAME` env var
* `faas.version` with value extracted from the `AWS_LAMBDA_FUNCTION_VERSION` env var

**Describe the solution you'd like**
Create a new `ResourceProvider` implementation.

**Describe alternatives you've considered**
A similar provider could be added to the instrumentation repo, but since this one already contains some AWS-specific resource providers it seemed like a more proper place for that.

**Additional context**
* the instrumentation issue that's the motivation for this change: open-telemetry/opentelemetry-java-instrumentation#1773

PR #2142, opened 2020-11-26
Add support for AWS Lambda resource attributes

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java/issues/2141

### DIFF
--- a/sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource/LambdaResource.java
+++ b/sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource/LambdaResource.java
@@ -31,7 +31,7 @@ public final class LambdaResource extends ResourceProvider {
     String functionName = environmentVariables.getOrDefault("AWS_LAMBDA_FUNCTION_NAME", "");
     String functionVersion = environmentVariables.getOrDefault("AWS_LAMBDA_FUNCTION_VERSION", "");
 
-    if (!isLambda(region, functionName, functionVersion)) {
+    if (!isLambda(functionName, functionVersion)) {
       return Attributes.empty();
     }
 
--- a/sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource/LambdaResource.java
+++ b/sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource/LambdaResource.java
@@ -10,6 +10,7 @@ import io.opentelemetry.api.common.AttributesBuilder;
 import io.opentelemetry.api.trace.attributes.SemanticAttributes;
 import io.opentelemetry.sdk.resources.ResourceProvider;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


jkwatson (approved):
Thanks! @anuraaga can you give this a once-over as our AWS rep?

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource/LambdaResource.java
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.extension.aws.resource;
+
+import static com.google.common.base.Strings.isNullOrEmpty;
```
anuraaga, line 8: Can you rebase? I've removed the guava dependency from this module. You can use `getOrDefault` and just `isEmpty` instead I believe.

anuraaga (commented):
LGTM but can you rebase? Thanks!

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource/LambdaResource.java
mateuszrzeszutek: Done.

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource/LambdaResource.java
```diff
@@ -0,0 +1,54 @@
...
+      return Attributes.empty();
+    }
+    String region = environmentVariables.getOrDefault("AWS_REGION", "");
+    String functionName = environmentVariables.get("AWS_LAMBDA_FUNCTION_NAME");
+    String functionVersion = environmentVariables.getOrDefault("AWS_LAMBDA_FUNCTION_VERSION", "");
+
+    AttributesBuilder builder =
+        Attributes.builder()
+            .put(SemanticAttributes.CLOUD_PROVIDER, SemanticAttributes.CloudProviderValues.AWS)
+            .put(SemanticAttributes.FAAS_NAME, functionName);
```
anuraaga, line 39: Shouldn't we check empty for this too? It's not clear why we check for the others only but not this one.

mateuszrzeszutek: This one cannot be empty, it's being checked in the `isLambda()` function.

mateuszrzeszutek: I refactored it a bit so that all env vars are retrieved in exactly the same way.

anuraaga (approved):
Thanks!

Issue #1908, opened 2020-10-28
Support process.runtime.* resource attributes


Oberon00:
There should be support for the process.runtime.* resource attributes, which should contain information about the monitored JVM.

Spec: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/semantic_conventions/process.md#java-runtimes

Note that I made a spec PR to split process.runtime from process, so that these resources can be reported independently: https://github.com/open-telemetry/opentelemetry-specification/pull/1137
Since these values are trivial to gather (just `System.getProperty`), it might make sense to always report them out of the box along with the telemetry SDK.

Or instead it may make sense to add them to a provider separate from the current process provider.

Oberon00:
A provider-based implementation could look like this:

```java
public class ProcessRuntimeResourceProvider extends ResourceProvider {

  @Override
  protected Attributes getAttributes() {
    return Attributes.of(
        PROCESS_RUNTIME_NAME,
        System.getProperty("java.runtime.name"),
        PROCESS_RUNTIME_VERSION,
        System.getProperty("java.runtime.version"),
        PROCESS_RUNTIME_DESCRIPTION,
        System.getProperty("java.vm.vendor") + ' ' + System.getProperty("java.vm.version"));
  }
}
```java

jkwatson:
My opinion: we should report them by default.

Oberon00:
@anuraaga From some looking around, it seems that `java.vm.name` is usually more interesting and specific than `java.runtime.name`. Do you think the spec should be changed? E.g. OpenJDK-based JVMs seem to have `java.runtime.name=OpenJDK Runtime Environment` but report something at least a bit more specific in `java.vm.name`, e.g. on Windows I can see for the same Adopt OpenJDK 15 version  and `java.vm.name=OpenJDK 64-Bit Server VM` for 64 bit but `OpenJDK Client VM` for 32 bit.

anuraaga:
Here's my gist for reference

https://gist.github.com/anuraaga/c79219a7e4401515c3de998f9d077be0

I think one thing that lead me to the current properties is I think this allowed OpenJ9 to model well. It's still OpenJDK, but with different VM. So keeping `runtime.name` constant with the other OpenJDK I think works well.

anuraaga:
Admittedly, I thought client VM had already been deleted by now... But I'd also assume any use case for telemetry would force server mode on anyways, which I guess requires a flag on 32-bit?

[...]

PR #2143, opened 2020-11-26
Support process.runtime.* resource attributes

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java/issues/1908

### DIFF
--- a/sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ProcessRuntimeResource.java
+++ b/sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ProcessRuntimeResource.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.extension.resources;
+
+import static io.opentelemetry.sdk.resources.ResourceAttributes.PROCESS_RUNTIME_DESCRIPTION;
+import static io.opentelemetry.sdk.resources.ResourceAttributes.PROCESS_RUNTIME_NAME;
+import static io.opentelemetry.sdk.resources.ResourceAttributes.PROCESS_RUNTIME_VERSION;
+
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.sdk.resources.ResourceProvider;
+
+/** {@link ResourceProvider} which provides information about the Java runtime. */
+public final class ProcessRuntimeResource extends ResourceProvider {
+  @Override
+  protected Attributes getAttributes() {
+    try {
+      String name = System.getProperty("java.runtime.name");
+      String version = System.getProperty("java.runtime.version");
+      String description =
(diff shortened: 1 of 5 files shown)


sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ProcessRuntimeResource.java
```diff
@@ -0,0 +1,40 @@
...
+
+/** {@link ResourceProvider} which provides information about the Java runtime. */
+public final class ProcessRuntimeResource extends ResourceProvider {
+  @Override
+  protected Attributes getAttributes() {
+    try {
+      String name = System.getProperty("java.runtime.name");
+      String version = System.getProperty("java.runtime.version");
+      String description =
+          String.join(
```
Oberon00, line 23: I think I would slightly prefer using `+` here, but no strong opinion.

sdk-extensions/resources/src/test/java/io/opentelemetry/sdk/extension/resources/ProcessRuntimeResourceTest.java
```diff
@@ -0,0 +1,37 @@
...
+import org.junit.jupiter.api.Test;
+
+class ProcessRuntimeResourceTest {
+  @Test
+  void shouldCreateRuntimeAttributes() {
+    // when
+    ReadableAttributes attributes = new ProcessRuntimeResource().getAttributes();
+
+    // then
+    assertThat(attributes.get(ResourceAttributes.PROCESS_RUNTIME_NAME)).isNotNull();
```
Oberon00, line 22:
```suggestion
    assertThat(attributes.get(ResourceAttributes.PROCESS_RUNTIME_NAME)).isNotBlank();
```

Slightly stricter assertion: https://joel-costigliola.github.io/assertj/core/api/org/assertj/core/api/AbstractCharSequenceAssert.html#isNotBlank()

Oberon00 (approved):
I think that's nice: We have both a ResourceProvider and it is included in the default attributes.
The only problem I can see over "hard-coding" this like the telemetry SDK resource, is that the SPI approach might not work with some fat Jar approaches.

sdk-extensions/resources/src/test/java/io/opentelemetry/sdk/extension/resources/ProcessRuntimeResourceTest.java
mateuszrzeszutek: Good point, done.

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ProcessRuntimeResource.java
mateuszrzeszutek: I'm fine with both versions.

anuraaga (approved):
Nice

jkwatson (approved):
Thanks!

Issue #2139, opened 2020-11-26
buildscripts/generate.sh is broken


mateuszrzeszutek:
**Describe the bug**
1. It generates the `SemanticAttributes.java` in wrong directory, with wrong package
2. `./generate.sh: line 28: ./gradlew: No such file or directory`
3. The script behaves slightly differently depending from where it's called: `opentelemetry-specification` is always downloaded to the current working directory, but .gitignore only ignores it when it's in `buildscripts/semantic-convention`
4. For some reason the spec repo is frozen in time at sha1 662baae949f01a8ecc950426bf09283be1b657de - this block us from getting any recent changes in the semantic attributes.

**Steps to reproduce**
```sh
./buildscripts/semantic-convention/generate.sh
```

Oberon00:
> For some reason the spec repo is frozen in time at sha1 662baae949f01a8ecc950426bf09283be1b657de

This is intentional. We want the committed `generate.sh` to reproducibly generate `SemanticAttributes.java`. It is important that we know from which version of the spec this was generated.

mateuszrzeszutek:
Hmm, are we going to update it manually to a newer hash when new attributes are needed? I mean, that's perfectly good reason for keeping the hash hard-coded, we should add a comment in the script saying why it's done that way.

PR #2144, opened 2020-11-26
Fix buildscripts/semantic-convention/generate.sh script

mateuszrzeszutek:
Including:
* update spec version to latest commit
* regenerate SemanticAttributes

Resolves https://github.com/open-telemetry/opentelemetry-java/issues/2139

### DIFF
--- a/api/src/main/java/io/opentelemetry/api/trace/attributes/SemanticAttributes.java
+++ b/api/src/main/java/io/opentelemetry/api/trace/attributes/SemanticAttributes.java
@@ -739,6 +739,12 @@ public final class SemanticAttributes {
    */
   public static final AttributeKey<String> ENDUSER_SCOPE = stringKey("enduser.scope");
 
+  /** Current &#34;managed&#34; thread ID (as opposed to OS thread ID). */
+  public static final AttributeKey<Long> THREAD_ID = longKey("thread.id");
+
+  /** Current thread name. */
+  public static final AttributeKey<String> THREAD_NAME = stringKey("thread.name");
+
   /**
    * The method or function name, or equivalent (usually rightmost part of the code unit&#39;s
    * name).
--- a/api/src/main/java/io/opentelemetry/api/trace/attributes/SemanticAttributes.java
+++ b/api/src/main/java/io/opentelemetry/api/trace/attributes/SemanticAttributes.java
@@ -7,9 +7,11 @@ package io.opentelemetry.api.trace.attributes;
 
 import static io.opentelemetry.api.common.AttributeKey.booleanKey;
 import static io.opentelemetry.api.common.AttributeKey.longKey;
+import static io.opentelemetry.api.common.AttributeKey.stringArrayKey;
 import static io.opentelemetry.api.common.AttributeKey.stringKey;
 
 import io.opentelemetry.api.common.AttributeKey;
+import java.util.List;
 
 // DO NOT EDIT, this is an Auto-generated file from /templates/SemanticAttributes.java.j2
 public final class SemanticAttributes {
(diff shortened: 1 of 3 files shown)


iNikem (approved):

api/src/main/java/io/opentelemetry/api/trace/attributes/SemanticAttributes.java
```diff
@@ -96,7 +160,7 @@
 
   /**
    * The version string of the VM image as defined in [Version
-   * Attributes](https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/resource/semantic_conventions#version-attributes).
+   * Attributes](README.md#version-attributes).
```
jkwatson, line 163: this doesn't look right, if line 136 is right.

buildscripts/semantic-convention/generate.sh
```diff
@@ -1,30 +1,32 @@
 #!/bin/bash
 
-DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+ROOT_DIR="${SCRIPT_DIR}/../../"
+# freeze the spec version to make SemanticAttributes generation reproducible
+SPEC_VERSION=d328331ce9dc6dec871b647b4e37a84b5396516e
```
jkwatson, line 6: can we also reference the *version* of the spec that we're referencing, so it's clear we're not pulling in arbitrary SHAs?

api/src/main/java/io/opentelemetry/api/trace/attributes/SemanticAttributes.java
Oberon00: It is wrong, but that is currently a limitation of the md generator: https://github.com/open-telemetry/build-tools/issues/20

Oberon00: The links are currently relative regarding the place where the markdown tables are generated in the spec. Nothing we can do about it without fixing the tool first, then we can make the YAML right again.

buildscripts/semantic-convention/generate.sh
mateuszrzeszutek: But we are pulling arbitrary SHAs 😅 

We can use tags here instead of hashes, e.g. `v0.7.0` but I'm afraid that just using tags will be very limiting. Spec `v0.7.0` is pretty recent, but `v0.6.0` was released 5 months ago.

Oberon00: We could maybe add a comment with the `git describe` of that version next to it. (Or can we use `git describe` result directly?)

Oberon00: In this case it would be `v0.7.0-15-gd328331`. And git seems to be able to parse this as revision:
```
$git log v0.7.0-15-gd328331
commit d328331ce9dc6dec871b647b4e37a84b5396516e
Author: Christian Neumüller <christian.neumueller@dynatrace.com>
Date:   Thu Nov 26 13:15:29 2020 +0100

    Add .gitattributes & normalize line endings (#1230)

    Co-authored-by: Armin Ruech <armin.ruech@dynatrace.com>

commit 2aa62dfe1efff9d25d935c52d579c44418c925d8
Author: William Armiros <54150514+willarmiros@users.noreply.github.com>
Date:   Tue Nov 24 20:17:42 2020 -0600

    Add semantic conventions for cloud provider-specific resource attributes (#1099)
```

mateuszrzeszutek: Nice! That looks great, much better than sha1.

mateuszrzeszutek: Unfortunately it doesn't work

```
fatal: couldn't find remote ref v0.7.0-15-gd328331
```

jkwatson: I fine using the SHA, but would like to just have a comment about how we got it (i.e. from a release tag, etc).

buildscripts/semantic-convention/generate.sh
```diff
@@ -3,6 +3,8 @@
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 ROOT_DIR="${SCRIPT_DIR}/../../"
 # freeze the spec version to make SemanticAttributes generation reproducible
+# this hash was obtained by calling `git rev-parse master`
```
jkwatson, line 6: can we not lock this to the recent 0.7.0 release for now?

jkwatson: And if not, please reference the date that this version was pulled from.

mateuszrzeszutek: Added the date.

buildscripts/semantic-convention/generate.sh
```diff
@@ -3,9 +3,9 @@
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 ROOT_DIR="${SCRIPT_DIR}/../../"
 # freeze the spec version to make SemanticAttributes generation reproducible
-# this hash was obtained by calling `git rev-parse master`
+# this hash was obtained by calling `git rev-parse master` on 1st Dec 2020
 # we can't use a version tag here because otel-spec releases are very rare
```
jkwatson, line 7: fyi I brought this up in the spec meeting this morning, so it's at least on people's radar.

jkwatson (approved):
Thanks!

PR #2145, opened 2020-11-27
Add asMap for Attributes.

anuraaga:
If this makes sense, I'll follow up with the rest of the types.

Implementation copied from https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/internal/collect/UnsafeArrayMap.java

While it's possible to directly implement `Map`, I think `asMap` is better since it makes sure these are compile errors which I think they should be.

```java
Map<AttributeKey<?>, Object> attributes = Attributes.builder().build() // Users should be using Attributes type

Attributes attributes = Attributes.builder().build();
attributes.put() // Because of Map, these write operations leak into Attributes
```

`asMap` provides an intentional step to get into a read-only Map view. It's very unlikely Java would heap-allocate the wrapper in most code I think.

I considered making the unsupported operations no-op instead of throw, but in this case, I feel it's conventional enough to go with it (e.g., `Collections.emptyMap()` also throws Unsupported for all of these). But not too sure on it.

### DIFF
--- a/api/src/main/java/io/opentelemetry/api/common/Attributes.java
+++ b/api/src/main/java/io/opentelemetry/api/common/Attributes.java
@@ -23,7 +23,7 @@ import javax.annotation.concurrent.Immutable;
  */
 @SuppressWarnings("rawtypes")
 @Immutable
-public abstract class Attributes extends ImmutableKeyValuePairs<AttributeKey, Object>
+public abstract class Attributes extends ImmutableKeyValuePairs<AttributeKey<?>, Object>
     implements ReadableAttributes {
   private static final Attributes EMPTY = Attributes.builder().build();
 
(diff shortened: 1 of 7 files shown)


jkwatson (approved):
Looks good. Not sure what you mean about 'not heap-allocating the wrapper in most cases', but I think this is a fine addition.

anuraaga:
> Looks good. Not sure what you mean about 'not heap-allocating the wrapper in most cases', but I think this is a fine addition.

Just that escape analysis should basically always kick in for common usages like

`attributes.asMap().entrySet().stream().foobar`

I went into a history dive as part of this in particular since I considered reusing the same `Entry` object in `EntrySet`. It's a bit wonky and less safe than you'd expect to not let a user keep `Entry` objects but was considered OK until Java 1.6 due to the significant performance benefit. With Java 1.7, those optimizations were removed in favor of safety + escape analysis making sure the `Entry` stays on the stack for standard patterns of iteration. Was an interesting history lesson :)

PR #2146, opened 2020-11-27
Replace SpanData.hasRemoteParent/getParentSpanId with getParentSpanContext

### LINKED ISSUE #987: Remove SpanData.getHasRemoteParent() and the logic to determine that
This property was removed from the proto definition and is no longer a requirement in OpenTelemetery.

anuraaga:
Fixes #987 

~~From what I can tell, @Oberon00's concern is addressed by now since we have `Context` everywhere.~~ added getParentSpanContext to keep the existing functionality while also being able to remove more methods.

~~Since this is export-side, I just deleted it but let me know if it's worth doing `@Deprecated` first.~~


Oberon00 (changes requested):
I think this feature is useful.

anuraaga:
@Oberon00 Went ahead and added `getParentSpanContext`

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/data/SpanData.java
```diff
@@ -50,13 +50,36 @@
...
+
   /**
    * Returns the parent {@code SpanId}. If the {@code Span} is a root {@code Span}, the SpanId
    * returned will be invalid..
    *
    * @return the parent {@code SpanId} or an invalid SpanId if this is a root {@code Span}.
+   * @deprecated Use {@link #getParentSpanContext()}
    */
-  String getParentSpanId();
+  @Deprecated
```
Oberon00, line 66: I don't know if we even want to deprecate that. So far, it seems exporters have only needed this, so it seems OK to keep it as convenience shortcut.
I'm fine with deprecating/removing it though, just a thought.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -135,10 +131,8 @@ private RecordEventsReadableSpan(
    * @param context supplies the trace_id and span_id for the newly started span.
    * @param name the displayed name for the new span.
    * @param kind the span kind.
-   * @param parentSpanId the span_id of the parent span, or {@code Span.INVALID} if the new span is
-   *     a root span.
-   * @param hasRemoteParent {@code true} if the parentContext is remote. {@code false} if this is a
-   *     root span.
+   * @param parentSpanContext the span_id of the parent span, or {@code Span.INVALID} if the new
```
Oberon00, line 134:
```suggestion
   * @param parentSpanContext the parent span context, or {@link SpanContext#getInvalid()} if this
```

Oberon00 (approved):

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/data/SpanData.java
jkwatson: I'm fine either way.

jkwatson (approved):
Thanks!

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/data/SpanData.java
anuraaga: Makes sense to me to keep it so undeprecated

PR #2147, opened 2020-11-27
Remove Status.value

### LINKED ISSUE #1744: The spec doesn't seem to define numeric values for the status codes, do we need these?
https://www.github.com/open-telemetry/opentelemetry-specification/tree/master/specification%2Ftrace%2Fapi.md

anuraaga:
Fixes #1744 

Not quite merged but https://github.com/open-telemetry/opentelemetry-specification/pull/1257 has so many approvals I just went with its spec.


Oberon00:
This PR also changes attribute names in the Jaeger exporter. Intentionally?

anuraaga:
Yeah went ahead and reflected the almost merged https://github.com/open-telemetry/opentelemetry-specification/pull/1257 since it makes clearest what to do instead of using value.

Oberon00 (approved):
I think this makes sense. 👍

jkwatson (approved):
Thanks!

api/src/main/java/io/opentelemetry/api/trace/StatusCode.java
```diff
@@ -5,36 +5,18 @@
...
-  StatusCode(int value) {
-    this.value = value;
-  }
-
-  /**
-   * Returns the numerical value of the code.
-   *
-   * @return the numerical value of the code.
-   */
-  public int value() {
```
jkwatson, line 37: since this is a public API should we deprecate it before deleting?

anuraaga: Makes sense

PR #2148, opened 2020-11-27
Remove Guava dependencies from SDK

### LINKED ISSUE #1944: Consider dropping guava from core SDK dependencies

anuraaga:
For #1944 

Remaining are gRPC-based artifacts (always use guava), opencensus shim (it's going away in new implementation type), and jfr (will probably use weak-lock-free for it after https://github.com/raphw/weak-lock-free/pull/12)

### DIFF
--- a/sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
+++ b/sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
@@ -252,6 +252,7 @@ public final class OpenTelemetrySdk extends DefaultOpenTelemetry {
    * @see Obfuscated
    */
   @ThreadSafe
+  // Visible for testing
   static class ObfuscatedTracerProvider implements TracerProvider, Obfuscated<TracerProvider> {
 
     private final TracerProvider delegate;
--- a/sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
+++ b/sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
@@ -7,7 +7,6 @@ package io.opentelemetry.sdk;
 
 import static java.util.Objects.requireNonNull;
 
-import com.google.common.annotations.VisibleForTesting;
 import io.opentelemetry.api.DefaultOpenTelemetry;
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.api.internal.Obfuscated;
@@ -253,7 +252,6 @@ public final class OpenTelemetrySdk extends DefaultOpenTelemetry {
    * @see Obfuscated
    */
   @ThreadSafe
-  @VisibleForTesting
   static class ObfuscatedTracerProvider implements TracerProvider, Obfuscated<TracerProvider> {
 
     private final TracerProvider delegate;
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/common/export/ConfigBuilder.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/common/export/ConfigBuilder.java
@@ -5,8 +5,6 @@
 
 package io.opentelemetry.sdk.common.export;
 
-import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.Maps;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
@@ -26,7 +24,7 @@ import javax.annotation.Nullable;
  */
 public abstract class ConfigBuilder<T> {
 
-  @VisibleForTesting
+  // Visible for testing
   protected enum NamingConvention {
     DOT {
       @Override
@@ -65,7 +63,9 @@ public abstract class ConfigBuilder<T> {
 
   /** Sets the configuration values from the given {@link Properties} object. */
   public T readProperties(Properties properties) {
-    return fromConfigMap(Maps.fromProperties(properties), NamingConvention.DOT);
+    Map<String, String> map = new HashMap<>(properties.size());
+    properties.forEach((key, value) -> map.put((String) key, (String) value));
+    return fromConfigMap(map, NamingConvention.DOT);
   }
 
   /** Sets the configuration values from environment variables. */
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/common/export/ConfigBuilder.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/common/export/ConfigBuilder.java
@@ -63,8 +63,9 @@ public abstract class ConfigBuilder<T> {
 
   /** Sets the configuration values from the given {@link Properties} object. */
   public T readProperties(Properties properties) {
-    Map<String, String> map = new HashMap<>(properties.size());
-    properties.forEach((key, value) -> map.put((String) key, (String) value));
+    // Properties incorrectly implements Map<Object, Object> but we know it only has Strings.
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    Map<String, String> map = (Map) properties;
     return fromConfigMap(map, NamingConvention.DOT);
   }
 
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/resources/ResourcesConfig.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/resources/ResourcesConfig.java
@@ -5,15 +5,16 @@
 
 package io.opentelemetry.sdk.resources;
 
+import static java.util.Objects.requireNonNull;
+
 import com.google.auto.value.AutoValue;
-import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Preconditions;
-import com.google.common.base.Splitter;
-import com.google.common.collect.ImmutableSet;
 import io.opentelemetry.sdk.common.export.ConfigBuilder;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.stream.Collectors;
 import javax.annotation.concurrent.Immutable;
 
 /**
@@ -69,7 +70,8 @@ public abstract class ResourcesConfig {
    * @return a new {@link Builder}.
    */
   public static Builder builder() {
-    return new AutoValue_ResourcesConfig.Builder().setDisabledResourceProviders(ImmutableSet.of());
+    return new AutoValue_ResourcesConfig.Builder()
+        .setDisabledResourceProviders(Collections.emptySet());
   }
 
   /**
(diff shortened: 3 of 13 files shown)


Oberon00:
CC @thisthat

sdk-extensions/zpages/src/main/java/io/opentelemetry/sdk/extension/zpages/ZPageHttpHandler.java
```diff
@@ -42,24 +37,36 @@
...
+              List<String> keyValuePair =
+                  Arrays.stream(param.split("="))
+                      .filter(s -> !s.isEmpty())
+                      .map(String::trim)
+                      .collect(Collectors.toList());
+              if (keyValuePair.size() > 1) {
+                if (keyValuePair.get(0).equals(PARAM_SPAN_NAME)) {
+                  try {
+                    queryMap.put(
+                        keyValuePair.get(0), URLDecoder.decode(keyValuePair.get(1), "UTF-8"));
```
Oberon00, line 60: Why only decode selectively here? Shouldn't all keys and values be decoded?

sdk/common/src/main/java/io/opentelemetry/sdk/common/export/ConfigBuilder.java
```diff
@@ -65,7 +63,9 @@ protected abstract T fromConfigMap(
 
   /** Sets the configuration values from the given {@link Properties} object. */
   public T readProperties(Properties properties) {
-    return fromConfigMap(Maps.fromProperties(properties), NamingConvention.DOT);
+    Map<String, String> map = new HashMap<>(properties.size());
+    properties.forEach((key, value) -> map.put((String) key, (String) value));
```
Oberon00, line 67: I think a normal foreach loop would be more readable.

sdk/common/src/main/java/io/opentelemetry/sdk/resources/ResourcesConfig.java
```diff
@@ -103,8 +105,11 @@ protected Builder fromConfigMap(
       String stringValue = getStringProperty(OTEL_JAVA_DISABLED_RESOURCES_PROVIDERS, configMap);
       if (stringValue != null) {
         this.setDisabledResourceProviders(
-            ImmutableSet.copyOf(
-                Splitter.on(',').omitEmptyStrings().trimResults().split(stringValue)));
+            Collections.unmodifiableSet(
+                Arrays.stream(stringValue.split(","))
+                    .filter(s -> !s.isEmpty())
+                    .map(String::trim)
```
Oberon00, line 111: I think you might need to swap the order of trim/isEmpty to match Guava. AFAIK `omitEmptyStrings` only sets a boolean. But I may be wrong here.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
```diff
@@ -352,16 +350,26 @@ public Builder setTraceIdRatioBased(double samplerRatio) {
...
-          traceConfig.getMaxNumberOfAttributes() > 0, "maxNumberOfAttributes");
-      Preconditions.checkArgument(traceConfig.getMaxNumberOfEvents() > 0, "maxNumberOfEvents");
-      Preconditions.checkArgument(traceConfig.getMaxNumberOfLinks() > 0, "maxNumberOfLinks");
-      Preconditions.checkArgument(
-          traceConfig.getMaxNumberOfAttributesPerEvent() > 0, "maxNumberOfAttributesPerEvent");
-      Preconditions.checkArgument(
-          traceConfig.getMaxNumberOfAttributesPerLink() > 0, "maxNumberOfAttributesPerLink");
-      Preconditions.checkArgument(
-          traceConfig.getMaxLengthOfAttributeValues() >= -1, "maxLengthOfAttributeValues");
+      if (traceConfig.getMaxNumberOfAttributes() <= 0) {
```
Oberon00, line 353: Didn't we have some `Utils.checkArgument` function?

Oberon00 (commented):
Thank you for the work! Since this is a rather big PR, maybe it could be split up a bit. zPages may be worth splitting out (if we want to get rid of Guava there at all), and the core SDK (sdk-all + dependencies + maybe core propagators/exporters) could be split out since that part would be straightforward.

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource/EksResource.java
```diff
@@ -115,12 +114,31 @@ private String getClusterName() {
...
-      String content = Files.asCharSource(file, Charsets.UTF_8).read();
+      String content = new String(readAllBytes(K8S_TOKEN_PATH), StandardCharsets.UTF_8);
       return "Bearer " + content;
     } catch (IOException e) {
       logger.log(Level.WARNING, "Unable to load K8s client token.", e);
     }
     return "";
   }
+
+  private static byte[] readAllBytes(String path) throws IOException {
```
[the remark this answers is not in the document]
anuraaga: Not on Android - but I realized, this module should not have the Android sniffer on... Nice!

sdk-extensions/zpages/src/main/java/io/opentelemetry/sdk/extension/zpages/TracezZPageHandler.java
```diff
@@ -540,41 +546,55 @@ private static String latencyBoundaryToString(LatencyBoundary latencyBoundary) {
...
+    return latencyBoundaryMap;
   }
 
-  private static final class EventComparator implements Comparator<Event>, Serializable {
-    private static final long serialVersionUID = 0;
-
-    @Override
-    public int compare(Event e1, Event e2) {
-      return Long.compare(e1.getEpochNanos(), e2.getEpochNanos());
+  private static String escapeHtml(String html) {
```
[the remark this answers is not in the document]
anuraaga: Yeah this is the one line that gets most benefit from Guava. I'm mixed, 2MB dependency just for this function, mostly :)

anuraaga (commented):
Sure let me split zpages out

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource/Ec2Resource.java
```diff
@@ -212,9 +212,9 @@ private static void readTo(@Nullable InputStream is, ByteArrayOutputStream os)
     if (is == null) {
       return;
     }
-    int b;
-    while ((b = is.read()) != -1) {
-      os.write(b);
+    byte[] buf = new byte[8192];
+    while (is.read(buf) != -1) {
+      os.write(buf);
```
Oberon00, line 217: You can't write the full buffer here. You need to check how many bytes you actually read into the buffer, which may be less.

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource/DockerHelper.java
```diff
@@ -25,7 +24,7 @@
     this(DEFAULT_CGROUP_PATH);
   }
 
-  @VisibleForTesting
+  // Visible for testing
```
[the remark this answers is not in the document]
Oberon00: Hmm, no strong opinion. If you think the annotation fits badly, comments should be fine too.

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource/Ec2Resource.java
```diff
@@ -206,4 +206,15 @@ private String fetchIdentity(String token) {
...
   }
+
+  private static void readTo(@Nullable InputStream is, ByteArrayOutputStream os)
+      throws IOException {
+    if (is == null) {
+      return;
+    }
+    int b;
+    while ((b = is.read()) != -1) {
+      os.write(b);
```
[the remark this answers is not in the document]
Oberon00: Wouldn't the overhead be even worse then, relatively speaking?

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
```diff
@@ -253,7 +252,6 @@ private MeterProvider buildMeterProvider() {
    * @see Obfuscated
    */
   @ThreadSafe
-  @VisibleForTesting
```
Oberon00, line 256: Add a comment?

sdk/common/src/main/java/io/opentelemetry/sdk/common/export/ConfigBuilder.java
Oberon00: Alternatively, how about some casting? https://stackoverflow.com/a/17209434/2128694

```java
Map<String, String> map = (Map<String, String>)(Map)properties;
```

Oberon00 (approved):
Some cleanups should be considered, but all in all 👍

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
anuraaga: Oops

sdk/common/src/main/java/io/opentelemetry/sdk/common/export/ConfigBuilder.java
anuraaga: Thanks!

sdk/common/src/main/java/io/opentelemetry/sdk/resources/ResourcesConfig.java
jkwatson: Whether or not it matches guava, this logic seems right. Values that are only whitespace shouldn't be used, I don't think.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfig.java
jkwatson: Yeah, the SDK uses API classes quite a bit (obviously because it has to implement it!). I think this is a fine usage, as long as it's not in a strictly "internal" package.

sdk/common/src/main/java/io/opentelemetry/sdk/resources/ResourcesConfig.java
anuraaga: My not splitting the PRs up in advance caused this review to be a bit hectic, sorry about that. While GitHub doesn't have the outdated mark here, if I remember correctly, the code used to be in the wrong order and @Oberon00 commented and I fixed it so it should be good now.

jkwatson (approved):
Thanks!

PR #2149, opened 2020-11-27
Remove Guava from SDK extensions, except zpages

anuraaga:
For #1944

### DIFF
--- a/sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource/Ec2Resource.java
+++ b/sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource/Ec2Resource.java
@@ -214,7 +214,7 @@ public class Ec2Resource extends ResourceProvider {
     }
     byte[] buf = new byte[8192];
     int read;
-    while ((read = is.read(buf)) != -1) {
+    while ((read = is.read(buf)) > 0) {
       os.write(buf, 0, read);
     }
   }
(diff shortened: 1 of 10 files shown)


sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource/Ec2Resource.java
```diff
@@ -206,4 +206,16 @@ private String fetchIdentity(String token) {
...
   }
+
+  private static void readTo(@Nullable InputStream is, ByteArrayOutputStream os)
+      throws IOException {
+    if (is == null) {
+      return;
+    }
+    byte[] buf = new byte[8192];
+    int read;
+    while ((read = is.read(buf)) != -1) {
```
Oberon00, line 217:
```suggestion
    while ((read = is.read(buf)) > 0) {
```

Although -1 should also work according to <https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#read-byte:A->, I think this is slightly more clear.

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource/JdkHttpClient.java
```diff
@@ -134,4 +134,16 @@ private static KeyStore getKeystoreForTrustedCert(String certPath) {
       return null;
     }
   }
+
+  private static void readTo(@Nullable InputStream is, ByteArrayOutputStream os)
```
Oberon00, line 138: A bit of a shame that we have to duplicate this. But I guess that is the best way, we don't want to be in the business of providing some public utils artifact.

Oberon00 (approved):

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource/JdkHttpClient.java
anuraaga: I'll migrate `Ec2Resource` to use this class instead of duplicating logic soon :)

jkwatson (approved):
Thanks!

PR #2154, opened 2020-11-28
Enable tooling for mrjar and use it to get epoch time in best way for…

### LINKED ISSUE #1527: Use microsecond precision on Java 9+

anuraaga:
… Java 8 or 9+.

Fixes #1527

### DIFF
--- a/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/RateLimitingSampler.java
+++ b/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/RateLimitingSampler.java
@@ -15,7 +15,7 @@ import io.opentelemetry.api.common.ReadableAttributes;
 import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.api.trace.Span.Kind;
 import io.opentelemetry.context.Context;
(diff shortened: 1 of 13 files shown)


Oberon00:
I'm a bit worried about how this MultiJar feature will play with tooling such as the shadow gradle plugin that the auto-instrumentation uses. Will shadow be able to correctly handle classes added under META-INF, e.g. will it patch the classes there when asked to? E.g. the auto-instrumenting agent does this for java.util.logging.

jkwatson (approved):
Thanks!

anuraaga:
@Oberon00 Will require verification but I see support for it here https://github.com/johnrengelman/shadow/pull/526/files

jkwatson:
> @Oberon00 Will require verification but I see support for it here https://github.com/johnrengelman/shadow/pull/526/files

So, if we merge, can you verify before we do a release?

anuraaga:
@jkwatson Yup will check it

Issue #2011, opened 2020-11-04
Consider making TextMapPropagator.fields() a Collection/Iterable object.


carlosalberto:
In all other places in the API/SDK we expose, for similar (read-only) cases, either `Collection` or `Iterable`. I suspect we use `List` for legacy reasons in OC, but that's it.

Oberon00:
I think we now usually only export a forEach function. But I would definitely prefer even just Iterable to that.

carlosalberto:
I'd definitely stick to `Iterable` as this is what we use for the closely-related `Getter.keys()` method.

SCoceD:
it must look like this ?
`return C carrier.getOrDefault(keys, null);`

ChrisJBurns:
I'm looking for a good first issue to get stuck into. Is this one good to pickup?

anuraaga:
I think this is a good one to take on. Let's change the return type here

https://github.com/open-telemetry/opentelemetry-java/blob/master/context/src/main/java/io/opentelemetry/context/propagation/TextMapPropagator.java#L95

from `List` to `Collection`.

@carlosalberto `Iterable` doesn't seem to make sense even per the javadoc

https://github.com/open-telemetry/opentelemetry-java/blob/master/context/src/main/java/io/opentelemetry/context/propagation/TextMapPropagator.java#L93

Fields are a constant of a propagator, so them having a fixed-size `Collection` seems most useful. I guess we went with `Iterable` for `Getter.keys` to allow easier wrapping of the carrier, which isn't needed here.

ChrisJBurns:
Sure, I can make ` Collection` can someone possibly assign this to me? :)

ChrisJBurns:
@jkwatson I believe this can be closed now due to #2156 being merged.

anuraaga:
Yup - thanks @ChrisJBurns!

PR #2156, opened 2020-11-29
Changes TextMapPropagator.fields() return type from List to Collection

ChrisJBurns:
- changes the return type of the fields() method in TextMapPropagator from List<String> to Collection<String>

Addresses: #2011

### DIFF
--- a/api/src/main/java/io/opentelemetry/api/trace/propagation/HttpTraceContext.java
+++ b/api/src/main/java/io/opentelemetry/api/trace/propagation/HttpTraceContext.java
@@ -7,7 +7,7 @@ package io.opentelemetry.api.trace.propagation;
 
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.propagation.TextMapPropagator;
-import java.util.List;
+import java.util.Collection;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
@@ -33,7 +33,7 @@ public final class HttpTraceContext implements TextMapPropagator {
   }
 
   @Override
-  public List<String> fields() {
+  public Collection<String> fields() {
     return W3CTraceContextPropagator.getInstance().fields();
   }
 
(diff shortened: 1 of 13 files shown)


context/src/main/java/io/opentelemetry/context/propagation/TextMapPropagator.java
```diff
@@ -87,12 +88,12 @@ static TextMapPropagator noop() {
    * clear fields as they couldn't have been set before. If it is a mutable, retryable object,
    * successive calls should clear these fields first.
    *
-   * @return list of fields that will be used by this formatter.
+   * @return collection of fields that will be used by this formatter.
```
anuraaga, line 91:
```suggestion
   * @return the fields that will be used by this formatter.
```

anuraaga (approved):
Thanks! Just small nit to decouple javadoc from implementation detail

api/src/main/java/io/opentelemetry/api/trace/propagation/HttpTraceContext.java
```diff
@@ -33,7 +33,7 @@ public static HttpTraceContext getInstance() {
   }
 
   @Override
-  public List<String> fields() {
+  public Collection<String> fields() {
     return W3CTraceContextPropagator.getInstance().fields();
```
Oberon00, line 37: It seems the `W3CTraceContextPropagator`'s implementation should also be changed.

ChrisJBurns: Additionally, when doing a CMD + SHFT + F in IntelliJ for `fields()`.. It also highlights the following `fields()` methods in the propagators.

image

Do you think we should also change these implementations? As although they are `List` which is a subtype of `Collection`, it may be best to be explicit with the types for clarify and consistency.

jkwatson: Yes, let's change them all for consistency. good catch.

ChrisJBurns: Pushed changes for all implementations also

jkwatson (approved):
Thanks!

jkwatson:
@marcingrzejszczak this is a potentially breaking change in the API.

PR #2164, opened 2020-12-01
Detect GC leaks of scopes in StrictContextStorage.

anuraaga:
Also switches Guava dependency in JFR extension to use vendored weak-lock-free.

While debugging context issues in a production service using x-ray, it became apparent that not only in unit tests, but being able to detect leaks of Scope in a staging server can really help have peace of mind. As a followup to this PR I will move `StrictContextStorage` from `testing` to `context` - I think it's helpful for everyone in servers too, I'd recommend everyone to enable it on staging servers all the time.

weak-lock-free is a good, high-performant map with weak keys and much smaller than using Guava for it. We use it in opentelemetry-java-instrumentation already. We should avoid adding dependencies (especially on pre-1.0) so it makes sense to vendor it in. I considered another approach where I shade in `opentelemetry-context` only and use the shaded class in other artifacts, but it would be impossible for IntelliJ to not treat that import as red until `shadowJar` is run at least once. But for some reason, even after running `shadowJar`, I couldn't get it to ever be non-red. So instead, this shades in all the users as well. Another, possibly simplest is best approach, would be to just copy the files into our repo. Let me know if you have any thoughts.


build.gradle
```diff
@@ -509,6 +515,33 @@ configure(opentelemetryProjects) {
         // Always run jmhReport after jmh task.
         tasks.jmh.finalizedBy tasks.jmhReport
     }
+
+    // We vendor in a shaded version of weak-lock-free into opentelemetry-context, which is a
```
anuraaga, line 519: @trask Please review the shading I've set up in this PR.

Oberon00: Have you verified the .pom/.module and META-INF still look OK for this? IIRC shadow by default does not preserve the MANIFEST.mf of any patched jars.

jkwatson:
If we do this, I think we need to add some sort of attribution file, correct?

@sfriberg some tweaks to the JFR code here, if you're interested.

jkwatson:
looks like it needs a rebase

sdk-extensions/jfr-events/src/main/java/io/opentelemetry/sdk/extension/jfr/JfrSpanProcessor.java
```diff
@@ -45,6 +46,9 @@ public boolean isStartRequired() {
 
   @Override
   public void onEnd(ReadableSpan rs) {
+    if (closed) {
```
sfriberg, line 49: This will leave data in the map as we never touch it again. Since things are weak we won't hold to span structures but we will keep the weak refs around until JVM shutdown.

This was a one reason I went with the NoopMap to avoid the issue to figure out where to place the closed checked :)

probably best to just put it in the existing if statement as the first check.

sdk-extensions/jfr-events/build.gradle
```diff
@@ -8,7 +10,9 @@ ext.moduleName = 'io.opentelemetry.sdk.extension.jfr'
 dependencies {
     implementation project(':opentelemetry-api'),
             project(':opentelemetry-sdk')
-    implementation libraries.guava
+
+    compileOnly libraries.weak_lock_free
```
sfriberg, line 14: How do I ensure that we have this library around when loading JFR extension? Expect that opentelemetry is loaded and reachable and pulls in this dependency?

build.gradle
anuraaga: Thanks! Realized weak-lock-free was ending up in `opentelemetry-context` POM, changed so that doesn't happen.

sdk-extensions/jfr-events/build.gradle
anuraaga: Yeah this has the pattern where we vendor into `opentelemetry-context`, which is always pulled in as a dependency.

sdk-extensions/jfr-events/src/main/java/io/opentelemetry/sdk/extension/jfr/JfrSpanProcessor.java
anuraaga: Thanks!

anuraaga (commented):
Shadow plugin automatically puts the pom file of the included dependency into the JAR, e.g. `META-INF/maven/com.blogspot.mydailyjava/weak-lock-free/pom.xml`. The POM has credit, including the license specification and link to repo, so I guess this is why it does it.

This is what we're relying on in the agent right now also (I think) - it's not clear if this is enough, we have https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/836 open for the same issue. Maybe it's ok for now since we do it in the agent too.

sdk-extensions/jfr-events/build.gradle
anuraaga: By the way if there's a better pattern I'm happy to try it :) This is the only one I could come up with without duplicating the class into each artifact that uses it.

anuraaga:
I simplified this to just vendor the code into our repo directly since it's only two files. With this, the attributes issue can probably be decoupled cleanly as it's the same issue that we already have for e.g. `PersistentHashTrie`.

jkwatson:
looks like there are some broken tests...

sfriberg (approved):
JFR changes look good

build.gradle
```diff
@@ -328,6 +333,7 @@ configure(opentelemetryProjects) {
                 prometheus_client_common: "io.prometheus:simpleclient_common:${prometheusVersion}",
                 protobuf                : "com.google.protobuf:protobuf-java",
                 protobuf_util           : "com.google.protobuf:protobuf-java-util",
+                weak_lock_free          : "com.blogspot.mydailyjava:weak-lock-free:${weakLockFreeVersion}",
```
jkwatson, line 336: with the code vendor'd in, do we still need this?

anuraaga:
@jkwatson I'm a bit worried I've added a flaky test here (raised the timeout from 10s to 30s) but we can see how it goes I guess

build.gradle
anuraaga: Oops thanks

sdk-extensions/jfr-events/src/main/java/io/opentelemetry/sdk/extension/jfr/JfrSpanProcessor.java
```diff
@@ -26,11 +22,16 @@
  */
 public class JfrSpanProcessor implements SpanProcessor {
 
-  private volatile Map<SpanContext, SpanEvent> spanEvents =
-      new MapMaker().concurrencyLevel(16).initialCapacity(128).weakKeys().makeMap();
+  private final WeakConcurrentMap<SpanContext, SpanEvent> spanEvents =
+      new WeakConcurrentMap.WithInlinedExpunction<>();
```
jkwatson, line 26: I ask you... Is "expunction" a word? really?

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/context/StrictContextStorage.java
```diff
@@ -111,8 +116,9 @@ public Scope attach(Context context) {
 
     stackTrace = Arrays.copyOfRange(stackTrace, from, stackTrace.length);
     caller.setStackTrace(stackTrace);
+    pendingScopes.put(scope, caller);
```
jkwatson, line 119: why not have this done in the StrictScope constructor, so it controls the full lifecycle, rather having it put in here, but removed internal to the `StrictScope.close()`?

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/context/StrictContextStorage.java
```diff
@@ -132,35 +138,28 @@ public Context current() {
...
+    pendingScopes.expungeStaleEntries();
+    for (CallerStackTrace caller : pendingScopes.drainPendingCallers()) {
       // Sometimes unit test runners truncate the cause of the exception.
       // This flattens the exception as the caller of close() isn't important vs the one that leaked
-      AssertionError toThrow =
-          new AssertionError(
-              "Thread [" + caller.threadName + "] opened a scope of " + caller.context + " here:");
-      toThrow.setStackTrace(caller.getStackTrace());
-      throw toThrow;
+      throw callerError(caller);
```
jkwatson, line 145: so, we're only throwing the first that we run across, rather than alert about all the possibilities?

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/context/StrictContextStorage.java
```diff
@@ -132,35 +138,28 @@ public Context current() {
    */
   // AssertionError to ensure test runners render the stack trace
   public void ensureAllClosed() {
-    List<CallerStackTrace> leakedCallers = new ArrayList<>();
-    currentCallers.drainTo(leakedCallers);
-    for (CallerStackTrace caller : leakedCallers) {
+    pendingScopes.expungeStaleEntries();
+    for (CallerStackTrace caller : pendingScopes.drainPendingCallers()) {
       // Sometimes unit test runners truncate the cause of the exception.
       // This flattens the exception as the caller of close() isn't important vs the one that leaked
```
jkwatson, line 144: this comment is duplicated down in `callerError`

sdk-extensions/jfr-events/src/main/java/io/opentelemetry/sdk/extension/jfr/JfrSpanProcessor.java
anuraaga: Don't ask me ask raph :P

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/context/StrictContextStorage.java
anuraaga: Added logging when multiple, still throwing since this is mostly for unit tests where an exception is best

anuraaga: Went ahead and added logging if more than one.

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/context/StrictContextStorage.java
```diff
@@ -132,35 +137,34 @@ public Context current() {
...
-      toThrow.setStackTrace(caller.getStackTrace());
-      throw toThrow;
+    pendingScopes.expungeStaleEntries();
+    List<CallerStackTrace> leaked = pendingScopes.drainPendingCallers();
+    if (!leaked.isEmpty()) {
+      if (leaked.size() > 1) {
+        logger.log(Level.SEVERE, "Multiple scopes leaked - first will be thrown as an error.");
+      }
+      for (CallerStackTrace caller : leaked) {
+        logger.log(Level.SEVERE, "Scope leaked", callerError(caller));
```
jkwatson, line 147: we could skip doing this if there's only one, so you don't get the same one twice? Not sure if that's good or bad, but just a thought.

jkwatson (approved):
Thanks! Looking forward to the next PR when we see the recommended way of enabling this for staging/testing but not accidentally getting it for production.

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/context/StrictContextStorage.java
anuraaga: Oops - yeah this is a bug I meant to have it in the if statement, thanks

PR #2165, opened 2020-12-01
Have Node itself implement Context to remove wrapper object.

anuraaga:
We only use the Trie in our Context storage, so we can just inline Context semantics into it. This was @bogdandrutu's suggestion when we consider replacing this with a simple array.

I haven't rerun benchmarks for this PR since I think it's obvious there would be an improvement. I'll rerun later to compare this version vs the array, and if the array holds up fine, we may go ahead and remove the trie later.


jkwatson (approved):
Thanks!

PR #2168, opened 2020-12-01
Deprecate trace-utils

anuraaga:
If anyone wants to keep these classes, let me know. `MessageEvent` seems very strange to me, I guess it should be related to the semantic attributes somehow. I'm pushing to delete it. `CurrentSpanUtils` is not useful really unless using the `boolean endSpan` feature. I've marked it for deprecation, but if anyone wants to keep this, I would probably remove the `boolean endSpan` since I don't think it should ever be used if not for it's auto-end-span functionality.

### DIFF
--- a/extensions/trace-utils/src/main/java/io/opentelemetry/extension/trace/CurrentSpanUtils.java
+++ b/extensions/trace-utils/src/main/java/io/opentelemetry/extension/trace/CurrentSpanUtils.java
@@ -11,7 +11,13 @@ import io.opentelemetry.context.Context;
 import io.opentelemetry.context.Scope;
 import java.util.concurrent.Callable;
 
-/** Util methods/functionality to interact with the {@link Span} in the {@link Context}. */
+/**
+ * Util methods/functionality to interact with the {@link Span} in the {@link Context}.
+ *
+ * @deprecated Span should never be propagated by itself. Instead of this class, use {@code
+ *     Context.with(span).wrap(...)}. This class will be removed in SDK 0.13.0.
+ */
+@Deprecated
 public final class CurrentSpanUtils {
   // No instance of this class.
   private CurrentSpanUtils() {}
--- a/extensions/trace-utils/src/main/java/io/opentelemetry/extension/trace/MessageEvent.java
+++ b/extensions/trace-utils/src/main/java/io/opentelemetry/extension/trace/MessageEvent.java
@@ -21,8 +21,11 @@ import javax.annotation.concurrent.Immutable;
  *
  * <p>It requires a {@link Type type} and a message id that serves to uniquely identify each
  * message. It can optionally have information about the message size.
+ *
+ * @deprecated This class will be removed in SDK 0.13.0.
  */
 @Immutable
+@Deprecated
 public final class MessageEvent {
 
   private static final String EVENT_NAME = "message";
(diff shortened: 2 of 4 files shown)


anuraaga:
One notable thing about `CurrentSpanUtils` is it seems to be broken - `Context.current` needs to be run when instantiating the callbacks, not when running them. So it makes me thing this class cannot be currently used for anything meaningful :)

extensions/trace-utils/src/main/java/io/opentelemetry/extension/trace/CurrentSpanUtils.java
```diff
@@ -11,7 +11,13 @@
 import io.opentelemetry.context.Scope;
 import java.util.concurrent.Callable;
 
-/** Util methods/functionality to interact with the {@link Span} in the {@link Context}. */
+/**
+ * Util methods/functionality to interact with the {@link Span} in the {@link Context}.
+ *
+ * @deprecated Span should never be propagated by itself. Instead of this class, use {@code
```
Oberon00, line 17:
```suggestion
 * @deprecated A Span should never be propagated by itself. Instead of this class, use {@code
```

extensions/trace-utils/src/main/java/io/opentelemetry/extension/trace/MessageEvent.java
```diff
@@ -21,8 +21,11 @@
...
  * <p>It requires a {@link Type type} and a message id that serves to uniquely identify each
  * message. It can optionally have information about the message size.
+ *
+ * @deprecated This class will be removed in SDK 0.13.0.
  */
 @Immutable
+@Deprecated
 public final class MessageEvent {
 
   private static final String EVENT_NAME = "message";
```
Oberon00, line 31: Note that this stuff is still in the gRPC semantic conventions (it was copied from OpenCensus AFAIK and never really discussed/updated).

Oberon00 (approved):

extensions/trace-utils/src/main/java/io/opentelemetry/extension/trace/MessageEvent.java
```diff
@@ -21,8 +21,11 @@
  *
  * <p>It requires a {@link Type type} and a message id that serves to uniquely identify each
  * message. It can optionally have information about the message size.
+ *
+ * @deprecated This class will be removed in SDK 0.13.0.
  */
 @Immutable
+@Deprecated
 public final class MessageEvent {
```
Oberon00, line 29: Add a note about the SemanticAttributes class maybe?

extensions/trace-utils/src/main/java/io/opentelemetry/extension/trace/MessageEvent.java
anuraaga: Yeah the gRPC instrumentation doesn't use it which shows how much this class is helping ;)

Oberon00:
If we are deprecating the whole artifact, we should add something to the README.

carlosalberto:
`MessageEvent` was used by OC, so @bogdandrutu may know about it.

jkwatson (approved):
Seems fine to me to get rid of it. I used it really early on to get a handle on context management in some toy projects, but I definitely don't need it now.

extensions/trace-utils/src/main/java/io/opentelemetry/extension/trace/MessageEvent.java
anuraaga: Added one

jkwatson:
Merging this. If anyone needs this, we can un-deprecate in 0.13.0 and no harm no foul.

PR #2170, opened 2020-12-01
Adding micro-benchmarks for traceId and spanId generation for AwsXrayIdGenerator

srprash:


jkwatson (approved):
It would be good to include the benchmark data in the PR description, so we can look back at how it looked at this point in time.

srprash:
> It would be good to include the benchmark data in the PR description, so we can look back at how it looked at this point in time.

Great suggestion. Will do so shortly. :)

anuraaga (approved):
Thanks! Feel free to edit the PR description when you're ready, we can still merge it.

PR #2171, opened 2020-12-01
change the default timeouts for otlp and jaeger to 10s

### LINKED ISSUE #2064: Change the default grpc deadline for OTLP and Jaeger exporters from 1s to 10s
I think this maps to `deadlineMs`, and the default should be bumped from `1s` to `10s` according to [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/protocol/exporter.md#configuration-options)

It doesn't look like there's a corresponding env var for Jaeger, but should we bump default Jaeger grpc deadline from `1s` to `10s` also?

(we've had a couple of `DEADLINE_EXCEEDED` issues reported in otel-java-instrumentation which is why I'm interested 😄)

jkwatson:
resolves #2064


carlosalberto (approved):

trask:
Thanks!

PR #2174, opened 2020-12-02
Replace AttributeConsumer with BiConsumer

anuraaga:
Our current interface has a wildcard type parameter - it seems to provide no real benefit in almost all cases, but prevents using lambdas with `Attributes.forEach`. For the cases it helped a bit, `putAll` works fine.


anuraaga:
This is breaking, but I don't think I can do it with deprecation - if we had two `forEach` methods, the caller would need to pick with an explicit cast.

api/src/main/java/io/opentelemetry/api/common/ArrayBackedAttributesBuilder.java
```diff
@@ -90,8 +90,11 @@ public AttributesBuilder put(String key, boolean... value) {
   }
 
   @Override
-  public AttributesBuilder putAll(Attributes attributes) {
-    attributes.forEach(this::put);
+  @SuppressWarnings({"unchecked", "rawtypes"})
+  public AttributesBuilder putAll(ReadableAttributes attributes) {
+    // Attributes must iterate over their entries with matching types for key / value, so this
+    // downcast to the raw type is safe.
+    attributes.forEach((key, value) -> put((AttributeKey) key, value));
```
anuraaga, line 97: Populating another attributes is basically the only time `<T>` helped - but we can manage that ourselves without sacrificing the ability to lambda

jkwatson:
Both options end up being kind-of ugly. At least this one has one fewer public thing we have to maintain. The lack of union types is definitely hurting us here, for sure.

jkwatson:
@marcingrzejszczak another small breaking change here.

jkwatson (approved):
👍

PR #2175, opened 2020-12-02
Use key as AttributeKey.toString

anuraaga:
I find unit test assertion messages and IntelliJ debugging to be quite hard to grok with this looking like `{type = STRING, value = foo}`. While using the key name does lose the type, I don't think people will find that a loss since presumably the value is right next to it.


api/src/main/java/io/opentelemetry/api/common/AttributeKeyImpl.java
```diff
@@ -24,6 +24,11 @@ public String getKey() {
   @Nullable
   abstract String key();
 
+  @Override
+  public final String toString() {
+    return getKey();
```
jkwatson, line 29: It might be nice to at least delineate that this *isn't* just a String. Maybe put a colon in front of it, so it looks like an elixir atom or a ruby symbol?

jkwatson: or after it?

Oberon00: I think the type should stay in the toString. Maybe `name(type)`? But I don't hate the current/old string representation.

api/src/test/java/io/opentelemetry/api/common/AttributesTest.java
```diff
@@ -370,4 +367,20 @@ void nullsAreNoOps() {
...
+        Attributes.builder()
+            .put("otel.status_code", "OK")
+            .put("http.response_size", 100)
+            .put("process.cpu_consumed", 33.44)
+            .put("error", true)
+            .build();
+
+    assertThat(attributes.toString())
+        .isEqualTo(
+            "{error=true, http.response_size=100, "
```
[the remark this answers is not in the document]
jkwatson: Of course, the counter-argument (and I'd say a pretty strong one) is that with your change, you now can't distinguish between `{error=true}` and `{error=true}` where one of them is an actual boolean value (correct) and one is a string value (incorrect). And, even worse, with spock, your test would pass when it shouldn't!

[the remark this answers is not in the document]
jkwatson: any thoughts here, @anuraaga ?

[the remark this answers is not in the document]
anuraaga: Sorry for taking forever to get back to this. Added quoting of string values, wonder if it's enough to have the clarity along with the brevity.

api/common/src/test/java/io/opentelemetry/api/common/AttributesTest.java
```diff
@@ -118,10 +118,7 @@ void asMap() {
     assertThat(map.values().contains("value1")).isTrue();
     assertThat(map.values().contains("value3")).isFalse();
 
-    assertThat(map.toString())
-        .isEqualTo(
-            "ReadOnlyArrayMap{AttributeKeyImpl{getType=STRING, key=key1}=value1,"
-                + "AttributeKeyImpl{getType=LONG, key=key2}=333}");
+    assertThat(map.toString()).isEqualTo("ReadOnlyArrayMap{key1=value1,key2=333}");
```
jkwatson, line 121: shouldn't this one have quotes around the `value1` ?

jkwatson:
I don't love this change, but I'm willing to give it a try and see how it feels. Since it's 100% debug, non-breaking, we can revisit when we decide it makes things more confusing. ;)

jkwatson (approved):
Thanks?

api/common/src/test/java/io/opentelemetry/api/common/AttributesTest.java
anuraaga: This is `Map`, not `Attributes`

anuraaga:
> Thanks?

Will keep an eye out for grumbles :)

PR #2177, opened 2020-12-02
Expose BaggageEntry / Metadata as interfaces and add asMap

anuraaga:
Baggage has been an interface for a while so skipped @jkwatson's recent refactorings. But metadata wasn't yet so this completes the conversion by exposing it as an interface. It also exposes `BaggageEntry` to be used with `asMap`. When making `Entry` private, I always figured it would come back for use in `asMap` and now it has.


api/src/main/java/io/opentelemetry/api/baggage/BaggageBuilder.java
```diff
@@ -51,7 +51,7 @@
    * @param entryMetadata the {@code EntryMetadata} associated with this {@link Entry}.
    * @return this
    */
-  BaggageBuilder put(String key, String value, EntryMetadata entryMetadata);
+  BaggageBuilder put(String key, String value, BaggageEntryMetadata entryMetadata);
```
anuraaga, line 54: API compatible but ABI incompatible. I could add an overload for the latter too, but guess it shouldn't matter at this stage.

jkwatson (approved):
Thanks!

jkwatson:
@marcingrzejszczak possible breaking change here.

PR #2187, opened 2020-12-03
Remove ReadableAttributes

anuraaga:
Now that `Attributes` is an interface, we don't need the almost identical `ReadableAttributes`.

As it's used in return values, there isn't a good way to do deprecation I think.


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/ParentBasedSampler.java
```diff
@@ -52,9 +53,9 @@ public SamplingResult shouldSample(
       Context parentContext,
       String traceId,
       String name,
-      Span.Kind spanKind,
-      ReadableAttributes attributes,
-      List<SpanData.Link> parentLinks) {
+      Kind spanKind,
+      Attributes attributes,
+      List<Link> parentLinks) {
```
jkwatson, line 58: interesting...are we allowing inner-class imports now? [I think it's fine...I just thought this would fail spotless]

jkwatson (approved):
Good catch!

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -519,7 +518,7 @@ int getTotalRecordedLinks() {
   }
 
   @GuardedBy("lock")
-  private ReadableAttributes getImmutableAttributes() {
+  private Attributes getImmutableAttributes() {
```
jkwatson, line 521: now that the ordering requirement on attributes has been relaxed, we could go back to returning our array-backed ones here, if we wanted to. Not a change that needs to be in this PR.

jkwatson:
@marcingrzejszczak this is potentially a breaking change.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
anuraaga: Yup noticed this too :)

PR #2189, opened 2020-12-03
Add Labels.asMap and hide the builder implementation.

anuraaga:
Looks like builder was accidentally left in the interface.


jkwatson:
Same question as with TraceState: do we need this now? Can we add it later, if there is demand?

(note: good catch on the builder being accidentally public)

anuraaga:
@jkwatson I would go ahead and get this in to be consistent with all our key/value pairs.

jkwatson (approved):
thanks!

PR #2190, opened 2020-12-03
Add TraceState.asMap

anuraaga:
Looks like builder was accidentally left in the interface.

### DIFF
--- a/api/src/main/java/io/opentelemetry/api/trace/TraceState.java
+++ b/api/src/main/java/io/opentelemetry/api/trace/TraceState.java
@@ -5,6 +5,7 @@
 
 package io.opentelemetry.api.trace;
 
+import java.util.Map;
 import java.util.function.BiConsumer;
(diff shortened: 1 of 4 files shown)


api/src/main/java/io/opentelemetry/api/trace/TraceState.java
```diff
@@ -69,6 +70,9 @@ static TraceStateBuilder builder() {
   /** Iterates over all the key-value entries contained in this {@link TraceState}. */
   void forEach(BiConsumer<String, String> consumer);
 
+  /** Returns a read-only view of this {@link TraceState} as a {@link Map}. */
+  Map<String, String> asMap();
```
jkwatson, line 74: So... hard question. Do we need this now, or could we add this to the API later, when needed? We should definitely limit the public API as much as possible, and I'd like to understand the use-case for this.

jkwatson (approved):
approved for consistency.

PR #2195, opened 2020-12-04
Add a system property that can enable strict context checking for sta…

anuraaga:
…ging environments.


jkwatson (approved):
This is great. Thanks!!!

PR #2199, opened 2020-12-04
Split API packages, move context in the same directory

bogdandrutu:
There are some things left to be done:
* Decide if context package should also be opentelemetry-api-context.
* Cleanup internal packages, things that are used by only one artifact should be moved there.


jkwatson:
I think this is a good change, to make users more able to mix & match pieces of the APIs that they wish to use. 

I wonder if having `opentelemetry-api-internal` be a separate module that is published to maven-central is a good idea, though. Could that code live in common, still in the internal package, to make it (slightly) more obvious that this is not code that should be used outside the project?

bogdandrutu:
That is why I have the second point in the description. Need to take another look at that package, for the moment I did a simple split

jkwatson (approved):
I'd like @anuraaga to take a look before we merge, but I think this is generally the right direction.

anuraaga (approved):
Thanks!

PR #2202, opened 2020-12-07
Remove SDK obfuscation since javaagent instruments SDK instead now.

anuraaga:
If I'm not mistaken, after https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1830 we don't need this stuff here. @trask can you confirm?


jkwatson:
I thought one of the goals of this was to (very strongly) discourage people from using the global SDK to write instrumentation.  Is that not the case?

Oberon00:
Yes, AFAIR the motivation was to prevent code like `((TracerProviderSdk)OpenTelemetry.getGlobalTracerProvider()).addSpanProcessor(...)`. Which is still a valid reason.

jkwatson:
> Yes, AFAIR the motivation was to prevent code like `((TracerProviderSdk)OpenTelemetry.getGlobalTracerProvider()).addSpanProcessor(...)`. Which is still a valid reason.

which is the opposite of what I said, but *agreed*...using the API to configure the SDK also seems like something we don't want people to be doing.

anuraaga:
Ah you mean without the agent? Yeah now I remember, makes sense.

PR #2205, opened 2020-12-07
Remove deprecated methods/classes before 1.0 release

bogdandrutu:
Only one deprecated left in metrics, since that package is not going to 1.0 yet.


jkwatson:
we should get rid of the deprecated method in the metrics API as well, since we said we'd get rid of it for 0.13.0

breedx-splk (approved):
This looks good to me.  Thanks for the deletions!

Oberon00:
Is the trace 1.0 release close?

bogdandrutu:
@jkwatson will do it in a separate PR, a bit more to take care there.

Oberon00 (approved):
I trust all of these were already deprecated in the latest released version?

Issue #2067, opened 2020-11-12
Remove 3rd party propagators from this repository


jkwatson:
See recently merged spec PR: https://github.com/open-telemetry/opentelemetry-specification/pull/1144

anuraaga:
For me, I've always felt like the `extension` folders in this repo are our version of a contrib repo - they used to even be called `contrib` before being renamed to `extension`. I guess we could choose from

1) Documenting that extensions are contrib and consider adding CODEOWNERS for the folders (this seems like a good idea already for the logging extension for example)
2) Move all the extensions to `opentelemetry-java-contrib`
3) Move only propagators to `opentelemetry-java-contrib`

Lean a bit towards 1) since it seems simpler

jkwatson:
If we think 1) is still in the spirit of the spec, then I'm totally ok with it. @carlosalberto what do you think?

carlosalberto:
The change in the Specification precisely mentions 'core repos', so I think this is a very liberal interpretation of it 😉 But ultimately the decision is yours @jkwatson about whether you want to support such code. Other SIG maintainers were not specially happy about maintaining such code, but if you are fine maintaining it (same repo, in the end), I'm definitely cool.

(If we decide to separate the OT one, I'm fine with moving it somewhere else and maintain it, btw).

jkwatson:
If the owning vendors (@carlosalberto and @anuraaga ) are ok with "best effort" maintenance of the propagators, then I'm ok with it. I think we'd want to add some very clear verbiage to the READMEs for the propagators that makes it clear that critical bugs should be logged with the vendors themselves, with at least some kind of link to how to do that.

Otherwise, I wouldn't guess there's a lot of actual work to be done maintaining this code. And, if the owning vendors aren't satisfied with the service, they are absolutely free to move the code out to where they can provide the support they need to.

carlosalberto:
@jkwatson Makes sense. Also we can apply what other maintainers suggested, about removing the related code in case the vendor doesn't help maintaining it (worst case). Sounds good?

jkwatson:
> @jkwatson Makes sense. Also we can apply what other maintainers suggested, about removing the related code in case the vendor doesn't help maintaining it (worst case). Sounds good?

That sounds good to me, yes.

breedx-splk:
I think this can be closed now, right?

PR #2207, opened 2020-12-07
Update README for extensions propagators

breedx-splk:
This is intended to address #2067.  

I took a best guess at documenting where to go for help, so maybe this is just a start...and hopefully @anuraaga and @carlosalberto are both ok with the direct mentions (if not, please say something).  Are there other maintainers to mention as well?

We can talk about this more in #2067 but @anuraaga mentioned adding a [CODEOWNERS](https://docs.github.com/en/free-pro-team@latest/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-file-location) file, but those github docs didn't really suggest a good (supported) subdir location.

### DIFF
--- a/extensions/trace-propagators/README.md
+++ b/extensions/trace-propagators/README.md
@@ -5,3 +5,18 @@ OpenTelemetry Contrib Trace Propagators
 
 [javadoc-image]: https://www.javadoc.io/badge/io.opentelemetry/opentelemetry-contrib-trace-propagators.svg
 [javadoc-url]: https://www.javadoc.io/doc/io.opentelemetry/opentelemetry-contrib-trace-propagators
+
+This project contains [trace propagators](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md) that support
+3rd-party propagation formats.  These propagators are here for convenience and are not officially
+supported by the OpenTelemetry Java maintainers.
+
+Issues/support for these propagators is only provided as a minimal "best effort", and critical
+bugs should be filed with the respective vendors themselves.
+
+* AWS X-Ray (file an issue here and mention @anuraaga)
+* OpenTracing (file an issue here and mention @carlosalberto)
+* B3 (file an issue on [the OpenZipkin repo](https://github.com/openzipkin/b3-propagation) that points here)
+* Jaeger (file an issue on [the Jaeger repo](https://github.com/jaegertracing/jaeger) that points here)
+
+Extension providers that do not receive adequate support/maintenance by their respective vendors 
+will become candidates for future removal.
--- a/extensions/trace-propagators/README.md
+++ b/extensions/trace-propagators/README.md
@@ -14,7 +14,7 @@ Issues/support for these propagators is only provided as a minimal "best effort"
 bugs should be filed with the respective vendors themselves.
 
 * AWS X-Ray (file an issue here and mention @anuraaga)
-* ListStep OpenTracing (file an issue here and mention @carlosalberto)
+* LightStep OpenTracing (file an issue here and mention @carlosalberto)
 * B3 (file an issue on [the OpenZipkin repo](https://github.com/openzipkin/b3-propagation) that points here)
 * Jaeger (file an issue on [the Jaeger repo](https://github.com/jaegertracing/jaeger) that points here)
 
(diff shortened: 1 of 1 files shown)


bogdandrutu:
@breedx-splk  can you please sign or fix the CLA?

breedx-splk:
> @breedx-splk can you please sign or fix the CLA?

Yes, working on that.

extensions/trace-propagators/README.md
```diff
@@ -5,3 +5,18 @@ OpenTelemetry Contrib Trace Propagators
...
+
+This project contains [trace propagators](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md) that support
+3rd-party propagation formats.  These propagators are here for convenience and are not officially
+supported by the OpenTelemetry Java maintainers.
+
+Issues/support for these propagators is only provided as a minimal "best effort", and critical
+bugs should be filed with the respective vendors themselves.
+
+* AWS X-Ray (file an issue here and mention @anuraaga)
+* OpenTracing (file an issue here and mention @carlosalberto)
```
jkwatson, line 17: Maybe "Lightstep OpenTracing" here, rather than just "OpenTracing"?

extensions/trace-propagators/README.md
```diff
@@ -5,3 +5,18 @@ OpenTelemetry Contrib Trace Propagators
...
+* B3 (file an issue on [the OpenZipkin repo](https://github.com/openzipkin/b3-propagation) that points here)
+* Jaeger (file an issue on [the Jaeger repo](https://github.com/jaegertracing/jaeger) that points here)
```
jkwatson, line 19: I thought we were on the hook to support these 2. Has that changed?

breedx-splk: I think that's up to us to decide...but the discussion in #2067 led me to believe that this entire subproject was external/contrib/3rd party.  If these are first-class supported exporters, should we move them into the main codebase?

jkwatson: I would very much prefer if we weren't having to support these propagators, but it's unclear to me whether these are considered 1st-class or 3rd-party or somewhere in between.

breedx-splk: I put a note in the SIG agenda for Wed. about the `TraceMultiPropagator`, so maybe we can figure this out then too?

anuraaga: It's pretty explicit that they are official

https://github.com/open-telemetry/opentelemetry-specification/blob/84d2b533eb7f1120b476b2ba7f301e467c890304/specification/context/api-propagators.md#propagators-distribution

Natural too since without them, no one can migrate to OpenTelemetry :P I think B3 and Jaeger were here in this package without much consideration to how supports them, more just something like "why not keep all the propagators together". But we're definitely on the hook for them.

I'd be ok with moving those into `opentelemetry-api-trace` if the wording here would make the OTel support aspect convoluted.

extensions/trace-propagators/README.md
```diff
@@ -5,3 +5,18 @@ OpenTelemetry Contrib Trace Propagators
...
+
+This project contains [trace propagators](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md) that support
+3rd-party propagation formats.  These propagators are here for convenience and are not officially
+supported by the OpenTelemetry Java maintainers.
+
+Issues/support for these propagators is only provided as a minimal "best effort", and critical
+bugs should be filed with the respective vendors themselves.
+
+* AWS X-Ray (file an issue here and mention @anuraaga)
+* ListStep OpenTracing (file an issue here and mention @carlosalberto)
```
anuraaga, line 17:
```suggestion
* LightStep OpenTracing (file an issue here and mention @carlosalberto)
```

extensions/trace-propagators/README.md
breedx-splk: I'll propose changing the wording here and will file a separate issue for moving B3 and Jaeger to `opentelemetry-api-trace` to be peers of `W3CTraceContextPropagator`.

breedx-splk:
Ok, I filed #2228 as a follow-up, and the readme changes here are ready for another read.  Thanks all!

extensions/trace-propagators/README.md
jkwatson: FYI, still not sure where this is going to land:  https://github.com/open-telemetry/oteps/issues/142 and whether it applies to propagators or not, and what set of propagators.

jkwatson (approved):
Thanks!

anuraaga (approved):

jkwatson:
let's go with it until such time as we need to move stuff out (if ever)!

PR #2208, opened 2020-12-07
Move obfuscated to sdk, make it package protected

bogdandrutu:

### DIFF
--- a/sdk/all/src/main/java/io/opentelemetry/sdk/Obfuscated.java
+++ b/sdk/all/src/main/java/io/opentelemetry/sdk/Obfuscated.java
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package io.opentelemetry.api.internal;
+package io.opentelemetry.sdk;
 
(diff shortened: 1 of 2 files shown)


jkwatson:
there's an open PR to remove this altogether. #2202 . If we don't remove it, then hiding it is good.

bogdandrutu:
I would do the hiding first since it unblocks my work of reducing the internal package (or even removing if possible).

bogdandrutu:
@jkwatson this should be uncontroversial, the removing still has open questions.

jkwatson:
@trask @anuraaga does this have any impact on the agent (I suspect not, but thought I'd give a heads-up in case this is a terrible change)?

jkwatson (approved):
👍

bogdandrutu:
@jkwatson do you suggest to wait for this then? Maybe at least @trask can confirm :)

jkwatson:
> @jkwatson do you suggest to wait for this then? Maybe at least @trask can confirm :)

we can revert if need-be. Just wanted to raise awareness.

Issue #2203, opened 2020-12-07
Decide on cross-version compatibility requirement or not


anuraaga:
We can either require all artifacts published from this repo to be aligned to the same version in a user's app or not. If the latter, we cannot share internal code among artifacts (for example, `TemporaryBuffers.java` is useful in API and in trace-propagators artifacts) or otherwise we would need to make sure we don't make breaking changes to them to prevent problems in the face of misaligned versions, eliminating any advantage of making them internal in the first place.

I think it's fairly conventional to expect all artifacts from a repo to be aligned, and we provide the BOM to make it easy. So it's probably reasonable for us to have a policy where all artifacts are aligned. Let me know any thoughts or concerns.

anuraaga:
By the way, if yes on expecting alignment, I'll move the opentelemetry-api-internal all into common

Oberon00:
Note that we need to fulfill https://github.com/open-telemetry/oteps/pull/143. In particular I am wondering if we can expect SDK and API versions to always align. If a library was compiled with API version 1.x but the app uses SDK + API version 1.y (x < y) that should normally not be a problem since at runtime, only the app's version would be used? And as long as 1.y is ABI compatible to 1.x, that should just work.

jkwatson:
Yes, API and SDK versions should definitely *not* be required to be aligned. 
Yes, we should move everything in `opentelemetry-api-internal` into common.
I think it is reasonable to require all the SDK modules to be aligned, since that's intended to be something configured at the application level.

I think it will be impossible to require that versions of everything being used in the API be aligned, however. Imagine a world where OTel instrumentation is baked into libraries. Some of those libraries will instrument with spans, and some with metrics, and we can't require that they're using the same versions across transitive dependencies, since we don't control those dependencies.

@marcingrzejszczak decisions on this could impact you.

marcingrzejszczak:
> I think it will be impossible to require that versions of everything being used in the API be aligned, however. Imagine a world where OTel instrumentation is baked into libraries. Some of those libraries will instrument with spans, and some with metrics, and we can't require that they're using the same versions across transitive dependencies, since we don't control those dependencies.

When you say aligned what do you mean exactly? Sleuth uses currently opentelemetry-instrumentation-java and opentelemetry-java. We must align on the opentelemetry-java version that means that we're using the same version in all of our modules. That version must be compatible with whatever opentelemety-instrumentation-java provides - otherwise our code breaks.

[...]

PR #2212, opened 2020-12-07
Add rationale and description of the release and versioning strategy.

jkwatson:
To fulfill the ask from https://github.com/open-telemetry/opentelemetry-specification/issues/1267

### DIFF
--- a/docs/rationale.md
+++ b/docs/rationale.md
@@ -24,3 +24,68 @@ try (Scope scope = TracingContextUtils.currentContextWith(span)) {
 
 It would not be possible to call `recordException` if `span` was also using try-with-resources.
 Because this is a common usage for spans, we do not support try-with-resources.
+
+
+## Versioning and Releases
+
+### Assumptions
+
+- This project uses semver v2, as does the rest of OpenTelemetry.
+
+### Goals
+
+- API Stability: 
+    - Once the API for a given signal (spans, logs, metrics, baggage) has been officially released, that API module will 
+function, *with no recompilation required*, with any SDK that has the same major version, and equal or greater minor or patch version. 
+    - For example, libraries that are instrumented with `opentelemetry-api-trace:1.0.1` will function with 
+SDK library `opentelemetry-sdk-trace:1.11.33`. 
+    - We call this requirement the "ABI" compatibility requirement for "Application Binary Interface" compatibility. 
+- SDK Stability:
+    - Public portions of the SDK (constructors, configuration, und-user interfaces) must remain backwards compatible. 
+    - Internal interfaces are allowed to break. 
+    - Strict ABI compatibility is not required.
+
+### Methods
+
+- Mature signals
+    - API modules for mature (i.e. released) signals will be transitive dependencies of the `opentelemetry-api` module.
+    - Methods for accessing mature APIs will be added, as appropriate to the `OpenTelemetry` interface.
+    - SDK modules for mature (i.e. released) signals will be transitive dependencies of the `opentelemetry-sdk` module.
+    - Configuration options for the SDK modules for mature signals will be exposed, as appropriate, on the `OpenTelemetrySdk` class.
+
+- Immature or experimental signals
+    - API modules for immature signals will not be transitive dependencies of the `opentelemetry-api` module.
+    - API modules will be named with an "-experimental" suffix to make it abundantly clear that depending on them is at your own risk.
+    - API modules for immature signals will be co-versioned along with mature API modules.
+    - The java packages for immature APIs will be used as if they were mature signals. This will enable users to easily transition from immature to 
+    mature usage, without having to change imports. 
+    - SDK modules for immature signals will also be named with an "-experiment" suffix, in parallel to their API modules.
+    
+### Examples
+
+Purely for illustration purposes, not intended to represent actual releases:
+
+- `v1.0.0` release:
+    - `io.opentelemetry:opentelemetry-api:1.0.0`
+        - Contains APIs for tracing, baggage, propagators (via the context dependency)
+    - `io.opentelemetry:opentelemetry-api-metrics-experimental:1.0.0`
+        - Note: packages here are the final package structure: `io.opentelemetry.api.metrics.*`
+    - `io.opentelemetry:opentelemetry-sdk-trace:1.0.0`
+    - `io.opentelemetry:opentelemetry-sdk-common:1.0.0`
+        - Shared code for metrics/trace implementations (clocks, etc)
+    - `io.opentelemetry:opentelemetry-sdk-metrics-experimental:1.0.0`
+        - Note: packages here are the final package structure: `io.opentelemetry.sdk.metrics.*`
+    - `io.opentelemetry:opentelemetry-sdk-all:1.0.0`
+        - The SDK side of `io.opentelemetry:opentelemetry-api:1.0.0`
+        - No mention of metrics in here!
+- `v1.15.0` release (with metrics)
+    - `io.opentelemetry:opentelemetry-api:1.15.0`
+        - Contains APIs for tracing, baggage, propagators (via the context dependency), metrics
+    - `io.opentelemetry:opentelemetry-sdk-trace:1.15.0`
+    - `io.opentelemetry:opentelemetry-sdk-common:1.15.0`
+        - Shared code for metrics/trace implementations (clocks, etc)
+    - `io.opentelemetry:opentelemetry-sdk-metrics:1.15.0` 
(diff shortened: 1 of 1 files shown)


breedx-splk (approved):
Seems ok to me, if not overspecified.  😁 . It could be nice to have an assumption about everything else being intended to improve compatibility and foster collaboration and explicitly _not_ to stifle/slow development.

No other real feedback/concerns.

docs/rationale.md
```diff
@@ -23,4 +23,69 @@ try (Scope scope = TracingContextUtils.currentContextWith(span)) {
...
+### Assumptions
+
+- This project uses semver v2, as does the rest of OpenTelemetry.
+
+### Goals
+
+- API Stability: 
+    - Once the API for a given signal (spans, logs, metrics, baggage) has been officially released, that API module will 
+function, *with no recompilation required*, with any SDK that has the same major version, and equal or greater minor or patch version. 
+    - For example, libraries that are instrumented with `opentelemetry-api-trace:1.0.1` will function with
```
anuraaga, line 40: Looking through https://github.com/open-telemetry/oteps/pull/143/files I couldn't find a point on this so I guess it's just #2203 :)

Given using a bom is easy and we can strongly encourage it, I don't know if we need to require this

docs/rationale.md
```diff
@@ -23,4 +23,69 @@ try (Scope scope = TracingContextUtils.currentContextWith(span)) {
...
+### Goals
+
+- API Stability: 
+    - Once the API for a given signal (spans, logs, metrics, baggage) has been officially released, that API module will 
+function, *with no recompilation required*, with any SDK that has the same major version, and equal or greater minor or patch version. 
+    - For example, libraries that are instrumented with `opentelemetry-api-trace:1.0.1` will function with 
+SDK library `opentelemetry-sdk-trace:1.11.33`. 
+    - We call this requirement the "ABI" compatibility requirement for "Application Binary Interface" compatibility. 
+- SDK Stability:
+    - Public portions of the SDK (constructors, configuration, und-user interfaces) must remain backwards compatible.
```
anuraaga, line 44: I think these first two points aren't unique to SDK

docs/rationale.md
```diff
@@ -23,4 +23,69 @@ try (Scope scope = TracingContextUtils.currentContextWith(span)) {
...
+- API Stability: 
+    - Once the API for a given signal (spans, logs, metrics, baggage) has been officially released, that API module will 
+function, *with no recompilation required*, with any SDK that has the same major version, and equal or greater minor or patch version. 
+    - For example, libraries that are instrumented with `opentelemetry-api-trace:1.0.1` will function with 
+SDK library `opentelemetry-sdk-trace:1.11.33`. 
+    - We call this requirement the "ABI" compatibility requirement for "Application Binary Interface" compatibility. 
+- SDK Stability:
+    - Public portions of the SDK (constructors, configuration, und-user interfaces) must remain backwards compatible. 
+    - Internal interfaces are allowed to break. 
+    - Strict ABI compatibility is not required.
```
anuraaga, line 46: This can't be allowed right? I think many SDK classes are still too important (`OpenTelemetrySdk.Builder` for example, span processors, sampler) to risk dependency hell by breaking ABI. Imagine a custom sampler someone published, we can't break it.

jkwatson: I think this is worth having a discussion about. I'm pretty hesitant to say we're going to maintain ABI compatibility across the entire swath of the SDK at this point, given how youthful it is. A custom sampler should be something that it is easy for an SDK user to update, correct? It's not going to be backed into a library or instrumentation, is it?

docs/rationale.md
jkwatson: What are you saying we don't need to require?  Shouldn't an 1.x API work with any 1.y SDK as long as y >= x?

docs/rationale.md
jkwatson: I was thinking this is a lesser requirement than the API requirements.

docs/rationale.md
anuraaga: It requires the sampler to also update to the newer SDK first, something we shouldn't force post-1.0 for SDK either. For whatever reason, it might be slow to update, or may be abandoned but still useful - the moment we create a situation where a user can't upgrade OTel because they have a dependency linked to the SDK, which IMO is quite valid), then we lose most of the value of semver in the first place. If we didn't, I'd say we'd need to list even trace as `sdk-trace-experimental`, we need the experimental flag to reflect whether we are going to introduce dependency hell on users or not.

Another example is Spring, which will be using the SDK classes. If our SDK is 1.0, we should be in a state where we don't cause version conflicts between Spring and OTel (Spring 4.2.3 requires OTel SDK 1.3.4, yikes).

jkwatson: So... what SDK APIs are going to be locked in stone forever? How do we distinguish one public SDK API from another? I feel like if we're going to require strict ABI compatibility with some subset of the SDK, we need a way to publish that module separately, as something like `opentelemetry-sdk-api` or something. Any ideas on how to achieve this goal?

anuraaga: All of it I think - the reason we have any public methods at all is because we expect code outside our control to use them. As soon as that happens, we need to commit to those public methods, or we'll cause conflicts - any library that causes conflicts becomes unpopular fast. I don't see anything special about the SDK vs the API in this regard.

We should be as diligent as we can by 1.0 to get things into a good place - we probably won't get everything and we'll have methods deprecated until the next major, etc. It is what it is, but it's safer than introducing ABI breaks that could cause dependency locks for users.

docs/rationale.md
jkwatson: But, if there is a library that is instrumented with API v. 1.0.0, that would mean you can't use SDK v. 1.1.0 with it? Doesn't that defeat the entire purpose of ABI compatibility, let alone API compatibility? This seems to directly contradict what you're saying about SDK ABI requirements! 

Or else, we're talking at cross purposes about something!

iNikem: @anuraaga this statement, as written here, MUST be true. If my app has a dependency on a library, which is instrumented (meaning was compiled against) API v1.0.1, then it MUST work if offered SDK v1.11.33 during runtime.

I think what you want to fix in writing is "should we allow having different versions of API and SDK during _runtime_". E.g. is it OK for application developer to specify in their build.gradle SDK v1.11.33 and API v1.0.1? I agree with Anuraag here that we should discourage that.

But! What if an application developer has SDK v1.11.33 and some library depends on API v1.20. What will SDK do?

docs/rationale.md
iNikem: First, I agree with Anuraag, that SDK should have stable public interface. With the same API/ABI compatibility guarantees as API.

Second, as I read it, OTEP 143 allows as to stabilise SDK separate from API. Thus we don't have to claim SDK stability yet. We can just release stable API and continue working on SDK.

Third, if we are going to provide stability guarantees for SDK, we have to distinguish public and internal parts of it. And guarantee only public parts. How to achieve that: separate packages, separate artifacts, JPMS, OSGI - is a separate question :)

docs/rationale.md
jkwatson: I probably need to update this to be more clear. I think rather than "public", we need to provide a way to specify what pieces of the SDK are under this backwards-compatibility requirement and what isn't. Unless we put the whole SDK into one package, I think it will be hard to make sure that all public classes/interfaces are fully supported.

docs/rationale.md
jkwatson: What will it do will obviously depend on the nature of the changes. If they happen to be backward compatible, then things should work. And, if they aren't, then probably all hell will break loose and NoClassDefFoundErrors will abound at runtime, unfortunately. 

Can we maybe detect versions at startup and fail fast if there are likely to be compatibility issues?

docs/rationale.md
jkwatson: yes, after consideration and discussion, I also agree with this. I don't know how we're going to achieve it, but I also think it's important. 

Ideas on how to effectively version the various parts of the project independently are very welcome!

trask: Ya, being able to release breaking SDK 2.0 without having to bump API to 2.0 is nice

I think we have similar issue in instrumentation repo, e.g. being able to release grpc library instrumentation 2.0 without having to bump `opentelemetry-instrumentation-api`.

This will require a bunch more release infra to manage versioning/releasing modules independently (and using prefixed tags, e.g. `api-1.1`, `sdk-2.0`).

jkwatson:
closing in favor of #2250

PR #2213, opened 2020-12-07
Remove version of checkArgument not used

bogdandrutu:
This PR fixes an inconsistency that MaxLengthOfAttributeValues was allowed to be set to 0, then thrown internal exception.
We should check for arguments only at public APIs.


jkwatson (approved):
Thanks! Deleting code is good!

PR #2186, opened 2020-12-02
Refactoring span benchmarking as per the spec

srprash:
Making the following changes to span creation according to the [span configuration](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/performance-benchmark.md#span-configuration) for performance benchmarking:
- Adding a service resource
- Adding the always on sampler
- Adding a single long type attribute
- Adding an event with no attributes

The benchmark result on a m5.xlarge EC2 instance:

```shell
Benchmark                                                                         Mode  Cnt     Score     Error   Units
SpanBenchmark.simpleSpanStartAddEventEnd_01Thread                                thrpt   10  3274.915 ±  16.454  ops/ms
SpanBenchmark.simpleSpanStartAddEventEnd_01Thread:·gc.alloc.rate                 thrpt   10   882.267 ±   3.930  MB/sec
SpanBenchmark.simpleSpanStartAddEventEnd_01Thread:·gc.alloc.rate.norm            thrpt   10   424.000 ±   0.001    B/op
SpanBenchmark.simpleSpanStartAddEventEnd_01Thread:·gc.churn.G1_Eden_Space        thrpt   10   893.436 ± 102.937  MB/sec
SpanBenchmark.simpleSpanStartAddEventEnd_01Thread:·gc.churn.G1_Eden_Space.norm   thrpt   10   429.436 ±  51.007    B/op
SpanBenchmark.simpleSpanStartAddEventEnd_01Thread:·gc.churn.G1_Old_Gen           thrpt   10     0.001 ±   0.002  MB/sec
SpanBenchmark.simpleSpanStartAddEventEnd_01Thread:·gc.churn.G1_Old_Gen.norm      thrpt   10     0.001 ±   0.001    B/op
SpanBenchmark.simpleSpanStartAddEventEnd_01Thread:·gc.count                      thrpt   10    63.000            counts
SpanBenchmark.simpleSpanStartAddEventEnd_01Thread:·gc.time                       thrpt   10    34.000                ms
SpanBenchmark.simpleSpanStartAddEventEnd_02Threads                               thrpt   10  4865.498 ±  73.035  ops/ms
SpanBenchmark.simpleSpanStartAddEventEnd_02Threads:·gc.alloc.rate                thrpt   10  1309.461 ±  20.750  MB/sec
SpanBenchmark.simpleSpanStartAddEventEnd_02Threads:·gc.alloc.rate.norm           thrpt   10   424.001 ±   0.002    B/op
[...]

PR #2214, opened 2020-12-07
Span benchmark with OTLP exporter and collector

srprash:
Refactoring the span pipeline benchmark to align it to the throughput measurement in the spec. 
The OTLP collector is spun up using testContainers and OTLP exporter is being used to send spans to the collector.

running the benchmark with `./gradlew -PjmhIncludeSingleClass=SpanPipelineBenchmark :opentelemetry-sdk:jmh` on an m5.xlarge EC2 instance gives the following result:

```shell
Benchmark                                                                                                 Mode  Cnt       Score       Error   Units
SpanPipelineBenchmark.BatchSpanProcessorBenchmark.measureSpanPipeline                                    thrpt   15  527317.589 ± 42314.216   ops/s
SpanPipelineBenchmark.BatchSpanProcessorBenchmark.measureSpanPipeline:·gc.alloc.rate                     thrpt   15     983.419 ±    81.052  MB/sec
SpanPipelineBenchmark.BatchSpanProcessorBenchmark.measureSpanPipeline:·gc.alloc.rate.norm                thrpt   15    2938.474 ±    93.300    B/op
SpanPipelineBenchmark.BatchSpanProcessorBenchmark.measureSpanPipeline:·gc.churn.G1_Eden_Space            thrpt   15     994.126 ±   176.230  MB/sec
SpanPipelineBenchmark.BatchSpanProcessorBenchmark.measureSpanPipeline:·gc.churn.G1_Eden_Space.norm       thrpt   15    2969.188 ±   428.534    B/op
SpanPipelineBenchmark.BatchSpanProcessorBenchmark.measureSpanPipeline:·gc.churn.G1_Old_Gen               thrpt   15       0.201 ±     0.180  MB/sec
SpanPipelineBenchmark.BatchSpanProcessorBenchmark.measureSpanPipeline:·gc.churn.G1_Old_Gen.norm          thrpt   15       0.595 ±     0.527    B/op
SpanPipelineBenchmark.BatchSpanProcessorBenchmark.measureSpanPipeline:·gc.churn.G1_Survivor_Space        thrpt   15       0.355 ±     0.455  MB/sec
SpanPipelineBenchmark.BatchSpanProcessorBenchmark.measureSpanPipeline:·gc.churn.G1_Survivor_Space.norm   thrpt   15       1.033 ±     1.314    B/op
SpanPipelineBenchmark.BatchSpanProcessorBenchmark.measureSpanPipeline:·gc.count                          thrpt   15      32.000              counts
SpanPipelineBenchmark.BatchSpanProcessorBenchmark.measureSpanPipeline:·gc.time                           thrpt   15     382.000                  ms
SpanPipelineBenchmark.SimpleSpanProcessorBenchmark.measureSpanPipeline                                   thrpt   15   22402.867 ±  6041.441   ops/s
SpanPipelineBenchmark.SimpleSpanProcessorBenchmark.measureSpanPipeline:·gc.alloc.rate                    thrpt   15     237.753 ±    59.024  MB/sec
SpanPipelineBenchmark.SimpleSpanProcessorBenchmark.measureSpanPipeline:·gc.alloc.rate.norm               thrpt   15   16559.668 ±   461.969    B/op
SpanPipelineBenchmark.SimpleSpanProcessorBenchmark.measureSpanPipeline:·gc.churn.G1_Eden_Space           thrpt   15     235.371 ±    79.240  MB/sec
SpanPipelineBenchmark.SimpleSpanProcessorBenchmark.measureSpanPipeline:·gc.churn.G1_Eden_Space.norm      thrpt   15   16494.578 ±  4326.637    B/op
SpanPipelineBenchmark.SimpleSpanProcessorBenchmark.measureSpanPipeline:·gc.churn.G1_Old_Gen              thrpt   15       2.665 ±    11.034  MB/sec
SpanPipelineBenchmark.SimpleSpanProcessorBenchmark.measureSpanPipeline:·gc.churn.G1_Old_Gen.norm         thrpt   15     222.285 ±   920.363    B/op
SpanPipelineBenchmark.SimpleSpanProcessorBenchmark.measureSpanPipeline:·gc.churn.G1_Survivor_Space       thrpt   15       2.863 ±     3.719  MB/sec
SpanPipelineBenchmark.SimpleSpanProcessorBenchmark.measureSpanPipeline:·gc.churn.G1_Survivor_Space.norm  thrpt   15     237.116 ±   332.514    B/op
SpanPipelineBenchmark.SimpleSpanProcessorBenchmark.measureSpanPipeline:·gc.count                         thrpt   15      33.000              counts
SpanPipelineBenchmark.SimpleSpanProcessorBenchmark.measureSpanPipeline:·gc.time                          thrpt   15    4541.000                  ms
```

Ref: [This](https://github.com/open-telemetry/opentelemetry-java/pull/2186) PR for span benchmarks without the exporter.
@jkwatson I hope this covers the part which I missed in the last PR.

### DIFF
--- a/sdk/all/src/jmh/java/io/opentelemetry/sdk/trace/SpanPipelineBenchmark.java
+++ b/sdk/all/src/jmh/java/io/opentelemetry/sdk/trace/SpanPipelineBenchmark.java
@@ -68,7 +68,8 @@ public class SpanPipelineBenchmark {
 
       String address = collector.getHost() + ":" + collector.getMappedPort(EXPOSED_PORT);
 
-      TraceConfig alwaysOn = TraceConfig.getDefault().toBuilder().setSampler(Sampler.alwaysOn()).build();
+      TraceConfig alwaysOn =
+          TraceConfig.getDefault().toBuilder().setSampler(Sampler.alwaysOn()).build();
 
       TracerSdkProvider tracerProvider =
           TracerSdkProvider.builder().setTraceConfig(alwaysOn).build();
--- a/sdk/all/src/jmh/java/io/opentelemetry/sdk/trace/SpanPipelineBenchmark.java
+++ b/sdk/all/src/jmh/java/io/opentelemetry/sdk/trace/SpanPipelineBenchmark.java
@@ -40,7 +40,9 @@ public class SpanPipelineBenchmark {
     private static final int EXPOSED_PORT = 5678;
     private static final int HEALTH_CHECK_PORT = 13133;
     private SpanBuilderSdk spanBuilderSdk;
+
     protected abstract SpanProcessor getSpanProcessor(String collectorAddress);
+
     protected abstract void runThePipeline();
 
     protected void doWork() {
@@ -66,13 +68,11 @@ public class SpanPipelineBenchmark {
 
       String address = collector.getHost() + ":" + collector.getMappedPort(EXPOSED_PORT);
 
-      TracerSdkProvider tracerProvider = TracerSdkProvider.builder().build();
+      TraceConfig alwaysOn = TraceConfig.getDefault().toBuilder().setSampler(Sampler.alwaysOn()).build();
 
+      TracerSdkProvider tracerProvider =
+          TracerSdkProvider.builder().setTraceConfig(alwaysOn).build();
       tracerProvider.addSpanProcessor(getSpanProcessor(address));
-
-      TraceConfig alwaysOn =
-          tracerProvider.getActiveTraceConfig().toBuilder().setSampler(Sampler.alwaysOn()).build();
-      tracerProvider.updateActiveTraceConfig(alwaysOn);
 
       Tracer tracerSdk = tracerProvider.get("PipelineBenchmarkTracer");
       spanBuilderSdk = (SpanBuilderSdk) tracerSdk.spanBuilder("PipelineBenchmarkSpan");
--- a/sdk/all/src/jmh/java/io/opentelemetry/sdk/trace/SpanPipelineBenchmark.java
+++ b/sdk/all/src/jmh/java/io/opentelemetry/sdk/trace/SpanPipelineBenchmark.java
@@ -5,24 +5,12 @@
 
 package io.opentelemetry.sdk.trace;
 
-import static io.opentelemetry.api.common.AttributeKey.booleanKey;
-import static io.opentelemetry.api.common.AttributeKey.doubleKey;
-import static io.opentelemetry.api.common.AttributeKey.longKey;
-import static io.opentelemetry.api.common.AttributeKey.stringKey;
-
-import io.opentelemetry.api.OpenTelemetry;
-import io.opentelemetry.api.common.AttributeKey;
-import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.trace.Span;
-import io.opentelemetry.api.trace.Span.Kind;
-import io.opentelemetry.api.trace.StatusCode;
 import io.opentelemetry.api.trace.Tracer;
-import io.opentelemetry.sdk.OpenTelemetrySdk;
-import io.opentelemetry.sdk.common.CompletableResultCode;
-import io.opentelemetry.sdk.trace.data.SpanData;
+import io.opentelemetry.exporter.otlp.OtlpGrpcSpanExporter;
+import io.opentelemetry.sdk.trace.config.TraceConfig;
 import io.opentelemetry.sdk.trace.export.SimpleSpanProcessor;
-import io.opentelemetry.sdk.trace.export.SpanExporter;
-import java.util.Collection;
+import io.opentelemetry.sdk.trace.samplers.Sampler;
 import java.util.concurrent.TimeUnit;
 import org.openjdk.jmh.annotations.Benchmark;
 import org.openjdk.jmh.annotations.Fork;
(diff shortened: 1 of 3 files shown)


sdk/all/src/jmh/resources/otel.yaml
```diff
@@ -0,0 +1,27 @@
...
+    endpoint: backend:8080
+    insecure: true
+
+service:
+  extensions: [health_check]
+  pipelines:
+    traces:
+      receivers: [otlp]
+      processors: [batch, queued_retry]
+      exporters: [logging, otlp]
```
jkwatson, line 27: we don't need otlp in here, do we? What's "backend:8080" pointing at, anyway?

jkwatson:
When this benchmark is run, do we know how many spans end up getting dropped on the floor because the exporter can't keep up?

sdk/all/src/jmh/resources/otel.yaml
srprash: You're right. We don't need the otlp exporter here. It was a copy paste error, sorry. I'll remove it.

srprash:
> When this benchmark is run, do we know how many spans end up getting dropped on the floor because the exporter can't keep up?

I didn't see a ["Failed to export spans"](https://github.com/open-telemetry/opentelemetry-java/blob/ed2cbf75f8a59fececbcfb1a116d76b50bf9cae8/exporters/otlp/src/main/java/io/opentelemetry/exporter/otlp/OtlpGrpcSpanExporter.java#L148-L149) message in the stdout when I ran the benchmark. Does that mean none of the spans were dropped by the exporter? I was seeing this message when I didn't configure the collector properly during tests.

jkwatson:
> > When this benchmark is run, do we know how many spans end up getting dropped on the floor because the exporter can't keep up?
> 
> I didn't see a ["Failed to export spans"](https://github.com/open-telemetry/opentelemetry-java/blob/ed2cbf75f8a59fececbcfb1a116d76b50bf9cae8/exporters/otlp/src/main/java/io/opentelemetry/exporter/otlp/OtlpGrpcSpanExporter.java#L148-L149) message in the stdout when I ran the benchmark. Does that mean none of the spans were dropped by the exporter? I was seeing this message when I didn't configure the collector properly during tests.

No, they will get silently dropped by the BatchSpanProcessor. We do generate metrics about this when it happens: https://github.com/open-telemetry/opentelemetry-java/blob/master/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java#L196 but other than that, we silently drop in order to not impact the running application.

jkwatson:
@srprash You can see here how I tracked this in the stress test that I built:  https://github.com/open-telemetry/opentelemetry-java/blob/master/perf-harness/src/test/java/io/opentelemetry/perf/OtlpPipelineStressTest.java#L180-L183

srprash:
> > > When this benchmark is run, do we know how many spans end up getting dropped on the floor because the exporter can't keep up?
> > 
> > 
> > I didn't see a ["Failed to export spans"](https://github.com/open-telemetry/opentelemetry-java/blob/ed2cbf75f8a59fececbcfb1a116d76b50bf9cae8/exporters/otlp/src/main/java/io/opentelemetry/exporter/otlp/OtlpGrpcSpanExporter.java#L148-L149) message in the stdout when I ran the benchmark. Does that mean none of the spans were dropped by the exporter? I was seeing this message when I didn't configure the collector properly during tests.
> 
> No, they will get silently dropped by the BatchSpanProcessor. We do generate metrics about this when it happens: https://github.com/open-telemetry/opentelemetry-java/blob/master/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java#L196 but other than that, we silently drop in order to not impact the running application.

I see. But I'm using a SimpleSpanProcessor here and it doesn't seem to have a counter for dropped spans. Should I be using a BatchSpanProcessor instead?

jkwatson:
> I see. But I'm using a SimpleSpanProcessor here and it doesn't seem to have a counter for dropped spans. Should I be using a BatchSpanProcessor instead?

Certainly the OTLP exporter is probably going to work more optimally if it doesn't have to send the spans one by one. I would never recommend anyone use OTLP with the SSP, although I guess it's useful to have a benchmark of how much throughput it can do! This is one of the issues I had with the original benchmark specification. There are configuration options you can tweak here if you're using the BatchSpanProcessor. Also, there's an `opentelemetry-sdk-async-process` that uses a Disruptor to move the spans around. It's going to be the highest performance (at least hopefully). 

Honestly, I'm still unclear on what the goals of this particular benchmark are. You can measure something (and you are!), but is it something that's going to be useful to anyone? I'm not so sure, due to all of these issues.

srprash:
> > I see. But I'm using a SimpleSpanProcessor here and it doesn't seem to have a counter for dropped spans. Should I be using a BatchSpanProcessor instead?
> 
> Certainly the OTLP exporter is probably going to work more optimally if it doesn't have to send the spans one by one. I would never recommend anyone use OTLP with the SSP, although I guess it's useful to have a benchmark of how much throughput it can do! This is one of the issues I had with the original benchmark specification. There are configuration options you can tweak here if you're using the BatchSpanProcessor. Also, there's an `opentelemetry-sdk-async-process` that uses a Disruptor to move the spans around. It's going to be the highest performance (at least hopefully).
> 
> Honestly, I'm still unclear on what the goals of this particular benchmark are. You can measure something (and you are!), but is it something that's going to be useful to anyone? I'm not so sure, due to all of these issues.

My goal of having this benchmark is to give a baseline idea of the SDK performance in creating and exporting spans. As you mentioned that the BSP is better than SSP (and more common to be used) and then the `opentelemetry-sdk-asyn-process` is even better, there are clearly many knobs and dials to tuning performance of the SDK. It is difficult to performance benchmark when there are multiple parameters, but I believe that its best to have a benchmark with a simple and common use case so that anyone using the SDK would have some idea of its performance and can extrapolate the benchmark according to his/her usage.
Benchmarking this span pipeline is a bit subjective, but I think having a simple span with few attributes and a BSP with OTLP exporter is a common use case that we can try to benchmark in the SDK.

anuraaga:
@jkwatson The main goal is to have something that reflects the spec, but it actually doesn't clearly specify the span processor

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/performance-benchmark.md#cpu-usage-measurement

Not too sure what "with both default configured simple and batching span processors" is but @srprash Maybe we can have two versions, one with each?

sdk/all/src/jmh/java/io/opentelemetry/sdk/trace/SpanPipelineBenchmark.java
```diff
@@ -5,96 +5,123 @@
...
+      // Configuring the collector test-container
+      GenericContainer<?> collector =
+          new GenericContainer<>(OTLP_COLLECTOR_IMAGE)
+              .withExposedPorts(EXPOSED_PORT, HEALTH_CHECK_PORT)
+              .waitingFor(Wait.forHttp("/").forPort(HEALTH_CHECK_PORT))
+              .withCopyFileToContainer(
+                  MountableFile.forClasspathResource("/otel.yaml"), "/etc/otel.yaml")
+              .withCommand("--config /etc/otel.yaml");
+
+      collector.start();
```
anuraaga, line 65: While ryuk cleans things up usually, let's go ahead and add `collector.stop()` in a teardown

sdk/all/src/jmh/java/io/opentelemetry/sdk/trace/SpanPipelineBenchmark.java
```diff
@@ -5,96 +5,123 @@
...
+
+      String address = collector.getHost() + ":" + collector.getMappedPort(EXPOSED_PORT);
+
+      TracerSdkProvider tracerProvider = TracerSdkProvider.builder().build();
+
+      tracerProvider.addSpanProcessor(getSpanProcessor(address));
```
anuraaga, line 71: In #2223 I'm trying to add this to the builder to be able to configure the tracer provider in one go.

In the meantime, builder already has `setTraceConfig` so how about using that instead of the update?

sdk/all/src/jmh/java/io/opentelemetry/sdk/trace/SpanPipelineBenchmark.java
srprash: An issue with doing `collector.stop()` in the teardown method is that the collector is shutdown before the span processor is able to export all the spans from the last benchmark iteration. For BSP, the spans can be flushed out before stopping the collector container but SSP doesn't offer any such functionality.
Letting ryuk take care of the collector testContainer for us is probably the best way right now.

jkwatson:
looks like this might need spotless run on it

jkwatson (approved):
Thanks!

PR #2215, opened 2020-12-07
Remove internal public validateLabelPairs, move to the class that uses it

bogdandrutu:


jkwatson (approved):
Good cleanup!

breedx-splk (approved):
This seems like a good change, I like that the usage is closer to the point implementation.

PR #2216, opened 2020-12-07
Remove deprecated setCallback, change SDK to avoid volatile

bogdandrutu:

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractAsynchronousInstrument.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractAsynchronousInstrument.java
@@ -11,43 +11,37 @@ import io.opentelemetry.sdk.metrics.aggregator.Aggregator;
 import io.opentelemetry.sdk.metrics.data.MetricData;
 import java.util.Collections;
 import java.util.List;
-import java.util.Objects;
 import java.util.concurrent.locks.ReentrantLock;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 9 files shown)


anuraaga (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractAsynchronousInstrument.java
```diff
@@ -11,45 +11,39 @@
...
 import java.util.Collections;
 import java.util.List;
-import java.util.Objects;
 import java.util.concurrent.locks.ReentrantLock;
 import javax.annotation.Nullable;
 
 abstract class AbstractAsynchronousInstrument<T extends AsynchronousInstrument.Result>
     extends AbstractInstrument implements AsynchronousInstrument<T> {
-  @Nullable private volatile Callback<T> metricUpdater = null;
+  @Nullable private final Callback<T> metricUpdater;
```
jkwatson, line 19: yes, this was always in my head as a thing that needed to be done. Thanks!

jkwatson (approved):
Thanks!

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractAsynchronousInstrument.java
bogdandrutu: That's why I separated from the previous PR, because I wanted to fix this deeper

PR #2219, opened 2020-12-08
Move small internal package under common

bogdandrutu:


api/common/src/main/java/io/opentelemetry/api/common/ArrayBackedAttributes.java
```diff
@@ -6,7 +6,7 @@
 package io.opentelemetry.api.common;
 
 import com.google.auto.value.AutoValue;
-import io.opentelemetry.api.internal.ImmutableKeyValuePairs;
+import io.opentelemetry.api.common.internal.ImmutableKeyValuePairs;
```
anuraaga, line 9: I would make the internal package `io.opentelemetry.api.internal` so we only have one internal prefix to deal with (e.g., exclude from javadocs). FWIU, having an internal package across multiple artifacts is fine JPMS-wise since they're not public.

bogdandrutu: Not sure how to handle the java9 module package name then...

anuraaga: I don't think it should matter since it's not public - the java 9 name is still `io.opentelemetry.api.common`. Technically moving it out may even hide the packages from Java 9 module users (though I've never tried using it 😆 )

bogdandrutu: Done.

jkwatson (approved):
👍

anuraaga (approved):

PR #2221, opened 2020-12-08
Remove Obfuscated marker interface.

anuraaga:
While we do need to keep obfuscation, we don't need this one which used to be used by the javaagent but not anymore.


trask (approved):
Nice

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
```diff
@@ -271,14 +271,15 @@ private MeterProvider buildMeterProvider() {
   }
 
   /**
-   * A {@link TracerProvider} wrapper that forces users to access the SDK specific implementation
-   * via the SDK, instead of via the API and casting it to the SDK specific implementation.
+   * This interface allows the SDK to unobfuscate an obfuscated static global provider.
```
bogdandrutu, line 274: This is not an interface.

jkwatson (approved):
oh good. tiny javadoc update needed, looks like, but 👍

PR #2222, opened 2020-12-08
Move ResourcesConfig builder private

anuraaga:
`ResourcesConfig` is specifically to configure SPI, a non-programmatic tool. I guess having a programmatic way of specifying it has no use case.


jkwatson (approved):
Thanks!

bogdandrutu:
> Though another option is to just make this private - ResourcesConfig is specifically to configure SPI, a non-programmatic tool. I guess having a programmatic way of specifying it has no use case.

Please apply the rule of minimizing the API if possible. No clear use-case to have this public => make it non-public :)

bogdandrutu (approved):

jkwatson:
Thinking further on this...is this something that @iNikem was going to use in his vendor-configuration case? @anuraaga do you not use this for the AWS distribution?

PR #2223, opened 2020-12-08
Allow adding span processors when building a tracer provider.

anuraaga:
Similar to how we added to `OpenTelemetrySdk.Builder` I think we should have it here for symmetry and for users that only interact with the tracer-sdk.


sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
```diff
@@ -216,10 +216,6 @@ public OpenTelemetrySdk build() {
       MeterProvider meterProvider = buildMeterProvider();
       TracerSdkProvider tracerProvider = buildTracerProvider();
 
-      for (SpanProcessor spanProcessor : spanProcessors) {
-        tracerProvider.addSpanProcessor(spanProcessor);
-      }
```
jkwatson, line 221: the downside about removing this is that now we won't be adding span processors to a pre-configured SDK. Not sure if we're ok with that or not. I know we've chatted about removing the ability to provide the fully configured SDK altogether, which would make this (obviously) just fine.

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
```diff
@@ -252,6 +248,7 @@ private TracerSdkProvider buildTracerProvider() {
       if (traceConfig != null) {
         tracerProviderBuilder.setTraceConfig(traceConfig);
       }
+      spanProcessors.forEach(tracerProviderBuilder::addSpanProcessor);
```
jkwatson, line 251: should we also do this up on line 236 to preserve backward compatibility? I was very surprised that there wasn't a unit test that verified this.

bogdandrutu (changes requested):
I think the reason to add this is wrong, we should remove the capability from the OpenTelemetry.Builder and add only this one.

jkwatson:
> I think the reason to add this is wrong, we should remove the capability from the OpenTelemetry.Builder and add only this one.

Can you explain why you say this? @iNikem has very much wanted this capability in the configuration of OpenTelemetrySdk itself. Providing the option in either place will let people either configure the whole public API, or just the tracing part of it, as they desire.

bogdandrutu:
Here are some reasons:
1. Duplicate ways to achieve the same thing as well as duplicate code. I don't think it needs explanation why this is wrong, but a small example is anything changes in one of these APIs and you have two deprecated APIs instead of one, confusion for the users, etc.
2. It is possible to configure the TracerProvider, so no need to have the global way.

We need to remember that larger our API surface is, more things we need to maintain and be careful about. So limiting the API is always the best decision. The trade-off is 1 extra line on the configuration code so not worth it.

bogdandrutu:
In https://github.com/open-telemetry/opentelemetry-java/pull/2231, I removed all duplicate ways. You can see in the test that I can still set all the options, and with not that much extra code.

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
anuraaga: Thanks, didn't notice / intend the change in behavior.

anuraaga (commented):
@bogdandrutu I thought the PR description being simple would be easier to understand but guess not :)

The real motivation for this is to remove the mutating `addSpanProcessor` - there's no use for it, unless we add `removeSpanProcessor` to parallel logging libraries ability to configure handlers. If you have code that only calls `add` once, it can happen when building the SDK itself no matter what.

After adding this method, I can refactor our tests, and then deprecate the mutator. This is orthogonal to #2231

bogdandrutu:
> The real motivation for this is to remove the mutating addSpanProcessor - there's no use for it, unless we add removeSpanProcessor to parallel logging libraries ability to configure handlers. If you have code that only calls add once, it can happen when building the SDK itself no matter what.

There is one problem with this approach "circular initialization dependencies". One of the reason "SpanProcessor" was not in the builder is because an implementation of a "SpanProcessor" (or an exporter that is needed to construct the SpanProcessor) may need/depend on a "[Tracer/Meter]Provider" to record metrics or some tracing.

Now the problem is:
* If implementation relies on the "global" then they will statically initialize their "Meter/Tracer" with a no-op or via SPI, and all their telemetry will not be correctly recorded.
* If they accept a provider during construction you don't have one because you are constructing it.

So how are you going to solve this initialization ordering issue?

anuraaga:
> If implementation relies on the "global" then they will statically initialize their "Meter/Tracer" with a no-op or via SPI, and all their telemetry will not be correctly recorded.

Hmm - it seems the problem already exists when trying to configure the global `OpenTelemetry`, there's an implicit order that the constructor (or maybe even just classloading) can't be called before calling the mutable `addSpanProcessor`.

Have also gotten feedback from teammates that it's very mysterious having to do a mutation, not being able to configure the SDK in one go.

So need to come up with something I think :)

The issue seems to mostly be about details of tracing / monitoring (exporters, remote samplers, etc). So it would be good if we can push the complexity down from the end-user layer (`OpenTelemetry` / `OpenTelemetrySdk`) to the observability layer, the exporters (well span processors) themselves as an example. With a DI framework, I'd just inject a lazy meter into and I think this will be fairly common for custom logic in apps. For us, we can have setters on the exporters and call them directly. Or we could generalize a little bit with interface like `LazyTracerInitializer` or such which builders are aware of. Using observability within one of the dependencies of observability is a special case so I think there are options where we can shift the complexity away from the actual entrypoint of OpenTelemetry.

bogdandrutu:
The problem is that then you make creating of other telemetry components more of a special case, if you don't care about that, then this may be a decent solution.

A second option is to always rely on SPI to load the core parts of the sdk, then the current OtelSDK.Builder is not a builder, it just sets the configuration to the already loaded core part.

From user perspective this happens in one shot, and it also allows telemetry components (except core part loaded with SPI) to access the global correctly.

That means user does not need to set the global when using our SDK which is also fine I think.

jkwatson:
ok, I think we have a plan to go forward with this now.

1) Add this option to the tracer provider builder
2) remove all non-component level options from the opentelemetry sdk builder

I'm ok with having 1 happen in a separate PR from 2. @bogdandrutu Your thoughts on sequencing?

anuraaga:
Is this PR blocked on anything? Let me know if there're any changes

jkwatson (approved):
Thanks!

jkwatson:
@bogdandrutu is there anything left to block this, since we removed the options from the main sdk builder itself?

bogdandrutu (approved):

PR #2225, opened 2020-12-08
Update coverage merging to follow Gradle example.

anuraaga:
The api folder split was not reflected into the manual list of projects we merge for coverage. I took this chance to update to Gradle's recommended pattern for merging, and doing so by automatically finding all projects we `maven-publish`.

https://docs.gradle.org/current/samples/sample_jvm_multi_project_with_code_coverage.html


jkwatson (approved):
Thanks!

PR #2226, opened 2020-12-08
Remove deprecated runtime-metrics and trace-utils packages.

anuraaga:
May as well get some code coverage back


bogdandrutu (approved):

jkwatson:
needs a rebase.

jkwatson (approved):
Thanks!

extensions/trace-utils/src/main/java/io/opentelemetry/extension/trace/MessageEvent.java
```diff
@@ -1,69 +0,0 @@
...
- * <p>It requires a {@link Type type} and a message id that serves to uniquely identify each
- * message. It can optionally have information about the message size.
- *
- * @deprecated This class will be removed in SDK 0.13.0. If these attributes are useful to you,
- *     consider filing an issue or a PR to have them added to {@link
- *     io.opentelemetry.api.trace.attributes.SemanticAttributes}.
- */
-@Immutable
-@Deprecated
-public final class MessageEvent {
```
bogdandrutu, line 31: FYI: This is for backwards compatibility with OpenCensus, so cannot be removed, if we still want to achieve the goal of backwards compatibility :)

bogdandrutu: This may leave in opencensus artifact, so ok to remove it.

bogdandrutu (approved):

PR #2227, opened 2020-12-08
Detect invalid use of makeCurrent in Kotlin coroutine.

anuraaga:
Since this reflects Kotlin runtime internals, naturally there may be some cat/mouse involved in checking this across Kotlin versions. But I think a best effort is still significantly better than nothing.


api/context/src/main/java/io/opentelemetry/context/StrictContextStorage.java
```diff
@@ -92,6 +92,17 @@ public Scope attach(Context context) {
     CallerStackTrace caller = new CallerStackTrace(context);
     StackTraceElement[] stackTrace = caller.getStackTrace();
 
+    // Detect invalid use from kotlin.
+    for (StackTraceElement element : stackTrace) {
+      if (element.getClassName().equals("kotlin.coroutines.jvm.internal.BaseContinuationImpl")
```
jkwatson, line 97: urgh. Any thoughts on how to keep this up-to-date if kotlin renames or changes their continuation implementation? Do we know what versions of kotlin this will work for, at this point?

jkwatson: as you say...best effort is better than nothing! Just wondering if we can provide some documentation about what is supported by this.

jkwatson:
looks like the build is failing due to this check kicking in?

trask (approved):
Nice

anuraaga:
Yup - realized I need to check the function is a suspending function, not any one. Will tweak

api/context/src/main/java/io/opentelemetry/context/StrictContextStorage.java
anuraaga: I added some doc to `Context` about the use of internal details, not sure how better to phrase it.

anuraaga:
Fixed this up and added more detection of suspending functions as well

extensions/kotlin/src/test/kotlin/io/opentelemetry/extension/kotlin/StrictContextWithCoroutinesTest.kt
```diff
@@ -141,20 +141,23 @@ class StrictContextWithCoroutinesTest {
         Context.current().with(ANIMAL, animal).makeCurrent().use {
             assertThat(Context.current().get(ANIMAL)).isEqualTo(animal)
             delay(10)
+            // May be on a different thread, in which case ANIMAL != animal!
```
jkwatson, line 144: is this comment explaining why we're not re-asserting? If so, can you just add that bit to the comment? Otherwise, explain what it is for. :)

extensions/kotlin/src/test/kotlin/io/opentelemetry/extension/kotlin/StrictContextWithCoroutinesTest.kt
```diff
@@ -141,20 +141,23 @@ class StrictContextWithCoroutinesTest {
...
         }
     }
 
     suspend fun suspendingFunctionMakeCurrentWithManualClose(animal: String) {
         val scope = Context.current().with(ANIMAL, animal).makeCurrent()
         assertThat(Context.current().get(ANIMAL)).isEqualTo(animal)
         delay(10)
+        // May be on a different thread, in which case ANIMAL != animal!
```
jkwatson, line 152: ditto on more explanatory comment

jkwatson (approved):
Just needs a teency comment clarification, please

Issue #2022, opened 2020-11-05
Confusing configuration story


iNikem:
I tried to gather into one place all bits and pieces that SDK allows to configure. So far I got (I concentrated only on tracing, no metrics here):

- SpanProcessor
- SpanExporter
- Sampler
- TracerProvider
- IdGenerator
- ContextPropagator
- BaggagePropagator
- Various limits on TraceConfig, e.g. max attribute count

Ignoring the last point from the above, other configurable objects can be split in two parts: unique, such as Sampler, IdGenerator and TracerProvider, and all others. Configuring unique objects obviously means replacing or _setting_ them. For non-unique objects we may want both to _add_ new objects in addition to defaults or to _set_ and replace default ones.

Now, let us take a look at the example code that tries to configure everything:

```
    OpenTelemetrySdk sdk = OpenTelemetrySdk.get();
    OpenTelemetrySdk.Builder sdkBuilder = sdk.toBuilder();
    TracerSdkManagement tracerSdkManagement = sdk.getTracerManagement();

    tracerSdkManagement.addSpanProcessor(new DemoSpanProcessor());
    TraceConfig currentConfig = tracerSdkManagement.getActiveTraceConfig();
    TraceConfig newConfig = currentConfig.toBuilder()
        .setSampler(new DemoSampler())
        .setMaxLengthOfAttributeValues(128)
        .build();
    tracerSdkManagement.updateActiveTraceConfig(newConfig);

    TracerSdkProvider tracerSdkProvider = TracerSdkProvider.builder().setIdsGenerator(new DemoIdGenerator()).build();
    sdkBuilder.setTracerProvider(tracerSdkProvider);

    sdkBuilder.setPropagators(DefaultContextPropagators.builder().addTextMapPropagator(new DemoPropagator()).build());

    OpenTelemetry.set(sdkBuilder.build());
```

First, is the code above idiomatic for this SDK? If yes, then let us analyse it.

- Different pieces of configuration must be set on different objects: TracerSdkManagement, TraceConfig and OpenTelemetrySdk.Builder
- Why TraceConfig encapsulates different limits (as numbers) and the single object, Sampler?
- If I want to add new SpanExporter, I have to think about SpanProcessor as well.
- If I want to add two SpanExporters, should I add two SpanProcessor? Should I use MultiSpanProcessor?
- If I want to change IdGenerator, I have to know about TracerSdkProvider and to set it.
- Which leads to #2018
[...]

PR #2231, opened 2020-12-08
Remove duplicate getters and builder options from SDK

bogdandrutu:
For reasons see https://github.com/open-telemetry/opentelemetry-java/pull/2223#issuecomment-741095083, and just apply the basic rule of defining an API, when in doubt don't add it, we can always add more functionality, but we cannot remove it.

### DIFF
--- a/sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
+++ b/sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
@@ -5,8 +5,6 @@
 
 package io.opentelemetry.sdk;
 
-import static java.util.Objects.requireNonNull;
-
 import io.opentelemetry.api.DefaultOpenTelemetry;
 import io.opentelemetry.api.DefaultOpenTelemetryBuilder;
 import io.opentelemetry.api.GlobalOpenTelemetry;
@@ -15,17 +13,9 @@ import io.opentelemetry.api.metrics.MeterProvider;
 import io.opentelemetry.api.trace.Tracer;
 import io.opentelemetry.api.trace.TracerProvider;
 import io.opentelemetry.context.propagation.ContextPropagators;
-import io.opentelemetry.sdk.common.Clock;
-import io.opentelemetry.sdk.internal.SystemClock;
 import io.opentelemetry.sdk.metrics.MeterSdkProvider;
-import io.opentelemetry.sdk.resources.Resource;
-import io.opentelemetry.sdk.trace.IdGenerator;
 import io.opentelemetry.sdk.trace.SdkTracerManagement;
 import io.opentelemetry.sdk.trace.SdkTracerProvider;
-import io.opentelemetry.sdk.trace.SpanProcessor;
-import io.opentelemetry.sdk.trace.config.TraceConfig;
-import java.util.ArrayList;
-import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 import javax.annotation.concurrent.ThreadSafe;
 
@@ -64,28 +54,11 @@ public final class OpenTelemetrySdk extends DefaultOpenTelemetry {
 
   private static final AtomicBoolean INITIALIZED_GLOBAL = new AtomicBoolean();
 
-  private final Clock clock;
-  private final Resource resource;
-
   private OpenTelemetrySdk(
       TracerProvider tracerProvider,
       MeterProvider meterProvider,
-      ContextPropagators contextPropagators,
-      Clock clock,
-      Resource resource) {
+      ContextPropagators contextPropagators) {
     super(tracerProvider, meterProvider, contextPropagators);
-    this.clock = clock;
-    this.resource = resource;
-  }
-
-  /** Returns the {@link Resource} for this {@link OpenTelemetrySdk}. */
-  public Resource getResource() {
-    return resource;
-  }
-
-  /** Returns the {@link Clock} for this {@link OpenTelemetrySdk}. */
-  public Clock getClock() {
-    return clock;
   }
 
   /** Returns the {@link SdkTracerManagement} for this {@link OpenTelemetrySdk}. */
@@ -95,12 +68,6 @@ public final class OpenTelemetrySdk extends DefaultOpenTelemetry {
 
   /** A builder for configuring an {@link OpenTelemetrySdk}. */
   public static class Builder extends DefaultOpenTelemetryBuilder {
-    private Clock clock;
-    private Resource resource;
-    private final List<SpanProcessor> spanProcessors = new ArrayList<>();
-    private IdGenerator idGenerator;
-    private TraceConfig traceConfig;
-
     /**
      * Sets the {@link SdkTracerProvider} to use. This can be used to configure tracing settings by
      * returning the instance created by a {@link SdkTracerProvider.Builder}.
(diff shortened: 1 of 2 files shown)


sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
```diff
@@ -144,130 +111,28 @@ public Builder setPropagators(ContextPropagators propagators) {
...
-    /**
-     * Sets the {@link Clock} to be used for measuring timings.
-     *
-     * <p>If you use {@link #setTracerProvider(TracerProvider)}, this will be ignored for purposes
-     * of configuring the TracerProvider.
-     *
-     * @param clock The clock to use for all temporal needs.
-     * @return this
-     */
-    public Builder setClock(Clock clock) {
```
anuraaga, line 156: While there could be an argument for removing the tracerprovider knobs here, I don't see why we would remove global concerns like clock / resource.

anuraaga (commented):
> when in doubt don't add it, we can always add more functionality, but we cannot remove it.

I have no doubt this makes the SDK significantly harder for users - it might take some guts but there are obvious shortcuts that reduce cognitive load for users in a clear way. Discoverability of configuration by only having one entry point is a win IMO. 

For some context, we're considering removing `setTracerProvider` from the `OpenTelemetrySdk.Builder` to reflect that the latter is the builder of the full SDK, the vastly common use case, not just parts, perhaps that makes the Sdk.Builder knobs feel more natural?

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
bogdandrutu: Because they are properties of the providers why would we need to have getters for Clock and Resource in the SDK?

bogdandrutu: Also it is a very poor design choice, if user sets TracerProvider and Resource the Resource is not set to the provider so it is an unexpected behavior for the user

bogdandrutu:
> I have no doubt this makes the SDK significantly harder for users - it might take some guts but there are obvious shortcuts that reduce cognitive load for users in a clear way. Discoverability of configuration by only having one entry point is a win IMO.

It is still one entry point, but with multiple levels of configuration:
```java
    OpenTelemetrySdk openTelemetry =
        OpenTelemetrySdk.builder()
            .setTracerProvider(
                TracerSdkProvider.builder()
                    .setClock(clock)
                    .setResource(resource)
                    .setIdGenerator(idGenerator)
                    .setTraceConfig(traceConfig)
                    .build())
            .setMeterProvider(
                MeterSdkProvider.builder().setResource(resource).setClock(clock).build())
            .build();
```

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
anuraaga: It's poor design to make it easy to set a different resource for tracer and meter right? The getters can probably be removed, they were with the thought that the providers might read them from the OpenTelemetry, but they can be pushed down via the builders instead. The idea is to remove `setTracerProvider` / `setMeterProvider` from here to make sure cross-signal concerns are handled by the SDK itself. We're making steps towards that and after making sure SDK can be configured as one unit, and each signal can also be configured as one unit, we can remove them.

bogdandrutu: Not sure about this, I can see reasons to have different clock instances (for metrics you don't need `us` precision, `ms` is more than enough). So the only part that is really common is `Resource`, does it worth all this duplicated code? Also as a user right now I am totally confused, where do I set `Resource`? Where do I set `SpanProcessors`?

1. So far we clearly agreed that the getters even if available to pass to the providers, they do not need to be public.

Do we agree that having duplicate way to configure the same thing is a maintainers problem?

anuraaga: Yeah - two ways to do the same thing is an issue of maintenance, which takes the back seat to user experience.

> Also as a user right now I am totally confused, where do I set Resource? Where do I set SpanProcessors?

I don't understand this - the configuration is on the SDK builder, why are you confused? (after we remove setTracerProvider / setMeterProvider)

iNikem:
@bogdandrutu Have you seen https://github.com/open-telemetry/opentelemetry-java/issues/2022 ?

jkwatson (changes requested):

iNikem:
First, I totally agree with @jkwatson : these PRs lack clear understanding of the final destination. I am sure that @bogdandrutu has one in his head, but external readers, like me, a confused.

Second, I am beginning to agree with Bogdan. Take a look at the following code:
```
    OpenTelemetrySdk.Builder sdkBuilder = OpenTelemetrySdk.builder()
        .addSpanProcessor(mock(SpanProcessor.class))
        .addSpanProcessor(SimpleSpanProcessor.builder(mock(SpanExporter.class)).build())
        .setIdGenerator(mock(IdGenerator.class))
        .setPropagators(ContextPropagators.create(mock(TextMapPropagator.class)))
        .setClock(mock(Clock.class))
        .setResource(mock(Resource.class))
        .setTracerProvider(mock(TracerProvider.class));
```

Can you predict the final state of the OpenTelemetrySdk from this? As a hint, here is a javadoc for the last line, `setTracerProvider`:

> If you use this method, it is assumed that you are providing a fully configured TracerSdkProvider, and other settings will be ignored.

Will ALL my previous configuration be ignored? Or only some parts of it? Which ones?

iNikem:
As a user of SDK configuration code, I want to see examples for two extremes:

- The maximum that I can configure (including TraceConfig that is mostly ignored by everybody)
- The minimum that I **have** to configure in order to have working SDK.

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
jkwatson: > Yeah - two ways to do the same thing is an issue of maintenance, which takes the back seat to user experience.

But, we already have tons of ways to configure things...env vars, system properties, programmatic, SPI. How do we draw the line at what the "right" number of ways is to configure the API and SDK?

jkwatson:
I believe we've decided that this is the route we need to go, since we're going to need to have stand-alone signal-specific SDK components, and we don't want to be left with a confusing extra way of doing the configuration. Thanks @bogdandrutu !

[also, this needs a rebase]

jkwatson (approved):
Thanks!

bogdandrutu:
@jkwatson done

anuraaga (commented):
Do we need to deprecate instead of remove @jkwatson ?

anuraaga (approved):

anuraaga:
I'm good with either though

jkwatson:
I'd deprecate anything that we've already released, but obviously if we haven't released it yet, we don't need to deprecate it.

bogdandrutu:
I think in the SDK we didn’t do that (see PR that changed the name of SDK classes)

anuraaga:
We've been doing a best-effort, without leaving too much duplication like duplicating entire classes, I think. But yeah for this release, we have some big things anyways so let's not worry too much about deprecations.

jkwatson:
> I think in the SDK we didn’t do that (see PR that changed the name of SDK classes)

Yeah...in that case we probably should have, but water under the bridge at this point, I think.

PR #2232, opened 2020-12-08
Small nit in internal members order

bogdandrutu:

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/DefaultOpenTelemetry.java
+++ b/api/all/src/main/java/io/opentelemetry/api/DefaultOpenTelemetry.java
@@ -24,6 +24,11 @@ import javax.annotation.concurrent.ThreadSafe;
 @ThreadSafe
 public class DefaultOpenTelemetry implements OpenTelemetry {
   private static final Object mutex = new Object();
+  @Nullable private static volatile OpenTelemetry globalOpenTelemetry;
+
+  private final TracerProvider tracerProvider;
+  private final MeterProvider meterProvider;
+  private volatile ContextPropagators propagators;
 
   static OpenTelemetry getGlobalOpenTelemetry() {
     if (globalOpenTelemetry == null) {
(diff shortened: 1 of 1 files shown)


api/all/src/main/java/io/opentelemetry/api/DefaultOpenTelemetry.java
```diff
@@ -24,6 +24,11 @@
 @ThreadSafe
 public class DefaultOpenTelemetry implements OpenTelemetry {
   private static final Object mutex = new Object();
+  @Nullable private static volatile OpenTelemetry globalOpenTelemetry;
```
anuraaga, line 27: static looks good, I would keep the instance fields where they were and move the constructor to be just below them, instead of moving them above the statics

jkwatson: I personally prefer this order, with the fields all above all the methods. Does the google style guide specify one way or another? If not, we should have a documented order, so we don't need to think about it. :)

bogdandrutu: I always have (in my code) fields first then methods.

jkwatson: @anuraaga you ok with switching to this style? If so, I can add it to the contributing doc, so it's not a point of debate. I checked the google style guide and it explicitly doesn't specify this. Only that things are in a "logical" order.

anuraaga: Oops sorry for missing replying

I've always found statics before non-statics to be easier to reason about - lifecycle and end-user usage is totally different. The instance fields can't be used in the static methods no matter what so they seem distracting to be grouped with them. I wouldn't block the change since 2/3 wins but please give it one more consideration given this reasoning :)

anuraaga (approved):

jkwatson (approved):

PR #2233, opened 2020-12-08
Split global functionality into a separate class

bogdandrutu:
This ensure clear separation of functionality. This PR does not change any functionality, it just restructures the code to separate global functionality from
the Default implementation. Also helps to ensure that Global initialization does not happen by mistake when working only with the default implementation.

Also it helps with methods like `OpenTelemetry.get()` and `OpenTelemetry.set()` to understand they are interacting with global instance

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
+++ b/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
@@ -1,0 +1,166 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.api;
+
+import static java.util.Objects.requireNonNull;
+
+import io.opentelemetry.api.metrics.Meter;
+import io.opentelemetry.api.metrics.MeterProvider;
+import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.api.trace.TracerProvider;
+import io.opentelemetry.context.propagation.ContextPropagators;
+import io.opentelemetry.spi.OpenTelemetryFactory;
+import io.opentelemetry.spi.metrics.MeterProviderFactory;
+import io.opentelemetry.spi.trace.TracerProviderFactory;
+import javax.annotation.Nullable;
+
+/**
+ * A global singleton for the entrypoint to telemetry functionality for tracing, metrics and
+ * baggage.
+ *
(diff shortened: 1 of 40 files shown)


api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
```diff
@@ -0,0 +1,166 @@
...
+ * <p>If using the OpenTelemetry SDK, you may want to instantiate the {@link OpenTelemetry} to
+ * provide configuration, for example of {@code Resource} or {@code Sampler}. See {@code
+ * OpenTelemetrySdk} and {@code OpenTelemetrySdk.builder} for information on how to construct the
+ * SDK {@link OpenTelemetry}.
+ *
+ * @see TracerProvider
+ * @see MeterProvider
+ * @see ContextPropagators
+ */
+public final class GlobalOpenTelemetry {
```
anuraaga, line 38: This is a big improvement. I think we need to compare with removing the `getGlobalTracer` type methods in favor of `OpenTelemetry.get().getTracer()` or `OpenTelemetry.getGlobal().getTracer()` or `getGlobalOpentelemetry().getTracer()` (static import). I think they're all fine, latter might lead to more testable user code by reducing the number of static methods involved.

bogdandrutu: I would still not have the "global" functionality on the interface `OpenTelemetry`. Having this separation is a win in my opinion. Not sure I understand what comparison you are asking me to do.

jkwatson:
I think this is a fantastic change. It's definitely a big breaking change, though. Should we go through a deprecation cycle on the global methods on the OpenTelemetry API to give users a chance to adapt? This seems like a big enough break that it warrants a deprecation cycle.

bogdandrutu:
@jkwatson done.

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
jkwatson: @anuraaga any further thoughts at this point? I'm inclined to this this PR the 👍 , but want to make sure your concerns are considered.

jkwatson (approved):
Thanks!

jkwatson:
@bogdandrutu this needs a rebase. Thanks!

bogdandrutu:
@jkwatson done.

PR #2235, opened 2020-12-09
Remove unnecessary builder for OpenTelemetry instance

bogdandrutu:
Every implementation should provide their own way of constructing the instance.


anuraaga (approved):
Agree that after removing `toBuilder` this makes a lot of sense

jkwatson (changes requested):

bogdandrutu:
@jkwatson this is not about SDK configuration, this is an API change. And I don't think we should require all API implementations to expose a Builder.

jkwatson:
If we do decide to remove this, it's a breaking API change, and so we should deprecate the method for a release before we remove it.

bogdandrutu:
Dropping this, since it was easier to recreate a new PR with deprecations.

PR #2239, opened 2020-12-09
Remove unnecessary setPropagators from OpenTelemetry interface

bogdandrutu:
Example usage in the `Application.java`.

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/DefaultOpenTelemetry.java
+++ b/api/all/src/main/java/io/opentelemetry/api/DefaultOpenTelemetry.java
@@ -28,10 +28,10 @@ public class DefaultOpenTelemetry implements OpenTelemetry {
 
   private final TracerProvider tracerProvider;
   private final MeterProvider meterProvider;
-
   private volatile ContextPropagators propagators;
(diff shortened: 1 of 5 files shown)


jkwatson (changes requested):

bogdandrutu:
@jkwatson this is not about SDK configuration, this is an API change. And I don't think we should require all API implementations to support set propagators.

jkwatson:
Sorry, yes, this isn't SDK configuration code, it's API configuration code. I'm definitely not opposed to this change, but it is definitely a breaking change to the API, so we should deprecate for at least one release before removing it. 

Just to be clear: I think this is a good reduction in surface, but I need to see a full story around API & SDK configuration to understand where we're going and where we think we're going to end up with how to configure this thing.

jkwatson:
ok! I think this is a good change, moving toward only setting propagators at implementation creation time. 
But, we should definitely deprecate before removing.

And, this needs a rebase. :)

Thanks!

carlosalberto (approved):

jkwatson (approved):
Thanks!

PR #2241, opened 2020-12-09
Move TraceConfig.Builder to the top level

anuraaga:
Consistency, and it looks like we get some better validation behavior (throw on set, not build) to boot.

I hadn't actually noticed we were using `Utils` here. In our meeting I think we were all ok with expecting the bom to be used to align dependencies so I'm perfectly fine with cross-referencing internal packages in this repo.

### DIFF
--- a/sdk-extensions/zpages/src/main/java/io/opentelemetry/sdk/extension/zpages/TraceConfigzZPageHandler.java
+++ b/sdk-extensions/zpages/src/main/java/io/opentelemetry/sdk/extension/zpages/TraceConfigzZPageHandler.java
@@ -7,6 +7,7 @@ package io.opentelemetry.sdk.extension.zpages;
 
 import io.opentelemetry.sdk.trace.TracerSdkManagement;
 import io.opentelemetry.sdk.trace.config.TraceConfig;
+import io.opentelemetry.sdk.trace.config.TraceConfigBuilder;
 import io.opentelemetry.sdk.trace.samplers.Sampler;
 import java.io.OutputStream;
 import java.io.PrintStream;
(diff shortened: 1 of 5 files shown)


bogdandrutu (changes requested):
The idea of offering only ToBuilder was important, because we want to ensure that you change only the parts you want and leave everything else the same in case of updating the current active config.

I think this PR changes also that behavior which should not be changed in a refactoring PR.

jkwatson:

bogdandrutu:
@jkwatson I don't think this is a change of configuration, this is just trying to align on the Builder not being a sub-class.

anuraaga:
@bogdandrutu Our standard pattern is `builder()` returning a builder populated with defaults, isn't it the same but less round-about than `getDefault().toBuilder()`?

bogdandrutu:
> @bogdandrutu Our standard pattern is builder() returning a builder populated with defaults, isn't it the same but less round-about than getDefault().toBuilder()?

Not necessary, the idea being that most likely devs will do mistakes when they try to Update the current config sampler for example. With your current change I am confident that someone will write without even thinking that they should start from the current config:

```java
TraceConfig traceConfig TraceConfig.builder().setSampler(foo).build();
tracerProviderSdk.UpdateConfig(traceConfig);
```

What user should do is, I think having an only `toBuilder` will trigger a thought that wait I don't want to start from default:

```java
TraceConfig traceConfig = tracerProviderSdk.CurrentConfig();
tracerProviderSdk.UpdateConfig(traceConfig.ToBuilder().setSampler(foo).build());
```

bogdandrutu:
@anuraaga also the addition of the `builder()` method is a different change than what is described in the PR, and should definitely be a separate PR.

jkwatson (changes requested):
Blocking until we can have a comprehensive design doc that the addition of a `builder()` method either does or doesn't align with.

bogdandrutu:
@jkwatson I think this is not constructive, my suggestion of not adding the new method does not change the current code, so I am not sure I understand the motivation of doing a change that we agree already.

jkwatson:
> @jkwatson I think this is not constructive, my suggestion of not adding the new method does not change the current code, so I am not sure I understand the motivation of doing a change that we agree already on (changing the builder pattern to not be a sub-class) is blocked by a new design doc requirement

Happy to unblock if we get rid of the `builder()` method from this PR. That piece (as you rightly called out) is exposing a new configuration surface.

anuraaga:
Makes sense - removed `builder()` for now

bogdandrutu (approved):

jkwatson (approved):
Thanks! Sorry for being a PitA about this!

bogdandrutu:
@jkwatson always good to have check and balance :)

PR #2242, opened 2020-12-09
Implement more Span default methods.

anuraaga:
I noticed that shims like in https://github.com/open-telemetry/opentelemetry-java/pull/2060 would be easier to write with boring implementations in default methods.


jkwatson (approved):
Thanks! java 8 FTW!

bogdandrutu (approved):

PR #2243, opened 2020-12-09
Update types to reflect the names in proto

bogdandrutu:
NON_MONOTONIC_SUM will be soon merged with SUM and offer the "isMonotonic" as a property.

### DIFF
--- a/exporters/logging/src/test/java/io/opentelemetry/exporter/logging/LoggingMetricExporterTest.java
+++ b/exporters/logging/src/test/java/io/opentelemetry/exporter/logging/LoggingMetricExporterTest.java
@@ -74,7 +74,7 @@ class LoggingMetricExporterTest {
                 "counterOne",
                 "A simple counter",
                 "one",
-                MetricData.Type.MONOTONIC_LONG,
+                MetricData.Type.SUM_LONG,
                 Collections.singletonList(
                     LongPoint.create(
                         nowEpochNanos, nowEpochNanos + 245, Labels.of("z", "y", "x", "w"), 1010))),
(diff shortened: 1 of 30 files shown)


breedx-splk (commented):
The execution of the rename seems fine, and the comments were updated.  

Is the plan then to have `isMonotonic()` on all of the `Type`s, so that one can also ask `SUMMARY.isMonotonic()` or `GAUGE_LONG.isMonotonic()`?  That doesn't seem great...but maybe you had a different approach in mind?

exporters/logging/src/test/java/io/opentelemetry/exporter/logging/LoggingMetricExporterTest.java
```diff
@@ -74,7 +74,7 @@ void testExport() {
                 "counterOne",
                 "A simple counter",
                 "one",
-                MetricData.Type.MONOTONIC_LONG,
+                MetricData.Type.SUM_LONG,
```
Oberon00, line 77: LONG_SUM would sound more natural IMHO.

bogdandrutu:
@breedx-splk I will design in a same way as the proto, based on the type we return an Object that has the Points + extra metadata specific to that type (see https://github.com/open-telemetry/opentelemetry-proto/blob/master/opentelemetry/proto/metrics/v1/metrics.proto)

exporters/logging/src/test/java/io/opentelemetry/exporter/logging/LoggingMetricExporterTest.java
bogdandrutu: good point. will change.

jkwatson (approved):
Thanks!

PR #2245, opened 2020-12-09
Rename batcher to instrumentaccumulator, remove active batcher (not used)

bogdandrutu:
The difference compared with the specs and go implementation is that the Accumulator has a map between InstrumentDescription and InstrumentAccumulator which will have it's own map LabelSet to Aggregator instead of a global map (InstrumentDescription + LabelSet) -> Aggregator


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/MeterSdk.java
```diff
@@ -112,8 +112,11 @@ public BatchRecorderSdk newBatchRecorder(String... keyValuePairs) {
     return result;
   }
 
-  /** Creates a {@link Batcher}, by using the {@link ViewRegistry} to do the actual work. */
-  Batcher createBatcher(
+  /**
+   * Creates a {@link InstrumentAccumulator}, by using the {@link ViewRegistry} to do the actual
+   * work.
+   */
+  InstrumentAccumulator createBatcher(
```
jkwatson, line 119: nit: this method name would probably need to be updated to `createAccumulator`

bogdandrutu: It will be moved and renamed in a followup.

jkwatson (approved):
Thanks!

PR #2248, opened 2020-12-09
Remove all implementations of the InstrumentAccumulator, kepp only the AllLabels

bogdandrutu:
This PR fixes a small bug where the AllLabels implementation was expecting "unmapped" boolean, which was passed correctly by the Sync instruments, but incorrectly by the Async.
This causes an unnecessary allocation for all the async instruments recordings.


jkwatson (approved):
Thanks! Good cleanup.

PR #2249, opened 2020-12-09
remove test dependency on grpc

breedx-splk:
This removes an unnecessary dependency from `jaeger-thrift` on grpc


bogdandrutu (approved):

jkwatson (approved):
Thanks!

PR #2250, opened 2020-12-09
Another option for release versioning proposal

jkwatson:
As an alternative approach to #2212 

This options adds `-alpha` to the released *versions* of immature libraries, rather than adding `-experimental` to the artifacts.

Another option to fulfill  https://github.com/open-telemetry/opentelemetry-specification/issues/1267

### DIFF
--- a/docs/rationale.md
+++ b/docs/rationale.md
@@ -70,7 +70,7 @@ Purely for illustration purposes, not intended to represent actual releases:
 
 - `v1.0.0` release:
     - `io.opentelemetry:opentelemetry-api:1.0.0`
-        - Contains APIs for tracing, baggage, propagators (via the context dependency)
+        - Includes APIs for tracing, baggage, context, propagators (via the context dependency)
     - `io.opentelemetry:opentelemetry-api-metrics:1.0.0-alpha`
         - Note: packages here are the final package structure: `io.opentelemetry.api.metrics.*`
     - `io.opentelemetry:opentelemetry-sdk-trace:1.0.0`
--- a/docs/rationale.md
+++ b/docs/rationale.md
@@ -41,7 +41,7 @@ function, *with no recompilation required*, with any API+SDK that has the same m
 SDK library `opentelemetry-sdk-trace:1.11.33` plus `opentelemetry-api-trace:1.11.33`. 
     - We call this requirement the "ABI" compatibility requirement for "Application Binary Interface" compatibility. 
 - SDK Stability:
-    - Public portions of the SDK (constructors, configuration, und-user interfaces) must remain backwards compatible.
+    - Public portions of the SDK (constructors, configuration, end-user interfaces) must remain backwards compatible.
         - Precisely what this includes has yet to be delineated. 
     - Internal interfaces are allowed to break.
 
@@ -91,4 +91,3 @@ Purely for illustration purposes, not intended to represent actual releases:
         - Note: packages here have not changed from the experimental jar...just a jar rename happened.
     - `io.opentelemetry:opentelemetry-sdk-all:1.15.0`
         - The SDK side of io.opentelemetry:opentelemetry-api:1.15.0
-
--- a/docs/rationale.md
+++ b/docs/rationale.md
@@ -38,12 +38,14 @@ Because this is a common usage for spans, we do not support try-with-resources.
     - Once the API for a given signal (spans, logs, metrics, baggage) has been officially released, code instrumented with that API module will 
 function, *with no recompilation required*, with any API+SDK that has the same major version, and equal or greater minor or patch version. 
     - For example, libraries that are instrumented with `opentelemetry-api-trace:1.0.1` will function, at runtime with 
-SDK library `opentelemetry-sdk-trace:1.11.33` plus `opentelemetry-api-trace:1.11.33`. 
+SDK library `opentelemetry-sdk-trace:1.11.33` plus `opentelemetry-api-trace:1.11.33` (or whatever specific versions are specified by 
+      the bom version `1.11.33`, if the individual versions have diverged). 
     - We call this requirement the "ABI" compatibility requirement for "Application Binary Interface" compatibility. 
 - SDK Stability:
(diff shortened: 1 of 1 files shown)


docs/rationale.md
```diff
@@ -23,4 +23,72 @@ try (Scope scope = TracingContextUtils.currentContextWith(span)) {
...
+    mature usage, without having to change imports. 
+    - SDK modules for immature signals will also be versioned with an "-alpha" suffix, in parallel to their API modules.
+    
+### Examples
+
+Purely for illustration purposes, not intended to represent actual releases:
+
+- `v1.0.0` release:
+    - `io.opentelemetry:opentelemetry-api:1.0.0`
+        - Contains APIs for tracing, baggage, propagators (via the context dependency)
```
bogdandrutu, line 73: And context itself correct?

jkwatson: yes...this was illustrative, not intended to be complete or 100% accurate.

jkwatson: added for clarity, though

tedsuo:
👍 I think this is an improvement, provided that package managers don't get confused into thinking that -alpha libraries are not part of the same release as non-alpha libraries.

jkwatson:
> 👍 I think this is an improvement, provided that package managers don't get confused into thinking that -alpha libraries are not part of the same release as non-alpha libraries.

Java will be fine, especially since we won't include the -alpha libraries in the BOM at all. You'll have to depend on them explicitly and intentionally in order to use them.

trask:
I like this option also. It gives a nice experience / smooth transition for early adopters from alpha/beta -> RC -> GA.

maxgolov:
Although I'm working on C++ SDK, I like this option much better than appending `-experimental` to package name. Adding `-experimental` to package name would not scale in Azure environment, and is unnatural for most C++ package managers.

docs/rationale.md
```diff
@@ -23,4 +23,72 @@ try (Scope scope = TracingContextUtils.currentContextWith(span)) {
...
+- API Stability: 
+    - Once the API for a given signal (spans, logs, metrics, baggage) has been officially released, code instrumented with that API module will 
+function, *with no recompilation required*, with any API+SDK that has the same major version, and equal or greater minor or patch version. 
+    - For example, libraries that are instrumented with `opentelemetry-api-trace:1.0.1` will function, at runtime with 
+SDK library `opentelemetry-sdk-trace:1.11.33` plus `opentelemetry-api-trace:1.11.33`. 
+    - We call this requirement the "ABI" compatibility requirement for "Application Binary Interface" compatibility. 
+- SDK Stability:
+    - Public portions of the SDK (constructors, configuration, und-user interfaces) must remain backwards compatible.
+        - Precisely what this includes has yet to be delineated. 
+    - Internal interfaces are allowed to break.
```
anuraaga, line 46: I guess this is referring to interfaces in the SDK that have the public modifier but don't fall into the above bucket. While I know `Precisely what this includes has yet to be delineated. ` do you have any example of it? `BatchSpanProcessor`?

jkwatson: I copied this out of Ted's otep...I think it just means non-public "internals" of the SDK. It might not be a valuable thing to call out.

anuraaga: Ok - yeah it might be worth calling out but it applies to both API and SDK so weird for me to see it here, which I guess is my comment on the OTEP too so just repeating the feedback it seems :)

anuraaga (commented):
Doc mostly LGTM, thanks

docs/rationale.md
```diff
@@ -23,4 +23,72 @@ try (Scope scope = TracingContextUtils.currentContextWith(span)) {
...
+### Goals
+
+- API Stability: 
+    - Once the API for a given signal (spans, logs, metrics, baggage) has been officially released, code instrumented with that API module will 
+function, *with no recompilation required*, with any API+SDK that has the same major version, and equal or greater minor or patch version. 
+    - For example, libraries that are instrumented with `opentelemetry-api-trace:1.0.1` will function, at runtime with 
+SDK library `opentelemetry-sdk-trace:1.11.33` plus `opentelemetry-api-trace:1.11.33`. 
+    - We call this requirement the "ABI" compatibility requirement for "Application Binary Interface" compatibility. 
+- SDK Stability:
+    - Public portions of the SDK (constructors, configuration, und-user interfaces) must remain backwards compatible.
```
iNikem, line 44:
```suggestion
    - Public portions of the SDK (constructors, configuration, end-user interfaces) must remain backwards compatible.
```

iNikem (approved):

docs/rationale.md
```diff
@@ -23,4 +23,72 @@ try (Scope scope = TracingContextUtils.currentContextWith(span)) {
...
+
+- This project uses semver v2, as does the rest of OpenTelemetry.
+
+### Goals
+
+- API Stability: 
+    - Once the API for a given signal (spans, logs, metrics, baggage) has been officially released, code instrumented with that API module will 
+function, *with no recompilation required*, with any API+SDK that has the same major version, and equal or greater minor or patch version. 
+    - For example, libraries that are instrumented with `opentelemetry-api-trace:1.0.1` will function, at runtime with 
+SDK library `opentelemetry-sdk-trace:1.11.33` plus `opentelemetry-api-trace:1.11.33`.
```
iNikem, line 41: But we planned to version SDK independently from API, right?

anuraaga: I think this depended on how hard we go on preserving SDK compatibility. The more precise statement is indeed probably the use of the BOM for the latest version rather than mentioning them explicitly.

jkwatson: I can update to refer to the BOM. Whether we actually end up versioning the SDK independently from the API remains to be seen (I sure hope we don't have to).

docs/rationale.md
jkwatson: updated

docs/rationale.md
jkwatson: done

anuraaga (approved):

bogdandrutu (approved):

PR #2255, opened 2020-12-10
Rename accumulator to processor to match Go

bogdandrutu:
I was wrong initially, what we had as Batcher is actually the Processor in Go.

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractSynchronousInstrument.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractSynchronousInstrument.java
@@ -17,14 +17,14 @@ abstract class AbstractSynchronousInstrument<B extends AbstractBoundInstrument>
     extends AbstractInstrument {
   private final ConcurrentHashMap<Labels, B> boundLabels;
   private final ReentrantLock collectLock;
-  private final InstrumentAccumulator instrumentAccumulator;
+  private final InstrumentProcessor instrumentProcessor;
 
   AbstractSynchronousInstrument(
-      InstrumentDescriptor descriptor, InstrumentAccumulator instrumentAccumulator) {
+      InstrumentDescriptor descriptor, InstrumentProcessor instrumentProcessor) {
     super(descriptor);
     boundLabels = new ConcurrentHashMap<>();
     collectLock = new ReentrantLock();
-    this.instrumentAccumulator = instrumentAccumulator;
+    this.instrumentProcessor = instrumentProcessor;
   }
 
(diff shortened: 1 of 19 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractSynchronousInstrument.java
```diff
@@ -68,14 +68,13 @@ public B bind(Labels labels) {
...
       }
-      return instrumentAccumulator.completeCollectionCycle();
+      return instrumentProcessor.completeCollectionCycle();
     } finally {
       collectLock.unlock();
     }
   }
 
-  abstract B newBinding(InstrumentAccumulator instrumentAccumulator);
+  abstract B newBinding(InstrumentProcessor instrumentProcessor);
```
jkwatson, line 79: I was following this rename until I hit this. Why are we passing a processor to create a new binding for an instrument? That doesn't seem like the role of the processor to me. I could think that the binding would be registered with the accumulator, not the processor.

bogdandrutu: Actually I need to pass only an Aggregator you can see that all the Bound constructor just call getAggregator. I decided to do that immediately after https://github.com/open-telemetry/opentelemetry-java/pull/2257.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/LongCounterSdk.java
```diff
@@ -35,15 +34,15 @@ public void add(long increment) {
   }
 
   @Override
-  BoundInstrument newBinding(InstrumentAccumulator instrumentAccumulator) {
-    return new BoundInstrument(instrumentAccumulator);
+  BoundInstrument newBinding(InstrumentProcessor instrumentProcessor) {
+    return new BoundInstrument(instrumentProcessor);
```
jkwatson, line 38: this definitely doesn't feel right to me. The Accumulator should be handling the binding, so it can keep track of when to released the resources associated with the labelset. I don't think that's the job of the Processor, which needs to hold long-term state of all the labelsets against which recordings have been made, even if they're not currently being actively recorded.

bogdandrutu: It is not, as explained in the other comment. I think you are keep reviewing this.

jkwatson (approved):
I'm not 100% sure I see where this is all headed, but let's give it a go.

PR #2256, opened 2020-12-10
Change ComponentRegistry to be a final class that accepts a factory method

bogdandrutu:
Java8 ftw :)


jkwatson (approved):
👍

jkwatson:
I like this change. Removing an extra class hierarchy is a win.

PR #2257, opened 2020-12-10
Change AbstractSynchronousInstrument to accept a bound factory instead of abstract method

bogdandrutu:

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/DoubleCounterSdk.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/DoubleCounterSdk.java
@@ -16,7 +16,7 @@ final class DoubleCounterSdk extends AbstractSynchronousInstrument<BoundInstrume
 
   private DoubleCounterSdk(
       InstrumentDescriptor descriptor, InstrumentProcessor instrumentProcessor) {
-    super(descriptor, instrumentProcessor);
+    super(descriptor, instrumentProcessor, BoundInstrument::new);
   }
 
   @Override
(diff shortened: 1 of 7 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/DoubleCounterSdk.java
```diff
@@ -16,7 +16,7 @@
 
   private DoubleCounterSdk(
       InstrumentDescriptor descriptor, InstrumentAccumulator instrumentAccumulator) {
-    super(descriptor, instrumentAccumulator);
+    super(descriptor, instrumentAccumulator, BoundInstrument::new);
```
jkwatson, line 19: If all the implementations are the same, can we just inline this into the AbstractSynchronousInstrument constructor and call it a day? Are there ever likely to be other implementations for creating a bound instrument? Or is this the only way to cleanly get the type capture to work right?

bogdandrutu: They are not the same. BoundInstrument is a different sub-class in the parent caller class

jkwatson: oh, you're right. I totally missed that in the github view. That's what I get for not pulling down the code and looking at it fully. 👍

jkwatson (approved):
👍

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/DoubleCounterSdk.java
bogdandrutu: Thanks :)

PR #2258, opened 2020-12-10
Change bound factory to accept Aggregator

bogdandrutu:


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/DoubleCounterSdk.java
```diff
@@ -37,8 +38,8 @@ public void add(double increment) {
   static final class BoundInstrument extends AbstractBoundInstrument
       implements DoubleCounter.BoundDoubleCounter {
 
-    BoundInstrument(InstrumentProcessor instrumentProcessor) {
-      super(instrumentProcessor.getAggregator());
+    BoundInstrument(Aggregator aggregator) {
+      super(aggregator);
```
bogdandrutu, line 42: @jkwatson see what I mentioned in the previous PR :)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractSynchronousInstrument.java
```diff
@@ -41,7 +42,7 @@ public B bind(Labels labels) {
     }
 
     // Missing entry or no longer mapped, try to add a new entry.
-    binding = boundFactory.apply(instrumentProcessor);
+    binding = boundFactory.apply(instrumentProcessor.getAggregator());
```
jkwatson, line 45: interesting that the coupling is the reverse of what is in @jmacd's design in the spec. I don't think it's necessarily a problem, but it might point at some tweaks to the SDK spec being necessary.

jkwatson (approved):
👍

PR #2261, opened 2020-12-10
Use standard Java interfaces instead of defining custom ones

bogdandrutu:


jkwatson (approved):
Thanks! Agreed preferring built-in interfaces to custom is good.

Issue #1909, opened 2020-10-28
Provide an SPI module for SDK interactions


kenfinnigan:
**Is your feature request related to a problem? Please describe.**
Currently, any SPI between the SDK and exporters is within the SDK modules.

Having the SPI within the SDK implementation has several impacts:

- There's no API boundary between the SDK and exporters. Exporters can potentially utilize parts of the SDK that weren't intended
- It complicates implementing exporters for those not familiar with the project, as it requires knowing which interfaces/classes from the SDK need to be implemented
- A separate module would make it possible to independently version, resulting in longer compatibility between SDK and exporters if the SPI has not been modified

**Describe the solution you'd like**
Either a pure SPI module that only contains interfaces or more of a "common" SPI module that includes the interfaces but also common classes that are required by both the SDK and exporters. Whichever approach can be taken will be influenced by the current SDK implementation and how easily some of those "common" SPI pieces can be extracted.

I'm happy to prototype this work if there's interest

jkwatson:
@kenfinnigan since we've moved all the SPI into the autoconfigure module, does that satisfy your needs?

kenfinnigan:
I don't believe it does, as something like https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/SpanExporter.java I would consider an exporter SPI which is present in the SDK.

Anything implementing an exporter needs to depend on the entire SDK to access that interface to implement it, leading to the possible problems I described above.

Is there any particular need or requirement for exporters to depend on the full SDK?

jkwatson:
oh yeah... I forgot about that specific bit. Basically you'd like a stable set of interfaces (not SPI in particular) for what @tedsuo has been calling "SDK Plugin Interfaces". Maybe we can update the description of this issue so I don't get confused by the "SPI" bit in the title in 2 months when I scrub though the issues again.  :)

kenfinnigan:
I took SPI as just meaning interfaces between the SDK and external bits, maybe there's a specific meaning I'm not aware of which is where the confusion came from.

I've updated the title to hopefully something more meaningful. If not, feel free to adjust

anuraaga:
[...]

PR #2262, opened 2020-12-10
Add a doc outlining possible approach to SDK configuration.

anuraaga:
@jkwatson asked me to make a proposal of SDK configuration, and I've tried to outline goals and approaches here. This might not be the best approach, but it keeps these principles in mind

- Keep the end-user API as simple as possible
- It's ok to push some complexity downstream for less-common use cases like dynamic config, partial SDKs
- It's ok to expect more domain knowledge from authors of SDK components / extensions than of end users
- Keep core lightweight and let external configuration mechanisms take on the brunt of configuration

Perhaps these principles aren't good principles :) But this can hopefully give something to talk about.

### DIFF
--- a/docs/sdk-configuration.md
+++ b/docs/sdk-configuration.md
@@ -16,6 +16,9 @@ their application developers. May write custom SDK extensions such as exporters,
 with their internal infrastructure and as such have some familiarity with tracing at least as the
 SDK presents it.
 
+- Telemetry extension authors. Write custom SDK extensions, often to support a particular backend.
+Verify familiar with telemetry.
+
 - OpenTelemetry maintainers. Write the SDK code.
 
 When making decisions, especially about complexity, we always prioritize the application developers,
@@ -29,13 +32,15 @@ to be much more of them than other audiences.
 ### Goals
 
 - Provide a single entrypoint to configuring the SDK. For end-users, less familiar with the SDK, we
-want to have everything together to provide discoverability and simpler end-user code.
+want to have everything together to provide discoverability and simpler end-user code. If there are
+several, clear use cases which benefit from different entrypoints, we could have multiple
+corresponding to each one.
 
 - Fit well with common Java idioms such as dependency injection, or common frameworks like Spring.
 
 - Reduce the chance of gotchas or configuration mishaps.
 
-- Aim for optimal performance for the most common use case.
+- Aim for optimal performance of the configured SDK for the most common use case.
 
 ### Non-goals
 
@@ -51,8 +56,8 @@ components.
 
 ## Configuring an instance of the SDK
 
-The SDK consists of configuration options for all the signals it supports. Users all have different
-requirements for how they use the SDK, for example they may use different exporters depending on
+The SDK exposes configuration options for all the signals it supports. Users all have different
+requirements for how they use the SDK; for example they may use different exporters depending on
 their backend. Because we cannot guess the configuration the user needs, we expect that the SDK must
 be configured by the user before it can be used.
 
@@ -78,7 +83,7 @@ OpenTelemetrySdk openTelemetry = OpenTelemetrySdk.builder()
 
 This
 
-- Exports spans using OTLP to spans-service
+- Exports spans using OTLP to `spans-service`
   - Uses the BatchSpanProcessor (or a batching exporter wraper) which we consider the best default 
   for exporting. It has defaults for queue size, timeout, etc
 - Exposes metrics using Prometheus format
@@ -99,7 +104,8 @@ Let's look at a more complicated example
 OpenTelemetrySdk openTelemetry = OpenTelemetrySdk.builder()
   .setResource(Resource.getDefault().merge(CoolResource.getDefault()))
   .setClock(AtomicClock.create())
-  .setPropagators(B3Propagator.getInstance())
+  .addPropatator(B3Propagator.getInstance())
+  .addPropagator(TracingVendorPropagator.getInstance())
   .addSpanExporter(OtlpGrpcSpanExporter.builder()
     .setEndpoint("spans-service:4317")
     .setTimeout(Duration.ofSeconds(10))
@@ -119,7 +125,7 @@ OpenTelemetrySdk openTelemetry = OpenTelemetrySdk.builder()
   .build()
 ```
 
-This configures resource, clock, exporters, span processors, sampler, trace limits, and metrics.
+This configures resource, clock, exporters, span processors, propagators, sampler, trace limits, and metrics.
 Everything is exposed on the `OpenTelemetrySdkBuilder` to provide discoverability of all the parts
 of the SDK and to let the SDK manage shared components - it is not possible to configure a different
 `Resource` for tracing and metrics, something that would be a gotcha. Exporters are handled natively
@@ -251,7 +257,7 @@ public class MyService {
 ## The global instance of the SDK
 
 A built instance is convenient to use in most Java apps because of dependency injection. Because it
-has a easy-to-reason lifecycle, being tied into the dependency ordering (even if dependency 
+has a easy-to-reason initialization ordering, being tied into the dependency ordering (even if dependency 
 injection happened to be done manually through constructor invocation), we encourage application
 developers to only use it.
 
@@ -270,7 +276,7 @@ should not have any effect one way or the other in the vast majority of cases.
 MySQL is the only known corner case that requires the global SDK instance. If such a corner case
 didn't exist, we may not even support it in the first place.
 
-See special note with Java Agent below though.
+See special note about Java Agent below though.
 
 ## Telemetry within SDK components
 
@@ -296,12 +302,15 @@ public class BatchExporter implements SpanExporter {
   
   @Override
   public void setOpenTelemetry(OpenTelemetry openTelemetry) {
-    traer = openTelemetry.getTracerProvider().get("spanexporter");
+    tracer = openTelemetry.getTracerProvider().get("spanexporter");
   }
   
   @Override
   public void export() {
-    tracer.spanBuilder("export").startSpan();
+    Tracer tracer = this.tracer;
+    if (tracer != null) {
+      tracer.spanBuilder("export").startSpan();
+    }
   }
 }
 public class OpenTelemetrySdkBuilder {
@@ -396,7 +405,7 @@ class SleuthUsingService {
 
 ## Auto-configuration
 
-The above presents programmatic configuration for the SDK and proposeds that the core SDK has no
+The above presents programmatic configuration for the SDK and proposes that the core SDK has no
 other mechanism for configuration. There is no SPI nor processing of environment variables or
 system properties. There are many mechanisms for configuration, for example Spring Boot. 
 Integration with these systems becomes easier to reason about if we consider auto-configuration at
--- a/docs/sdk-configuration.md
+++ b/docs/sdk-configuration.md
@@ -131,7 +131,7 @@ of the SDK and to let the SDK manage shared components - it is not possible to c
 `Resource` for tracing and metrics, something that would be a gotcha. Exporters are handled natively
 given they are so important - even though the exporters are specified before span processors, they
 will always be executed last to make sure all span processors appy to them. The exporter builders
-inherit from a `ExporterBuilder` type of abstract class - it takes cafe of batching configuration to
+inherit from a `ExporterBuilder` type of abstract class - it takes care of batching configuration to
 avoid having to understand separate concepts like `BatchSpanProcessor`. Exporters that don't support
 batching at all, like the custom `FullyAsyncZipkinExporter` don't have to. Note this can be
 similarly achieved by having wrappers like `BatchExporter.wrap(OtlpGrpcSpanExporter.builder()...`
@@ -196,7 +196,7 @@ public class OpenTelemetryModule {
   }
   
   @Bean
-  public MeterProvider tracerProvider(OpenTelemetry openTelemetry) {
+  public MeterProvider meterProvider(OpenTelemetry openTelemetry) {
     return openTelemetry.getMeterProvider();
   }
   
--- a/docs/sdk-configuration.md
+++ b/docs/sdk-configuration.md
@@ -1,0 +1,492 @@
+# Design for configuring the SDK
+
+This document outlines some of the goals we have for user configuration of the SDK. It is a
+continuation of discussion started with https://github.com/open-telemetry/opentelemetry-java/issues/2022.
+
+## Target audiences
+
+There are a few different target audiences that are related to our configuration story.
+
+- Application developers, aka end-users. Often have no knowledge of tracing but want to add 
+OpenTelemetry to their app and see traces show up in a console. Application developers will increase
+in number forever, while the below are more constant.
+
+- Dev-ops / framework developers. Write components or frameworks to support providing tracing to
+their application developers. May write custom SDK extensions such as exporters, samplers, to fit
+with their internal infrastructure and as such have some familiarity with tracing at least as the
+SDK presents it.
+
+- OpenTelemetry maintainers. Write the SDK code.
+
+When making decisions, especially about complexity, we always prioritize the application developers,
+then framework developers, then maintainers. This is because we expect those with less domain
+knowledge about tracing to require a simpler experience than those with more. There is also more
+bang-for-the-buck by making the end-user experience as streamlined as possible since we expect there
+to be much more of them than other audiences.
+
+## Goals and non-goals
+
+### Goals
+
+- Provide a single entrypoint to configuring the SDK. For end-users, less familiar with the SDK, we
+want to have everything together to provide discoverability and simpler end-user code.
+
+- Fit well with common Java idioms such as dependency injection, or common frameworks like Spring.
+
+- Reduce the chance of gotchas or configuration mishaps.
+
+- Aim for optimal performance for the most common use case.
+
+### Non-goals
+
+- Provide the best possible experience for custom SDKs. Generally the burden of the experience for
+custom SDKs can fall on their authors, we optimize for our standard usage, the full SDK. Any
+reference to "the SDK" in this document refers to the full SDK with all signals.
+
+- Make sure everything is auto-configurable. This is out of the scope of the SDK, and instead is
+left to auto-configuration layers, which are also described below but not as part of the core SDK. 
+In partiular, SignalProvider SPIs which currently exist are proposed to be removed from the SDK. The 
+SDK may provide an autoconfiguration extension as an option which is not internal to the main SDK 
+components.
+
+## Configuring an instance of the SDK
+
+The SDK consists of configuration options for all the signals it supports. Users all have different
+requirements for how they use the SDK, for example they may use different exporters depending on
+their backend. Because we cannot guess the configuration the user needs, we expect that the SDK must
+be configured by the user before it can be used.
+
+Goals for configuring the SDK are
+
+- Discoverability of options
+- Ease of use by end users, e.g., less complicated code required
+- Avoid requiring duplicate configuration, which can lead to errors or confusion
+- Provide good defaults where possible
+
+In Java, the builder pattern is common for configuring instances. Let's look at what that may look
+like. The simplest configuration will be when a user wants to get a default experience, exporting
+with a specific exporter to an endpoint.
+
+```java
+OpenTelemetrySdk openTelemetry = OpenTelemetrySdk.builder()
+  .addSpanExporter(OtlpGrpcSpanExporter.builder()
+    .setEndpoint("spans-service:4317")
+    .build())
+  .addMetricExposition(PrometheusExposition.getDefault())
+  .build();
+```
+
+This
+
+- Exports spans using OTLP to spans-service
+  - Uses the BatchSpanProcessor (or a batching exporter wraper) which we consider the best default 
+  for exporting. It has defaults for queue size, timeout, etc
+- Exposes metrics using Prometheus format
+- Uses ParentBased(AlwaysOn) sampler
+- Uses standard random IDs
+- Uses a default Resource which simply consists of the SDK (or any other resources we decide to)
+include in the core SDK, not extensions
+- Uses the default Clock, which uses Java 8 / 9+ optimized APIs for getting time
+  - The only real reason a user would set this is for unit tests, not for production
+- Enforces default numeric limits related to number of attribute, etc
+
+Because the exporting is often the only aspect an end user needs to configure, this is the simplest
+possible API for configuring the SDK.
+
+Let's look at a more complicated example
+
+```java
+OpenTelemetrySdk openTelemetry = OpenTelemetrySdk.builder()
+  .setResource(Resource.getDefault().merge(CoolResource.getDefault()))
+  .setClock(AtomicClock.create())
+  .setPropagators(B3Propagator.getInstance())
+  .addSpanExporter(OtlpGrpcSpanExporter.builder()
+    .setEndpoint("spans-service:4317")
+    .setTimeout(Duration.ofSeconds(10))
+    .setBatchQueueSize(10000)
+    .build())
+  .addSpanExporter(ZipkinExporter.builder()
+    .setEndpoint("spans-service:4317")
+    .disableBatching()
+    .build())
+  .addSpanExporter(FullyAsyncZipkinExporter.create())
+  .addSpanProcessor(CustomAttributeAddingProcessor.create())
+  .addSpanProcessor(CustomEventAddingProcessor.create())
+  .setTraceSampler(Sampler.rateLimiting())
+  .setTraceLimits(TraceLimits.builder().setMaxAttributes(10).build())
+  .setIdGenerator(TimestampedIdGenerator.create())
+  .setMeterStuff(InProgressMetricsStuff.create())
+  .build()
+```
+
+This configures resource, clock, exporters, span processors, sampler, trace limits, and metrics.
+Everything is exposed on the `OpenTelemetrySdkBuilder` to provide discoverability of all the parts
+of the SDK and to let the SDK manage shared components - it is not possible to configure a different
+`Resource` for tracing and metrics, something that would be a gotcha. Exporters are handled natively
+given they are so important - even though the exporters are specified before span processors, they
+will always be executed last to make sure all span processors appy to them. The exporter builders
+inherit from a `ExporterBuilder` type of abstract class - it takes cafe of batching configuration to
+avoid having to understand separate concepts like `BatchSpanProcessor`. Exporters that don't support
+batching at all, like the custom `FullyAsyncZipkinExporter` don't have to. Note this can be
+similarly achieved by having wrappers like `BatchExporter.wrap(OtlpGrpcSpanExporter.builder()...`
+as well and is presented as one option.
+
+Note that parts can be separated to separate subconfig classes if it seems to make the API easier to
+use. We would want to avoid duplication though, e.g., having two configurations that expect a
+`Resource`.
+
+One thing to keep in mind is that it will be less common for an application developer to go this far
+and we can expect it is actually framework developers that use the full configuration capabilities
+of the SDK, likely by tying it to a separate configuration system.
+
+### Why build instances
+
+We have found that even at such an early stage of user adoption, users want to build instances of
+the SDK.
+
+- Integrates with dependency injection, e.g., Spring, in a similar way as many other libraries
+- Can allow having multiple instances in the same app, in multi-concern single-classloader scenarios
+- Allow managing lifecycle of SDK, e.g., shutting down and starting along with the lifecycle of a
+serverless runtime
+
+### Configuring the SDK in a framework
+
+It is extremely common for Java apps to be written using a dependency injection framework like
+Spring, Guice, Dagger, HK, and many more. They will all follow a very similar pattern though.
+
+```java
+@Component
+public class OpenTelemetryModule {
+  
+  @Bean
+  public OpenTelemetry openTelemetry(MonitoringConfig config) {
+    OpenTelemetrySdk.builder()
+      .setResource(Resource.getDefault().merge(CoolResource.getDefault()))
+      .setClock(AtomicClock.create())
+      .setPropagators(B3Propagator.getInstance())
+      .addSpanExporter(OtlpGrpcSpanExporter.builder()
+        .setEndpoint(config.getOtlpExporter().getEndpoint())
+        .setTimeout(config.getOtlpExporter().getTimeout())
+        .setBatchQueueSize(config.getOtlpExporter().getQueueSize())
+        .build())
+      .addSpanExporter(ZipkinExporter.builder()
+        .setEndpoint(config.getZipkinExporter().getEndpoint())
+        .disableBatching()
+        .build())
+      .addSpanExporter(FullyAsyncZipkinExporter.create())
+      .addSpanProcessor(CustomAttributeAddingProcessor.create())
+      .addSpanProcessor(CustomEventAddingProcessor.create())
+      .setTraceSampler(config.getSamplingRate() != 0 
+          ? Sampler.rateLimiting(config.getSamplingRate()) : Sampler.getDefault())
+      .setTraceLimits(TraceLimits.builder().setMaxAttributes(config.getMaxSpanAttributes()).build())
+      .setIdGenerator(TimestampedIdGenerator.create())
+      .setMeterStuff(InProgressMetricsStuff.create())
+      .build();
+  }
+  
+  @Bean
+  public TracerProvider tracerProvider(OpenTelemetry openTelemetry) {
+    return openTelemetry.getTracerProvider();
+  }
+  
+  @Bean
+  public MeterProvider tracerProvider(OpenTelemetry openTelemetry) {
+    return openTelemetry.getMeterProvider();
+  }
+  
+  @Bean
+  public AuthServiceStub authService(OpenTelemetry openTelemetry, AuthConfig config) {
+    return AuthServiceGrpc.newBlockingStub(ManagedChannelBuilder.forEndpoint(config.getEndpoint()))
+      .withInterceptor(TracingClientInterceptor.create(openTelemetry));
+  }
+  
+  @Bean
+  public ServletFilter servletFilter(OpenTelemetry openTelemetry) {
+    return TracingServletFilter.create(openTelemetry);
+  }
+}
+
+// Use some instrumented client
+@Component
+public class MyAuthInterceptor {
+
+  private final AuthServiceStub authService;
+  
+  @Inject
+  public MyAuthInterceptor(AuthServiceStub authService) {
+    this.authService = authService;
+  }
+  
+  public void doAuth() {
+    if (!authService.getToken("credential").isAuthenticated()) {
+      throw new HackerException();
+    }
+  }
+}
+
+// Use tracer directly, not so common
+@Component
+public class MyService {
+
+  private final Tracer tracer;
+  private final Meter meter;
+  
+  @Inject
+  public MyService(TracerProvider tracerProvider, MeterProvider meterProvider) {
+    tracer = tracerProvider.get("my-service");
+    meter = meterProvider.get("my-service");
+  }
+  
+  public void doLogic() {
+    Span span = tracer.spanBuilder("logic").startSpan();
+    try (Scope ignored = span.makeCurrent()) {
+      Thread.sleep(1000);
+    } finally {
+      span.end();
+    }
+  }
+}
+```
+
+## The global instance of the SDK
+
+A built instance is convenient to use in most Java apps because of dependency injection. Because it
+has a easy-to-reason lifecycle, being tied into the dependency ordering (even if dependency 
+injection happened to be done manually through constructor invocation), we encourage application
+developers to only use it.
+
+However, there are corner cases where an instance cannot be injected. The famous example is MySQL -
+MySQL interceptors are initialized by calling a default constructor, and there is no way to pass a
+built instance of a signal provider. For this case, we must store an SDK instance into a global
+variable. To reduce gotchas for application developers, the first SDK that is instantiated is stored
+automatically in the global variable.
+
+Before an SDK has been set, access to the global SDK will throw an exception and will not return a
+default. The global can only be used after an SDK has been configured. SDKs must be configured early
+in an application to ensure it applies to all of the logic in the app, and this will generally be
+ensured by the configuration framework such as Spring. For application developers, this restriction
+should not have any effect one way or the other in the vast majority of cases.
+
+MySQL is the only known corner case that requires the global SDK instance. If such a corner case
+didn't exist, we may not even support it in the first place.
+
+See special note with Java Agent below though.
+
+## Telemetry within SDK components
+
+SDK components, such as exporters or remote samplers, may want to emit telemetry for their own
+processing. However, the SDK components must be initialized before the SDK can be fully built. We do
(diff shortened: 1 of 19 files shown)


docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+variable. To reduce gotchas for application developers, the first SDK that is instantiated is stored
+automatically in the global variable.
+
+Before an SDK has been set, access to the global SDK will throw an exception and will not return a
+default. The global can only be used after an SDK has been configured. SDKs must be configured early
+in an application to ensure it applies to all of the logic in the app, and this will generally be
+ensured by the configuration framework such as Spring. For application developers, this restriction
+should not have any effect one way or the other in the vast majority of cases.
+
+MySQL is the only known corner case that requires the global SDK instance. If such a corner case
```
anuraaga, line 270: At least in Brave

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+  .build()
+```
+
+This configures resource, clock, exporters, span processors, sampler, trace limits, and metrics.
+Everything is exposed on the `OpenTelemetrySdkBuilder` to provide discoverability of all the parts
+of the SDK and to let the SDK manage shared components - it is not possible to configure a different
+`Resource` for tracing and metrics, something that would be a gotcha. Exporters are handled natively
+given they are so important - even though the exporters are specified before span processors, they
+will always be executed last to make sure all span processors appy to them. The exporter builders
+inherit from a `ExporterBuilder` type of abstract class - it takes cafe of batching configuration to
```
bogdandrutu, line 128: s/cafe/care?

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+  .setPropagators(B3Propagator.getInstance())
+  .addSpanExporter(OtlpGrpcSpanExporter.builder()
+    .setEndpoint("spans-service:4317")
+    .setTimeout(Duration.ofSeconds(10))
+    .setBatchQueueSize(10000)
+    .build())
+  .addSpanExporter(ZipkinExporter.builder()
+    .setEndpoint("spans-service:4317")
+    .disableBatching()
+    .build())
```
bogdandrutu, line 111: How can I set https://github.com/open-telemetry/opentelemetry-java/tree/master/sdk-extensions/async-processor to be used instead of the normal batch span processor?

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+      .build();
+  }
+  
+  @Bean
+  public TracerProvider tracerProvider(OpenTelemetry openTelemetry) {
+    return openTelemetry.getTracerProvider();
+  }
+  
+  @Bean
+  public MeterProvider tracerProvider(OpenTelemetry openTelemetry) {
```
bogdandrutu, line 193: s/tracerProvider/meterProvider?

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+
+// Use tracer directly, not so common
+@Component
+public class MyService {
+
+  private final Tracer tracer;
+  private final Meter meter;
+  
+  @Inject
+  public MyService(TracerProvider tracerProvider, MeterProvider meterProvider) {
```
bogdandrutu, line 235: I would encourage passing Otel instance always. So if we add more functionality to the otel instance we don't break the API.

docs/sdk-configuration.md
bogdandrutu: So there is no world outside Spring? :)

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+A built instance is convenient to use in most Java apps because of dependency injection. Because it
+has a easy-to-reason lifecycle, being tied into the dependency ordering (even if dependency 
+injection happened to be done manually through constructor invocation), we encourage application
+developers to only use it.
+
+However, there are corner cases where an instance cannot be injected. The famous example is MySQL -
+MySQL interceptors are initialized by calling a default constructor, and there is no way to pass a
+built instance of a signal provider. For this case, we must store an SDK instance into a global
+variable. To reduce gotchas for application developers, the first SDK that is instantiated is stored
+automatically in the global variable.
```
bogdandrutu, line 262: I think this is a very unexpected behavior....

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+    tracerProvider.setSampler(sampler);
+    components.add(sampler);
+  }
+
+  public OpenTelemetry build() {
+    OpenTelemetrySdk sdk = new OpenTelemetrySdk(tracerProvider.build(), meterProvider.build());
+    for (OpenTelemetryComponent component : components) {
+      component.setOpenTelemetry(sdk);
+    }
+  }
```
bogdandrutu, line 323: What if someone is very strict and want to actually use a different sdk instance + pipeline to monitor the main monitoring sdk?

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+}
+```
+
+A framework author will have an even easier time since most dependency injection frameworks
+natively support lazy injection.
```
bogdandrutu, line 328: So if I am not using Spring I am not a "framework" author? :)

I would rewrite with something like:
When using a framework that supports dependency injection and lazy initialization ....

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+within telemetry components to those components, where the maintainers will have more domain
+knowledge. It allows this mutator method to be removed from the end-user API.
+
+- `updateTraceConfig` - instead of allowing updates at the top level, we should consider making
+`TraceConfig` an interface, and the SDK default implementation is static. It allows the above
+benefits of immutability to be in place for the common case where dynamic updates are not needed.
+Where dynamic updates are needed, it can be replaced with a mutable implementation instead of making
+the SDK configuration mutable. This keeps update methods out of the end-user APIs and will generally
+give framework developers more control by handling dynamicism themselves without the chance of
+end-users to affect it negatively.
```
bogdandrutu, line 375: If TraceConfig is not "immutable" you have hard time and undefined behaviors. What does it happen if we decide to call sampler twice and gets changed between the calls? Or any other property that we use?

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+```java
+
+class SleuthUsingService {
+  
+  @Inject
+  private OpenTelemetry openTelemetry;
+
+  public void doLogic() {
+    // My logic is important, so always sample it!
+    OpenTelemetrySdk.getTracerManagement().updateTraceConfig(config -> config.setSampler(ALWAYS_ON));
```
bogdandrutu, line 388: Who is the author of this line? Only one person should configure or control the SDK config. Any random "service" running in an application service should be able to affect only their instance of SDK if any, otherwise should not be able to control the config.

docs/sdk-configuration.md
bogdandrutu: By making immutable we can guarantee that one config is applied for the entire lifetime of one span, and you cannot do this with a mutable config.

You may say that you add a "Factory" that returns an immutable TraceConfig, that is a lot of unnecessary allocations (you may cache the result). But if there is a factory why someone cannot do the same thing and update the Factory?

docs/sdk-configuration.md
anuraaga: It could also be reasonable to leave this up to frameworks - but it will almost always be required.

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+  for exporting. It has defaults for queue size, timeout, etc
+- Exposes metrics using Prometheus format
+- Uses ParentBased(AlwaysOn) sampler
+- Uses standard random IDs
+- Uses a default Resource which simply consists of the SDK (or any other resources we decide to)
+include in the core SDK, not extensions
+- Uses the default Clock, which uses Java 8 / 9+ optimized APIs for getting time
+  - The only real reason a user would set this is for unit tests, not for production
+- Enforces default numeric limits related to number of attribute, etc
+
```
iNikem, line 92: What about propagation?

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+```java
+OpenTelemetrySdk openTelemetry = OpenTelemetrySdk.builder()
+  .setResource(Resource.getDefault().merge(CoolResource.getDefault()))
+  .setClock(AtomicClock.create())
+  .setPropagators(B3Propagator.getInstance())
+  .addSpanExporter(OtlpGrpcSpanExporter.builder()
+    .setEndpoint("spans-service:4317")
+    .setTimeout(Duration.ofSeconds(10))
+    .setBatchQueueSize(10000)
```
iNikem, line 106: batching becomes the property of exporters?

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+    .build())
+  .addSpanExporter(ZipkinExporter.builder()
+    .setEndpoint("spans-service:4317")
+    .disableBatching()
+    .build())
+  .addSpanExporter(FullyAsyncZipkinExporter.create())
+  .addSpanProcessor(CustomAttributeAddingProcessor.create())
+  .addSpanProcessor(CustomEventAddingProcessor.create())
+  .setTraceSampler(Sampler.rateLimiting())
+  .setTraceLimits(TraceLimits.builder().setMaxAttributes(10).build())
```
iNikem, line 116: I like this

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+  .setMeterStuff(InProgressMetricsStuff.create())
+  .build()
+```
+
+This configures resource, clock, exporters, span processors, sampler, trace limits, and metrics.
+Everything is exposed on the `OpenTelemetrySdkBuilder` to provide discoverability of all the parts
+of the SDK and to let the SDK manage shared components - it is not possible to configure a different
+`Resource` for tracing and metrics, something that would be a gotcha. Exporters are handled natively
+given they are so important - even though the exporters are specified before span processors, they
+will always be executed last to make sure all span processors appy to them. The exporter builders
```
iNikem, line 127: How do you plan to chain `SpanProcessor.end()` calls?

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+
+- Application developers, aka end-users. Often have no knowledge of tracing but want to add 
+OpenTelemetry to their app and see traces show up in a console. Application developers will increase
+in number forever, while the below are more constant.
+
+- Dev-ops / framework developers. Write components or frameworks to support providing tracing to
+their application developers. May write custom SDK extensions such as exporters, samplers, to fit
+with their internal infrastructure and as such have some familiarity with tracing at least as the
+SDK presents it.
+
```
iNikem, line 18: I think we bundle two types here: authors of custom exporters/samplers and frameworks authors. The latter will handle configuration, but the former will just write a component according to an interface.

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+      span.end();
+    }
+  }
+}
+```
+
+## The global instance of the SDK
+
+A built instance is convenient to use in most Java apps because of dependency injection. Because it
+has a easy-to-reason lifecycle, being tied into the dependency ordering (even if dependency
```
iNikem, line 254: I agree with this section, but not with "easy-to-reason lifecycle". Shared instance kept in a field will break if somebody somewhere calls `shutdown` on the same shared instance.

docs/sdk-configuration.md
iNikem: I agree with Bogdan. Let it be explicit.

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+Before an SDK has been set, access to the global SDK will throw an exception and will not return a
+default. The global can only be used after an SDK has been configured. SDKs must be configured early
+in an application to ensure it applies to all of the logic in the app, and this will generally be
+ensured by the configuration framework such as Spring. For application developers, this restriction
+should not have any effect one way or the other in the vast majority of cases.
+
+MySQL is the only known corner case that requires the global SDK instance. If such a corner case
+didn't exist, we may not even support it in the first place.
+
+See special note with Java Agent below though.
```
iNikem, line 273:
```suggestion
See special note about Java Agent below though.
```

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+}
+interface SpanExporter extends OpenTelemetryComponent {
+}
+public class BatchExporter implements SpanExporter {
+  
+  private volatile Tracer tracer;
+  
+  @Override
+  public void setOpenTelemetry(OpenTelemetry openTelemetry) {
+    traer = openTelemetry.getTracerProvider().get("spanexporter");
```
iNikem, line 299:
```suggestion
    tracer = openTelemetry.getTracerProvider().get("spanexporter");
```

docs/sdk-configuration.md
iNikem: How are we going to support dynamic Samplers?

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+    // "manage the SDK". Unlike the javaagent, it can't block access to SDK methods we may provide.
+    doSampledLogicWhileOtherServicesAlsoGetSampled();
+  }
+}
+```
+
+
+## Auto-configuration
+
+The above presents programmatic configuration for the SDK and proposeds that the core SDK has no
```
iNikem, line 399:
```suggestion
The above presents programmatic configuration for the SDK and proposes that the core SDK has no
```

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+other mechanism for configuration. There is no SPI nor processing of environment variables or
```
iNikem, line 400: Wait :) I understand your purpose and mostly support it, but configuration from environment is useful. If core SDK does not support (which is fine), we may need an example of how one can achieve it.

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+or internal frameworks developed by devops teams at companies) is also a mechanism for automatically
+configuring the SDK. In general, we would expect Sleuth users to not be using the java agent.
+
+Examples of how Sleuth could work are presented above in examples using `@Bean`. In particular, we
+expect it to have its own set of configuration properties - by making sure we don't implement
+configuration properties in the core SDK, only configuration layers like the agent or a possible
+configuration wrapper, we avoid the possibility of confusion by having duplicate variables (in
+practice, OpenTelemetry naming would likely be ignored and overwritten by Spring naming).
+
+## Partial SDKs
```
iNikem, line 439: I got very confused by this section. It seems like configuration duplication. What if I by mistake do both `OpenTelemetrySdk.builder().xxxxx` as above and `DefaultOpenTelemetry.builder().xxxx` as below?

iNikem (commented):
I mostly support this proposal. Three open questions:

- Should java agent set Global SDK?
- Confusing part about partial SDKs
- Dynamic Samplers

docs/sdk-configuration.md
anuraaga: Thanks for reminder :)

docs/sdk-configuration.md
anuraaga: It's a random idea - more that batching configuration is the job of exporter configuration. One goal was to make it more obvious when an exporter is tied to a batching (or not) strategy.

But as with the comment about async processor this might be too limiting so not liking it so much anymore but still want to talk it through a bit.

docs/sdk-configuration.md
anuraaga: Oops forgot to reply - indeed this may make the fluent model too limiting. But ideally we can at least find a way to share timeout settings between processor and exporter, major pain right now.

docs/sdk-configuration.md
anuraaga: This proposal doesn't address the issue of not being able to have branching `SpanProcessor`. It just calls the processors in order, but trying to help with the exporter gotcha by making sure it comes at the end.

iNikem: This seems to be as violating spec, et least formally

docs/sdk-configuration.md
anuraaga: Sampler is already an interface I was always expecting them to be a dynamic implementation of it.

docs/sdk-configuration.md
anuraaga: Agree it is a decision on whether we prioritize the "wrapper artifact" I mentioned below. But do want to move it to a different layer.

docs/sdk-configuration.md
anuraaga: `OpenTelwetrySdk` is only present in the full SDK artifact (`-all`). So I guess it won't be there with a partial SDK?

Or is it potential confusion for full SDK users having both entry points? I don't think our repo should have any documentation related to `DefaultOpenTelemetry`. Maybe even hide it from JavaDoc. Full SDK users use `OpenTelemetrySdk` that's it, and it's all we support. Supporting partial SDK is up to the authors of partial SDK components IMO.

docs/sdk-configuration.md
iNikem: All current Samplers are immutable and are changed by swapping instances.

docs/sdk-configuration.md
iNikem: The second part, yes. So you are talking here about some other (partial) implementation of the SDK, not this one?

anuraaga: Not sure what you're referring to with "this one" - my hypothetical micrometer SDK?

So I am imagining an app with dependencies on `opentelemetry-sdk-trace` and `opentelemetry-micrometer` and use `DefaultOpenTelemetry`.

Or there might be a `opentelemetry-sdk-with-micrometer` which depends on the above and provides `OpenTelemetrySdkWithMicrometerBuilder`. Obtuse but just an example of what I was thinking thinking here.

iNikem: "this one" is "this repo". I mean, we provide full SDK and we don't plan to provide partial SDKs, right? Partial SDKs come from outside, right?

anuraaga: Ah yeah that's what I'm thinking

jkwatson:
ref #1909 which I had forgotten existed.

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+
+- Dev-ops / framework developers. Write components or frameworks to support providing tracing to
+their application developers. May write custom SDK extensions such as exporters, samplers, to fit
+with their internal infrastructure and as such have some familiarity with tracing at least as the
+SDK presents it.
+
+- OpenTelemetry maintainers. Write the SDK code.
+
+When making decisions, especially about complexity, we always prioritize the application developers,
+then framework developers, then maintainers. This is because we expect those with less domain
```
jkwatson, line 22: this is an excellent call-out, and something we all definitely need to keep in mind. Sometimes this will mean we add "more than one way to do things" because it will make app developers lives easier, even if it makes our jobs as maintainers more annoying/painful.

docs/sdk-configuration.md
jkwatson: the framework authors might need to do both (see Sleuth), but your point is still a good one.

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+### Goals
+
+- Provide a single entrypoint to configuring the SDK. For end-users, less familiar with the SDK, we
+want to have everything together to provide discoverability and simpler end-user code.
+
+- Fit well with common Java idioms such as dependency injection, or common frameworks like Spring.
+
+- Reduce the chance of gotchas or configuration mishaps.
+
+- Aim for optimal performance for the most common use case.
```
jkwatson, line 38: is performance at configuration time really much of an issue and something we should worry about?

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+knowledge about tracing to require a simpler experience than those with more. There is also more
+bang-for-the-buck by making the end-user experience as streamlined as possible since we expect there
+to be much more of them than other audiences.
+
+## Goals and non-goals
+
+### Goals
+
+- Provide a single entrypoint to configuring the SDK. For end-users, less familiar with the SDK, we
+want to have everything together to provide discoverability and simpler end-user code.
```
jkwatson, line 32: is this a "single entrypoint" per use-case, or only a single entrypoint, period?

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+
+- Make sure everything is auto-configurable. This is out of the scope of the SDK, and instead is
+left to auto-configuration layers, which are also described below but not as part of the core SDK. 
+In partiular, SignalProvider SPIs which currently exist are proposed to be removed from the SDK. The 
+SDK may provide an autoconfiguration extension as an option which is not internal to the main SDK 
+components.
+
+## Configuring an instance of the SDK
+
+The SDK consists of configuration options for all the signals it supports. Users all have different
```
jkwatson, line 54:
```suggestion
The SDK exposes configuration options for all the signals it supports. Users all have different
```

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+requirements for how they use the SDK, for example they may use different exporters depending on
```
jkwatson, line 55:
```suggestion
requirements for how they use the SDK; for example they may use different exporters depending on
```

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+- Avoid requiring duplicate configuration, which can lead to errors or confusion
+- Provide good defaults where possible
+
+In Java, the builder pattern is common for configuring instances. Let's look at what that may look
+like. The simplest configuration will be when a user wants to get a default experience, exporting
+with a specific exporter to an endpoint.
+
+```java
+OpenTelemetrySdk openTelemetry = OpenTelemetrySdk.builder()
+  .addSpanExporter(OtlpGrpcSpanExporter.builder()
```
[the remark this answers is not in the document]
jkwatson: My opinion: no. But, I'm certainly up to being convinced otherwise. 

My reasoning is: people are going to try out the SDK on their local laptops first (probably), and if the default is to point at Otlp, they'll just get a bunch of errors in their logs.

I could be convinced that the default could be the logging exporter...that seems like an option that wouldn't break anybody and make it clear what's happening.

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+    .setEndpoint("spans-service:4317")
+    .build())
+  .addMetricExposition(PrometheusExposition.getDefault())
```
jkwatson, line 75: This example mixes two signals, which is good. The question is: are we likely to end up with an unwieldy builder if it contains the configuration for multiple signals, especially when we (eventually) add logs in here as well.

Would we be better off (as I think Bogdan has suggested) requiring the SDK components for each signal to be configured indivually, then have the `OpenTelemetrySdk` builder only compose the other components? I can see pros and cons.

docs/sdk-configuration.md
jkwatson: I think there's been some thought that maybe we should swap the responsibility and have the exporter be created with a processor (if the specific exporter allows for more than one option), rather than the other way around. That would definitely be a change to our processor/exporter APIs, but maybe now is a good time to think about it.

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+        .build())
+      .addSpanExporter(FullyAsyncZipkinExporter.create())
+      .addSpanProcessor(CustomAttributeAddingProcessor.create())
+      .addSpanProcessor(CustomEventAddingProcessor.create())
+      .setTraceSampler(config.getSamplingRate() != 0 
+          ? Sampler.rateLimiting(config.getSamplingRate()) : Sampler.getDefault())
+      .setTraceLimits(TraceLimits.builder().setMaxAttributes(config.getMaxSpanAttributes()).build())
+      .setIdGenerator(TimestampedIdGenerator.create())
+      .setMeterStuff(InProgressMetricsStuff.create())
+      .build();
```
jkwatson, line 184: Here is a sticking point that I see with the omnibus build-everything-at-the-OpenTelemetrySdk-level: For 1.0, we need to remove metrics from the OpenTelemetrySdk, which will mean that we *must* provide a way to build the pre-release MeterSdkProvider in a stand-alone manner. And, there may (probably will) be people who only want to use one of the signals, without the omnibus OpenTelemetrySdk, so we'll need to provide individual builders for each of the signals.

If we need to provide individual builders for each SDK signal, then are we ok with having both ways of building things? If so, then 👍 . This does seem to be a case where we should favor the end-user and make sure that they can do it either way.

docs/sdk-configuration.md
jkwatson: By "Let it be explicit" you mean "require the SDK user to explicitly assign the global"? Does this mean that the operator/app developer will need to understand the internals of all the instrumentation that happens to be included with the project?

I'm not necessarily arguing in favor of auto-setting the global instance, but I wanted to bring up a problem with the alternative.

My opinion would probably be: get rid of the global altogether and make instrumentation authors work around it in some way [I'm definitely not an expert here...just trying to optimize for simplicity as much as possible].

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+  private volatile Tracer tracer;
+  
+  @Override
+  public void setOpenTelemetry(OpenTelemetry openTelemetry) {
+    traer = openTelemetry.getTracerProvider().get("spanexporter");
+  }
+  
+  @Override
+  public void export() {
+    tracer.spanBuilder("export").startSpan();
```
jkwatson, line 304: obviously would need a null check here for the tracer, but that's a nit, obviously

docs/sdk-configuration.md
jkwatson: sure, but that doesn't mean that other Samplers could be more dynamic (or we could even update our existing Samplers to be dynamic without altering the interface, I think).

docs/sdk-configuration.md
jkwatson: It seems like this might be an argument against exposing a global where anyone could muck about with it and effect everything running in the current classloader (or JVM itself depending on where the global is class loaded first).

docs/sdk-configuration.md
jkwatson: not true, though, since we'll be pulling metrics out of the "main" SDK and only providing that piece as an "experimental" signal until it's ready to go. We'll have a "partial" SDK for metrics, at least in some sense.

docs/sdk-configuration.md
```diff
@@ -0,0 +1,492 @@
...
+reasonable. We can also hope that where it is important, it is the author of partial SDKs that
+provide a one-stop-shop entrypoint.
+
+```java
+@Bean
+public OpenTelemetry openTelemetry() {
+  return OpenTelemetrySdkWithMicrometer.builder()
+    .addSpanExporter()
+    .setMeterRegistry()
+    .build();
```
jkwatson, line 470: this is definitely something I hadn't pondered. This wouldn't work with pre-release signals, though, unless the pre-release metrics SDK (for instance) also provided an extension to the OpenTelemetry interface, too. 🤔

jkwatson:
1) Thanks for writing this up!!!
2) I think we also should consider about the configuration story with the "API only" users, where they only need to decide on propagators, and maybe some custom context backing store. Doesn't the question of global behavior of the SDK also hinge somewhat on the existence and use-cases for the global API in the absence of an SDK? In some sense, the API provides an SDK implementation, albeit a super minimal mostly no-op one. I feel like we should have the same story for configuring *that* minimal SDK that we do for a full-featured SDK.

jkwatson:
> 1. Thanks for writing this up!!!
> 2. I think we also should consider the configuration story with the "API only" users, where they only need to decide on propagators, and maybe some custom context backing store. Doesn't the question of global behavior of the SDK also hinge somewhat on the existence and use-cases for the global API in the absence of an SDK? In some sense, the API provides an SDK implementation, albeit a super minimal mostly no-op one. I feel like we should have the same story for configuring _that_ minimal SDK that we do for a full-featured SDK.

docs/sdk-configuration.md
iNikem: If we need to provide a way to configure and use just some of the signals, then I think that we should go this way:
> requiring the SDK components for each signal to be configured indivually, then have the OpenTelemetrySdk builder only compose the other components

The problem with supporting mix and match is this piece of code: https://github.com/open-telemetry/opentelemetry-java/pull/2231#issuecomment-741903911 .

anuraaga:
I have rewritten the builder to accept signals instead of signal knobs - while the duplication of `Resource` pains me, I think it could be better than the alternative. I'm not 100% convinced though as that duplication did hurt to write it ;) I have fleshed out the auto-configuration wrapper, which I hope can satisfy the itch for the magic global for end users that like that. We would expect it not to be used alongside Spring, or the javaagent.

I have also removed some of the proposals around reducing the complexity of registering exporters - I think it's somewhat orthogonal to the "overall configuration story" and will require quite a bit of thought anyways.

docs/sdk-configuration.md
```diff
@@ -0,0 +1,608 @@
...
+
+In Java, the builder pattern is common for configuring instances. Let's look at what that may look
+like. The simplest configuration will be when a user wants to get a default experience, exporting
+with a specific exporter to an endpoint.
+
+The SDK builder will simply allow accept its components as builder parameters. It only allows
+setting SDK implementations and is not meant for use with partial SDKs.
+
+```java
+class OpenTelemetrySdkBuilder {
```
anuraaga, line 79: Realized there's probably no point for this class.

jkwatson:
I think this is really solid. Thanks again for the work.  

I have only one thing I think we need to include and that is calling out how the "API only" case ends up working. We can't let it throw exceptions before the global has been set because library instrumentation might be using a global (against recommendations, but we can't stop it), and we can't let a library dependency crash an application. The user may not even know that the library they are using has been instrumented with OTel, and so we can't require the app dev to configure the API global.

jkwatson:
Oh...one more piece I'm not sure about...are we going to end up auto-setting the global the first time a full SDK has been built or no?

docs/sdk-configuration.md
```diff
@@ -0,0 +1,676 @@
...
+  public OpenTelemetrySdkBuilder setTracerProvider(TracerSdkProvider tracerProvider);
+  public OpenTelemetrySdkBuilder setMeterProvider(MeterSdkProvider meterProvider);
+  public OpenTelemetrySdk buildAndSetAsGlobal();
```
anuraaga, line 82: @jkwatson I had removed the auto-set as global due to controversy, but do think it's the best option in most cases. If the problem is mystery and surprise, then this naming, while obtuse, can make it clear. I would not have a normal `build()` method as that would reduce the discoverability of this mechanism - it has to be `buildWithoutSetting` or something similar to make sure a user makes a conscious decision without just assuming to use `build()`.

docs/sdk-configuration.md
```diff
@@ -0,0 +1,676 @@
...
+    }
+
+    @Bean
+    OpenTelemetrySdk openTelemetry(TracerSdkProvider tracerProvider) {
+        return OpenTelemetrySdk.builder().setTracerProvider(tracerProvider).build();
+    }
+}
+```
+
+The `OpenTelemetrySdk` and `TracerSdkProvider` are still exposed to provide access to `forceFlush`.
```
anuraaga, line 557: @jkwatson I realized it could make sense to remove `shutdown` from the `TracerSdkProvider`. It seems like it could make some lifecycle woes simpler to reason about, maybe even enable `toBuilder()`?

docs/sdk-configuration.md
```diff
@@ -0,0 +1,676 @@
...
+However, there are corner cases where an instance cannot be injected. The famous example is MySQL -
+MySQL interceptors are initialized by calling a default constructor, and there is no way to pass a
+built instance of a signal provider. For this case, we must store an SDK instance into a global
+variable. It is expected that frameworks or end-users will set the SDK as the global to support
+instrumentation that requires this.
+
+Before an SDK has been set, access to the global `OpenTelemetry` will return a no-op 
+`DefaultOpenTelemetry`. This is because it is possible library instrumentation is using the global,
+and it may even use it during the processing of a request (rather than only at initialization time).
+For this reason, we cannot throw an exception. Instead, if the SDK is detected on the classpath, we
```
anuraaga, line 363: @jkwatson API-only case

docs/sdk-configuration.md
```diff
@@ -0,0 +1,676 @@
...
+users to use `OpenTelemetrySdkBuilder` - while there is some duplication with the signal provider
+builder, it is work maintainers can do to present the simplest interface for the most common use
+case of using the whole SDK.
+
+Without SPI, the way to initialize a partial SDK would be to use `DefaultOpenTelemetry`.
+
+```java
+@Bean
+public OpenTelemetry openTelemetry() {
+  return DefaultOpenTelemetry.builder()
```
iNikem, line 637: So what is the essential difference between this case and the "full" `OpenTelemetrySdkBuilder` above?

anuraaga: The SDK builder returns types that expose forceFlush and shutdown as required by the spec. Hence me filing issues to try to get rid of those :-)

docs/sdk-configuration.md
```diff
@@ -0,0 +1,676 @@
...
+    }
+    
+    @Override
+    public void export() {
+    tracer.get().spanBuilder("export").startSpan();
+    }
+}
+```
+
+## Immutability of OpenTelemetry configuration
```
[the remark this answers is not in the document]
anuraaga: Please let me know what's confusing since I want to clarify it :)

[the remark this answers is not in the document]
iNikem: E.g. I don't understand the connection between `addSpanProcessor` method (on which class, btw?) and telemetry within telemetry (btw, is this the same as what you called "telemetry within SDK components" above?).

"the SDK default implementation is static" - do you mean "immutable"? And in general, sorry Anuraag, this section reads much sloppier than the preceding document :)

Also, I understand that Bogdan suggests having immutable SDK. As this does make total case on the first glance, may be this section better address those cases, when we potentially need mutability. E.g. "why obvious choice of immutable SDK is not that simple?"

docs/sdk-configuration.md
```diff
@@ -0,0 +1,676 @@
...
+    // My logic is important, so always sample it!
+    OpenTelemetrySdk.getTracerManagement().updateTraceConfig(config -> config.setSampler(ALWAYS_ON));
+    // This service was able to affect other services, even though Sleuth intends to
+    // "manage the SDK". Unlike the javaagent, it can't block access to SDK methods we may provide.
+    doSampledLogicWhileOtherServicesAlsoGetSampled();
+  }
+}
+```
+
+## TracerSdkManagement
```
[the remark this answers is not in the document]
iNikem: And now I understand why :) This section does not talk about SDK configuration at all :) But some completely separate concerns of public interfaces of the SDK. Should this section be in document at all?

docs/sdk-configuration.md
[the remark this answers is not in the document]
anuraaga: Thanks will try to rewrite it - one thing that's always tricky for me is talking about immutability. The SDK isn't immutable in the strictest sense since, e.g., `BatchSpanProcessor` has a mutable queue of spans. I guess the word to use is shallow-immutable which is what I meant by configuration being static. Let me see what I can do.

docs/sdk-configuration.md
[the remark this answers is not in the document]
anuraaga: This came to mind while considering what's left after removing the mutability methods from management interface, since ideally we don't have a management interface at all. But agree it's not really related with others, happy to have it just in commit history.

docs/sdk-configuration.md
```diff
@@ -0,0 +1,676 @@
...
+    }
+}
+```
+
+The `OpenTelemetrySdk` and `TracerSdkProvider` are still exposed to provide access to `forceFlush`.
+This makes it seem like `forceFlush`, though implemented in the SDK, may be more appropriate for the
+API as it's an end-user API - it could make the separation of concerns of the initialization of the
+SDK and the end-user tracing API complete. But the spec currently does not allow it.
+
+## Library instrumentation
```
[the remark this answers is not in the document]
anuraaga: Hmm - was I too nice to allow library authors an option of starting with global? I was imagining Bogdan's use case where some codebases may just want to make everything global, even in their instrumentation. But maybe it's too unclear in the context of this doc. Should I remove everything fro `An alternative method...`?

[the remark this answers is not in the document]
iNikem: I think so, yes.

docs/sdk-configuration.md
jkwatson: the only bit that might be confusing is that the 2nd one *wouldn't* set it as global, right?

docs/sdk-configuration.md
```diff
@@ -0,0 +1,676 @@
...
+           .setTracerProvider(tracerProvider)
+           .setMeterProvider(meterProvider)
+           .build();
+   }
+}
+```
+
+This configures resource, clock, exporters, span processors, propagators, sampler, trace limits, and metrics.
+It configures two SDKs with different propagators. We unfortunately cannot achieve our goal of only
+setting a property once - `Resource` and `Clock` are shared among signals but configured repeatedly.
```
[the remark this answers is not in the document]
jkwatson: @bogdandrutu has (I think) a use-case where sharing a `Resource` across MeterProviders will be prolematic. I'll let him explain.

docs/sdk-configuration.md
jkwatson: Thanks. This logging I think will be very helpful at finding where people are waiting until too late to set up their SDK. 👍

docs/sdk-configuration.md
anuraaga: Yeah - I think you're saying we need to document these methods well?

iNikem:
Are we good to merge this? Or what is the plan with this?

jkwatson:
> Are we good to merge this? Or what is the plan with this?

I'll give this one last once-over before approving!

docs/sdk-configuration.md
jkwatson: lol. yes, and maybe the 2nd call to `buildAndSetAsGlobal()` would fail? Not sure what the should be, but that can be decided at a later date.

docs/sdk-configuration.md
```diff
@@ -0,0 +1,612 @@
...
+
+The SDK builder will simply accept its components as builder parameters. It only allows
+setting SDK implementations and is not meant for use with partial SDKs.
+
+```java
+class OpenTelemetrySdkBuilder {
+  public OpenTelemetrySdkBuilder setTracerProvider(TracerSdkProvider tracerProvider);
+  public OpenTelemetrySdkBuilder setMeterProvider(MeterSdkProvider meterProvider);
+  public OpenTelemetrySdk buildAndSetAsGlobal();
+  public OpenTelemetrySdk buildWithoutSettingAsGlobal();
```
jkwatson, line 83: I might just call this one `build()`.  ;)

jkwatson (approved):
I think this is excellent. I suspect we'll evolve some bits of it as we go, but this seems like a fantastic starting point. Thanks again!

docs/sdk-configuration.md
anuraaga: See my note that I think this needs to be a conscious decision, though admittedly don't like the name much either :P

PR #2263, opened 2020-12-10
Some example of possible SDK configuration

iNikem:
To better see the effect of proposed changes

### DIFF
--- a/sdk/all/src/test/java/io/opentelemetry/sdk/OpenTelemetrySdkTest.java
+++ b/sdk/all/src/test/java/io/opentelemetry/sdk/OpenTelemetrySdkTest.java
@@ -196,35 +196,39 @@ class OpenTelemetrySdkTest {
         () -> OpenTelemetrySdk.builder().setTracerProvider(mock(TracerProvider.class)));
   }
 
-  //This is just a demonstration of maximum that one can do with OpenTelemetry configuration.
-  //Demonstrates how clear or confusing is SDK configuration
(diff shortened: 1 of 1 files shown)


sdk/all/src/test/java/io/opentelemetry/sdk/OpenTelemetrySdkTest.java
```diff
@@ -191,4 +195,74 @@ void onlySdkInstancesAllowed() {
...
+  // This is just a demonstration of two small but not trivial configurations.
+  // Demonstrates how clear or confusing is SDK configuration
+  @Test
+  void minimalOpenTelemetrySdkConfigurationDemo() {
+    OpenTelemetrySdk.builder()
+        .addSpanProcessor(SimpleSpanProcessor.builder(mock(SpanExporter.class)).build())
+        .setPropagators(ContextPropagators.create(mock(TextMapPropagator.class)))
+        .setTraceConfig(
+            TraceConfig.getDefault().toBuilder().setSampler(mock(Sampler.class)).build())
+        .build();
```
jkwatson, line 260: We could turn these all into tests by adding a simple `isNotNull` assert on the results. But, that's only if we feel like we need to have these tests do assertions. Just the fact that they don't throw exceptions is probably enough, TBH.

jkwatson (approved):
Thanks! I think these are useful cases to consider while we work on solidifying & simplifying the configuration APIs.

iNikem:
@jkwatson Do you want me to do anything more on this PR or can it be merged?

PR #2269, opened 2020-12-10
Introduce accumulator concept

bogdandrutu:
The code already existed as part of the instrument, but it was extracted in a separate concept,
and all Sync/Async instruments look very similar: Descriptor + Accumulator.

I will work to improve the initialization story in a followup PR, but current code is good enough to introduce the new concept.


jkwatson (approved):
This is looking good and getting close to being easily mappable to the current spec.

Two things I see as follow-ups:
1) I think I would rename the MetricProducer to "MetricProcessor" to more closely match the current naming in the spec.
2) It would be good to make it more obvious when the Accumulator's aggregator gets reset in the cycle

Thanks!

PR #2271, opened 2020-12-10
Adding instructions to run benchmarks

srprash:


jkwatson (approved):
Awesome. Thanks! I always end up having to dig into the gradle file to find these, so this is a good doc update!

anuraaga (approved):

PR #2273, opened 2020-12-10
Multi propagation

### LINKED ISSUE #2244: TraceMultiPropagator vs. MultiTextMapPropagator
There are two very similar ways and classes that allow multiple trace context propagators to be in play at the same time:
* TraceMultiPropagator - lives in extensions module, created by invoking `TraceMultiPropagator.create(...)`
* MultiTextMapPropagator - created by invoking `TextMapPropagator.composite(...)`

The implementations are very similar -- they both allow context injection and extraction by delegating to a list of other propagators.  However, there is a subtle difference between them,  due to the fact that `TraceMultiPropagator.extract()` stops extracting after the first success.

This subtle difference could be confusing to users.  
Since it does seem likely that users will want/need to support multiple formats, we should keep this...but we should avoid the duplicate implementation.  

Let's remove the `TraceMultiPropagator` and add a "stopAfterFirstFound" flag to the `MultiTextMapPropagator` constructor that will be referenced when extracting.  This should default to `false` to preserve the existing behavior.  
The implementation should also beef up the javadoc in `MultiTextMapPropagator`.

breedx-splk:
This resolves #2244 by merging `TraceMultiPropagator` functionality into `MultiTextMapPropagator` and removing `TraceMultiPropagator` from the extensions.


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 #### Breaking Changes
 
+- TraceMultiPropagator extension has been removed. Users should migrate to the instance returned
```
jkwatson, line 9: this should probably move down a section about extensions rather than the API. And this is an enhancement to the API, rather than a breaking change, I think.

api/context/src/main/java/io/opentelemetry/context/propagation/MultiTextMapPropagator.java
```diff
@@ -17,39 +20,133 @@
...
+      Context updatedContext = textPropagators[i].extract(context, carrier, getter);
+      if (stopExtractAfterFirst && (updatedContext != context)) {
+        return updatedContext;
+      }
+      context = updatedContext;
     }
     return context;
   }
+
+  static class Builder {
```
jkwatson, line 109: we've been favoring pulling these out into their own class, rather than making them inner classes. Still good to keep nice and package-access, though.

api/context/src/main/java/io/opentelemetry/context/propagation/MultiTextMapPropagator.java
```diff
@@ -17,39 +20,133 @@
...
   private final Collection<String> allFields;
+  private final boolean stopExtractAfterFirst;
+  private final boolean iterateBackwards;
 
-  MultiTextMapPropagator(List<TextMapPropagator> textPropagators) {
-    this.textPropagators = new TextMapPropagator[textPropagators.size()];
-    textPropagators.toArray(this.textPropagators);
-    this.allFields = Collections.unmodifiableList(getAllFields(this.textPropagators));
+  /** Returns a {@link MultiTextMapPropagator} for the given {@code propagators}. */
+  public static TextMapPropagator create(TextMapPropagator... propagators) {
```
jkwatson, line 27: these methods don't need to be public, right, since the class itself is not?

jkwatson:
looks like it needs a spotless run, and a few small cleanups, then looks good to go. Thanks!

CHANGELOG.md
breedx-splk: We talked about this and agreed to let the release manager work it out.  😉

api/context/src/main/java/io/opentelemetry/context/propagation/MultiTextMapPropagator.java
breedx-splk: I didn't see too much prior art to support this, but I did it anyway.  👍

api/context/src/main/java/io/opentelemetry/context/propagation/MultiTextMapPropagator.java
breedx-splk: Sure.  Is the idea to force users to go through the interface `create()` methods then?  Hide all implementations?  Is that the idea?

api/context/src/main/java/io/opentelemetry/context/propagation/TextMapPropagator.java
```diff
@@ -62,17 +60,19 @@ static TextMapPropagator composite(TextMapPropagator... propagators) {
...
+
+  /**
+   * Returns a {@link TextMapPropagator} which simply delegates injection and extraction to the
+   * provided propagators. Extraction will short circuit when the first propagator updates the
+   * context.
+   *
+   * <p>Invocation order of {@code TextMapPropagator#inject()} and {@code
+   * TextMapPropagator#extract()} for registered trace propagators is undefined.
+   */
+  static TextMapPropagator compositeWithShortCircuit(Iterable<TextMapPropagator> propagators) {
```
anuraaga, line 74: Let's add the varargs version too

api/context/src/main/java/io/opentelemetry/context/propagation/TextMapPropagatorBuilder.java
```diff
@@ -0,0 +1,58 @@
...
+      this.delegates.add(propagator);
+    }
+  }
+
+  public TextMapPropagatorBuilder stopExtractAfterFirst() {
+    this.stopExtractAfterFirst = true;
+    return this;
+  }
+
+  public TextMapPropagatorBuilder iterateBackwards() {
```
anuraaga, line 35: Do we need / use iterateBackwards?

anuraaga (commented):
Thanks for the help

api/context/src/main/java/io/opentelemetry/context/propagation/TextMapPropagatorBuilder.java
breedx-splk: Good call out. Because the original impl was in the external (fake contrib) module, it's hard to get a sense of how widely it might be used...so I am inclined to remove it.  Curious what other folks think. 

I mean, if you want reverse iteration order, reverse the list first?  🤓

breedx-splk:
Ok, I think we might be taking a different approach -- simply deprecating `TraceMultiPropagator` and deleting it in a future release.  @carlosalberto are you ok with this idea?

api/context/src/main/java/io/opentelemetry/context/propagation/MultiTextMapPropagator.java
jkwatson: yessir

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 #### Breaking Changes
 
+- TraceMultiPropagator extension has been removed. Users should migrate to the instance returned 
+  from `TextMapPropagator.compositeWithShortCircuit()`.
```
jkwatson, line 10: were you going to remove this changelog entry?

jkwatson: nm. looks like you haven't done anything with the comments yet. sorry for the noise!

breedx-splk: I mean, I said I was going to leave it for the release manager to wrangle, but if that's just noise I'm happy to remove it too.

jkwatson: yeah, remove it for now, so it doesn't end up in the wrong place and I forget about it. :)

breedx-splk:
Closing this one so that we can make a new one that more simply just deprecates the `TraceMultiPropagator`.

Issue #2125, opened 2020-11-24
AggregationTemporality is set incorrectly in the export pipeline


sirianni:
**Describe the bug**
`AggregationTemporality` is set incorrectly in export pipeline

**Steps to reproduce**
Instantiate the `OtlpGrpcMetricExporter` using a custom `AggregationConfiguration` with non-default `Temporality`
```java
    MeterSdkProvider meterProvider = OpenTelemetrySdk.getGlobalMeterProvider();

    InstrumentSelector instrumentSelector = InstrumentSelector.newBuilder()
        .instrumentType(InstrumentType.COUNTER)
        .build();
    AggregationConfiguration viewSpecification = AggregationConfiguration.create(
        Aggregations.sum(),
        Temporality.DELTA
    );
    meterProvider.registerView(instrumentSelector, viewSpecification);

    IntervalMetricReader intervalMetricReader = IntervalMetricReader.builder()
        .setMetricExporter(OtlpGrpcMetricExporter.getDefault())
        .setMetricProducers(Collections.singleton(meterProvider.getMetricProducer()))
        .setExportIntervalMillis(30000)
        .build();
```

**What did you expect to see?**
I expected to see the generated `Metric` proto with `AGGREGATION_TEMPORALITY_DELTA`

**What did you see instead?**
I saw the generated `Metric` proto with `AGGREGATION_TEMPORALITY_CUMULATIVE`
```
InstrumentationLibraryMetrics #0
Metric #0
Descriptor:
     -> Name: http_requests
     -> Description: 
     -> Unit: 1
     -> DataType: IntSum
     -> IsMonotonic: true
     -> AggregationTemporality: AGGREGATION_TEMPORALITY_CUMULATIVE
```

**What version and what artifacts are you using?**
Artifacts: `opentelemetry-sdk` `opentelemetry-exporter-otlp`
Version: v0.11.0
How did you reference these artifacts? 
```xml
      <dependency>
        <groupId>io.opentelemetry</groupId>
        <artifactId>opentelemetry-bom</artifactId>
        <version>0.11.0</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
    <dependency>
      <groupId>io.opentelemetry</groupId>
      <artifactId>opentelemetry-sdk</artifactId>
    </dependency>
    <dependency>
      <groupId>io.opentelemetry</groupId>
      <artifactId>opentelemetry-exporter-otlp</artifactId>
    </dependency>
```

**Environment**
[...]

PR #2274, opened 2020-12-11
Change MetricData to be consistent with protos, move creation to the Aggregation

bogdandrutu:
Currently the temporality is not handled correctly, and in all of the cases I am setting it as CUMULATIVE, but in this PR I adding the capability to set temporality from the SDK.

Few fixes in this PR:
* No empty metrics emitted from the SDK. We used to always emit metadata about the metrics without any points even if there were no recordings.

### DIFF
--- a/exporters/prometheus/src/test/java/io/opentelemetry/exporter/prometheus/PrometheusCollectorTest.java
+++ b/exporters/prometheus/src/test/java/io/opentelemetry/exporter/prometheus/PrometheusCollectorTest.java
@@ -55,23 +55,27 @@ class PrometheusCollectorTest {
 
   private static ImmutableList<MetricData> generateTestData() {
     return ImmutableList.of(
-        MetricData.create(
+        MetricData.createLongSum(
             Resource.create(Attributes.of(stringKey("kr"), "vr")),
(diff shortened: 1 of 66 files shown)


exporters/prometheus/src/test/java/io/opentelemetry/exporter/prometheus/PrometheusCollectorTest.java
```diff
@@ -55,23 +55,27 @@ void registerToDefault() throws IOException {
 
   private static ImmutableList<MetricData> generateTestData() {
     return ImmutableList.of(
-        MetricData.create(
+        MetricData.createDoubleSum(
```
jkwatson, line 58: previously this was a long sum; seems like there would be value in testing both long and double here, rather than only double.

bogdandrutu: Sorry :) good catch.

jkwatson:
This looks like it may fix #2125 as well, although I can't quite reason it through without testing it.

jkwatson (approved):
Looks good. Thanks!

PR #2275, opened 2020-12-11
Remove OpenTelemetryBuilder interface, every implementation will have it's own builder

bogdandrutu:
This PR also fixes the DefaultOpenTelemetry builder to be it's own class and not a sub-class.


jkwatson (approved):
This looks like it's in alignment with where we're headed in #2262 . Approving, but let's let @anuraaga also give it approval before merging.

bogdandrutu:
@anuraaga already approved the PR I closed because of the ask to deprecate instead of removing. But will wait.

anuraaga (approved):

PR #2278, opened 2020-12-11
Call point constructor with the value got instead of getting the value again

bogdandrutu:
Not a problem in the Double version where this is implemented correctly.


breedx-splk (approved):
Pretty sure JIT would be all over this, but I like this change because it reads better anyway.  👍

bogdandrutu:
> Pretty sure JIT would be all over this, but I like this change because it reads better anyway. 👍

JIT cannot handle a failing case :) you get a value check if not `null` then get the value again (may become `null`) between these 2 moments, so we will crash. So because the value is atomic JIT has to read the new value to ensure it did not change.

jkwatson (approved):

breedx-splk:
> > Pretty sure JIT would be all over this, but I like this change because it reads better anyway. 👍
> 
> JIT cannot handle a failing case :) you get a value check if not `null` then get the value again (may become `null`) between these 2 moments, so we will crash. So because the value is atomic JIT has to read the new value to ensure it did not change.

Oh, great point!  I missed that.  Thank you for clarifying @bogdandrutu .

PR #2279, opened 2020-12-11
Deprecate metrics from opentelemetry-api, add a global version in metrics api

bogdandrutu:
One small caveat, the GlobalMeterProvider class may be removed when we do the stable release. When the metrics will be stable/release candidate we need to find a different solution for that class in order to avoid removing it just before the release.


exporters/otlp/src/main/java/io/opentelemetry/exporter/otlp/OtlpGrpcSpanExporter.java
```diff
@@ -105,6 +99,12 @@
    *     or to a negative value, the exporter will wait indefinitely.
    */
   private OtlpGrpcSpanExporter(ManagedChannel channel, long deadlineMs) {
+    Meter meter = GlobalMetricsProvider.getMeter("io.opentelemetry.exporters.otlp");
+    this.spansSeen =
+        meter.longCounterBuilder("spansSeenByExporter").build().bind(EXPORTER_NAME_LABELS);
+    LongCounter spansExportedCounter = meter.longCounterBuilder("spansExportedByExporter").build();
+    this.spansExportedSuccess = spansExportedCounter.bind(EXPORT_SUCCESS_LABELS);
+    this.spansExportedFailure = spansExportedCounter.bind(EXPORT_FAILURE_LABELS);
```
jkwatson, line 107: I thought we had already done this. Good catch!

jkwatson (approved):
Thanks!

exporters/otlp/src/main/java/io/opentelemetry/exporter/otlp/OtlpGrpcSpanExporter.java
bogdandrutu: I need to unbind them in shutdown.

jkwatson:
looks like this needs a rebase, @bogdandrutu

exporters/otlp/src/main/java/io/opentelemetry/exporter/otlp/OtlpGrpcSpanExporter.java
bogdandrutu: Done.

bogdandrutu:
@anuraaga does this look good for you?

PR #2281, opened 2020-12-11
[DO_NOT_MERGE] Remove metrics from the first stable version of otel

bogdandrutu:
Depends on https://github.com/open-telemetry/opentelemetry-java/pull/2279

This PR will be merged after the next release, but it is here to prove that we can remove metrics, and we deprecated all the right classes/methods.


bogdandrutu:
Replaced by https://github.com/open-telemetry/opentelemetry-java/pull/2341

PR #2284, opened 2020-12-13
Make AttributeKey compare normally with both type/key.

### LINKED ISSUE #2282: Attributes.get() method does not provide type safety
Since `AttributeKey.equals()` does not include the type, the compiler thinks that this code is type safe, but it will throw `ClassCastException`:

```
AttributeKey<Long> key = AttributeKey.longKey("mykey");
Attributes attributes = Attributes.of(key, 5L);
String value = attributes.get(AttributeKey.stringKey("mykey"));
```

If nothing can be done about this, a warning in the `Attributes.get()` javadoc would be helpful.

anuraaga:
Instead passes a special `Comparator` to sortAndFilter where the old behavior is actually needed. A bit annoying, but all inside an internal class.

Also tried out a testing library I've been wanting to check out :) It was reporting the lack of getType in equals before the change.

Fixes #2282


jkwatson (approved):
Thanks! This is a good way to solve the deduping issue.

PR #2285, opened 2020-12-14
Make classes that can be shutdown Closeable

anuraaga:
Regardless of where https://github.com/open-telemetry/opentelemetry-specification/issues/1288 goes, I think it's a no-brainer to allow using Java's mechanism for cleanup.


Oberon00:
Why Closeable instead of AutoCloseable?

anuraaga:
@Oberon00 AFAIK, the only difference is that `Closeable.close` is required to be idempotent (second+ calls to close are ignored), while `AutoCloseable.close` doesn't. I guess we like the former here?

Oberon00:
And the `throws Exception` vs `throws IOException`.

anuraaga:
> And the throws Exception vs throws IOException.

Yeah - currently none of our methods throw checked exceptions so these close methods don't even declare it. I wonder if they should though, hmm.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SpanProcessor.java
```diff
@@ -99,4 +101,13 @@ default CompletableResultCode shutdown() {
...
     return CompletableResultCode.ofSuccess();
   }
+
+  /**
+   * Closes this {@link SpanProcessor} after processing any remaining spans, releasing any
+   * resources.
+   */
+  @Override
+  default void close() {
+    shutdown().join(10, TimeUnit.SECONDS);
```
Oberon00, line 111: This seems a bit questionable. If we have a shutdown timeout, should it be hardcoded or configurable? If it is hardcoded, maybe use a constant?

Oberon00: Also, maybe it would be better to throw a timeout exception if the shutdown does not complete.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SpanProcessor.java
```diff
@@ -99,4 +101,13 @@ default CompletableResultCode shutdown() {
   default CompletableResultCode forceFlush() {
     return CompletableResultCode.ofSuccess();
   }
+
+  /**
+   * Closes this {@link SpanProcessor} after processing any remaining spans, releasing any
+   * resources.
+   */
+  @Override
+  default void close() {
```
bogdandrutu, line 110: Do we need a shutdown and a close?

jkwatson:
I can understand the desire to have a more "standard" API to use, but I'm not sure that `close()` really evokes what we're doing with these instances. We're not necessarily "closing" anything at all with this operation. 

I'm not 100% opposed to this, but I also agree with @bogdandrutu that we should probably only have one, consistent way to shut down SDK components. I don't think we're ever recommend that someone use a try-with-resources block on SDK components, so I'm not really sure how this will help an end user. Is there a particular use-case or user that you are targeting with this change?

anuraaga:
> I can understand the desire to have a more "standard" API to use, but I'm not sure that close() really evokes what we're doing with these instances. We're not necessarily "closing" anything at all with this operation.

I don't know if I follow - isn't `close` an idiomatic way to turn off an instance, releasing its resources?

> I don't think we're ever recommend that someone use a try-with-resources block on SDK components, so I'm not really sure how this will help an end user. Is there a particular use-case or user that you are targeting with this change?

If I was writing a from-scratch Java app, I would probably use try-with-resources since why-not? But the case I'm more thinking about is exposing these as Spring beans - then they'd be automatically closed by Spring for free.

jkwatson:
spring beans is a very interesting case. I didn't know that spring would auto-close Closeable beans when they are removed from use!

I'm not sure how you'd configured an SDK and have it run for the lifetime of the application and then shut it down using try-with-resources, but maybe I'm missing something obvious.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SpanProcessor.java
anuraaga: I don't think so if the spec is ok with changing names.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SpanProcessor.java
anuraaga: I think we expect this to finish relatively quickly, just finish things up. So maybe 10 seconds was too large, let me try 1 second. I'd prefer hard-coded just to keep things simpler, if extracting a constant are you thinking `internal` package?

> Also, maybe it would be better to throw a timeout exception if the shutdown does not complete.

I don't know if I've ever been able to handle an exception during shutdown so am ambivalent, but can add it. How useful does it seem?

anuraaga (commented):
@jkwatson Well, technically I wouldn't since Armeria uses non-daemon background threads for server processing, but I recall seeing servers that use daemon background threads and block `main()`. Imagined wrapping that in try/resources.

But really the main reason is Spring or similar DI containers that generally autoclose. And the main reason I mention anything else is to not be called a Spring-fanatic by @bogdandrutu  :P

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SpanProcessor.java
Oberon00: > I think we expect this to finish relatively quickly

Hmm, since shutdown includes flush, it will often make network requests.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SpanProcessor.java
jkwatson: What was the outcome of the discussion of having the shutdown method be optional? Were people ok with using a language-appropriate idiom if one was available?

Oberon00: I think using the name `close` instead of `shutdown` would still qualify as providing `shutdown`. But that aspect was not discussed as it was concluded that generally a shutdown functionality is *not* optional (see https://github.com/open-telemetry/opentelemetry-specification/issues/1288).

jkwatson: hm. Is it worth adding a clarification to the spec, or do we think this is within our control as language experts?

Oberon00: I think no one will complain about some SDK method name. But a general clarification in the spec (library-guidelines.md maybe) would be good like

> Method/constant/... names defined in the spec SHOULD be adjusted to language naming conventions (snake_case, PascalCase, ...) but otherwise used as-is. As an exception, if a well-established name/naming pattern exists in a language for a particular functionality, it SHOULD be used instead. In general, the names should be choosen to adhere to the spec as closely as possible without being awkward or unnatural in the target language.

jkwatson:
So, we have 3 options with this:

1) have two methods..one blocking and one non-blocking (potentially) for shutting down components. [note: we should make all the `shutdown()` methods return `CompletableResultCode` for consistency, if we keep the possibly non-blocking version.]
2) have just one method, java-idiomatic `close()` method that is always blocking. we need to decide on timeout behavior, though. This might also require a bit of spec clarification to make sure we're still compliant with a differently named method.
3) have just one method, `shutdown()` that is possibly non-blocking (see note above about consistency).

I think I'm in favor of option 1, but I'm on the fence a bit about how to handle timeouts with `close()`. I think we either a) block indefinitely until the `shutdown()` result completes or b) fire-and-forget and always return immediately, ignoring the result entirely. Having an arbitrary hard-coded timeout is a bit of a mushy middle that would probably be unexpected.

anuraaga:
> I think we either a) block indefinitely until the shutdown() result completes or b) fire-and-forget and always return immediately, ignoring the result entirely. Having an arbitrary hard-coded timeout is a bit of a mushy middle that would probably be unexpected.

I don't think we can do fire-and-forget since it defeats the purpose of shutdown on program exit, which is to make sure spans get sent. I'm ok with blocking indefinitely, but is a 10s timeout then all that mushy? It seems excessively long, enough to get the spans out, without being too long in that if for some reason it's stuck, at least other components in the app can be shutdown.

jkwatson:
@anuraaga thoughts on one method vs. two, and whether all the `shutdown()`s should return `CompletableResultCode` (assuming we keep them)?

anuraaga:
@jkwatson I would lean towards just one `close`, but if we kept `shutdown` would definitely want all of them to return the code

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkTracerManagement.java
```diff
@@ -70,4 +71,22 @@
...
+   *
+   * <p>After this is called, newly created {@code Span}s will be no-ops.
+   *
+   * <p>After this is called, further attempts at re-using or reconfiguring this instance will
+   * result in undefined behavior. It should be considered a terminal operation for the SDK
+   * implementation.
+   */
+  @Override
+  default void close() {
+    shutdown();
```
jkwatson, line 90: can you create an issue to make the `shutdown()` return `CompletableResultCode` for consistency? Thanks!

jkwatson (approved):
I think this is fine. We can take on the issue of consistency and possible removal of the `shutdown()` method altogether as a separate issue.

Oberon00:
I'm a bit skeptical of this. `shutdown` might be an expensive operation doing network I/O, for short-running processes it might be the only time it does so, so it might need to do DNS lookup + TLS handshake + actually transferring data + waiting for the response code. This can take time, and it may depend very much on the app how much time is acceptable. If we are shutting down a server or a batch job and we know we have a bad connection, maybe one minute is fine. If we have a short-running app and it's OK if we have no telemetry on some runs, then maybe only two seconds are OK.

`close` is very easy to use, e.g. with a try-with-resources block, but maybe it is too easy in that case. The user should maybe think a bit of how they want to handle the shutdown.

Oberon00:
One idea would be to have something like `try (Scope ignored = myTracerProvider.shutdownAfterwards(10, TimeUnit.SECONDS)`.

jkwatson:
> One idea would be to have something like `try (Scope ignored = myTracerProvider.shutdownAfterwards(10, TimeUnit.SECONDS) { ... }`.

If you want to handle the timeout, you can use the `shutdown()` method as-is and wait for as long (or as short) as you want to the returned result code. Is that not enough to satisfy this use-case?

Oberon00:
If we are OK with shutdown staying in addition to close, then I think that's fine, yes.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkTracerManagement.java
jkwatson: Note: #2422 implements this, so I don't know that we need an issue.

anuraaga:
@Oberon00 Is this PR ok as it is, and given #2422 ?

anuraaga:
Going to go ahead and merge this, let me know if there's any followup needed.

PR #2288, opened 2020-12-14
Move DaemonThreadFactory to internal package.

anuraaga:
Doesn't seem like something we need to share outside our repo.


bogdandrutu (commented):
What is the rule about internal? Do we allow other artifacts in extensions to depend on this?

jkwatson:
> What is the rule about internal? Do we allow other artifacts in extensions to depend on this?

Great question. My opinion is that it's ok for our own extensions to use these, but I don't feel super strongly about this.

anuraaga:
Yeah - we decided that we expect users to use a BOM. The main reason to do so would be to share internal among artifacts :)

jkwatson (approved):
Thanks!

bogdandrutu (approved):

PR #2289, opened 2020-12-14
Add Context.toString

anuraaga:
It's difficult to debug context issues, such as IntelliJ debugger view or the error messages from `StringContextStorage`, without a toString on Context.

### DIFF
--- a/api/context/src/main/java/io/opentelemetry/context/ArrayBasedContext.java
+++ b/api/context/src/main/java/io/opentelemetry/context/ArrayBasedContext.java
@@ -81,4 +81,18 @@ final class ArrayBasedContext implements Context {
     newEntries[newEntries.length - 1] = value;
     return new ArrayBasedContext(newEntries);
   }
+
+  @Override
+  public String toString() {
+    StringBuilder sb = new StringBuilder("{");
+    for (int i = 0; i < entries.length; i += 2) {
+      sb.append(entries[i]).append('=').append(entries[i + 1]).append(", ");
+    }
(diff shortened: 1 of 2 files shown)


api/context/src/main/java/io/opentelemetry/context/ArrayBasedContext.java
```diff
@@ -81,4 +81,18 @@ private ArrayBasedContext(Object[] entries) {
     newEntries[newEntries.length - 1] = value;
     return new ArrayBasedContext(newEntries);
   }
+
+  @Override
+  public String toString() {
```
bogdandrutu, line 86: Add a comment that the string returned is not considered stable and cannot be used for other reasons than debugging.

jkwatson: If we're going to do that here, we should probably put that everywhere, since I think we want that to be true across the board. I wonder if we should just put this somewhere in our docs, rather than put it on all the methods?

jkwatson (approved):

api/context/src/main/java/io/opentelemetry/context/ArrayBasedContext.java
bogdandrutu: That is also good :)

jkwatson: I'll add a line to the CONTRIBUTING PR that I just opened.

PR #2291, opened 2020-12-14
Move Sampler builder to top level

anuraaga:


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/ParentBasedSampler.java
```diff
@@ -19,7 +19,7 @@
 /**
  * A Sampler that uses the sampled flag of the parent Span, if present. If the span has no parent,
  * this Sampler will use the "root" sampler that it is built with. See documentation on the {@link
- * Builder} methods for the details on the various configurable options.
+ * ParentBasedSamplerBuilder} methods for the details on the various configurable options.
  */
 @Immutable
 public class ParentBasedSampler implements Sampler {
```
jkwatson, line 25: can we make this class non-public now? That would be an improvement, if we can.

jkwatson (approved):
Thanks!

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/ParentBasedSampler.java
bogdandrutu: Also final :). Cannot be non-public if the "builder" method is on this class.

jkwatson: yeah, I realized that and forgot to remove the comment.

anuraaga: Nice, thanks!

Issue #2290, opened 2020-12-14
TracerSdkProvider or TracerProviderSdk?


anuraaga:
I always get confused by the naming of the SDK implementations since they feel inconsistent. `SpanBuilderSdk`, `TracerSdk` add a `Sdk` suffix to the interface they are implementing. `TracerSdkProvider` has `Sdk` in the middle. Presumably the latter name is because it's a provider of `TracerSdk` but I would lean towards consistency in suffixing API interface name with `Sdk`. `TracerSdkProvider` is the only public class, though, so I guess it's name is most important. In that sense too, since `TracerSdk` is an internal class, naming the API a provider of that hidden class has no real meaning, and I guess just the suffix could be easier to grok for end users.

What do people think?

anuraaga:
Ah, or `SdkTracerProvider` might make the most sense, I guess it's common to _prefix_ an interface, not suffix

iNikem:
+1 for `SDKXxxx`

Oberon00:
If possible, we should make that class package-private anyway (now that we have the Management interface implemented by it).

jkwatson:
> If possible, we should make that class package-private anyway (now that we have the Management interface implemented by it).

we'll need a public builder, but it would be a good restriction if the class itself could be non-public, 👍

jkwatson:
BTW, much prefer `Sdk` as a prefix, so 👍

anuraaga:
Thanks!

PR #2292, opened 2020-12-14
Move Sdk to be prefix of implementation classes instead of suffix/mid…

anuraaga:
…fix.

Sending this out to illustrate #2290. No worries if it doesn't seem like nice naming and we can close but figured it's easier to take a look. One observation is that we probably would continue with `OpenTelemetrySdk` instead of `SdkOpenTelemetry` since we use the wording OpenTelemetry SDK so much. While a bit inconsistent, I think it's probably ok.


Oberon00 (approved):
Rolls off the tongue much better now. Thanks!

jkwatson (approved):
I think this is a great change. I remember being confused about the naming choice long ago, but never bothered to go and make the change. 👍 🚀

jkwatson:
@bogdandrutu any objection to this naming update?

bogdandrutu (approved):

PR #2293, opened 2020-12-14
Manually structure output of log exporter instead of delegating to un…

anuraaga:
…stable toString

For now, I mostly copied the logging from instrumentation repo

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/javaagent-exporters/logging/src/main/java/io/opentelemetry/javaagent/exporters/logging/LoggingExporter.java#L32

I think we agree we should use something more stable than `toString` for this exporter, but I don't know what format we'd all like. I'm thinking of something like `{name: foo, traceId: aefaefaefefafe, spanId: afeeafaefaef, attributes: {foo="bar"}, events: [], links: []}`. Any preferences?


trask:
maybe otlp json format?

anuraaga:
@trask Hmm - I guess that would be best implemented somehow sharing code with `OtlpGrpcSpanExporter` which converts to OTLP. Perhaps that's a better approach and this should just stick to tostring - any thoughts on changing the logging exporter the agent configures to an OtlpLoggingSpanExporter?

trask:
> any thoughts on changing the logging exporter the agent configures to an OtlpLoggingSpanExporter

I think this would be great. I like the idea of a stable (even spec'd) logging format.

Oberon00:
What is the trace/span ID format in OTLP/JSON? Still Base64 or hex? I think having a base64 format in the logging exporter would be very confusing.

anuraaga:
OTLP/JSON uses hex - hence it will be very annoying to actually implement it here, heh. Will wee what we can do.

jkwatson:
I wrote this originally as something quick-and-dirty; not intended to really be the final implementation. So, I'm definitely in favor of having some sort of stable, standardized format. Makes me wonder if this format is something that could/should be specced, cross-language.

Anyway...I have few opinions about the specific format, but I do think sticking with hex/base16 is a good idea, to reduce confusion with all of the other internal toString formats.

exporters/logging/src/main/java/io/opentelemetry/exporter/logging/LoggingSpanExporter.java
```diff
@@ -15,12 +15,23 @@
...
   @Override
   public CompletableResultCode export(Collection<SpanData> spans) {
+    // We always have 32 + 16 + name + several whitespace, 60 seems like an OK initial guess.
+    StringBuilder sb = new StringBuilder(60);
     for (SpanData span : spans) {
-      logger.log(Level.INFO, "span: {0}", span);
+      sb.setLength(0);
+      sb.append(span.getName())
+          .append(" ")
+          .append(span.getTraceId())
```
jkwatson, line 29: any thoughts to prefixing `"t"` before the trace id and `"s"` before the span id?

anuraaga: The drastically different length of the values makes me ok without the prefix.

anuraaga:
Well given we'll probably have this logger even if having an OTLP logger as well, undrafting it. Still open to any suggestions on what format to use here, keeping in mind that most agent users are seeing this output.

exporters/logging/src/main/java/io/opentelemetry/exporter/logging/LoggingSpanExporter.java
jkwatson: yeah, I guess at the moment, we can't get shorter traceIds into here, since we'll always zero-pad if we get on. 👍

jkwatson (approved):
This seems cleaner, for sure, than what we have right now. Thanks!

We can always add more fields here if someone asks for it.

PR #2295, opened 2020-12-14
Add a OTLP JSON span exporter to logs.

anuraaga:
This uses a protobuf-json [library](https://github.com/curioswitch/curiostack/tree/master/common/grpc/protobuf-jackson) I wrote a while ago, and we use it in Armeria for production. It's based on Jackson vs gson, which I think has a higher chance of being on an app's classpath which is nice. The use of Jackson also made it quite easy to switch the binary marshalling to hex from base64. But it's designed for speed (for use in production) so uses bytebuddy for runtime code generation as opposed to runtime reflection like the official json-format does. That does unfortunately make it more heavyweight than it could be, but at least it's fast 😅.

Wondering where to put this, probably not in `-otlp`, new `-otlp-logging`, or replace the current `-logging`? Leaning towards `-otlp-logging`.


anuraaga:
For reference, in official, the base64 encoding is hard coded inside this private method, so no go

https://github.com/protocolbuffers/protobuf/blob/master/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java#L1235

jkwatson:
how "heavyweight" does it end up being? My only concern with replacing the existing logging exporter is how much users would be taking on to get it. `-otlp-logging` seems good to me, otherwise. We can always decide to replace later, if there's appetite for it.

anuraaga:
@jkwatson I think ~4MB for byte-buddy + jackson. That's after I do some pruning since it's currently getting ~3MB of guava too due to a dependency on protobuf-java-util for a very small amount of logic, as it stands it's about 7MB. Let me go ahead and make a new module for it at least for now.

trask:
are we going to break out `otlp-metrics` and `otlp-spans` modules due to versioning? maybe this should be `logging-otlp`, it seems more a variant of logging to me than a variant of otlp (not sure if I'm making sense)

anuraaga:
Ah forgot to mention one more caveat - we'll need an `otlp-common` to contain the proto translation code. How's that sound?

> are we going to break out otlp-metrics and otlp-spans modules due to versioning? 

Ah yeah this was the plan

> maybe this should be logging-otlp, it seems more a variant of logging to me than a variant of otlp (not sure if I'm making sense)

Makes sense to me

anuraaga:
Hmm - well I guess the README is saying it's ok for me to move the code into the extension?

https://github.com/open-telemetry/opentelemetry-java/tree/master/sdk-extensions/otproto

bogdandrutu:
@anuraaga haven’t read the code, but why did you mention JSON encoding in the other PR? Can you add JSON format? We should prioritize proto

anuraaga:
@bogdandrutu This does use JSON serialization of the proto. Meaning it relies on e.g., `SpanAdapter`, classes that technically aren't needed if adding #2304

bogdandrutu:
I read the PR and I think the tittle is confusing, this is not an exporter for OTLP logs, but an exporter that writes to the logs. Do we need this?

anuraaga:
@bogdandrutu There was interest in a fully spec'd, stable format for spans when writing to logs in #2293. It seems nice, but could be overkill. @jkwatson @trask any thoughts? I've heard Splunk wanting to use logs for exporting spans in AWS Lambda, maybe @bogdandrutu @iNikem you know if this could help with that?

trask:
We've seen interest in using the logging exporter beyond getting started / toy apps (e.g. https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1600).

So it would be great if the output of the logging exporter can have some sort of stability guarantee, similar to the other stability guarantees we have been discussing.

OTLP/JSON[*] seems ideal, only because we get it for "free" and wouldn't have to invent something new.

[*]without the weird traceid/spanid format

exporters/otlpjson/build.gradle
```diff
@@ -0,0 +1,22 @@
...
+    id "maven-publish"
+
+    id "ru.vyarus.animalsniffer"
+}
+
+description = 'OpenTelemetry Protocol JSON Exporters'
+ext.moduleName = "io.opentelemetry.exporter.otlpjson"
+
+dependencies {
+    compileOnly project(':sdk:trace')
```
anuraaga, line 12: @jkwatson So I tried a different approach to splitting all exporters for trace vs metrics here since it seemed like it could be a bit too much. If going with this, I'd revert the change to split out the `otlp` exporters in favor of `compileOnly` dependencies.

One of the reasons it felt like much is we would need to split otproto up as well - they all end up having a common, trace, metrics split. Lots of artifacts.

But it means that previously, a user could only depend on `exporter-otlp` to have the SDK as well, but with such a scheme, they'd have to additionally add the SDK as a dependency. I wonder if this is too annoying or reasonable.

exporters/otlpjson/README.md
```diff
@@ -0,0 +1,9 @@
+# OpenTelemetry - OTLP JSON Exporters
```
jkwatson, line 1: If this is purely a logging exporter, this module should be named `logging-otlp` I think.

exporters/otlpjson/src/test/java/io/opentelemetry/exporter/otlpjson/OtlpJsonLoggingMetricExporterTest.java
```diff
@@ -0,0 +1,163 @@
...
+              + "          \"timeUnixNano\": \"2\",\n"
+              + "          \"value\": 4.0\n"
+              + "        }],\n"
+              + "        \"aggregationTemporality\": \"AGGREGATION_TEMPORALITY_CUMULATIVE\",\n"
+              + "        \"isMonotonic\": true\n"
+              + "      }\n"
+              + "    }]\n"
+              + "  }]\n"
+              + "}",
+          logged.get(0).getMessage(),
```
jkwatson, line 146: If I read this right...the log message will be split across multiple lines? Don't we want to make sure that every log message is only a single line, so it can work with normal logging systems?

exporters/otlpjson/src/test/java/io/opentelemetry/exporter/otlpjson/OtlpJsonLoggingSpanExporterTest.java
```diff
@@ -0,0 +1,192 @@
...
+              + "            \"boolValue\": true\n"
+              + "          }\n"
+              + "        }]\n"
+              + "      }],\n"
+              + "      \"status\": {\n"
+              + "        \"code\": \"STATUS_CODE_OK\"\n"
+              + "      }\n"
+              + "    }]\n"
+              + "  }]\n"
+              + "}",
```
jkwatson, line 174: same question about multi-line log message

exporters/otlpjson/build.gradle
jkwatson, line 12: Would "provided" be a semantically more correct dependency declaration?

jkwatson: So, since the trace SDK is supposed to be stable, I think it would ok to declare this as an `api` dependency and get that for free, like people would probably expect. Keeping the metrics dependency as `provided` might be ok, though, since someone has to explicitly opt-in to depending on it.

exporters/otlpjson/README.md
anuraaga: Yeah - there's some tension with the logging signal. But went with that name.

exporters/otlpjson/build.gradle
anuraaga: Do you think so long term? e.g., what if someone replaces the trace SDK with a custom one, would we want the exporter artifacts to still bring in the trace SDK? I guess it could be ok, as custom SDKs still seem rare enough to me. Just want to confirm.

anuraaga: I think provided is "expected to be there", e.g., tomcat, and optional is "add if you want". `compileOnly` roughly corresponds with optional.

exporters/otlpjson/src/test/java/io/opentelemetry/exporter/otlpjson/OtlpJsonLoggingMetricExporterTest.java
anuraaga: Thanks for noticing! This is a JSON assert so whitespace is ignored. I added additional checks that the message has no newlines and realized I needed to set that.

exporters/otlpjson/build.gradle
jkwatson: If it's a completely custom SDK, it might have a completely different exporter interface, yeah? This does make me think we should have a separate artifact with the "SDK Extension Points" in it, including exporter, span processor, sampler interfaces, so at least custom SDKs can use the same extension points.

exporters/otlpjson/README.md
jkwatson: heh...yeah, I had forgotten about that bit.

exporters/logging-otlp/build.gradle
```diff
@@ -0,0 +1,22 @@
+plugins {
+    id "java-library"
+    id "maven-publish"
+
+    id "ru.vyarus.animalsniffer"
+}
+
+description = 'OpenTelemetry Protocol JSON Logging Exporters'
+ext.moduleName = "io.opentelemetry.exporter.logging.otlp"
```
jkwatson, line 9: my only possible quibble is that maybe this should be `"io.opentelemetry.exporter.logging-otlp"` but I'm not quite sure which is going to be more discoverable by end-users. Maybe the description could be more .. descriptive about exactly what's going on here.  Maybe `description = 'OpenTelemetry Logging exporters with OTLP JSON formatting'` ?

jkwatson (approved):
Thanks! Can't wait to try it out!

exporters/otlpjson/README.md
trask: oh no, I hadn't thought about that either, that's going to be very confusing, should we rename our existing logging exporter to "file" exporter, or "console" exporter, or ...

PR #2304, opened 2020-12-15
Add custom proto marshaler to avoid unnecessary allocation

bogdandrutu:
Some of the differences with the generated code:
1. Avoid builders to construct the marshalers;
2. Use less objects for Attributes because of the way `oneof` generated code is implemented; Also proto files supports arrays of any combination which is not the case for us.
3. Take advantage on the fact that data we know are immutable, no need to wrap to unmodifiable list/maps etc;

Unfortunately we cannot get the whole performance back because we need to write a custom Marshaler for grpc (which is possible but try to limit the PR size).

As the numbers show we have ~25% improvement in CPU and ~15% in memory with this PR. With a custom marshaler we can get anoter 7% CPU and 7% memory.

CPU benchmarks:
Screen Shot 2020-12-14 at 8 53 23 PM

Memory rate/op:
Screen Shot 2020-12-14 at 8 53 53 PM

**UPDATE**

After one more round of improvements (based on @anuraaga hint that I can use `byte[]`, I changed all the places where we wrapped the `byte[]` in `BytesString` and won a lot more performance, now we are 55% CPU usage win and 57% Memory. 

CPU benchmarks:
Screen Shot 2020-12-15 at 8 58 14 PM

Memory rate/op:
Screen Shot 2020-12-15 at 8 58 24 PM


anuraaga:
@bogdandrutu Can you describe at high level the main difference between the protoc generated code?

bogdandrutu:
@anuraaga will do that tomorrow, but in the meantime would be good to send a PR to remove this annoying check for tests and benchmarks (good for code):
```
GitHub Actions
/ Build (ubuntu-latest)
exporters/otlp/src/jmh/java/io/opentelemetry/exporter/otlp/RequestMarshalBenchmarks.java#L24
[Task :opentelemetry-exporter-otlp:checkstyleJmh FAILED] [JavadocMethod] Missing a Javadoc comment.
```

anuraaga (commented):
Ah I see this is manually serializing the SDK types directly to the bytestream. Nice but do you plan on removing the existing conversion code? It's useful for JSON encoding like in https://github.com/open-telemetry/opentelemetry-java/pull/2295

bogdandrutu:
For the moment is good to have the generated code to sanity check and test our manual conversion.

bogdandrutu:
I am using the ids and few things from the generated code like the enums

exporters/otlp/src/main/java/io/opentelemetry/exporter/otlp/AttributeMarshaler.java
```diff
@@ -0,0 +1,293 @@
...
+      case STRING_ARRAY:
+        return new KeyValueArrayStringMarshaler(key, (List<String>) value);
+      case LONG_ARRAY:
+        return new KeyValueArrayLongMarshaler(key, (List<Long>) value);
+      case BOOLEAN_ARRAY:
+        return new KeyValueArrayBooleanMarshaler(key, (List<Boolean>) value);
+      case DOUBLE_ARRAY:
+        return new KeyValueArrayDoubleMarshaler(key, (List<Double>) value);
+    }
+    throw new IllegalArgumentException("");
```
jkwatson, line 64: we should have a better description here, even if it won't happen under any normal circumstance

bogdandrutu: :))))

jkwatson (approved):
Thanks. I like that we compare the output with the generated code output as a test. 👍

exporters/otlp/src/jmh/java/io/opentelemetry/exporter/otlp/RequestMarshalBenchmarks.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  @Benchmark
+  @Threads(1)
+  public ByteBuffer marshalProto(RequestMarshalState state) throws IOException {
+    ExportTraceServiceRequest protoRequest =
+        ExportTraceServiceRequest.newBuilder()
+            .addAllResourceSpans(SpanAdapter.toProtoResourceSpans(state.spanDataList))
+            .build();
+    ByteBuffer protoOutput = ByteBuffer.allocate(protoRequest.getSerializedSize());
```
anuraaga, line 46: If I recall correctly, when size is known gRPC allocates `byte[]` how about using `byte[]` instead of `ByteBuffer`?

exporters/otlp/src/main/java/io/opentelemetry/exporter/otlp/AttributeMarshaler.java
```diff
@@ -0,0 +1,293 @@
...
+import java.io.IOException;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+
+abstract class AttributeMarshaler extends MarshalerWithSize {
+  private static final AttributeMarshaler[] EMPTY_REPEATED = new AttributeMarshaler[0];
+  private final ByteString key;
+  private final int valueSize;
+
+  static AttributeMarshaler[] createRepeated(Attributes attributes) {
```
anuraaga, line 25: It seems unfortunate to instantiate marshallers for each object. Isn't it better to define `Marshaler` something like

```java
interface MarshalerWithSize<T> {

  int serializedSize(T value);

  void writeValue(T value, CodedOutputStream stream);
}
```

and have singleton marshalers?

anuraaga (commented):
Initial high level comment

exporters/otlp/src/main/java/io/opentelemetry/exporter/otlp/AttributeMarshaler.java
anuraaga: Wilder would be to make everything static methods since we don't need to be generic, we only have a single API. But that's probably overboard.

bogdandrutu: The problem is that the size of the key needs a conversion to utf8 which is very expensive and allocating one object I measured to be better and do the conversion once. We can always improve more once we have the initial structure in place

bogdandrutu: I am looking also into a different approach that gogoproto uses which is to calculate size once and then write from the end to the buffer so to avoid the need of caching the size result per message. We have more ways to improve this and I will do a bit more on this during the holidays but still believe it is a good start

exporters/otlp/src/jmh/java/io/opentelemetry/exporter/otlp/RequestMarshalState.java
```diff
@@ -0,0 +1,108 @@
...
+                  AttributeKey.stringArrayKey("key_string_array"),
+                  Arrays.asList("string", "string"))
+              .put(AttributeKey.longArrayKey("key_long_array"), Arrays.asList(12L, 23L))
+              .put(AttributeKey.doubleArrayKey("key_double_array"), Arrays.asList(12.3, 23.1))
+              .put(AttributeKey.booleanArrayKey("key_boolean_array"), Arrays.asList(true, false))
+              .build());
+
+  private static final InstrumentationLibraryInfo INSTRUMENTATION_LIBRARY_INFO =
+      InstrumentationLibraryInfo.create("name", null);
+  private static final byte[] TRACE_ID_BYTES =
```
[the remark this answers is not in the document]
bogdandrutu: I will change them but for bytes in proto it shouldn’t matter (maybe just from hex to bytes which for comparison that we do is irrelevant both of them doing it

exporters/otlp/src/jmh/java/io/opentelemetry/exporter/otlp/RequestMarshalBenchmarks.java
bogdandrutu: Good call, will change

exporters/otlp/src/main/java/io/opentelemetry/exporter/otlp/AttributeMarshaler.java
anuraaga: Ok - also don't forget we have a fast concurrent weak map now, so it can often be used as an efficient cache when needed. If the attribute key is the main reason for the instantiation, then that'd be easy to cache and maybe a simple win.

anuraaga (approved):

exporters/otlp/src/jmh/java/io/opentelemetry/exporter/otlp/RequestMarshalBenchmarks.java
bogdandrutu: Done.

exporters/otlp/src/jmh/java/io/opentelemetry/exporter/otlp/RequestMarshalState.java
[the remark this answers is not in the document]
bogdandrutu: Done.

bogdandrutu:
After one more round of improvements (based on @anuraaga hint that I can use `byte[]`, I changed all the places where we wrapped the `byte[]` in `BytesString` and won a lot more performance, now we are 55% CPU usage win and 57% Memory. See the new benchmarks in the description.

exporters/otlp/src/main/java/io/opentelemetry/exporter/otlp/AttributeMarshaler.java
bogdandrutu: ???

PR #2306, opened 2020-12-15
Move OTLP adapters to otproto module for use outside of gRPC exporters.

anuraaga:
In #2295 we are thinking of using these outside the gRPC exporter artifacts. And we probably will want to provide HTTP exporters eventually as well, so we need these conversions to live outside. `otproto` seems designed for it based on the README.


jkwatson (approved):
Seems reasonable to me.

carlosalberto (approved):

anuraaga:
Thinking of merging this since we need the shared stuff (Resource) somewhere anyways to be able to split the exporter artifacts. Even if not going with #2295 I think it's still intuitive enough to have trace/metric-specific stuff here grouped with the rest of the conversion.

PR #2309, opened 2020-12-15
Make public src/main classes final where possible.

anuraaga:
Realized I forgot to make the top-level builders final, also did a scrub.


exporters/jaeger-thrift/src/test/java/io/opentelemetry/exporter/jaeger/thrift/JaegerThriftSpanExporterTest.java
```diff
@@ -235,14 +234,12 @@ void testExportMultipleResources() throws SenderException, UnknownHostException
   void configTest() {
     Map<String, String> options = new HashMap<>();
     String serviceName = "myGreatService";
-    String endpoint = "http://127.0.0.1:9090";
     options.put("otel.exporter.jaeger.service.name", serviceName);
-    options.put("otel.exporter.jaeger.endpoint", endpoint);
```
anuraaga, line 240: Don't think I'm too sad to lose this check

exporters/otlp/src/test/java/io/opentelemetry/exporter/otlp/OtlpGrpcSpanExporterTest.java
```diff
@@ -50,23 +91,36 @@
   @Test
   void configTest() {
     Map<String, String> options = new HashMap<>();
-    options.put("otel.exporter.otlp.span.timeout", "12");
-    options.put("otel.exporter.otlp.span.endpoint", "http://localhost:6553");
```
anuraaga, line 54: endpoint isn't allowed to start with http:// btw

exporters/jaeger-thrift/src/main/java/io/opentelemetry/exporter/jaeger/thrift/JaegerThriftSpanExporter.java
```diff
@@ -41,7 +41,8 @@
   private static final String IP_KEY = "ip";
   private static final String IP_DEFAULT = "0.0.0.0";
   private final ThriftSender thriftSender;
-  private final Process process;
+  // Visible for testing
+  final Process process;
```
bogdandrutu, line 45: I would expose a package protected getter for testing, to keep the logic that all members should be `private final` when possible.

exporters/jaeger/src/main/java/io/opentelemetry/exporter/jaeger/JaegerGrpcSpanExporter.java
```diff
@@ -49,8 +49,11 @@
   private static final String IP_KEY = "ip";
   private static final String IP_DEFAULT = "0.0.0.0";
   private final CollectorServiceGrpc.CollectorServiceFutureStub stub;
-  private final Model.Process.Builder processBuilder;
-  private final ManagedChannel managedChannel;
+
+  // Visible for testing
+  final Model.Process.Builder processBuilder;
+  // Visible for testing
+  final ManagedChannel managedChannel;
```
bogdandrutu, line 56: ditto

exporters/otlp/build.gradle
```diff
@@ -18,7 +18,9 @@ dependencies {
             libraries.protobuf,
             libraries.protobuf_util
 
-    testImplementation project(':opentelemetry-sdk-testing')
+    testImplementation project(':opentelemetry-sdk-testing'),
+            'com.linecorp.armeria:armeria-grpc',
+            'com.linecorp.armeria:armeria-junit5'
```
bogdandrutu, line 23: Why this change?

exporters/otlp/src/test/java/io/opentelemetry/exporter/otlp/OtlpGrpcSpanExporterTest.java
```diff
@@ -34,9 +42,42 @@
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import org.mockito.Mockito;
+import org.junit.jupiter.api.extension.RegisterExtension;
 
 class OtlpGrpcSpanExporterTest {
+
+  @RegisterExtension
+  public static ServerExtension server =
```
bogdandrutu, line 50: This are great test, but why adding them in a PR that says it does make things final?

exporters/otlp/build.gradle
jkwatson: yeah, feels like this snuck in to the wrong PR

exporters/otlp/src/test/java/io/opentelemetry/exporter/otlp/OtlpGrpcSpanExporterTest.java
anuraaga: Without the spy (can't because of final), we lose a significant test that the key-value splitting logic is correct (a regression test since it was broken for a while).

jkwatson (approved):

PR #2310, opened 2020-12-15
Implement a fallback service.name for all Resource instances

### LINKED ISSUE #2298: Always use a fallback service.name in the Resource, if none is provided.
With https://github.com/open-telemetry/opentelemetry-specification/pull/1269 merged, we need to implement it.

jkwatson:
resolves #2298 

Note: this ended up being way more complicated than I had guessed it would!

### DIFF
--- a/sdk-extensions/aws/src/test/java/io/opentelemetry/sdk/extension/aws/resource/LambdaResourceTest.java
+++ b/sdk-extensions/aws/src/test/java/io/opentelemetry/sdk/extension/aws/resource/LambdaResourceTest.java
@@ -8,7 +8,6 @@ package io.opentelemetry.sdk.extension.aws.resource;
 import static java.util.Collections.emptyMap;
 import static java.util.Collections.singletonMap;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.trace.attributes.SemanticAttributes;
@@ -28,7 +27,8 @@ class LambdaResourceTest {
     Attributes attributes = resource.create().getAttributes();
 
     // then
-    assertTrue(attributes.isEmpty());
+    // we get one from the fallback service.name
+    assertThat(attributes.asMap()).hasSize(1);
   }
 
   @Test
@@ -41,13 +41,14 @@ class LambdaResourceTest {
     Attributes attributes = resource.create().getAttributes();
 
     // then
-    assertThat(attributes)
-        .isEqualTo(
+    assertThat(attributes.asMap())
+        .containsAllEntriesOf(
             Attributes.of(
-                SemanticAttributes.CLOUD_PROVIDER,
-                "aws",
-                SemanticAttributes.FAAS_NAME,
-                "my-function"));
+                    SemanticAttributes.CLOUD_PROVIDER,
+                    "aws",
+                    SemanticAttributes.FAAS_NAME,
+                    "my-function")
+                .asMap());
   }
 
   @Test
@@ -64,17 +65,18 @@ class LambdaResourceTest {
     Attributes attributes = resource.create().getAttributes();
 
     // then
-    assertThat(attributes)
-        .isEqualTo(
+    assertThat(attributes.asMap())
+        .containsAllEntriesOf(
             Attributes.of(
-                SemanticAttributes.CLOUD_PROVIDER,
-                "aws",
-                SemanticAttributes.CLOUD_REGION,
-                "us-east-1",
-                SemanticAttributes.FAAS_NAME,
-                "my-function",
-                SemanticAttributes.FAAS_VERSION,
-                "1.2.3"));
+                    SemanticAttributes.CLOUD_PROVIDER,
+                    "aws",
+                    SemanticAttributes.CLOUD_REGION,
+                    "us-east-1",
+                    SemanticAttributes.FAAS_NAME,
+                    "my-function",
+                    SemanticAttributes.FAAS_VERSION,
+                    "1.2.3")
+                .asMap());
   }
 
   @Test
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
@@ -24,7 +24,7 @@ import javax.annotation.concurrent.Immutable;
 
 /**
  * {@link Resource} represents a resource, which capture identifying information about the entities
- * for which signals (stats or traces) are reported.
+ * for which signals (metrics, logs, or traces) are reported.
  */
 @Immutable
 @AutoValue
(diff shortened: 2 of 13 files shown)


sdk/common/src/test/java/io/opentelemetry/sdk/resources/ResourceTest.java
```diff
@@ -154,15 +167,15 @@ void testMergeResources() {
         Attributes.of(stringKey("a"), "1", stringKey("b"), "2", stringKey("c"), "4");
 
     Resource resource = DEFAULT_RESOURCE.merge(resource1).merge(resource2);
-    assertThat(resource.getAttributes()).isEqualTo(expectedAttributes);
+    assertThat(resource.getAttributes().asMap()).containsAllEntriesOf(expectedAttributes.asMap());
```
bogdandrutu, line 170: Why changing the previous call way of checking equality? Same question everywhere.

jkwatson: Mostly because if I'm on Windows, the fallback name is different than on other systems, and doing it as a map let's me compare that it at least has the right stuff in it.

carlosalberto (approved):

sdk/common/src/test/java/io/opentelemetry/sdk/resources/ResourceTest.java
jkwatson: the other approach, I guess, would be to expose the fallback value so tests could access it and put it into the expected. Not sure if that's better or worse, TBH.

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -36,11 +36,15 @@
...
       " should be a ASCII string with a length not exceed " + MAX_LENGTH + " characters.";
-  private static final Resource EMPTY = create(Attributes.empty());
 
   private static final Resource TELEMETRY_SDK;
+  private static final String FALLBACK_SERVICE_NAME;
 
   static {
+    String osName = System.getProperty("os.name");
+    boolean windows = osName != null && osName.toLowerCase().contains("windows");
+    FALLBACK_SERVICE_NAME = windows ? "unknown_service:java.exe" : "unknown_service:java";
```
anuraaga, line 46: Can we leave off `java`? The spec says if we don't have access to the executable name, we don't need to include it. Since we're basically taking a random guess here, I guess it means we don't have access to the executable name?

jkwatson: well, it's not *quite* a random guess. We know that the code is at least running on a JVM (graal native binaries aside for now), "java" is at least more descriptive than nothing. I'd be amenable to just hard-coding "unknown_service:java" so it  is at least known that the thing they've misconfigured is a java service. It would narrow down the search and at least eliminate non-java apps when trying to find which app was misconfigured.

jkwatson: (it would also make the testing easier to just have something hard-coded, so I'm not opposed to leaving windows users without their `.exe`)

anuraaga: The constant seems fine to me too, removing doesn't seem that bad function-wise since we have the language in the resource already, but either works

sdk-extensions/aws/src/test/java/io/opentelemetry/sdk/extension/aws/resource/LambdaResourceTest.java
```diff
@@ -41,13 +41,14 @@ void shouldAddNonEmptyAttributes() {
     Attributes attributes = resource.create().getAttributes();
 
     // then
-    assertThat(attributes)
-        .isEqualTo(
+    assertThat(attributes.asMap())
```
anuraaga, line 44: Do we need to make these changes? Since we're creating the resource here I guess it doesn't have service.name etc in it.

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
Oberon00: I think the main class name would make much more sense for java than the exe name. How precise/strict is the spec here? Would that be allowed?

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -155,8 +161,29 @@ public Resource merge(@Nullable Resource other) {
...
+          currentAttributes.toBuilder()
+              .put(ResourceAttributes.SERVICE_NAME, otherServiceName)
+              .build();
+    }
+
+    attrBuilder.putAll(currentAttributes);
+    return createResource(attrBuilder.build());
+  }
+
+  private static AutoValue_Resource createResource(Attributes attributes) {
```
Oberon00, line 179: I don't think we should add the resource attributes everywhere. The service name should be treated just the same as the telemetry.sdk.name, which is also (slated to be) "always" required.

Oberon00 (changes requested):

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
Oberon00: If I create a resource with certain attributes explicitly without merging the default resource, I would be very surprised to have any additional attributes. And it seems to make all the code more complicated.

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
anuraaga: The spec says executable so I read it as currently no, but it would be simple to add a clarification. It would probably bring all script languages into play, e.g., python main.py should have main.py but what about /foo/bar/main.py. Would we want the fully qualified name here? It seems possibly too long, though simple name seems too short.

Pretty complicated, I wish not setting service.name could be a startup exception 😨

sdk-extensions/aws/src/test/java/io/opentelemetry/sdk/extension/aws/resource/LambdaResourceTest.java
jkwatson: Funny thing...all Resources now have to have a service.name. Otherwise, how could you make sure there's always a fallback, no matter what?

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
jkwatson: The spec requires it now. I think you have a disagreement with the spec, not with the implementation?

Oberon00: This is not how I interpret the spec. I think it would be enough if every TracerProvider / Metric merged in a resource with the service name. But even that is quite a strict interpretation. I think it would be enough if it is in the default resource.

@carlosalberto You wrote this spec, WDYT?

Oberon00: I created https://github.com/open-telemetry/opentelemetry-specification/pull/1294

jkwatson: If you merged the default resource with one that contains the service.name, the new one would not get picked up, right?

Oberon00: Correct. Normally you would instead merge your custom resource with the default resource instead of the other way round (the left side is preferred, the right side is supplementary).

jkwatson:
I'm very happy to pause on this until https://github.com/open-telemetry/opentelemetry-specification/pull/1294 is resolved. I definitely agree that it would be better if we only added in the service.name when we assigned the resource to a provider.

Oberon00:
@jkwatson Feel free to leave your review at the spec PR.

jkwatson:
closing this, as we've gone another way.

PR #2312, opened 2020-12-15
Omit metrics artifacts (alpha) from the bill of materials (bom)

### LINKED ISSUE #2268: Remove the metrics modules from the BOM
We don't want non-stable signals to be in the otel java BOM. It will be required for users to manually add them as dependencies and keep the versions in sync themselves.

breedx-splk:
This resolves #2268.

### DIFF
--- a/bom/build.gradle
+++ b/bom/build.gradle
@@ -11,6 +11,7 @@ dependencies {
         parent.childProjects.sort { "$it.value" }
                 .collect { it.value }
                 .findAll { it.name != project.name }
+                .findAll { !it.name.endsWith('-metrics') }
                 .each { project ->
                     api project
                     println project


bom/build.gradle
```diff
@@ -11,6 +11,7 @@ dependencies {
         parent.childProjects.sort { "$it.value" }
                 .collect { it.value }
                 .findAll { it.name != project.name }
+                .findAll { !it.name.endsWith('-metrics') }
```
anuraaga, line 14: Does `it.version.endsWith('-alpha')` work?

jkwatson: maybe `.contains`, since SNAPSHOT always ends up last.

breedx-splk: It doesn't work.  It seems to be the case that the dependency resolution happens early enough that the version has not been customized in the submodule yet...or at least not consistently/reliably.

anuraaga (approved):

bom/build.gradle
anuraaga: Too bad - thanks for checking!

breedx-splk:
Since this was already covered by #2329 I am going to close this out.  Thanks for looking!

Issue #2272, opened 2020-12-10
Verify and update jaeger and zipkin exporters for error code handling.


jkwatson:
This PR in the specs updated how span errors should be handled in jaeger & zipkin:  https://github.com/open-telemetry/opentelemetry-specification/pull/1257

We should verify that our code is doing what it should and update it if it isn't.

ChrisJBurns:
I can get stuck into this one for a good second issue 👍🏼

ChrisJBurns:
I am assuming for this, I am more specifically looking at the `exporters/jaeger` `exporters/jaeger-thrift` and `exporters/zipkin` modules?

ChrisJBurns:
So, for brevity, clarity and future reference I'll condense all of the details on this issue with regards to my findings.

## Jaeger
The error handling has been detailed in the [`jaeger`](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk_exporters/jaeger.md#error-flag) exporter specification.
>When Span Status is set to ERROR an error tag MUST be added with the Boolean value of true. The added error tag MAY override any previous value.

### Jaeger-gRPC
This has been consistently implemented in the following [code](https://github.com/open-telemetry/opentelemetry-java/blob/91dc1191813a887caba594d483fe8718ce79d7fc/exporters/jaeger/src/main/java/io/opentelemetry/exporter/jaeger/Adapter.java#L125-L127):
```java
if (span.getStatus().getStatusCode() == StatusCode.ERROR) {
    target.addTags(toKeyValue(KEY_ERROR, true));
}
```

### Jaeger-Thrift
The same has been done in the `jaeger-thrift` `Adapter.java` [class](https://github.com/open-telemetry/opentelemetry-java/blob/91dc1191813a887caba594d483fe8718ce79d7fc/exporters/jaeger-thrift/src/main/java/io/opentelemetry/exporter/jaeger/thrift/Adapter.java#L123-L125):
```java
if (span.getStatus().getStatusCode() == StatusCode.ERROR) {
    tags.add(toTag(KEY_ERROR, true));
}
```

## Zipkin
[...]

PR #2314, opened 2020-12-16
Update Zipkin exporter error handling

ChrisJBurns:
The Zipkin exporter specification for error handling has slightly changed regarding the removal of the `otel.status_description` tag in favour for the error tag.

Due to this, we now move the `error` tag setting into the `!status.isUnset()` condition.

Resolves [2272](https://github.com/open-telemetry/opentelemetry-java/issues/2272)

### DIFF
--- a/exporters/zipkin/src/main/java/io/opentelemetry/exporter/zipkin/ZipkinSpanExporter.java
+++ b/exporters/zipkin/src/main/java/io/opentelemetry/exporter/zipkin/ZipkinSpanExporter.java
@@ -128,16 +128,15 @@ public final class ZipkinSpanExporter implements SpanExporter {
     spanAttributes.forEach(
         (key, value) -> spanBuilder.putTag(key.getKey(), valueToString(key, value)));
     SpanData.Status status = spanData.getStatus();
-    // include status code & description.
+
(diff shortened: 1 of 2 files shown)


exporters/zipkin/src/main/java/io/opentelemetry/exporter/zipkin/ZipkinSpanExporter.java
```diff
@@ -158,6 +157,10 @@ static Span generateSpan(SpanData spanData, Endpoint localEndpoint) {
     return spanBuilder.build();
   }
 
+  private static String getValueOrReturnEmpty(String value) {
```
anuraaga, line 160:
```suggestion
  private static String nullToEmpty(String value) {
```

Just small nit, we're not really "getting the value" from anything here

anuraaga (commented):
Thanks! Can you add a test with null description?

ChrisJBurns:
@anuraaga Great minds think alike. I noticed the missing null test just after I committed (just waiting for it to build).

anuraaga (approved):
Thanks!

jkwatson (approved):
Thanks!

PR #2315, opened 2020-12-16
Split trace/metrics OTLP exporters.

anuraaga:


exporters/otlp-metrics/build.gradle
```diff
@@ -0,0 +1,27 @@
+plugins {
+    id "java"
+    id "maven-publish"
+
+    id "ru.vyarus.animalsniffer"
+}
+
+description = 'OpenTelemetry Protocol Metrics Exporter'
+ext.moduleName = "io.opentelemetry.exporter.otlp.metrics"
```
jkwatson, line 9: do we want to put the `-alpha` on the version numbers?

anuraaga: Yeah added

exporters/otlp-metrics/build.gradle
```diff
@@ -0,0 +1,30 @@
...
+
+version = "${version}".replaceFirst(/^(\d+)\.(\d+).(\d+)/) { _, major, minor, patch ->
+    "${major}.${minor}.${patch}-alpha"
+}
```
jkwatson, line 13: heh. now that we have this in 3 places, we should ponder how to not repeat this. No need to change it in this PR, but something to definitely think about.

settings.gradle
```diff
@@ -39,7 +39,8 @@ include ":opentelemetry-all",
         ":opentelemetry-exporter-jaeger",
         ":opentelemetry-exporter-jaeger-thrift",
         ":opentelemetry-exporter-logging",
-        ":opentelemetry-exporter-otlp",
```
jkwatson, line 42: do we envision eventually having a module that ties the two (or maybe 3 with logs) together? Should we keep that module now for the 1.0 release so that people don't have to adjust their dependencies?

jkwatson:
I think we should keep the same top-level module, but split things out into 2 submodules that the top-level depends on. Then, people won't have to adjust their dependencies when we add metrics back in. Thoughts?

anuraaga:
@jkwatson I guess the top level would be alpha, since we don't want non-alpha depending on alpha right? Not sure if it's very useful if it's alpha though

anuraaga:
Ah wonder if you're thinking of having the top level only depend on trace for now. That makes sense, let me add it

jkwatson:
> Ah wonder if you're thinking of having the top level only depend on trace for now. That makes sense, let me add it

sorry..yeah, that's what I meant. I didn't write it very well, though.

jkwatson (approved):
The repackaging is an unfortunate but necessary breaking change (for the module system, I assume).  Thanks!

PR #2319, opened 2020-12-16
Only apply java-library to projects and configure reactively.

anuraaga:
I noticed a weird situation where we are applying `java` plugin in all projects individually, and then also applying `java` and `java-library` at top level. Since we're mostly libraries, I've switched the project configurations to `java-library`, except for testing-only projects which don't expose any API, removed the top level application, and instead respond to the plugin application as opposed to proactively configuring (this ensures it runs after `java-library` is applied to the project instead of before). 

Signing plugin configuration is also moved into the maven-publish configuration block where it makes more sense.

### DIFF
--- a/integration-tests/build.gradle
+++ b/integration-tests/build.gradle
@@ -1,5 +1,5 @@
 plugins {
-    id "java"
+    id "java-library"
     id "ru.vyarus.animalsniffer"
 }
 
--- a/integration-tests/build.gradle
+++ b/integration-tests/build.gradle
@@ -1,5 +1,5 @@
 plugins {
-    id "java-library"
+    id "java"
     id "ru.vyarus.animalsniffer"
 }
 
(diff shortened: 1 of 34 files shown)


integration-tests/build.gradle
```diff
@@ -14,13 +14,12 @@ task fatJar(type: Jar) {
...
 
 dependencies {
-    api project(':opentelemetry-api')
-
-    implementation project(':opentelemetry-sdk'),
-        project(':opentelemetry-exporter-jaeger'),
-        libraries.grpc_protobuf,
-        libraries.protobuf,
-        "io.grpc:grpc-netty-shaded:${grpcVersion}"
+    implementation project(':opentelemetry-api'),
```
jkwatson, line 17: curious why this is necessary. Doesn't the sdk depend on the api?

jkwatson (approved):
Thanks!

PR #2320, opened 2020-12-16
Have signal SDK depend on only signal API.

anuraaga:


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -66,8 +61,6 @@
 
   private static final String WORKER_THREAD_NAME =
       BatchSpanProcessor.class.getSimpleName() + "_WorkerThread";
-  private static final String SPAN_PROCESSOR_TYPE_LABEL = "spanProcessorType";
```
anuraaga, line 69: We'll need to decide whether we expose a `Listener` interface to be able to push these metrics via an extension. Presumably after this PR.

api/trace/src/main/java/io/opentelemetry/spi/trace/TracerProviderFactory.java
```diff
@@ -5,7 +5,6 @@
 
 package io.opentelemetry.spi.trace;
```
anuraaga, line 6: I moved this file into the `api-trace` folder for now. Later we'd presumably move all the SPI into an `autoconfigure` or such artifact based on #2262

bogdandrutu (changes requested):
Do not remove metrics from processors yet.

jkwatson:
I agree with Bogdan on this. I think we really do want those metrics in the processors and exporters; they are super valuable signals.

bogdandrutu:
> I agree with Bogdan on this. I think we really do want those metrics in the processors and exporters; they are super valuable signals.

One caveat the dependency I think can be "implementation" because we do not expose any API from the metrics publicly in this artifact

anuraaga:
I was thinking we'd need to expose a listener or such - how can we keep the metrics without forcing a user to use alpha APIs?

bogdandrutu:
> I was thinking we'd need to expose a listener or such - how can we keep the metrics without forcing a user to use alpha APIs?

Let's discuss that in a different PR :))

Also you are going to invent a new metrics API "listener", we must go deeper :))

anuraaga:
I was thinking that the next release seems like it will be quite breaking, so we would want to get as much breaking stuff in as possible. If we're splitting metrics out, ideally we split it everywhere by the next release so I made this PR. What "different PR" would be appropriate for discussiong? :P

anuraaga:
Got one idea for a different PR let me play with it.

anuraaga:
This looks like it got gobbled into #2341

Issue #2316, opened 2020-12-16
MetricExporter shutdown method is not consistent with SpanExporter shutdown method


trask:
MetricExporter shutdown returns void while SpanExporter shutdown method returns CompletableResultCode.

PR #2321, opened 2020-12-16
Make MetricExporter shutdown method returns CompletableResultCode.

dengliming:
closes [https://github.com/open-telemetry/opentelemetry-java/issues/2316](https://github.com/open-telemetry/opentelemetry-java/issues/2316)

### DIFF
--- a/exporters/otlp/src/main/java/io/opentelemetry/exporter/otlp/OtlpGrpcMetricExporter.java
+++ b/exporters/otlp/src/main/java/io/opentelemetry/exporter/otlp/OtlpGrpcMetricExporter.java
@@ -153,11 +153,12 @@ public final class OtlpGrpcMetricExporter implements MetricExporter {
    * cancelled. The channel is forcefully closed after a timeout.
    */
   @Override
-  public void shutdown() {
+  public CompletableResultCode shutdown() {
     try {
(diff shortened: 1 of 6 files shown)


bogdandrutu (approved):

exporters/otlp/src/main/java/io/opentelemetry/exporter/otlp/OtlpGrpcMetricExporter.java
```diff
@@ -153,11 +153,12 @@ public static OtlpGrpcMetricExporter getDefault() {
    * cancelled. The channel is forcefully closed after a timeout.
    */
   @Override
-  public void shutdown() {
+  public CompletableResultCode shutdown() {
     try {
       managedChannel.shutdown().awaitTermination(5, TimeUnit.SECONDS);
     } catch (InterruptedException e) {
       logger.log(Level.WARNING, "Failed to shutdown the gRPC channel", e);
```
jkwatson, line 160: should we return a failure in this case?

dengliming: yes. missing that. will fix it, thanks. : )

PR #2322, opened 2020-12-16
Fix small nits in method names and patterns we use in metrics SDK

bogdandrutu:


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/InstrumentSelector.java
```diff
@@ -30,45 +35,30 @@ public static Builder newBuilder() {
...
+   * null if none was specified.
    */
   @Nullable
-  @Memoized
-  public Pattern instrumentNamePattern() {
-    return instrumentNameRegex() == null ? null : Pattern.compile(instrumentNameRegex());
-  }
-
-  /** Returns whether the InstrumentType been specified. */
-  public boolean hasInstrumentType() {
```
jkwatson, line 53: I think it reads a little better to have this helper methods, rather than requiring callers to look for nulls.

bogdandrutu: Helper public methods means more to maintain and take care of, there is only one usage of this and I think it is worth the downside of reading to not have to maintain ABI compatibility.

anuraaga (approved):

jkwatson:
needs a rebase. Thanks!

jkwatson (approved):

PR #2323, opened 2020-12-16
Consistent naming of SDK implementation classes

bogdandrutu:

### DIFF
--- a/exporters/jaeger-thrift/src/test/java/io/opentelemetry/exporter/jaeger/thrift/JaegerThriftIntegrationTest.java
+++ b/exporters/jaeger-thrift/src/test/java/io/opentelemetry/exporter/jaeger/thrift/JaegerThriftIntegrationTest.java
@@ -10,7 +10,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import io.opentelemetry.api.GlobalOpenTelemetry;
 import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.api.trace.Tracer;
-import io.opentelemetry.sdk.OpenTelemetrySdk;
+import io.opentelemetry.sdk.SdkOpenTelemetry;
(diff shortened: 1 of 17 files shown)


jkwatson:
I think in the original rename PR, we had decided to keep this one the same. This is, in some sense, the SDK proper, and I so I don't think it makes sense to use "Sdk" as a modifier in this case. 

The other ones that were renamed are the "SDK"'s implementation of the interfaces, and thus the thing (eg. Tracer) is modified by "Sdk" making "SdkTracer".

bogdandrutu:
> The other ones that were renamed are the "SDK"'s implementation of the interfaces, and thus the thing (eg. Tracer) is modified by "Sdk" making "SdkTracer".

OpenTelemetrySdk is indeed an implementation of the OpenTelemetry interface correct?

iNikem:
I always felt that "SdkTracer" == "SDK's Tracer" == "Tracer of SDK". By the same logic "SDK of OpenTelemetry" == "OpenTelemtry's SDK" == "OpenTelemetrySdk".

bogdandrutu:
@iNikem it is very confusing to have half of the class names prefix with `Sdk` and the other half suffix with `Sdk`. Users will get very confused by this pattern, I am trying to offer a consistent experience, all SDK instances are prefixed with `Sdk`

iNikem:
@bogdandrutu there is ONE class with `Sdk` suffix. But I am not a maintainer of this repo :) As a user I find `SdkTracer` and `OpenTelemetrySdk` quite intuitive.

anuraaga:
Agree with @iNikem I'd lean towards having just this one class suffix with SDK, since it corresponds well with what we always call the "OpenTelemetry SDK".

jkwatson:
the majority seem ok with leaving this as-is, so I'm closing this PR.

PR #2327, opened 2020-12-16
Add note about autotracer name

nathanleclaire:


QUICKSTART.md
```diff
@@ -46,6 +46,8 @@ Tracer tracer =
     OpenTelemetry.getGlobalTracer("instrumentation-library-name", "1.0.0");
 ```
 
+*Note:* For the tracer generated by the Java agent, use `io.opentelemetry.auto` for the library name.
```
anuraaga, line 49: Hi @nathanleclaire! This would probably be more appropriate in the documentation for the java agent

https://github.com/open-telemetry/opentelemetry-java-instrumentation

But I'm wondering what you're attempting - normally we'd expect a user to use something to identify their app (even as simple as `app`) rather than reuse the name of the agent. Do you need access to this tracer specifically for something?

/cc @trask @iNikem

nathanleclaire: Hi @anuraaga thanks for the quick response. I can certainly add it there, but that documentation links here for info on `getGlobalTracer`.

> But I'm wondering what you're attempting - normally we'd expect a user to use something to identify their app (even as simple as app) rather than reuse the name of the agent. Do you need access to this tracer specifically for something?

We were looking to start tracing an app with the agent, then add more custom spans and attributes to those same traces. In our testing, the only way we could get this to work was by accessing the global tracer of the agent. We started out trying to get a custom tracer like you mention, but the result we were seeing was that would create spans in a separate trace.

Oberon00: This is a misunderstanding. All tracers share the same "trace tree"/active span/etc. So there is no need to fake a particular instrumenting library name to extend the spans/traces created by an auto-instrumenting agent. At least that is the spec perspective. Maybe the java agent has a bug / nonconformance here?

Also, ideally each instrumentation within the agent would use it's own instrumenting library name, so that they can be distinguished (e.g something like `io.opentelemetry.auto.jdbc`).

Oberon00 (changes requested):

QUICKSTART.md
jkwatson: If this doesn't work regardless of the tracer name then we have a gigantic bug that would need to be fixed ASAP (or else the agent does!).

nathanleclaire: So if I understand what you're saying correctly, I should be able to pass any name to the `getGlobalTracer` function when using the Java Agent and get back the correct tracer?

Oberon00: Basically yes, but "the correct tracer" does not exist. All tracers manipulate the same shared state. The only thing that depends on the name you pass to the tracer is the InstrumentationLibraryInfo that the exporter will receive as part of the span data.

jkwatson: note: I have recently heard confusion about this since it differs from OpenTracing. OpenTelemetry tracers are not the same thing as OpenTracing tracers. As @Oberon00 says...they are just a holder for the name/version information, and otherwise just use the same state as every other tracer created from a single TracerProvider (which is more akin to the OpenTracing Tracer concept).

nathanleclaire: Hmm. I think I get what you are saying but it is inconsistent with what we were seeing in practice, where a new global tracer would assume a totally new state and context. It's possible we missed something, but I will see if I can put up an example that reproduces the issue.

trask: hey @nathanleclaire, a repro would be great

puckpuck:
To add to this conversation, the docs should indicate if you are wanting to create your own custom span from the Global Tracer, then you should pass in your own name.  This would avoid some confusion on how to use the library. Perhaps even a sentence or two on what happens to that name (ie: becomes a span attribute).

jkwatson:
I'm going to close this, as it isn't correct. I've created an issue to improve the documentation (#2393).

PR #2311, opened 2020-12-15
inject alpha into the version for metrics api and sdk.

breedx-splk:
Resolves #2266.

After doing a `./gradlew clean build` the jars with alpha look like:

```
$ find . -name '*.jar' | grep alpha
./sdk/metrics/build/libs/opentelemetry-sdk-metrics-0.13.0-alpha-SNAPSHOT.jar
./sdk/metrics/build/libs/opentelemetry-sdk-metrics-0.13.0-alpha-SNAPSHOT-sources.jar
./sdk/metrics/build/libs/opentelemetry-sdk-metrics-0.13.0-alpha-SNAPSHOT-javadoc.jar
./api/metrics/build/libs/opentelemetry-api-metrics-0.13.0-alpha-SNAPSHOT-javadoc.jar
./api/metrics/build/libs/opentelemetry-api-metrics-0.13.0-alpha-SNAPSHOT-sources.jar
./api/metrics/build/libs/opentelemetry-api-metrics-0.13.0-alpha-SNAPSHOT.jar
```

PR #2329, opened 2020-12-17
Metrics are alpha take 2

breedx-splk:
This rolls back #2311 and reimplements #2328.

It does feel a bit hacky, but in my testing it does actually include `alpha` in the artifact names, poms, etc...and the bom change omits the metrics artifacts from the bom entirely.

### DIFF
--- a/api/metrics/build.gradle
+++ b/api/metrics/build.gradle
@@ -6,10 +6,6 @@ plugins {
     id "ru.vyarus.animalsniffer"
 }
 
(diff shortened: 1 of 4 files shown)


anuraaga (approved):
Thanks for digging, this approach seems sound to me.

bogdandrutu (approved):

jkwatson (approved):
let's give it a try!

PR #2330, opened 2020-12-17
Add setMeterProvider to BSP.

anuraaga:
I'm playing with the basics for https://github.com/open-telemetry/opentelemetry-java/pull/2262/files#diff-9c65892c78d75723d499d2f433695dfb0eaced043768a3803c1f2a39e201599aR375

Thinking about it led me to #2320 since it seemed a bit strange to be using the metrics API in our code while it's alpha, but I guess we can try the `@AlphaApi` annotation approach. Any thoughts?

### DIFF
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
@@ -140,8 +140,8 @@ public final class BatchSpanProcessor implements SpanProcessor {
 
   /**
    * Sets the {@link MeterProvider} to use to record metrics related to processed spans. By default,
-   * this processor will not record metrics. Call this method with a configured {@link
-   * MeterProvider} to enable recording.
+   * the {@link BatchSpanProcessor} will not record metrics. Call this method with a configured
+   * {@link MeterProvider} to enable recording.
    */
   @AlphaApi
   public void setMeterProvider(MeterProvider meterProvider) {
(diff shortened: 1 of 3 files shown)


anuraaga:
I guess another thought is during alpha stage we require setting up metrics before setting up tracing. Anyways trying to sort through what the trace/metrics split actually means for configuration.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -137,13 +139,21 @@ public CompletableResultCode forceFlush() {
     return worker.forceFlush();
   }
 
+  /**
+   * Sets the {@link MeterProvider} to use to record metrics related to processed spans. By default,
+   * the {@link BatchSpanProcessor} will record metrics using the {@link GlobalMetricsProvider}.
+   * Call this method with a configured {@link MeterProvider} to enable recording metrics if the
+   * {@link GlobalMetricsProvider} is not configured..
+   */
+  @AlphaApi
```
bogdandrutu, line 148: Why not just leave as is and only be possible to use metrics with SPI for the moment!?

anuraaga: Yeah that's my comment - it means metrics needs to be set up before tracing, which could be ok for now but want to confirm that with everyone.

bogdandrutu: I think while still alpha we are ok, we don’t want to expose this publicly in any stable api so I would not do what is proposed in this PR

anuraaga:
The way I'm looking at this is that during alpha, actually end users shouldn't be using metrics at all, even semi-adventorous ones. We could even say metrics are only for use from our SDK / instrumentation for now. So having some rough edges seems fine to me, thanks.

PR #2333, opened 2020-12-17
Upload coverage HTML report as action artifact.

anuraaga:
Lately codecov has been flaky for unknown reasons - it works well for many commits then stops working for many commits. This uploads the HTML report to GitHub Actions so at least there's something to look at.


Oberon00:
How long are these retained? Do we have some sort of storage quota?

anuraaga:
Default is 90 days though can be configured.

https://github.blog/changelog/2020-10-08-github-actions-ability-to-change-retention-days-for-artifacts-and-logs/

I think there is also a space limit but it also gets automatically purged of old things when it gets full. Can't find a number for the actual size limit but think it's in gigs.

jkwatson (approved):
Thanks!

PR #2334, opened 2020-12-17
Return a failure code while OtlpGrpcMetricExporter shutdown failed

dengliming:
missing return failure code in this case [https://github.com/open-telemetry/opentelemetry-java/pull/2321#discussion_r544478041](https://github.com/open-telemetry/opentelemetry-java/pull/2321#discussion_r544478041)


anuraaga (approved):
Thanks!

jkwatson:
Is it possible to write a unit test for this case?

anuraaga:
Tests for InterruptedException are pretty tough, especially for a gRPC call, not sure we can do it here

jkwatson (approved):

PR #2335, opened 2020-12-17
Exclude AutoValue generated classes from code coverage.

anuraaga:


Oberon00:
Do we really want to exclude these? If we never use a property in tests, perhaps it's an indicator that the property (or accessor) should be removed?

anuraaga:
They add a lot of missed branches for equals and hashcode. If we always use equals verifier it's another approach to reduce their impact on coverage. But I don't know if coverage is the right tools for finding unused properties and lean towards always excluding generated code.

jkwatson (approved):
Thanks! Requiring code coverage for generated classes would be silly.

PR #2342, opened 2020-12-17
Deprecate trace mutli propagator

### LINKED ISSUE #2244: TraceMultiPropagator vs. MultiTextMapPropagator

breedx-splk:
This resolves #2244...even if the implementation here doesn't perfectly match what's in that issue.  There were discussions on the previous PR (#2273) and we agreed to deprecate the extension class and remove it in the next release.  Since 0.13.0 went out today, I assumed we could remove it in 0.14.0 but maybe people think 0.15.0 is better?

In any case, I also brought some missing tests over, hopefully coverage goes up slightly.

### DIFF
--- a/api/context/src/test/java/io/opentelemetry/context/propagation/MultiTextMapPropagatorTest.java
+++ b/api/context/src/test/java/io/opentelemetry/context/propagation/MultiTextMapPropagatorTest.java
@@ -24,6 +24,7 @@ import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 
 @ExtendWith(MockitoExtension.class)
+@SuppressWarnings("deprecation")
 class MultiTextMapPropagatorTest {
 
   @Mock private TextMapPropagator propagator1;
--- a/api/context/src/test/java/io/opentelemetry/context/propagation/MultiTextMapPropagatorTest.java
+++ b/api/context/src/test/java/io/opentelemetry/context/propagation/MultiTextMapPropagatorTest.java
@@ -1,0 +1,152 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.context.propagation;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import io.opentelemetry.context.Context;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class MultiTextMapPropagatorTest {
+
+  @Mock private TextMapPropagator propagator1;
+  @Mock private TextMapPropagator propagator2;
+  @Mock private TextMapPropagator propagator3;
+
+  private static final TextMapPropagator.Getter<Map<String, String>> getter =
+      new TextMapPropagator.Getter<Map<String, String>>() {
+        @Override
+        public Iterable<String> keys(Map<String, String> carrier) {
+          return carrier.keySet();
+        }
+
+        @Nullable
+        @Override
+        public String get(Map<String, String> carrier, String key) {
+          return carrier.get(key);
+        }
+      };
+
+  @Test
+  void addPropagator_null() {
+    assertThrows(
+        NullPointerException.class,
+        () -> new MultiTextMapPropagator((List<TextMapPropagator>) null));
+  }
+
+  @Test
+  void fields() {
+    when(propagator1.fields()).thenReturn(Arrays.asList("foo", "bar"));
+    when(propagator2.fields()).thenReturn(Arrays.asList("hello", "world"));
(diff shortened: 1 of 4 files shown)


anuraaga (approved):

extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/TraceMultiPropagator.java
```diff
@@ -48,8 +48,12 @@
  * Context context = OpenTelemetry.getPropagators().getTextMapPropagator()
  *   .extract(context, carrier, carrierGetter);
  * }</pre>
+ *
+ * This class is deprecated and will be removed in 0.14.0. Users should migrate to the
```
anuraaga, line 52: Oops nit - `@deprecated -`

breedx-splk: I guess that makes it easier to find?  Anyway, added that `@`...

extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/TraceMultiPropagator.java
```diff
@@ -49,7 +49,7 @@
  *   .extract(context, carrier, carrierGetter);
  * }</pre>
  *
- * This class is deprecated and will be removed in 0.14.0. Users should migrate to the
+ * This class is @deprecated and will be removed in 0.14.0. Users should migrate to the
```
jkwatson, line 52: I think @anuraaga was asking to add a `@deprecated` tag to the javadoc itself, not embed the `@` in the comment.

extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/TraceMultiPropagator.java
```diff
@@ -49,7 +49,7 @@
...
  * MultiTextMapPropagator in the api.
  */
```
jkwatson, line 54:
```suggestion
 * @deprecated Migrate to the MultiTextPropagator in the API.
 */
```

anuraaga: Yeah sorry for lack of clarity. This format will render a bit better in IntelliJ I think

jkwatson (approved):
Thanks!

jkwatson:
looks like something is grumbling about using deprecated methods in the build. Probably tests, I'd reckon.

api/context/src/test/java/io/opentelemetry/context/propagation/MultiTextMapPropagatorTest.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class MultiTextMapPropagatorTest {
```
anuraaga, line 27:
```suggestion
@SuppressWarnings("deprecation")
class MultiTextMapPropagatorTest {
```

extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/TraceMultiPropagator.java
```diff
@@ -48,8 +48,13 @@
  * Context context = OpenTelemetry.getPropagators().getTextMapPropagator()
  *   .extract(context, carrier, carrierGetter);
  * }</pre>
+ *
+ * This class is @deprecated and will be removed in 0.14.0. Users should migrate to the
+ * MultiTextMapPropagator in the api.
+ * @deprecated Migrate to the MultiTextPropagator in the API.
```
anuraaga, line 54:
```suggestion
 * @deprecated Use {@link TextMapPropagator#composite}
```

anuraaga (approved):

api/context/src/test/java/io/opentelemetry/context/propagation/MultiTextMapPropagatorTest.java
```diff
@@ -0,0 +1,136 @@
...
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+@SuppressWarnings("deprecation")
```
jkwatson, line 27:
```suggestion
@SuppressWarnings("deprecation") // we're testing deprecated stuff in here.
```

jkwatson:
looks like spotless needs a little more help on one of the tests

PR #2344, opened 2020-12-17
Add default implementation for provider get

bogdandrutu:

### DIFF
--- a/api/trace/src/main/java/io/opentelemetry/api/trace/TracerProvider.java
+++ b/api/trace/src/main/java/io/opentelemetry/api/trace/TracerProvider.java
@@ -5,6 +5,7 @@
 
 package io.opentelemetry.api.trace;
 
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
@@ -31,7 +32,9 @@ public interface TracerProvider {
    *     instrument*ed* library (e.g., "io.opentelemetry.contrib.mongodb"). Must not be null.
    * @return a tracer instance.
    */
-  Tracer get(String instrumentationName);
+  default Tracer get(String instrumentationName) {
+    return get(instrumentationName, null);
+  }
 
   /**
    * Gets or creates a named and versioned tracer instance.
@@ -41,5 +44,5 @@ public interface TracerProvider {
    * @param instrumentationVersion The version of the instrumentation library (e.g., "1.0.0").
    * @return a tracer instance.
    */
-  Tracer get(String instrumentationName, String instrumentationVersion);
+  Tracer get(String instrumentationName, @Nullable String instrumentationVersion);
 }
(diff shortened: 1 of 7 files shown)


api/trace/src/main/java/io/opentelemetry/api/trace/TracerProvider.java
```diff
@@ -41,5 +44,5 @@ static TracerProvider getDefault() {
    * @param instrumentationVersion The version of the instrumentation library (e.g., "1.0.0").
    * @return a tracer instance.
    */
-  Tracer get(String instrumentationName, String instrumentationVersion);
+  Tracer get(String instrumentationName, @Nullable String instrumentationVersion);
```
anuraaga, line 47: I think we wanted to leave `@Nullable` off of this as the contract for our "public API". `null` is not a valid value for a version and shouldn't be passed, but we defensively support it anyways.

bogdandrutu: That is not what I see in the implementation of get with just name... we explicitly pass null so would be incorrect to not officially support null if we by default we call it with null

bogdandrutu: All implementation must implement null support otherwise they are not spec compliant

anuraaga: That's the implementation - but nullable is how we are presenting the API here. If null version has a spec-defined meaning than the annotation makes sense but I thought that was just a defensive check.

bogdandrutu: Then why by default we don't use `""` (empty string) for the other version with just the name? If we don't expect null?

bogdandrutu: > That's the implementation - but nullable is how we are presenting the API here.

This is wrong, because the `@Nullable` annotation tells alternative implementations that `null` has to be supported, it is not a defensive mechanism, we clearly call it will `null`.

bogdandrutu: My point is if we don’t add Nullable we cannot default implement the other one with calling with Null, which may be fine, we just drop this PR

jkwatson: We support the "null" case from the spec by providing an overloaded method that doesn't include the version at all. I think it's better to have this method, which explicitly has the version as a parameter, to not say that "null" is an ok value. If someone passes a `null` into it, we can't fail (and nor can any other spec compliant implementation) because we can't have bad inputs cause exceptions. 

From the user's perspective, if they have a null version, they should call the method without a version, which is how we support that use-case. By marking this as `@Nullable`, we're telling the user that it's totally ok to pass us a null, which is really not how we want people using the API.

bogdandrutu: Per offline discussion, we should close this PR.

PR #2346, opened 2020-12-17
Stop gap measure to fix api and sdk pom breakage dependency on metrics.

breedx-splk:
The gradle publishing hacks that put `-alpha` into the version string of the metrics modules missed the poms for the api and sdk modules, as seen here: https://bintray.com/beta/#/open-telemetry/maven/opentelemetry-java/0.13.0?tab=files

This PR is a stop-gap measure that makes the api and sdk modules depend directly on the metrics-api and metrics-sdk (respectively) at the `v0.13.0-alpha` version.  It's not great, but it should work around the current breakage.  This will end up being removed when the api and sdk no longer depend on the metrics modules at all.

### DIFF
--- a/api/all/build.gradle
+++ b/api/all/build.gradle
@@ -9,6 +9,15 @@ plugins {
 description = 'OpenTelemetry API'
 ext.moduleName = "io.opentelemetry.api"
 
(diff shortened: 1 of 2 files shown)


jkwatson (approved):
🤞

PR #2347, opened 2020-12-17
allow the list of shas to cherry-pick to be blank.

jkwatson:
if we already have the release branch in a good state, we don't need to add the list of shas.


jkwatson:
closing for now

Issue #2340, opened 2020-12-17
Flaky Test: IntervalMetricReaderTest [x1]


jkwatson:
This test fails occasionally with this error:

```
IntervalMetricReaderTest > intervalExport() FAILED
    java.lang.AssertionError: 
    Expecting:
      <[[MetricData{resource=Resource{attributes={}}, instrumentationLibraryInfo=InstrumentationLibraryInfo{name=IntervalMetricReaderTest, version=null}, name=my metric, description=my metric description, unit=us, type=DOUBLE_SUM, data=DoubleSumData{monotonic=true, aggregationTemporality=CUMULATIVE, points=[LongPoint{startEpochNanos=1000, epochNanos=3000, labels={}, value=1234567}]}}],
        [MetricData{resource=Resource{attributes={}}, instrumentationLibraryInfo=InstrumentationLibraryInfo{name=IntervalMetricReaderTest, version=null}, name=my metric, description=my metric description, unit=us, type=DOUBLE_SUM, data=DoubleSumData{monotonic=true, aggregationTemporality=CUMULATIVE, points=[LongPoint{startEpochNanos=1000, epochNanos=3000, labels={}, value=1234567}]}}],
        [MetricData{resource=Resource{attributes={}}, instrumentationLibraryInfo=InstrumentationLibraryInfo{name=IntervalMetricReaderTest, version=null}, name=my metric, description=my metric description, unit=us, type=DOUBLE_SUM, data=DoubleSumData{monotonic=true, aggregationTemporality=CUMULATIVE, points=[LongPoint{startEpochNanos=1000, epochNanos=3000, labels={}, value=1234567}]}}],
        [MetricData{resource=Resource{attributes={}}, instrumentationLibraryInfo=InstrumentationLibraryInfo{name=IntervalMetricReaderTest, version=null}, name=my metric, description=my metric description, unit=us, type=DOUBLE_SUM, data=DoubleSumData{monotonic=true, aggregationTemporality=CUMULATIVE, points=[LongPoint{startEpochNanos=1000, epochNanos=3000, labels={}, value=1234567}]}}]]>
    to contain exactly (and in same order):
[...]

PR #2352, opened 2020-12-18
Disable flaky interval test

anuraaga:
This is flaking a lot it seems, let's disable for now

Should be reverted as part of #2340


jkwatson:
@breedx-splk is working on fixing this test...should we wait for that before disabling?

anuraaga:
@jkwatson I'm seeing CI fail quite a bit. Not a big deal but I guess it's easy to add the revert for this into that change if it's going to be a bit more?

PR #2355, opened 2020-12-18
Use http helper class in EC2 Resource.

anuraaga:


trask (approved):

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource/Ec2Resource.java
```diff
@@ -69,70 +62,17 @@ public Ec2Resource() {
   // Generic HTTP fetch function for IMDS.
   // TODO(anuraaga): Migrate to JdkHttpClient
```
jkwatson, line 63: comment can now go away?

jkwatson (approved):
👍

PR #2356, opened 2020-12-18
Remove dependency on metrics from sdk:all

anuraaga:
Same as api:all

### DIFF
--- a/exporters/otlp/trace/build.gradle
+++ b/exporters/otlp/trace/build.gradle
@@ -10,7 +10,7 @@ description = 'OpenTelemetry Protocol Trace Exporter'
 ext.moduleName = "io.opentelemetry.exporter.otlp.trace"
 
 dependencies {
-    api project(':sdk:all')
+    api project(':sdk:trace')
 
     implementation project(':sdk-extensions:otproto'),
             libraries.grpc_api,
--- a/sdk/all/build.gradle
+++ b/sdk/all/build.gradle
@@ -24,7 +24,7 @@ dependencies {
     testImplementation project(':sdk:testing')
     testImplementation libraries.junit_pioneer
 
-    jmh(project(':sdk:metrics'))
+    jmh project(':sdk:metrics')
     jmh(project(':sdk:testing')) {
         // JMH doesn't handle dependencies that are duplicated between the main and jmh
         // configurations properly, but luckily here it's simple enough to just exclude transitive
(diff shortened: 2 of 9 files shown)


sdk/all/build.gradle
```diff
@@ -25,6 +24,7 @@ dependencies {
     testImplementation project(':sdk:testing')
     testImplementation libraries.junit_pioneer
 
+    jmh(project(':sdk:metrics'))
```
Oberon00, line 27:
```suggestion
    jmh project(':sdk:metrics')
```

sdk/testing/build.gradle
```diff
@@ -9,6 +9,7 @@ ext.moduleName = 'io.opentelemetry.sdk.testing'
 dependencies {
     api project(':api:all')
     api project(':sdk:all')
+    api project(':sdk:metrics')
```
Oberon00, line 12: If we have testing depend on metrics, we need to give it an alpha version number.

Oberon00 (approved):

sdk/testing/build.gradle
anuraaga: Yeah - presumably we have many extensions we'll need to give alpha too (logging for example). Let me try addressing that more generically outside this PR.

exporters/otlp/trace/build.gradle
```diff
@@ -10,7 +10,7 @@ description = 'OpenTelemetry Protocol Trace Exporter'
 ext.moduleName = "io.opentelemetry.exporter.otlp.trace"
 
 dependencies {
-    api project(':sdk:all')
+    api project(':sdk:trace')
```
bogdandrutu, line 13: This needs only implementation dependency not api correct?

bogdandrutu: Actually it could be runtime dependency in this case.

bogdandrutu:
Please use the right dependency type for all the cases.

sdk/testing/build.gradle
jkwatson: I don't think we have a requirement of testing artifact stability, do we? [not that I care too much if the testing artifact is versioned `alpha`, really]

exporters/otlp/trace/build.gradle
jkwatson: since the exporter implements an interface in the sdk:trace module, it effectively exposes that dependency as a part of it's API surface, so this has to be an `api` dependency, I believe.

jkwatson (approved):
Thanks!

exporters/otlp/trace/build.gradle
bogdandrutu: Sorry I read this metrics :))

PR #2357, opened 2020-12-18
Use public readProperties instead of protected fromConfigMap to avoid…

anuraaga:
… annoying pattern of initializing a subclass for testing.


jkwatson (approved):
Thanks!

PR #2359, opened 2020-12-18
Implement some simple builder implementation methods as default methods.

anuraaga:


api/common/src/main/java/io/opentelemetry/api/common/ArrayBackedAttributesBuilder.java
```diff
@@ -98,23 +50,35 @@ public AttributesBuilder putAll(Attributes attributes) {
     return this;
   }
 
-  private static List<Double> toList(double... values) {
+  @Nullable
+  static List<Double> toList(double... values) {
```
Oberon00, line 54: I think we should either mark the `double...` as `@Nullable` or drop the `@Nullable` from the return type annotation. Or we could actually leave the code as it was here, or return `Collections.emptyList()`.

jkwatson: yeah, we're already guarding all the calls so they don't pass in `null` here, so I think we can just return emptyList if someone figures out a way around it.

anuraaga: Agree it's hard enough to pass null so anything is fine including empty list.

anuraaga: Actually I went with null (and added `@Nullable`) to the parameter. Otherwise the repeated values would have different behavior on null vs the singular values, didn't see a good enough reason to do that.

Oberon00 (approved):
I would prefer removing nulls & nullable as much as possible, but the affected parts are package-private anyway.

jkwatson (approved):
Thanks!

PR #2361, opened 2020-12-18
Require type and name (default ".*") in the InstrumentSelector 

bogdandrutu:
Aggregation and temporality are very specific to every instrument type, so we need to explicitly ask users what instrument type they intend to match for the specific view.

This will simplify the logic of pre-computed what operation the processor needs to do in order to calculate the given temporality.

Temporary implement a logic of last added view wins.


jkwatson (approved):
This seems fine. I think there's a use-case for exporters to say "give me deltas for everything", but we can figure that out over time, or just provide a helper that will do that for all the types.

bogdandrutu:
@jkwatson give me delta for everything may be inefficient for observers, also delta does not apply to ValueObserver :)

bogdandrutu:
But I will address these corner cases in a follow up for the moment just focus on having type every time

PR #2363, opened 2020-12-18
Merge ViewRegistry and AggregationChooser 

bogdandrutu:
Depends on https://github.com/open-telemetry/opentelemetry-java/pull/2361


jkwatson:
What's the motivation for this change? I had pulled it out into its own class to provide some separation of responsibilities, and to keep the individual classes small.

bogdandrutu:
> What's the motivation for this change? I had pulled it out into its own class to provide some separation of responsibilities, and to keep the individual classes small.

I think the ViewManager right now was only a static method (from configuration to batcher(a.k.a processor)) so no reason to have it. I actually consider to move that logic later in the createProcessor.

jkwatson (approved):

PR #2366, opened 2020-12-18
Create a MANDATORY Resource instance containing mandatory attributes. Merge that into the default resource.

### LINKED ISSUE #2298: Always use a fallback service.name in the Resource, if none is provided.

jkwatson:
resolves #2298 in a less drastic way, which seems to be what the spec is being updated to require.

compare to #2310 

spec PR reference: https://github.com/open-telemetry/opentelemetry-specification/pull/1294


sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/spi/SdkMeterProviderFactoryTest.java
```diff
@@ -17,4 +20,26 @@
   void testDefault() {
     assertThat(GlobalMetricsProvider.get()).isInstanceOf(SdkMeterProvider.class);
   }
+
+  @Test
+  void builder_fallbackServiceName() {
+    Resource resourceWithFallbacks =
+        Resource.create(Attributes.of(ResourceAttributes.SERVICE_NAME, "unknown_service:java"));
```
anuraaga, line 27: Think it's good if at least one of the tests has another attribute in it.

sdk/trace/src/test/java/io/opentelemetry/sdk/trace/SdkTracerProviderTest.java
```diff
@@ -42,15 +44,44 @@ void setUp() {
...
-    assertThat(
-            SdkTracerProvider.builder()
-                .setClock(mock(Clock.class))
-                .setResource(mock(Resource.class))
-                .setIdGenerator(mock(IdGenerator.class))
-                .setTraceConfig(mock(TraceConfig.class))
-                .build())
-        .isNotNull();
+  void builder_fallbackServiceName() {
+    Resource resource = mock(Resource.class);
```
anuraaga, line 48: Do we need this mock?

anuraaga (approved):
Much cooler

sdk/trace/src/test/java/io/opentelemetry/sdk/trace/SdkTracerProviderTest.java
jkwatson: probably not. Can just use `Resource.getEmpty()`, for sure.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProviderBuilder.java
```diff
@@ -52,6 +54,9 @@ public SdkMeterProviderBuilder setResource(@Nonnull Resource resource) {
    * @return An initialized TracerSdkFactory.
    */
   public SdkMeterProvider build() {
-    return new SdkMeterProvider(clock, resource);
+    Resource resourceWithFallbacks =
+        Resource.create(
+            FALLBACK_MANDATORY_ATTRIBUTES.toBuilder().putAll(resource.getAttributes()).build());
```
Oberon00, line 59: This is how I initially proposed the spec PR, but not how I reworked it shortly after and how it got approved 😃 
Now the spec basically says that we should treat the service.name exactly the same as we already treat the telemetry.sdk.name: It's there by default, but if the user set's their own resource explicitly, they need to merge it in manually.

See https://github.com/open-telemetry/opentelemetry-specification/pull/1294#issuecomment-747589223 and https://github.com/open-telemetry/opentelemetry-specification/blob/39fa82f0a7e58975782520c304768d63da0c7612/specification/resource/sdk.md (not yet merged).

If you think the first version of the PR was better, you should say so on the spec. 

@anuraaga @jkwatson You approved the spec PR after I made this change, but maybe you want to check again?

Oberon00 (changes requested):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProviderBuilder.java
jkwatson: I removed my approval, because I was mostly going on the PR description, which seems to match what I've implemented here. What's now written in the spec is no longer comprehensible to me. :(

Oberon00: @jkwatson To really dismiss your approval you have to expand the box at the bottom of the PR and click the dots next to your approval there (or add a request-changes review). Adding a comment-only review does not remove the approval.

jkwatson:
@Oberon00 Do you think this PR is now inline with the current state of your in-flight spec PR?  https://github.com/open-telemetry/opentelemetry-specification/pull/1294

jkwatson:
this is dependent on #2542 to get the merging to happen in the right order

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -52,9 +53,15 @@
...
       " should be a ASCII string with a length not exceed " + MAX_LENGTH + " characters.";
   private static final Resource EMPTY = create(Attributes.empty());
-
   private static final Resource TELEMETRY_SDK;
 
+  /**
+   * The MANDATORY Resource instance contains the mandatory attributes that must be used if they are
+   * not provided by the Resource that is given to an SDK signal provider.
+   */
+  public static final Resource MANDATORY =
```
Oberon00, line 62: I don't think this needs to be separated from DEFAULT. Also, this would be the only `public` field in this class, the others are `private` and have getters.

Oberon00 (approved):

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
jkwatson: I like having it be a separate thing in the code, but agree it should be private, and can be accessed via the default Resource.

jkwatson: done

PR #2367, opened 2020-12-18
remove some deprecated SpanData methods

jkwatson:


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/data/SpanData.java
```diff
@@ -228,17 +215,6 @@ static Link create(SpanContext spanContext, Attributes attributes, int totalAttr
...
     SpanContext getSpanContext();
 
-    /**
-     * Returns the {@code SpanContext}.
-     *
-     * @return the {@code SpanContext}.
-     * @deprecated Use {@link #getSpanContext()}
-     */
-    @Deprecated
-    default SpanContext getContext() {
```
bogdandrutu, line 238: Should we deprecate the other accessors for ids in favor of this one?

It has already helpers for bytes for ids and we can remove the helpers to convert.

jkwatson: This one was deprecated in favor of `getSpanContext()`, so I think we want to delete this one, either way.

bogdandrutu: Good point :)

bogdandrutu (approved):

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/data/SpanData.java
bogdandrutu: This is in the link not in the SpanData :)

PR #2368, opened 2020-12-21
Avoid spying the class under test.

anuraaga:


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
```diff
@@ -159,6 +159,11 @@ protected Builder fromConfigMap(
       }
       return this;
     }
+
+    // Visible for testing
+    long getExportIntervalMillis() {
```
Oberon00, line 164: I like to use a prefix such as `test_` for methods that are not only visible for testing but exist solely for testing.

anuraaga: I don't think I've ever done that but don't mind it either. Will wait for another vote and rename.

Oberon00: No strong opinion here either. I just think it would be nice to mark the method as "exists only for testing" vs "visibility extended for testing".

exporters/otlp/metrics/src/test/java/io/opentelemetry/exporter/otlp/metrics/OtlpGrpcMetricExporterTest.java
```diff
@@ -70,11 +68,9 @@ void configTest() {
     Properties options = new Properties();
     options.put("otel.exporter.otlp.metric.timeout", "12");
     options.put("otel.exporter.otlp.insecure", "true");
-    OtlpGrpcMetricExporterBuilder config = OtlpGrpcMetricExporter.builder();
-    OtlpGrpcMetricExporterBuilder spy = Mockito.spy(config);
-    spy.readProperties(options);
-    verify(spy).setDeadlineMs(12);
-    verify(spy).setUseTls(false);
+    OtlpGrpcMetricExporterBuilder config = OtlpGrpcMetricExporter.builder().readProperties(options);
+    assertThat(config.getUseTls()).isFalse();
```
jkwatson, line 72: rather than exposing a method, we can dig into the internals and keep the production code a bit cleaner. I can't say which is better or worse, but at least if we just dig into the internals, we've reduced a couple lines of production code and a discussion about naming those methods. ;)

anuraaga: Nice

jkwatson (approved):
Thanks!

PR #2369, opened 2020-12-21
Merge non-metrics API into API artifact.

anuraaga:
Not sure, but realized this may be closer to what we want. Less artifacts published and less chance of users depending on e.g., `api-trace`, which we presumably don't want them to do. And when metrics is ready, we copy the code into the api artifact and delete the metrics artifact from our repository.

I didn't do the same for `exporters/otlp` since that is an SDK-related artifact. I guess even long term we want that to be splittable?


Oberon00:
> less chance of users depending on e.g., api-trace, which we presumably don't want them to do

I thought enabling this was the whole reason why the API was split? Was it only to separate out the metrics?

anuraaga:
I thought it was only to allow separating metrics but I may have misunderstood :)

Oberon00:
I'm not sure either. From a versioning & maintainability perspective, assuming we want the same version number for all the api components, it would be easier to have only one api artifact, to restrict mixing & matching of different versions.

Oberon00:
For reference, the split was done in #2199 by @bogdandrutu. And @jkwatson did comment in https://github.com/open-telemetry/opentelemetry-java/pull/2199#issuecomment-739005313:

> I think this is a good change, to **make users more able to mix & match pieces of the APIs** that they wish to use.

I think we are not all on the same page here yet 😃

Oberon00:
I notice we also list the APIs separately in the README, but we have already done so before they were actually separate for some reason.

jkwatson:
I think my original thought process was: if someone want to use metrics before they are stable, then they don't have to do anything except change the version number they depend on when metrics does go stable. And, if this is the pattern we like, we should be consistent across the signals.

Otherwise, the metrics API module will completely disappear when we make it stable, which feels like an unexpected and potentially confusing change to the end users.

anuraaga:
Since metrics are alpha, and there will be some UX issues even besides breaking changes like https://github.com/open-telemetry/opentelemetry-java/pull/2330#issuecomment-747809958 I think a POM update to remove a dependency instead of bumping the version is fairly lightweight and reasonable to expect for an alpha module I think.

I'd say a decision of splitting the API module proactively and long-term should be about whether we want to enable the ability for users to depend on only parts of the API. It seems ok to me, but also increases the number of artifacts and I don't have a strong preference for either. But I guess the metrics split by itself I wouldn't expect to be a reason for the split.

jkwatson:
I'm fine with this approach, if we're all on board with it. My main goal is just to get to stability so we can cut 1.0. :)

api/all/build.gradle
```diff
@@ -11,10 +11,7 @@ ext.moduleName = "io.opentelemetry.api"
 archivesBaseName = "opentelemetry-api"
 
 dependencies {
-    api project(':api:context'),
-            project(':api:baggage'),
-            project(':api:common'),
-            project(':api:trace')
+    api project(':api:context')
```
jkwatson, line 14: ok... just thinking here...  if `:api` is going to contain baggage and traces, why not context?

api/all/src/test/java/io/opentelemetry/api/trace/propagation/W3CTraceContextPropagatorTest.java
```diff
@@ -450,8 +449,8 @@ void fieldsList() {
 
   @Test
   void headerNames() {
-    Assertions.assertThat(W3CTraceContextPropagator.TRACE_PARENT).isEqualTo("traceparent");
-    Assertions.assertThat(W3CTraceContextPropagator.TRACE_STATE).isEqualTo("tracestate");
+    assertThat(W3CTraceContextPropagator.TRACE_PARENT).isEqualTo("traceparent");
```
jkwatson, line 452: not sure why this changed leaked into this PR, but I also don't care too much, because it's just in a test and a good change.

jkwatson:
@bogdandrutu do you have an opinion here? Our options are:

1) Separate API artifact for each signal (trace, baggage, context, metrics, logs) and common, and a dependency-only artifact that depends on the stable bits.
  - pros: everything is well modularized, and users can opt to pick-and-choose which micro-pieces of the API they care about
  - cons: this is a lot of artifacts, and definitely is more complex setup
2) One API artifact for everything stable, which we move things into as they become stable
  - pros: very simple, not much to think about for an end user
  - cons: you get everything, whether you want it or not (maybe a few extra kb of dependency, I guess)
3) One API artifact for everything stable except the context, which we move things into as they become stable (this PR).
  - pros: see above for options 2
  - cons: why is context off all by itself?

Oberon00:
@jkwatson 

> why is context off all by itself?

Wasn't there a vision somewhere that the context could be used independently of telemetry? But maybe we don't want to encourage that anyway?

If in doubt it is probably safer to put it in the API package too. Separating it out (with an `api` dependency) is probably less breaking than not providing a separate artifact anymore if we change our mind.

anuraaga:
If we merged context into the API, I'd very much like to add `getSpan` `getBaggage` to `Context`. Guessing that's the main reason not to merge it? :-)

Oberon00:
>  I'd very much like to add getSpan getBaggage to Context

At that point we can radically simplify our implementation from `Object[] items` and key allocation to just `Baggage currentBaggage; Span activeSpan;` too. At which point we don't really have a Context anymore.

I wouldn't like having another class that statically depends on all different signals. With Context it would even be a cyclic dependency. Even within a single artifact, I like having loose coupling where easily possible.

anuraaga:
Yup - so I think keeping it separate makes sense :)

jkwatson (approved):
👍  thanks!

jkwatson:
this needs a (probably painful) rebase.

PR #2372, opened 2020-12-21
Restore accidentally deleted :all project

anuraaga:


settings.gradle
```diff
@@ -26,7 +26,8 @@ plugins {
 }
 
 rootProject.name = "opentelemetry-java"
-include ":api:all",
+include ":all",
```
Oberon00, line 29: What is the purpose of the `all` project anyway?

Oberon00: Oh, I see it has this archunit test and the coverage report collection... Maybe adding a short README to the subfolder would be helpful? I think it's not that obvious.

Oberon00 (approved):

settings.gradle
anuraaga: Sure

PR #2373, opened 2020-12-21
Remove most addspanprocessor mutator invocations.

anuraaga:
What's left is ZPages which will be a small API change there so will do it separately.


opencensus-shim/src/test/java/io/opentelemetry/opencensusshim/TraceInteroperabilityTest.java
```diff
@@ -56,20 +59,32 @@
...
+  private static final OpenTelemetry openTelemetry;
 
-  @BeforeEach
-  public void init() {
+  static {
+    spanExporter = spy(SpanExporter.class);
     when(spanExporter.export(anyList())).thenReturn(CompletableResultCode.ofSuccess());
+
     SpanProcessor spanProcessor = SimpleSpanProcessor.builder(spanExporter).build();
-    OpenTelemetrySdk.getGlobalTracerManagement().addSpanProcessor(spanProcessor);
```
anuraaga, line 67: span processor leak fixed :)

sdk/testing/src/test/java/io/opentelemetry/sdk/testing/exporter/InMemorySpanExporterTest.java
```diff
@@ -23,13 +23,17 @@
...
-  private final Tracer tracer = sdkTracerProvider.get("InMemorySpanExporterTest");
   private final InMemorySpanExporter exporter = InMemorySpanExporter.create();
 
+  private Tracer tracer;
+
   @BeforeEach
   void setup() {
-    sdkTracerProvider.addSpanProcessor(SimpleSpanProcessor.builder(exporter).build());
+    tracer =
+        SdkTracerProvider.builder()
```
bogdandrutu, line 33: I don't think Simple processor needs this, but it is a good pattern to follow and shutdown the ask all the time.

integration-tests/src/main/java/io/opentelemetry/SendTraceToJaeger.java
```diff
@@ -78,8 +83,8 @@ public static void main(String[] args) {
 
     // Start the example
     SendTraceToJaeger example = new SendTraceToJaeger(ip, port);
-    example.setupJaegerExporter();
-    example.myWonderfulUseCase();
+    OpenTelemetry openTelemetry = example.initOpenTelemetry();
+    SendTraceToJaeger.myWonderfulUseCase(openTelemetry);
```
jkwatson, line 87: it's a little odd to use this class name to call a static method on itself, but it's not a deal breaker.

jkwatson (approved):
Thanks. A couple of tiny tweaks would be fine, but nothing blocking.

integration-tests/src/main/java/io/opentelemetry/SendTraceToJaeger.java
bogdandrutu: you mean just write this as `myWonderfulUseCase(openTelemetry);` ?

anuraaga: Error prone doesn't like

bogdandrutu: Can you paste the error? I am curious

anuraaga: Sorry my bad - error prone didn't like example.static(), but static() no problem. Error message earlier threw me off

```
/Users/aanuraag/git/opentelemetry-java/integration-tests/src/main/java/io/opentelemetry/SendTraceToJaeger.java:87: warning: [static] static method should be qualified by type name, SendTraceToJaeger, instead of by an expression
    example.myWonderfulUseCase(openTelemetry);
           ^
```

PR #2374, opened 2020-12-21
Add SimpleSpanProcessor.create and deprecate SimpleSpanProcessorBuilder

anuraaga:
The spec doesn't have a `sampled` parameter

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#simple-processor

So I think we should remove the builder at least for now, if anyone needs the sampled parameter we can add to spec first. I'd expect no though since the job of the simple processor is to be simple.


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/SimpleSpanProcessor.java
```diff
@@ -48,19 +49,30 @@
   private final boolean sampled;
   private final AtomicBoolean isShutdown = new AtomicBoolean(false);
 
+  /**
+   * Returns a new {@link SimpleSpanProcessor} which exports spans to the {@link SpanExporter}
+   * synchronously.
+   */
+  public static SpanProcessor create(SpanExporter exporter) {
```
Oberon00, line 56: Maybe the PR title should be "Add SimpleSpanProcessor.create()"?

Oberon00, line 56: I have no opinion on this point, but often we hid such classes in the past and instead used a static method in the interface. So what about making this class deprecated and later package-private and instead adding a `SpanProcessor.simple(Exporter)` method? (or call it `wrap`)

anuraaga: I tend to prefer the pattern but I don't know if we want the `export` package to leak into the top level. It feels a bit weird to me though maybe it's not such a big deal.

A different option with better grouping would be for `SpanExporter` to expose the factory. But also weird. Default methods `asSimpleSpanProcessor` etc seems better.

Another idea in my head is to make this private completely and only have `SdkTracerProvider.addSpanExporter(exporter, BatchSettings.noBatching())`.

Not sure if it works but is also a reason I wanted to simplify this builder away :)

Oberon00: I think we should come to a (potentially) final decision here before removing/deprecating any public stuff. The builder is not that big a burden to maintain in the meantime. If we decide to leave the SimpleSpanProcessor for the foreseeable future then this PR is fine, otherwise I'd prefer we leave the builder as-is for now.

Oberon00: I think `Exporter.asSimpleSpanProcessor` would be fine. Otherwise, if we somehow want to keep SpanProcessor/Exporter statically decoupled (so far the interfaces don't know about each other), we could add an utility class like `ExportingSpanProcessor` with a static method `simple` (and `batching`).

bogdandrutu (commented):
The reason to use builder was to be consistent with other processors and to allow easily extension (which can also b achieved with create). I think I prefer consistency between processors.

jkwatson:
My opinion (not incredibly strongly held) is that we should keep the creational patterns consistent across the various bits, so even if the SSP builder seems like a bit of overkill, at least it won't surprise people that the creational patterns are different between the SSP and BSP. The "principle of least surprise" would seem to dictate consistency to me.

We can always explore additional creational patterns in the future, but I wouldn't worry about this yet.

anuraaga:
rIs the reason to have the unspec'd `sampled` parameter to have an excuse for a builder? I don't think we would have a builder if there were no parameters at all, that seems weirder. A processor with no options not having a builder, sure it's not consistent with other processors, but is consistent with other creatable objects in general I think.

bogdandrutu:
@anuraaga nobody said this "rIs the reason to have the unspec'd sampled parameter to have an excuse for a builder?". This is your interpretation which is wrong, and I think @jkwatson and I are not complaining about removing that option, but we are complaining about changing to factory method.

anuraaga:
@bogdandrutu "rIs the reason" has a typo, it was supposed to be "Is the reason" and just a normal question. Since otherwise, we have a completely empty builder. I've never seen that before, are you sure we want an empty builder here?

bogdandrutu:
Not sure if empty builder is something that bad... But just to clarify having or not an unspeced parameter is a different problem. I think long time ago we were asked to support that because the specs says if I remember correctly that all spans with isRecording true will be passed. I think specs needs clarification. The question is if specs adds this do we want create or Builder?

anuraaga:
@bogdandrutu I would have the `create` factory until a builder is needed, similar to some other places we've done that like propagators. I think it's reasonable for `BatchSpanProcessor.create` to exist which uses the default settings and would add the symmetry. I don't see `BatchSpanProcessorBuilder`'s existence requiring a `SimpleSpanProcessorBuilder` - some implementations of interfaces are complex and have settings, some don't. No builder makes obvious the fact that an interface has no settings and provides a lot of information in a self-documenting way. Once we have options, we have builder.

Oberon00:
@bogdandrutu 

> because the specs says if I remember correctly that all spans with isRecording true will be passed. I think specs needs clarification.

The spec was already clarified some time ago here (https://github.com/open-telemetry/opentelemetry-specification/blob/v0.7.0/specification/trace/sdk.md#sampling):

> Span Processor MUST receive only those spans which have this
  field \[IsRecording] set to `true`. However, Span Exporter SHOULD NOT
  receive them unless the `Sampled` flag was also set.
> ...
> Span Exporters MUST
  receive those spans which have `Sampled` flag set to true and they SHOULD NOT receive the ones
  that do not.

jkwatson (approved):
Since we shouldn't be providing the option of sampled or not here, I approve this PR. There's no need for an empty builder for something as simple as this processor. We can add a builder if at some point in the future we get more features, and we can keep the `create` method with the default without causing pain.

opencensus-shim/src/test/java/io/opentelemetry/opencensusshim/TraceInteroperabilityTest.java
```diff
@@ -66,8 +66,20 @@
...
+            .setTracerProvider(SdkTracerProvider.builder().addSpanProcessor(spanProcessor).build())
+            .build();
+    GlobalOpenTelemetry.set(openTelemetry);
+  }
+
+  @Captor private ArgumentCaptor<Collection<SpanData>> spanDataCaptor;
+
+  @BeforeEach
+  void resetMocks() {
+    reset(spanExporter);
```
jkwatson: this might be a very rare case where I would use the full `Mockito.reset` since I had no idea where this method was coming from without looking at imports.

anuraaga: Commenting on the merge commit :)

Not going to send a PR to rename it but if it feels any better, hopefully the method name `resetMocks` helps.

PR #2375, opened 2020-12-21
Migrate assertThrows -> assertThatThrownBy

anuraaga:
This makes all the assertions consistently assertj. It also restores asserting on the exception message. During JUnit5 migration, I misinterpreted the `message` parameter to `assertThrows` as a check on the message, but it isn't - the correct way would have been to assert on the returned `Throwable` instead.

This PR is 1) Structural replace 2) Fix failing tests due to drift in exception message assertions mostly by changing `hasMessage` to `hasMessageContaining`

### DIFF
--- a/api/all/src/test/java/io/opentelemetry/api/OpenTelemetryTest.java
+++ b/api/all/src/test/java/io/opentelemetry/api/OpenTelemetryTest.java
@@ -6,7 +6,7 @@
 package io.opentelemetry.api;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
@@ -101,7 +101,8 @@ class OpenTelemetryTest {
   @Test
   void testTracerNotFound() {
     System.setProperty(TracerProviderFactory.class.getName(), "io.does.not.exists");
-    assertThrows(IllegalStateException.class, () -> GlobalOpenTelemetry.getTracer("testTracer"));
+    assertThatThrownBy(() -> GlobalOpenTelemetry.getTracer("testTracer"))
+        .isInstanceOf(IllegalStateException.class);
   }
 
   @Test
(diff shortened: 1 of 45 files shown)


Oberon00:
Oh, I only knew `assertThatExceptionOfType().isThrownBy`.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkTracerProviderBuilder.java
```diff
@@ -5,13 +5,14 @@
 
 package io.opentelemetry.sdk.trace;
 
+import static java.util.Objects.requireNonNull;
```
Oberon00, line 8: Intentional change?

Oberon00 (approved):

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkTracerProviderBuilder.java
anuraaga: I needed to add `"traceConfig"` to the requireNonNull check for it and used my normal static import style. If it was 100 lines of drift I'd probably revert it but hoping just three is ok :)

api/all/src/test/java/io/opentelemetry/api/OpenTelemetryTest.java
```diff
@@ -6,7 +6,7 @@
 package io.opentelemetry.api;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
```
bogdandrutu, line 9: Is there no equivalent in jupiter? I liked the fact that we always use the jupiter imports.

bogdandrutu (approved):

api/all/src/test/java/io/opentelemetry/api/OpenTelemetryTest.java
Oberon00: The jupiter equivalent would be using `assertThat()` around the return value of the `asserThrows` which is the exception. But we use asserj for all other assertions, so I think either way is equally consistent.

PR #2376, opened 2020-12-21
Return SpanProcessor from zpages instead of automatically mutating gl…

anuraaga:
…obal.


jkwatson (approved):
This is a breaking change to this module, correct?

bogdandrutu (approved):

anuraaga:
@jkwatson Yeah guess it is

PR #2379, opened 2020-12-21
Add a notion of Accumulation as the value passed between Accumulator …

bogdandrutu:
…and Processor

Partially implement the logic described in https://github.com/open-telemetry/opentelemetry-java/discussions/2378. Still more cleanup required but at this point this is good enough to be the first step.


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregation/Aggregations.java
```diff
@@ -186,15 +215,21 @@ public AggregatorFactory getAggregatorFactory(InstrumentValueType instrumentValu
           : DoubleLastValueAggregator.getFactory();
     }
 
+    @Override
+    public Accumulation merge(Accumulation a1, Accumulation a2) {
+      // TODO: Define the order between accumulation.
```
jkwatson, line 220: Do we need a spec issue for this, or just a Java issue?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/AbstractAggregator.java
```diff
@@ -13,24 +16,17 @@
...
-   * {@code Aggregator}.
-   *
-   * <p>If this method is called, you can assume that the passed in aggregator can be cast to your
-   * self-type.
-   *
-   * @param aggregator The aggregator to merge with.
-   */
-  abstract void doMergeAndReset(Aggregator aggregator);
+  /** Implementation of the {@code toAccumulationThenReset}. */
+  abstract Accumulation doToAccumulationThenReset();
```
jkwatson, line 29: urgh. this method name is getting ugly. we should think up a better name at some point.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/Aggregator.java
```diff
@@ -15,23 +14,11 @@
...
-   * Returns the {@code Point} with the given properties and the value from this Aggregation.
-   *
-   * @param startEpochNanos the startEpochNanos for the {@code Point}.
-   * @param epochNanos the epochNanos for the {@code Point}.
-   * @param labels the labels for the {@code Point}.
-   * @return the {@code Point} with the value from this Aggregation.
    */
   @Nullable
-  Point toPoint(long startEpochNanos, long epochNanos, Labels labels);
+  Accumulation toAccumulationThenReset();
```
jkwatson, line 21: maybe `accumulateAndReset()` ?

jkwatson (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregation/Aggregations.java
bogdandrutu: I think right now go uses a timestamp as well associated with the entry. I left a TODO because I intend to look into this soon.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/Aggregator.java
bogdandrutu: I like `accumulate`, but I would use ThenReset because it is consistent with java stream APIs.

PR #2380, opened 2020-12-21
Ensure that the global Otel is set once and before any usage.

bogdandrutu:

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
+++ b/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
@@ -10,7 +10,7 @@ import io.opentelemetry.api.trace.TracerProvider;
 import io.opentelemetry.context.propagation.ContextPropagators;
 import io.opentelemetry.spi.OpenTelemetryFactory;
 import io.opentelemetry.spi.trace.TracerProviderFactory;
-import javax.annotation.Nullable;
+import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * A global singleton for the entrypoint to telemetry functionality for tracing, metrics and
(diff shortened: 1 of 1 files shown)


bogdandrutu:
@jkwatson @anuraaga this PR must pass all tests before we can declare that the global initialization ordering problem is solved.

jkwatson:
I'm not sure I understand what you mean. We have tests that call `set` right now, so of course they're going to fail if you make it only settable once, and have the 2nd time throw an exception.

anuraaga:
I think at last week's meeting we discussed making the global set-once-only, but yeah looking at this, we do find the valid case of setting multiple times for tests, don't think we can block that. How about a log message?

With the idea of making setting the global more explicit

https://github.com/open-telemetry/opentelemetry-java/blob/master/docs/sdk-configuration.md#configuring-an-instance-of-the-sdk

and logging if the global is used before it's been set but the SDK is on the classpath

https://github.com/open-telemetry/opentelemetry-java/blob/master/docs/sdk-configuration.md#the-global-instance-of-the-sdk

hopefully it's reasonable enough even if set is allowed multiple times with a log message.

bogdandrutu:
I think if we cannot do our tests to pass this it means we did not defined the story of configuration right, how do we expect others to get it right?

I am completely opposite to release any API related to global if we cannot make this work (and we should not release otel API without global) so we need to fix everything to ensure that in all the cases our order and story of initialization fits this.

bogdandrutu:
@jkwatson 
> I'm not sure I understand what you mean. We have tests that call set right now, so of course they're going to fail if you make it only settable once, and have the 2nd time throw an exception.

I think we should figure out a way to structure the code so that this not happen and we have a strong guarantee that the global is not changed after it was initialized (SPI/Default/User set)

bogdandrutu:
@anuraaga I think just for tests (or actually between tests) we can make the JUNIT extension to "reset" the global between tests. But inside one test we should not have the problem.

bogdandrutu:
@jkwatson @anuraaga not necessary we need to have this merged in this way:
* we can return a checked exception
* we can log and proceed with set
* we can log and not proceed with set

There are options, but as long as we cannot write all the tests (at test level) to not crash with this change it means users will also not be able to configure things correctly and some components will have unexpected (most likely default) instances.

anuraaga:
@bogdandrutu Yeah - found a few of the were using `OpenTelemetry.getPropagators()` to get a list of empty propagators.

PR #2382, opened 2020-12-21
Change aggregator to use generics for Accumulation

bogdandrutu:
Followup PR will propagate the Accumulation type to the Aggregation to avoid type cast in merge.

### DIFF
--- a/sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/aggregation/LastValueAggregationTest.java
+++ b/sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/aggregation/LastValueAggregationTest.java
@@ -20,13 +20,12 @@ import io.opentelemetry.sdk.resources.Resource;
 import java.util.Collections;
 import org.junit.jupiter.api.Test;
 
-/** Unit tests for {@link AggregationFactory#lastValue()}. */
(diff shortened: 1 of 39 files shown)


sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/aggregation/LastValueAggregationTest.java
```diff
@@ -20,13 +20,13 @@
 import java.util.Collections;
 import org.junit.jupiter.api.Test;
 
-/** Unit tests for {@link Aggregations#sum()}. */
+/** Unit tests for {@link Aggregations#lastValue()}. */
```
jkwatson, line 23: I'd just delete this javadoc altogether. We've been removing them as they don't provide much value to anyone and are too easy to become out of date.

jkwatson (approved):
👍

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/aggregation/LastValueAggregationTest.java
bogdandrutu: Done.

PR #2383, opened 2020-12-21
Split different aggregations in their own files

bogdandrutu:

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregation/Aggregations.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregation/Aggregations.java
@@ -5,28 +5,8 @@
 
 package io.opentelemetry.sdk.metrics.aggregation;
 
-import io.opentelemetry.api.common.Labels;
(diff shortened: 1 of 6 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregation/Aggregations.java
```diff
@@ -5,36 +5,16 @@
...
-
 public final class Aggregations {
+  private Aggregations() {}
 
   /**
    * Returns an {@code Aggregation} that calculates sum of recorded measurements.
    *
    * @return an {@code Aggregation} that calculates sum of recorded measurements.
    */
   public static Aggregation sum() {
```
anuraaga, line 16: For followup guess we can move these into `Aggregation.java` itself

anuraaga (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregation/Aggregations.java
bogdandrutu: Yes, but I try to limit the PR sizes :)

PR #2384, opened 2020-12-22
Remove AbstractBoundInstrument, move mapping logic to Aggregator

bogdandrutu:
Remove one allocation when calling instruments without bounds.


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SynchronousInstrumentAccumulator.java
```diff
@@ -14,45 +14,41 @@
...
-      return binding;
+      return aggregator;
     }
 
     // Missing entry or no longer mapped, try to add a new entry.
-    binding = boundFactory.apply(instrumentProcessor.getAggregator());
+    aggregator = instrumentProcessor.getAggregator();
     while (true) {
-      B oldBound = boundLabels.putIfAbsent(labels, binding);
+      Aggregator oldBound = aggregatorLabels.putIfAbsent(labels, aggregator);
```
jkwatson, line 40: tiny nit: might rename this variable to `boundAggregator`

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/SynchronousInstrumentAccumulatorTest.java
```diff
@@ -0,0 +1,52 @@
...
+      accumulator.collectAll();
+      Aggregator anotherDuplicateAggregator = accumulator.bind(Labels.of("K", "V"));
+      try {
+        assertThat(anotherDuplicateAggregator).isEqualTo(aggregator);
+      } finally {
+        anotherDuplicateAggregator.release();
+      }
+    } finally {
+      aggregator.release();
+      aggregator.release();
```
jkwatson, line 49: could we add a verification that there is nothing bound at this point? Not sure if that would require dipping into the internals, but it seems like a valuable assertion to be able to make.

jkwatson (approved):
👍 good change. Thanks!

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SynchronousInstrumentAccumulator.java
bogdandrutu: Done.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/SynchronousInstrumentAccumulatorTest.java
bogdandrutu: So it is a bit of logic implied here. But because all of them are same instance based on the assertions in the test, just trying to unmap will confirm that the aggregator is no longer referenced (in use). This will unfortunately get a bit into internals and calling collectAll will not work anymore after this, but for testing is good enough.

PR #2385, opened 2020-12-22
Deprecate addSpanProcessor mutator

anuraaga:
Not used in our code anymore.


bogdandrutu:
I think before we prove we can pass the tests with the PR I created that global is set once per test and before anyone reads it we are not confident to remove this.

jkwatson:
> I think before we prove we can pass the tests with the PR I created that global is set once per test and before anyone reads it we are not confident to remove this.

Discussion ongoing over in https://github.com/open-telemetry/opentelemetry-java/discussions/2371 Your participation would be very helpful. Thanks!

jkwatson (approved):
Thanks!

jkwatson:
I'm going to merge this. If we decide we can't remove this due to unforeseen issues, we can un-deprecate it.

PR #2386, opened 2020-12-22
Reduce usage of GlobalOpenTelemetry

anuraaga:
Mostly to make the tests more readable rather than relying on what happens to be in the global at the time (possibly mutated in another test in the package).

### DIFF
--- a/perf-harness/src/test/java/io/opentelemetry/perf/OtlpPipelineStressTest.java
+++ b/perf-harness/src/test/java/io/opentelemetry/perf/OtlpPipelineStressTest.java
@@ -26,6 +26,7 @@ import io.opentelemetry.sdk.metrics.export.IntervalMetricReader;
 import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.sdk.resources.ResourceAttributes;
 import io.opentelemetry.sdk.testing.exporter.InMemoryMetricExporter;
+import io.opentelemetry.sdk.trace.SdkTracerManagement;
 import io.opentelemetry.sdk.trace.SdkTracerProvider;
 import io.opentelemetry.sdk.trace.export.BatchSpanProcessor;
 import java.io.IOException;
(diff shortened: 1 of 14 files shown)


perf-harness/src/test/java/io/opentelemetry/perf/OtlpPipelineStressTest.java
```diff
@@ -243,16 +245,22 @@ private static IntervalMetricReader setupSdk(MetricExporter metricExporter) {
...
-        .setMetricProducers(
-            Collections.singleton(
-                ((SdkMeterProvider) GlobalMetricsProvider.get()).getMetricProducer()))
-        .setExportIntervalMillis(1000)
-        .build();
-  }
+    Resource resource =
+        Resource.create(
+            Attributes.builder()
+                .put(ResourceAttributes.SERVICE_NAME, "PerfTester")
```
jkwatson, line 251: heh...this is more of a "how does the SDK react to bad situations" than perf per se, but this is fine. :)

jkwatson: And, I had put `perf` in the name of the tracer, so I'm mostly to blame here. ;)

perf-harness/src/test/java/io/opentelemetry/perf/OtlpPipelineStressTest.java
```diff
@@ -101,6 +102,8 @@
 
   private final InMemoryMetricExporter metricExporter = InMemoryMetricExporter.create();
 
+  private SdkTracerProvider tracerProvider;
```
jkwatson, line 105: Do we need the whole `SdkTracerProvider` here, or is the management interface enough?  I actually struggled with this kind of decision in the examples, as well.  It's so easy to end up with the SDK tracer provider and to use it for instrumentation, and we'd definitely like to discourage that pattern as much as possible.

jkwatson (approved):
Thanks!

perf-harness/src/test/java/io/opentelemetry/perf/OtlpPipelineStressTest.java
anuraaga: Oops forgot to remove the unnecessary system property too, phew

PR #2387, opened 2020-12-22
Rename OpenTelemetrySdkBuilder.build to buildAndRegisterGlobal and ad…

anuraaga:
…d non-registering version.

From https://github.com/open-telemetry/opentelemetry-java/blob/master/docs/sdk-configuration.md#configuring-an-instance-of-the-sdk - this is to reduce the magic of a `build` method affecting the global.

Sending as draft for now since want to merge #2386 first


sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdkBuilder.java
```diff
@@ -51,19 +49,31 @@ public OpenTelemetrySdkBuilder setPropagators(ContextPropagators propagators) {
...
+   * Returns a new {@link OpenTelemetrySdk} built with the configuration of this {@link
+   * OpenTelemetrySdkBuilder}. This SDK is not registered as the global {@link
+   * io.opentelemetry.api.OpenTelemetry}. It is recommended that you register one SDK using {@link
+   * OpenTelemetrySdkBuilder#buildAndRegisterGlobal()} for use by instrumentation that requires
+   * access to a global instance of {@link io.opentelemetry.api.OpenTelemetry}.
+   *
+   * @see GlobalOpenTelemetry
    */
-  public OpenTelemetrySdk build() {
+  public OpenTelemetrySdk buildWithoutRegisteringGlobal() {
```
jkwatson, line 72: I still think that just calling this `build()` would be fine. If you're looking at your auto-complete options, it should be obvious that the one that has `AndRegisterGlobal()` on the end of it would do what it says, and, hence, this one wouldn't.

anuraaga: Ok :)

jkwatson (approved):

PR #2388, opened 2020-12-22
Add some metric types to ocshim tests

anuraaga:
I tried various invocations with `AggregationWindow.Interval` but could not get a summary to be exported. Since summaries are deprecated in OC anyways, I wonder if we can just remove their conversion code and consider them unsupported? Or if anyone can tell me how to add coverage for them would be nice.


jkwatson:
We definitely need some OC support in this project to help maintain the OC backward compatibility. @jsuereth are you able to help out here?

jkwatson (approved):
Thanks!

PR #2390, opened 2020-12-22
Allow registering TraceConfigSupplier and use it from zpages.

anuraaga:
This is to move mutability out of the SDK and into those that need mutability.

ZPages still eventually needs some rework to make it applicable to non-global SDK, for now it is a mix but I think we should update that as a followup since it's an extension.

An idea for
```java
interface SdkTracerProviderExtension {
  void configure(SdkTracerProviderBuilder builder);
}

class SdkTracerProviderBuilder {
  addExtension(SdkTracerProviderExtension extension) { extension.configure(this)) }
```

came to mind but for now not pursuing it.


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkTracerProviderBuilder.java
```diff
@@ -67,7 +68,20 @@ public SdkTracerProviderBuilder setResource(Resource resource) {
...
+    return this;
+  }
+
+  /**
+   * Assign a {@link Supplier} of {@link TraceConfig}. {@link TraceConfig} will be retrieved each
+   * time a {@link io.opentelemetry.api.trace.Span} is started.
+   *
+   * @return this
+   */
+  public SdkTracerProviderBuilder setTraceConfigSupplier(
```
jkwatson, line 81: My only concern here is that we now have 2 methods that override each other's selection, which could potentially add some confusion for the user. Could we possibly get away only with this one, rather than providing both and dealing with the potential confusion?

anuraaga: Yeah it's definitely a concern. This didn't feel "that bad" and I think it's important to preserve the vastly more common case of static config, `setTraceConfig(TraceConfig.builder()....build())`. A couple of other ideas that come to mind

- `TraceConfigBuilder.build` returns a `Supplier<TraceConfig>`. This is easy to use, just wonder if it's confusing for a builder to not return the precise type in its name. I guess not
- Rename this one to `setTraceConfig` so it's an overload. I think having multiple overloads where only one should be used is quite common. In this case it's the opposite problem of the above, where I wonder if it's confusing for the same method name to accept both the type and a supplier of the type. I guess not

Accepting help on deciding among the options :)

jkwatson: My opinion on this is that it is fine to have two methods. And, having the same method name will probably help to make it clear that they are doing the same thing, and hence more obvious that only one of them applies.

I do like this approach to TraceConfig dynamism, as I said.  So, 👍  from me!

build.gradle
```diff
@@ -131,6 +131,9 @@ subprojects {
             it.options.errorprone.disable("JdkObsolete")
             it.options.errorprone.disable("UnnecessaryAnonymousClass")
 
+            // Limits APIs
+            it.options.errorprone.disable("NoFunctionalReturnType")
```
jkwatson, line 135: ❤️

jkwatson (approved):
I approve

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkTracerProviderBuilder.java
anuraaga: Thanks - went with same method name approach

PR #2391, opened 2020-12-22
Add aggregation factory and remove specialized only the aggregator by…

bogdandrutu:
… value type

This will allow to specialize merge arguments and other methods in the Aggregation by the type of the value recorded.


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregation/CountAggregation.java
```diff
@@ -10,19 +10,20 @@
...
-enum CountAggregation implements Aggregation {
-  INSTANCE;
+final class CountAggregation implements Aggregation {
+  static final CountAggregation INSTANCE = new CountAggregation();
+
+  private CountAggregation() {}
 
   @Override
-  public AggregatorFactory getAggregatorFactory(InstrumentValueType instrumentValueType) {
+  public AggregatorFactory getAggregatorFactory() {
```
jkwatson, line 26: I wonder if we could hide the `AggregatorFactory` on the Aggregation and just have a method like `getAggregator()` on it directly, so simplify the call chain. I might be missing something, though.

jkwatson (approved):
Seems like a move in the right direction. It might be nice, when things are stable, to see if we can collapse some of the calls and hide some of the details. 👍

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregation/CountAggregation.java
bogdandrutu: Will look later, I think I need the factory, but you will see that probably needs a different name.

PR #2392, opened 2020-12-22
Add ability to create accumulation to the Aggregator

bogdandrutu:
This way the Accumulator will produce the right accumulations, for example in case of count. Added tests for all instrument types.

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/AggregatorFactory.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/AggregatorFactory.java
@@ -10,12 +10,37 @@ import javax.annotation.concurrent.Immutable;
 
 /** Factory class for {@link Aggregator}. */
 @Immutable
-public interface AggregatorFactory<T extends Accumulation> {
+public abstract class AggregatorFactory<T extends Accumulation> {
 
   /**
-   * Returns a new {@link Aggregator}.
+   * Returns a new {@link Aggregator}. This MUST by used by the synchronous to aggregate recorded
+   * measurements during the collection cycle.
    *
    * @return a new {@link Aggregator}.
    */
-  Aggregator<T> getAggregator();
+  public abstract Aggregator<T> getAggregator();
+
+  /**
+   * Returns a new {@code Accumulation} for the given value. This MUST be used by the asynchronous
+   * instruments to create {@code Accumulation} that are passed to the processor.
+   *
+   * @param value the given value to be used to create the {@code Accumulation}.
+   * @return a new {@code Accumulation} for the given value.
+   */
+  public Accumulation accumulateLong(long value) {
+    throw new UnsupportedOperationException(
+        "This aggregator does not support recording long values.");
+  }
+
+  /**
+   * Returns a new {@code Accumulation} for the given value. This MUST be used by the asynchronous
+   * instruments to create {@code Accumulation} that are passed to the processor.
+   *
+   * @param value the given value to be used to create the {@code Accumulation}.
+   * @return a new {@code Accumulation} for the given value.
+   */
+  public Accumulation accumulateDouble(double value) {
+    throw new UnsupportedOperationException(
+        "This aggregator does not support recording double values.");
+  }
 }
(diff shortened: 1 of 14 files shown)


sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/SdkMeterProviderTest.java
```diff
@@ -0,0 +1,460 @@
...
+                MetricData.LongSumData.create(
+                    /* isMonotonic= */ true,
+                    MetricData.AggregationTemporality.CUMULATIVE,
+                    Collections.singletonList(
+                        MetricData.LongPoint.create(
+                            testClock.now(), testClock.now(), Labels.empty(), 1)))));
+  }
+
+  @Test
+  void collectAllASyncInstruments() {
```
jkwatson, line 246: nit : collectAllAsyncInstruments  (lower-case the 's' in async)

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/SdkMeterProviderTest.java
```diff
@@ -0,0 +1,460 @@
...
+                "",
+                "1",
+                MetricData.DoubleGaugeData.create(
+                    Collections.singletonList(
+                        MetricData.DoublePoint.create(
+                            testClock.now(), testClock.now(), Labels.empty(), 10.1)))));
+  }
+
+  @Test
+  void collectAllASyncInstruments_CustomAggregation() {
```
jkwatson, line 346: same naming nit here

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/AggregatorFactory.java
```diff
@@ -5,16 +5,42 @@
...
+  public abstract Aggregator getAggregator();
+
+  /**
+   * Returns a new {@code Accumulation} for the given value. This MUST be used by the asynchronous
+   * instruments to create {@code Accumulation} that are passed to the processor.
+   *
+   * @param value the given value to be used to create the {@code Accumulation}.
+   * @return a new {@code Accumulation} for the given value.
+   */
+  public Accumulation accumulateLong(long value) {
```
jkwatson, line 30: discussed offline that we should probably move these methods into the Aggregator. Not sure if this should be done in this PR or a follow-up.

Also, it might be nice to call it something like "directAccumulate" or something that makes it clear it should be used by async instruments.

bogdandrutu: If you are ok I would do it in the next PR :)

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/SdkMeterProviderTest.java
bogdandrutu: Done.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/SdkMeterProviderTest.java
bogdandrutu: Done.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/SdkMeterProviderTest.java
```diff
@@ -0,0 +1,460 @@
...
+import io.opentelemetry.sdk.metrics.common.InstrumentType;
+import io.opentelemetry.sdk.metrics.data.MetricData;
+import io.opentelemetry.sdk.metrics.view.AggregationConfiguration;
+import io.opentelemetry.sdk.metrics.view.InstrumentSelector;
+import io.opentelemetry.sdk.resources.Resource;
+import java.util.Arrays;
+import java.util.Collections;
+import org.junit.jupiter.api.Test;
+
+public class SdkMeterProviderTest {
```
jkwatson, line 31: excellent test. thanks!

jkwatson (approved):
👍

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/SdkMeterProviderTest.java
bogdandrutu: Thanks

PR #2394, opened 2020-12-23
Make Aggregation generic, so we can ensure type safety between compon…

bogdandrutu:
…ents

This ensures that we pass the correct type of Accumulation between components inside the SDK.

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregation/Aggregation.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregation/Aggregation.java
@@ -20,7 +20,7 @@ import javax.annotation.concurrent.Immutable;
  * given {@code Instrument} into the equivalent {@code MetricData}.
  */
 @Immutable
-public interface Aggregation {
+public interface Aggregation<T extends Accumulation> {
 
   /**
    * Returns an {@code AggregationFactory} that can be used to produce the {@link
@@ -29,14 +29,14 @@ public interface Aggregation {
    *
    * @return the {@code AggregationFactory}.
    */
-  AggregatorFactory<?> getAggregatorFactory();
+  AggregatorFactory<T> getAggregatorFactory();
 
   /**
    * Returns the result of the merge of the given {@link Accumulation}s.
    *
    * @return the result of the merge of the given {@link Accumulation}s.
    */
-  Accumulation merge(Accumulation a1, Accumulation a2);
+  T merge(T a1, T a2);
 
   /**
    * Returns the {@link MetricData} that this {@code Aggregation} will produce.
@@ -55,7 +55,7 @@ public interface Aggregation {
       Resource resource,
       InstrumentationLibraryInfo instrumentationLibraryInfo,
       InstrumentDescriptor descriptor,
-      Map<Labels, Accumulation> accumulationMap,
+      Map<Labels, ? extends Accumulation> accumulationMap,
       long startEpochNanos,
       long epochNanos);
 }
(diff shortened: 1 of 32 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregation/ImmutableAggregationFactory.java
```diff
@@ -9,36 +9,40 @@
...
+  public <V extends Accumulation> Aggregation<V> create(InstrumentValueType instrumentValueType) {
     switch (instrumentValueType) {
       case LONG:
-        return longAggregation;
+        return (Aggregation<V>) longAggregation;
       case DOUBLE:
-        return doubleAggregation;
+        return (Aggregation<V>) doubleAggregation;
     }
     throw new IllegalArgumentException("Invalid instrument value type");
```
bogdandrutu, line 47: This hides the actual type behind the <V> still safe in the Accumulator and Processor but because of this you can see in DoubleMinMaxSumCountAggregatorTest.java that I need to add an annotation (errorprone complained), then discovered that I cannot use `MinMaxSumCountAccumulation` there because everything is actually hidden behind this conversion.

I will most likely need to make the factory have 2 generic types and 2 methods `createLong` and `createDouble` which will return their own generic.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregation/Aggregation.java
```diff
@@ -55,7 +55,7 @@ MetricData toMetricData(
       Resource resource,
       InstrumentationLibraryInfo instrumentationLibraryInfo,
       InstrumentDescriptor descriptor,
-      Map<Labels, Accumulation> accumulationMap,
+      Map<Labels, ? extends Accumulation> accumulationMap,
```
bogdandrutu, line 58: This will be transformed in T as well soon, but trying to limit the size of the PR.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregation/AggregationFactory.java
```diff
@@ -61,5 +61,5 @@ static AggregationFactory minMaxSumCount() {
    * @param instrumentValueType the type of recorded values for the {@code Instrument}.
    * @return a new {@link Aggregation}.
    */
-  Aggregation create(InstrumentValueType instrumentValueType);
+  <V extends Accumulation> Aggregation<V> create(InstrumentValueType instrumentValueType);
```
jkwatson, line 64: `V`?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregation/MinMaxSumCountAggregation.java
```diff
@@ -18,40 +18,33 @@
...
+        Math.max(a1.getMax(), a2.getMax()));
   }
 
   @Override
   public MetricData toMetricData(
       Resource resource,
       InstrumentationLibraryInfo instrumentationLibraryInfo,
       InstrumentDescriptor descriptor,
-      Map<Labels, Accumulation> accumulationMap,
+      Map<Labels, ? extends Accumulation> accumulationMap,
```
jkwatson, line 47: for future, I might name this parameter "accumulationsByLabels". No need to change it now.

jkwatson (approved):
👍

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregation/AggregationFactory.java
bogdandrutu: Changed to `T` :)

PR #2395, opened 2020-12-23
Rename Aggregator -> AggregatorHandle; AggregatorFactory -> Aggregator

bogdandrutu:
Left a TODO to move `getInstance` to the new `Aggregator` interface.


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregation/AggregationFactory.java
```diff
@@ -5,11 +5,11 @@
 
 package io.opentelemetry.sdk.metrics.aggregation;
 
-import io.opentelemetry.sdk.metrics.aggregator.Aggregator;
+import io.opentelemetry.sdk.metrics.aggregator.AggregatorHandle;
 import io.opentelemetry.sdk.metrics.common.InstrumentValueType;
 import javax.annotation.concurrent.Immutable;
 
-/** Factory class for {@link Aggregator}. */
+/** Factory class for {@link AggregatorHandle}. */
```
jkwatson, line 12: This comment is no longer correct.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleMinMaxSumCountAggregator.java
```diff
@@ -11,77 +11,85 @@
...
+  public MinMaxSumCountAccumulation accumulateDouble(double value) {
+    return MinMaxSumCountAccumulation.create(1, value, value, value);
   }
 
-  private static final class DoubleState {
-    private long count;
-    private double sum;
-    private double min;
-    private double max;
+  static final class Handle extends AggregatorHandle<MinMaxSumCountAccumulation> {
```
jkwatson, line 40: If we're ok with not testing the exact type returned by the createHandle method, we could make these handle implementations private.

jkwatson (approved):
Looks good! Thanks!

anuraaga (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleMinMaxSumCountAggregator.java
bogdandrutu: I had them initially private, then when I had to fix the tests I did this. We should follow up on this, for the moment I am keeping this way to not remove tests.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregation/AggregationFactory.java
bogdandrutu: Updated.

PR #2396, opened 2020-12-23
Log warning if SDK is detected on classpath but has not been initialized when accessing global.

anuraaga:
This is on the read side, whereas #2380 is on the write side. We'll probably want both.


bogdandrutu (dismissed):
This makes the assumption that we remove SPI, otherwise this is fine to happen. Do we want this if user expects to use SPI?

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
```diff
@@ -115,4 +131,35 @@ public static Tracer getTracer(String instrumentationName, String instrumentatio
...
+        logger.log(
+            Level.SEVERE,
+            "Attempt to access GlobalOpenTelemetry.get before OpenTelemetrySdk has been "
+                + "initialized. This generally means telemetry will not be recorded for parts of "
+                + "your application. Make sure to initialize OpenTelemetrySdk, using "
+                + "OpenTelemetrySdk.builder(), as early as possible in your application. "
+                + "If you do not need to use the OpenTelemetry SDK, either exclude it from your "
+                + "classpath or set the 'otel.sdk.suppress-sdk-initialized-warning' system "
+                + "property to true.",
+            new Throwable());
```
jkwatson, line 159: does severe require a throwable, or is this here to generate a stack trace where the user can see where the issue is?

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
```diff
@@ -115,4 +131,35 @@ public static Tracer getTracer(String instrumentationName, String instrumentatio
...
+      }
+    }
+
+    static void logIfSdkFound() {}
```
jkwatson, line 163: ????

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
```diff
@@ -115,4 +131,35 @@ public static Tracer getTracer(String instrumentationName, String instrumentatio
...
+      boolean hasSdk = false;
+      try {
+        Class.forName("io.opentelemetry.sdk.OpenTelemetrySdk");
+        hasSdk = true;
+      } catch (Throwable t) {
+        // Ignore
+      }
+
+      if (hasSdk) {
+        // TODO(anuraaga): Play https://www.youtube.com/watch?v=tBA-1ENblN4 if JavaFX detected on
```
jkwatson, line 148: 😂 😹

jkwatson, line 148: also JavaFX???

anuraaga:
@bogdandrutu I am proposing having all autoconfiguration in a separate artifact. That artifact could use SPI to allow configuration of non-env var controlled stuff. And this code could be updated to detect whether the artifact (which is under our control so doesn't need a generic entry point mechanism) is on the classpath and call it's initialization entry point with reflection if we want it to be a completely automatic process, and this log wouldn't be thrown for users of that SPI-approach.

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
anuraaga: Not required, it's indeed to generate a stack trace for the user to find the problem

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
anuraaga: I thought it's what all the cool kids are using now adays for multimedia in Java :P

anuraaga: Removed in latest commit since you had your laugh :)

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
anuraaga: Added a comment

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
```diff
@@ -30,6 +32,12 @@
  * @see ContextPropagators
  */
 public final class GlobalOpenTelemetry {
+  // Visible for testing
+  static final Logger logger = Logger.getLogger(GlobalOpenTelemetry.class.getName());
+
+  private static final boolean suppressSdkCheck =
+      Boolean.getBoolean("otel.sdk.suppress-sdk-initialized-warning");
```
jkwatson, line 39: this jdk API continues to be one of the weirdest one out there.

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
```diff
@@ -115,4 +131,37 @@ public static Tracer getTracer(String instrumentationName, String instrumentatio
...
+  // Use an inner class that checks and logs in its static initializer to have log-once behavior
+  // using initial classloader lock and no further runtime locks or atomics.
+  private static class SdkChecker {
+    static {
+      boolean hasSdk = false;
+      try {
+        Class.forName("io.opentelemetry.sdk.OpenTelemetrySdk");
+        hasSdk = true;
+      } catch (Throwable t) {
+        // Ignore
```
jkwatson, line 144: might be marginally clearer if we just returned immediately in this case

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
jkwatson: haha. awesome

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
anuraaga: It's a static initializer so can't return

jkwatson: you are so right. can static initializers call static methods that could return? Just curious...doesn't really matter that much.

anuraaga: It can - but would just add some indirection so inclined to leave it like this.

jkwatson (approved):
I approve the removal of SPI from the main API, so I also approve of this as a stop-gap warning to show the user that they need to force their SDK init earlier.

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
bogdandrutu, line 39: Why do we refer to SDK in the API?

bogdandrutu:
@jkwatson you said you approve removing of the SPI, but SPI is still there. So I am confused what this PR gives us (especially if I use SPI, which is still available).

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
```diff
@@ -115,4 +131,37 @@ public static Tracer getTracer(String instrumentationName, String instrumentatio
   public static ContextPropagators getPropagators() {
     return get().getPropagators();
   }
+
+  // Use an inner class that checks and logs in its static initializer to have log-once behavior
+  // using initial classloader lock and no further runtime locks or atomics.
+  private static class SdkChecker {
```
bogdandrutu, line 137: Call it "ImplementationChecker". We are currently refer as "SDK" to our official implementation.

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
```diff
@@ -115,4 +131,37 @@ public static Tracer getTracer(String instrumentationName, String instrumentatio
...
+    static {
+      boolean hasSdk = false;
+      try {
+        Class.forName("io.opentelemetry.sdk.OpenTelemetrySdk");
```
bogdandrutu, line 141: Why do we hardcode our implementation in the API artifact?

anuraaga: Bigger question is why not? We have an official implementation, the OpenTelemetry SDK, which we expect to be used in the vast majority of cases. If there's something we get from picking into other artifacts we publish, there isn't really a reason not to. It's a bit like how we have the comments "package-private to allow usage from auto-instrumentation" - that's even another repository, but we think it's worth it to allow good instrumentation experience.

anuraaga (commented):
> you said you approve removing of the SPI, but SPI is still there. So I am confused what this PR gives us (especially if I use SPI, which is still available).

@bogdandrutu This PR gets us a warning when the global is used before the user initializes the SDK. It doesn't have any downside on the current SPI. And the current SPI will probably be removed after #2401, we've discussed that it isn't helping much in #2371. An API-level SPI for autoloading a custom implementation can be added later if it helps a lot, until we understand better the actual usage patterns of custom implementations it's hard to say. In the meantime, let's focus on the UX of what we expect, the OpenTelemetry SDK.

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
jkwatson: Because our SDK is going to be the solution for 99.99% of all users. I think it's fine to refer to our SDK like this.

anuraaga (commented):
@jkwatson @bogdandrutu Do you think we can add this in now that the SPI isn't there?

anuraaga:
@jkwatson @bogdandrutu I realized that we do need this PR (https://github.com/open-telemetry/opentelemetry-java/pull/2548#pullrequestreview-571871153 was only about warning when `set` is called in the wrong order, but does not help in the case that `buildAndRegisterGlobal` is forgotten completely). Can we go ahead and add this?

anuraaga:
For reference: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1550#issuecomment-791217067

trask (approved):
> For reference: [open-telemetry/opentelemetry-java-instrumentation#1550 (comment)](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1550#issuecomment-791217067)

👍

asarkar:
@anuraaga  I suggest that the warning log clearly say what the problem is, and what to do to fix it, like "No OpenTelemetry implementation found; consider adding <Maven GAV> to classpath" instead of get/set stuff which dumb users like me won't understand.

anuraaga:
@asarkar It says `Make sure to initialize OpenTelemetrySdk, using OpenTelemetrySdk.builder()...buildAndRegisterGlobal(), as early as possible in your application`. Isn't that what you need to know?

asarkar:
> It says Make sure to initialize OpenTelemetrySdk, using OpenTelemetrySdk.builder()...buildAndRegisterGlobal(), as early as possible in your application. Isn't that what you need to know?

I'd need to know where to get it. Remember, if someone is missing the SDK, they don't have `OpenTelemetrySdk` on classpath.

anuraaga:
@asarkar Yeah this message is only thrown when the SDK is on the classpath, where it's almost always the case that it's an error to have forgotten to register it. It makes this PR a relative no-brainer.

This PR does not change the behavior when SDK isn't on the classpath, in which case noop is used - this actually does have real use cases for example, a library may be instrumented with `OpenTelemetry` but a user doesn't actually want to enable OpenTelemetry in their app by including the SDK. It might still be worth logging something for this case but it's not as clear cut and we can think about it separate from this one.

jkwatson:
I still think this is a good addition. @bogdandrutu any new thoughts?

jkwatson:
> I still think this is a good addition. @bogdandrutu any new thoughts?

@bogdandrutu seems like we should unblock and get this merged. You ok with that?

jkwatson:
@anuraaga I think we can unblock this if you are interested.

anuraaga:
@jkwatson Yeah it'd be nice to merge this

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
```diff
@@ -168,6 +175,39 @@ private static OpenTelemetry maybeAutoConfigure() {
...
+      if (hasSdk) {
+        logger.log(
+            Level.SEVERE,
+            "Attempt to access GlobalOpenTelemetry.get before OpenTelemetrySdk has been "
+                + "initialized. This generally means telemetry will not be recorded for parts of "
+                + "your application. Make sure to initialize OpenTelemetrySdk, using "
+                + "OpenTelemetrySdk.builder()...buildAndRegisterGlobal(), as early as possible in "
+                + "your application. If you do not need to use the OpenTelemetry SDK, either "
+                + "exclude it from your classpath or set the "
+                + "'otel.sdk.suppress-sdk-initialized-warning' system property to true.",
```
kubawach, line 199: Looking at how we name configuration properties enabling/disabling various options here and in the "-instrumentation" I believe that a better one could be `otel.sdk.not-initialized-warning.enable` set by default to `true`. In all of the other cases we use "enable" rather that "supress (disable)".

kubawach (approved):

jkwatson:
@anuraaga I'm sure this needs a rebase. Thanks!

PR #2398, opened 2020-12-23
Replace Batch/SimpleSpanProcessor with BatchSettings managed by SdkTr…

anuraaga:
…acerProvider.

This is just one idea to try to make it easier to do a very common thing, register an exporter for tracing. It makes batching an implementation detail of the tracer provider rather than a separate concept. Wondering what people think.


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkTracerProviderBuilder.java
```diff
@@ -94,14 +98,53 @@ public SdkTracerProviderBuilder addSpanProcessor(SpanProcessor spanProcessor) {
...
+   * provided, no batching will be done.
+   */
+  // Using deprecated for public use class
+  @SuppressWarnings({"deprecation", "UnnecessarilyFullyQualified"})
+  public SdkTracerProviderBuilder addExporter(SpanExporter exporter, BatchSettings batchSettings) {
+    requireNonNull(exporter, "exporter");
+    requireNonNull(batchSettings, "batchSettings");
+    if (batchSettings == BatchSettings.noBatching()) {
+      exporterSpanProcessors.add(
+          io.opentelemetry.sdk.trace.export.SimpleSpanProcessor.create(exporter));
```
jkwatson, line 123: odd fully qualified import here

anuraaga: It seems to be the only way to avoid the `deprecation` warning that would other get applied to the import statement itself amazingly.

jkwatson: weird!

jkwatson:
This is a cool idea, but I'd like to propose that we shelve this sort of change until after we've got the rest of the configuration story solid and agreed to.

anuraaga:
@jkwatson Rebased this on the recent changes - what do you think?

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -61,7 +61,11 @@
  *   <li>{@code OTEL_BSP_EXPORT_TIMEOUT}: sets the maximum allowed time to export data.
  *   <li>{@code OTEL_BSP_EXPORT_SAMPLED}: sets whether only sampled spans should be exported.
  * </ul>
+ *
+ * @deprecated Use {@link
+ *     io.opentelemetry.sdk.trace.SdkTracerProviderBuilder#addExporter(SpanExporter)}
  */
+@Deprecated
```
jkwatson, line 68: I don't think we should deprecate the batch span processor, or remove it. This is something that is basically required by the current SDK specification and I would not remove it without some significant spec rewording.

jkwatson (changes requested):

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
anuraaga: I think even with this approach we are satisfying everything here

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#span-processor

- Built in processors can be registered (`BatchSettings`)

- Custom processors can be registered (`addSpanProcessor`)

- Built in processors can be decorated (`addSpanProcessor`)

I would rather expose BSP later if it becomes really necessary if we are ok with the approach in this PR. Or otherwise close the PR and live with registering an exporter being a bit tedious.

jkwatson:
I spent a lot of time thinking about this over the weekend. 

1) I think that our current span processor (not just java..what is specced) model is pretty limited and definitely will not fulfill all the use-cases that we will need. But, it will work for 99% of the users for now.
2) At this late date, I don't think it's a good idea to try and upend the current model, flawed though it might be. Any solution that we patch in at this point will probably only be addressing a narrow solution space, and not be a comprehensive re-think of the model of the span pipeline.

So, I'd prefer that we leave the current system in place for now. Yes, it's non-ideal. Yes, it does lead to a bit of awkwardness when configuring the SDK. And, if we want to address the configuration issue independently, perhaps by providing some new APIs for the simple use-cases, I would absolutely support that, post 1.0.

jkwatson:
@anuraaga Do you have plans to re-open this idea as a easier-to-use mode of batch operations? I still think it could be good to have that convenience API for configuring the SDK, although perhaps the auto-configuration module has made it relatively moot?

anuraaga:
Let's go ahead and close this for now, it can use another think

PR #2400, opened 2020-12-23
Make prometheus client API dependency of exporter.

anuraaga:
`PrometheusCollector` extends `Collector` so the prometheus artifact is actually in the public API of the module.


jkwatson (approved):
Thanks!

PR #2401, opened 2020-12-23
Add autoconfiguration wrapper artifact

anuraaga:
This implements autoconfiguration moved into a separate layer as discussed a few times before. It's aim is to remove autoconfiguration from the core SDK so frameworks with their own configuration system, e.g., sleuth, can use it with full control, and other users can rely on our autoconfiguration if they want. Since we have control of the artifact, while slightly weird, it seems reasonable for us to automatically call into it with reflection, and otherwise we just return no-op (and log in the case of a misconfigured SDK as in #2396).

I mostly followed the parameters supported by the java agent except

- Updated names to use latest from spec, some small tweaks
- Didn't add support for "max attributes per event" type variables since not in spec
- Didn't add jaeger_thrift https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1973


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
```diff
@@ -0,0 +1,351 @@
...
+
+    return OpenTelemetrySdk.builder()
+        .setTracerProvider(tracerProviderBuilder.build())
+        .setPropagators(ContextPropagators.create(TextMapPropagator.composite(propagators)))
+        .build();
+  }
+
+  private static void configureOtlpMetrics(
+      ConfigProperties config, SdkMeterProvider meterProvider) {
+    OtlpGrpcMetricExporterBuilder builder = OtlpGrpcMetricExporter.builder();
```
anuraaga, line 175: Will need to add linkage error guards to return better error messages for when a configuration option is set but implementation not available.

marcingrzejszczak:
@anuraaga thanks for mentioning me! BTW can you start mentioning @jonatan-ivanov too (that's why I suggested @jkwatson to create a Github Team and just add people to the team and mention that team instead of concrete people).

build.gradle
```diff
@@ -199,6 +199,7 @@ subprojects {
                     jsr305                      : "com.google.code.findbugs:jsr305:${findBugsJsr305Version}",
                     prometheus_client           : "io.prometheus:simpleclient:${prometheusVersion}",
                     prometheus_client_common    : "io.prometheus:simpleclient_common:${prometheusVersion}",
+                    prometheus_client_httpserver: "io.prometheus:simpleclient_httpserver:${prometheusVersion}",
```
jkwatson, line 202: this doesn't belong in this PR, does it?

jkwatson: oh, I think I see..if autoconfigure is going to configure it...  nm

jkwatson:
> @anuraaga thanks for mentioning me! BTW can you start mentioning @jonatan-ivanov too (that's why I suggested @jkwatson to create a Github Team and just add people to the team and mention that team instead of concrete people).

It turns out that I don't have permissions to create a team (or at least I get a 404 from github when trying to click the "new team" link). So, @bogdandrutu will have to do this, I think, unless he gives me (and @anuraaga) owner permissions on the repository.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
```diff
@@ -0,0 +1,351 @@
...
+      }
+    }
+
+    return OpenTelemetrySdk.builder()
+        .setTracerProvider(tracerProviderBuilder.build())
+        .setPropagators(ContextPropagators.create(TextMapPropagator.composite(propagators)))
+        .build();
+  }
+
+  private static void configureOtlpMetrics(
```
jkwatson, line 173: I think I'd break each of these methods out into their own class so it's clear how to add new ones to the auto-config, rather than having to dig into this beast to find them. :)

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
```diff
@@ -115,4 +122,25 @@ public static Tracer getTracer(String instrumentationName, String instrumentatio
...
+      return null;
+    }
+
+    try {
+      Method initialize = openTelemetrySdkAutoConfiguration.getMethod("initialize");
+      return (OpenTelemetry) initialize.invoke(null);
+    } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
+      throw new IllegalStateException(
+          "OpenTelemetrySdkAutoConfiguration detected on classpath "
+              + "but could not invoke initialize method. This is a bug in OpenTelemetry.",
```
jkwatson, line 142: this could also be due to a restrictive security manager

jkwatson:
this is pretty slick. It's SPI (albeit a slightly different flavor, with the builder), but it doesn't demand SPI be used. It makes it easy to manually auto-configure or manually configure, or mix-and-match however you like.

👍  to this approach. Obviously, if we want to go this route, we'll want to write some tests for the configuration (or move them over from the existing configuration code/tests), and do some refactor/cleanup so it's not one enormous class, but as a proof-of-concept, this is very nice.  🚀

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
anuraaga: For a public method, as far as I know there wouldn't be any situation that could be blocked by security manager, is there one?

jkwatson: I thought security managers could block all reflection, but I wouldn't swear to it.

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
```diff
@@ -48,6 +51,13 @@ public static OpenTelemetry get() {
...
       synchronized (mutex) {
         if (globalOpenTelemetry == null) {
+
+          OpenTelemetry autoConfigured = maybeAutoConfigure();
+          if (autoConfigured != null) {
+            set(autoConfigured);
+            return autoConfigured;
+          }
+
           OpenTelemetryFactory openTelemetryFactory = Utils.loadSpi(OpenTelemetryFactory.class);
```
anuraaga, line 61: Unlike the first commit with PoC, I haven't removed the API SPI yet, that we would do in a separate PR.

buildscripts/checkstyle.xml
```diff
@@ -55,20 +55,6 @@
              value="LITERAL_TRY, LITERAL_FINALLY, LITERAL_IF, LITERAL_ELSE, LITERAL_SWITCH"/>
         </module>
         <module name="NeedBraces"/>
-        <module name="LeftCurly"/>
```
anuraaga, line 58: See #2418

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
```diff
@@ -0,0 +1,69 @@
...
+import io.opentelemetry.sdk.metrics.SdkMeterProvider;
+import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.sdk.trace.SdkTracerProvider;
+import java.util.List;
+import java.util.Locale;
+
+/**
+ * Auto-configuration for the OpenTelemetry SDK. As an alternative to programmatically configuring
+ * the SDK using {@link OpenTelemetrySdk#builder()}, this package can be used to automatically
+ * configure the SDK using environment properties specified by OpenTelemetry.
```
anuraaga, line 21: I'm hoping to link to a docs page from here rather than outlining all properties here - presuably autoconfiguration users are the last to read the javadocs

anuraaga (commented):
I have a bit more grungework in adding unit tests for ConfigProperties and some failure cases, but I think this is mostly ready for another look

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
jkwatson: Yes, a security manager can block all reflection, even for public methods, and even just calls to "getMethod", although I would suspect that's extremely rare these days.

jkwatson:
This is looking like a fantastic start. @anuraaga What do you think needs to be done to get it out of draft mode? I'd love to move forward with this ASAP, so we can move toward a 1.0 release.

anuraaga:
@jkwatson Just a few more boring unit tests I'll knock out today :)

anuraaga:
@jkwatson Ok should be good to go

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
```diff
@@ -0,0 +1,135 @@
...
+    return new ConfigProperties(System.getProperties(), System.getenv());
+  }
+
+  // Visible for testing
+  @SuppressWarnings({"unchecked", "rawtypes"})
+  static ConfigProperties createForTest(Map<String, String> properties) {
+    return new ConfigProperties((Map) properties, Collections.emptyMap());
+  }
+
+  private final Map<String, String> config;
```
jkwatson, line 31: this should be up above the static methods

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
```diff
@@ -0,0 +1,135 @@
...
+
+  private ConfigProperties(
+      Map<Object, Object> systemProperties, Map<String, String> environmentVariables) {
```
jkwatson, line 34: the left map could be `Map<?,?>` then you don't have to cast above.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
```diff
@@ -0,0 +1,135 @@
...
+            keyValuePair ->
+                Arrays.stream(keyValuePair.split("=", 2))
+                    .map(String::trim)
+                    .filter(s -> !s.isEmpty())
+                    .collect(Collectors.toList()))
+        .map(
+            splitKeyValuePairs -> {
+              if (splitKeyValuePairs.size() != 2) {
+                throw new ConfigurationException(
+                    "Map property key missing value: " + name + "=" + config.get(name));
```
jkwatson, line 114: small nit. if you have an entry like `=f` it's technically missing the key, not the value. ;)

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  Map<String, String> getCommaSeparatedMap(String name) {
+    return getCommaSeparatedValues(name).stream()
+        .map(
+            keyValuePair ->
+                Arrays.stream(keyValuePair.split("=", 2))
+                    .map(String::trim)
+                    .filter(s -> !s.isEmpty())
+                    .collect(Collectors.toList()))
```
jkwatson, line 109: These 4 lines of code could be extracted to a method like "filterBlanksAndNulls" and re-used above at line 96-99

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigurationException.java
```diff
@@ -0,0 +1,16 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.autoconfigure;
+
+/** An exception that is thrown if the user-provided configuration is invalid. */
+public final class ConfigurationException extends RuntimeException {
+
+  private static final long serialVersionUID = 4717640118051490483L;
```
jkwatson, line 11: I don't think this is necessary or useful. No one should be using default serialization for anything, anywhere.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
```diff
@@ -0,0 +1,116 @@
...
+
+  static boolean configureExporter(
+      String name,
+      ConfigProperties config,
+      boolean metricsAlreadyRegistered,
+      SdkMeterProvider meterProvider) {
+    switch (name) {
+      case "otlp":
+      case "otlp_metrics":
+        if (metricsAlreadyRegistered) {
```
jkwatson, line 27: it's a little annoying that this same check has to be done twice, but I don't see any easy away around it without some sort of ugly nesting of switch statements

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
```diff
@@ -0,0 +1,116 @@
...
+          throw new ConfigurationException(
+              "Multiple metrics exporters configured. Only one metrics exporter can be "
+                  + "configured at a time.");
+        }
+        configurePrometheusMetrics(config, meterProvider);
+        return true;
+      default:
+        return false;
+    }
```
jkwatson, line 44: I guess you could do:
```
    switch (name) {
      case "otlp":
      case "otlp_metrics":
      case "prometheus":
        if (metricsAlreadyRegistered) {
          throw new ConfigurationException(
              "Multiple metrics exporters configured. Only one metrics exporter can be "
                  + "configured at a time.");
        }
        if ("prometheus".equals(name)) {
          configurePrometheusMetrics(config, meterProvider);
          return true;
        }
        configureOtlpMetrics(config, meterProvider);
        return true;
      default:
        return false;
    }
```

but I don't know if that's really much better.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
```diff
@@ -0,0 +1,80 @@
...
+
+    Set<String> exporterNames =
+        new LinkedHashSet<>(config.getCommaSeparatedValues("otel.exporter"));
+
+    configureMeterProvider(resource, exporterNames, config);
+
+    SdkTracerProvider tracerProvider =
+        TracerProviderConfiguration.configureTracerProvider(resource, exporterNames, config);
+
+    if (!exporterNames.isEmpty()) {
```
jkwatson, line 44: I really really don't like that the previous code mutates the set and we're relying on that fact here. That was super surprising to me...in fact, I was sure this line was a bug until I dug into the implementations. 

Even if it ends up being uglier, I'd very strongly prefer it if we didn't have hidden side-effects in the code like this.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java
```diff
@@ -0,0 +1,151 @@
...
+        ServiceLoader.load(SdkTracerProviderConfigurer.class)) {
+      configurer.configure(tracerProviderBuilder);
+    }
+
+    List<SpanExporter> spanExporters = new ArrayList<>();
+    for (String name : new ArrayList<>(exporterNames)) {
+      SpanExporter exporter = SpanExporterConfiguration.configureExporter(name, config);
+      if (exporter != null) {
+        spanExporters.add(exporter);
+        exporterNames.remove(name);
```
jkwatson, line 44: 👎

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfigurationTest.java
```diff
@@ -0,0 +1,26 @@
...
+
+class OpenTelemetrySdkAutoConfigurationTest {
+
+  @Test
+  void resourcePrioritizesUser() {
+    Resource resource =
+        OpenTelemetrySdkAutoConfiguration.configureResource(
+            ConfigProperties.createForTest(
+                Collections.singletonMap("otel.resource.attributes", "telemetry.sdk.name=test")));
+    assertThat(resource.getAttributes().get(SemanticAttributes.TELEMETRY_SDK_NAME))
```
jkwatson, line 23: I didn't even realize that this was a thing that end-users could do. Is this desirable behavior?

sdk-extensions/autoconfigure/src/testConfigError/java/io/opentelemetry/sdk/autoconfigure/ConfigErrorTest.java
```diff
@@ -0,0 +1,155 @@
...
+        .isInstanceOf(ConfigurationException.class)
+        .hasMessage("Unrecognized value for otel.exporter: cat,dog");
+  }
+
+  @Test
+  @SetSystemProperty(key = "otel.exporter", value = "bar")
+  void globalOpenTelemetryWhenError() {
+    Logger logger = Logger.getLogger(GlobalOpenTelemetry.class.getName());
+    AtomicReference<LogRecord> logged = new AtomicReference<>();
+    Handler handler =
```
jkwatson, line 113: It'll be nice to convert to the logunit code!

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
anuraaga: Yeah I mention it in a spec issue too :)

https://github.com/open-telemetry/opentelemetry-specification/pull/1318#discussion_r551854325

Do you see an alternative? Should I mutate a `handledExportNames` with `.add` instead?

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
anuraaga: Heh - I'm going to have trouble getting used to fields being so far from the constructor :P

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigurationException.java
anuraaga: I think Java throws a warning which fails our build if I don't have it (agree no one should serialize it :)

warning: [serial] serializable class ConfigurationException has no definition of serialVersionUID

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfigurationTest.java
anuraaga: I think we'll need it for service name? I don't think a user should do something bad but it seems to make sense semantically to prioritize them if they do go to the effort.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
jkwatson: You could return an updated copy of the set, maybe?

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
trask: that's why static methods should go at the bottom of the class 😁

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
anuraaga: It's tricky since already returning a tracerprovider from one of them. It's a bit of duplication but I went with exposing the recognized names and checking eagerly instead of mutations.

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
anuraaga: Had no idea - do you think I should change the message? Wondering how realistic it is to disable public reflection vs adding noise to this code.

jkwatson: Maybe just "This could be a bug in OpenTelemetry" ?  Really, it doesn't matter all that much.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigurationException.java
jkwatson: that must be an errorprone warning. I can't imagine that javac would emit that.

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfigurationTest.java
jkwatson: I don't know if it's generally desirable or not, but it certainly is the way it works today, so have a test for it is good. :)

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
```diff
@@ -0,0 +1,79 @@
...
+    ContextPropagators propagators = PropagatorConfiguration.configurePropagators(config);
+
+    Resource resource = configureResource(config);
+
+    Set<String> exporterNames =
+        new LinkedHashSet<>(config.getCommaSeparatedValues("otel.exporter"));
+
+    Set<String> unrecognizedExporters = new LinkedHashSet<>(exporterNames);
+    unrecognizedExporters.removeAll(SpanExporterConfiguration.RECOGNIZED_NAMES);
+    unrecognizedExporters.removeAll(MetricExporterConfiguration.RECOGNIZED_NAMES);
```
jkwatson, line 41: this is a good solution. thanks!

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java
```diff
@@ -0,0 +1,150 @@
...
+      case "always_on":
+        return Sampler.alwaysOn();
+      case "always_off":
+        return Sampler.alwaysOff();
+      case "traceidratio":
+        {
+          Double ratio = config.getDouble("otel.trace.sampler.arg");
+          if (ratio == null) {
+            throw new ConfigurationException(
+                "otel.trace.sampler=traceidratio but otel.trace.sampler.arg is not provided. "
```
jkwatson, line 124: stuff like this is a good argument for extracting the keys into constants, but that can be done as a separate step later, for sure.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
```diff
@@ -0,0 +1,133 @@
...
+class ConfigProperties {
+
+  private final Map<String, String> config;
+
+  static ConfigProperties get() {
+    return new ConfigProperties(System.getProperties(), System.getenv());
+  }
+
+  // Visible for testing
+  @SuppressWarnings({"unchecked", "rawtypes"})
```
jkwatson, line 28: I don't think this supression is needed any more.

jkwatson, line 28:
```suggestion
```

jkwatson (approved):
Thanks!

PR #2402, opened 2020-12-23
Remove usage of updateTraceConfig in tests.

anuraaga:


sdk/trace/src/test/java/io/opentelemetry/sdk/trace/TestUtils.java
```diff
@@ -57,41 +51,4 @@ public static SpanData makeBasicSpan() {
...
         .build();
   }
-
-  /**
-   * Create a very basic SpanData instance, suitable for testing. It has the bare minimum viable
-   * data.
-   *
-   * @return A SpanData instance.
-   */
-  public static SpanBuilder startSpanWithSampler(
```
jkwatson, line 67: very glad we can get rid of this. 👍

jkwatson (approved):
Nice!

PR #2408, opened 2020-12-25
Add some test coverage to api module

anuraaga:


api/all/src/main/java/io/opentelemetry/api/internal/ImmutableKeyValuePairs.java
```diff
@@ -29,9 +29,7 @@
 @Immutable
 public abstract class ImmutableKeyValuePairs<K, V> {
 
-  protected List<Object> data() {
-    return Collections.emptyList();
```
anuraaga, line 33: This was only needed because `ImmutableBaggage` wasn't abstract which it should be

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/W3CBaggagePropagator.java
```diff
@@ -40,7 +40,7 @@ private W3CBaggagePropagator() {}
   @Override
   public <C> void inject(Context context, C carrier, Setter<C> setter) {
     Baggage baggage = Baggage.fromContext(context);
-    if (baggage == null) {
+    if (baggage.isEmpty()) {
```
anuraaga, line 43: Never null

jkwatson (approved):
Thanks!

PR #2409, opened 2020-12-28
Fix README tracing and baggage api links

kacperfronc:
README seems to have two broken links, this small change fixes them


anuraaga (approved):
Thanks! Sorry for the bad links

PR #2410, opened 2020-12-28
fix gradle version to 6.7.1

shalk:


anuraaga (approved):
Thanks!

build.gradle
```diff
@@ -556,7 +556,7 @@ subprojects {
 }
 
 wrapper {
-    gradleVersion = '6.7'
+    gradleVersion = '6.7.1'
```
Oberon00, line 559: Note that this does not actually update the wrapper. To update the wrapper you'd need to change `gradle/wrapper/gradle-wrapper.properties` like done for the examples.

I wonder if we should remove this block. Wouldn't it default to the current gradle version otherwise? That is exactly what we want.

anuraaga: Oof sorry - on phone couldn't see the path was for the examples.

Yeah I think we can remove the block, otherwise you have to run the wrapper task twice anyways to make sure the gradlew script is actually updated.

PR #2413, opened 2020-12-28
Change MetricData to expose the Point type that is expected for each data type

bogdandrutu:
Depends on https://github.com/open-telemetry/opentelemetry-java/pull/2412

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregation/DoubleLastValueAggregation.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregation/DoubleLastValueAggregation.java
@@ -7,12 +7,9 @@ package io.opentelemetry.sdk.metrics.aggregation;
 
 import io.opentelemetry.api.common.Labels;
 import io.opentelemetry.sdk.common.InstrumentationLibraryInfo;
-import io.opentelemetry.sdk.metrics.accumulation.Accumulation;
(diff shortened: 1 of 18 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregation/DoubleLastValueAggregation.java
```diff
@@ -0,0 +1,72 @@
...
+        return MetricData.createDoubleGauge(
+            resource,
+            instrumentationLibraryInfo,
+            descriptor.getName(),
+            descriptor.getDescription(),
+            descriptor.getUnit(),
+            MetricData.DoubleGaugeData.create(points));
+      case COUNTER:
+      case UP_DOWN_COUNTER:
+      case VALUE_RECORDER:
```
jkwatson, line 63: this could be replaced with a single `default:`

jkwatson (approved):

anuraaga (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregation/DoubleLastValueAggregation.java
bogdandrutu: I tend to not do that so when we add a new option it forces me to update all switch statements and think about the implications, if I add default etrorprone will not complain when adding a new value

PR #2416, opened 2020-12-30
Add CodeQL security scans

amanbrar1999:
## Motivation
Follow up to issue https://github.com/open-telemetry/oteps/issues/144

[CodeQL](https://github.com/github/codeql-action) is GitHub's static analysis engine which scans repos for security vulnerabilities. As the project grows and we near GA it might be useful to have a workflow which checks for security vulnerabilities with every PR so we can ensure every incremental change is following best development practices. Also passing basic security checks will also make sure that there aren't any glaring issues for our users.
## Changes
* This PR adds [CodeQL](https://github.com/github/codeql-action) security checks to the repo
* After every run the workflow uploads the results to GitHub. Details on the run and security alerts will show up in the `security` tab of this repo.

**Workflow Triggers**
* daily cron job at 1:30am
* workflow_dispatch (in case maintainers want to trigger a security check manually)

cc- @alolita

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -33,28 +33,12 @@ jobs:
           languages: java
 
       # Build the code
-      # - uses: actions/checkout@v2
-      #   with:
-      #     fetch-depth: 0
-      # - id: setup-java-8
-      #   name: Setup Java 8
-      #   uses: actions/setup-java@v1
-      #   with:
-      #     java-version: 8
-      #     java-package: jre
       - name: Setup Java 11
         uses: actions/setup-java@v1
         with:
           java-version: 11
 
       - run: ./gradlew build --stacktrace jacocoTestReport
-      # - uses: burrunan/gradle-cache-action@v1.6
-      #   with:
-      #     remote-build-cache-proxy-enabled: false
-      #     arguments: build --stacktrace jacocoTestReport
-      #     properties: |
-      #       testAdditionalJavaVersions=true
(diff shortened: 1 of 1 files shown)


jkwatson (approved):

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,42 @@
+name: "Code Scanning - Action"
+
+on:
+  workflow_dispatch:
+  schedule:
+    #        ┌───────────── minute (0 - 59)
```
anuraaga, line 6: Let's remove these they add some noise, and we have a few cron workflows without the comments so lose some consistency.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,42 @@
...
+
+jobs:
+  CodeQL-Build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      # Initializes the CodeQL tools for scanning.
```
anuraaga, line 25: Don't need any of these comments they repeat the step names.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v1
+        with:
+          languages: java
+
+      # Build the code
```
anuraaga, line 31: Incidentally not correct - so yeah don't need these comments :)

anuraaga (commented):
Thanks! Just some cleanup nits.

anuraaga (approved):

PR #2417, opened 2021-01-04
Delegate constant traceidratio samplers to the constant implementations.

anuraaga:
Should be equivalent with a bit better performance.


anuraaga:
Realized we have an attribute set by the ratio sampler too so it's not a 1:1 equivalency.

Oberon00:
@anuraaga Actually I think we should remove that attribute, as it is off-spec (https://github.com/open-telemetry/opentelemetry-specification/pull/570 was never merged). Should I create an issue for that?

anuraaga:
Ah sure sounds good

Oberon00:
Created #2420.

PR #2418, opened 2021-01-04
Remove formatting checks from checkstyle

anuraaga:
I found `LeftCurly` is incompatible with formatting produced by googlejavaformat some times. But anyways, we don't need checkstyle to look at any sort of formatting since it's handled by googlejavaformat / spotless so we may as well prevent possibility of conflicts and get a bit of checkstyle performance.


jkwatson (approved):
Thanks!

PR #2421, opened 2021-01-04
prevent workflow 'Publish snapshots to JFrog' in fork repo

shalk:
Because  github [enable actions by default](https://github.blog/changelog/2020-12-15-github-actions-environments-environment-protection-rules-and-environment-secrets-beta/) ,every time I sync my repo with upstream, it will run actions ,thgen I get notifications about some jobs fail.

Though I can disable action in settings，but i  think some job will never run in fork repo like ''Publish snapshots to JFrog' .

So i add `if condition` here


jkwatson (approved):
This seems reasonable to me. I do also recommend disabling actions in your fork, though.

PR #2422, opened 2021-01-04
Change the SdkTracerManagement shutdown method to return CompletableResultCode

jkwatson:
This makes the shutdown method more consistent across the SDK.

### DIFF
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkTracerManagement.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkTracerManagement.java
@@ -64,8 +64,11 @@ public interface SdkTracerManagement {
    * <p>After this is called, further attempts at re-using or reconfiguring this instance will
    * result in undefined behavior. It should be considered a terminal operation for the SDK
    * implementation.
+   *
+   * @return a {@link CompletableResultCode} which is completed when all the span processors have
+   *     been shut down.
    */
-  void shutdown();
+  CompletableResultCode shutdown();
 
   /**
    * Requests the active span processor to process all span events that have not yet been processed
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/TracerSharedState.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/TracerSharedState.java
@@ -29,7 +29,7 @@ final class TracerSharedState {
 
   @GuardedBy("lock")
   @Nullable
-  private volatile CompletableResultCode isStopped = null;
+  private volatile CompletableResultCode shutdownResult = null;
 
   @GuardedBy("lock")
   private final List<SpanProcessor> registeredSpanProcessors;
@@ -106,7 +106,7 @@ final class TracerSharedState {
    */
   boolean isStopped() {
     synchronized (lock) {
-      return isStopped != null && isStopped.isSuccess();
+      return shutdownResult != null && shutdownResult.isSuccess();
     }
   }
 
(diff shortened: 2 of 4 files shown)


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkTracerManagement.java
```diff
@@ -64,8 +64,11 @@
    * <p>After this is called, further attempts at re-using or reconfiguring this instance will
    * result in undefined behavior. It should be considered a terminal operation for the SDK
    * implementation.
+   *
+   * @return a {@link CompletableResultCode} which is completed when all the span processors have
+   *     been shut down.
    */
-  void shutdown();
+  CompletableResultCode shutdown();
```
bogdandrutu, line 71: Does this make impossible to change to `close` and implement `AutoClosable`? Do we care about that?

jkwatson: It shouldn't. The goal here was to make it so that we could have 2 modes of "closing". One is "shutdown()" like we have today, which will universally return `CompletableResultCode`, and then the `close` method proposed in #2285 as the auto-closable version.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/TracerSharedState.java
```diff
@@ -24,8 +25,11 @@
   // Reads and writes are atomic for reference variables. Use volatile to ensure that these
   // operations are visible on other CPUs as well.
   private volatile Supplier<TraceConfig> traceConfigSupplier;
-  private volatile SpanProcessor activeSpanProcessor = NoopSpanProcessor.getInstance();
-  private volatile boolean isStopped = false;
+  private volatile SpanProcessor activeSpanProcessor;
+
+  @GuardedBy("lock")
+  @Nullable
+  private volatile CompletableResultCode isStopped = null;
```
anuraaga, line 32:
```suggestion
  private volatile CompletableResultCode stopResult = null;
```

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/TracerSharedState.java
```diff
@@ -101,19 +105,24 @@ void addSpanProcessor(SpanProcessor spanProcessor) {
...
   }
 
   /**
    * Stops tracing, including shutting down processors and set to {@code true} {@link #isStopped()}.
+   *
+   * @return a {@link CompletableResultCode} that will be completed when the span processor is shut
+   *     down.
    */
-  void stop() {
+  CompletableResultCode stop() {
```
anuraaga, line 119: Maybe it's worth renaming this to `shutdown` for consistency (even though it's an internal method)

sdk/trace/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorTest.java
```diff
@@ -45,7 +46,14 @@
...
   private SdkTracerProvider sdkTracerProvider;
   private final BlockingSpanExporter blockingSpanExporter = new BlockingSpanExporter();
-  @Mock private SpanExporter mockServiceHandler;
+
+  @Mock(lenient = true)
+  private SpanExporter mockServiceHandler;
+
+  @BeforeEach
+  void setUp() {
+    when(mockServiceHandler.shutdown()).thenReturn(CompletableResultCode.ofSuccess());
```
anuraaga, line 55: Maybe add to the test method instead of using lenient?

anuraaga (approved):

sdk/trace/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorTest.java
jkwatson: I can look into it. It wasn't actually clear what test needed this. :)

jkwatson: I figured it out. This is one reason I don't like having mocks as fields if they're not used in every test. :(

anuraaga: From what I understand it's the use of `when` in beforeeach that triggers the strict checking, not the presence of a field. But no worries either way.

anuraaga (approved):

PR #2423, opened 2021-01-04
Move the semantic conventions generated code into its own module

### LINKED ISSUE #2264: Extract the semantic conventions classes into their own module.
The [OTEP for versioning](https://github.com/open-telemetry/oteps/pull/143) strongly suggests that semantic conventions should be considered a separate module. They will probably evolve more quickly than other parts of the API, and it makes sense to be able to consume them independently.

jkwatson:
Since the semantic conventions may be versioned independently of the rest of the API, it seems like a good idea to move them to their own module before we hit 1.0. 

Resolves #2264


sdk/trace/build.gradle
```diff
@@ -12,6 +12,7 @@ ext.propertiesDir = "build/generated/properties/io/opentelemetry/sdk/trace"
 
 dependencies {
     api project(':api:all'),
+            project(':api:semconv'),
```
bogdandrutu, line 15: maybe just `semconv`? Not under `api`?

bogdandrutu (approved):

sdk/trace/build.gradle
jkwatson: I wasn't really sure which was better. Looking at the specs, it looks like semantic conventions are described under a section called "data", so I agree that it probably doesn't make sense to have it under API here, either.  👍

Oberon00 (approved):

README.md
```diff
@@ -16,12 +16,12 @@ project.
 This project contains the following top level components:
 
 * [OpenTelemetry API](api/):
-  * [tracing api](api/all/src/main/java/io/opentelemetry/api/trace/) includes `Tracer`, `Span` and `SpanContext`.
-  * [baggage api](api/all/src/main/java/io/opentelemetry/api/baggage/) defines a collection of entries in the form of key-value pairs of data that can be propagated to provide contextual information.
-  * [metrics api](api/metrics/src/main/java/io/opentelemetry/api/metrics/).
-  * [context api](api/context/src/main/java/io/opentelemetry/context/).
+  * [stable apis](api/all/src/main/java/io/opentelemetry/api/all/) including `Tracer`, `Span`, `SpanContext`, and `Baggage`
+  * [semantic conventions](api/semconv/) Generated code for the OpenTelemetry semantic conventions.
```
anuraaga, line 20:
```suggestion
  * [semantic conventions](semconv/) Generated code for the OpenTelemetry semantic conventions.
```

anuraaga (approved):
I'm hoping we won't have to version it differently since it seems like it could confusing for one artifact to have its own version (otherwise, we would wonder why we don't do the same for other modules like `context).

But a separate artifact makes sense anyways since we can expect backends to just need proto and semconv, and not API.

jkwatson:
> I'm hoping we won't have to version it differently since it seems like it could confusing for one artifact to have its own version (otherwise, we would wonder why we don't do the same for other modules like `context).


Oh for sure. I really don't want to, but if we need to, it would only be possible if it's broken out. :)

bogdandrutu (approved):
Should the package be `io.opentelemetry.semconv`? Without `api`?

jkwatson:
> Should the package be `io.opentelemetry.semconv`? Without `api`?

oh yes. that does make sense. good eye! I'll regenerate the file.

jkwatson:
ok, I have repackaged the `SemanticAttributes` into `io.opentelemetry.semconv`, and deprecated the ones in the api package. Thanks for the reviews!

semconv/build.gradle
```diff
@@ -0,0 +1,16 @@
+plugins {
+    id "java-library"
+    id "maven-publish"
+
+    id "me.champeau.gradle.jmh"
```
anuraaga, line 5:
```suggestion
```

semconv/build.gradle
```diff
@@ -0,0 +1,16 @@
...
+    id "ru.vyarus.animalsniffer"
+}
+
+description = 'OpenTelemetry Semantic Conventions'
+ext.moduleName = "io.opentelemetry.api.semconv"
```
anuraaga, line 10:
```suggestion
ext.moduleName = "io.opentelemetry.semconv"
```

anuraaga (approved):
Found just small nits

PR #2425, opened 2021-01-05
Use logunit instead of tedious jul handler modifications for testing

anuraaga:
Been looking for this and finally found it!


jkwatson (approved):
Thanks! This is cleaner, for sure.

PR #2161, opened 2020-11-30
Migrate docker image mirroring to ghcr.

anuraaga:
Will change the references in integration tests after merging / running

PR #2426, opened 2021-01-05
Use health port for jaeger container startup check.

### LINKED ISSUE #1871: Jaeger (and presumably similar) testcontainers test flaky
Lately I see jaeger tests fail with a timeout relatively frequently - I've even noticed this on my machine, not just on CI. Docker Hub has been introducing pull limits from what I understand and perhaps it's affecting us. We probably want to cache `.docker` or whatever to reduce docker pulls because of the rate limiting, though it wouldn't solve the problem for a new contributor. Another option is to rehost on ghcr or bintray, but that seems not great either.

anuraaga:
Also migrate images to the ghcr ones I had set up but forgot to apply the change for.

Fixes #1871 (hopefully, will reopen if this doesn't actually fix it)

### DIFF
--- a/perf-harness/src/test/java/io/opentelemetry/perf/OtlpPipelineStressTest.java
+++ b/perf-harness/src/test/java/io/opentelemetry/perf/OtlpPipelineStressTest.java
@@ -67,7 +67,7 @@ public class OtlpPipelineStressTest {
   public static GenericContainer<?> collectorContainer =
       new GenericContainer<>(
               DockerImageName.parse(
-                  "open-telemetry-docker-dev.bintray.io/java-test-containers:otel-collector-dev"))
+                  "ghcr.io/open-telemetry/java-test-containers:otel-collector-dev"))
           .withNetwork(network)
           .withNetworkAliases("otel-collector")
           .withExposedPorts(OTLP_RECEIVER_PORT)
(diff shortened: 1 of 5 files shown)


perf-harness/src/test/java/io/opentelemetry/perf/OtlpPipelineStressTest.java
```diff
@@ -67,7 +67,7 @@
   public static GenericContainer<?> collectorContainer =
       new GenericContainer<>(
               DockerImageName.parse(
-                  "open-telemetry-docker-dev.bintray.io/java-test-containers:otel-collector-dev"))
+                  "ghcr.io/open-telemetry/java-test-containers:otel-collector-dev"))
```
Oberon00, line 70: This change seems unrelated to the PR purpose. What is the reason for using a different registry?

anuraaga: For a long time ever since starting to use bintray for agent smoke tests, we found the push reliability to be close to 0. GHCR has yet to fail :)

I had already updated our mirror workflow

https://github.com/open-telemetry/opentelemetry-java/pull/2161

But forgot to update the tests after that run once.

It's indeed not strongly related to the purpose but touches the same lines mostly, please let me be a bit lazy with it still being a small PR :D

Oberon00 (approved):

breedx-splk:
@anuraaga Thanks for the quick response to my findings with this quick fix!  🏆

PR #2427, opened 2021-01-05
Update dependencies

anuraaga:


api/context/build.gradle
```diff
@@ -28,9 +28,9 @@ dependencies {
     grpcInOtelTestImplementation libraries.grpc_context
     otelInGrpcTestImplementation libraries.grpc_context
 
-    braveInOtelTestImplementation "io.zipkin.brave:brave:5.13.1"
-    otelAsBraveTestImplementation "io.zipkin.brave:brave:5.13.1"
-    otelInBraveTestImplementation "io.zipkin.brave:brave:5.13.1"
+    braveInOtelTestImplementation "io.zipkin.brave:brave:5.13.3"
```
Oberon00, line 31: I think we should define that version once in the shared build.gradle, with the other versions. Also, at least for test dependencies, I think using `5.13.+` would be better.

build.gradle
```diff
@@ -157,25 +157,25 @@ subprojects {
...
-            grpcVersion = '1.33.1'
-            guavaVersion = '30.0-android'
-            jacksonVersion = '2.11.3'
+            grpcVersion = '1.34.1'
+            guavaVersion = '30.1-android'
+            jacksonVersion = '2.12.0'
             jmhVersion = '1.26'
             junitVersion = '5.7.0'
-            mockitoVersion = '3.6.0'
+            mockitoVersion = '3.7.0'
```
Oberon00, line 165: Same here: IMHO we should use a `+` for the patch version of test-only dependencies.

Oberon00 (approved):

build.gradle
bogdandrutu: This is not reproducible, I would encourage not to use `+` in any of the dependencies. It is very important to have reproducible builds and tests.

Oberon00: I think there is a gradle feature for that https://docs.gradle.org/current/userguide/dependency_locking.html, but I haven't looked into it much yet.

jkwatson (approved):
👍

build.gradle
anuraaga: Agree that dynamic versions without locking isn't a good idea. And since we'd want to bump on minor versions anyways, I think bumping manually on patch versions isn't a huge deal and keeps the test dependencies consistent with non-test.

PR #2428, opened 2021-01-05
Have projects set their release state to alpha instead of recognizing…

anuraaga:
… word metrics

We will apply alpha to many more modules for sure, most sdk-extensions I think.


jkwatson:
this seems like a good upgrade. And yes, there are several things that we should probably tag as alpha before release. I can follow up with that PR.

jkwatson (approved):
Thanks!

PR #2429, opened 2021-01-05
Allow setting trusted certs for OTLP Trace exporter

anuraaga:
While working on autoconfiguration, I noticed we support setting TLS but very few people will actually be able to use it since we don't support setting trusted certs. Now they can be set programattically - will add support for the env variable in a followup, and add to metrics exporter in a followup.

### DIFF
--- a/exporters/otlp/trace/build.gradle
+++ b/exporters/otlp/trace/build.gradle
@@ -3,14 +3,24 @@ plugins {
     id "maven-publish"
 
     id "me.champeau.gradle.jmh"
+    id "org.unbroken-dome.test-sets"
     id "ru.vyarus.animalsniffer"
 }
 
 description = 'OpenTelemetry Protocol Trace Exporter'
 ext.moduleName = "io.opentelemetry.exporter.otlp.trace"
 
+testSets {
+    testGrpcNetty
+    testGrpcNettyShaded
+    testGrpcOkhttp
+}
+
 dependencies {
     api project(':sdk:trace')
+
+    compileOnly "io.grpc:grpc-netty"
+    compileOnly "io.grpc:grpc-netty-shaded"
 
     implementation project(':sdk-extensions:otproto'),
             libraries.grpc_api,
@@ -19,11 +29,25 @@ dependencies {
             libraries.protobuf
 
     testImplementation project(':sdk:testing'),
-            'com.linecorp.armeria:armeria-grpc',
-            'com.linecorp.armeria:armeria-junit5',
             "io.grpc:grpc-testing:${grpcVersion}"
 
-    testRuntime "io.grpc:grpc-netty-shaded:${grpcVersion}"
+    testGrpcNettyImplementation 'com.linecorp.armeria:armeria-grpc',
+            'com.linecorp.armeria:armeria-junit5'
+
+    testGrpcNettyShadedImplementation 'com.linecorp.armeria:armeria-grpc',
+            'com.linecorp.armeria:armeria-junit5'
+
+    testGrpcOkhttpImplementation 'com.linecorp.armeria:armeria-grpc',
+            'com.linecorp.armeria:armeria-junit5'
+
+    testGrpcNettyRuntimeOnly "io.grpc:grpc-netty:${grpcVersion}",
+            libraries.slf4jsimple
+
+    testGrpcNettyShadedRuntimeOnly "io.grpc:grpc-netty-shaded:${grpcVersion}",
+            libraries.slf4jsimple
+
+    testGrpcOkhttpRuntimeOnly "io.grpc:grpc-okhttp:${grpcVersion}",
+            libraries.slf4jsimple
 
     jmh project(':sdk:testing')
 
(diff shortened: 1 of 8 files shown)


exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterBuilder.java
```diff
@@ -114,6 +129,51 @@ public OtlpGrpcSpanExporter build() {
         managedChannelBuilder.intercept(MetadataUtils.newAttachHeadersInterceptor(metadata));
       }
 
+      if (trustedCertificatesPem != null) {
+        // gRPC does not abstract TLS configuration so we need to check the implementation and act
+        // accordingly.
+        if (managedChannelBuilder
```
anuraaga, line 135: This horror is all I could find :(

exporters/otlp/trace/build.gradle
```diff
@@ -3,27 +3,51 @@ plugins {
...
             libraries.grpc_protobuf,
             libraries.grpc_stub,
             libraries.protobuf
 
     testImplementation project(':sdk:testing'),
-            'com.linecorp.armeria:armeria-grpc',
-            'com.linecorp.armeria:armeria-junit5',
             "io.grpc:grpc-testing:${grpcVersion}"
 
-    testRuntime "io.grpc:grpc-netty-shaded:${grpcVersion}"
```
anuraaga, line 26: Since test sets automatically inherit the main test configuration, I needed to move this and the corresponding test out.

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterBuilder.java
```diff
@@ -114,6 +129,51 @@ public OtlpGrpcSpanExporter build() {
...
+            .getClass()
+            .getName()
+            .equals("io.grpc.netty.NettyChannelBuilder")) {
```
Oberon00, line 138: Interesting, I was wondering why you didn't use `instanceof` and then I found that `instanceof` does load the class: https://bugs.openjdk.java.net/browse/JDK-6592631

breedx-splk: I also didn't realize that `instanceof` caused classload, so this is interesting!  Thanks for sharing that @Oberon00 . 
 
Is the goal here really to prevent loading one of two classes, though?  It's definitely less readable and more fragile this way, so maybe I'm missing some some other awful byproduct of the classload?

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterBuilder.java
```diff
@@ -114,6 +129,51 @@ public OtlpGrpcSpanExporter build() {
...
+                    + "X.509 in PEM format?",
+                e);
+          }
+        } else {
+          throw new IllegalStateException(
+              "TLS cerificate configuration only supported with Netty. "
+                  + "If you need to configure a certificate, switch to grpc-netty or "
+                  + "grpc-netty-shaded.");
+        }
+        // TODO(anuraaga): Support okhttp.
```
jkwatson, line 174: can you add an issue for this, so we don't forget?

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterBuilder.java
jkwatson: Isn't the real reason for this to make sure we don't need to pull in these dependencies into the project, not to actually save class loading time/memory?

breedx-splk: I could believe that.

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterBuilder.java
anuraaga: Sure filed https://github.com/open-telemetry/opentelemetry-java/issues/2430

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterBuilder.java
anuraaga: Yeah user brings in the gRPC implementation they want to use (usually one of grpc-netty or grpc-netty-shaded)

jkwatson (approved):
Thanks!

PR #2431, opened 2021-01-06
Replace max-workers in favor of priority setting, enable local build …

anuraaga:
…caching, and remove unnecessary Xmx

We've been using these settings in instrumentation repo and they seem to work fairly well.

Caching enables the local build cache, so builds don't have to rerun after cleans (btw our CI already uses local build cache).

And we don't need to set Xmx4096 for Gradle's jvmargs since that is only for the wrapper, the Gradle daemon which actually does builds can use a lot of memory.


jkwatson (approved):
Not sure if I've seen any issues related to this, but this seems fine to me.

PR #2432, opened 2021-01-06
Remove sampling probability attribute when was not accepted to spec.

### LINKED ISSUE #2420: Remove off-spec sampling.probability from TraceIdRatioBasedSampler attribute (for now)
This was added in #1116 (for #1087) but the follow-up spec PR open-telemetry/opentelemetry-specification#570 was not merged. So I propose to remove this attribute from the TraceIdRatioBasedSampler for GA. The final spec definition might be different from what we have now, and I think other languages also don't implement this.

See also https://github.com/open-telemetry/opentelemetry-java/pull/2417#issuecomment-753880089

anuraaga:
If anyone is relying on this currently, it's not too difficult to use a custom sampler implementation until the spec is updated.

Fixes #2420


Oberon00 (approved):

jkwatson (approved):
Thanks!

PR #2433, opened 2021-01-06
Have a default span name in case instrumentation sends in a blank/empty/null name

### LINKED ISSUE #2424: The SDK will raise an NPE to the client code if span name is null.
**Describe the bug**
The SDK should never adversely impact (aside from possible performance implications) client code. If instrumentation sets a null span name, an NPE is thrown by the SDK when trying to convert to SpanData for the export pipeline. This NPE ends up in the client's application code, which is not acceptable.

**Steps to reproduce**
Install the SDK, start and end a span with a null `name`.

Maybe a logged warning, maybe nothing? I'm honestly not sure exactly what is the correct behavior here. Should the Span be sent to the span processor with a null name? Should it be silently dropped?

**What did you see instead?**
A NullPointerException interrupting client code.

OTel API & SDK 0.13.1.

jkwatson:
Resolves #2424


carlosalberto (approved):

carlosalberto:
I feel like this is something we should have clarified in the Spec itself, but 'sadly' now it's frozen. +1 to this change.

anuraaga (approved):

PR #2436, opened 2021-01-06
Mark a few more modules as alpha release status

jkwatson:
I got 2 more here that I don't think we should declare as stable for 1.0. Any others that people think we aren't ready to support forever yet?

### DIFF
--- a/opencensus-shim/gradle.properties
+++ b/opencensus-shim/gradle.properties
@@ -1,6 +1,1 @@
-#
-# Copyright The OpenTelemetry Authors
-# SPDX-License-Identifier: Apache-2.0
-#
-
 otel.release=alpha
--- a/opencensus-shim/gradle.properties
+++ b/opencensus-shim/gradle.properties
@@ -1,0 +1,6 @@
+#
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+#
+
+otel.release=alpha
--- a/opentracing-shim/gradle.properties
+++ b/opentracing-shim/gradle.properties
@@ -1,6 +1,1 @@
-#
-# Copyright The OpenTelemetry Authors
-# SPDX-License-Identifier: Apache-2.0
-#
-
 otel.release=alpha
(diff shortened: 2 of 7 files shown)


carlosalberto (approved):

anuraaga:
async-processor - don't think it's actually been tested well, want to encourage BSP unless we can properly benchmark / compare them. If disruptor doesn't really help even at high load, we would delete it I think.

jfr-events - implicitly has a public API in terms of the data it exports to JFR. We need more feedback on if this data works well / is useful before we can commit to it

zpages - need to rework it to support non-global OTel SDK. Not a 1.0 priority for me though

opencensus / opentracing shim I'd also like confirmation that they work well and as expected in apps before marking them stable if possible as they're pretty complicated pieces. I'm actually a bit concerned by the opentracing shim since there was significant work done in OC shim to support bidirectional compat in #2060 which I don't think the OT shim is doing, is it ok?

jkwatson:
> async-processor - don't think it's actually been tested well, want to encourage BSP unless we can properly benchmark / compare them. If disruptor doesn't really help even at high load, we would delete it I think.
> 
> jfr-events - implicitly has a public API in terms of the data it exports to JFR. We need more feedback on if this data works well / is useful before we can commit to it
> 
> zpages - need to rework it to support non-global OTel SDK. Not a 1.0 priority for me though
> 
> opencensus / opentracing shim I'd also like confirmation that they work well and as expected in apps before marking them stable if possible as they're pretty complicated pieces. I'm actually a bit concerned by the opentracing shim since there was significant work done in OC shim to support bidirectional compat in #2060 which I don't think the OT shim is doing, is it ok?

I asked @jsuereth about the OC shim stuff this morning and he's going to split the metrics out and we can mark that one alpha as a part of that effort. 

We definitely need to tackle the OT shim tasks that are still outstanding if we want that to not be alpha, for sure.

I'll add those ones to this PR as well. Thanks for the thoughts!

bogdandrutu (approved):

anuraaga (approved):

PR #2439, opened 2021-01-06
Add a Duration parser for configuration.

### LINKED ISSUE #2435: Unit-Parser for configuration
Create a way to parse units in configuration parameters, of limited syntax.

E.g. `1s` would parse into 1 second duration, and `10ms` would be 10 milliseconds.

A set of methods to pull a string in a specific format with known unit-conversion.

```
class ConfigParser {
  static long getDurationInMillis(String rawValue) {...}
  static long getSizeInBytes(String rawValue) {...}
}
```

We could just not allow convenient unit syntax in config.

**Additional context**
See: https://github.com/open-telemetry/opentelemetry-specification/issues/1312

jsuereth:
- Update ConfigBuilder in `sdk/common/export` to include a `getDurationMillisProperty` method.
- Add a short + limited duration parser that simply extract that "units" string and relies
  on substring efficiency (we assume config doesn't have to be fast as it's likely read about once)
- Add tests to verify various incarnations of duration are parsed to appropriate millisecond values.

Fixes #2435

This is my first PR to OTel-Java so sending this out for early review/evaluation. Want to verify the following:

- Is this the right location for this utility?
- Am I violating unit-test best practices?
- Do you want better documentation on the format, or can I try to contribute that directly to the speicfication?

### DIFF
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
@@ -132,7 +132,7 @@ public class ConfigProperties {
   }
 
   /**
-   * Get a duration property from the map, {@code null} if it cannot be found or it has a wrong
+   * Get a duration property from the map, or {@code null} if it cannot be found or it has a wrong
    * type.
    *
    * <p>Durations can be of the form "{number}{unit}", where unit is one of:
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
@@ -132,8 +132,8 @@ public class ConfigProperties {
   }
 
   /**
-   * Get a duration property from the map, or {@code null} if it cannot be found or it has a wrong
-   * type.
+   * Returns a duration property from the map, or {@code null} if it cannot be found or it has a
+   * wrong type.
    *
    * <p>Durations can be of the form "{number}{unit}", where unit is one of:
    *
@@ -154,7 +154,7 @@ public class ConfigProperties {
    */
   @Nullable
   @SuppressWarnings("UnusedException")
-  Duration getDuration(String name) {
+  public Duration getDuration(String name) {
     String value = config.get(name);
     if (value == null || value.isEmpty()) {
       return null;
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
@@ -5,6 +5,7 @@
 
 package io.opentelemetry.sdk.autoconfigure;
 
+import java.time.Duration;
 import java.util.AbstractMap;
 import java.util.Arrays;
 import java.util.Collections;
@@ -13,6 +14,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
@@ -130,6 +132,56 @@ public class ConfigProperties {
   }
 
   /**
+   * Get a duration property from the map, {@code null} if it cannot be found or it has a wrong
+   * type.
+   *
+   * <p>Durations can be of the form "{number}{unit}", where unit is one of:
+   *
+   * <ul>
+   *   <li>ms
+   *   <li>s
+   *   <li>m
+   *   <li>h
+   *   <li>d
+   * </ul>
+   *
+   * <p>If no unit is specified, milliseconds is the assumed duration unit.
+   *
+   * @param name The property name
+   * @return the {@link Duration} value of the property, {@code null} if the property cannot be
+   *     found.
+   * @throws ConfigurationException for malformed duration strings.
+   */
+  @Nullable
+  @SuppressWarnings("UnusedException")
+  Duration getDuration(String name) {
+    String value = config.get(name);
+    if (value == null || value.isEmpty()) {
+      return null;
+    }
+    String unitString = getUnitString(value);
+    // TODO: Environment variables have unknown encoding.  `trim()` may cut codepoints oddly
+    // but likely we'll fail for malformed unit string either way.
+    String numberString = value.substring(0, value.length() - unitString.length());
+    try {
+      long rawNumber = Long.parseLong(numberString.trim());
+      TimeUnit unit = getDurationUnit(unitString.trim());
+      return Duration.ofMillis(TimeUnit.MILLISECONDS.convert(rawNumber, unit));
+    } catch (NumberFormatException ex) {
+      throw new ConfigurationException(
+          "Invalid duration property "
+              + name
+              + "="
+              + value
+              + ". Expected number, found: "
+              + numberString);
+    } catch (ConfigurationException ex) {
+      throw new ConfigurationException(
+          "Invalid duration property " + name + "=" + value + ". " + ex.getMessage());
+    }
+  }
+
+  /**
    * Returns a map-valued configuration property. The format of the original value must be
    * comma-separated for each key, with an '=' separating the key and value. For instance, <code>
    * service.name=Greatest Service,host.name=localhost</code> Empty values will be removed.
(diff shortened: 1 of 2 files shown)


sdk/common/src/main/java/io/opentelemetry/sdk/common/export/ConfigBuilder.java
```diff
@@ -158,4 +159,89 @@ protected static Double getDoubleProperty(String name, Map<String, String> map)
...
+      String unitString = getDurationUnitString(value);
+      TimeUnit units = getDurationUnit(unitString);
+      String numberString = value.substring(0, value.length() - unitString.length()).trim();
+      long rawNumber = Long.parseLong(numberString);
+      return TimeUnit.MILLISECONDS.convert(rawNumber, units);
+    } catch (NumberFormatException | NullPointerException ex) {
+      return null;
+    }
+  }
+  /** Returns the TimeUnit associated with a unit string. Defaults to milliseconds. */
```
jkwatson, line 195:
```suggestion

  /** Returns the TimeUnit associated with a unit string. Defaults to milliseconds. */
```

sdk/common/src/main/java/io/opentelemetry/sdk/common/export/ConfigBuilder.java
```diff
@@ -158,4 +159,89 @@ protected static Double getDoubleProperty(String name, Map<String, String> map)
...
+    while (i >= 0) {
+      char c = rawValue.charAt(i);
+      if (!isDurationUnitToken(c)) {
+        break;
+      }
+      i -= 1;
+    }
+    return rawValue.substring(i + 1);
+  }
+  /** Returns true if the given character is allowable as a unit of duration of size. */
```
jkwatson, line 230:
```suggestion

  /** Returns true if the given character is allowable as a unit of duration of size. */
```

sdk/common/src/main/java/io/opentelemetry/sdk/common/export/ConfigBuilder.java
```diff
@@ -158,4 +159,91 @@ protected static Double getDoubleProperty(String name, Map<String, String> map)
       return null;
     }
   }
+
+  /**
+   * Get a duration property from the map, {@code null} if it cannot be found or it has a wrong
```
anuraaga, line 164: Thanks a lot for the help! Configuration is still in flux, and this functionality just moved to a different location. Can you put it there instead?

https://github.com/open-telemetry/opentelemetry-java/blob/master/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java

sdk/common/src/main/java/io/opentelemetry/sdk/common/export/ConfigBuilder.java
```diff
@@ -158,4 +159,91 @@ protected static Double getDoubleProperty(String name, Map<String, String> map)
...
+   *   <li>d
+   * </ul>
+   *
+   * @param name The property name
+   * @param map The map where to look for the property
+   * @return the {@link Long} value of the property in milliseconds, {@code null} in case of error
+   *     or if the property cannot be found.
+   */
+  @Nullable
+  protected static Long getDurationMillisProperty(String name, Map<String, String> map) {
```
anuraaga, line 183: I guess we could just return a `Duration`

sdk/common/src/main/java/io/opentelemetry/sdk/common/export/ConfigBuilder.java
```diff
@@ -158,4 +159,91 @@ protected static Double getDoubleProperty(String name, Map<String, String> map)
...
+    }
+    throw new NumberFormatException("Invalid duration string, found: " + unitString);
+  }
+
+  /**
+   * Fragments the 'units' portion of a config value from the 'value' portion.
+   *
+   * <p>E.g. "1ms" would return the string "ms".
+   */
+  private static String getDurationUnitString(String rawValue) {
```
anuraaga, line 220: Not a big difference, but I think I would search for the first non-number to avoid the additional `isDurationUnitToken` method.

sdk/common/src/test/java/io/opentelemetry/sdk/common/export/ConfigBuilderTest.java
```diff
@@ -59,6 +59,34 @@ void doubleProperty() {
...
+            "duration", Collections.singletonMap("duration", "2m"));
+    assertThat(durationProperty).isEqualTo(120000L);
+    durationProperty =
+        ConfigBuilder.getDurationMillisProperty(
+            "duration", Collections.singletonMap("duration", "1h"));
+    assertThat(durationProperty).isEqualTo(3600000);
+    durationProperty =
+        ConfigBuilder.getDurationMillisProperty(
+            "duration", Collections.singletonMap("duration", "3d"));
+    assertThat(durationProperty).isEqualTo(259200000L); // 2592e8 = 3 * 24 * 3600 * 1000
```
anuraaga, line 87: Instead of needing comments, we could use e.g., `TimeUnit.DAYS.toMillis` in these tests too for self-description.

sdk/common/src/main/java/io/opentelemetry/sdk/common/export/ConfigBuilder.java
jsuereth: Done.  Not sure if it's too heavyweight, but it is convenient.

sdk/common/src/main/java/io/opentelemetry/sdk/common/export/ConfigBuilder.java
jsuereth: Done.

sdk/common/src/main/java/io/opentelemetry/sdk/common/export/ConfigBuilder.java
jsuereth: I moved to that. There's a lot of "oddities" w/ i18n and codepoints here, depending on what people paste in.  LIKELY we should throw a ConfigException on most attempts to hack this value, but my brain's still not quite up to the task of remembering the problematic codepoints + encodings to try out in tests.

sdk/common/src/test/java/io/opentelemetry/sdk/common/export/ConfigBuilderTest.java
jsuereth: Done.

jkwatson (approved):
Seems good to me. We should probably hold off on merging until the relevant spec stuff has been approved so we don't end up supporting dead code.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
```diff
@@ -129,4 +181,40 @@ private static ConfigurationException newInvalidPropertyException(
         .filter(s -> !s.isEmpty())
         .collect(Collectors.toList());
   }
+
+  /** Returns the TimeUnit associated with a unit string. Defaults to milliseconds. */
+  private static TimeUnit getDurationUnit(String unitString) {
+    // If statement in likely-to-be-used order.
```
anuraaga, line 187: Let's just go in order of size to reduce confusion when reading this code, the optimization of one less branch for going out-of-order is just not worth it.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
```diff
@@ -129,4 +181,40 @@ private static ConfigurationException newInvalidPropertyException(
...
+    if ("s".equals(unitString)) {
```
anuraaga, line 188: Consider `switch(unitString)`, ends up as same bytecode but less `.equals` repetition

anuraaga (approved):
Just some very small nits, thanks!

anuraaga (commented):
@jsuereth With https://github.com/open-telemetry/opentelemetry-specification/issues/1351 filed I think we can merge this. Do you mind taking a look at the remaining comments?

jkwatson:
@jsuereth ping! I think we could get this merged, but there are some outstanding comments, and it needs a rebase.

jsuereth:
Sorry! Fixing now.  Was swamped in gradle + too many meetings hell for the past 48 hours.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
```diff
@@ -129,4 +181,40 @@ private static ConfigurationException newInvalidPropertyException(
...
+  private static String getUnitString(String rawValue) {
+    int i = rawValue.length() - 1;
+    while (i >= 0) {
+      char c = rawValue.charAt(i);
+      if (Character.isDigit(c)) {
+        break;
+      }
+      i -= 1;
+    }
+    return rawValue.substring(i + 1);
```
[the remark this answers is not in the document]
jsuereth: Two points:

1. Reverse parsing is common tactic.   If you'd like I can fix up readability.  Your code finds the first non-digit, my code find the last digit.  Renaming "i" to "lastDigitIndex" should clarify that
2. Reverse-parsing is a LOT more efficient here.  Usually my while loop is a no-op, and it should iterate over far less characters.  I'd hate to swap this in case someone does try to use this where performance matters (or copy-pasta is somewhere else without understanding that nuance).

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
jsuereth: Fixed

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
jsuereth: Removed

jsuereth:
@jkwatson Should be updated and ready to go.

anuraaga (approved):
Thanks1

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
```diff
@@ -86,6 +88,56 @@ Double getDouble(String name) {
     }
   }
 
+  /**
+   * Get a duration property from the map, {@code null} if it cannot be found or it has a wrong
```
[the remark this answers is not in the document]
anuraaga: Hmm this reminds me of the spec repo PR - in case you're not familiar, you need to press "Commit suggestion" to reflect it into the code, not only resolve the conversation.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
```diff
@@ -129,6 +131,56 @@ public Double getDouble(String name) {
...
+   * <p>If no unit is specified, milliseconds is the assumed duration unit.
+   *
+   * @param name The property name
+   * @return the {@link Duration} value of the property, {@code null} if the property cannot be
+   *     found.
+   * @throws ConfigurationException for malformed duration strings.
+   */
+  @Nullable
+  @SuppressWarnings("UnusedException")
+  public Duration getDuration(String name) {
```
anuraaga, line 157: Added a commit to change visibility and start of docstring to match a recent PR

PR #2440, opened 2021-01-06
Add a test for the RECORD_ONLY span handling in the BatchSpanProcessor

### LINKED ISSUE #1588: Deal with the TODOs in the BatchSpanProcessorTest

jkwatson:
And remove the old TODOs from the BSP test.

resolves #1588

### DIFF
--- a/sdk/trace/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorTest.java
+++ b/sdk/trace/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorTest.java
@@ -11,14 +11,17 @@ import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
+import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.context.Context;
 import io.opentelemetry.sdk.common.CompletableResultCode;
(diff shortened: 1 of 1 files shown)


sdk/trace/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorTest.java
```diff
@@ -378,48 +371,57 @@ void exportNotSampledSpans() {
...
-            .build());
-
-    createNotSampledRecordingEventsEndedSpan(SPAN_NAME_1);
-    ReadableSpan sampledSpan = createSampledEndedSpan(SPAN_NAME_2);
-    List<SpanData> exported = waitingSpanExporter.waitForExport(1);
-    assertThat(exported).containsExactly(sampledSpan.toSpanData());
-    */
+    createEndedSpan(SPAN_NAME_2);
+
+    List<SpanData> exported = waitingSpanExporter.waitForExport();
```
anuraaga, line 423: I think you need to do a similar pattern as the other test of exporting a sampled span after the record only span and checking it's the only exported span. Otherwise, this isn't doing any waiting I believe since nothing to wait for.

jkwatson: The exporter is waiting for 1, and I was relying on the timeout to ensure nothing was getting through. Am I misunderstanding how the waiting exporter works?

anuraaga (approved):

carlosalberto (approved):

PR #2441, opened 2021-01-07
Set archiveBaseName pattern for sdk-extensions too

anuraaga:
Had just forgotten for this folder


jkwatson (approved):
gradle is mystifying

PR #2442, opened 2021-01-07
Deprecate ConfigBuilder.

anuraaga:
Autoconfiguration now job of autoconfigure module

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -117,6 +117,10 @@ subprojects {
             it.options.errorprone.disable("AutoValueImmutableFields")
             // "-Xep:AutoValueImmutableFields:OFF"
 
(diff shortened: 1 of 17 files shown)


build.gradle
```diff
@@ -117,6 +117,10 @@ subprojects {
             it.options.errorprone.disable("AutoValueImmutableFields")
             // "-Xep:AutoValueImmutableFields:OFF"
 
+            // Fully qualified names may be necessary when deprecating a class to avoid
+            // deprecation warning.
+            it.options.errorprone.disable("UnnecessarilyFullyQualified")
```
jkwatson, line 122: bah. this seems like an errorprone bug to me.

jkwatson (approved):
Thanks!

PR #2443, opened 2021-01-07
Make JaegerRemoteSamplerBuilder consistent with other builders.

anuraaga:
While I had some mind to mark this module as alpha as well but it's not that much work to make it ready I guess.

- Builder to top level
- setEndpoint similar to other gRPC builders
- setInterval with TimeUnit / Duration instead of only millis
- set, not with
- changed initial sampler default to parent-based since surely it's a better default

### DIFF
--- a/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerBuilder.java
+++ b/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerBuilder.java
@@ -118,7 +118,7 @@ public class JaegerRemoteSamplerBuilder {
    *
    * @return the remote sampler instance.
    */
-  public JaegerRemoteSampler build() {
+  public Sampler build() {
     if (channel == null) {
       channel = ManagedChannelBuilder.forTarget(endpoint).usePlaintext().build();
     }
--- a/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerBuilder.java
+++ b/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerBuilder.java
@@ -87,38 +87,11 @@ public class JaegerRemoteSamplerBuilder {
   }
 
   /**
-   * Sets the polling interval. Optional.
-   *
-   * @param pollingIntervalMs the polling interval in Ms.
-   * @return this.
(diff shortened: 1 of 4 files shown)


sdk-extensions/jaeger-remote-sampler/src/test/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerIntegrationTest.java
```diff
@@ -38,13 +37,12 @@
...
   @Test
   void remoteSampling_perOperation() {
-    String jaegerHost =
-        String.format("127.0.0.1:%d", jaegerContainer.getMappedPort(COLLECTOR_PORT));
     final JaegerRemoteSampler remoteSampler =
-        JaegerRemoteSampler.builder()
-            .setChannel(ManagedChannelBuilder.forTarget(jaegerHost).usePlaintext().build())
-            .setServiceName(SERVICE_NAME)
-            .build();
+        (JaegerRemoteSampler)
```
jkwatson, line 41: any reason to not have the builder return the concrete class, rather than the interface? Is that for the agent code to work more easily?

jkwatson (approved):
Thanks!

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerBuilder.java
```diff
@@ -0,0 +1,129 @@
...
+   *
+   * @param initialSampler the initial sampler to use.
+   * @return this.
+   * @deprecated Use {@link #setInitialSampler(Sampler)}
+   */
+  @Deprecated
+  public JaegerRemoteSamplerBuilder withInitialSampler(Sampler initialSampler) {
+    this.initialSampler = initialSampler;
+    return this;
+  }
```
bogdandrutu, line 114: This is already a breaking change, do we need to deprecate then?

anuraaga: Ah yeah, thanks

sdk-extensions/jaeger-remote-sampler/src/test/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerIntegrationTest.java
anuraaga: I thought it could make things more future-proof, but after a rethink, I can't come up with a benefit indeed, reverted.

PR #2444, opened 2021-01-07
Accept TimeUnit and Duration for BSP values instead of forcing millis.

anuraaga:
Also doesn't seem to hurt to use nanos for the internal stored value.


sdk/trace/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorTest.java
```diff
@@ -412,16 +442,24 @@ public String getDescription() {
...
                     .build())
             .setTraceConfig(traceConfig::get)
             .build();
 
-    createEndedSpan(SPAN_NAME_2);
+    createEndedSpan(SPAN_NAME_1);
+    traceConfig.set(traceConfig.get().toBuilder().setSampler(Sampler.alwaysOn()).build());
+    ReadableSpan span = createEndedSpan(SPAN_NAME_2);
 
+    // Spans are recorded and exported in the same order as they are ended, we test that a non
```
anuraaga, line 455: @jkwatson I ended up adding this as discussed on the other PR so the test ends in 100ms instead of 10s

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorBuilder.java
```diff
@@ -104,15 +127,36 @@ boolean getExportOnlySampled() {
...
 
   // Visible for testing
-  long getScheduleDelayMillis() {
-    return scheduleDelayMillis;
+  long getScheduleDelayNanos() {
+    return scheduleDelayNanos;
+  }
+
+  /**
+   * Sets the maximum time an exporter will be allowed to run before being cancelled. If unset,
```
jkwatson, line 143:
```suggestion
   * Sets the maximum time an export will be allowed to run before being cancelled. If unset,
```

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorBuilder.java
```diff
@@ -104,15 +127,36 @@ boolean getExportOnlySampled() {
...
+   */
+  public BatchSpanProcessorBuilder setExporterTimeout(long timeout, TimeUnit unit) {
+    requireNonNull(unit, "unit");
+    checkArgument(timeout >= 0, "timeout must be non-negative");
+    exporterTimeoutNanos = unit.toNanos(timeout);
+    return this;
+  }
+
+  /**
+   * Sets the maximum time an exporter will be allowed to run before being cancelled. If unset,
```
jkwatson, line 154:
```suggestion
   * Sets the maximum time an export will be allowed to run before being cancelled. If unset,
```

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -284,7 +284,7 @@ private void exportCurrentBatch() {
 
       try {
         final CompletableResultCode result = spanExporter.export(batch);
-        result.join(exporterTimeoutMillis, TimeUnit.MILLISECONDS);
+        result.join(exporterTimeoutNanos, TimeUnit.NANOSECONDS);
```
jkwatson, line 287: Nanos feels a bit like overkill here. There are no processors on which you can actually join at nanosecond precision are there? Or even less than microsecond?

sdk/trace/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorTest.java
jkwatson: doh. I merged that one already, so you'll need a rebase.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
anuraaga: Timeouts are always handled with nanos against `System.nanoTime()` (wall time is what doesn't have less-than-microsecond precision). So what we pass gets converted into a nanosecond value. I agree it's overkill to ever use this, but the advantage of using nanos here is no need to document anything like "values are internally converted to millis precision", and I guess there isn't really a disadvantage.

jkwatson (approved):
Thanks!

PR #2445, opened 2021-01-07
Add TraceConfig.builder()

anuraaga:
Now that `updateTraceConfig` is deprecated, I think we do want this to avoid the tedium / lack of discoverability of `getDefault().toBuilder()`


jkwatson (approved):
Yes, thanks!

PR #2446, opened 2021-01-07
Cleanup OTLP exporter timeout settings.

anuraaga:
- deadline (absolute timestamp) -> timeout (relative time)
- Support TimeUnit / Duration
- Replaced internal value with nanos which is what gRPC converts to as well
- Make constants private that don't seem to need to be public


exporters/otlp/metrics/src/main/java/io/opentelemetry/exporter/otlp/metrics/OtlpGrpcMetricExporter.java
```diff
@@ -94,8 +105,8 @@ public CompletableResultCode export(Collection<MetricData> metrics) {
 
     final CompletableResultCode result = new CompletableResultCode();
     MetricsServiceFutureStub exporter;
-    if (deadlineMs > 0) {
-      exporter = metricsService.withDeadlineAfter(deadlineMs, TimeUnit.MILLISECONDS);
+    if (timeoutNanos > 0) {
+      exporter = metricsService.withDeadlineAfter(timeoutNanos, TimeUnit.NANOSECONDS);
```
jkwatson, line 109: same comment about nanos being overkill from the other PR.

jkwatson (approved):
Thanks!

jkwatson:
this needs a wee rebase.

PR #2448, opened 2021-01-07
Exclude new location of semattrs from coverage

anuraaga:


jkwatson (approved):
oh shoot. thanks for this. I thought about this when putting in the original PR, but then it totally slipped my mind.

PR #2451, opened 2021-01-07
Use ParentBased for jaeger remote sampler results

anuraaga:
And fix mutable field not volatile

While I haven't used Jaeger before, I would be very surprised if the results were not supposed to be parent-based, at least not without a way for the server to control that.


jkwatson (approved):
seems good to me. Do we have a Jaeger expert around who can confirm this? @yurishkuro ?

yurishkuro:
The change to `volatile` is fine, but there's a larger bug that stateful samplers are always replaced instead of being updated while maintaining their internal state. Specifically, rate limiting samplers, when implemented as leaky bucket or similar, have an internal state ("water level"), which gets reset if a brand new sampler is created every time the SDK polls for a strategy (even if strategy did not change).

jkwatson:
> The change to `volatile` is fine, but there's a larger bug that stateful samplers are always replaced instead of being updated while maintaining their internal state. Specifically, rate limiting samplers, when implemented as leaky bucket or similar, have an internal state ("water level"), which gets reset if a brand new sampler is created every time the SDK polls for a strategy (even if strategy did not change).

How about using ParentBased as the base? I think that was the main point of this particular PR.

yurishkuro:
Yes, ParentBased is consistent with Jaeger SDK behavior.

anuraaga:
Thanks @yurishkuro filed #2465

PR #2452, opened 2021-01-07
Update to latest semantic conventions.

anuraaga:


Oberon00 (approved):

buildscripts/semantic-convention/generate.sh
```diff
@@ -3,9 +3,9 @@
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 ROOT_DIR="${SCRIPT_DIR}/../../"
 # freeze the spec version to make SemanticAttributes generation reproducible
-# this hash was obtained by calling `git rev-parse master` on 14th Dec 2020
+# this hash was obtained by calling `git rev-parse master` on 2020/01/07
 # we can't use a version tag here because otel-spec releases are very rare
-SPEC_VERSION=7fb1f01e6a4d9a6b8b7f8780385809136179fd45
+SPEC_VERSION=d01d110b41b384e3d90169ec8d66d07f8b5963cd
```
anuraaga, line 8: By the way, I guess we don't expect to automatically run this script in CI. Does it mean there's no benefit in using this fixed version? If it's HEAD, then anytime someone wants to update conventions they just need to run the script and setup a PR without affecting build reproducibility.

jkwatson: I really don't want to target HEAD if we can help it. Having a specific SHA means we have traceability about what we generated the code from. For me, it's not about build reproducibility, and just knowing a specific version that we are using. I'd much much prefer that we actually did spec releases that we could use, but until that happens, this at least let's us backtrack and figure out what we are using.

buildscripts/semantic-convention/generate.sh
```diff
@@ -3,9 +3,9 @@
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 ROOT_DIR="${SCRIPT_DIR}/../../"
 # freeze the spec version to make SemanticAttributes generation reproducible
-# this hash was obtained by calling `git rev-parse master` on 14th Dec 2020
+# this hash was obtained by calling `git rev-parse master` on 2020/01/07
```
jkwatson, line 6:
```suggestion
# this hash was obtained by calling `git rev-parse master` on 2021/01/07
```

bogdandrutu (approved):

PR #2453, opened 2021-01-07
Deprecate OpenTelemetry SPI.

anuraaga:
We have autoconfigure module for SDK customization and would like to see patterns related to custom SDKs before SPIing to much initially.

Hopefully this helps with #2022

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
+++ b/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
@@ -8,8 +8,6 @@ package io.opentelemetry.api;
 import io.opentelemetry.api.trace.Tracer;
 import io.opentelemetry.api.trace.TracerProvider;
 import io.opentelemetry.context.propagation.ContextPropagators;
-import io.opentelemetry.spi.OpenTelemetryFactory;
-import io.opentelemetry.spi.trace.TracerProviderFactory;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.logging.Level;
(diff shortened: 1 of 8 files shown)


api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
```diff
@@ -33,6 +26,7 @@
  * @see TracerProvider
  * @see ContextPropagators
  */
+@SuppressWarnings("deprecation") // Remove after deleting OpenTelemetry SPI
```
jkwatson, line 29: Wait up. We're deleting the global altogether? I think I might have missed a memo. I'm assuming this is a mistake?

jkwatson (changes requested):
I don't think we are planning on deprecating the global...

bogdandrutu:
@jkwatson I think you need to agree with @anuraaga on this, because I pointed the same in other PRs (or I got the same impression that this is the intent) and I tried to block it, but could not hold on this because everyone seem to want this.

jkwatson:
> @jkwatson I think you need to agree with @anuraaga on this, because I pointed the same in other PRs (or I got the same impression that this is the intent) and I tried to block it, but could not hold on this because everyone seem to want this.

The intent is absolutely to remove the SPI from here, but not the global. At least, that's my understanding.

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
anuraaga: This is a suppresswarnings, no a Deprecated :P

jkwatson: oh. yes. :disappear:

jkwatson (approved):
Thanks!

jkwatson:
> I don't think we are planning on deprecating the global...

And I'm right! And, totally misread an annotation.

bogdandrutu (approved):

PR #2467, opened 2021-01-08
Cleanup ViewRegistry, move processor creation to InstrumentProcessor

bogdandrutu:

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/InstrumentProcessor.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/InstrumentProcessor.java
@@ -11,12 +11,12 @@ import io.opentelemetry.sdk.common.InstrumentationLibraryInfo;
 import io.opentelemetry.sdk.metrics.aggregator.Aggregator;
 import io.opentelemetry.sdk.metrics.common.InstrumentDescriptor;
 import io.opentelemetry.sdk.metrics.data.MetricData;
+import io.opentelemetry.sdk.metrics.view.AggregationConfiguration;
 import io.opentelemetry.sdk.resources.Resource;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 
 /**
  * An {@code InstrumentProcessor} represents an internal instance of an {@code Accumulator} for a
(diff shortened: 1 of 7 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/InstrumentProcessor.java
```diff
@@ -37,41 +37,40 @@
...
+            /* delta= */ false);
+      case DELTA:
+        return new InstrumentProcessor<>(
+            descriptor,
+            aggregator,
+            meterProviderSharedState.getResource(),
+            meterSharedState.getInstrumentationLibraryInfo(),
+            meterProviderSharedState.getClock(),
+            /* delta= */ true);
+    }
```
breedx-splk, line 72: The only difference between the two cases is the delta boolean.  If you make a local var like 
```
boolean delta = configuration.getTemporality() == DELTA;
```
you can simplify the switch, remove the `/* delta */` comment, and it cleans up a little bit.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/ViewRegistryTest.java
```diff
@@ -7,95 +7,16 @@
...
-
-    InstrumentProcessor<?> result =
-        viewRegistry.createBatcher(providerSharedState, meterSharedState, descriptor);
-
-    assertThat(result.generatesDeltas()).isTrue();
-    assertThat(result).isEqualTo(expectedInstrumentProcessor);
-
-    assertThat(result).isNotNull();
-  }
-
```
breedx-splk, line 98: Did we lose coverage here?  Should this have moved to the InstrumentProcessorTest instead?

breedx-splk (changes requested):
This looks good to me.  The whole `sharedState.getViewRegistry().createBatcher(...)` felt like a law of Demeter violation, so this change simplifies that for sure.  👍 Had a small suggestion and a question around test coverage.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/ViewRegistryTest.java
bogdandrutu: Already tested when we test all the instrument types

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/InstrumentProcessor.java
bogdandrutu: Good suggestion, done.

jkwatson (approved):
👍

PR #2468, opened 2021-01-08
Reduce number of allocations by passing an output list to collectAll

bogdandrutu:

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractAsynchronousInstrument.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractAsynchronousInstrument.java
@@ -19,7 +19,7 @@ abstract class AbstractAsynchronousInstrument extends AbstractInstrument {
   }
 
   @Override
-  public final List<MetricData> collectAll() {
-    return accumulator.collectAll();
+  final void collectAll(List<MetricData> output) {
+    accumulator.collectAll(output);
   }
 }
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeter.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeter.java
@@ -99,7 +99,7 @@ final class SdkMeter implements Meter {
     Collection<AbstractInstrument> instruments = instrumentRegistry.getInstruments();
     List<MetricData> result = new ArrayList<>(instruments.size());
     for (AbstractInstrument instrument : instruments) {
-      result.addAll(instrument.collectAll());
+      instrument.collectAll(result);
     }
     return result;
   }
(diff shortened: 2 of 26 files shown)


jkwatson:
While I appreciate the desire to micro-optimize, introducing this side-effect-driven approach to remove a few extra list creations seems premature. Do we have benchmarks that shows this is worth introducing side-effects to the internals of our code like this?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractAsynchronousInstrument.java
```diff
@@ -19,7 +19,7 @@
   }
 
   @Override
-  public final List<MetricData> collectAll() {
-    return accumulator.collectAll();
+  final void collectAll(List<MetricData> output) {
```
anuraaga, line 22: `collectAllTo`?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeter.java
```diff
@@ -99,7 +99,7 @@ public BatchRecorderSdk newBatchRecorder(String... keyValuePairs) {
     Collection<AbstractInstrument> instruments = instrumentRegistry.getInstruments();
     List<MetricData> result = new ArrayList<>(instruments.size());
     for (AbstractInstrument instrument : instruments) {
-      result.addAll(instrument.collectAll());
+      instrument.collectAll(result);
```
anuraaga, line 102: As this is an internal method and there does seem to be significant allocation churn here I suspect we will want to make the change. But I agree that we should always provide the before / after jmh for performance improvements.

anuraaga: For reference, rewrite proto serialization had a much larger negative effect on maintainability :) But we think it's worth it thanks to the numbers.

jkwatson: One list per instrument per collection cycle doesn't feel like much to me.

anuraaga: I may be wrong but isn't an instrument what actually measures a given metric? I guess apps often have hundreds of them which seems like a point that could use some tweaks.

jkwatson: Compared to the allocations with everything else involved in creating measurements, I doubt this would have much impact, but to re-iterate, we should measure the difference so we have data to base a decision on.

jkwatson:
@bogdandrutu I'm going to close this. The SDK has moved significantly since this was initially PRd. @jsuereth Can you tell if an optimization like this potentially relevant any more with the rework?

PR #2469, opened 2021-01-08
Extract IntervalMetricReaderBuilder as a top-level class

bogdandrutu:

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReaderBuilder.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReaderBuilder.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.metrics.export;
+
+import io.opentelemetry.api.internal.Utils;
+import java.util.Collection;
+import java.util.Map;
+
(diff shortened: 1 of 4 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReaderBuilder.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Map;
+
+/** Builder for {@link IntervalMetricReader}. */
+@SuppressWarnings("deprecation") // Remove after ConfigBuilder is deleted
+public final class IntervalMetricReaderBuilder
+    extends io.opentelemetry.sdk.common.export.ConfigBuilder<IntervalMetricReaderBuilder> {
+  private final IntervalMetricReader.InternalState.Builder optionsBuilder;
+  private static final String KEY_EXPORT_INTERVAL = "otel.imr.export.interval";
+
+  IntervalMetricReaderBuilder(IntervalMetricReader.InternalState.Builder optionsBuilder) {
```
jkwatson, line 19: this is a little weird, to have to point a doubly-nested class in here. Could we just take the 3 properties for the internal state and move them into this builder?

bogdandrutu: This is not public. I can do it in a follow-up PR.

jkwatson (approved):
👍

PR #2474, opened 2021-01-09
Deadline -> timeout for jaeger exporter

anuraaga:


jkwatson (approved):
👍  This verbiage will definitely be more universally understood.

PR #2475, opened 2021-01-11
Add a PropagatorProvider for custom propagators with autoconfiguration.

anuraaga:
While it would be sort of nice to allow Configurer of `OpenTelemetrySdkBuilder`, it creates a race condition where everything can be overwritten by `setTracerProvider` in a confusing way. It's not such a big deal to provide top-level configuration as separate providers I guess.


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/PropagatorProvider.java
```diff
@@ -0,0 +1,21 @@
...
+
+import io.opentelemetry.api.OpenTelemetry;
+import io.opentelemetry.context.propagation.TextMapPropagator;
+
+/**
+ * A service provider interface (SPI) for providing additional propagators to the autoconfigured
+ * SDK. The {@link TextMapPropagator} returned from any found {@link PropagatorProvider} will be
+ * available as part of {@link OpenTelemetry#getPropagators()}. This should generally only be used
+ * for custom propagators - for any propagator supported by OpenTelemetry, it is recommended to use
+ * the {@code otel.propagators} property instead.
```
jkwatson, line 16: How will this interact with `otel.propagators`? Are propagators provided by this SPI additive, or will they replace whatever is specified by the system property/env var?  I see from the implementation that it is additive, but we should absolutely document that here, for maximum clarity.

jkwatson (approved):
Just one documentation clarification needed, then this is 👍

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/PropagatorConfiguration.java
```diff
@@ -23,6 +25,11 @@ static ContextPropagators configurePropagators(ConfigProperties config) {
     for (String propagatorName : config.getCommaSeparatedValues("otel.propagators")) {
       propagators.add(PropagatorConfiguration.getPropagator(propagatorName));
     }
+
+    for (PropagatorProvider provider : ServiceLoader.load(PropagatorProvider.class)) {
+      propagators.add(provider.get());
+    }
+
     return ContextPropagators.create(TextMapPropagator.composite(propagators));
```
jkwatson, line 33: somewhere we need to du-dupe this vs. the env var/system prop. Where should that be done?

anuraaga: Went ahead and deduped here.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/PropagatorProvider.java
anuraaga: Yeah makes sense, added a note.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/ConfigurablePropagatorProvider.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+public interface ConfigurablePropagatorProvider {
+  /** Returns a {@link TextMapPropagator} to register to OpenTelemetry. */
+  TextMapPropagator getPropagator();
+
+  /**
+   * Returns the name of this propagator, which can be specified with the {@code otel.propagators}
+   * property to enable it.
+   */
+  String getName();
```
anuraaga, line 26: @jkwatson Talking with @trask, realized it makes more sense to allow adding a recognized propagator name rather than one that's automatically enabled. We can imagine the `jaeger-remote-sampler` artifact to have a compileOnly dependency on autoconfigure and add a `ConfigurablePropagatorProvider` implementation to allow configuring it by command line if users have both on classpath.

We might also want a provider that automatically enables it but it's not quite as important I think.

sdk-extensions/autoconfigure/src/testFullConfig/java/io/opentelemetry/sdk/autoconfigure/FullConfigTest.java
```diff
@@ -162,32 +153,18 @@ void configures() throws Exception {
...
-                    Span.wrap(
-                        SpanContext.create(
-                            TraceId.fromLongs(1, 1),
-                            SpanId.fromLong(2),
-                            TraceFlags.getDefault(),
-                            TraceState.builder().set("cat", "meow").build())))
-                .with(Baggage.builder().put("airplane", "luggage").build()),
-            headers,
-            Map::put);
+    Collection<String> fields =
```
anuraaga, line 156: Previous test was unnecessarily complex

anuraaga:
@jkwatson Rewrote to make the provider configure a propagator that can be registered rather than always registering, since it's more useful for things like adding autoconfiguration to a custom propagator.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/ConfigurablePropagatorProvider.java
jkwatson: :shipit:

jkwatson:
I like it. :shipit:

trask (approved):
nice, really looking forward to using the autoconfigure module in the agent 👍

PR #2480, opened 2021-01-12
Deprecate TraceProtoUtils

anuraaga:
Realized we have this class we probably don't want to release as a stable artifact, at least for now.


sdk-extensions/otproto/src/main/java/io/opentelemetry/sdk/extension/otproto/TraceProtoUtils.java
```diff
@@ -21,7 +21,9 @@ private TraceProtoUtils() {}
    *
    * @param spanId the spanId to convert.
    * @return a ByteString representation.
+   * @deprecated Will be removed without replacement
    */
+  @Deprecated
```
jkwatson, line 26: Is it worth also deprecating the class, as well? Not sure that it makes much difference, but just thought I'd ask.

jkwatson (approved):
Thanks!

jkwatson:
I'm going to merge this as-is. If we feel it's important to also deprecate the class itself, and not only the methods, we can do that separately.

PR #2481, opened 2021-01-12
Deprecate trace-propagators StringUtils to make package private.

anuraaga:
Found this presumably mistakenly public class.


Oberon00:
I thought that was intentional since I assumed the purpose of that package was to provide utility functions that may be useful for propagators. Although they are not really trace-specific...

anuraaga:
@Oberon00 IIRC, Dynatrace has a custom propagator - are you using these methods? And if so, how sad would you be to lose them? Indeed, because they're not trace-related hoping to avoid exposing them unless there's extreme demand. It's already a bit weird since custom propagators should only need to depend on `opentelemetry-api`, not on `opentelemetry-extension-trace-propagators` just to pick up `StringUtils`.

Oberon00:
It would not be a big deal for Dynatrace. I don't know from the top of my head if we use these methods, but they would be easy enough to copy & paste if we do.

anuraaga:
Got un-lazy and went ahead and did proper deprecation

jkwatson (approved):
Good deprecation. No need to publish and maintain this forever. Thanks!

PR #2483, opened 2021-01-12
CHANGELOG updates for Jan 12, 2021

jkwatson:


anuraaga (approved):

jkwatson:
I think this needs to be updated to include the deprecation of some metrics SDK bits, but that can be done post-release, since `-alpha` if we want.

Issue #2393, opened 2020-12-23
Update the quickstart, readme and examples to better reflect our recommended best practices


jkwatson:
For example: 

* cut-and-paste examples of using the bom to pull in dependencies as the various kinds of users that we're targetting.
* more explicit recommendations on configuring the SDK as a non-agent user
* links to the instrumentation repo on how to configure the agent
* how to get access to API and SDK for the non-mature signals (like metrics and eventually logs)

jkwatson:
I think I got everything updated. If we come across other things, we can tackle them individually.

PR #2484, opened 2021-01-12
Update the zipkin example to better match best practices, and update all to work with 0.14.1

jkwatson:
As a part of #2393 I'd like to get the examples to better separate SDK and API usage. Here I've updated the zipkin example to do that. If we like this structure, I'll follow-up and update all the examples (or someone else can pick that up, of course).

### DIFF
--- a/examples/zipkin/src/main/java/io/opentelemetry/example/zipkin/ExampleConfiguration.java
+++ b/examples/zipkin/src/main/java/io/opentelemetry/example/zipkin/ExampleConfiguration.java
@@ -33,15 +33,17 @@ public class ExampleConfiguration {
             .build();
 
     // Set to process the spans by the Zipkin Exporter
-    OpenTelemetrySdk openTelemetry = OpenTelemetrySdk.builder()
-        .setTracerProvider(SdkTracerProvider.builder()
-            .addSpanProcessor(SimpleSpanProcessor.create(zipkinExporter))
-            .build())
-        .buildAndRegisterGlobal();
+    OpenTelemetrySdk openTelemetry =
+        OpenTelemetrySdk.builder()
+            .setTracerProvider(
+                SdkTracerProvider.builder()
+                    .addSpanProcessor(SimpleSpanProcessor.create(zipkinExporter))
+                    .build())
+            .buildAndRegisterGlobal();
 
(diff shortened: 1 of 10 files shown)


examples/zipkin/src/main/java/io/opentelemetry/example/zipkin/ExampleConfiguration.java
```diff
@@ -0,0 +1,52 @@
...
+    String httpUrl = String.format("http://%s:%s", ip, port);
+    ZipkinSpanExporter zipkinExporter =
+        ZipkinSpanExporter.builder()
+            .setEndpoint(httpUrl + ENDPOINT_V2_SPANS)
+            .setServiceName(SERVICE_NAME)
+            .build();
+    // save the management interface so we can shut it down at the end of the example.
+    sdkTracerManagement = openTelemetrySdk.getTracerManagement();
+    // Set to process the spans by the Zipkin Exporter
+    sdkTracerManagement.addSpanProcessor(SimpleSpanProcessor.builder(zipkinExporter).build());
```
anuraaga, line 42: This separation looks nice, and I suspect we should go for it. But it'd be best if we can update to the new `Builder.addSpanProcessor.buildAndRegisterGlobal()` type of pattern to confirm what the end result looks like.

jkwatson: We can't do that until 0.14.0, correct? The examples are still using 0.13.1 for now, until we release.

anuraaga: I forget what's in what version 😅  Yeah I think so.

jkwatson:
ok, I've also updated all of the examples to at least not have any deprecations for 0.14.1. After this is merged, I'll follow up by updating the rest of the examples to have better structure.

examples/zipkin/src/main/java/io/opentelemetry/example/zipkin/ExampleConfiguration.java
jkwatson: ok, updated for 0.14.1!

anuraaga (approved):

PR #2489, opened 2021-01-12
Implement MetricProducer directly by the SdkMeterProvider

bogdandrutu:

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
@@ -26,13 +26,17 @@ import javax.annotation.Nullable;
  *
  * <p>This class is not intended to be used in application code and it is used only by {@link
  * OpenTelemetry}.
+ *
+ * <p>WARNING: A MetricProducer is stateful. It will only return changes since the last time it was
+ * accessed. This means that if more than one {@link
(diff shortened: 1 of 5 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
```diff
@@ -58,20 +61,14 @@ public SdkMeter get(String instrumentationName, @Nullable String instrumentation
...
-   *
-   * <p>WARNING: A MetricProducer is stateful. It will only return changes since the last time it
-   * was accessed. This means that if more than one {@link
-   * io.opentelemetry.sdk.metrics.export.MetricExporter} has a handle to this MetricProducer, the
-   * two exporters will not receive copies of the same metric data to export.
-   *
-   * @return the {@link MetricProducer} that can be used to retrieve metrics from this {@code
-   *     MeterSdkProvider}.
-   */
-  public MetricProducer getMetricProducer() {
```
jkwatson, line 73: I know metrics is alpha still, but is it worth deprecating this method (and changing it to return `this`) for one release, to ease the transition for alpha metrics users?

bogdandrutu: Your call ... I can do that easily do let's do it

bogdandrutu: Done.

jkwatson (approved):
👍

PR #2500, opened 2021-01-13
Deprecate StatusData.isUnset

anuraaga:
There is some inconsistency here. I think there are three options

1. Remove `isUnset` because there is no similar `isError`, and its behavior is different from `isOk` which checks two codes
2. Remove both `isUnset` and `isOk`
3. Add `isError`, change `isOk` to only check OK

I like `isOk` since it's annoying to check two codes to know whether a status is erroneous or not so I've picked the first option. I'm ok with 2 as well. 3 seems silly since it just duplicates the fairly simple code-vs-enum comparison into methods, but it would be more consistent than the current method names.

### DIFF
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/data/StatusData.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/data/StatusData.java
@@ -76,8 +76,8 @@ public interface StatusData {
       case OK:
         return true;
       case ERROR:
-      default:
         return false;
     }
+    return false;
   }
 }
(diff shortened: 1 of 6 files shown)


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/data/StatusData.java
```diff
@@ -68,8 +70,14 @@ default boolean isUnset() {
    *
    * @return {@code true} if this {@code Status} is OK or UNSET.
    */
-  // TODO: Consider to remove this in a future PR. Avoid too many changes in the initial PR.
   default boolean isOk() {
```
anuraaga, line 73: Also considering inverting this to `isFailure` so it uses a word that doesn't match one of the code names.

bogdandrutu: Maybe better to deprecate this as well....

bogdandrutu (approved):

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/data/StatusData.java
bogdandrutu: In a separate PR if you decire

jkwatson:
yeah, I'd be fine with deprecating both of them, and only adding back if there is sufficient demand.

anuraaga:
Ok went with deprecating both

PR #2505, opened 2021-01-13
Don't use ConfigBuilder for ResourceProvider allow-list

anuraaga:
As I was removing `ConfigBuilder`, I found that there is just one location in the SDK that still reads a property. I've inlined the logic rather than keeping something like `ConfigBuilder` around for it (it was somewhat obtuse usage of that builder pattern anyways), and as it's just an implementation detail we can change it later if something happens.


sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -16,19 +16,34 @@
...
+ * fully qualified class names of the {@link ResourceProvider} implementations as a comma separated
+ * list to the system property {@code -Dotel.java.disabled.resource_providers} or the {@code
+ * OTEL_JAVA_DISABLED_RESOURCE_PROVIDERS} environment variable.
  */
 @Immutable
 @AutoValue
 public abstract class Resource {
+
+  private static final String OTEL_JAVA_DISABLED_RESOURCES_PROVIDERS_PROPERTY_KEY =
+      "otel.java.disabled.resource_providers";
```
trask, line 42: instrumentation repo uses `-` instead of `_` in property names, e.g. `otel.java.disabled.resource-providers`, following spring convention: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1414#issuecomment-733181398

jkwatson (approved):
Seems fine to me! Thanks!

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
jkwatson: this isn't a new property. If we want to change it, that should be a separate thing to track.

trask: 👍 I'll open an issue for consideration

anuraaga:
Completely bizarre that even though this is already part of #2506 it could still be merged again since no changes

PR #2517, opened 2021-01-13
Extract inner classes from metrics API

bogdandrutu:

### DIFF
--- a/api/metrics/src/main/java/io/opentelemetry/api/metrics/AsynchronousInstrument.java
+++ b/api/metrics/src/main/java/io/opentelemetry/api/metrics/AsynchronousInstrument.java
@@ -6,7 +6,6 @@
 package io.opentelemetry.api.metrics;
 
 import io.opentelemetry.api.common.Labels;
-import java.util.function.Consumer;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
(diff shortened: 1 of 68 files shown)


api/metrics/src/main/java/io/opentelemetry/api/metrics/DoubleCounterBound.java
```diff
@@ -0,0 +1,24 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.api.metrics;
+
+import javax.annotation.concurrent.ThreadSafe;
+
+/** A {@code Bound Instrument} for a {@link DoubleCounter}. */
+@ThreadSafe
+public interface DoubleCounterBound extends BoundSynchronousInstrument {
```
jkwatson, line 12: why the rename? I think `BoundDoubleCounter` was a better name.

jkwatson:
ok, aside from renaming all the bound instrument classes. I much prefer having the "bound" adjective at the front of the class name.

api/metrics/src/main/java/io/opentelemetry/api/metrics/DoubleCounterBound.java
bogdandrutu: Was mostly trying to have classes together grouped by name. I understand your concern, will fix.

bogdandrutu: Done.

jkwatson (approved):
Thanks!

PR #2520, opened 2021-01-14
Update metric data util in the aggregator, remove unnecessary helpers

bogdandrutu:


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/LongLastValueAggregator.java
```diff
@@ -64,24 +64,41 @@ public MetricData toMetricData(
       Map<Labels, Long> accumulationByLabels,
       long startEpochNanos,
       long epochNanos) {
-    List<LongPoint> points =
```
jkwatson, line 67: is this a good change, to now have 3 places to make this call, rather than just the one?

bogdandrutu: I think so, because in certain cases we do the conversion and not use it.

jkwatson (approved):
👍

PR #2528, opened 2021-01-14
When creating aggregator pass resource and instrumentation library info.

bogdandrutu:
Avoid passing these info again when converting to metrics, and make every aggregator store this.


jkwatson (approved):
👍

PR #2534, opened 2021-01-14
Add initial support to define temporality per AggregatorFactory

bogdandrutu:
In this PR there are quiet some changes:
* Default aggregation for Sum/UpDownSum observers is changed to Sum, now that the sum knows how to produce cumulative for different types of instruments.
* SumAggregationFactory has an argument that allows to configure to produce always cumulative or delta for all types except sum/updownsum observers.
* `toMetricPoint` now accepts all available timestamps start, last collection, now and can decide which one to use.
* LastValue produces only gauge metrics now. The start time is always 0 (a.k.a. unset).

Fixes https://github.com/open-telemetry/opentelemetry-java/issues/2125


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
```diff
@@ -95,16 +95,13 @@ public static SdkMeterProviderBuilder builder() {
...
    *
    * //register the view with the MeterSdkProvider
-   * meterProvider.registerView(instrumentSelector, viewSpecification);
+   * meterProvider.registerView(instrumentSelector, aggregationFactory);
    * }</pre>
-   *
-   * @see AggregationConfiguration
    */
-  public void registerView(InstrumentSelector selector, AggregationConfiguration specification) {
+  public void registerView(InstrumentSelector selector, AggregatorFactory specification) {
```
jkwatson, line 104: this parameter name doesn't seem to make sense any more. rename to `aggregationFactory` I think.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/AbstractMinMaxSumCountAggregator.java
```diff
@@ -0,0 +1,52 @@
...
+import io.opentelemetry.api.common.Labels;
+import io.opentelemetry.sdk.common.InstrumentationLibraryInfo;
+import io.opentelemetry.sdk.metrics.common.InstrumentDescriptor;
+import io.opentelemetry.sdk.metrics.data.DoubleSummaryData;
+import io.opentelemetry.sdk.metrics.data.MetricData;
+import io.opentelemetry.sdk.resources.Resource;
+import java.util.Map;
+
+abstract class AbstractMinMaxSumCountAggregator
+    extends AbstractAggregator<MinMaxSumCountAccumulation> {
```
jkwatson, line 17: Since we're coercing all the longs into doubles when we accumulate, is worth having separate aggregators for long & double? Could we get away with just using double?

bogdandrutu: Separate PR? Already large this one and very complicated

jkwatson: sounds good!

jkwatson (approved):
👍

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
bogdandrutu: Done.

PR #2536, opened 2021-01-15
Use endpoint scheme in OTLP exporters

### LINKED ISSUE #2525: Update OTLP exporter(s) configuration to require a scheme for the endpoint.
See this spec PR: https://github.com/open-telemetry/opentelemetry-specification/pull/1234

anuraaga:
Fixes #2525

### DIFF
--- a/exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterBuilder.java
+++ b/exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterBuilder.java
@@ -16,20 +16,28 @@ import io.grpc.netty.GrpcSslContexts;
 import io.grpc.netty.NettyChannelBuilder;
 import io.grpc.stub.MetadataUtils;
 import java.io.ByteArrayInputStream;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.time.Duration;
 import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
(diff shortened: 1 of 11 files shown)


exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterBuilder.java
```diff
@@ -123,9 +150,9 @@ public OtlpGrpcSpanExporterBuilder addHeader(String key, String value) {
   public OtlpGrpcSpanExporter build() {
     if (channel == null) {
       final ManagedChannelBuilder<?> managedChannelBuilder =
-          ManagedChannelBuilder.forTarget(endpoint);
+          ManagedChannelBuilder.forTarget(endpoint.getAuthority());
```
jkwatson, line 153: authority, if I read the spec correctly, also can include auth info. Are we ok with that possibly ending up here, or would be better off explicitly grabbing the host and port and putting them together?

jkwatson: either, way, we should probably have a test that uses the full form of the authority section and makes sure we deal with it correctly. :)

anuraaga: Good callout - added tests and they actually seem to work. Not sure about authority vs host/port but since it seems to work leaning towards keeping it simple.

jkwatson (approved):

PR #2540, opened 2021-01-15
Generate the ResourceAttributes class

### LINKED ISSUE #2457: Current semconv are under a "trace" package but they are also resource semconv
See https://github.com/open-telemetry/opentelemetry-java/blob/master/semconv/src/main/java/io/opentelemetry/semconv/trace/attributes/SemanticAttributes.java#L34

jkwatson:
* deprecate the old one
* convert usages to the new one

resolves #2457


semconv/src/main/java/io/opentelemetry/semconv/trace/attributes/SemanticAttributes.java
```diff
@@ -7,350 +7,14 @@
...
 import io.opentelemetry.api.common.AttributeKey;
-import java.util.List;
 
 // DO NOT EDIT, this is an Auto-generated file from
-// buildscripts/semantic-convention//templates/SemanticAttributes.java.j2
+// buildscripts/semantic-convention/templates/SemanticAttributes.java.j2
 public final class SemanticAttributes {
 
-  /** Name of the cloud provider. */
-  public static final AttributeKey<String> CLOUD_PROVIDER = stringKey("cloud.provider");
```
jkwatson, line 21: my only question: do we need to deprecate each and every one of these that were in the wrong place?

bogdandrutu: If follow the rules yes, but we can be a bit more relaxed here

buildscripts/semantic-convention/templates/ResourceAttributes.java.j2
```diff
@@ -0,0 +1,109 @@
+/*
```
Oberon00, line 1: Can't we use the same template for both?

jkwatson: at the moment, there's a hand-added one at the end of the other template, so it would require some better jinja hacking than I'm currently qualified for.

jkwatson: I figured it out. easier than I thought. updated, thanks!

semconv/src/main/java/io/opentelemetry/semconv/trace/attributes/SemanticAttributes.java
anuraaga: If these stay with deprecation, I would be able to deflect the pain of these from when I update the instrumentation repo to you doing it after, presumably you don't want that :P

anuraaga (approved):

semconv/src/main/java/io/opentelemetry/semconv/trace/attributes/SemanticAttributes.java
jkwatson: I'll work on deprecating these. I might be able to figure out how to generate that automatically.

Oberon00 (approved):

semconv/src/main/java/io/opentelemetry/semconv/trace/attributes/SemanticAttributes.java
jkwatson: ok, deprecated. I did it manually since there didn't seem to be an obvious way to do it programmatically in the template.

anuraaga (approved):

PR #2547, opened 2021-01-18
Deprecate tracermanagement and SDK global access

anuraaga:
The main use for tracer management was mutating the global SDK - now that it's not mutable, this isn't important. The remaining use cases are

- shutdown. This does not need to be a separate management interface since whoever constructed the `SdkTracerProvider` should be managing it's shutdown anyways, using the methods on `SdkTracerProvider`

- force flush. Force flush is already fishy since it's used in business logic but requires using the SDK. `SdkTracerManagement` isn't the right interface for it anyways, we should instead consider an extension that can hide the SDK away behind a small wrapper. In the meantime, force flush is rare enough that use cases like awslambda Java instrumentation can just do it through the `SdkTracerProvider`.

The main use for the global SDK was to shutdown the SPI SDK or force flush the global SDK. Now, there is no SPI SDK, and `autoconfigure` shuts down automatically. Force flush for the `autoconfigure` SDK is still needed, but that should be exposed as a class in that module, not here in the SDK. Will add it separately. Incidentally, auto instrumentation is affected in a positive way - we currently block access to the global SDK using bytecode instrumentation and will be able to just remove that if this global access is gone.


sdk/testing/src/main/java/io/opentelemetry/sdk/testing/junit4/OpenTelemetryRule.java
```diff
@@ -80,8 +79,13 @@ public OpenTelemetry getOpenTelemetry() {
...
 
-  /** Returns the {@link SdkTracerManagement} created by this extension. */
-  public SdkTracerManagement getTracerManagement() {
+  /**
+   * Returns the {@link io.opentelemetry.sdk.trace.SdkTracerManagement} created by this extension.
+   *
+   * @deprecated Will be removed without replacement
+   */
+  @Deprecated
+  public io.opentelemetry.sdk.trace.SdkTracerManagement getTracerManagement() {
```
anuraaga, line 88: These also shouldn't be needed with the mutators gone.

Oberon00:
SdkTracerManagement was introduced to deter users from calling API methods on SdkTracerProvider mainly. I would still think it can be useful for that purpose, but probably we'd need to do more than we currently do, e.g.:

- Make `SdkTracerProvider` an internal class.
- The builder returns a `SdkTracerManagement`.
- The `SdkTracerManagement` gains a method `getTracerProvider` that returns an API `TracerProvider`

anuraaga:
I think that could potentially reduce direct usage of the provider. But it's a bit convoluted in my opinion, the classes become harder to understand. I don't think anything we do here really affects the policy that instrumentation only uses API. Do we have to add so many hoops to prevent users from using `SdkTracerProvider` in their apps instead of documentation? I think documenting caveats is common, and there is a balance between that vs hard-coding if with compile time tricks, don't know if they buy us that much though.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/SpanExporter.java
```diff
@@ -77,8 +76,8 @@ static SpanExporter composite(Iterable<SpanExporter> exporters) {
   CompletableResultCode flush();
 
   /**
-   * Called when {@link SdkTracerManagement#shutdown()} is called, if this {@code SpanExporter} is
-   * registered to a {@code TracerSdkManagement} object.
+   * Called when {@link io.opentelemetry.sdk.trace.SdkTracerProvider#shutdown()} is called, if this
```
jkwatson, line 79: this can be imported, yes? Since it's not deprecated?

jkwatson (approved):

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkTracerProvider.java
```diff
@@ -69,11 +72,28 @@ public Tracer get(String instrumentationName, @Nullable String instrumentationVe
...
   @Override
   public TraceConfig getActiveTraceConfig() {
     return sharedState.getActiveTraceConfig();
   }
 
+  /**
+   * Attempts to stop all the activity for this {@link Tracer}. Calls {@link
+   * SpanProcessor#shutdown()} for all registered {@link SpanProcessor}s.
+   *
+   * <p>This operation may block until all the Spans are processed. Must be called before turning
```
iNikem, line 85: It may block AND return `CompletableResultCode`?

anuraaga: Nice catch

PR #2548, opened 2021-01-18
Prevent setting global OpenTelemetry twice.

anuraaga:


sdk/testing/src/main/java/io/opentelemetry/sdk/testing/junit4/OpenTelemetryRule.java
```diff
@@ -68,8 +68,6 @@ public static OpenTelemetryRule create() {
   private final OpenTelemetrySdk openTelemetry;
   private final InMemorySpanExporter spanExporter;
 
-  private volatile OpenTelemetry previousGlobalOpenTelemetry;
```
anuraaga, line 71: With SPI gone and double-registration blocked, the global OTel is explicitly configured, and needs to be explicitly reset, so we don't need this backing up anymore.

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
```diff
@@ -67,14 +68,31 @@ public static OpenTelemetry get() {
...
+    }
   }
 
-  // for testing
-  static void reset() {
+  /**
+   * Unsets the global {@link OpenTelemetry}. This is only meant to be used from tests which need to
+   * reconfigure {@link OpenTelemetry}.
+   */
+  public static void resetForTest() {
```
jkwatson, line 95: oof. this being public makes me pretty uncomfortable. Is there any way we can restrict this to being package access again? Perhaps with a same-package testing bridge?

anuraaga: I moved to a separate class in the same package, not sure how much better it is, maybe a little.

api/all/src/main/java/io/opentelemetry/api/OpenTelemetryTesting.java
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.api;
+
+/** Utilities for use from tests related to OpenTelemetry. */
+public final class OpenTelemetryTesting {
```
jkwatson, line 9: Can this live in the testing artifact, or do we need it in the API? Look like it's needed for testing the SDK itself. I wonder if the SDK could provide its own little bridge for testing, outside of the official testing artifact.

bogdandrutu: I would rather make the `resetForTest` public than adding a new class, what do you think?

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
```diff
@@ -36,6 +36,7 @@
   private static final Object mutex = new Object();
 
   @Nullable private static volatile OpenTelemetry globalOpenTelemetry;
+  @Nullable private static Throwable setGlobalCaller;
```
bogdandrutu, line 39: Add a annotation that this is guarded by the mutex to clarify synchronization story.

bogdandrutu (commented):
Should get also call set with NOOP?

api/all/src/main/java/io/opentelemetry/api/OpenTelemetryTesting.java
jkwatson: If we can somehow document that people calling it in non-test code will not be supported, and that it isn't a part of the official public API, then I would be fine with it. I fear that if we keep it public on the API that it will get abused and we'll be stuck with it forever.

anuraaga: I think it's a bit too indirect to add this only to the testing artifact, and would be impossible for custom implementations to test with then.

This is a public API, I don't think we can run away from that at least in terms of an API break perspective. I feel the `This is only meant to be used from tests which need to reconfigure {@link OpenTelemetry}.` caveat and the method name itself having ForTest is reassuring enough to me. Should I move it back to OpenTelemetry?

anuraaga:
> Should get also call set with NOOP?

This sounds reasonable. Can you take another look at #2396? :) If registering the SDK late becomes impossible, I definitely want to warn the user.

bogdandrutu:
> This sounds reasonable. Can you take another look at #2396? :) If registering the SDK late becomes impossible, I definitely want to warn the user.

But do we need #2396 if we simply call `set(NOOP)` on the get? Then if user tries to set will get an error that was already registered when reading first time?

api/all/src/main/java/io/opentelemetry/api/OpenTelemetryTesting.java
jkwatson: If we need it to be a public API, then yes, it should be on the GlobalOpenTelemetry instance. I'm still not convinced that we need this public outside of the testing artifact, though. The testing artifact is a testing API, and if it exposed a public method that called a non-public method on the GOT, I'm not sure how that would inhibit 3rd parties from doing what they needed to do. 

I feel like I must be missing something, though. Why is it problematic to have the public method only in the testing artifact?

anuraaga: One issue is the testing artifact is for the SDK and depends on it, but this method is for the API, so a bit out of scope. It could still be ok though.

But to expose a method in the testing artifact calling a package private method, I guess it'd need to be in the same package. Don't believe we could have two artifacts exposing classes from the same package for java module reasons. Presumably module system also thinks `reset` should be next to `set`, which is its sibling method :P

jkwatson: The Java module thing is a good callout that I hadn't considered. Ugh. Perhaps we have to just live with it being a part of the public API, and hope people don't use it to do evil things that we don't want them to.

anuraaga: Thanks - moved it back, hoping the `ForTest` is clear enough.

I did come up with a "friend class" implementation - this method could be left public but throw an exception if `new Throwable().getStackTrace()` doesn't contain a specially recognized class from the testing artifact. Think it's too annoying though.

anuraaga (commented):
> But do we need #2396 if we simply call set(NOOP) on the get? Then if user tries to set will get an error that was already registered when reading first time?

It seems ok - the error message is less precise but I think we can live with it yeah.

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
```diff
@@ -110,6 +127,14 @@ public static Tracer getTracer(String instrumentationName, String instrumentatio
     return get().getTracer(instrumentationName, instrumentationVersion);
   }
 
+  /**
+   * Unsets the global {@link OpenTelemetry}. This is only meant to be used from tests which need to
+   * reconfigure {@link OpenTelemetry}.
+   */
+  public static void resetForTest() {
+    GlobalOpenTelemetry.globalOpenTelemetry = null;
```
jkwatson, line 135: why prefix `globalOpenTelemetry` with the class name here, but not other places? Is this an artifact of it being moved in & out of this class?

jkwatson, line 135:
```suggestion
    globalOpenTelemetry = null;
```

jkwatson (approved):

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
anuraaga: Yes :)

PR #2550, opened 2021-01-19
Use archivesBaseName for Implementation-Title

### LINKED ISSUE #2549: OpenTelemetry 0.14.1 JAR has manifest with Implementation-Title "all"
MANIFEST.MF of opentelemetry-api-0.14.1.jar looks like this:

    Manifest-Version: 1.0
    Implementation-Title: all
    Implementation-Version: 0.14.1
    Built-By: runner
    Built-JDK: 11.0.9
    Source-Compatibility: 1.8
    Target-Compatibility: 1.8
    Automatic-Module-Name: io.opentelemetry.api

I think Implementation-Title should be improved here. Some tools rely on information provided by manifest. Previous versions of the the jar used value "opentelemetry-api".

anuraaga:
When looking around, I realized many Gradle projects don't set this since there's no `addDefaultImplementationEntries` like in Maven. But doesn't hurt.

Fixes #2549

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -147,13 +147,15 @@ subprojects {
             options.compilerArgs += ["-Xlint:-serial"]
         }
 
-        jar.manifest {
-            attributes('Implementation-Title': name,
-                    'Implementation-Version': version,
-                    'Built-By': System.getProperty('user.name'),
-                    'Built-JDK': System.getProperty('java.version'),
-                    'Source-Compatibility': sourceCompatibility,
-                    'Target-Compatibility': targetCompatibility)
+        afterEvaluate {
+            jar.manifest {
+                attributes('Implementation-Title': archivesBaseName,
+                        'Implementation-Version': version,
+                        'Built-By': System.getProperty('user.name'),
+                        'Built-JDK': System.getProperty('java.version'),
+                        'Source-Compatibility': sourceCompatibility,
+                        'Target-Compatibility': targetCompatibility)
+            }
         }
 
         ext {


Oberon00 (approved):

jkwatson (approved):

quijote:
I just had a look at the manifest of 1.0.0 JARs.

API

    Manifest-Version: 1.0
    Automatic-Module-Name: io.opentelemetry.api

SDK

    Manifest-Version: 1.0
    Automatic-Module-Name: io.opentelemetry.sdk

There is no longer any information about the version of those JARs in manifest. Was this removed on purpose?

SDK contains io/opentelemetry/sdk/version.properties with content:

    sdk.version=1.0.0

But I did not find a good way to detect API version at runtime. 

At the moment we can no longer detect which OpenTelemetry API version is used by our monitored applications.

jkwatson:
> I just had a look at the manifest of 1.0.0 JARs.
> 
> API
> 
> ```
> Manifest-Version: 1.0
> Automatic-Module-Name: io.opentelemetry.api
> ```
> 
> SDK
> 
> ```
> Manifest-Version: 1.0
> Automatic-Module-Name: io.opentelemetry.sdk
> ```
> 
> There is no longer any information about the version of those JARs in manifest. Was this removed on purpose?
> 
> SDK contains io/opentelemetry/sdk/version.properties with content:
> 
> ```
> sdk.version=1.0.0
> ```
> 
> But I did not find a good way to detect API version at runtime.
> 
> At the moment we can no longer detect which OpenTelemetry API version is used by our monitored applications.

If this is something you need, please create an issue to track it. Thanks.

PR #2552, opened 2021-01-19
Register global in autoconfigure

anuraaga:
Currently calling initialize explicitly does not register the global, meaning a subsequent call to get the global actually reinitializes the SDK by calling initialize again. This isn't intended nor efficient/safe.


jkwatson (approved):
Thanks!

PR #2555, opened 2021-01-19
Use BiConsumer instead of custom interface

bogdandrutu:
Tried to use ObjDoubleConsumer but then it will be inconsistent with the `add` method on the sync instrument.

### DIFF
--- a/api/metrics/src/main/java/io/opentelemetry/api/metrics/AsynchronousInstrument.java
+++ b/api/metrics/src/main/java/io/opentelemetry/api/metrics/AsynchronousInstrument.java
@@ -5,7 +5,6 @@
 
 package io.opentelemetry.api.metrics;
 
-import io.opentelemetry.api.common.Labels;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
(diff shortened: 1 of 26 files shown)


jkwatson:
are we ok with 
1) boxing all the numeric recordings like this (I think probably yes, since this is async)?
2) handling the possibility of getting `null` passed in with `Long` and `Double` as the types?

bogdandrutu:
are we ok with

> boxing all the numeric recordings like this (I think probably yes, since this is async)?
Probably not a problem.

> handling the possibility of getting null passed in with Long and Double as the types?
Alternative is to use the ObjDoubleConsumer and ObjLongConsumer but unfortunately that will be (Labels, value) which is different than (value, Labels) like in the sync instruments.

I am open to consider that alternative and simplify things.

jkwatson:
If we do keep this as-is, what would the null handling look like? Would we silently drop those recordings? Or default them to 0? I don't think we would want to throw an exception. 

TBH, I wasn't even aware that `ObjDoubleConsumer` and `ObjLongConsumer` existed until you pointed them out! 

I don't have strong feelings either way, as long as we have good documentation on how nulls are handled if we go with the standard BiConsumer.

anuraaga:
I think it's pretty common to define custom interfaces to use a primitive type for a bit of performance and not worrying about null, I lean towards leaving it as is.

jkwatson:
I'm going to close this due to significant drift in the API since it was initially proposed.

PR #2559, opened 2021-01-19
Update the Jaeger exporters to properly use the service name from the Resource

### LINKED ISSUE #2557: Jaeger exporter(s): map Resource service.name into the Process.ServiceName
Currently, if the Resource has a service.name in it, it simply gets copied into a span tag, which is incorrect. This value should be set as the Process serviceName.

jkwatson:
resolves #2557

### DIFF
--- a/exporters/jaeger-thrift/src/main/java/io/opentelemetry/exporter/jaeger/thrift/JaegerThriftSpanExporter.java
+++ b/exporters/jaeger-thrift/src/main/java/io/opentelemetry/exporter/jaeger/thrift/JaegerThriftSpanExporter.java
@@ -13,6 +13,7 @@ import io.jaegertracing.thriftjava.Tag;
 import io.jaegertracing.thriftjava.TagType;
 import io.opentelemetry.sdk.common.CompletableResultCode;
 import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.sdk.resources.ResourceAttributes;
 import io.opentelemetry.sdk.trace.data.SpanData;
 import io.opentelemetry.sdk.trace.export.SpanExporter;
 import java.net.InetAddress;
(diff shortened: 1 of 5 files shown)


exporters/jaeger-thrift/src/main/java/io/opentelemetry/exporter/jaeger/thrift/JaegerThriftSpanExporter.java
```diff
@@ -13,6 +13,7 @@
 import io.jaegertracing.thriftjava.TagType;
 import io.opentelemetry.sdk.common.CompletableResultCode;
 import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.sdk.resources.ResourceAttributes;
```
jkwatson, line 16: This will need to be updated after #2540 lands, obviously.

exporters/jaeger-thrift/src/main/java/io/opentelemetry/exporter/jaeger/thrift/JaegerThriftSpanExporter.java
```diff
@@ -110,6 +111,12 @@ public CompletableResultCode export(Collection<SpanData> spans) {
 
   private Process createProcess(Resource resource) {
     Process result = new Process(this.process);
+
+    String serviceName = resource.getAttributes().get(ResourceAttributes.SERVICE_NAME);
+    if (serviceName != null && !serviceName.isEmpty()) {
+      result.setServiceName(serviceName);
```
anuraaga, line 117: Didn't we still need to fallback to a constant in exporters (though not make that configurable)?

jkwatson: that will be a follow-on PR after the relevant PRs to have the default service.name set. I'm already working on it. :)

anuraaga (approved):

PR #2562, opened 2021-01-20
Include only prod artifacts in BOM.

anuraaga:


jkwatson (approved):
👍 you beat me to it!

Oberon00:
I don't understand the reason for this. The alpha artifacts should be aligned with the corresponding release versions & dependency versions too, shouldn't they? Removing them from the BOM makes them harder to use, but does it bring any advantages for maintainability?

anuraaga:
@Oberon00 We're thinking of providing an `-alpha-bom` artifact with them. We want to make sure a user has the word `alpha` somewhere in their build file so they know they are opting into to something that will likely break them.

Oberon00:
Understood. That makes sense.

PR #2565, opened 2021-01-20
Handle old-style otlp endpoints properly.

jkwatson:
non-scheme'd endpoints would blow up before this change.

### DIFF
--- a/exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterBuilder.java
+++ b/exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterBuilder.java
@@ -80,7 +80,11 @@ public final class OtlpGrpcSpanExporterBuilder {
    */
   public OtlpGrpcSpanExporterBuilder setEndpoint(String endpoint) {
     requireNonNull(endpoint, "endpoint");
-    final URI uri;
+    if (!endpoint.matches("^[a-zA-Z]+:\\/\\/.*")) {
+      endpoint = "http://" + endpoint;
+    }
+
+    URI uri;
     try {
       uri = new URI(endpoint);
     } catch (URISyntaxException e) {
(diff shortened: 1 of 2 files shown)


exporters/otlp/trace/src/test/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterTest.java
```diff
@@ -83,13 +87,12 @@ void invalidConfig() {
     assertThatThrownBy(() -> OtlpGrpcSpanExporter.builder().setEndpoint(null))
         .isInstanceOf(NullPointerException.class)
         .hasMessage("endpoint");
-    assertThatThrownBy(() -> OtlpGrpcSpanExporter.builder().setEndpoint("😺://localhost"))
```
jkwatson, line 86: sorry, no more kitty-cats. they'll have to fail when grpc tries to connect

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterBuilder.java
```diff
@@ -80,7 +80,11 @@ public OtlpGrpcSpanExporterBuilder setTimeout(Duration timeout) {
    */
   public OtlpGrpcSpanExporterBuilder setEndpoint(String endpoint) {
     requireNonNull(endpoint, "endpoint");
-    final URI uri;
+    if (!endpoint.matches("^[a-zA-Z]+:\\/\\/.*")) {
```
anuraaga, line 83: How about just `endpoint.contains("://")`?  That way we won't miss restoring invalid URL check when we remove support for schemaless URLs.

anuraaga (approved):
Sorry thought there was still a residual test for this but there wasn't, oops.

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterBuilder.java
jkwatson: Seems fine. I'll update

jkwatson: done

exporters/otlp/trace/src/test/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterTest.java
anuraaga: Aww thought we saved the cat 😿

jkwatson: ah well.. we can bring it back if you think it's an important use-case that people might fall into.  ;)

PR #2569, opened 2021-01-22
Allow setting trusted certificate variable.

anuraaga:
I was originally hoping this would solve https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1829 but realized that's about mTLS, not TLS. I don't think the spec defines variables for mTLS yet (technically the current variable is misdocced I guess). But now that I made this change it doesn't hurt to get it in and the variable will likely be renamed in the future.

### DIFF
--- a/sdk-extensions/autoconfigure/build.gradle
+++ b/sdk-extensions/autoconfigure/build.gradle
@@ -88,7 +88,7 @@ testJaeger {
 testOtlpTls {
     environment("OTEL_RESOURCE_ATTRIBUTES", "service.name=test,cat=meow")
     environment("OTEL_TRACE_EXPORTER", "otlp")
-    environment("OTEL_BSP_SCHEDULE_DELAY_MILLIS", "10")
+    environment("OTEL_BSP_SCHEDULE_DELAY", "10")
 }
 
 testZipkin {
--- a/sdk-extensions/autoconfigure/build.gradle
+++ b/sdk-extensions/autoconfigure/build.gradle
@@ -15,6 +15,7 @@ testSets {
     testInitializeRegistersGlobal
     testJaeger
     testPrometheus
+    testOtlpTls
     testZipkin
 }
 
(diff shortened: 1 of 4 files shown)


sdk-extensions/autoconfigure/build.gradle
```diff
@@ -81,6 +85,12 @@ testJaeger {
     environment("OTEL_BSP_SCHEDULE_DELAY_MILLIS", "10")
 }
 
+testOtlpTls {
+    environment("OTEL_RESOURCE_ATTRIBUTES", "service.name=test,cat=meow")
+    environment("OTEL_TRACE_EXPORTER", "otlp")
+    environment("OTEL_BSP_SCHEDULE_DELAY_MILLIS", "10")
```
jkwatson, line 91: this shouldn't have MILLIS any more, correct?

jkwatson: also the ones above and below.

jkwatson: and yes, I missed this in the original PR to remove the MILLIS.  😢

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/SpanExporterConfiguration.java
```diff
@@ -59,6 +63,21 @@ static OtlpGrpcSpanExporter configureOtlpSpans(ConfigProperties config) {
       builder.setTimeout(Duration.ofMillis(timeoutMillis));
     }
 
+    String certificate = config.getString("otel.exporter.otlp.certificate");
```
jkwatson, line 66: the spec also suggests span and metric specific configuration. do we need to support that, or is this good enough (I think this is good enough for now...we can add the signal-specific ones later if needed)

sdk-extensions/autoconfigure/build.gradle
jkwatson: https://github.com/open-telemetry/opentelemetry-java/pull/2574

jkwatson (approved):

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/SpanExporterConfiguration.java
anuraaga: Yup missing the signal-specific ones for all these variables, will add in followup

PR #2576, opened 2021-01-25
Update to Gradle 6.8.1

anuraaga:
Take 2 - POMs look ok.

Also removed task setting which we don't really use / need.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -571,10 +571,6 @@ subprojects {
     }
 }
 
-wrapper {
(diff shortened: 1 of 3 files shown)


Oberon00 (approved):
So we can upgrade now because 6.8.1 includes the fix for https://github.com/gradle/gradle/issues/15731 (https://docs.gradle.org/6.8.1/release-notes.html#fixed-issues) -- Great!

jkwatson (approved):
🤞

PR #2578, opened 2021-01-25
OtTracerPropagator: send only 64 bits for trace id

malafeev:
cc @carlosalberto

### DIFF
--- a/extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/OtTracerPropagator.java
+++ b/extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/OtTracerPropagator.java
@@ -58,7 +58,7 @@ public final class OtTracerPropagator implements TextMapPropagator {
     if (!spanContext.isValid()) {
       return;
     }
-    // OpenTracing trace id MUST be 64-bits therefore OpenTelemetry trace id is truncated to 64-bits
+    // Lightstep trace id MUST be 64-bits therefore OpenTelemetry trace id is truncated to 64-bits
     // by retaining least significant (right-most) bits.
     setter.set(
         carrier,
--- a/extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/OtTracerPropagator.java
+++ b/extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/OtTracerPropagator.java
@@ -58,6 +58,8 @@ public final class OtTracerPropagator implements TextMapPropagator {
     if (!spanContext.isValid()) {
       return;
     }
+    // OpenTracing trace id MUST be 64-bits therefore OpenTelemetry trace id is truncated to 64-bits
+    // by retaining least significant (right-most) bits.
     setter.set(
         carrier,
         TRACE_ID_HEADER,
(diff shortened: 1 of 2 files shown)


extensions/trace-propagators/src/test/java/io/opentelemetry/extension/trace/propagation/OtTracerPropagatorTest.java
```diff
@@ -27,6 +27,7 @@
 
   private static final TraceState TRACE_STATE_DEFAULT = TraceState.builder().build();
   private static final String TRACE_ID = "ff000000000000000000000000000041";
+  private static final String TRACE_ID_RIGHT_PART = TRACE_ID.substring(TraceId.getHexLength() / 2);
```
carlosalberto, line 30: In order to have this in a more explicit fashion, I'd simply go and use the hardcoded value itself, as other values do here, i.e. `TRACE_ID_RIGHT_PART = "0000000000000041"`.

extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/OtTracerPropagator.java
```diff
@@ -58,7 +58,10 @@ public static OtTracerPropagator getInstance() {
     if (!spanContext.isValid()) {
       return;
     }
-    setter.set(carrier, TRACE_ID_HEADER, spanContext.getTraceIdAsHexString());
+    setter.set(
```
carlosalberto, line 61: Leave a comment explaining we are doing this explicitly as the format only supports 64-bit ids?

carlosalberto (approved):

carlosalberto:
Looks good! Left a pair of comments but overall LGTM

anuraaga (commented):
Do we need a custom id generator too?

extensions/trace-propagators/src/test/java/io/opentelemetry/extension/trace/propagation/OtTracerPropagatorTest.java
malafeev: done

extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/OtTracerPropagator.java
malafeev: done

extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/OtTracerPropagator.java
```diff
@@ -58,7 +58,12 @@ public static OtTracerPropagator getInstance() {
     if (!spanContext.isValid()) {
       return;
     }
-    setter.set(carrier, TRACE_ID_HEADER, spanContext.getTraceIdAsHexString());
+    // OpenTracing trace id MUST be 64-bits therefore OpenTelemetry trace id is truncated to 64-bits
```
anuraaga, line 61:
```suggestion
    // Lightstep trace id MUST be 64-bits therefore OpenTelemetry trace id is truncated to 64-bits
```

This is specific to Lightstep I guess? Opentracing spec says 128-bit

https://github.com/opentracing/specification/blob/master/rfc/trace_identifiers.md#trace-context-http-headers

malafeev:
> Do we need a custom id generator too?

I don't think we need custom id generator. cc @carlosalberto

anuraaga:
> I don't think we need custom id generator. cc @carlosalberto

I'm guessing this is because collector, or whatever other receiver, will also truncate IDs it receives, so they match up. Makes sense to me

anuraaga (approved):

carlosalberto:
>  will also truncate IDs it receives

Yes, this is the case :)

malafeev:
Looks like PR is ready to merge, all comments resolved. Can anybody merge it?

carlosalberto:
Merged (although not a maintainer anymore. Hope that is ok @jkwatson @anuraaga ;) )

PR #2575, opened 2021-01-25
Supports Histogram Aggregation

beanliu:
Support fixed bucket histogram aggregation and exporting them with OTLP/Prometheus exporter.

Since the View API is not part of the public Metric API yet ([optentelemetry-specification#466](https://github.com/open-telemetry/opentelemetry-specification/issues/466)), there's no universal way to configure custom aggregators for different instruments. However, the `SdkMeterProvider` provides a `registerView` method which allows developers to specify aggregations for certain instruments. This suggests that we could do the following to achieve histogram aggregations:

```Java
private static DoubleValueRecorder tryNewHistogramValueRecorder(
    MeterProvider meterProvider,
    Meter meter,
    String name,
    double[] boundaries) {
  if (meterProvider.getClass() == SdkMeterProvider.class) {
    // `registerView` is not part of the public Meter API yet, so here we use a cast to do a bit
    // of trick just to demonstrate the histogram usage.
    SdkMeterProvider sdkMeterProvider = (SdkMeterProvider) meterProvider;
    sdkMeterProvider
        .registerView(InstrumentSelector.builder().setInstrumentNameRegex(name).setInstrumentType(
            InstrumentType.VALUE_RECORDER).build(), AggregatorFactory.histogram(boundaries, /* stateful= */ false));
  }

// in the application/instrumentation code
DoubleValueRecorder mRequestLatency = tryNewHistogramValueRecorder(meterProvider, meter,
        "example_request_latency", new double[] {50, 200, 300});
DoubleValueRecorder mResponseSize = tryNewHistogramValueRecorder(meterProvider, meter,
    "example_response_size", new double[] {1_000, 5_000, 20_000});
```

And we could upgrade the example to use View API once they are ready.

PR #2581, opened 2021-01-25
Histogram Aggregation - core SDK implementation

beanliu:
the original PR #2575 was divided up to 3 PR following the advice from the comments.
they are not fully independent of each other and share some common commits, here's a simple description of their relationships in commits:
```
------>(add histogram to MetricData)------->(implement histogram aggregator, #2581 )
                                    ------->(OTLP exporter, #2582)
                                    ------->(Prometheus exporter, #2583 )
```
links of follow up PRs: #2582 #2583

---

Support fixed bucket histogram aggregation and exporting them with OTLP/Prometheus exporter.

Since the View API is not part of the public Metric API yet ([optentelemetry-specification#466](https://github.com/open-telemetry/opentelemetry-specification/issues/466)), there's no universal way to configure custom aggregators for different instruments. However, the `SdkMeterProvider` provides a `registerView` method which allows developers to specify aggregations for certain instruments. This suggests that we could do the following to achieve histogram aggregations:

```Java
private static DoubleValueRecorder tryNewHistogramValueRecorder(
    MeterProvider meterProvider, Meter meter, String name, double[] boundaries) {
  if (meterProvider.getClass() == SdkMeterProvider.class) {
    SdkMeterProvider sdkMeterProvider = (SdkMeterProvider) meterProvider;

    sdkMeterProvider.registerView(
        InstrumentSelector.builder()
            .setInstrumentNameRegex(name)
            .setInstrumentType(InstrumentType.VALUE_RECORDER)
            .build(),
        AggregatorFactory.histogram(
            ImmutableDoubleArray.copyOf(boundaries), /* stateful= */ false));
  }

  return meter.doubleValueRecorderBuilder(name).build();
}

// in the application/instrumentation code
DoubleValueRecorder mRequestLatency = tryNewHistogramValueRecorder(meterProvider, meter,
        "example_request_latency", new double[] {50, 200, 300});
DoubleValueRecorder mResponseSize = tryNewHistogramValueRecorder(meterProvider, meter,
    "example_response_size", new double[] {1_000, 5_000, 20_000});
```

And we could upgrade the example to use View API once they are ready.


sdk/metrics/src/jmh/java/io/opentelemetry/sdk/metrics/aggregator/BucketSearch.java
```diff
@@ -0,0 +1,244 @@
...
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Threads;
+import org.openjdk.jmh.annotations.Warmup;
+
+@State(Scope.Benchmark)
+public class BucketSearch {
```
jkwatson, line 20: what is the purpose of this benchmark? It doesn't seem to be benchmarking anything but how fast we can loop through an array. And, it does it with randomly selected values, which I'm not sure will end up providing a useful benchmark for anything.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/AggregatorFactory.java
```diff
@@ -45,6 +45,18 @@ static AggregatorFactory count(AggregationTemporality temporality) {
...
+  /**
+   * Returns an {@code AggregatorFactory} that calculates an approximation of the distribution of
+   * the measurements taken.
+   *
+   * @param stateful configures if the aggregator is stateful.
+   * @param boundary configures the fixed bucket boundaries.
+   * @return an {@code AggregationFactory} that calculates histogram of recorded measurements.
+   */
+  static AggregatorFactory histogram(double[] boundary, boolean stateful) {
+    return new HistogramAggregatorFactory(boundary, stateful);
```
jkwatson, line 57:
```suggestion
  static AggregatorFactory histogram(double[] boundaries, boolean stateful) {
    return new HistogramAggregatorFactory(boundaries, stateful);
```

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
```diff
@@ -0,0 +1,168 @@
...
+
+  @Override
+  public AggregatorHandle<HistogramAccumulation> createHandle() {
+    return new Handle(this.boundaries);
+  }
+
+  @Override
+  public final HistogramAccumulation merge(HistogramAccumulation x, HistogramAccumulation y) {
+    if (!x.getBoundaries().equals(y.getBoundaries())) {
+      throw new IllegalArgumentException("can't merge histograms with different boundaries");
```
jkwatson, line 43: what happens if this exception is thrown? Will this stop the collection from happening? Is there something we can do to make sure that this doesn't completely stop the process and break telemetry from being reported?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
```diff
@@ -0,0 +1,168 @@
...
+
+    long[] mergedCounts = new long[x.getCounts().size()];
+    for (int i = 0; i < x.getCounts().size(); ++i) {
+      mergedCounts[i] = x.getCounts().get(i) + y.getCounts().get(i);
+    }
+    return HistogramAccumulation.create(
+        x.getCount() + y.getCount(),
+        x.getSum() + y.getSum(),
+        x.getBoundaries(),
+        Arrays.stream(mergedCounts).boxed().collect(Collectors.toList()));
```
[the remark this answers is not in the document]
jkwatson: yeah, not sure why we can't just store these as arrays in the accumulation, rather than boxing them all up.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    @Override
+    protected void doRecordLong(long value) {
+      doRecordDouble((double) value);
+    }
+
+    private static final class State {
+      private long count;
+      private double sum;
```
jkwatson, line 135: since these two values can always be derived from the contents of the buckets, could we not keep them in the live state, and only calculate them when we need to report/merge the data, which happens off the hot path?

jkwatson: if we did that, we could only lock the specific bucket, rather than having to lock the whole data structure, when we want to record a value.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
```diff
@@ -0,0 +1,168 @@
...
+      private final double[] boundaries;
+      private final long[] counts;
+
+      public State(double[] boundaries) {
+        this.boundaries = Arrays.copyOf(boundaries, boundaries.length);
+        this.counts = new long[this.boundaries.length + 1];
+        reset();
+      }
+
+      // Benchmark shows that linear search performs better with ordinary buckets.
```
jkwatson, line 145: better than what?

bogdandrutu (commented):
I am confused about the series of PRs sent. Can we start with a minimal PR that just adds the data type, then one PR that adds the aggregator, then PRs to plug it to the exporters

jkwatson:
> I am confused about the series of PRs sent. Can we start with a minimal PR that just adds the data type, then one PR that adds the aggregator, then PRs to plug it to the exporters

heh. this was my request to break up the original into 3 PRs. this is the first one in the series at the moment.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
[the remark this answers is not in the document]
beanliu: iirc the reason is that AutoValue would complain about creating HistogramAccumulation with mutable primitive arrays.
but yep I could change it to the array approach and document the potential mutable issue.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
```diff
@@ -0,0 +1,168 @@
...
+    @GuardedBy("lock")
+    private final State current;
+
+    Handle(double[] boundaries) {
+      current = new State(boundaries);
+    }
+
+    @Override
+    protected HistogramAccumulation doAccumulateThenReset() {
+      lock.writeLock().lock();
```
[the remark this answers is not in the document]
beanliu: took the same approach as other `XyzAggregator`s to make them consistent. but I guess I could make it simpler.

sdk/metrics/src/jmh/java/io/opentelemetry/sdk/metrics/aggregator/BucketSearch.java
beanliu: The purpose is to justify the linear search of `DoubleHistogramAggregator.Handle.State.findBucketIndex`.

Here's the basic setup:
- the bucket boundaries are grabbed from some real cases.
- the values are selected in a uniform-distribution with proper upper bound related to the boundaries.
    - in real world, most value distribution would bias toward to the left end. so the results would serve as an estimated upper bound of the performance.

but I'm not sure if I should put the benchmark in the PR actually since it's not too general. Do u suggest removing this file?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
beanliu: those are very good questions. thanks for pointing that out.

my assumption is that the exception would never be thrown because `x` and `y` should be created from the same aggregator and belong to the same synchronous instrument.
tbh I'm not sure this is a strong one which always holds so it may be better to take a more resilient approach here.

current thought is that just return one of them or an empty one. do you have any suggestions with regard to that?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
beanliu: nice one. will do.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
beanliu: better than the binary search. will update the comment.

beanliu:
> I am confused about the series of PRs sent. Can we start with a minimal PR that just adds the data type, then one PR that adds the aggregator, then PRs to plug it to the exporters

sorry for the confusion. the original PR #2575 was divided up to 3 PR following the advice from the comments.
they are not fully independent of each other and share some common commits, here's a simple description of their relationships in commits:
```
------>(add histogram to MetricData)------->(implement histogram aggregator, #2581 )
                                    ------->(OTLP exporter, #2582)
                                    ------->(Prometheus exporter, #2583 )
```

will update the description to make it more clear. if you think the `(add histogram to MetricData)` part is worth another separate PR to have a better review process please let me know.

sdk/metrics/src/jmh/java/io/opentelemetry/sdk/metrics/aggregator/BucketSearch.java
jkwatson: Yes, I don't think that benchmarking linear searching, out of context of the actual usage of that search, has much value. I would definitely recommend removing it from this PR.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
jkwatson: If this happens, it's a pretty severe bug in the SDK, I think. In that case, we should log a message (only one per the lifetime of the SDK, I think) at SEVERE and just return an empty accumulation, I think.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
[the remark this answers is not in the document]
beanliu: > We may as well construct the list up above instead of copying through a stream I think (by presizing and using set, not add).

yep, benchmark shows that it's much faster than boxing a stream so I've updated the codes accordingly.
```
Benchmark                    Mode  Cnt      Score     Error   Units
ArrayCopy.copy_arrayListLoop thrpt   10  12776.515 ± 437.142  ops/ms
ArrayCopy.copy_streamBoxed  thrpt   10   4524.458 ±  90.628  ops/ms
```

I'm not comfortable with primitive arrays since it's impossible to mark them as immutable. and I've wrapper the list as a unmodifiable one. `ImmutableList` would be ideal here but guava is not in the dependencies so out of options.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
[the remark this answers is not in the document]
beanliu: unfortunately, the benchmark shows that the `synchronized` approach is much slower in a multi-threaded environment (not a performance expert so I have no idea why).

```
# explicit lock
Benchmark                                           Mode  Cnt      Score      Error   Units
DoubleHistogramBenchmark.aggregate_10Threads       thrpt   10  27148.945 ±   97.655  ops/ms
DoubleHistogramBenchmark.aggregate_1Threads        thrpt   10  40392.887 ±  591.675  ops/ms
DoubleHistogramBenchmark.aggregate_5Threads        thrpt   10  25110.300 ± 2177.205  ops/ms

# synchronized
Benchmark                                           Mode  Cnt       Score      Error   Units
DoubleHistogramBenchmark.aggregate_10Threads       thrpt   10    6688.401 ± 1013.960  ops/ms
DoubleHistogramBenchmark.aggregate_1Threads        thrpt   10  102054.704 ± 8556.397  ops/ms
DoubleHistogramBenchmark.aggregate_5Threads        thrpt   10    5047.048 ±  984.942  ops/ms
```

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
```diff
@@ -12,14 +12,21 @@
...
 import java.util.concurrent.locks.ReentrantReadWriteLock;
-import java.util.stream.Collectors;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import javax.annotation.concurrent.GuardedBy;
 
 final class DoubleHistogramAggregator extends AbstractAggregator<HistogramAccumulation> {
+  private static final Logger logger = Logger.getLogger(DoubleHistogramAggregator.class.getName());
+
+  private static boolean LoggedMergingInvalidBoundaries = false;
```
jkwatson, line 28: this should be lower-cased, and probably also volatile, so it can be updated cross-thread safely.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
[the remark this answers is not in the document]
jkwatson: a wrapper class around the array would probably be a viable option.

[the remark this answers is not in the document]
beanliu: > a wrapper class around the array would probably be a viable option.

100% agree. but it seems to me it would be reinventing the wheels like `ImmutableList`. and where should we put the class in the hierarchy? not sure sdk is the right place tho.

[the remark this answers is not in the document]
jkwatson: ImmutableList boxes every item in the list individually, meaning you have n extra objects per list, rather than 1 extra object in a single wrapper for an array.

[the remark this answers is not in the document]
beanliu: are you suggesting something like [ImmutableLongArray](https://guava.dev/releases/22.0/api/docs/com/google/common/primitives/ImmutableLongArray.html) and [ImmutableDoubleArray](https://guava.dev/releases/22.0/api/docs/com/google/common/primitives/ImmutableDoubleArray.html) to completely get rid of the boxing? since #2148 I can't use these classes directly but have to implement a similar & simpler version in the sdk.  I don't have a strong opinion of that since I don't have much understanding of the project context, and I would add them if you think that's a good idea.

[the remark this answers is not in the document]
jkwatson: Something like that, yes, but I don't think we need all the methods and functionality that the guava classes currently provide.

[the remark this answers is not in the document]
beanliu: I've replaced all read-only primitive arrays with ImmutableLongArray/ImmutableDoubleArray.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  @Override
+  public HistogramAccumulation accumulateLong(long value) {
+    return HistogramAccumulation.create(
+        1, value, ImmutableDoubleArray.of(), ImmutableLongArray.of(1));
+  }
+
+  static final class Handle extends AggregatorHandle<HistogramAccumulation> {
+    private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
```
jkwatson, line 108: to follow-up on the lock...if we only need write, could we just use a normal `ReentrantLock` rather than the more complex read/write version?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/common/ImmutableDoubleArray.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.Arrays;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+public class ImmutableDoubleArray {
+  private static final ImmutableDoubleArray EMPTY = new ImmutableDoubleArray(new double[0]);
+
+  /** Returns the empty array. */
+  public static ImmutableDoubleArray of() {
```
jkwatson, line 17: do we need this method? let's keep the API as narrow as possible for now. Please remove methods that we don't need. thanks!

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/common/ImmutableDoubleArray.java
```diff
@@ -0,0 +1,112 @@
...
+    return EMPTY;
+  }
+
+  /** Returns an immutable array containing a single value. */
+  public static ImmutableDoubleArray of(double e0) {
```
jkwatson, line 22: this we can probably remove, too, since a single-bucket histogram seems...not very useful

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/common/ImmutableDoubleArray.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+    if (!(object instanceof ImmutableDoubleArray)) {
+      return false;
+    }
+    ImmutableDoubleArray that = (ImmutableDoubleArray) object;
+    if (this.length() != that.length()) {
+      return false;
+    }
+    for (int i = 0; i < length(); i++) {
+      if (!areEqual(this.get(i), that.get(i))) {
```
jkwatson, line 71: can we use `Arrays.equals()` here rather than re-implementing it?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/common/ImmutableLongArray.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.Arrays;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+public class ImmutableLongArray {
+  private static final ImmutableLongArray EMPTY = new ImmutableLongArray(new long[0]);
+
+  /** Returns the empty array. */
+  public static ImmutableLongArray of() {
```
jkwatson, line 17: same comments as for the double version. let's keep the API super narrow

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/MetricData.java
```diff
@@ -29,6 +29,8 @@
           /* isMonotonic= */ false, AggregationTemporality.CUMULATIVE, Collections.emptyList());
   private static final DoubleSummaryData DEFAULT_DOUBLE_SUMMARY_DATA =
       DoubleSummaryData.create(Collections.emptyList());
+  private static final DoubleHistogramData DEFAULT_DOUBLE_HISTOGRAM_DATA =
+      DoubleHistogramData.create(AggregationTemporality.CUMULATIVE, Collections.emptyList());
```
jkwatson, line 33: can we really hard-code CUMULATIVE here? Shouldn't it depend on how the aggregation is configured?

jkwatson:
This seems pretty close to me, but I'd like our local metrics expert, @bogdandrutu to review before we merge.

Also, it's worth noting that the OTel metrics data model is about to get a re-think, which could lead to needing to update things in here, depending on the outcome of those discussions.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/common/ImmutableDoubleArray.java
beanliu: this is required to construct an empty histogram accumulation when merging two accumulation with different boundaries. I won't argue that it's strictly needed though.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/common/ImmutableDoubleArray.java
beanliu: yep, not very useful in practise, but I mainly use it in the unit tests. otherwise it's a bit pain to write lots of `ImmutableDoubleArray.of(new double[]{1})` instead of `ImmutableDoubleArray.of(1)`.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/common/ImmutableLongArray.java
beanliu: this one is not used actually. I added it mainly to keep symmetry with `ImmutableDoubleArray`. I could get rid of this one for sure.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/MetricData.java
beanliu: it's a default & fallback value when we try to get the histogram data from an instance of `MetricData` whose type is not `HISTOGRAM`. I would assume something is wrong if this happens.

beanliu:
> This seems pretty close to me, but I'd like our local metrics expert, @bogdandrutu to review before we merge.
> Also, it's worth noting that the OTel metrics data model is about to get a re-think, which could lead to needing to update things in here, depending on the outcome of those discussions.

Thanks for the review and all the comments. as of the data model re-thinking, is there any reference that I could follow up?

jkwatson:
> Thanks for the review and all the comments. as of the data model re-thinking, is there any reference that I could follow up?

I think there are meetings being put on the public calendar where this is being discussed. I'd start by looking at recent metrics SIG meeting notes.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/HistogramAccumulation.java
```diff
@@ -0,0 +1,63 @@
...
+   * Creates a new {@link HistogramAccumulation} with the given values.
+   *
+   * @return a new {@link HistogramAccumulation} with the given values.
+   */
+  static HistogramAccumulation create(
+      long count, double sum, ImmutableDoubleArray boundaries, ImmutableLongArray counts) {
+    return new AutoValue_HistogramAccumulation(count, sum, boundaries, counts);
+  }
+
+  HistogramAccumulation() {}
```
aimiebell, line 28: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleHistogramData.java
```diff
@@ -0,0 +1,34 @@
...
+package io.opentelemetry.sdk.metrics.data;
+
+import com.google.auto.value.AutoValue;
+import java.util.Collection;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+@AutoValue
+public abstract class DoubleHistogramData implements Data<DoubleHistogramPointData> {
+  DoubleHistogramData() {}
```
aimiebell, line 15: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleHistogramPointData.java
```diff
@@ -0,0 +1,85 @@
...
+      Labels labels,
+      double sum,
+      long count,
+      ImmutableDoubleArray boundaries,
+      ImmutableLongArray counts) {
+    return new AutoValue_DoubleHistogramPointData(
+        startEpochNanos, epochNanos, labels, sum, count, boundaries, counts);
+  }
+
+  DoubleHistogramPointData() {}
```
aimiebell, line 45: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/MetricData.java
```diff
@@ -140,6 +142,28 @@ public static MetricData createDoubleSummary(
...
+        resource,
+        instrumentationLibraryInfo,
+        name,
+        description,
+        unit,
+        MetricDataType.HISTOGRAM,
+        data);
+  }
+
   MetricData() {}
```
aimiebell, line 167: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/HistogramAccumulation.java
anuraaga: Hi @aimiebell - this bot seems to have incorrect detection of method name problems. We also have many different checks that any new one would need to be aligned with. And I guess it's not really good practice to enable a bot against production repositories without getting permission first - currently this is causing problems for contributors. Can you disable the bot please? Thanks!

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/common/ImmutableDoubleArray.java
bogdandrutu: I think we should remove this as well as the boundaries from the Accumulation.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/HistogramAccumulation.java
```diff
@@ -0,0 +1,63 @@
...
+import com.google.auto.value.AutoValue;
+import io.opentelemetry.api.common.Labels;
+import io.opentelemetry.sdk.metrics.common.ImmutableDoubleArray;
+import io.opentelemetry.sdk.metrics.common.ImmutableLongArray;
+import io.opentelemetry.sdk.metrics.data.DoubleHistogramPointData;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+@AutoValue
+public abstract class HistogramAccumulation {
```
bogdandrutu, line 17: A bit of explanation about `Accumulation` that may help to understand why boundaries don't need to be here.

The framework that we build and uses the Aggregator guarantees that an instance of the `Accumulation` that is created by a specific instance of an Aggregator will be `merged` and converted to metrics by the same specific Aggregator instance.

So as long as one Aggregator instance produces all Accumulations with constant boundaries you don't need to worry about merging accumulations with different boundaries.

bogdandrutu:
@beanliu sorry for taking too long to look into this. We just finished the tracing review, and we can now look into metrics again :)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/HistogramAccumulation.java
bogdandrutu, line 17: I think this can be made package protected, no need to be public.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/common/ImmutableDoubleArray.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+package io.opentelemetry.sdk.metrics.common;
+
+import java.util.Arrays;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+public class ImmutableDoubleArray {
```
bogdandrutu, line 13: Even if you use this internally (see my previous comment why not needed), we should make this package protected and when we get boundries from the user we get them as array or list and internally copy them and store like this class.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/common/ImmutableLongArray.java
```diff
@@ -0,0 +1,102 @@
...
+ */
+
+package io.opentelemetry.sdk.metrics.common;
+
+import java.util.Arrays;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+public class ImmutableLongArray {
```
bogdandrutu, line 13: Same comment as for the Double version, I would prefer to not expose this publicly.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleHistogramPointData.java
```diff
@@ -0,0 +1,85 @@
...
+   */
+  public abstract long getCount();
+
+  /**
+   * The bucket boundaries. For a Histogram with N defined boundaries, e.g, [x, y, z]. There are N+1
+   * counts: [-inf, x), [x, y), [y, z), [z, +inf].
+   *
+   * @return the bucket boundaries in increasing order.
+   */
+  public abstract ImmutableDoubleArray getBoundaries();
```
bogdandrutu, line 67: We can have a List here, and we can return "unmodifiableList".

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleHistogramPointData.java
```diff
@@ -0,0 +1,85 @@
...
+
+  /**
+   * The counts in each bucket.
+   *
+   * @return the counts in each bucket.
+   */
+  public abstract ImmutableLongArray getCounts();
```
bogdandrutu, line 74: Same. (May be able to also use Arrays.asList, but need to check if that is immutable)

bogdandrutu:
@beanliu @jkwatson I do understand that this PR was initially split, but I still see 2 different PRs in this one:
* One that adds the "data" (changes in the data package)
* One that adds the "Aggregator" (changes in the aggregator package)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleHistogramPointData.java
beanliu: assume this is the result of marking `ImmutableDoubleArray` non-public so we can't use the class here?

imo, `ImmutableDoubleArray` is a better contract than `unmodifiableList` because the immutable property is guaranteeed at compile-time while the latter could cause a run time exception.

beanliu: I've mark `ImmutableDoubleArray` and `ImmutableLongArray` non-public so that it's only visible in the aggregator package. But this kind of makes the interface of the `DoubleHistogramPointData` dependent on the mutable `List` interface (it's `Collections.unmodifiableList` tho).

beanliu:
> but I still see 2 different PRs in this one: ...

@bogdandrutu thanks for the comments. not sure if I understand correctly, given the review is in progress, are u suggesting splitting this PR again?

bogdandrutu:
Unfortunately I have some time sensitive tasks at work for the next 2 days. I will come back to you asap.

Regarding splitting. I know I review but was a very high level review to be honest because there are a lot of changes in this PR. But I think if you were to split probably would be way faster to merge this.

beanliu:
this PR is split into #2923 (changes in the data package) and #2924 (changes in the aggregator package). will create follow-up PR to enable histogram exporters once the prerequisite PRs got merged. closing this off.

PR #2585, opened 2021-01-26
Move Sampler to SdkTracerProvider

anuraaga:
For #2022

Didn't bother with deprecating since it seems complicated to do here.

Given how often our unit tests set samplers, and rarely they set limits, there does seem to be a nice win here.

/cc @iNikem

### DIFF
--- a/sdk-extensions/otproto/src/main/java/io/opentelemetry/sdk/extension/otproto/TraceProtoUtils.java
+++ b/sdk-extensions/otproto/src/main/java/io/opentelemetry/sdk/extension/otproto/TraceProtoUtils.java
@@ -8,9 +8,6 @@ package io.opentelemetry.sdk.extension.otproto;
 import com.google.protobuf.ByteString;
 import io.opentelemetry.api.trace.SpanId;
 import io.opentelemetry.api.trace.TraceId;
-import io.opentelemetry.proto.trace.v1.ConstantSampler;
-import io.opentelemetry.sdk.trace.config.TraceConfig;
-import io.opentelemetry.sdk.trace.samplers.Sampler;
 
 /** Utilities for converting various objects to protobuf representations. */
 final class TraceProtoUtils {
@@ -22,7 +19,7 @@ final class TraceProtoUtils {
    * @param spanId the spanId to convert.
    * @return a ByteString representation.
    */
-  public static ByteString toProtoSpanId(String spanId) {
+  static ByteString toProtoSpanId(String spanId) {
     return ByteString.copyFrom(SpanId.bytesFromHex(spanId, 0));
   }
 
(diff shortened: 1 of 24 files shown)


sdk-extensions/otproto/src/main/java/io/opentelemetry/sdk/extension/otproto/TraceProtoUtils.java
```diff
@@ -32,50 +29,7 @@ public static ByteString toProtoSpanId(String spanId) {
...
     return ByteString.copyFrom(TraceId.bytesFromHex(traceId, 0));
   }
-
-  /**
-   * Returns a {@code TraceConfig} from the given proto.
-   *
-   * @param traceConfigProto proto format {@code TraceConfig}.
-   * @return a {@code TraceConfig}.
-   */
-  public static TraceConfig traceConfigFromProto(
```
anuraaga, line 45: This class was already package-private, forgot to clean up the methods

sdk-extensions/zpages/src/main/java/io/opentelemetry/sdk/extension/zpages/TraceConfigzZPageHandler.java
```diff
@@ -121,7 +121,7 @@ private static void emitChangeTable(PrintStream out) {
         /* rowName= */ "SamplingProbability to",
         /* paramName= */ QUERY_STRING_SAMPLING_PROBABILITY,
         /* inputPlaceHolder= */ "[0.0, 1.0]",
-        /* paramDefaultValue= */ TraceConfig.getDefault().getSampler().getDescription(),
```
anuraaga, line 124: Pretty sure this was wrong, it would look like `SamplingProbability to ParentBased{AlwaysOnSampler}`

sdk-extensions/zpages/src/main/java/io/opentelemetry/sdk/extension/zpages/TraceConfigzZPageHandler.java
```diff
@@ -210,7 +210,7 @@ private void emitActiveTable(PrintStream out) {
     emitActiveTableRow(
         /* out= */ out,
         /* paramName= */ "Sampler",
-        /* paramValue=*/ configSupplier.get().getSampler().getDescription(),
+        /* paramValue=*/ configSupplier.getSampler().getDescription(),
```
anuraaga, line 213: Now will always show something like `TraceIdRatio{0.2}`

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/config/TraceConfigBuilder.java
```diff
@@ -31,39 +26,6 @@
...
-  }
-
-  /**
-   * Sets the global default {@code Sampler}. It must be not {@code null} otherwise {@link #build()}
-   * will throw an exception.
-   *
-   * @param samplerRatio the global default ratio used to make decisions on {@link Span} sampling.
-   * @return this.
-   */
-  public TraceConfigBuilder setTraceIdRatioBased(double samplerRatio) {
```
anuraaga, line 54: Never noticed this crazy method. While I understand the sentiment, having some branches be parentBased and some not was extremely weird!

iNikem (approved):
Thank you! :)

jkwatson (approved):
Moving the Sampler into the shared state seems like a safe move. Thanks for getting this in as the last big change for 1.0!

bogdandrutu:
@anuraaga  @jkwatson why is better to have sampler in a different place than the other config? I feel this is a very subjective change with not necessary very strong arguments, the fact that tests set more one config than other does not make that config a special case...

jkwatson:
> @anuraaga @jkwatson why is better to have sampler in a different place than the other config? I feel this is a very subjective change with not necessary very strong arguments, the fact that tests set more one config than other does not make that config a special case...

This was a specific request from @iNikem to make it easier/clearer on how to configure the SDK. Yes, totally subjective, but from direct user feedback.

jkwatson:
Merging so we can get ready for a 1.0 RC. @bogdandrutu if you have further issues, we can revisit and consider reverting.

bogdandrutu:
> This was a specific request from @iNikem to make it easier/clearer on how to configure the SDK. Yes, totally subjective, but from direct user feedback.

I do believe consistency is more important.

@iNikem where do you set the `Sampler` that much that it bothers you? I think Sampler will be configured only at the initialization time (99% of the time). So I don't see any big improvement

PR #2586, opened 2021-01-26
Migrate to using gradle platform to manage dependency versions instea…

anuraaga:
…d of global variables.

I've generally found the pattern of using dependency management to be easier to work with than global variables for version management.

- No reinventing names - the name to reference a dependency is it's official group/module
  - No ambiguity on whether it's supposed to be underscore, camelcase, etc just copy-paste from maven search
  - Uses mental memory of artifact names, instead of per-project variable names
- No double-checking the base file for the variable name, especially problematic with groovy not being compiled so easy to get it wrong
- More familiar to uses coming from Maven, this is very similar to maven dependency management

One main goal is to make the project consistent with instrumentation repo - we already find the downside of reinventing names, we use completely different pattern between the two repos. Could settle on one, or just use normal maven coordinates instead as here, which I believe is less cognitive load in the long run especially for occasional contributors.


build.gradle
```diff
@@ -315,42 +225,44 @@ subprojects {
...
         dependencies {
             configurations.all {
-                // Kotlin compiler classpaths don't support BOM nor need it.
-                if (it.name.endsWith('Classpath') && !it.name.startsWith('kotlin')) {
-                    boms.each {bom ->
-                        add(it.name, enforcedPlatform(bom))
-                    }
-                }
+                // Gradle and newer plugins will set these configuration properties correctly.
+                if (it.canBeResolved && !it.canBeConsumed
```
Oberon00, line 229: This looks like a workaround.

dependencyManagement/build.gradle.kts
```diff
@@ -0,0 +1,114 @@
+plugins {
+    `java-platform`
+}
+
+data class DependencySet(val group: String, val version: String, val modules: List<String>)
```
Oberon00, line 5: To me, all this looks rather unfamiliar, not knowing kotlin. Do we really want to add another language to our buildscripts?

And is this `DependencySet` class, along with the boilerplate loop required to apply it, really better than inventing variable names for each dependency set?

Oberon00 (commented):
I'm a bit skeptical about this. I get that it theoretically makes some things simpler (less names to invent), but to me it seems that all in all, the build becomes even more black magic with this.

build.gradle
anuraaga: The below line is the workaround :) This line is the recommendation I got from Gradle team a while ago on how to apply dependency management with platforms.

dependencyManagement/build.gradle.kts
anuraaga: The plan is to remove Groovy since lack of completion / compile safety makes it quite tedious for me to maintain the scripts. We already have kotlin in opentelemetry-kotlin, after a migration, we would have one less programming language, Groovy, though.

> And is this DependencySet class, along with the boilerplate loop required to apply it, really better than inventing variable names for each dependency set?

I think it's an implementation detail of this script so that the downstream projects just have to use maven coordinates for defining dependencies. It's more decoupling of the base and downstream projects than we have right now.

opencensus-shim/src/test/java/io/opentelemetry/opencensusshim/InteroperabilityTest.java
```diff
@@ -335,6 +357,151 @@ public void testNoRecordDoesNotExport() {
     verify(spanExporter, never()).export(anyCollection());
   }
 
+  @Test
+  @SuppressWarnings("deprecation") // Summary is deprecated in census
+  void testSupportedMetricsExportedCorrectly() {
```
jkwatson, line 362: seems like some non-gradle work crept into this PR. Can we limit it to just gradle changes?

anuraaga: Yup moved

sdk/common/src/test/java/io/opentelemetry/sdk/common/CompletableResultCodeTest.java
```diff
@@ -149,7 +150,7 @@ void join() {
     CompletableResultCode result = new CompletableResultCode();
     new Thread(
             () -> {
-              Uninterruptibles.sleepUninterruptibly(50, TimeUnit.MILLISECONDS);
+              Uninterruptibles.sleepUninterruptibly(Duration.ofMillis(50));
```
jkwatson, line 153: eh? more stuff creeping in!

anuraaga: Oops sorry, moved this too

jkwatson (approved):
anything that moves us away from groovy is ok by me. 😉

jkwatson:
It looks like we lost publishing of our source jars with this change. all the other publishing seems to be fine, though

PR #2589, opened 2021-01-27
Merge metric and trace OC shim interop tests.

anuraaga:
Metric tests were using junit4 even though we want to converge on junit5. But Switching to junit5 has both interop tests run in the same classloader and conflict on the shim's static initialization requirements. Easiest workaround is to merge them into one test.


jkwatson (approved):
👍

PR #2590, opened 2021-01-27
Migrate to guava JRE version

anuraaga:
We only use guava in tests now, so we can use JRE version which has less chance of conflicts with other libraries.

### DIFF
--- a/opencensus-shim/src/test/java/io/opentelemetry/opencensusshim/InteroperabilityTest.java
+++ b/opencensus-shim/src/test/java/io/opentelemetry/opencensusshim/InteroperabilityTest.java
@@ -62,8 +62,10 @@ import io.opentelemetry.sdk.trace.data.StatusData;
 import io.opentelemetry.sdk.trace.export.SimpleSpanProcessor;
 import io.opentelemetry.sdk.trace.export.SpanExporter;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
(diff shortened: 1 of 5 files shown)


opencensus-shim/src/test/java/io/opentelemetry/opencensusshim/InteroperabilityTest.java
```diff
@@ -412,28 +414,23 @@ void testSupportedMetricsExportedCorrectly() {
       statsRecorder.newMeasureMap().put(latency2, 60).record();
     }
     List<List<MetricData>> exported = metricExporter.waitForNumberOfExports(3);
-    List<MetricData> metricData = exported.get(2);
+    List<MetricData> metricData =
```
anuraaga, line 417: I don't think this is supposed to relate to Guava, but somehow this test has a different order in #2586. Sorting is the sane thing to do though I think.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/StressTestRunner.java
```diff
@@ -44,7 +43,7 @@ final void run() {
                 for (int i = 0; i < operation.getNumOperations(); i++) {
                   operation.getUpdater().update();
                   Uninterruptibles.sleepUninterruptibly(
-                      operation.getOperationDelayMs(), TimeUnit.MILLISECONDS);
+                      Duration.ofMillis(operation.getOperationDelayMs()));
```
anuraaga, line 46: error prone

jkwatson (approved):
👍

PR #2592, opened 2021-01-27
Migrate build.gradle to kotlin

anuraaga:
Did this first since it's the doozy. The code's too different to have any reasonable diff with the groovy so just going to have to eyeball I think.


build.gradle.kts
```diff
@@ -0,0 +1,559 @@
...
+import net.ltgt.gradle.errorprone.ErrorPronePlugin
+import org.gradle.api.plugins.JavaPlugin.*
+import org.gradle.api.tasks.testing.logging.TestExceptionFormat
+import org.jfrog.gradle.plugin.artifactory.dsl.ArtifactoryPluginConvention
+import org.jfrog.gradle.plugin.artifactory.dsl.PublisherConfig
+import org.jfrog.gradle.plugin.artifactory.dsl.ResolverConfig
+import ru.vyarus.gradle.plugin.animalsniffer.AnimalSniffer
+import ru.vyarus.gradle.plugin.animalsniffer.AnimalSnifferPlugin
+
+plugins {
```
jkwatson, line 19: would it make sense to have 2 `plugins` closures? One for the things we apply universally and one for the bits we don't?

jkwatson: hm. looks like we can't have 2 of them, but maybe we can separate the `apply false` from the others, just for clarity?

build.gradle.kts
```diff
@@ -0,0 +1,559 @@
...
+}
+
+if (!JavaVersion.current().isJava11Compatible()) {
+    throw GradleException("JDK 11 or higher is required to build. " +
+            "One option is to download it from https://adoptopenjdk.net/. If you believe you already " +
+            "have it, please check that the JAVA_HOME environment variable is pointing at the " +
+            "JDK 11 installation.")
+}
+
+// Nebula plugin will not configure if .git doesn't exist, let's allow building on it by stubbing it
```
jkwatson, line 39: how plausible will it be, now that we're using kotlin, to do some refactoring of the build script to make it more modular? Can we move pieces into files that are appropriately named to make it really easy to find the pieces related to publishing, for example?

build.gradle.kts
```diff
@@ -0,0 +1,559 @@
...
+                    allDisabledChecksAsWarnings.set(true)
+
+                    // Doesn't currently use Var annotations.
+                    disable("Var") // "-Xep:Var:OFF"
+
+                    // ImmutableRefactoring suggests using com.google.errorprone.annotations.Immutable,
+                    // but currently uses javax.annotation.concurrent.Immutable
+                    disable("ImmutableRefactoring") // "-Xep:ImmutableRefactoring:OFF"
+
+
```
jkwatson, line 164: there are a couple of extra CR/LFs scattered in here. on purpose, or can we clean this up?

build.gradle.kts
```diff
@@ -0,0 +1,559 @@
...
+                                license {
+                                    name.set("The Apache License, Version 2.0")
+                                    url.set("http://www.apache.org/licenses/LICENSE-2.0.txt")
+                                }
+                            }
+
+                            developers {
+                                developer {
+                                    id.set("opentelemetry")
+                                    name.set("OpenTelemetry Gitter")
```
jkwatson, line 458: we should definitely change this! [doesn't have to be in this PR]

jkwatson (approved):
less groovy makes me more happy. 
👍

build.gradle.kts
anuraaga: Good idea

build.gradle.kts
anuraaga: Yup I think we can do some splitting soon

anuraaga (commented):
Tried publishToMavenLocal with this PR and see sources jar. Weird, will see if that holds for snapshots too wouldn't expect a change though.

PR #2594, opened 2021-01-27
OtTracerPropagator: inject and extract baggage

malafeev:


extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/OtTracerPropagator.java
```diff
@@ -66,6 +70,21 @@ public static OtTracerPropagator getInstance() {
...
     setter.set(carrier, SAMPLED_HEADER, String.valueOf(spanContext.isSampled()));
+
+    Baggage baggage = Baggage.fromContext(context);
+    if (!baggage.isEmpty()) {
+      baggage.forEach(
+          (key, value, metadata) -> {
+            String metadataValue = metadata.getValue();
+            String baggageValue;
+            if (metadataValue != null && !metadataValue.isEmpty()) {
+              baggageValue = value + ";" + metadataValue;
```
carlosalberto, line 81: Metadata is not supported for OT, similarly to Jaeger's format ;(

carlosalberto:
Thanks for the PR. One thing that we need to ponder is whether `Baggage` can be injected even amidst an invalid `SpanContext` - and I think we should, to be on the safe side, just as the Jaeger Propagator [does](https://github.com/open-telemetry/opentelemetry-java/blob/main/extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/JaegerPropagator.java#L90), i.e.

```java
    SpanContext spanContext = Span.fromContext(context).getSpanContext();
    if (spanContext.isValid()) {
      injectSpan(spanContext, carrier, setter);
    }

    injectBaggage(Baggage.fromContext(context), carrier, setter);
```

extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/OtTracerPropagator.java
malafeev: I removed metadata from inject.
Should I also remove it from extract?

carlosalberto: Yes, please ;)

malafeev: done

carlosalberto:
After confirming to what other OT Propagators do, we don't really expect to inject/extract `Baggage` if `SpanContext` extraction/injection failed. So we are (almost) good to go! Left a pair comments but overall LGTM!

extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/OtTracerPropagator.java
```diff
@@ -66,6 +69,15 @@ public static OtTracerPropagator getInstance() {
         spanContext.getTraceIdAsHexString().substring(TraceId.getHexLength() / 2));
     setter.set(carrier, SPAN_ID_HEADER, spanContext.getSpanIdAsHexString());
     setter.set(carrier, SAMPLED_HEADER, String.valueOf(spanContext.isSampled()));
+
```
carlosalberto, line 72: Leave a note mentioning that `Baggage` is only injected if there was a valid `SpanContext` - likewise for extraction ;)

extensions/trace-propagators/src/test/java/io/opentelemetry/extension/trace/propagation/OtTracerPropagatorTest.java
```diff
@@ -114,6 +115,20 @@ void inject_NotSampledContext() {
     assertThat(carrier).containsEntry(OtTracerPropagator.SAMPLED_HEADER, "false");
   }
 
+  @Test
```
carlosalberto, line 118: Have an additional test, to make sure that upon no valid `SpanContext` in `Context`, the `Bagagge` **was not** injected? Likewise for extraction ;)

malafeev: added

extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/OtTracerPropagator.java
malafeev: done

extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/OtTracerPropagator.java
```diff
@@ -85,7 +98,29 @@ public static OtTracerPropagator getInstance() {
     if (!spanContext.isValid()) {
       return context;
     }
-    return context.with(Span.wrap(spanContext));
+
+    Context extractedContext = context.with(Span.wrap(spanContext));
+
+    // Baggage is only extracted if there is a valid SpanContext
+    if (carrier != null) {
```
carlosalberto, line 105: If we are in this line, it means the extracted `SpanContext` is valid, which means the carrier is not `null`, correct?

carlosalberto (approved):

carlosalberto:
Left a small comment but LGTM. Thanks for updating this PR!

extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/OtTracerPropagator.java
malafeev: there is a check above:
```
if (!spanContext.isValid()) { return context; }
```
in this line SpanContext is valid, but `carrier` could be potentially `null`.

PR #2595, opened 2021-01-27
Enable auto-configuration of custom, named span exporters.

jkwatson:
This need was uncovered while working on a custom distribution of the opentelemetry auto-instrumentation agent.

In order for this to work, we really wanted access to the `ConfigProperties` as well, so we could leverage the standardized property handling that it provides.

Alternative approaches would be welcome, but this seems like a pretty clean pattern. In fact, we might want to add the `ConfigProperties` to other auto-configuration SPI methods as well, so we don't have to figure out how to add them later.

### DIFF
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
@@ -46,7 +46,7 @@ public class ConfigProperties {
   }
 
   /**
-   * Get a string-valued configuration property.
+   * Returns a string-valued configuration property.
    *
    * @return null if the property has not been configured.
    */
@@ -56,7 +56,7 @@ public class ConfigProperties {
   }
 
   /**
-   * Get a integer-valued configuration property.
+   * Returns a integer-valued configuration property.
    *
    * @return null if the property has not been configured.
    * @throws NumberFormatException if the property is not a valid integer.
@@ -76,7 +76,7 @@ public class ConfigProperties {
   }
 
   /**
-   * Get a long-valued configuration property.
+   * Returns a long-valued configuration property.
    *
    * @return null if the property has not been configured.
    * @throws NumberFormatException if the property is not a valid long.
@@ -96,7 +96,7 @@ public class ConfigProperties {
   }
 
   /**
-   * Get a double-valued configuration property.
+   * Returns a double-valued configuration property.
    *
    * @return null if the property has not been configured.
    * @throws NumberFormatException if the property is not a valid double.
@@ -116,7 +116,7 @@ public class ConfigProperties {
   }
 
   /**
-   * Get a list-valued configuration property. The format of the original value must be
+   * Returns a list-valued configuration property. The format of the original value must be
    * comma-separated. Empty values will be removed.
    *
    * @return an empty list if the property has not been configured.
@@ -130,7 +130,7 @@ public class ConfigProperties {
   }
 
   /**
-   * Get a map-valued configuration property. The format of the original value must be
+   * Returns a map-valued configuration property. The format of the original value must be
    * comma-separated for each key, with an '=' separating the key and value. For instance, <code>
    * service.name=Greatest Service,host.name=localhost</code> Empty values will be removed.
    *
@@ -156,7 +156,7 @@ public class ConfigProperties {
   }
 
   /**
-   * Get a boolean-valued configuration property. Uses the same rules as {@link
+   * Returns a boolean-valued configuration property. Uses the same rules as {@link
    * Boolean#parseBoolean(String)} for handling the values.
    *
    * @return false if the property has not been configured.
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
@@ -16,7 +16,13 @@ import java.util.Map;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
-class ConfigProperties {
+/**
+ * Properties to be used for auto-configuration of the OpenTelemetry SDK components. These
+ * properties will be a combination of system properties and environment variables. The properties
+ * for both of these will be normalized to be all lower case, and underscores will be replaced with
+ * periods.
+ */
+public class ConfigProperties {
 
   private final Map<String, String> config;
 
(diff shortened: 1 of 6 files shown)


kenfinnigan (approved):
Seems reasonable to me

iNikem (approved):

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/ConfigurableSpanExporterProvider.java
```diff
@@ -0,0 +1,31 @@
...
+import io.opentelemetry.sdk.autoconfigure.ConfigProperties;
+import io.opentelemetry.sdk.trace.export.SpanExporter;
+
+/**
+ * A service provider interface (SPI) for providing additional exporters that can be used with the
+ * autoconfigured SDK. If the {@code otel.trace.exporter} property contains a value equal to what is
+ * returned by {@link #getName()}, the exporter returned by {@link
+ * #createExporter(ConfigProperties)} will be enabled and added to the SDK.
+ */
+public interface ConfigurableSpanExporterProvider {
```
breedx-splk, line 17: I know there's precedence with `ConfigurablePropagatorProvider`, but I think the `Configurable` prefix is kinda redundant/unnecessary.  We're already in the autoconfigure package space.  What about simply `SpanExporterProvider`?

sdk-extensions/autoconfigure/src/testFullConfig/java/io/opentelemetry/sdk/autoconfigure/TestConfigurableSpanExporterProvider.java
```diff
@@ -0,0 +1,54 @@
...
+      return shouldAlwaysFail
+          ? CompletableResultCode.ofFailure()
+          : CompletableResultCode.ofSuccess();
+    }
+
+    @Override
+    public CompletableResultCode shutdown() {
+      return shouldAlwaysFail
+          ? CompletableResultCode.ofFailure()
+          : CompletableResultCode.ofSuccess();
```
breedx-splk, line 51: factor out private internal method to eliminate this duplication?

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/ConfigurableSpanExporterProvider.java
```diff
@@ -0,0 +1,31 @@
...
+  /**
+   * Returns a {@link SpanExporter} that can be registered to OpenTelemetry by providing the
+   * property value specified by {@link #getName()}.
+   */
+  SpanExporter createExporter(ConfigProperties config);
+
+  /**
+   * Returns the name of this exporter, which can be specified with the {@code otel.trace.exporter}
+   * property to enable it. If the name is the same as any other defined exporter name, it is
+   * undefined which will be used.
```
breedx-splk, line 28:
```suggestion
   * property to enable it. The name returned should NOT be the same as any other exporter name.
   * If the name does conflict with another exporter name, the resulting behavior is undefined and 
   * it is explicitly unspecified which exporter will actually be used.
 ```

**Review by breedx-splk — CHANGES_REQUESTED** [review]
Looks good to me too, just a few small ideas.

File: sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/ConfigurableSpanExporterProvider.java (hunk 1 above)
Comment (jkwatson) [reply sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/ConfigurableSpanExporterProvider.java:17]: @anuraaga your opinion?

File: sdk-extensions/autoconfigure/src/testFullConfig/java/io/opentelemetry/sdk/autoconfigure/TestConfigurableSpanExporterProvider.java (hunk 2 above)
Comment (jkwatson) [reply sdk-extensions/autoconfigure/src/testFullConfig/java/io/opentelemetry/sdk/autoconfigure/TestConfigurableSpanExporterProvider.java:51]: sure...I was pondering ripping this out and just exposing the ConfigProperties on this test class which seems a little less obtuse and simpler.

Comment (jkwatson) [reply sdk-extensions/autoconfigure/src/testFullConfig/java/io/opentelemetry/sdk/autoconfigure/TestConfigurableSpanExporterProvider.java]: ripped this stuff out and made the test simpler. please take a look.

Comment (breedx-splk) [reply sdk-extensions/autoconfigure/src/testFullConfig/java/io/opentelemetry/sdk/autoconfigure/TestConfigurableSpanExporterProvider.java]: Seems more straightforward now.  Is was testing the config through the methods, now it can just get the config.  Simpler.

File: sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java [hunk 4]
```diff
@@ -39,14 +45,25 @@ private ConfigProperties(Map<?, ?> systemProperties, Map<String, String> environ
     this.config = config;
   }
 
+  /**
+   * Get a string-valued configuration property.
```
anuraaga, line 49: Returns instead of get for all

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/ConfigurableSpanExporterProvider.java
anuraaga: Without configurable, I'd expect the exporter to be automatically enabled, not only made configurable. Not a huge deal either way though.

anuraaga (approved):
Thanks for the change! I was expecting both public properties and configurable exporter provider to happen so no surprise here :-)

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
jkwatson: sure, I'll update them

jkwatson: done

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/ConfigurableSpanExporterProvider.java
breedx-splk: I'm not sure why the configurability speaks to the idea of enablement or not....but I am fine leaving it for now since there is prior art.  If it bothers us later, we can always change it.

breedx-splk (approved):

PR #2596, opened 2021-01-28
Publish artifacts directly to sonatype.

anuraaga:
This is the same configuration we use for instrumentation so is just bringing it over. While we haven't had anywhere near as many stability problems in this repo, it's good to be consistent, and fixes an issue we're seeing related to jars on artifactory. A lot less config, a lot more kotlin DSL to achieve a similar result :)

Ran `./gradlew snapshot` locally and it seems to check out (including javadoc / sources jar which currently artifactory doesn't have for whatever reason).
https://oss.sonatype.org/content/repositories/snapshots/io/opentelemetry/opentelemetry-api/0.15.0-SNAPSHOT/


README.md
```diff
@@ -99,7 +99,7 @@ We strongly recommend using our published BOM to keep all dependency versions in
   <repositories>
     <repository>
       <id>oss.sonatype.org-snapshot</id>
```
anuraaga, line 101: I didn't have to change this :P

iNikem (approved):

jkwatson (approved):
let's give it a whirl. I'll revert if it doesn't work when merged.

jkwatson:
unfortunately, we lost `bom` publishing with this change

jkwatson:
I think we actually had lost bom publishing with an earlier change.

PR #2599, opened 2021-01-28
apply the publishing configuration directly, rather than inside the java plugin config

jkwatson:
I think this fixes the issue with the bom not being published (at least it did locally)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -402,95 +402,95 @@ subprojects {
                 }
             }
         }
+    }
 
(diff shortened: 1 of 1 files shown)


jsuereth (approved):
Not that what I say means anything, but I just did similar build finagling for Google-Cloud-Exporters.  Yay for automation and "fun" with plugin magic.

anuraaga (approved):
Ah thanks, yeah the extra indent wasn't intended

PR #2601, opened 2021-01-29
Apply dependency management to gradle-protobuf-plugin classpath inste…

anuraaga:
…ad of implementation to prevent POM leak


jkwatson (approved):
🤞

jkwatson:
this is looking good. thanks!

PR #2605, opened 2021-01-29
Use OpenCensus implementation for binary propagation instead of copying.

anuraaga:
Less code is good code

/cc @jsuereth


opencensus-shim/src/test/java/io/opentelemetry/opencensusshim/OpenTelemetryPropagationComponentImplTest.java
```diff
@@ -33,8 +33,8 @@
 
   @Test
   public void implementationOfBinary() {
```
jsuereth, line 35: Is this test still useful? Maybe we shouldn't test the impl anymore as we don't control it, that way if I need to fix opencensus for whatever reason, this doesn't silently break.

jsuereth (approved):
Thanks for the fix!  Completely blew past the fact this class was NOT marked final (many others were).

I think when we first designed this impl was NOT a dependency and this should have been fixed as soon as it became one.  Thanks!

jkwatson (approved):

opencensus-shim/src/test/java/io/opentelemetry/opencensusshim/OpenTelemetryPropagationComponentImplTest.java
anuraaga: I think we can deal with it if it happens - I wanted to have this so that if the assertion fails because the class is now `NoopBinaryFormatImpl` for whatever reason, we will know something significant needs to be done :P

PR #2608, opened 2021-01-29
Use duration parser in autoconfiguration.

anuraaga:
Sweet


jsuereth (approved):
Thanks fo wiring this in so quickly!

jkwatson (approved):
Thanks!

PR #2609, opened 2021-01-29
Migrate api / sdk projects to build.gradle.kts

anuraaga:


sdk/all/build.gradle
```diff
@@ -1,47 +0,0 @@
...
-    testImplementation "org.junit-pioneer:junit-pioneer"
-}
-
-sourceSets {
-    main {
-        output.dir("build/generated/properties", builtBy: 'generateVersionResource')
-    }
-}
-
-animalsniffer {
```
Oberon00, line 33: What's up with the removal of this?

anuraaga: Was a no-op, leftover from before moving exclusions to top-level I think.

sdk/common/build.gradle.kts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+sourceSets {
+    main {
+        output.dir("build/generated/properties", "builtBy" to "generateVersionResource")
+    }
+}
+
+tasks {
+    register("generateVersionResource") {
```
jkwatson, line 32: is there a way we can not repeat this task registration? It's also up in sdk-all right now.

sdk/common/build.gradle.kts
```diff
@@ -0,0 +1,82 @@
...
+        val propertiesDir = file("build/generated/properties/io/opentelemetry/sdk/common")
+        outputs.dir(propertiesDir)
+
+        doLast {
+            File(propertiesDir, "version.properties").writeText("sdk.version=${project.version}")
+        }
+    }
+}
+
+for (version in mrJarVersions) {
```
jkwatson, line 42: this also seems like a good candidate for extraction, especially if we're going to have more MR jars in the future.

sdk/metrics/build.gradle.kts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+sourceSets {
+    main {
+        output.dir("build/generated/properties", "builtBy" to "generateVersionResource")
+    }
+}
+
+tasks {
+    register("generateVersionResource") {
```
jkwatson, line 31: and another one here for generating resources.

sdk/trace/build.gradle.kts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+sourceSets {
+    main {
+        output.dir("build/generated/properties", "builtBy" to "generateVersionResource")
+    }
+}
+
+tasks {
+    register("generateVersionResource") {
```
jkwatson, line 57: :thonk:

jkwatson (approved):
Thanks!

sdk/trace/build.gradle.kts
anuraaga: Yeah noticed the duplication we already have and want to share in another PR

PR #2610, opened 2021-01-29
Remove deprecated setServiceName

anuraaga:
Moved the fallback to default resource logic closer to where service name is picked up in Jaeger. And for Zipkin, I think we accidentally lost filling in endpoint, so restoring that is a lot of the change.

### DIFF
--- a/exporters/zipkin/src/main/java/io/opentelemetry/exporter/zipkin/ZipkinSpanExporter.java
+++ b/exporters/zipkin/src/main/java/io/opentelemetry/exporter/zipkin/ZipkinSpanExporter.java
@@ -15,6 +15,7 @@ import io.opentelemetry.api.trace.Span.Kind;
 import io.opentelemetry.api.trace.StatusCode;
 import io.opentelemetry.sdk.common.CompletableResultCode;
 import io.opentelemetry.sdk.common.InstrumentationLibraryInfo;
+import io.opentelemetry.sdk.resources.Resource;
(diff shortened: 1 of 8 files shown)


carlosalberto (approved):

exporters/zipkin/src/main/java/io/opentelemetry/exporter/zipkin/ZipkinSpanExporter.java
```diff
@@ -143,15 +142,15 @@ private static String nullToEmpty(String value) {
...
     Attributes resourceAttributes = spanData.getResource().getAttributes();
 
     // use the service.name from the Resource, if it's been set.
     String serviceNameValue = resourceAttributes.get(ResourceAttributes.SERVICE_NAME);
     if (serviceNameValue == null) {
-      return localEndpoint;
+      serviceNameValue = Resource.getDefault().getAttributes().get(ResourceAttributes.SERVICE_NAME);
     }
-    return Endpoint.newBuilder().serviceName(serviceNameValue).build();
+    return Endpoint.newBuilder().serviceName(serviceNameValue).ip(localAddress).build();
```
jkwatson, line 153: Is the Endpoint builder re-usable? Could we avoid an allocation or two by keeping a handle to 
`Endpoint.newBuilder().ip(localAddress)` and just tacking on the serviceName each time and building?

jkwatson (approved):

PR #2616, opened 2021-01-29
Add the span kind and the tracer name/version to the logging exporter

jkwatson:


kenfinnigan (approved):
I wondered about this in my testing today but forgot to mention it

anuraaga (approved):

PR #2630, opened 2021-02-01
Mark SamplingResult as immutable

bogdandrutu:

### DIFF
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/SamplingResult.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/SamplingResult.java
@@ -13,11 +13,13 @@ import io.opentelemetry.api.trace.SpanContext;
 import io.opentelemetry.api.trace.TraceState;
 import io.opentelemetry.context.Context;
 import java.util.List;
+import javax.annotation.concurrent.Immutable;
 
 /**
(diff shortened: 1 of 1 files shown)


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/SamplingResult.java
```diff
@@ -13,11 +13,13 @@
 import io.opentelemetry.api.trace.TraceState;
 import io.opentelemetry.context.Context;
 import java.util.List;
+import jdk.nashorn.internal.ir.annotations.Immutable;
```
anuraaga, line 16: Looks like an accidental import

bogdandrutu: Will fix it... No idea why idea decided to import this

jkwatson: I think I've configured IDEA to never import anything from nashorn to avoid this.

bogdandrutu: Done

jkwatson (approved):
👍

PR #2639, opened 2021-02-01
Migrate proto to kts

anuraaga:


proto/build.gradle.kts
```diff
@@ -0,0 +1,70 @@
...
+    api("com.google.protobuf:protobuf-java")
+    api("io.grpc:grpc-api")
+    api("io.grpc:grpc-protobuf")
+    api("io.grpc:grpc-stub")
+}
+
+val protoVersion = "0.7.0"
+// To generate checksum, download the file and run "shasum -a 256 ~/path/to/vfoo.zip"
+val protoChecksum = "0b581c654b2360485b99c2de3731dd59275b0fe7b91d78e7f6c5efd5997f4c82"
+val protoArchivePath = "$buildDir/archives/opentelemetry-proto-${protoVersion}.zip"
```
Oberon00, line 26: Could we make it a `file` here already? Does `zipTree(file)` work?

Oberon00 (approved):

proto/build.gradle.kts
anuraaga: Thanks, it does

jkwatson (approved):

PR #2641, opened 2021-02-01
Use shadow plugin to generate bundled jars.

anuraaga:
Since that's its job, so less configuration on our side. Removes use of a potentially non-inclusive term too.


jkwatson (approved):

jkwatson:
closing and re-opening to bounce the GH actions, which look stuck

PR #2647, opened 2021-02-01
Move labels outside of the stable package

bogdandrutu:


jkwatson:
I agree we need to get them out of the stable module. But, unfortunately, if we keep them in the same java package, we'll break anyone who uses the java module system, since you can consume two modules that publish classes in the same package.

bogdandrutu:
> I agree we need to get them out of the stable module. But, unfortunately, if we keep them in the same java package, we'll break anyone who uses the java module system, since you can consume two modules that publish classes in the same package.

This is a temporary fix and kind of allowed since metrics is alpha. We will find a better fix after the 1.0 release

jkwatson:
> > I agree we need to get them out of the stable module. But, unfortunately, if we keep them in the same java package, we'll break anyone who uses the java module system, since you can consume two modules that publish classes in the same package.
> 
> This is a temporary fix and kind of allowed since metrics is alpha. We will find a better fix after the 1.0 release

So, is it ok that pulling in an alpha metrics dependency could break your whole API usage if you're using java modules? I'm not sure that it is. That would basically be saying that metrics users can't use Java modules. Maybe that's ok, but that potentially rules out a large swath of users who might want to try out metrics before it's stable.

bogdandrutu:
> Maybe that's ok, but that potentially rules out a large swath of users who might want to try out metrics before it's stable

I am talking about a very limited amount of time until we fix it, and happy to break that than releasing an alpha API as stable.

jkwatson:
> > Maybe that's ok, but that potentially rules out a large swath of users who might want to try out metrics before it's stable
> 
> I am talking about a very limited amount of time until we fix it, and happy to break that than releasing an alpha API as stable.

Could we just repackage before 1.0, then, rather than keeping them in the same package for 1.0?

bogdandrutu:
> Could we just repackage before 1.0, then, rather than keeping them in the same package for 1.0?

Can you clarify?

jkwatson:
> > Could we just repackage before 1.0, then, rather than keeping them in the same package for 1.0?
> 
> Can you clarify?

you could put Labels in a different package as a part of this PR.  :)

bogdandrutu:
@jkwatson done moved them to `...metrics.common`

anuraaga (approved):

jkwatson (approved):
Thanks!

jkwatson:
looks like this needs a rebase.

PR #2648, opened 2021-02-01
Remove unnecessary class

bogdandrutu:


jkwatson (approved):
thanks. there was some reason at some point that we needed this extra class, but obviously that reason has gone away. good cleanup!

PR #2649, opened 2021-02-01
Fix consistency between implementation/test of fromContextOrNull

bogdandrutu:

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/baggage/Baggage.java
+++ b/api/all/src/main/java/io/opentelemetry/api/baggage/Baggage.java
@@ -51,7 +51,7 @@ public interface Baggage extends ImplicitContextKeyed {
    * Baggage} if there is no baggage in the context.
    */
   static Baggage fromContext(Context context) {
-    Baggage baggage = fromContextOrNull(context);
+    Baggage baggage = context.get(BaggageContextKey.KEY);
     return baggage != null ? baggage : empty();
   }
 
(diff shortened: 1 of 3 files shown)


anuraaga (commented):
IIRC, sleuth might have needed to differentiate between invalid and null for composite propagator compatibility with Brave, please check if this use can be replaced.

https://github.com/spring-cloud-incubator/spring-cloud-sleuth-otel/blob/eac53772fc7ca63324da498aaef5ccfec4c28105/spring-cloud-sleuth-otel/src/main/java/org/springframework/cloud/sleuth/otel/propagation/CompositeTextMapPropagator.java#L108

bogdandrutu:
@anuraaga  Sent one PR https://github.com/spring-cloud-incubator/spring-cloud-sleuth-otel/pull/8 for a different thing, please double check that it is fine :)

> IIRC, sleuth might have needed to differentiate between invalid and null for composite propagator compatibility with Brave, please check if this use can be replaced.

Not sure, we use to have this in our code-base something similar, what did it happen with that?

anuraaga:
> Not sure, we use to have this in our code-base something similar, what did it happen with that?

We removed the shortcircuiting propagator since it's not in spec. For Brave compatibility, sleuth may need short circuiting though, not sure.

jkwatson:
It doesn't bother me to keep these very simple methods in the API. Happy to maintain them, especially if they are useful to someone.

carlosalberto (approved):

anuraaga:
@bogdandrutu Let's keep the method, but make the implementations of `fromContext` consistent

bogdandrutu:
@jkwatson PR updated

jkwatson (approved):
Thanks!

PR #2660, opened 2021-02-02
Expose the SpanContext to the SpanData

bogdandrutu:
This helps exporter to have access directly to bytes, hex or any other format we will expose for the ids in the SpanContext.

Question: Should we deprecate and remove getTraceId/getSpanId/isSampled/getTraceState/getParentSpanId and enforce users to use the methods in the SpanContext?

Updates https://github.com/open-telemetry/opentelemetry-java/issues/2666


anuraaga (approved):
Returning span context seems to make sense. The discussion of keeping helpers came up when adding `parentSpanContext` - the general consensus was keeping `getParentSpanId()` as a helper since it's so commonly needed. So I guess we can keep them.

bogdandrutu:
> Returning span context seems to make sense. The discussion of keeping helpers came up when adding parentSpanContext - the general consensus was keeping getParentSpanId() as a helper since it's so commonly needed. So I guess we can keep them.

Why returning the `hex` vs `bytes`? Maybe rename them `TraceIdAsHex` so in the future we can add the bytes version as well and also consistent with SpanContext?

anuraaga:
Yeah the renames to be consistent with SpanContext seems helpful, these methods happened to be returning strings for a long time I think.

jkwatson (approved):
Thanks!

Issue #2658, opened 2021-02-01
Should we return the interface or the object in static methods like `getInstance`?


bogdandrutu:
In the trace propagator we return the interface:
https://www.javadoc.io/doc/io.opentelemetry/opentelemetry-api/latest/io/opentelemetry/api/trace/propagation/W3CTraceContextPropagator.html

In the baggage propagator we return the implementation class:
https://www.javadoc.io/doc/io.opentelemetry/opentelemetry-api/latest/io/opentelemetry/api/baggage/propagation/W3CBaggagePropagator.html

We need to be consistent and return always one or the other. I am in favor of returning the implementation class so in the future if we add extra methods on the implementation class those will be exposed.

anuraaga:
Yeah let's return implementation, if a class is public there's no reason to not expose it in factories too and can add flexibility.

Oberon00:
> We need to be consistent

Why? I would return the interface unless the class has additional public methods.

bogdandrutu:
> I would return the interface unless the class has additional public methods.

Then you lock yourself for the future. You cannot change the return type (ABI breaking change) so even if you add more additional methods you will not be able to expose them.

jkwatson:
I would generally favor returning the more restrictive type, in this case being the interface, since it exposes a little bit less API surface. 

But, I don't feel incredibly strong about it, and can see both sides of the argument, so if @bogdandrutu and @anuraaga agree on the approach, I'll go along with for the ride.

anuraaga:
@jkwatson Didn't you point out returning the implementation for jaeger remote sampler when I hadn't? :P But yeah I'd say the reason is that we're not able to reduce "publicness" at all by returning the interface - so may as well return the implementation which may provide some benefit in the future. For propagators, the probability of this is extremely low, but for other implementations like the remote sampler, I think it's reasonably possible and we may as well be consistent.

bogdandrutu:
We should check for all these cases and fix them before 1.0 :)

PR #2661, opened 2021-02-02
Have w3ctracecontextpropagator actually implement textmappropagator.

anuraaga:
Updates https://github.com/open-telemetry/opentelemetry-java/issues/2658

I was expecting a one-word change, woah!

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/trace/propagation/W3CTraceContextPropagator.java
+++ b/api/all/src/main/java/io/opentelemetry/api/trace/propagation/W3CTraceContextPropagator.java
@@ -36,7 +36,7 @@ import javax.annotation.concurrent.Immutable;
  * designed to support all the data propagated via W3C propagation natively.
  */
 @Immutable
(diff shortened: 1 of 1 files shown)


bogdandrutu (approved):

Oberon00:
This weakens encapsulation somewhat. Is this to make the code easier to read?

anuraaga:
@Oberon00 If we could make the class `W3cTraceContextPropagator` non-public, then that is great, but we can't since propagator implementations are separate from the interface. If it's public, there's really no reason for `getInstance` to not return the type - while it's not likely for propagators especially, it does give the option of adding implementation-specific methods if it ever comes up. There isn't really a downside I would say since the class will be exposed either way so is always safe to be ABI-linked against.

Oberon00:
Currently the propagator class is a private inner class. The public class is just the `getInstance` method. But I guess this additional encapsulation is not really much of an advantage in practice.

jkwatson (approved):
👍  This is fine to change, pre 1.0. Thanks!

PR #2662, opened 2021-02-02
Tweak javadoc options

### LINKED ISSUE #2652: Artifact description for sdk-trace says `trace 0.15.0 API`
https://www.javadoc.io/doc/io.opentelemetry/opentelemetry-sdk-trace/latest/index.html

anuraaga:
- Sets title explicitly to same as project description
- Enables HTML5 since we always build with Java 11 using toolchains
- Couple of other miscellaneous

Fixes #2652

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -107,10 +107,6 @@ subprojects {
         val javaToolchains = the<JavaToolchainService>()
 
         tasks {
-            withType(Javadoc::class) {
-                exclude("io/opentelemetry/internal/**")
-            }
-
             val testJava8 by registering(Test::class) {
                 javaLauncher.set(javaToolchains.launcherFor {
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -214,11 +210,24 @@ subprojects {
             }
 
             withType(Javadoc::class) {
+                exclude("io/opentelemetry/internal/**")
```
bogdandrutu, line 213: Should we exclude any internal directory not only the main one? If any subpackage has it's own internal for example?

bogdandrutu (approved):

build.gradle.kts
anuraaga: Good point, and since we shouldn't have `internal` in the package of any public class just expanded the glob a bit

bogdandrutu (approved):

PR #2668, opened 2021-02-02
Have TraceConfig supply itself.

anuraaga:
I don't think we can lose the ability to simply set the result of the traceconfig builder when configuring the SDK since it should be much more common than needing dynamic updates.

This change allows it - I don't know if I like it better than what we have right now though.


jkwatson:
This is definitely an odd look. I would guess that users would be confused by the fact that you can pass in an instance which also happens to be a supplier. If I were looking at the API, would would think that I should pass in a lambda that returned the constant value that I wanted to use, rather than just pass in the instance.

Long story short: I would rather just have two methods, to make the API clear to the end user.

anuraaga:
Yeah - going to close this, let's stick with two methods. Two method overloads with same name, different type, is pretty common I guess

PR #2669, opened 2021-02-02
Publish BOM for alpha artifacts.

anuraaga:


bom-alpha/gradle.properties
```diff
@@ -0,0 +1 @@
+otel.release=alpha
```
jkwatson, line 1: 😂

jkwatson (approved):
Thanks!

Issue #2613, opened 2021-01-29
Duplication of attributes in `SpanWrapper.toString()`


kenfinnigan:
**Describe the bug**
When printing out `SpanData` instances that were captured, all the attributes on the span are output once as a value on the delegate of `SpanWrapper`, `RecordEventsReadableSpan`, and then again as attributes on `SpanWrapper` directly.

**What did you expect to see?**
I would expect to see the attributes listed only once unless the sets are not meant to be identical.

**What did you see instead?**
It was initially confusing as it had me wondering whether there was a copy, or they were different attributes that just happened to be the same in the case I saw.

**What version and what artifacts are you using?**
Artifacts: `opentelemetry-sdk`, `opentelemetry-sdk-testing`
Version: 0.14.1
How did you reference these artifacts?
```
        <dependency>
            <groupId>io.opentelemetry</groupId>
            <artifactId>opentelemetry-sdk</artifactId>
        </dependency>
        <dependency>
            <groupId>io.opentelemetry</groupId>
            <artifactId>opentelemetry-sdk-testing</artifactId>
            <scope>test</scope>
        </dependency>
```

**Environment**
Compiler: openjdk version "11.0.8"
OS: Mac Big Sur

**Additional context**
From a chat with @jkwatson on gitter, maybe there's a way to suppress the output of attributes on `SpanWrapper` when calling `toString()`?

bogdandrutu:
Can you add an output example?

anuraaga:
Yeah added a test to demonstrate this, but haven't gotten around to fixing it yet

https://github.com/open-telemetry/opentelemetry-java/blob/15fcadd8d7e7b139f30260bd575b112b827c2734/sdk/trace/src/test/java/io/opentelemetry/sdk/trace/SdkSpanBuilderTest.java#L897

kenfinnigan:
Thanks @anuraaga.

Is that sufficient @bogdandrutu?

bogdandrutu:
Yes, I actually found the bug but waiting for a decision on the https://github.com/open-telemetry/opentelemetry-java/pull/2660 before moving forward with the fix because that will influence a bit how I will fix it :)

kenfinnigan:
not a problem, just wanted to check nothing else was needed from me

PR #2671, opened 2021-02-02
Fix SpanWrapper toString to avoid duplicates

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-java/issues/2613


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -497,45 +497,4 @@ private Attributes getImmutableAttributes() {
     // otherwise, make a copy of the data into an immutable container.
     return attributes.immutableCopy();
   }
-
-  @Override
-  public String toString() {
```
jkwatson, line 502: do we really want to get rid of this toString()?  We can fix the other one without getting rid of this one.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SpanWrapper.java
```diff
@@ -158,4 +158,57 @@ public int getTotalRecordedLinks() {
   public int getTotalAttributeCount() {
     return totalAttributeCount();
   }
+
+  @Override
+  public final String toString() {
```
jkwatson, line 163: final here? Seems unnecessarily restrictive.

jkwatson: oh, does this make autoValue not try to implement it? I forgot about that. nm

kenfinnigan (approved):

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
bogdandrutu: I don't want users to call `toString` on the `Span` and parse the String to extract fields, they should do that only in the processor via the `ReadableSpan`.

anuraaga: `toString` is important in debugging, e.g., it's presented well by the IDE. We should definitely not be removing `toString`'s we already have. While it'd be unfortunate if anyone parses the tostring, that's a bit too explicit foot-shooting to lose the debugging experience for normal cases.

jkwatson: Agree with @anuraaga . Let's not get rid of valuable debugging information, even if a user might do something unsupported. They can also use reflection if they want to do bad things to themselves.

bogdandrutu: I will move this to a separate PR and still fight it. I am not sure what debugging scenario you have in mind

bogdandrutu: Reverted.

bogdandrutu:
For some reasons this PR does not get updated, if you look at my branch https://github.com/open-telemetry/opentelemetry-java/compare/main...bogdandrutu:fixsdtostring the removing of the toString was reverted, I will give few more minutes to see if github picks this up

jkwatson (approved):

carlosalberto (approved):

PR #2673, opened 2021-02-02
Update the check in the tracer to see if the SDK has been shut down.

jkwatson:
This changes the logic so that we only need to have called shutdown on the TracerSharedState, not that it has successfully shut down. Otherwise, a failed shutdown would continue to produce spans when it shouldn't be.

### DIFF
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/TracerSharedState.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/TracerSharedState.java
@@ -81,18 +81,20 @@ final class TracerSharedState {
   }
 
   /**
-   * Returns {@code true} if tracing is stopped.
+   * Returns {@code true} if tracing has been shut down.
    *
-   * @return {@code true} if tracing is stopped.
+   * @return {@code true} if tracing has been shut down.
    */
-  boolean isStopped() {
+  boolean hasBeenShutdown() {
+    // todo: do we really need a lock around this check? This is on the hot path of tracing.
     synchronized (lock) {
-      return shutdownResult != null && shutdownResult.isSuccess();
+      return shutdownResult != null;
     }
   }
 
   /**
-   * Stops tracing, including shutting down processors and set to {@code true} {@link #isStopped()}.
+   * Stops tracing, including shutting down processors and set to {@code true} {@link
+   * #hasBeenShutdown()}.
    *
    * @return a {@link CompletableResultCode} that will be completed when the span processor is shut
    *     down.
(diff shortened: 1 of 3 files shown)


bogdandrutu (approved):

carlosalberto (approved):

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/TracerSharedState.java
```diff
@@ -81,18 +81,20 @@ SpanProcessor getActiveSpanProcessor() {
...
   /**
-   * Returns {@code true} if tracing is stopped.
+   * Returns {@code true} if tracing has been shut down.
    *
-   * @return {@code true} if tracing is stopped.
+   * @return {@code true} if tracing has been shut down.
    */
-  boolean isStopped() {
+  boolean hasBeenShutdown() {
+    // todo: do we really need a lock around this check? This is on the hot path of tracing.
```
anuraaga, line 89: We needed it when we had two checks, now we only have one - so yeah let's delete this lock now

bogdandrutu: Suggest separate PR since that is an internal implementation detail, this is a behavior change

bogdandrutu: Point being this is urgent (before 1.0) the lock improvement can wait :)

anuraaga: - Our 1.0 is about API stability, we generally wouldn't consider this PR as required-for-ga anyways, and definitely not required-for-next-rc

- It's a two line change that is a direct result of this PR, separating doesn't make sense it just decouples the cause/effect

jkwatson: @breedx-splk and I have been working separately on getting rid of the lock. The question we have is... how important is it that calls to `shutdown()` are idempotent? That is, how much safety do we need to provide around not calling shutdown more than once on the processors/exporters?

anuraaga: The lock on write, or a AtomicReference CAS seems important, but for this read we're just reading a volatile variable. Perhaps that's the confusion - it seems the variable was incorrectly marked volatile despite being guarded :) It's why I thought it's a no-brainer to remove the lock here.

jkwatson: ok, done. It's still a volatile read in the hot path, but I guess that's probably better than a volatile read, guarded by a lock on the hot path. :)

anuraaga (approved):

bogdandrutu (approved):

PR #2674, opened 2021-02-02
[BREAKING CHANGE] Change Baggage.forEach to map other forEach methods, same generics asMap

bogdandrutu:
* Removes one custom interface and uses BiConsumer as all the other places.
* Builder no longer needs to create new objects from parent and can reshare them.


jkwatson:
I think this is a good change, but I will not merge until we've all agreed this is an ok API breaking change. The other option would be to make this a purely additive API, since the method signatures wouldn't clash.

extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/OtTracerPropagator.java
```diff
@@ -73,11 +73,10 @@ public static OtTracerPropagator getInstance() {
...
     Baggage baggage = Baggage.fromContext(context);
     if (!baggage.isEmpty()) {
+      // Metadata is not supported by OpenTracing
       baggage.forEach(
-          (key, value, metadata) -> {
-            // Metadata is not supported by OpenTracing
-            setter.set(carrier, PREFIX_BAGGAGE_HEADER + key, value);
-          });
+          (key, baggageEntry) ->
+              setter.set(carrier, PREFIX_BAGGAGE_HEADER + key, baggageEntry.getValue()));
```
anuraaga, line 79: This seems to generally make iterating more tedious by requiring `getValue` - is it better for users? I think the talk about the generic interface is more theoretical and these actual usages are what would drive wanting to make a change.

bogdandrutu: Without this you can look at the Builder and we force an extra allocation. I think just for that and it is still worth it.

anuraaga: The builder's an implementation detail though - we can fix the allocation. So it's still more important to look at the iteration pattern itself - the previous seemed easier. It's not way easier though, so I wouldn't block a change but am not convinced it's much better this way either.

bogdandrutu: For me this is a clear improvement:
* Reduces allocation (that can be done in other ways I agree)
* Reduces API surface by removing one extra interface (better for auto-instrumentation)
* Improves consistency with other "collection like" classes that we expose
* Downside: One extra call for the user in some cases.

anuraaga: > Downside: One extra call for the user in some cases.

most cases :P I think generally we would prioritize the user experience among those bullets, so for me it's the hard to quantify value of consistency with other collection classes vs making it harder to access the baggage value. I also have an assumption that almost no users will be using metadata, which increases the negative impact of this change a bit. Still not a big deal but it's something so we should be sure.

iNikem: We talked today that users will not actually read baggage that much. They will write it more than read. And I slightly like @bogdandrutu 's change.

anuraaga (approved):
It's a very small different and usage, and it makes sense to stick with more common types like `BiConsumer` :+1:

bogdandrutu:
@jkwatson can I merge this one?

jkwatson (approved):

PR #2678, opened 2021-02-02
Update CompletableResultCode to use java8 CompletableFuture

bogdandrutu:
Question: Should we consider to use CompletableFuture<ResultCode> as the return type for all methods instead of CompletableResultCode?

### DIFF
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/common/CompletableResultCode.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/common/CompletableResultCode.java
@@ -7,13 +7,10 @@ package io.opentelemetry.sdk.common;
 
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.List;
-import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicInteger;
-import javax.annotation.Nullable;
-import javax.annotation.concurrent.GuardedBy;
+import java.util.concurrent.TimeoutException;
 
 /**
  * This class models JDK 8's CompletableFuture to afford migration should Open Telemetry's SDK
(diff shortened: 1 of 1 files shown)


anuraaga:
This is something I hadn't gotten to sorting out. We agreed, to have best-effort support for Android, to follow okhttp's practice with regard to animalsniffer. But we never bumped down from API 24 to API 21 which is what they use, and the latter doesn't have CF. It also doesn't have *Adder which is why I didn't just downgrade it but I was considering using 21 generally and keep metrics at 24 until sorting out what to do about it.

This is the background - while we do need the animalsniffer to keep our backs, but I'd still like to follow our decision on Android and not at least intentionally include CF in our code.

jkwatson:
> This is something I hadn't gotten to sorting out. We agreed, to have best-effort support for Android, to follow okhttp's practice with regard to animalsniffer. But we never bumped down from API 24 to API 21 which is what they use, and the latter doesn't have CF. It also doesn't have *Adder which is why I didn't just downgrade it but I was considering using 21 generally and keep metrics at 24 until sorting out what to do about it.
> 
> This is the background - while we do need the animalsniffer to keep our backs, but I'd still like to follow our decision on Android and not at least intentionally include CF in our code.

I don't remember saying we'd stick with 21. I vaguely remember discussing okhttp and using it as a model, but we obviously never followed-through on that. :(

bogdandrutu:
Couple of lessons from this PR:
* `join` API is equivalent with a get more than a `join`. Probably we should rename that.
* `whenComplete` has a different API, should we fix that?
* `ofAll` is very confusing. CompletableFuture has a `allOf` which does something different than our `ofAll`.

Even if we don't use CompletableFuture maybe worth at least fixing the API.

anuraaga:
> I don't remember saying we'd stick with 21. I vaguely remember discussing okhttp and using it as a model, but we obviously never followed-through on that. :(

Sorry for not following through on the animalsniffer - but we did follow through a little by not using CompletableFuture anywhere :D

> join API is equivalent with a get more than a join. Probably we should rename that.

The key point is no checked exceptions - Java took an unfortunate decision of having a method with checked exceptions + timeout, or a method with no exceptions but no timeout. Timeout is important, and checked exceptions are annoying to users, which is why it looks like this. Not sure if I'd rename to `get` or add checked exceptions instead.

> whenComplete has a different API, should we fix that?

The main reason for the difference is not returning a Throwable, just track success or failure, keep logging of errors a job of the exporters. This is just to reduce complexity of the class but we could make it closer to CompletableFuture. The class has been working pretty well so far though I think so don't know if we need to change it at this point.

jkwatson:
After thinking about pros and cons...I think I'm in the "let's stick with what we have" camp. It works, very few people will even need to worry about it, and I'm ok with maintaining a few lines of code, even if the API is slightly different than the JDK APIs (which I will almost never hold up as being great... `IoException` I'm looking at you)

anuraaga (commented):
I think the change that makes the most sense could be adding `whenComplete(Consumer<Boolean>)`, but that's backwards compatible so we can add later. I wouldn't change join for having the unchecked exception behavior, and `ofAll` is consistent with `ofSuccess` and `ofFailure`, I think this consistency is more important than matching `CompletableFuture`, and these names seem OK.

PR #2679, opened 2021-02-02
Rename BaggageEntryMetadata to BaggageMetadata

bogdandrutu:


anuraaga (commented):
It's the metadata for a baggage entry, not the whole baggage, so I think current name is better - what happens if spec adds metadata to the overall baggage?

bogdandrutu:
Probably is the `Entry` part that is annoying, we should still discuss separately if the Item is better or suitcase :))

jkwatson:
Yeah... it is metadata associated with the entry, not the baggage as a whole. So, I vote we stick with what we have now; the gain is very little at this point.

PR #2682, opened 2021-02-03
[BREAKING CHANGE] Simplify bytesFromHex by removing offset, consisten…

bogdandrutu:
…t isValid


anuraaga (approved):

carlosalberto (approved):

carlosalberto:
(We can add an overload with an offset later on _if_ needed ;) )

bogdandrutu:
@carlosalberto for sure, but for the moment it seems that is not needed

bogdandrutu:
Also the CharSequence is lightweight so it can be constructed with a different offset (no more extra copy anyway)

jkwatson (approved):

PR #2683, opened 2021-02-03
Downcast to reduce allocations when copying Baggage

anuraaga:


jkwatson (approved):
seems fine to me. should save some allocations in the most common case

Issue #2666, opened 2021-02-02
Cleanup TraceId and SpanId APIs


bogdandrutu:
After the refactoring when TraceId and SpanId were embedded in the SpanContext, these classes became utils/helpers. Since then the names of the methods are no longer consistent and some of them are unnecessary:
* ~~copyHexInto - not used outside the TraceID class~~
* fromLong[s] - return the hex representation. We should follow the same conventions as for the bytes/hex helpers and name these hexFromLong[s]
* SpanId.asLong accepts `CharSequence` but all other methods related to the hex representation accept `String`.
* traceIdHighBytesAsLong and traceIdLowBytesAsLong suggest that they convert from the bytes (probably I misunderstand this) but they actually accept hex version. And again inconsistent being the only once with `CharSequence`
* bytesFromHex/bytesToHex - are not necessary if in the SpanData we expose the SpanContext which will allow users to retrieve the ids in hex or bytes.

jkwatson:
I'm good with doing these cleanups. These APIs are probably not used by anyone outside of the project itself at this point, so a last cleanup here seems like a good idea.

PR #2684, opened 2021-02-03
[BREAKING CHANGE] Remove AsHexString suffix from ids getters.

bogdandrutu:
Updates https://github.com/open-telemetry/opentelemetry-java/issues/2666

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/trace/SpanContext.java
+++ b/api/all/src/main/java/io/opentelemetry/api/trace/SpanContext.java
@@ -65,33 +65,33 @@ public interface SpanContext {
   }
 
   /**
-   * Returns the trace identifier associated with this {@code SpanContext}.
+   * Returns the trace identifier associated with this {@link SpanContext} as lowercase hex.
    *
-   * @return the trace identifier associated with this {@code SpanContext}.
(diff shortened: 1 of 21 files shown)


api/all/src/main/java/io/opentelemetry/api/trace/SpanContext.java
```diff
@@ -65,33 +65,33 @@ static SpanContext createFromRemoteParent(
...
 
   /**
-   * Returns the trace identifier associated with this {@code SpanContext}.
+   * Returns the trace identifier associated with this {@link SpanContext} as lowercase hex.
    *
-   * @return the trace identifier associated with this {@code SpanContext}.
+   * @return the trace identifier associated with this {@link SpanContext} as lowercase hex.
    */
-  String getTraceIdAsHexString();
+  String getTraceIdHex();
```
anuraaga, line 72: Let's simplify even more :) `getTraceId()`. This means that we always refer to Trace ID, or Span ID, as the hex version by default. When we have a suffix like bytes, then it's the bytes version.

bogdandrutu: Done. Will followup to update TraceId/SpanId classes to reflect this change.

jkwatson (approved):

PR #2694, opened 2021-02-03
Add javadoc to the package-info.java for Sampling

### LINKED ISSUE #2632: Update javadoc for ParentBased methods
Currently there are 2 methods to create ParentBased samplers, one that directly creates the Sampler without allowing to set other options and one that returns the builder.

I would suggest to remove the optimization method that does not return the builder. It is an extra public method to maintain and returning a builder in Java world is probably one of the most common patter so nothing to worry.

### LINKED ISSUE #2644: Document samplers package
Right now we have in the same package as the main SDK components:
* IdsGenerator
* SpanProcessor

In a `config` package:
* Configs only for span limits

In a `sampler` package:
* Sampler interface and few implementations

I think `data` and `export` are very clear what they are and those don't need to change.

Questions:
* Should we move Sampler and Config to the main SDK components package for consistency?
* If not, why do we not move IdsGenerator separately?

jkwatson:
and update some javadoc for parent-based Samplers.

resolves #2632 
resolves #2644

### DIFF
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/Sampler.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/Sampler.java
@@ -59,9 +59,9 @@ public interface Sampler {
 
   /**
    * Returns a {@link ParentBasedSamplerBuilder} that enables configuration of the parent-based
-   * sampling strategy. The parent's sampling decision if one exists, otherwise following the root
-   * sampler and other optional sampler's decision. There are a several options available on the
-   * builder to configure the precise behavior with respect to how the decision will be made.
+   * sampling strategy. The parent's sampling decision is used if a parent span exists, otherwise
+   * this strategy uses the root sampler's decision. There are a several options available on the
+   * builder to control the precise behavior of how the decision will be made.
    *
    * @param root the required {@code Sampler} which is used to make the sampling decisions if the
    *     parent does not exist.
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/Sampler.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/Sampler.java
@@ -15,7 +15,7 @@ import io.opentelemetry.sdk.trace.data.LinkData;
 import java.util.List;
 import javax.annotation.concurrent.ThreadSafe;
 
-/** Sampler is used to make decisions on {@link Span} sampling. */
+/** A Sampler is used to make decisions on {@link Span} sampling. */
 @ThreadSafe
 public interface Sampler {
 
@@ -46,6 +46,8 @@ public interface Sampler {
    * whether or not to sample. If there is no parent, the Sampler uses the provided Sampler delegate
    * to determine the sampling decision.
    *
+   * <p>This method is equivalent to calling {@code #parentBasedBuilder(Sampler).build()}
+   *
    * @param root the {@code Sampler} which is used to make the sampling decisions if the parent does
    *     not exist.
    * @return a {@code Sampler} that follows the parent's sampling decision if one exists, otherwise
@@ -56,12 +58,14 @@ public interface Sampler {
   }
 
   /**
-   * Returns a {@link ParentBasedSamplerBuilder} that follows the parent's sampling decision if one
-   * exists, otherwise following the root sampler and other optional sampler's decision.
+   * Returns a {@link ParentBasedSamplerBuilder} that enables configuration of the parent-based
+   * sampling strategy. The parent's sampling decision if one exists, otherwise following the root
+   * sampler and other optional sampler's decision. There are a several options available on the
+   * builder to configure the precise behavior with respect to how the decision will be made.
    *
    * @param root the required {@code Sampler} which is used to make the sampling decisions if the
    *     parent does not exist.
-   * @return a {@code ParentBased.Builder}
+   * @return a {@code ParentBasedSamplerBuilder}
    */
   static ParentBasedSamplerBuilder parentBasedBuilder(Sampler root) {
     return new ParentBasedSamplerBuilder(root);
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/package-info.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/package-info.java
@@ -3,7 +3,31 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-/** Samplers for selecting traces that are recorded and exported. */
+/**
+ * This package contains {@link io.opentelemetry.sdk.trace.samplers.Sampler}s for selecting traces
+ * that are recorded and exported. <br>
+ *
+ * <p>Sampling is a mechanism to control the noise and overhead introduced by OpenTelemetry by
+ * reducing the number of samples of traces collected and sent to the backend. <br>
+ *
+ * <p>The following sampling strategies are provided here:
+ *
+ * <p>{@link io.opentelemetry.sdk.trace.samplers.Sampler#alwaysOff()} : This strategy will ensure
+ * that no Spans are ever sent to the export pipeline.
+ *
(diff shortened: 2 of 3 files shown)


bogdandrutu (approved):
This PR looks good. I just want to point out that there is a small strange thing that we expose a Builder only class for `ParentBased` maybe an option is to provide a factory method that accepts only root and one that accepts all the values? Not sure but wanted to write this so maybe someone has an idea :)

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/package-info.java
```diff
@@ -3,7 +3,31 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-/** Samplers for selecting traces that are recorded and exported. */
+/**
+ * This package contains {@link io.opentelemetry.sdk.trace.samplers.Sampler}s for selecting traces
```
anuraaga, line 7: Does a package-info have to fully qualify the names in the package?

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/package-info.java
```diff
@@ -3,7 +3,31 @@
...
+ * that no Spans are ever sent to the export pipeline.
+ *
+ * <p>{@link io.opentelemetry.sdk.trace.samplers.Sampler#alwaysOn()} : This strategy will ensure
+ * that every Span will be sent to the export pipeline.
+ *
+ * <p>{@link io.opentelemetry.sdk.trace.samplers.Sampler#traceIdRatioBased(double)} : This strategy
+ * will sample the provided fraction of Spans, deterministically based on the TraceId of the Spans.
+ * This means that all spans from the a given trace will have the same sampling result.
+ *
+ * <p>{@code Sampler#parentBased(Sampler)} : This strategy will always use the sampled state of the
```
anuraaga, line 25:
```suggestion
 * <p>{@link Sampler#parentBased(Sampler)} : This strategy will always use the sampled state of the
```

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/package-info.java
```diff
@@ -3,7 +3,31 @@
...
+ * parent span when deciding whether to sample a Span or not. If the the Span has no parent, the
```
anuraaga, line 26:
```suggestion
 * parent span when deciding whether to sample a Span or not. If the Span has no parent, the
```

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/package-info.java
```diff
@@ -3,7 +3,31 @@
...
+ * provided "root" Sampler will be used for that decision. The parent-based strategy is highly
+ * configurable, using the {@link io.opentelemetry.sdk.trace.samplers.ParentBasedSamplerBuilder}
+ * which can be acquired from the {@code Sampler#parentBasedBuilder(Sampler)} method.
```
anuraaga, line 29:
```suggestion
 * which can be acquired from the {@link Sampler#parentBasedBuilder(Sampler)} method.
```

anuraaga (approved):

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/package-info.java
jkwatson: yeah. it appears that it does. :(

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/package-info.java
jkwatson: unfortunately, then you have to fully qualify it and it fails the formatting check. Any idea how to turn off the line-length checks for package-info files?

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/package-info.java
jkwatson: see above for my sad tale of woe

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/package-info.java
anuraaga: Yeah, added a commit here to disable checkstyle LineLength for package-info.java specifically, I think it's safe since it's just documentation.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/Sampler.java
```diff
@@ -56,12 +58,14 @@ static Sampler parentBased(Sampler root) {
   }
 
   /**
-   * Returns a {@link ParentBasedSamplerBuilder} that follows the parent's sampling decision if one
-   * exists, otherwise following the root sampler and other optional sampler's decision.
+   * Returns a {@link ParentBasedSamplerBuilder} that enables configuration of the parent-based
+   * sampling strategy. The parent's sampling decision if one exists, otherwise following the root
```
iNikem, line 62: > The parent's sampling decision if one exists

It seems like some verb is missing?

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/package-info.java
```diff
@@ -3,7 +3,35 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-/** Samplers for selecting traces that are recorded and exported. */
+/**
+ * This package contains {@link io.opentelemetry.sdk.trace.samplers.Sampler}s for selecting traces
+ * that are recorded and exported. <br>
+ *
+ * <p>Sampling is a mechanism to control the noise and overhead introduced by OpenTelemetry by
```
iNikem, line 10: Should we reference specification here?

jkwatson: Probably; I copied this line directly out of the spec. I'm slightly worried about javadoc referencing potentially mobile spec docs, but it might be fine.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/Sampler.java
jkwatson: I re-worked the language in here a bit.

jkwatson: sorry...I *will* rework the language a bit.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/package-info.java
jkwatson: check it out. how stable do we think that URL is going to be?

bogdandrutu: you can get a permanent link that points to a specific commit.

jkwatson: That's an idea. Will that be something that we remember to update when the spec gets versioned?

bogdandrutu: :)) everything has it's own problems

jkwatson: switched to the latest SHA for now.

iNikem (approved):

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/Sampler.java
jkwatson: and now I *have* reworked the language.

PR #2698, opened 2021-02-04
[FEEDBACK REQUIRED] Change TraceFlags to follow the ids and be hex by default

bogdandrutu:
This PR changes how SpanContext exposes TraceFlags by default - from byte to hex representation. An alternative PR will be created to show how a cleanup could be made without changing the default representation.


jkwatson:
I think I prefer this option, but not by a large margin.

PR #2703, opened 2021-02-04
[FEEDBACK REQUIRED] Cleanup TraceFlags, make it consistent with ids, but use byte as default

bogdandrutu:
This uses 2 characters to represent the hex TraceFlags.

Alternative to https://github.com/open-telemetry/opentelemetry-java/pull/2698 that keeps the default representation as byte for TraceFlags.

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/trace/TraceFlags.java
+++ b/api/all/src/main/java/io/opentelemetry/api/trace/TraceFlags.java
@@ -8,47 +8,104 @@ package io.opentelemetry.api.trace;
 import javax.annotation.concurrent.Immutable;
 
 /**
- * Helper methods for dealing with trace flags options. These options are propagated to all child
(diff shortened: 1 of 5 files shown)


api/all/src/main/java/io/opentelemetry/api/trace/TraceFlags.java
```diff
@@ -8,47 +8,104 @@
...
+  }
+
+  /**
+   * Returns the second lowercase hex (base16) character of the {@code TraceFlags} converted from
+   * the given hex (base16) representation.
+   *
+   * @param traceFlags the byte representation of the {@code TraceFlags}.
+   * @throws IllegalArgumentException if not enough characters in the {@code dest}.
+   */
+  public static char secondHexCharacter(byte traceFlags) {
```
bogdandrutu, line 105: Better names can be provided as `secondCharacterAsHex` or `lowPartAsHex`, similar for the `firstHexCharacter`

PR #2704, opened 2021-02-04
Replace DefaultOpenTelemetry builder / class exposure with a factory …

### LINKED ISSUE #2629: Consider to not expose DefaultOpenTelemetry class
Right now this class does not have a clear use for being public. Because of this java API should not publicly expose this class and it should be package protected until we get a clear use-case to expose it publicly.

anuraaga:
…for configuring propagation only.

Fixes #2629


api/all/src/main/java/io/opentelemetry/api/OpenTelemetry.java
```diff
@@ -21,6 +21,20 @@
  * @see ContextPropagators
  */
 public interface OpenTelemetry {
+  /** Returns a default, completely no-op {@link OpenTelemetry}. */
+  static OpenTelemetry getDefault() {
```
anuraaga, line 25: Wondering if we should rename all these API `getDefault` to `noop` which is more obvious, and consistent with what ContextPropagators is right now.

anuraaga: Though I remember there was one concern that the default tracer returns noop spans, but is itself still an operation. But I don't know if that's really an important distinction.

trask: I like noop. Default sounds like a default "functioning" implementation to me.

api/all/src/main/java/io/opentelemetry/api/DefaultOpenTelemetry.java
```diff
@@ -15,26 +15,25 @@
  */
 @ThreadSafe
 public final class DefaultOpenTelemetry implements OpenTelemetry {
```
bogdandrutu, line 17: I think we can hide this class as well correct?

api/all/src/main/java/io/opentelemetry/api/OpenTelemetry.java
bogdandrutu: I like "Noop", only concern is we should probably do the same everywhere...

api/all/src/main/java/io/opentelemetry/api/OpenTelemetry.java
```diff
@@ -21,6 +21,20 @@
...
+    return DefaultOpenTelemetry.noop();
+  }
+
+  /**
+   * Returns an {@link OpenTelemetry} which will do remote propagation of {@link
+   * io.opentelemetry.context.Context} using the provided {@link ContextPropagators} and is no-op
+   * otherwise.
+   */
+  static OpenTelemetry getPropagating(ContextPropagators propagators) {
```
bogdandrutu, line 34: `getPropagationOnly`?

api/all/src/main/java/io/opentelemetry/api/DefaultOpenTelemetry.java
anuraaga: Oops forgot, yeah

api/all/src/main/java/io/opentelemetry/api/OpenTelemetry.java
anuraaga: It's a bit obtuse - propagating looks great next to noop I think since it's just the version that adds propagation so think the only is naturally implied. Compared to default, though, maybe not as much

api/all/src/main/java/io/opentelemetry/api/OpenTelemetry.java
jkwatson: "Noop" is more fun to say, and certainly more precise, but I think at this point we shouldn't change it everywhere, so I think we should stick with 'default'.

api/all/src/main/java/io/opentelemetry/api/OpenTelemetry.java
jkwatson: There is one corner case where it's a bit misleading to say "only"...if the auto-instrumentation agent instruments the API, you'll get full functionality here. Actually...in that case will the propagators provided here be ignored and replaced with the agent's internal configuration?

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
```diff
@@ -78,7 +75,7 @@ public static OpenTelemetry get() {
    * <p>If you are using the OpenTelemetry SDK, you should generally use {@code
    * OpenTelemetrySdk.builder().buildAndRegisterGlobal()} instead of calling this method directly.
    */
-  public static void set(OpenTelemetry openTelemetry) {
+  public static OpenTelemetry set(OpenTelemetry openTelemetry) {
```
jkwatson, line 78: should we add javadoc about the `@return` on here now? This API is slightly strange to simply return exactly what is passed in.

api/all/src/main/java/io/opentelemetry/api/DefaultOpenTelemetry.java
bogdandrutu: Can you fix this?

api/all/src/main/java/io/opentelemetry/api/OpenTelemetry.java
bogdandrutu: Are you? Why not via the Global?

jkwatson: If you use the Global, the agent also instruments it and uses it's own internal SDK instance.

anuraaga: @jkwatson Yeah you mentioned the agent overriding the API trick - do you know what version it was? Will confirm, but it should have changed to only override what's returned by `GlobalOpenTelemetry.get()`, will confirm. That's the behavior we want anyways

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
anuraaga: Doh, sorry I thought this was a private method so tried to save a line of code. Let me revert this

api/all/src/main/java/io/opentelemetry/api/OpenTelemetry.java
jkwatson: I don't remember what version it was. probably 13 or 14

anuraaga:
Updated, didn't rename anything

jkwatson (approved):

PR #2705, opened 2021-02-04
Return Tracer from DefaultTracer.getInstance

### LINKED ISSUE #2658: Should we return the interface or the object in static methods like `getInstance`?

anuraaga:
There don't seem to be anymore instances where public class has method `getInstance` and returns interface. I found this one location where non-public class returns non-public class instead of interface, which is not consistent with our other such cases.

Fixes #2658

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/trace/DefaultTracer.java
+++ b/api/all/src/main/java/io/opentelemetry/api/trace/DefaultTracer.java
@@ -20,7 +20,7 @@ final class DefaultTracer implements Tracer {
 
   private static final DefaultTracer INSTANCE = new DefaultTracer();
 
-  static DefaultTracer getInstance() {
+  static Tracer getInstance() {
     return INSTANCE;
   }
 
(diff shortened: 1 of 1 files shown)


api/all/src/main/java/io/opentelemetry/api/trace/DefaultTracer.java
```diff
@@ -20,7 +20,7 @@
 
   private static final DefaultTracer INSTANCE = new DefaultTracer();
```
Oberon00, line 21:
```suggestion
  private static final Tracer INSTANCE = new DefaultTracer();
```

bogdandrutu (approved):

bogdandrutu (approved):

PR #2706, opened 2021-02-04
[FEEDBACK REQUIRED] Traceflagsref precompute

anuraaga:
Draft of the 2' with the precomputed TraceFlags. Makes sure there aren't any allocations


api/all/src/main/java/io/opentelemetry/api/trace/ImmutableSpanContext.java
```diff
@@ -26,9 +26,13 @@ static SpanContext getInvalid() {
...
   static SpanContext create(
-      String traceIdHex, String spanIdHex, byte traceFlags, TraceState traceState, boolean remote) {
+      String traceIdHex,
+      String spanIdHex,
+      String traceFlags,
+      TraceState traceState,
+      boolean remote) {
     return new AutoValue_ImmutableSpanContext(
-        traceIdHex, spanIdHex, traceFlags, traceState, remote);
+        traceIdHex, spanIdHex, traceState, remote, TraceFlags.parsedFromHex(traceFlags));
```
bogdandrutu, line 35: The problem with this is that `traceFlags` may be invalid (not 2 hex characters). As a result of this `TraceFlags.parsedFromHex` may throw exceptions (and assuming we fix that) or may produce an invalid params. What do we do in that case? This is the biggest unknown for me with this solution.

PR #2709, opened 2021-02-04
Change TraceFlags to be a class, expose all helpers on the class itself.

bogdandrutu:


jkwatson:
I like this option. Having a class to wrap up the functionality is a good thing, and in this case, since we don't allocate anything extra per request, it's a very good option. 

👍

api/all/src/main/java/io/opentelemetry/api/trace/TraceFlags.java
```diff
@@ -8,47 +8,128 @@
...
+   *
+   * @return the lowercase hex (base16) representation of the {@code TraceFlags} with the sampling
+   *     flag bit on.
+   */
+  public static TraceFlags getSampled() {
+    return SAMPLED;
+  }
+
+  /** Extract the lowercase hex (base16) representation of the {@code TraceFlags} from a buffer. */
+  public static TraceFlags fromBuffer(CharSequence src, int srcOffset) {
```
bogdandrutu, line 61: Should this be `fromHex`?

api/all/src/main/java/io/opentelemetry/api/trace/TraceFlags.java
```diff
@@ -8,47 +8,128 @@
...
- * {@link Span spans}. These determine features such as whether a {@code Span} should be traced. It
- * is implemented as a bitmask.
+ * Helper methods for dealing with trace flags options. A valid trace flags is a 2 character
+ * lowercase hex (base16) String.
+ *
+ * <p>These options are propagated to all child {@link Span spans}. These determine features such as
+ * whether a {@code Span} should be traced. It is implemented as a bitmask.
  */
 @Immutable
 public final class TraceFlags {
```
anuraaga, line 18: @trask Will this prevent the interface bridging of `SpanContext` being a class? I guess so since we wouldn't be able to use the covariant return type.

@bogdandrutu Can we make this an interface and move everything into `ImmutableTraceFlags` class?

bogdandrutu: Don't see a problem with that

bogdandrutu: Can I do it in the next PR to limit the size? Or I can do it here if you really need it atomically.

anuraaga: Nah follow up is fine

api/all/src/main/java/io/opentelemetry/api/trace/SpanContext.java
```diff
@@ -36,14 +36,14 @@ static SpanContext getInvalid() {
   /**
    * Creates a new {@code SpanContext} with the given identifiers and options.
    *
-   * @param traceIdHex the trace identifier of the span context.
-   * @param spanIdHex the span identifier of the span context.
-   * @param traceFlags the byte representation of the {@link TraceFlags}
-   * @param traceState the trace state for the span context.
+   * @param traceIdHex the trace identifier of the {@code SpanContext}.
```
anuraaga, line 39: `{@link` everywhere in this PR

api/all/src/main/java/io/opentelemetry/api/trace/TraceFlags.java
```diff
@@ -8,47 +8,128 @@
...
 
 /**
- * Helper methods for dealing with trace flags options. These options are propagated to all child
- * {@link Span spans}. These determine features such as whether a {@code Span} should be traced. It
- * is implemented as a bitmask.
+ * Helper methods for dealing with trace flags options. A valid trace flags is a 2 character
+ * lowercase hex (base16) String.
+ *
+ * <p>These options are propagated to all child {@link Span spans}. These determine features such as
+ * whether a {@code Span} should be traced. It is implemented as a bitmask.
```
anuraaga, line 15:
```suggestion
 * whether a {@code Span} should be traced.
```

Don't think we need implementation detail here (it wasn't so clear to me either way)

anuraaga (approved):
I think all the `@code` can and should be `@link` but otherwise LGTM

carlosalberto (approved):

jkwatson (approved):
There are a few lingering outstanding questions/comments on this. This is the last breaking change before RC2, so let's get it over the finish line, including the conversion to an interface in a follow-up PR.

api/all/src/main/java/io/opentelemetry/api/trace/SpanContext.java
bogdandrutu: Done.

api/all/src/main/java/io/opentelemetry/api/trace/TraceFlags.java
bogdandrutu: Done.

PR #2716, opened 2021-02-05
Make TraceFlags interface to allow agent to re-implement it

bogdandrutu:

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/trace/SpanContext.java
+++ b/api/all/src/main/java/io/opentelemetry/api/trace/SpanContext.java
@@ -30,7 +30,7 @@ public interface SpanContext {
    * @return the invalid {@code SpanContext}.
    */
   static SpanContext getInvalid() {
-    return ImmutableSpanContext.getInvalid();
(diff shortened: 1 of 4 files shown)


carlosalberto (approved):

api/all/src/main/java/io/opentelemetry/api/trace/SpanContext.java
```diff
@@ -30,7 +30,7 @@
    * @return the invalid {@code SpanContext}.
    */
   static SpanContext getInvalid() {
-    return ImmutableSpanContext.getInvalid();
+    return ImmutableSpanContext.INVALID;
```
jkwatson, line 33: looks like this change snuck in here. I'm ok with it since it's not changing any APIs.

jkwatson (approved):
👍

api/all/src/main/java/io/opentelemetry/api/trace/SpanContext.java
bogdandrutu: I made it for consistency with accessing the variables directly as I did in this PR

Issue #2719, opened 2021-02-05
SpanContext does not respect the specification contract


bogdandrutu:
Right now the `SpanContext` API says:

```java
  /**
   * Returns the trace identifier associated with this {@link SpanContext} as 32 character lowercase
   * hex String.
   *
   * @return the trace identifier associated with this {@link SpanContext} as lowercase hex.
   */
  String getTraceId();

  /**
   * Returns the span identifier associated with this {@link SpanContext} as 16 character lowercase
   * hex String.
   *
   * @return the span identifier associated with this {@link SpanContext} as 16 character lowercase
   *     hex (base16) String.
   */
  String getSpanId();
```

This is the contract that we offer to the users, which we do not enforce in the public constructor:

```java
SpanContext spanContext = SpanContext.create("foo", "bar", TraceFlags.getDefault(), TraceState.getDefault());
System.out.println(spanContext.getTraceId());  // THIS WILL NOT PRINT 32 lowercase hex.
System.out.println(spanContext.getTraceIdBytes());  // this will throw an exception
```

There may be callers that rely on the returned string to be `32 character lowercase hex String` and we cannot break them, so we need to validate the input received in the `create` method: either during creation or just before returning the value but we MUST validate the input.

jkwatson:
let's try to get this tightened up by 1.0.0, if at all possible.

jkwatson:
When we do this, we can remove the extra validation in our propagators, so we don't double-validate.

bogdandrutu:
@jkwatson only if we do it in factory method otherwise still needs to happen because the specification for propagator is to not change Context if input is invalid

bogdandrutu:
See https://github.com/open-telemetry/opentelemetry-specification/pull/1403/files that actually was crafted when I was looking at this problem :)

jkwatson:
> @jkwatson only if we do it in factory method otherwise still needs to happen because the specification for propagator is to not change Context if input is invalid

Just saying that the propagator can ask the SpanContext if it's valid after creating and choose to return the original, rather than asking about the individual pieces, then asking again after it's been created.

bogdandrutu:
@jkwatson indeed, good point. As long as we do respect the specs there we are good to do that as well :)

PR #2721, opened 2021-02-05
Consistent name for Trace/Span ids getters with the specification

bogdandrutu:

See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#retrieving-the-traceid-and-spanid

Not breaking change since methods were just renamed anyway.

There is one more advantage to this, if we ever want to add back the wrapper classes for Trace/Span id we will be able to have nice getters. Related to https://github.com/open-telemetry/opentelemetry-java/issues/2719 and the fact that we may have to add the wrappers to avoid multiple validations, this does not mean that we will do it that way but leaves the option open.


jkwatson (approved):
Not opposed, since this is already broken, but let's get to stable real quick so we can cut a RC release ASAP.

anuraaga (approved):

PR #2723, opened 2021-02-05
[BREAKING CHANGE] Remove not very used `getTraceState` method from SpanData

bogdandrutu:
It will always be possible to add it back if this becomes a problem for our users.


anuraaga (commented):
@Oberon00 Can you confirm you don't use this? I recall this being added to allow transferring information from propagator to exporter

anuraaga (approved):
Ah nevermind it's just a shortcut

jkwatson (approved):

PR #2728, opened 2021-02-06
Validate the span id and trace id when creating the SpanContext

### LINKED ISSUE #2719: SpanContext does not respect the specification contract

jkwatson:
And, remove that extra validation from propagators.

I didn't remove it from the B3 Propagator(s) yet... that will take a bit more work to untangle.

Resolves #2719

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/trace/ImmutableSpanContext.java
+++ b/api/all/src/main/java/io/opentelemetry/api/trace/ImmutableSpanContext.java
@@ -14,12 +14,18 @@ import javax.annotation.concurrent.Immutable;
 abstract class ImmutableSpanContext implements SpanContext {
 
   static final SpanContext INVALID =
-      create(
+      createInternal(
           TraceId.getInvalid(),
           SpanId.getInvalid(),
           TraceFlags.getDefault(),
           TraceState.getDefault(),
           /* remote= */ false);
+
+  private static AutoValue_ImmutableSpanContext createInternal(
+      String traceId, String spanId, TraceFlags traceFlags, TraceState traceState, boolean remote) {
+    return new AutoValue_ImmutableSpanContext(
+        traceId, spanId, traceFlags, traceState, /* remote$= */ remote);
+  }
 
   static SpanContext create(
       String traceIdHex,
@@ -27,8 +33,11 @@ abstract class ImmutableSpanContext implements SpanContext {
       TraceFlags traceFlags,
       TraceState traceState,
       boolean remote) {
-    return new AutoValue_ImmutableSpanContext(
-        traceIdHex, spanIdHex, traceFlags, traceState, remote);
+    if (SpanId.isValid(spanIdHex) && TraceId.isValid(traceIdHex)) {
+      return createInternal(traceIdHex, spanIdHex, traceFlags, traceState, remote);
+    }
+    return createInternal(
+        TraceId.getInvalid(), SpanId.getInvalid(), traceFlags, traceState, remote);
   }
 
   @Override
--- a/api/all/src/main/java/io/opentelemetry/api/trace/propagation/W3CTraceContextPropagator.java
+++ b/api/all/src/main/java/io/opentelemetry/api/trace/propagation/W3CTraceContextPropagator.java
@@ -214,12 +214,10 @@ public final class W3CTraceContextPropagator implements TextMapPropagator {
       String traceId =
           traceparent.substring(TRACE_ID_OFFSET, TRACE_ID_OFFSET + TraceId.getLength());
       String spanId = traceparent.substring(SPAN_ID_OFFSET, SPAN_ID_OFFSET + SpanId.getLength());
-      if (TraceId.isValid(traceId) && SpanId.isValid(spanId)) {
-        TraceFlags traceFlags = TraceFlags.fromHex(traceparent, TRACE_OPTION_OFFSET);
-        return SpanContext.createFromRemoteParent(
-            traceId, spanId, traceFlags, TraceState.getDefault());
-      }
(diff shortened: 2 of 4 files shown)


api/all/src/main/java/io/opentelemetry/api/trace/ImmutableSpanContext.java
```diff
@@ -14,21 +14,30 @@
...
       String spanIdHex,
       TraceFlags traceFlags,
       TraceState traceState,
       boolean remote) {
-    return new AutoValue_ImmutableSpanContext(
-        traceIdHex, spanIdHex, traceFlags, traceState, remote);
+    if (SpanId.isValid(spanIdHex) && TraceId.isValid(traceIdHex)) {
+      return createInternal(traceIdHex, spanIdHex, traceFlags, traceState, remote);
+    }
+    return createInternal(
```
bogdandrutu, line 39: Should we return the INVALID SC?

bogdandrutu: The reason is that it may seem strange to propagate flags and state but not one of the IDs if the other one is invalid

bogdandrutu: So we either drop everything or we keep everything that is valid. I think the current implementation is in the middle and not ideal

api/all/src/main/java/io/opentelemetry/api/trace/ImmutableSpanContext.java
```diff
@@ -14,21 +14,30 @@
...
           TraceId.getInvalid(),
           SpanId.getInvalid(),
           TraceFlags.getDefault(),
           TraceState.getDefault(),
           /* remote= */ false);
 
+  private static AutoValue_ImmutableSpanContext createInternal(
+      String traceId, String spanId, TraceFlags traceFlags, TraceState traceState, boolean remote) {
+    return new AutoValue_ImmutableSpanContext(
+        traceId, spanId, traceFlags, traceState, /* remote$= */ remote);
```
bogdandrutu, line 27: Now that you calculate the isValid it can be a property in the AutoValue so no need to recalculate that or to use memoization.

bogdandrutu, line 27: Consider to remove the comment `/* remote$= */` not needed

api/all/src/main/java/io/opentelemetry/api/trace/ImmutableSpanContext.java
jkwatson: There were tests depending on this behavior. I'll need to figure out what they were trying to accomplish.

bogdandrutu: I think I know why. There are 2 cases:
1. is invalid hex string -  in this case I think we should return SpanContext.getInvalid()
2. is equal with TraceIs.getInvalid() - this is used in tests I guess, probably we should preserve all values.

Not sure if that is the behavior we want, but this is a guess for the problem.

jkwatson: yes...this will take a bit of research that we can create a follow-up issue to deal with, I think.

api/all/src/main/java/io/opentelemetry/api/trace/ImmutableSpanContext.java
jkwatson: excellent idea. let's create an issue to track that optimization.

jkwatson: it always baffles me when errorprone will complain about this, and when it won't. I'm sure I added it because errorprone complained, but there might have been some refactoring after that.

anuraaga (approved):

api/all/src/main/java/io/opentelemetry/api/trace/ImmutableSpanContext.java
jkwatson: Added #2751

api/all/src/main/java/io/opentelemetry/api/trace/ImmutableSpanContext.java
jkwatson: #2752

jkwatson:
Added issues to track the optimization and investigation of the slightly strange half-way-valid usage.

jkwatson:
Added issues to track the optimization and investigation of the slightly strange half-way-valid usage.

PR #2734, opened 2021-02-06
Simplify TraceFlags fromHex, less checks to perform

bogdandrutu:
getLength is no longer needed because fromHex cleary shows how many characters are needed.

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/trace/TraceFlags.java
+++ b/api/all/src/main/java/io/opentelemetry/api/trace/TraceFlags.java
@@ -16,15 +16,6 @@ import javax.annotation.concurrent.Immutable;
 @Immutable
 public interface TraceFlags {
   /**
-   * Returns the length of the lowercase hex (base16) representation of the {@link TraceFlags}.
-   *
-   * @return the length of the lowercase hex (base16) representation of the {@link TraceFlags}.
-   */
-  static int getLength() {
-    return ImmutableTraceFlags.HEX_LENGTH;
-  }
-
-  /**
    * Returns the default (with all flag bits off) byte representation of the {@link TraceFlags}.
    *
    * @return the default (with all flag bits off) byte representation of the {@link TraceFlags}.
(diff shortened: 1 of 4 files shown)


api/all/src/main/java/io/opentelemetry/api/trace/TraceFlags.java
```diff
@@ -45,17 +36,15 @@ static TraceFlags getSampled() {
...
+   * @param second the second hex character.
    * @return the {@link TraceFlags} converted from the given lowercase hex (base16) representation.
-   * @throws NullPointerException if {@code src} is null.
-   * @throws IndexOutOfBoundsException if {@code src} is too short.
-   * @throws IllegalArgumentException if invalid characters in the {@code src}.
+   * @throws IllegalArgumentException if invalid lowercase hex characters.
    */
-  static TraceFlags fromHex(CharSequence src, int srcOffset) {
-    return ImmutableTraceFlags.fromHex(src, srcOffset);
+  static TraceFlags fromHex(char first, char second) {
```
bogdandrutu, line 46: Alternative names "fromHexCharacters" or "fromHexChars" or this is good enough :) ?

api/all/src/main/java/io/opentelemetry/api/trace/TraceFlags.java
```diff
@@ -15,15 +15,6 @@
  */
 @Immutable
 public interface TraceFlags {
-  /**
-   * Returns the length of the lowercase hex (base16) representation of the {@link TraceFlags}.
-   *
-   * @return the length of the lowercase hex (base16) representation of the {@link TraceFlags}.
-   */
-  static int getLength() {
```
jkwatson, line 23: I don't think we should remove this at this point. It's not harmful to have it here, so let's keep it as-is.

api/all/src/main/java/io/opentelemetry/api/trace/TraceFlags.java
jkwatson: I'd very strongly prefer not to make any more breaking API changes at this point. Let's leave it as-is.

PR #2741, opened 2021-02-06
[BREAKING CHANGE] Remove getters for ids in SpanData

bogdandrutu:
As an alternative all getters can be renamed with a `Hex` suffix, but limiting the public API seems better.


jkwatson:
It's too late for this change. Keep it as is.

bogdandrutu:
I realized that the removing may be too aggressive, filed an alternative that I think it is better.

Issue #2724, opened 2021-02-05
Limit the API that we expose where TraceId/SpanId are String's


bogdandrutu:
Discussed offline with @jkwatson and we will try to limit the helper methods and the API that we expose for dealing with TraceId/SpanId which currently are represented as String.

Because of the possible performance implications of the https://github.com/open-telemetry/opentelemetry-java/issues/2719 we will try to leave our door open to add the TraceId/SpanId wrappers back - advantage of them is that once these wrapper are constructed no extra validation is required.

Resolving this issue does not imply reverting back to use wrapper classes for the ids, but:
* Rename accessors for [Trace/Span]Id with a suffix `Hex`
* Remove all (almost all helpers) that transform to from hex string from the public API. For example the long version are easy to remove and use thinks like `Long.parseLong(spanId, 16)`.

bogdandrutu:
I think it is very important to allow in the future to have if needed wrappers again for SpanId/TraceId classes. In order to do that we just need to merge 3 renaming PRs and this issue can be closed:
* #2742
* #2744
* #2740

jkwatson:
I honestly don't think this is very important. We can make it work without further changes.

bogdandrutu:
Can you document the plan here, so we agree on that and I will be fine

jkwatson:
I guess I don't think that we need to make a plan for a hypothetical that will most likely never come to pass. I don't think saving one allocation per span is a big deal, and I don't think we should try to guess at future designs. When we have an actual need to change something, we'll figure out the right way to change it. Until then, we leave things as is, and we let the future take care of itself.

anuraaga:
IIUC the discussion on this issue, there was concern on double validating but we found a pattern to avoid it, right? If so (or even not I guess) I agree we don't have a plan for any big changes and this change has extended far beyond what it seemed like in https://github.com/open-telemetry/opentelemetry-java/pull/2721. We could consider reverting that to have the consistency back, the spec mentions formats but not method names.

bogdandrutu:
What I understand we found so far: With String you have to validate that is hex lowercase every time you construct a SpanContext, indeed we found a solution to not do that twice per SpanContext instance, but we did not find a solution to the constructor validation. So even if this issue will be closed we haven't solved the issue that because we are using String we cannot guarantee that the values are valid (lowercase hex not non-zero).

I would suggest if you revert anything to revert all changes done since 0.15.0 version will simplify everything. I think that is the best you can do.

[...]

PR #2742, opened 2021-02-06
[BREAKING CHANGE] Update SpanData to be consistent with names in SpanContext

bogdandrutu:
This change will facilitate us to add (if decided to) interface for TraceId and SpanId. It does not add instability to the current API since it is only a renaming. 


jkwatson:
I don't agree with this change. It will break every exporter that is maintained externally, very shortly before the 1.0.0 release.

bogdandrutu:
To document the reason: maintainers voted no against this change

PR #2744, opened 2021-02-06
[BREAKING CHANGE] Rename SpanId/TraceId to SpanIdHex/TraceIdHex

bogdandrutu:
This change will facilitate us to add (if decided to) interface for TraceId and SpanId. It does not add instability to the current API since it is only a renaming. 


jkwatson:
I don't think this is a good change. The classes we have now are named in a way that is unsurprising and easy to find and understand. These names are obscure and point at a future design that I don't think is agreed on. I prefer what we have now.

bogdandrutu:
To document the reason: maintainers did not like this change so it was not allowed to be merged :)

Issue #2735, opened 2021-02-06
Hide StrictContextStorage implementation, expose only the flag to configure it


bogdandrutu:
If we do that we can remove the extra shaded dependencies from the Context package and the package becomes way smaller and simpler. I think this will be a huge win especially because we expect the Context package to be used the more.

Advantages like the number of classes we export (for platforms like Android) will be very appreciated by the users.

anuraaga:
It's meant for use in real servers, such as in alpha deployments, too not just tests, so it was moved into here. Debugging context issues using real requests is important especially since users will have varying test coverage. 

Android is fine since it always goes through Proguard.

And we will need to use the weak map in trace package too once we add support for tracking leaked spans, not just scopes.

bogdandrutu:
I think this is an overthinking on our side, and we will suffer because of this. Also right now there is a java property why does the class need to be public?

bogdandrutu:
I am still suggesting to move it back to test until we really have some users asking for this. Premature optimization and overthinking will cause us a lot of pain to maintain all the APIs.

bogdandrutu:
Last comment I promise :). Alternative we can have an extension with this, so at least if something breaks there or anything we can increase that version independently. Still believe `context` and `api` which will be distributed everywhere should be super slim and fewer APIs as possible

anuraaga:
Context debugging is necessary to use context safely - it's not an extension but a required feature. I've even been paged late at night due to context bugs and it's a constant issue that comes up. Debug by flipping a flag, no need to rebuild the app - this is key.

It's not premature optimization but a feature lending from real experience with users that go async before having a proper async framework, which is most of them.

jkwatson:
Yes. This was a concrete request from very experienced users. We need to keep this.

bogdandrutu:
@jkwatson ok I understand that we need it. But in terms of having this as an extension instead of being in the main package? 

Also a bigger problem is the WeakHashMap class which probably we make a bigger mistake exposing as public and also making a promise to keep it as is. How is that related to context?

jkwatson:
No more breaking changes at this point. We'll live with what we have.

bogdandrutu:
Removing the WeakMap (hiding) should not be a breaking change based on the rules of internal package correct? I think you and @anuraaga told me that internal should not be consider part of the API but looks like it is in this case so this is not respecting the your own rule.

bogdandrutu:
[...]

PR #2745, opened 2021-02-06
[Breaking Change] Limit the public api in the Context package

bogdandrutu:
Avoid exposing the public classes for StrictContextStorage, but preserve the functionality as it was protected by a "flag". This will give users the same experience as before and will allow us to move forward by improving the code without having to deal with ABI compatibility.

Part of this PR I had to remove 2 test classes which I will take as an action item to fix that and re-add them after this PR is merged, also I copied the `WeakConcurrentMap` in the jfr extension for the moment until we clarify where this code can leave so that multiple packages can share this.

Summary:
* Copying the `WeakConcurrentMap` and not exposing it publicly should not be considered breaking change, because no external user should depend on that.
* I do not believe that it was expected for users to consume directly `StrictContextStorage`, it was also final so could not be extended. So making this package protected should not be consider breaking change for the users.
* Counted the number of lines (not including tests and propagation) in the context package, we have ~900 lines in StrictContextStorage/AbstractWeakConcurrentMap/WeakConcurrentMap and ~950 all the other classes and interfaces. I think limiting the number of lines that we expose public is a clear win, and we don't break the users.

Fixes https://github.com/open-telemetry/opentelemetry-java/issues/2735


jkwatson:
I don't think this matters very much, as the number of users who will ever care is very very small. So, I would prefer not making further changes this close to 1.0.0.

bogdandrutu:
I think if it is has so little users is not been better to not have it

jkwatson:
I don't want to make changes on short notice, destabilizing what we have right now, that is working just fine.

bogdandrutu:
To document the reason: maintainers voted no against this change

PR #2747, opened 2021-02-07
Revert to version v0.15.0, no improvements are allowed in RC

bogdandrutu:
This reverts everything up to commit `68ee47108a808ad22bfa19618107263ac173e72f`

@open-telemetry/java-maintainers  please approve ASAP


anuraaga:
We had a set of changes we agreed to make on Wednesday - I don't think we have to revert them, but the changes started to drag on, and the issue in the recent changes is that we can't be dragging forever. But I don't think we have to necessarily revert anything - the suggestion of the trace/span id accessor name was just in case you thought it'd be better.

That being said I'm not opposed to reverting the code back to RC to maintain its intent - issues like Span.Kind were sad but still livable. We can't revert the tooling changes though which don't affect the RC. Some changes like adding the word master back (not sure how main is still there though) are clearly wrong.

bogdandrutu:
Let me know up to which commit you feel comfortable to approve the revert

anuraaga:
If reverting anything, my suggestion is either to

https://github.com/open-telemetry/opentelemetry-java/commit/e187c18112f999c06096e6d7faad0766cb477f6f

To restore to before the change of potentially restoring ID wrappers or

https://github.com/open-telemetry/opentelemetry-java/commit/6c30f411f366091d6e3fae8602fcc452aaeba9cf

Which if I read it right is the boundary of our plan from the triage meeting we had.

jkwatson:
I definitely don't think there is need to revert everything. I'll take a look later today to look for a good spot in the history.

jkwatson:
I actually think if we just rename `SpanContext.getSpanIdHex()` back to `SpanContext.getSpanId()` and likewise with the trace id, that we're in a really solid place without any further reverting.

bogdandrutu:
So to revert only that? Btw that makes this library inconsistent with the specs because in the specs it says the preferred name is Hex

bogdandrutu:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#retrieving-the-traceid-and-spanid

bogdandrutu:
So I would prefer not to do that revert since my name will be associated with that change. Not going to block a PR if you and @anuraaga will merge it.

jkwatson:
> So to revert only that? Btw that makes this library inconsistent with the specs because in the specs it says the preferred name is Hex


The specs say nothing about the method names, but only about the form of the results, unless I am mis-reading them.

jkwatson:
But, if you're not comfortable with that name, I'm happy to put in the PR.

PR #2750, opened 2021-02-08
Restore Trace/SpanId.fromBytes

anuraaga:
This reverts commit d6fea3a70af466f1ef358ab3126a667a9607fd5e.

There currently isn't any way to get an ID from OTLP, which is often used in integration tests. Insrumentation repo used these methods and for now I've inlined everything down through BigEndianEncoding, but I think we need to provide this.


jkwatson (approved):
Ok.

carlosalberto (approved):

PR #2753, opened 2021-02-08
Changelog updates, pre 0.16.0 release.

jkwatson:


CHANGELOG.md
```diff
@@ -56,13 +59,22 @@ return an implementation that contains propagators, but is otherwise no-op.
...
+
+- The `OtTracerPropagator` has been renamed to `OtTracePropagator` in the trace-propagators extension module.
 
 ### SDK
 
 #### Breaking Changes
 
 - `TraceConfig` has been renamed to `SpanLimits` and relocated to the `io.opentelemetry.sdk.tracing` package.
 All related method names have been renamed to match.
+- `SpanData.getTraceState()` has been removed. If this needed by an exporter, please let us know why and what your use-case is.
```
Oberon00, line 76: I can report that we need this in Dynatrace. We store some information about which server in a backend cluster a span should be sent to in the tracestate. This information needs to be propagated to child spans, also across processes, so the tracestate is the natural place to store it.

If exporters cannot access tracestate, then it seems to become quite useless.

jkwatson: oh! Thank you so much for looking this over. I'll get it restored before we release 0.16.0.

Oberon00: It's no huge, urgent, deal for us, as we currently use a fork anyway (and will continue for some time due to our need for Java 7), but just wanted to let you know 😃

jkwatson: @Oberon00  FYI...this is still available via `SpanData.getSpanContext().getTraceState()`. Do you need it to be top level as well or can we have just one way to get it?

jkwatson: I'm going to update this PR to make it clear that it's still available via the SpanContext.

Oberon00: Oh, from reading this changelog I thought it was removed without replacement. No, of course I don't need it at top level, as long as it's available at all that's fine.

anuraaga (approved):

Issue #2654, opened 2021-02-01
Document the requirements and guarantees for the internal packages


bogdandrutu:
I would add an extra sentence that this may break at any moment to scare users of using it.

PR #2757, opened 2021-02-08
Remove comment about internal backwards compatibility guarantees

bogdandrutu:

### DIFF
--- a/api/context/src/main/java/io/opentelemetry/context/internal/shaded/WeakConcurrentMap.java
+++ b/api/context/src/main/java/io/opentelemetry/context/internal/shaded/WeakConcurrentMap.java
@@ -39,9 +39,6 @@ import java.util.concurrent.atomic.AtomicLong;
  *
  * <p>This class has been copied as is from
  * https://github.com/raphw/weak-lock-free/blob/ad0e5e0c04d4a31f9485bf12b89afbc9d75473b3/src/main/java/com/blogspot/mydailyjava/weaklockfree/WeakConcurrentMap.java
- * This is used in multiple artifacts in OpenTelemetry and while it is in our internal API,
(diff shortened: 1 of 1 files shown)


api/context/src/main/java/io/opentelemetry/context/internal/shaded/WeakConcurrentMap.java
```diff
@@ -39,9 +39,6 @@
  *
  * <p>This class has been copied as is from
  * https://github.com/raphw/weak-lock-free/blob/ad0e5e0c04d4a31f9485bf12b89afbc9d75473b3/src/main/java/com/blogspot/mydailyjava/weaklockfree/WeakConcurrentMap.java
- * This is used in multiple artifacts in OpenTelemetry and while it is in our internal API,
- * generally backwards compatible changes should not be made to avoid a situation where different
- * versions of OpenTelemetry artifacts become incompatible with each other.
```
bogdandrutu, line 44: Per previous discussion, we don't want to offer this guarantee, we offer compatibility with internal only if using the bom

jkwatson: @anuraaga was this something you were going to address in #2654 ?

bogdandrutu: The reason to remove this particular one is that it is the only place where we have this kind of requirement. I think the policy should be for all packages and at the package level

jkwatson: yep...just making sure there isn't colliding work going on. :)

anuraaga (approved):

PR #2766, opened 2021-02-08
Null label values means label is not present

bogdandrutu:


api/metrics/src/main/java/io/opentelemetry/api/metrics/common/ArrayBackedLabels.java
```diff
@@ -22,7 +22,7 @@ static Labels empty() {
   }
 
   static Labels sortAndFilterToLabels(Object... data) {
-    return new ArrayBackedLabels(sortAndFilter(data, /* filterNullValues= */ false));
+    return new ArrayBackedLabels(sortAndFilter(data, /* filterNullValues= */ true));
```
jkwatson, line 25: I think we might be filtering null values everywhere now...so maybe we can just set true in the method itself?

bogdandrutu: Followup :) don't want to change everything in one shot :)

jkwatson (approved):

PR #2769, opened 2021-02-09
save the already known state of SpanContext validity

### LINKED ISSUE #2752: Optimize the SpanContext validity check
In #2728 we now validate the trace id and span id when the SpanContext is created, so we can optimize the `isValid()` check to store that value, rather than having to calculate it again, and then memoizing the result.

jkwatson:
Resolves #2752


api/all/src/main/java/io/opentelemetry/api/trace/ImmutableSpanContext.java
```diff
@@ -34,15 +39,18 @@ static SpanContext create(
...
+        remote,
+        /* valid= */ false);
   }
 
   @Override
-  @Memoized
-  public boolean isValid() {
-    return SpanContext.super.isValid();
-  }
+  public abstract boolean isValid();
```
jkwatson, line 55: needed here to force auto-value to provide the property.

api/all/src/main/java/io/opentelemetry/api/trace/ImmutableSpanContext.java
```diff
@@ -19,12 +18,18 @@
...
   private static AutoValue_ImmutableSpanContext createInternal(
-      String traceId, String spanId, TraceFlags traceFlags, TraceState traceState, boolean remote) {
+      String traceId,
+      String spanId,
+      TraceFlags traceFlags,
+      TraceState traceState,
+      boolean remote,
+      boolean valid) {
     return new AutoValue_ImmutableSpanContext(
-        traceId, spanId, traceFlags, traceState, /* remote$= */ remote);
```
bogdandrutu, line 27: You did not fix this from my previous comment :D

bogdandrutu (approved):

api/all/src/main/java/io/opentelemetry/api/trace/ImmutableSpanContext.java
jkwatson: I knew I was going to be in here hacking again, so I thought I'd just figure it out here. :)

PR #2770, opened 2021-02-09
Add throws NPE javadoc to Span addEvent methods

trask:
Thanks for pointing these out @samvaity!

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/trace/Span.java
+++ b/api/all/src/main/java/io/opentelemetry/api/trace/Span.java
@@ -219,6 +219,7 @@ public interface Span extends ImplicitContextKeyed {
    * @param attributes the attributes that will be added; these are associated with this event, not
    *     the {@code Span} as for {@code setAttribute()}.
    * @return this.
+   * @throws NullPointerException if {@code attributes} is {@code null}.
    */
   Span addEvent(String name, Attributes attributes);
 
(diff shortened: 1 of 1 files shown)


jkwatson:
Hmm. I wonder if, rather than documenting this, we should fix it so that it doesn't NPE.

trask:
for better or worse, it was documented as NPE in SpanBuilder.addLink():

https://github.com/open-telemetry/opentelemetry-java/blob/7c73ddb158ec39060941be8fafa85240d03c2b70/api/all/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java#L159-L161

bogdandrutu:
@trask we are using the `@Nullable` annotation, so maybe we can make a more general statement that everything that is not marked as `@Nullable` will throw NPE.

anuraaga (approved):
We're not supposed to throw anywhere in the API, so we do need to fix that. We could merge this first though to reflect the reality for now.

jkwatson:
> We're not supposed to throw anywhere in the API, so we do need to fix that. We could merge this first though to reflect the reality for now.

would it be a breaking change to no longer throw documented NPEs, though? If we don't document them, it can be considered a bug we can fix.

anuraaga:
> would it be a breaking change to no longer throw documented NPEs, though? If we don't document them, it can be considered a bug we can fix.

Ah interesting - I think in general, it's not a breaking change even if it's documented, as long as the replacement is a harmless no-op. But keeping it as a bug, rather than a feature, does make sense too.

anuraaga:
Superceded by #2777, thanks for bringing to attention @trask and @samvaity!

PR #2772, opened 2021-02-09
Make StrictContextStorage package-private

### LINKED ISSUE #2735: Hide StrictContextStorage implementation, expose only the flag to configure it

anuraaga:
Fixes #2735

### DIFF
--- a/extensions/kotlin/build.gradle
+++ b/extensions/kotlin/build.gradle
@@ -4,11 +4,16 @@ plugins {
 
     id "me.champeau.gradle.jmh"
     id "org.jetbrains.kotlin.jvm"
+    id "org.unbroken-dome.test-sets"
     id "ru.vyarus.animalsniffer"
 }
 
 description = "OpenTelemetry Kotlin Extensions"
 ext.moduleName = "io.opentelemetry.extension.kotlin"
+
+testSets {
+    testStrictContext
+}
 
 dependencies {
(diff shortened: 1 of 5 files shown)


jkwatson (approved):
👍 could you throw an item into the CHANGELOG for this as well?

anuraaga:
Sure sent #2773

extensions/kotlin/build.gradle
```diff
@@ -29,4 +34,8 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
 }
 
 // We don't have any public Java classes
-tasks.javadoc.enabled = false
\ No newline at end of file
+tasks.javadoc.enabled = false
+
+testStrictContext {
+    jvmArgs("-Dio.opentelemetry.context.enableStrictContext=true")
+}
```
bogdandrutu, line 41: New line?

bogdandrutu (approved):

extensions/kotlin/build.gradle
bogdandrutu: Can be done in a separate PR :)

PR #2774, opened 2021-02-09
Remove testonly code from context

bogdandrutu:
Not a breaking change :)


api/context/src/main/java/io/opentelemetry/context/StrictContextStorage.java
```diff
@@ -247,67 +206,9 @@ public String toString() {
...
-    }
-
-    List<CallerStackTrace> drainPendingCallers() {
-      List<CallerStackTrace> pendingCallers =
-          map.values().stream().filter(caller -> !caller.closed).collect(Collectors.toList());
-      map.clear();
-      return pendingCallers;
-    }
-
-    // Called by cleaner thread.
```
anuraaga, line 287: This is removing the functionality of the `StrictContextStorage` completely, which would be caught if the test wasn't deleted with it. 

Anyways we will need to figure out how to allow instrumentation to call this from tests as well and we don't need to delete this code.

anuraaga (changes requested):

api/context/src/main/java/io/opentelemetry/context/StrictContextStorage.java
bogdandrutu: I am lost in whole this code. How can it happen to have this check without getting an error in one of the other places:
https://github.com/open-telemetry/opentelemetry-java/blob/main/api/context/src/main/java/io/opentelemetry/context/ThreadLocalContextStorage.java#L34

OR 

https://github.com/open-telemetry/opentelemetry-java/blob/main/api/context/src/main/java/io/opentelemetry/context/StrictContextStorage.java#L226

anuraaga: The test for garbage leaked scopes is only in one place

https://github.com/open-telemetry/opentelemetry-java/blob/main/api/context/src/strictContextEnabledTest/java/io/opentelemetry/context/StrictContextStorageTest.java#L170

bogdandrutu: If the error pattern is to always start a new scope and never close we can add a check for the scope depth and log error when larger than 1000 or something like that

anuraaga: I think that assumes the user doesn't try to work around context issues with `Context.root().makeCurrent()`, which java instrumentation currently does since we have scope bugs and don't use this debug mechanism yet. Checking for GC-leaked scopes is more direct with less corner cases I think.

bogdandrutu: I will have to think about that. But we should keep this discussion open and if we are able to get rid of the GC part would be great. We can have the GC logic as a test helper for testing purposes, but if the logic controlled by this flag gets simpler and users can run with this with a reasonable overhead may actually be better for them. But let's brainstorm here

anuraaga: Sure - for reference, the majority of the overhead of the flag is from recording the stacktrace when attaching, which makes finding a leak much easier than only knowing the stack when detaching. The GC listener is something, but very minor in comparison to that.

jkwatson:
I'm going to close this for now, as I believe this ship has sailed several months ago. Please re-open with an issue if you think there is further discussion needed.

PR #2776, opened 2021-02-09
Make StrictContextStorage closeable.

anuraaga:
This doesn't make `StrictContextStorage` public but at least allows instrumentation tests to do something like this if they want.

```java
@AfterEach
void tearDown() {
  ((AutoCloseable) ContextStorage.get()).close();
}
```

@mateuszrzeszutek has been looking into activating strict context storage in our tests and we do need some way to call this.

### DIFF
--- a/api/context/src/main/java/io/opentelemetry/context/StrictContextStorage.java
+++ b/api/context/src/main/java/io/opentelemetry/context/StrictContextStorage.java
@@ -33,37 +33,11 @@ import java.util.stream.Collectors;
 
 /**
  * A {@link ContextStorage} which keeps track of opened and closed {@link Scope}s, reporting caller
- * information if a {@link Scope} is closed incorrectly or not at all. This is useful in
- * instrumentation tests to check whether any scopes leaked.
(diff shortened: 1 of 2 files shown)


api/context/src/main/java/io/opentelemetry/context/StrictContextStorage.java
```diff
@@ -65,7 +65,7 @@
  * > }
  * }</pre>
  */
-final class StrictContextStorage implements ContextStorage {
+final class StrictContextStorage implements ContextStorage, AutoCloseable {
```
jkwatson, line 68: could you add a bit of javadoc to this class that explains that the close implementation does. I see it on the method, but I think it would be additionally helpful to have it at the class level.  (Also, I think that the example is using out of date APIs while you're in here. There is no `ContextStorage.set(...)` any more)

anuraaga: Sure, thanks

mateuszrzeszutek (approved):

jkwatson (approved):
Thanks!

Issue #2093, opened 2020-11-18
Use a single named logger instance for all invalid API usages


jkwatson:
Currently, we silently fail invalid usages of the API as no-ops. It would be very useful for debugging to have a single named Logger that we can use to log all invalid API usages. It should not be enabled by default (maybe log at FINE or FINEST level), but its name should be publicly documented.

This is preferable to having individual loggers for each piece of the API, because that would entail having developers to have to manage many different logger instances, rather than having one central way to enable or disable this debugging information.

Oberon00:
Loggers can be configured by namespace. E.g. you can set the level of everything under `io.opentelemetry.api` and `.sdk` with one line each.

jkwatson:
> > This is preferable to having individual loggers for each piece of the API, because that would entail having developers to have to manage many different logger instances,
> 
> Loggers can be configured by package name (prefix). E.g. you can set the level of everything under `io.opentelemetry.api` and `.sdk` with one line each.

Sure, but that's likely to get a bunch of other things that are unrelated to API usage problems (for example, exporter or span processor logging in the SDK).

gcatanese:
@jkwatson Hi John, I would like to try and help, can you please point me to an example of invalid usage you have mentioned?

jkwatson:
> @jkwatson Hi John, I would like to try and help, can you please point me to an example of invalid usage you have mentioned?

One example would be passing `null` into any of the API methods that do not allow nulls.

gcatanese:
@jkwatson Ok, thanks, this is what I propose (given my first contribution I would rather check I understand the problem and follow the approach before creating a PR)

Create a shared logger in `DefaultOpenTelemetry`
```     
   public static final Logger NO_OPS_LOGGER = Logger.getLogger("NoOpsLogging");
```
**Alternative**: If the public logger is not a desirable solution I would still suggest having a shared constant for the logger name ("NoOpsLogging") while creating the Logger in each class

Log as finest when needed
```
  if (context == null) {     
      NO_OPS_LOGGER.finest("Span.fromContext: context is null");    
      return Span.getInvalid();
    }
```

Logging is enabled only on-demand (`logging.level.NoOpsLogging=TRACE`) 

Let me know what you think (and if you agree with any useful info/guidelines would be appreciated - I imagine I would have to 'touch' several classes)

jkwatson:
[...]

PR #2777, opened 2021-02-09
Use empty default for null attributes and discard for null unit

anuraaga:
Supercedes #2770

### DIFF
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
@@ -283,11 +283,8 @@ final class RecordEventsReadableSpan implements ReadWriteSpan {
 
   @Override
   public ReadWriteSpan addEvent(String name, long timestamp, TimeUnit unit) {
-    if (name == null) {
+    if (name == null || unit == null) {
       return this;
-    }
-    if (unit == null) {
-      unit = TimeUnit.MILLISECONDS;
     }
     addTimedEvent(EventData.create(unit.toNanos(timestamp), name, Attributes.empty(), 0));
     return this;
@@ -313,14 +310,11 @@ final class RecordEventsReadableSpan implements ReadWriteSpan {
 
   @Override
   public ReadWriteSpan addEvent(String name, Attributes attributes, long timestamp, TimeUnit unit) {
-    if (name == null) {
+    if (name == null || unit == null) {
       return this;
     }
     if (attributes == null) {
       attributes = Attributes.empty();
-    }
-    if (unit == null) {
-      unit = TimeUnit.MILLISECONDS;
     }
     int totalAttributeCount = attributes.size();
     addTimedEvent(
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
@@ -286,6 +286,9 @@ final class RecordEventsReadableSpan implements ReadWriteSpan {
     if (name == null) {
       return this;
     }
+    if (unit == null) {
+      unit = TimeUnit.MILLISECONDS;
+    }
     addTimedEvent(EventData.create(unit.toNanos(timestamp), name, Attributes.empty(), 0));
     return this;
   }
@@ -294,6 +297,9 @@ final class RecordEventsReadableSpan implements ReadWriteSpan {
   public ReadWriteSpan addEvent(String name, Attributes attributes) {
     if (name == null) {
       return this;
+    }
+    if (attributes == null) {
+      attributes = Attributes.empty();
     }
     int totalAttributeCount = attributes.size();
     addTimedEvent(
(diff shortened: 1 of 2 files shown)


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -286,6 +286,9 @@ public ReadWriteSpan addEvent(String name, long timestamp, TimeUnit unit) {
     if (name == null) {
       return this;
     }
+    if (unit == null) {
+      unit = TimeUnit.MILLISECONDS;
```
Oberon00, line 290: That seems questionable. In that case, maybe it would be better to fall back to using the current time as if `addEvent(name)` -- and log something.

anuraaga: Yeah that's true - I remembered our convention of env variables all being millisecond so thought of this. But ignoring the timestamp does seem better.

jkwatson: I'm not sure what the correct behavior might be. Can we add some javadocs to the API to say that malformed inputs like this will have indeterminate results?

anuraaga: Hmm - I wonder what a good technique is. One option is add a copy-paste blob to each of the `package-info` in the API artifact, another is to add it to `OpenTelemetry`, our "entry point". Or even both. Any thoughts?

jkwatson: I don't have any good thoughts on how to document this in a way that doesn't mean going to each method and examining it closely for possible weird corner cases. We could go so far as to just mark the `TimeUnit` nullable and specify the behavior exactly in the API, but that also feels weird. We really don't want people to send in null parameters, but we can't stop them, so what do we do?

anuraaga: I realized that there isn't so much we can do - we can expect the percentage that pass null here to not read docs to be large :) I think we just pick a behavior and stick with it - the goal is to not crash the app, it isn't to ensure telemetry works when used incorrectly.

So I don't think we should fall back to current time - it obfuscates the bug even more since the data will be slightly realistic. Instead passing a number and seeing a ridiculous timeline in the backend (more likely backends will filter them out if sophisticated enough) could make it easier to find the bug. Naturally, we do need to implement the named logger for parameter validation to present more info too.

So I would just pick a fallback unit and go with it everywhere there's a chance of an app crash. I lean towards milliseconds since we do use it for unitless timestamps elsewhere. And it's not important to have any reason since we just need something to prevent app crash.

Oberon00: Another option would be to discard the event completely, like we do with null attribute values.

Or you could add an extra event attribute with error message.

kenfinnigan: While I appreciate the desire to not crash a user's application, and I don't have all the history of previous discussions/decisions on similar topics, but is there a desire to "fail" as early as possible?

If a `null` value leads to milliseconds being used, it's highly likely this only gets noticed in monitoring systems, and even then only if the value associated with the milliseconds is outside a "normal" range of expected values. I could see the values being "normal" and therefore no problems are flagged, or that they're so bad alerts and other monitoring escalations are triggered.

Would it be reasonable to assume some level of application testing is performed before deployment to production?

Could we do the same as a `null` `name` and do nothing?

jkwatson: I the risk I see is that someone writes poor library instrumentation, built in to a library, then someone's app crashes when they upgrade the library. Is that ok?

anuraaga: > Or you could add an extra event attribute with error message.

@Oberon00 This is actually one of the reasons I was leaning towards not discarding, in the future adding attributes for such error messages may or may not make sense.

> While I appreciate the desire to not crash a user's application, and I don't have all the history of previous discussions/decisions on similar topics, but is there a desire to "fail" as early as possible?

There is a desire to fail at startup, but we don't want to cause anymore crashes when the user's app is actually running. These tend to be the scariest issues for users, and the reality is many users expect telemetry to get some data "for free" since it doesn't have direct business value. They're angry when the release of a new feature has to get rolled back due to crashes, but can live with some lost tracing until the next release. The spec describes it here

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md#error-handling-in-opentelemetry

and I guess it ended up like that because I'm probably not the only one that's dealt with a customer angry about crashes from a tracing SDK :)

anuraaga: Discard won so went with it!

trask (approved):

Oberon00 (approved):
LGTM.

I think we should create a follow-up issue how to better support debugging such usage errors generally, e.g. if we want to have some logging.

trask:
> e.g. if we want to have some logging

I'm a fan of log once at warn (and mentioning in the log message that only the first occurrence will be logged).

anuraaga:
We have #2093 for our (long standing) issue about reporting validation errors

jkwatson (approved):
This seems fine to me, and agree with eventually wanting to provide some sort of lightweight feedback to the user to let them know that APIs are being misused.

carlosalberto (approved):

PR #2778, opened 2021-02-09
Don't NPE on null keys when getting from baggage, tracestate, attribu…

anuraaga:
…tes.

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/baggage/ImmutableBaggage.java
+++ b/api/all/src/main/java/io/opentelemetry/api/baggage/ImmutableBaggage.java
@@ -5,8 +5,6 @@
 
 package io.opentelemetry.api.baggage;
 
-import static java.util.Objects.requireNonNull;
-
 import io.opentelemetry.api.internal.ImmutableKeyValuePairs;
 import io.opentelemetry.api.internal.StringUtils;
 import io.opentelemetry.context.Context;
@@ -69,7 +67,9 @@ final class ImmutableBaggage extends ImmutableKeyValuePairs<String, BaggageEntry
 
     @Override
     public BaggageBuilder setParent(Context context) {
-      requireNonNull(context, "context");
+      if (context == null) {
+        return this;
+      }
       parent = Baggage.fromContext(context);
       return this;
     }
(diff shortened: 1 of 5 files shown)


api/all/src/main/java/io/opentelemetry/api/baggage/ImmutableBaggage.java
```diff
@@ -69,7 +67,9 @@ private static Baggage sortAndFilterToBaggage(Object[] data) {
 
     @Override
     public BaggageBuilder setParent(Context context) {
-      requireNonNull(context, "context");
+      if (context == null) {
```
bogdandrutu, line 70: Why not root?

bogdandrutu (commented):
Probably better to exclude the context change from this PR

api/all/src/main/java/io/opentelemetry/api/baggage/ImmutableBaggage.java
jkwatson: do you mean: why not put the Baggage into the root context? That would seem like a pretty unexpected behavior, since the user didn't do anything with the root context.

bogdandrutu: No I mean using root as the parent, aka no Parent behavior

jkwatson: oh, I see. sorry. I was confused. isn't that the same thing as what this is doing, then? You're basically saying that it doesn't have a parent?

bogdandrutu: No, because if I do builder.setParent(nonNullContext).setParent(null) will result in the `nonNullContext` to be the parent.

bogdandrutu: Also I think we should spend a bit more time before doing this change, because I don't think we can easily change this in the future. So we need to make sure this is what we want. Having now throwing NPE allows us to easily change it in the future to no longer throw.

bogdandrutu: To be clear: the null behavior for get is great and I am on board with that, it also matches the `Map.get`

anuraaga: Makes sense, reverted it for now (think I forgot that was there by the time I finished the PR). Left the tweaked test case since it'll be needed sooner than latter.

bogdandrutu (approved):

carlosalberto (approved):

PR #2780, opened 2021-02-09
update the examples for 0.16.0

jkwatson:


examples/metrics/src/main/java/io/opentelemetry/example/metrics/DoubleCounterExample.java
```diff
@@ -21,7 +21,7 @@
  * extensions.
  */
 public class DoubleCounterExample {
-  private static final OpenTelemetry openTelemetry = DefaultOpenTelemetry.builder().build();
+  private static final OpenTelemetry openTelemetry = GlobalOpenTelemetry.get();
```
bogdandrutu, line 24: I like this :)

bogdandrutu (approved):

Issue #2765, opened 2021-02-08
Improve performance of Attributes.of(myKey, "value")


bogdandrutu:
Right now a call to `Attributes.of(myKey, "value")` allocates around 5 new objects:
* 1 array during the call to `ArrayBackedAttributes.sortAndFilterToAttributes`
* 1 array during the call to `ImmutableKeyValuePairs.mergeSort` + copy of the initial array
* 1 array + 1 object during the call to `ImmutableKeyValuePairs.dedup` (ArrayList object + underlying array in the ArrayList). 
* 1 object the ArrayBackedAttributes

JianhuaGong:
We has the performance issue, can you explain why sorting and filtering are performed during put a attribute?

jkwatson:
> We has the performance issue, can you explain why sorting and filtering are performed during put a attribute?

we just hadn't gotten to the optimization yet. :) PR submitted.

JianhuaGong:
> we just hadn't gotten to the optimization yet. :) PR submitted.

ok, we look forward to this version.

PR #2787, opened 2021-02-09
Improve ImmutableKeyValuePairs dedup, do everything inplace

bogdandrutu:
This change will reduce the number of allocations by 2 for the case where all entries are valid,
and by 1 when entries are filtered in the dedup.

**Benchmarks Before:**

Screen Shot 2021-02-09 at 12 18 11 PM

Screen Shot 2021-02-09 at 12 19 31 PM

**Benchmarks After:**

Screen Shot 2021-02-09 at 3 02 20 PM

Screen Shot 2021-02-09 at 3 02 48 PM

Updates https://github.com/open-telemetry/opentelemetry-java/issues/2765

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/internal/ImmutableKeyValuePairs.java
+++ b/api/all/src/main/java/io/opentelemetry/api/internal/ImmutableKeyValuePairs.java
@@ -7,8 +7,7 @@ package io.opentelemetry.api.internal;
 
 import static io.opentelemetry.api.internal.Utils.checkArgument;
 
-import java.util.ArrayList;
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


api/all/src/main/java/io/opentelemetry/api/internal/ImmutableKeyValuePairs.java
```diff
@@ -181,31 +182,39 @@ private static void splitAndMerge(
...
-      result.add(key);
+      previousKey = key;
+      data[size++] = key;
+      data[size++] = value;
     }
-    Collections.reverse(result);
-    return result;
+    // Elements removed from the array, copy the array. We optimize for the case where we don't have
+    // duplicates or invalid entries.
+    if (data.length != size) {
```
anuraaga, line 212: Do you think we'd add `size` as a field in a future PR to avoid this copy?

anuraaga (approved):

api/all/src/main/java/io/opentelemetry/api/internal/ImmutableKeyValuePairs.java
bogdandrutu: I am working on that, but the problem is with hash/equal which are no longer easy to implement. But yes I think we can do that.

Issue #2508, opened 2021-01-13
example to integrate metric with logging or jmx exporter


maneeshbhunwal123:
I have integrated with opentelemetery java sdk manually. And now i want to the metric that i am publishing to opentelemetery sdk. I am not finding any examples to do so, can someone please help here? I would be happy to add the steps as a PR.

jkwatson:
Yes, it would be very useful to have some metrics examples that include an exporter which isn't prometheus. 

If you look at the current metrics examples, it should be as simple as setting up an `IntervalMetricReader` with an exporter attached, and assigning the SdkMeterProvider's MetricProducer to it.

A PR to add this to the metrics examples would be great!

maneeshbhunwal123:
where should it be added, i got it working, 
should it be added https://opentelemetry.io/docs/java/?

jkwatson:
If you could add a running example to our `examples` project, that would be a great start.

jkwatson:
closing. resolved with #2789

PR #2789, opened 2021-02-10
2508: Logging Exporter Example for Trace and Metrics

reggiemcdonald:
This PR adds an example to the examples directory showing how to make use of the logging exporters for trace and metrics in a single runnable example. I thought this would help to address #2508 

The jaeger example was used as a basis for this example, altered to make use of the logging exporters

### DIFF
--- a/examples/logging/src/main/java/io/opentelemetry/example/logging/ExampleConfiguration.java
+++ b/examples/logging/src/main/java/io/opentelemetry/example/logging/ExampleConfiguration.java
@@ -42,6 +42,6 @@ public class ExampleConfiguration {
         SdkTracerProvider.builder()
             .addSpanProcessor(SimpleSpanProcessor.create(new LoggingSpanExporter()))
             .build();
-    return OpenTelemetrySdk.builder().setTracerProvider(tracerProvider).build();
+    return OpenTelemetrySdk.builder().setTracerProvider(tracerProvider).buildAndRegisterGlobal();
   }
 }
--- a/examples/logging/src/main/java/io/opentelemetry/example/logging/ExampleConfiguration.java
+++ b/examples/logging/src/main/java/io/opentelemetry/example/logging/ExampleConfiguration.java
@@ -24,7 +24,7 @@ public class ExampleConfiguration {
    *
    * @return A ready-to-use {@link OpenTelemetry} instance.
    */
-  public static OpenTelemetry initOpenTelemetry() {
+  public static void initOpenTelemetry() {
     // This will be used to create instruments
     SdkMeterProvider meterProvider = SdkMeterProvider.builder().buildAndRegisterGlobal();
 
@@ -35,13 +35,12 @@ public class ExampleConfiguration {
         .setMetricProducers(Collections.singleton(meterProvider))
         .setExportIntervalMillis(METRIC_EXPORT_INTERVAL_MS)
         .build();
-
     // Tracer provider configured to export spans with SimpleSpanProcessor using
     // the logging exporter.
     SdkTracerProvider tracerProvider =
         SdkTracerProvider.builder()
             .addSpanProcessor(SimpleSpanProcessor.create(new LoggingSpanExporter()))
             .build();
-    return OpenTelemetrySdk.builder().setTracerProvider(tracerProvider).build();
+    OpenTelemetrySdk.builder().setTracerProvider(tracerProvider).buildAndRegisterGlobal();
   }
 }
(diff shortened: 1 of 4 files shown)


examples/logging/src/main/java/io/opentelemetry/example/logging/ExampleConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+   *
+   * @return A ready-to-use {@link OpenTelemetry} instance.
+   */
+  public static OpenTelemetry initOpenTelemetry() {
+    // This will be used to create instruments
+    SdkMeterProvider meterProvider = SdkMeterProvider.builder().buildAndRegisterGlobal();
+
+    // Create an instance of IntervalMetricReader and configure it
+    // to read metrics from the meterProvider and export them to the logging exporter
+    IntervalMetricReader.builder()
```
bogdandrutu, line 33: Confused what you do with the result of the `build()`?

reggiemcdonald: I don't actually use the result. However, calling `build()` creates an instance of `IntervalMetricReader` and thus starts it: 
https://github.com/open-telemetry/opentelemetry-java/blob/e363e891dca81a8b07045cfe524990b639882d39/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java#L61-L70

bogdandrutu: facepalm :)

examples/logging/src/main/java/io/opentelemetry/example/logging/ExampleConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+        .setExportIntervalMillis(METRIC_EXPORT_INTERVAL_MS)
+        .build();
+
+    // Tracer provider configured to export spans with SimpleSpanProcessor using
+    // the logging exporter.
+    SdkTracerProvider tracerProvider =
+        SdkTracerProvider.builder()
+            .addSpanProcessor(SimpleSpanProcessor.create(new LoggingSpanExporter()))
+            .build();
+    return OpenTelemetrySdk.builder().setTracerProvider(tracerProvider).build();
```
bogdandrutu, line 45: Register to global to be consistent with metrics?

bogdandrutu (approved):

examples/logging/src/main/java/io/opentelemetry/example/logging/ExampleConfiguration.java
anuraaga, line 45:
```suggestion
    return OpenTelemetrySdk.builder().setTracerProvider(tracerProvider).buildAndRegisterGlobal();
```

anuraaga (approved):
Thanks for the help @reggiemcdonald!

examples/logging/src/main/java/io/opentelemetry/example/logging/ExampleConfiguration.java
reggiemcdonald: Yep good call

examples/logging/src/main/java/io/opentelemetry/example/logging/LoggingExporterExample.java
```diff
@@ -16,8 +16,8 @@
   private final Tracer tracer;
   private final LongCounter counter;
 
-  public LoggingExporterExample(OpenTelemetry openTelemetry) {
-    tracer = openTelemetry.getTracer(INSTRUMENTATION_NAME);
+  public LoggingExporterExample() {
+    tracer = GlobalOpenTelemetry.getTracer(INSTRUMENTATION_NAME);
```
anuraaga, line 20: Ah sorry for the confusion, we still use the returned instance here. The global is only to support special instrumentation such as mysql.

reggiemcdonald: gotcha sorry will revert

examples/logging/src/main/java/io/opentelemetry/example/logging/ExampleConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+    // This will be used to create instruments
+    SdkMeterProvider meterProvider = SdkMeterProvider.builder().buildAndRegisterGlobal();
+
+    // Create an instance of IntervalMetricReader and configure it
+    // to read metrics from the meterProvider and export them to the logging exporter
+    IntervalMetricReader.builder()
+        .setMetricExporter(new LoggingMetricExporter())
+        .setMetricProducers(Collections.singleton(meterProvider))
+        .setExportIntervalMillis(METRIC_EXPORT_INTERVAL_MS)
+        .build();
```
jkwatson, line 37: It does make me think that we should rename this to `start()` to make it super clear that it starts up the background thread.

jkwatson (approved):

PR #2792, opened 2021-02-10
Rename disabled resources property to be consistent with agent.

### LINKED ISSUE #2526: Consider changing property naming convention
The Java instrumentation repo is using `-` instead of `_` in property names, e.g. `otel.java.disabled.resource-providers` instead of  `otel.java.disabled.resource_providers`.

Copying comment here from https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1414#issuecomment-733181398 in favor of kebab-case:

[Spring Boot env var / system property mappings](https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-relaxed-binding) supports both camelCase and kebab-case, but recommends kebab-case:

> Property | Note
> -- | --
> acme.my-project.person.first-name | Kebab case, which is recommended for use in .properties and .yml files.

And also interestingly:

> The prefix value for the annotation must be in kebab case (lowercase and separated by -, such as acme.my-project.person).

anuraaga:
Lacking anything in the spec, this follows the rule of do what agent is doing right now. 

It's a small external configuration change which I think is within the scope of RC.

Fixes #2526

### DIFF
--- a/sdk/common/build.gradle.kts
+++ b/sdk/common/build.gradle.kts
@@ -89,12 +89,12 @@ tasks {
 
     named<Test>("testResourceDisabledByProperty") {
         jvmArgs("-Dotel.java.disabled.resource-providers=io.opentelemetry.sdk.extension.resources.OsResource,io.opentelemetry.sdk.extension.resources.ProcessResource")
+        // Properties win, this is ignored.
+        environment("OTEL_JAVA_DISABLED_RESOURCE_PROVIDERS", "io.opentelemetry.sdk.extension.resources.ProcessRuntimeResource")
     }
 
     named<Test>("testResourceDisabledByEnv") {
         environment("OTEL_JAVA_DISABLED_RESOURCE_PROVIDERS", "io.opentelemetry.sdk.extension.resources.OsResource,io.opentelemetry.sdk.extension.resources.ProcessResource")
-        // Environment variable wins, this is ignored.
-        jvmArgs("-Dotel.java.disabled.resource-providers=io.opentelemetry.sdk.extension.resources.ProcessRuntimeResource")
 
     }
 }
--- a/sdk/common/build.gradle.kts
+++ b/sdk/common/build.gradle.kts
@@ -3,12 +3,18 @@ plugins {
     id("maven-publish")
 
     id("ru.vyarus.animalsniffer")
+    id("org.unbroken-dome.test-sets")
 }
 
 description = "OpenTelemetry SDK Common"
 extra["moduleName"] = "io.opentelemetry.sdk.common"
 
 val mrJarVersions = listOf(9)
+
+testSets {
+    create("testResourceDisabledByProperty")
+    create("testResourceDisabledByEnv")
+}
 
 dependencies {
     api(project(":api:all"))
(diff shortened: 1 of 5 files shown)


trask (approved):

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -40,7 +40,7 @@
 public abstract class Resource {
 
   private static final String OTEL_JAVA_DISABLED_RESOURCES_PROVIDERS_PROPERTY_KEY =
-      "otel.java.disabled.resource_providers";
+      "otel.java.disabled.resource-providers";
```
bogdandrutu, line 43: Should this not be in autoconfig module?

bogdandrutu (commented):
Naive question: should some of this be part of the autoconfig?

jkwatson:
> Naive question: should some of this be part of the autoconfig?

I think this is a very good question. I've been wrestling with this question for a couple weeks now. I think it probably *should* be moved to autoconfig. It's definitely confusing to have this one SPI here in the main SDK, while the others are all over in the autoconfig module. 

What would the downsides be to moving it (aside from the obvious pre-1.0 behavior change)?

jkwatson (approved):
Let's decide on moving the SPI to autoconfigure as a separate issue. I'll open it.

jkwatson:
> Let's decide on moving the SPI to autoconfigure as a separate issue. I'll open it.

#2796

sdk/common/build.gradle.kts
```diff
@@ -79,4 +86,15 @@ tasks {
...
\ No newline at end of file
+
+    named<Test>("testResourceDisabledByProperty") {
+        jvmArgs("-Dotel.java.disabled.resource-providers=io.opentelemetry.sdk.extension.resources.OsResource,io.opentelemetry.sdk.extension.resources.ProcessResource")
+    }
+
+    named<Test>("testResourceDisabledByEnv") {
+        environment("OTEL_JAVA_DISABLED_RESOURCE_PROVIDERS", "io.opentelemetry.sdk.extension.resources.OsResource,io.opentelemetry.sdk.extension.resources.ProcessResource")
+        // Environment variable wins, this is ignored.
+        jvmArgs("-Dotel.java.disabled.resource-providers=io.opentelemetry.sdk.extension.resources.ProcessRuntimeResource")
```
mateuszrzeszutek, line 97: Hmm, does that mean that environment variable has precedence over system property in this case? In all other configuration props system properties > environment.

jkwatson: this is a great catch, @mateuszrzeszutek ! Yes, we should be consistent and do it one way. I don't have an opinion which is better, but consistency is needed.

anuraaga: Thanks a lot, must have gotten mixed up at some point! Glad I wrote these tests 😂

trask (approved):

anuraaga:
Let's get this in while making a final call on #2796

jkwatson:
Is this a breaking change?

anuraaga:
Changelog-wise I think it classifies as such, but not really semver wise since we don't guarantee telemetry stability which is what this is.

jkwatson:
> Changelog-wise I think it classifies as such, but not really semver wise since we don't guarantee telemetry stability which is what this is.

ok.  👍

PR #2793, opened 2021-02-10
Make kotlin stdlib dependencies compileOnly since a user will bring t…

anuraaga:
…hem in.

See https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2206/files#r571379473


iNikem (approved):

jkwatson (approved):
Thanks!

PR #2799, opened 2021-02-11
Add Span#setAttributes method to add Attributes to a Span

HaloFour:
Adds helper methods `Span#setAttributes` and `SpanBuilder#setAttributes` which accept an `Attributes` instance and copy the attributes into the span.

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/trace/Span.java
+++ b/api/all/src/main/java/io/opentelemetry/api/trace/Span.java
@@ -168,6 +168,23 @@ public interface Span extends ImplicitContextKeyed {
   }
 
   /**
+   * Sets attributes to the {@link Span}. If the {@link Span} previously contained a mapping for any
+   * of the keys, the old values are replaced by the specified values.
+   *
+   * @param attributes the attributes
+   * @return this.
+   */
+  @SuppressWarnings("unchecked")
+  default Span setAllAttributes(Attributes attributes) {
+    if (attributes == null || attributes.isEmpty()) {
+      return this;
+    }
+    attributes.forEach(
+        (attributeKey, value) -> this.setAttribute((AttributeKey<Object>) attributeKey, value));
+    return this;
+  }
+
+  /**
    * Adds an event to the {@link Span}. The timestamp of the event will be the current time.
    *
    * @param name the name of the event.
--- a/api/all/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java
+++ b/api/all/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java
@@ -229,6 +229,23 @@ public interface SpanBuilder {
   <T> SpanBuilder setAttribute(AttributeKey<T> key, T value);
 
   /**
+   * Sets attributes to the {@link SpanBuilder}. If the {@link SpanBuilder} previously contained a
+   * mapping for any of the keys, the old values are replaced by the specified values.
+   *
+   * @param attributes the attributes
+   * @return this.
+   */
+  @SuppressWarnings("unchecked")
+  default SpanBuilder setAllAttributes(Attributes attributes) {
+    if (attributes == null || attributes.isEmpty()) {
+      return this;
+    }
+    attributes.forEach(
+        (attributeKey, value) -> setAttribute((AttributeKey<Object>) attributeKey, value));
+    return this;
+  }
+
+  /**
    * Sets the {@link SpanKind} for the newly created {@code Span}. If not called, the implementation
    * will provide a default value {@link SpanKind#INTERNAL}.
    *
--- a/sdk/trace/src/test/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpanTest.java
+++ b/sdk/trace/src/test/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpanTest.java
@@ -475,6 +475,118 @@ class RecordEventsReadableSpanTest {
   }
 
   @Test
+  void setAllAttributes() {
+    RecordEventsReadableSpan span = createTestRootSpan();
+    Attributes attributes =
+        Attributes.builder()
+            .put("StringKey", "StringVal")
+            .put("NullStringKey", (String) null)
+            .put("EmptyStringKey", "")
+            .put(stringKey("NullStringAttributeValue"), null)
+            .put(stringKey("EmptyStringAttributeValue"), "")
+            .put("LongKey", 1000L)
+            .put(longKey("LongKey2"), 5)
+            .put(longKey("LongKey3"), 6L)
+            .put("DoubleKey", 10.0)
+            .put("BooleanKey", false)
+            .put(
+                stringArrayKey("ArrayStringKey"),
+                Arrays.asList("StringVal", null, "", "StringVal2"))
+            .put(longArrayKey("ArrayLongKey"), Arrays.asList(1L, 2L, 3L, 4L, 5L))
+            .put(doubleArrayKey("ArrayDoubleKey"), Arrays.asList(0.1, 2.3, 4.5, 6.7, 8.9))
+            .put(booleanArrayKey("ArrayBooleanKey"), Arrays.asList(true, false, false, true))
+            // These should be dropped
+            .put(stringArrayKey("NullArrayStringKey"), null)
+            .put(longArrayKey("NullArrayLongKey"), null)
+            .put(doubleArrayKey("NullArrayDoubleKey"), null)
+            .put(booleanArrayKey("NullArrayBooleanKey"), null)
+            // These should be maintained
+            .put(longArrayKey("ArrayWithNullLongKey"), Arrays.asList(new Long[] {null}))
+            .put(stringArrayKey("ArrayWithNullStringKey"), Arrays.asList(new String[] {null}))
+            .put(doubleArrayKey("ArrayWithNullDoubleKey"), Arrays.asList(new Double[] {null}))
+            .put(booleanArrayKey("ArrayWithNullBooleanKey"), Arrays.asList(new Boolean[] {null}))
+            .build();
+
+    try {
+      span.setAllAttributes(attributes);
+    } finally {
+      span.end();
+    }
+
+    SpanData spanData = span.toSpanData();
+    assertThat(spanData.getAttributes().size()).isEqualTo(16);
+    assertThat(spanData.getAttributes().get(stringKey("StringKey"))).isNotNull();
+    assertThat(spanData.getAttributes().get(stringKey("EmptyStringKey"))).isNotNull();
+    assertThat(spanData.getAttributes().get(stringKey("EmptyStringAttributeValue"))).isNotNull();
+    assertThat(spanData.getAttributes().get(longKey("LongKey"))).isNotNull();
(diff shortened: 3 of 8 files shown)


api/all/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java
```diff
@@ -233,6 +233,27 @@
...
   <T> SpanBuilder setAttribute(AttributeKey<T> key, @Nonnull T value);
 
+  /**
+   * Sets attributes to the {@code Span}.  If the {@code Span} previously contained a mapping for
+   * any of the keys, the old values are replaced by the specified values.
+   *
+   * @param attributes the attributes
+   * @return this.
+   */
+  default Span setAttributes(Attributes attributes) {
```
jkwatson, line 243: this needs to be a `SpanBuilder` returned here.

jkwatson:
Cool. This is a great addition to the APIs. I think we need some unit tests for these methods as well. Thanks!

api/all/src/main/java/io/opentelemetry/api/trace/Span.java
```diff
@@ -162,6 +163,26 @@ default Span setAttribute(AttributeKey<Long> key, int value) {
...
   }
 
+  /**
+   * Sets attributes to the {@code Span}. If the {@code Span} previously contained a mapping for any
+   * of the keys, the old values are replaced by the specified values.
+   *
+   * @param attributes the attributes
+   * @return this.
+   */
+  default Span setAttributes(Attributes attributes) {
```
iNikem, line 173: This seems more like `addAll` to me than `set`. The latter implies that new attributes completely replace the old ones.

HaloFour: I'd almost argue that `set*` should be called `put*` given the `Map`-like semantics.  But I'll let you guys argue/settle on the names.  😁 

I've worked out my Gradle/IDE issues and hopefully will get some unit tests up later this morning.  Apologies for the rapid fire commits last night if they resulted in notifications.

jkwatson: `setAll` is another option for this.

api/all/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java
```diff
@@ -233,6 +234,26 @@
...
+    if (attributes == null || attributes.isEmpty()) {
+      return this;
+    }
+    SpanBuilder span = this;
+    for (Map.Entry<AttributeKey<?>, Object> entry : attributes.asMap().entrySet()) {
+      @SuppressWarnings("unchecked")
+      AttributeKey<Object> attributeKey = (AttributeKey<Object>) entry.getKey();
+      span = span.setAttribute(attributeKey, entry.getValue());
+    }
+    return span;
```
jkwatson, line 254: Can't you replace this section with this:
```
    attributes.forEach((key, o) -> this.setAttribute((AttributeKey<Object>) key, o));
    return this;
```

With this, we'd need to add an `@SuppressWarnings("unchecked")` to the interface method, which would be a bit odd, I guess. At the very least, we could put this more efficient implementation down in the SDK's implemenation.

jkwatson: That annotation doesn't make it into bytecode or javadoc, so maybe it would be ok?

jkwatson: Ah, a little better:
```
    attributes.forEach((key, v) -> {
      AttributeKey<Object> k = (AttributeKey<Object>) key;
      setAttribute(k, v);
    });
    return this;
```

HaloFour: Sure.  I was thinking I'd want to be agnostic about whether `setAttribute` returned the same instance or a different immutable instance but if it's always guaranteed to be the former then it's a moot point.  I could always scope the unchecked operation to an assignment to a local within the lambda which I would think Hotspot should be able to elide.

api/all/src/main/java/io/opentelemetry/api/trace/Span.java
iNikem: Map-like would be then `putAll` :)

HaloFour: Yep.  I'm game with whatever name you guys prefer, let me know and I'll update the PR.

api/all/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java
jkwatson: hm. interesting. I suppose there could be an impl that was immutable and this wouldn't work with that. Perhaps just push the better one down into the SDK where we know we're mutable?

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkSpanBuilder.java
```diff
@@ -152,6 +152,17 @@ public SpanBuilder setAttribute(String key, boolean value) {
     return this;
   }
 
+  @Override
+  @SuppressWarnings("unchecked")
+  public SpanBuilder setAttributes(Attributes attributes) {
```
jkwatson, line 157: can you put some javadoc on this method explaining why we override the default (for performance, fewer allocations, etc)?

api/all/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java
anuraaga: The javadoc has `@return this`, which I think specifies the contract for this interface that it's always mutable so I think we don't need to worry too much.

HaloFour: Good point, I'm being too defensive here.  I'll also update the name of the method to `setAllAttributes` to I think more clearly describe what it does.

api/all/src/main/java/io/opentelemetry/api/trace/Span.java
jkwatson: well, IMO, all the rest of the methods, for good or for ill, are `set*`, so we should definitely start with `set`. :)

api/all/src/main/java/io/opentelemetry/api/trace/Span.java
```diff
@@ -162,6 +162,23 @@ default Span setAttribute(AttributeKey<Long> key, int value) {
...
+   * @param attributes the attributes
+   * @return this.
+   */
+  @SuppressWarnings("unchecked")
+  default Span setAllAttributes(Attributes attributes) {
+    if (attributes == null || attributes.isEmpty()) {
+      return this;
+    }
+    attributes.forEach(
+        (attributeKey, value) -> this.setAttribute((AttributeKey<Object>) attributeKey, value));
```
bogdandrutu, line 178: `s/Object/?`?

api/all/src/main/java/io/opentelemetry/api/trace/Span.java
```diff
@@ -162,6 +162,23 @@ default Span setAttribute(AttributeKey<Long> key, int value) {
     return setAttribute(key, (long) value);
   }
 
+  /**
+   * Sets attributes to the {@code Span}. If the {@code Span} previously contained a mapping for any
```
anuraaga, line 166:
```suggestion
   * Sets attributes to the {@link Span}. If the {@link Span} previously contained a mapping for any
```

api/all/src/main/java/io/opentelemetry/api/trace/Span.java
anuraaga, line 178: Thanks @HaloFour, this follows the pattern we have in a couple other places like `putAll` where we hide the unchecked cast from users with this sort of shortcut.

api/all/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java
```diff
@@ -233,6 +233,23 @@
    */
   <T> SpanBuilder setAttribute(AttributeKey<T> key, @Nonnull T value);
 
+  /**
+   * Sets attributes to the {@code Span}. If the {@code Span} previously contained a mapping for any
```
anuraaga, line 237:
```suggestion
   * Sets attributes to the {@link SpanBuilder}. If the {@link SpanBuilder} previously contained a mapping for any
```

sdk/trace/src/test/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpanTest.java
```diff
@@ -463,6 +463,118 @@ void setAttribute_nullAttributeValue() {
...
+            .build();
+
+    try {
+      span.setAllAttributes(attributes);
+    } finally {
+      span.end();
+    }
+
+    SpanData spanData = span.toSpanData();
+    assertThat(spanData.getAttributes().size()).isEqualTo(16);
```
anuraaga, line 506: You may consider a map assertion on `spanData.getAttributes.asMap()` though this is fine too

api/all/src/main/java/io/opentelemetry/api/trace/Span.java
HaloFour: `attributeKey` is already an `AttributeKey<?>`, it needs to be unchecked cast to `AttributeKey<Object>` in order to call `setAttribute` with `value` since it is an `Object`.

jkwatson:
Pending a decision on the name of the method, this looks great. Thanks!

anuraaga:
We are going to give the decision on `setAttribute` vs `putAttribute` another day to settle but have tentatively decided to go with the former, so we will likely continue with the current name `setAllAttributes`. Thanks for the patience @HaloFour

anuraaga (approved):
@HaloFour Sorry for the delay on the response on this - the method name looks great, do you mind rebasing and we'll get this in?

Oberon00:
Is this something that could be brought up in the spec?

anuraaga:
@Oberon00 I see, in the spec indeed there is a sentence about "single attribute"

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-attributes

Let me file a spec issue - hope it doesn't take too long :)

anuraaga:
https://github.com/open-telemetry/opentelemetry-specification/pull/1473 is the spec issue

HaloFour:
Cool beans.  Give me a heads up as to whether we're unblocked by the spec issue and I'll rebase.

jkwatson:
https://github.com/open-telemetry/opentelemetry-specification/pull/1473 has been merged, so this has been unblocked, @HaloFour .

HaloFour:
Excellent, I'll fix the merge conflict and rebase.  Thanks!

carlosalberto (approved):

jkwatson (approved):
Thanks!

jkwatson:
Sorry, I had forgotten about this PR until I ran into this myself. approved. 
@bogdandrutu Any objections to this API addition?

jkwatson:
Merging! Thanks @HaloFour, and sorry for the long delay!

PR #2800, opened 2021-02-11
introduction of View and first version of LabelsProessor

parallelstream:
* Added View as a wrapper around AggregatorFactory and LabelsProcessorFactory
* Added LabelsProcessor interfaces

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
@@ -12,7 +12,9 @@ import io.opentelemetry.sdk.internal.ComponentRegistry;
 import io.opentelemetry.sdk.metrics.aggregator.AggregatorFactory;
 import io.opentelemetry.sdk.metrics.data.MetricData;
 import io.opentelemetry.sdk.metrics.export.MetricProducer;
+import io.opentelemetry.sdk.metrics.processor.LabelsProcessorFactory;
 import io.opentelemetry.sdk.metrics.view.InstrumentSelector;
+import io.opentelemetry.sdk.metrics.view.View;
 import io.opentelemetry.sdk.resources.Resource;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -102,6 +104,12 @@ public final class SdkMeterProvider implements MeterProvider, MetricProducer {
    * }</pre>
    */
   public void registerView(InstrumentSelector selector, AggregatorFactory aggregatorFactory) {
-    sharedState.getViewRegistry().registerView(selector, aggregatorFactory);
+    sharedState.getViewRegistry().registerView(selector, View.builder().setAggregatorFactory(aggregatorFactory).build());
+  }
+  public void registerView(InstrumentSelector selector, AggregatorFactory aggregatorFactory,
+      LabelsProcessorFactory labelsProcessorFactory) {
+    sharedState.getViewRegistry().registerView(selector, View.builder()
+        .setAggregatorFactory(aggregatorFactory)
+        .setLabelsProcessorFactory(labelsProcessorFactory).build());
   }
 }
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
@@ -9,10 +9,8 @@ import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.api.metrics.MeterProvider;
 import io.opentelemetry.sdk.common.Clock;
 import io.opentelemetry.sdk.internal.ComponentRegistry;
-import io.opentelemetry.sdk.metrics.aggregator.AggregatorFactory;
 import io.opentelemetry.sdk.metrics.data.MetricData;
 import io.opentelemetry.sdk.metrics.export.MetricProducer;
-import io.opentelemetry.sdk.metrics.processor.LabelsProcessorFactory;
 import io.opentelemetry.sdk.metrics.view.InstrumentSelector;
 import io.opentelemetry.sdk.metrics.view.View;
 import io.opentelemetry.sdk.resources.Resource;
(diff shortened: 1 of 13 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
```diff
@@ -102,6 +104,12 @@ public static SdkMeterProviderBuilder builder() {
    * }</pre>
    */
   public void registerView(InstrumentSelector selector, AggregatorFactory aggregatorFactory) {
```
bogdandrutu, line 106: We should change this to be:

`public void registerView(InstrumentSelector selector, View view)`

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/processor/LabelsProcessor.java
```diff
@@ -0,0 +1,18 @@
+package io.opentelemetry.sdk.metrics.processor;
+
+import io.opentelemetry.api.metrics.common.Labels;
+import io.opentelemetry.context.Context;
+
+public interface LabelsProcessor {
```
bogdandrutu, line 6: Hard to review with a lot of errors in the UI, do you mind running `./gradlew spotlessApply`?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
parallelstream: done

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/processor/LabelsProcessor.java
parallelstream: done

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/processor/BaggageLabelsProcessor.java
```diff
@@ -0,0 +1,26 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.metrics.processor;
+
+import io.opentelemetry.api.metrics.common.Labels;
+import io.opentelemetry.api.metrics.common.LabelsBuilder;
+import io.opentelemetry.context.Context;
+
+public class BaggageLabelsProcessor implements LabelsProcessor {
```
bogdandrutu, line 12: * final
* javadoc

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/processor/LabelsProcessor.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package io.opentelemetry.sdk.metrics.processor;
+
+import io.opentelemetry.api.metrics.common.Labels;
+import io.opentelemetry.context.Context;
+
+public interface LabelsProcessor {
+  /**
+   * Called when bind() method is called. Allows to manipulate labels which this instrument is bound
```
bogdandrutu, line 13: Actually `bind()` is not very accurate, it is an implementation detail in our SDK that we implement both the "bounds" and "direct records" using the same internal `bind()` method in the accumulator. Probably you need to use API concepts when describing this instead of internal concepts in the SDK.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/processor/LabelsProcessor.java
```diff
@@ -0,0 +1,25 @@
...
+   * Context
+   *
+   * @param ctx context of the operation
+   * @param labels immutable labels. When processors are chained output labels of the previous one
+   *     is passed as an input to the next one. Last labels returned by a chain of processors are
+   *     used for bind() operation.
+   * @return labels to be used as an input to the next processor in chain or bind() operation if
+   *     this is the last processor
+   */
+  Labels onLabelsBound(Context ctx, Labels labels);
```
bogdandrutu, line 24: Worth also explaining that this is experimental for the moment, and does not work correctly with BoundInstruments.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/processor/LabelsProcessorFactory.java
```diff
@@ -0,0 +1,23 @@
...
+  static LabelsProcessorFactory baggageExtractor(BaggageMetricsLabelsExtractor labelsExtractor) {
+    return () -> new BaggageLabelsProcessor(labelsExtractor);
+  }
+
+  /**
+   * Returns a new {@link LabelsProcessorFactory}.
+   *
+   * @return new {@link LabelsProcessorFactory}
+   */
+  LabelsProcessor create();
```
bogdandrutu, line 22: I think we should use the same input as for AccumulatorFactory: "resource, libraryinfo, instrumentdescriptor" for consistency. Probably we can be smarter about this and embed them, but for the moment just be consistent.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/processor/NoopLabelsProcessor.java
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.metrics.processor;
+
+import io.opentelemetry.api.metrics.common.Labels;
+import io.opentelemetry.context.Context;
+
+public class NoopLabelsProcessor implements LabelsProcessor {
```
bogdandrutu, line 11: what about we make `create` return "nullable" value, and return null by default? That way we know if no processor is installed and we can optimize more in the accumulator (not even retrieving the current context, etc).

bogdandrutu (commented):
Initial feedback about the new interfaces

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/processor/LabelsProcessor.java
parallelstream: tried to provide a more accurate description. I had to double check, but seems that bind() as well as synchronous instrument is a part of API concepts. I did get your point though and tried avoid using implementation details to describe the behavior

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/processor/LabelsProcessorFactory.java
parallelstream: done

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/processor/LabelsProcessor.java
parallelstream: done

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/processor/BaggageLabelsProcessor.java
parallelstream: done

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/processor/NoopLabelsProcessor.java
parallelstream: I am not so sure about this one. Having nulls in public methods is usually considered an anti pattern, I do understand though that we have this call on a critical path for every metric recording. Seems like getting current context is pretty lightweight - just pulling it from thread local most of the time. Unless you insist I would avoid null's. If you do insist I am happy to change though

bogdandrutu: you are correct, probably we need to expose just the noop instance, so at least we can do the trick if we want to compare with noop reference which should be same as comparing with null.

jkwatson:
do we still need this PR now that the others have been merged?

parallelstream:
no

PR #2806, opened 2021-02-11
simplify the Jaeger integration test

### LINKED ISSUE #2804: Simplify the JaegerExporterIntegrationTest
This test currently has a lot of moving parts, which might be a cause of it failing quite often locally.

I recommend removing the separate jar file generation, and separate docker instance to run it, and just put the code in that class into the test itself.

jkwatson:
resolves #2804


integration-tests/build.gradle
```diff
@@ -18,9 +18,5 @@ dependencies {
             "org.testcontainers:junit-jupiter",
             "com.squareup.okhttp3:okhttp"
 
-    tasks.withType(Test) {
```
anuraaga, line 21: Can you remove the shadow plugin up top too (which I assume isn't needed anymore)?

anuraaga (approved):

iNikem (approved):

integration-tests/build.gradle
jkwatson: oh yes, definitely. good catch.

jkwatson: done

PR #2811, opened 2021-02-12
Default exporters to OTLP per spec.

anuraaga:
Default is `otlp` in spec. Also removed deprecated exporter names.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#exporter-selection

I've gone ahead and added `none` here in anticipation of https://github.com/open-telemetry/opentelemetry-specification/pull/1439/files and given this is still an alpha module.


jkwatson (approved):
Thanks!

trask (approved):

PR #2821, opened 2021-02-16
Use schemed endpoint for jaeger exporter.

anuraaga:
While looking at https://github.com/open-telemetry/opentelemetry-java-instrumentation/discussions/2266 I noticed that the spec defines endpoint as a schemed URL for Jaeger too, but we don't handle it. This means there's no environment variable for specifying an insecure or secure connection.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#jaeger-exporter

I just copied the code from the OTLP exporter, presumably we need to clean these up to remove support for schemaless though.

### DIFF
--- a/exporters/jaeger/src/main/java/io/opentelemetry/exporter/jaeger/JaegerGrpcSpanExporterBuilder.java
+++ b/exporters/jaeger/src/main/java/io/opentelemetry/exporter/jaeger/JaegerGrpcSpanExporterBuilder.java
@@ -10,15 +10,24 @@ import static java.util.Objects.requireNonNull;
 
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.time.Duration;
 import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /** Builder utility for this exporter. */
 public final class JaegerGrpcSpanExporterBuilder {
-  private static final String DEFAULT_ENDPOINT = "localhost:14250";
+
+  private static final Logger logger =
+      Logger.getLogger(JaegerGrpcSpanExporterBuilder.class.getName());
+
+  private static final String DEFAULT_ENDPOINT_URL = "http://localhost:14250";
+  private static final URI DEFAULT_ENDPOINT = URI.create(DEFAULT_ENDPOINT_URL);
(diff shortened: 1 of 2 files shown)


exporters/jaeger/src/main/java/io/opentelemetry/exporter/jaeger/JaegerGrpcSpanExporterBuilder.java
```diff
@@ -35,13 +44,34 @@ public JaegerGrpcSpanExporterBuilder setChannel(ManagedChannel channel) {
...
+      throw new IllegalArgumentException("Invalid endpoint, must be a URL: " + endpoint, e);
+    }
+    // TODO(anuraaga): Remove after announcing deprecation of schemaless URLs.
+    if (uri.getScheme() != null) {
+      if (!uri.getScheme().equals("http") && !uri.getScheme().equals("https")) {
+        throw new IllegalArgumentException("Invalid scheme, must be http or https: " + endpoint);
+      }
+    } else {
+      logger.log(
+          Level.WARNING,
```
jkwatson, line 69: should we also log this, or something similar, up on line 53 when we have to insert the scheme?

jkwatson (approved):
Thanks!

exporters/jaeger/src/main/java/io/opentelemetry/exporter/jaeger/JaegerGrpcSpanExporterBuilder.java
```diff
@@ -35,13 +44,34 @@ public JaegerGrpcSpanExporterBuilder setChannel(ManagedChannel channel) {
...
+      endpoint = "http://" + endpoint;
+    }
+
+    URI uri;
+    try {
+      uri = new URI(endpoint);
+    } catch (URISyntaxException e) {
+      throw new IllegalArgumentException("Invalid endpoint, must be a URL: " + endpoint, e);
+    }
+    // TODO(anuraaga): Remove after announcing deprecation of schemaless URLs.
```
trask, line 62: does this comment apply instead to lines 52-54 above?

trask (approved):

anuraaga (commented):
I just copied this from OTLP, let's stick with it for now. I think we need to follow up with our already deprecated removing of support for schemaless URLs (for jaeger it will be removed without deprecation period since it's late in the cycle)

PR #2823, opened 2021-02-16
Simplify ParentBased equals/hashcode and add coverage.

anuraaga:


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/ParentBasedSampler.java
```diff
@@ -104,28 +103,20 @@ public boolean equals(Object o) {
...
-      return false;
-    }
-    if (!Objects.equals(remoteParentNotSampled, that.remoteParentNotSampled)) {
-      return false;
-    }
-    if (!Objects.equals(localParentSampled, that.localParentSampled)) {
-      return false;
-    }
-    return Objects.equals(localParentNotSampled, that.localParentNotSampled);
+    return root.equals(that.root)
```
anuraaga, line 106: Fields are never null

sdk/trace/src/test/java/io/opentelemetry/sdk/trace/samplers/ParentBasedSamplerTest.java
```diff
@@ -407,4 +407,9 @@ void getDescription() {
                 + "remoteParentNotSampled:AlwaysOffSampler,localParentSampled:AlwaysOnSampler,"
                 + "localParentNotSampled:AlwaysOffSampler}");
   }
+
+  @Test
+  void equals() {
+    EqualsVerifier.forClass(ParentBasedSampler.class).verify();
```
breedx-splk, line 413: Slick!

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/ParentBasedSampler.java
breedx-splk: The annotations on the constructor params indicate otherwise.  If this is true, we should change that.  Maybe even `@NotNull` them?

breedx-splk (changes requested):
Thanks for the cleanup...this reads much better now.  Not sure about those `@Nullable` fields not being null tho.  😉

jkwatson (approved):
Isn't the equals/hashcode only there for testing assertions? I'm surprised it didn't have some coverage already due to that.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/ParentBasedSampler.java
jkwatson: note that they get replaced with the defaults if null in the constructor impl.

anuraaga:
@jkwatson No clue, I found a lot of red here https://codecov.io/gh/open-telemetry/opentelemetry-java/src/main/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/ParentBasedSampler.java

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/ParentBasedSampler.java
anuraaga: Yeah - it's a bit confusing, let me send another PR to replace Nullable with setting the defaults in the builder.

breedx-splk: Got it.

PR #2824, opened 2021-02-16
clean up the javadoc and make examples pasteable

### LINKED ISSUE #2817: Some javadoc with code examples in it has some mangled content
See here for example:  https://www.javadoc.io/doc/io.opentelemetry/opentelemetry-sdk-extension-tracing-incubator/latest/io/opentelemetry/sdk/extension/incubator/trace/data/DelegatingSpanData.html

the `{@literal @}` leaks into the actual javadoc. There is no perfect solution to this, but one approach is the one taken here:  https://www.javadoc.io/doc/io.opentelemetry/opentelemetry-context/latest/io/opentelemetry/context/ContextStorage.html . that is, to prepend a character before each line in the example code.

jkwatson:
resolves #2817

### DIFF
--- a/api/metrics/src/main/java/io/opentelemetry/api/metrics/DoubleSumObserver.java
+++ b/api/metrics/src/main/java/io/opentelemetry/api/metrics/DoubleSumObserver.java
@@ -21,28 +21,28 @@ import javax.annotation.concurrent.ThreadSafe;
  * <p>Example:
  *
  * <pre>{@code
- * class YourClass {
- *
- *   private static final Meter meter = OpenTelemetry.getMeterProvider().get("my_library_name");
- *   private static final DoubleSumObserver cpuObserver =
- *       meter.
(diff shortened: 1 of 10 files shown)


carlosalberto (approved):

api/metrics/src/main/java/io/opentelemetry/api/metrics/DoubleSumObserver.java
```diff
@@ -21,28 +21,28 @@
...
- *          {@literal @}Override
- *           public void update(DoubleResult result) {
- *             // Get system cpu usage
- *             result.observe(cpuIdle, "state", "idle");
- *             result.observe(cpuUser, "state", "user");
- *           }
- *         });
- *   }
- * }
+ * // class YourClass {
```
anuraaga, line 24: Nice idea!

anuraaga (approved):

PR #2828, opened 2021-02-16
Properly shutdown tracer providers synchronously via close()

### LINKED ISSUE #2825: Autoconfigure does not wait for shutdown of TracerProvider to complete
See: https://github.com/open-telemetry/opentelemetry-java/blob/23adacaf2e89584b2005e7a84333578b5e3f6700/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java#L51

Autoconfigure adds a shutdown hook that calls `TracerProviderSdk.shutdown()`. But shutdown is an async method that returns a CompletableResultCode. So any work on daemon threads will just be cancelled. I think a call to `CompletableResultCode.join` is missing (what the timeout should be is an interesting question).

I ran into this while playing around with the auto-instrumenting agent and the Jaeger exporter. When adding a custom span processor (as in examples/distro) that actually does something that takes a bit of time, the Jager exporter is not able to finish. CC @trask

jkwatson:
Resolves #2825

I updated the examples, as well, since they were also not doing it correctly.

### DIFF
--- a/examples/http/src/main/java/io/opentelemetry/example/http/ExampleConfiguration.java
+++ b/examples/http/src/main/java/io/opentelemetry/example/http/ExampleConfiguration.java
@@ -37,7 +37,7 @@ class ExampleConfiguration {
             .setPropagators(ContextPropagators.create(W3CTraceContextPropagator.getInstance()))
             .build();
 
-    Runtime.getRuntime().addShutdownHook(new Thread(sdkTracerProvider::shutdown));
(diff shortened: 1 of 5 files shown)


breedx-splk (approved):
LGTM

trask (approved):

anuraaga (approved):
Oops, thanks!

PR #2829, opened 2021-02-16
Obfuscate the GlobalOpenTelemetry instance.

### LINKED ISSUE #2788: Obfuscate GlobalOpenTelemetry.get()
I'm nervous about users casting the return value of `GlobalOpenTelemetry.get()` to `OpenTelemetrySdk`, e.g.

```
((OpenTelemetrySdk) GlobalOpenTelemetry.get()).getSdkTracerProvider()...
```

It seems like a reasonable way to get the global SDK now that we no longer have `OpenTelemetrySdk.get()`, so I suspect users will find their way to doing it.

The reason I'm nervous about this, is that the javaagent returns a bridged OpenTelemetry instance from GlobalOpenTelemetry.get(), which will then cause the user's app to throw ClassCastException, and a (the?) top priority of the java agent is to not break a user's application.

This is similar to the reason the return value from `GlobalOpenTelemetry.get().getTracerProvider()` is obfuscated, in order to prevent similar global access to the SDK:

```
((SdkTracerProvider) GlobalOpenTelemetry.get().getTracerProvider())...
```

jkwatson:
This is to prevent people from casting to the SDK implementation.
Resolves #2788

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
+++ b/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
@@ -14,6 +14,7 @@ import java.lang.reflect.Method;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.Nullable;
+import javax.annotation.concurrent.ThreadSafe;
 
 /**
  * A global singleton for the entrypoint to telemetry functionality for tracing, metrics and
@@ -33,7 +34,7 @@ public final class GlobalOpenTelemetry {
 
   private static final Object mutex = new Object();
 
-  @Nullable private static volatile OpenTelemetry globalOpenTelemetry;
+  @Nullable private static volatile ObfuscatedOpenTelemetry globalOpenTelemetry;
 
   @GuardedBy("mutex")
   @Nullable
@@ -86,7 +87,7 @@ public final class GlobalOpenTelemetry {
                 + "instead. Previous invocation set to cause of this exception.",
             setGlobalCaller);
       }
-      globalOpenTelemetry = openTelemetry;
+      globalOpenTelemetry = new ObfuscatedOpenTelemetry(openTelemetry);
       setGlobalCaller = new Throwable();
     }
   }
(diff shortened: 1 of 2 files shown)


trask (approved):
Thanks 🙏

carlosalberto (approved):

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
```diff
@@ -166,4 +167,29 @@ private static OpenTelemetry maybeAutoConfigure() {
...
+      this.delegate = delegate;
+    }
+
+    @Override
+    public TracerProvider getTracerProvider() {
+      return delegate.getTracerProvider();
+    }
+
+    @Override
+    public ContextPropagators getPropagators() {
```
anuraaga, line 191: Can you add this to the test too?

sdk/all/src/test/java/io/opentelemetry/sdk/OpenTelemetrySdkTest.java
```diff
@@ -44,10 +44,10 @@ void tearDown() {
...
     OpenTelemetrySdk sdk = OpenTelemetrySdk.builder().buildAndRegisterGlobal();
-    assertThat(sdk).isSameAs(GlobalOpenTelemetry.get());
-    assertThat(GlobalOpenTelemetry.get()).isSameAs(sdk);
-    assertThat(((OpenTelemetrySdk) GlobalOpenTelemetry.get()).getSdkTracerProvider().get(""))
-        .isSameAs(GlobalOpenTelemetry.getTracerProvider().get(""));
+    assertThat(GlobalOpenTelemetry.get()).extracting("delegate").isSameAs(sdk);
+    assertThat(sdk.getTracerProvider().get(""))
+        .isSameAs(GlobalOpenTelemetry.getTracerProvider().get(""))
+        .isSameAs(GlobalOpenTelemetry.get().getTracer(""));
   }
```
anuraaga, line 51: Can you add a test that attempts to cast the SDK?

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
jkwatson: done

sdk/all/src/test/java/io/opentelemetry/sdk/OpenTelemetrySdkTest.java
jkwatson: done

anuraaga (approved):

PR #2831, opened 2021-02-16
[Breaking Change] Rename getEmpty() to empty()

### LINKED ISSUE #2786: Future consistency of simple getter names / accessors in general
In general we use the `getFoo` even for static methods and default values, but we missed this in the `Baggage`
https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/baggage/Baggage.java#L33

This issue is not to change the API, but to ensure that from now on we stay with the `getFoo` pattern.

**Update** 

* `empty()` looks to be more like what Java prefers based on the usage in `Collections`, but we don't respect that in few places: https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/common/src/main/java/io/opentelemetry/sdk/common/InstrumentationLibraryInfo.java#L42 and https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java#L121

jkwatson:
Provisionally resolves #2786 

Do not merge unless we agree to make this change before 1.0.0

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/trace/ArrayBasedTraceStateBuilder.java
+++ b/api/all/src/main/java/io/opentelemetry/api/trace/ArrayBasedTraceStateBuilder.java
@@ -27,7 +27,7 @@ final class ArrayBasedTraceStateBuilder implements TraceStateBuilder {
   private final ArrayBasedTraceState parent;
   @Nullable private List<String> entries;
 
-  static TraceState getEmpty() {
+  static TraceState empty() {
     return EMPTY;
   }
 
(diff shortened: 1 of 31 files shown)


jsuereth:
Comments from the peanut gallery: 

1. We aren't using "empty" method in our exporters or their tests.  So this change looks ok to me.
2. These last minute changes are the ones I regret most later.  Are all of these like collections?  Is consistency within the class more important?  This is one where you could make persuasive arguments either way (for this change or not).

anuraaga:
@jsuereth Definitely hear you on the danger of last-second changes :) We discussed in detail at our meeting and do believe these are all collections where the word `empty` works well.

anuraaga (approved):

Issue #2816, opened 2021-02-12
AutoConfiguration: Use the SimpleSpanProcessor if the LoggingSpanExporter is selected


jkwatson:
If the LoggingSpanExporter is selected for autoconfiguration (via `otel.traces.exporter=logging`), it would make more sense to use the simple span exporter, rather than the batch.

reggiemcdonald:
I can do this if no one else is working on it?

PR #2832, opened 2021-02-16
2816: Autoconfigure to use SimpleSpanProcessor when logging exporter is used

reggiemcdonald:
This PR updates `TracerProviderConfiguration` to use a `SimpleSpanProcessor` whenever the environment variable `otel.traces.exporter` is set to `logging`. Right now on the main branch, an autoconfigured SDK always uses a `BatchSpanProcessor`.

fixes: #2816

### DIFF
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java
@@ -18,16 +18,9 @@ import io.opentelemetry.sdk.trace.export.SimpleSpanProcessor;
 import io.opentelemetry.sdk.trace.export.SpanExporter;
 import io.opentelemetry.sdk.trace.samplers.Sampler;
 import java.time.Duration;
-import java.util.Collections;
-import java.util.HashSet;
 import java.util.ServiceLoader;
-import java.util.Set;
 
 final class TracerProviderConfiguration {
-
-  /** A set of the exporter names that should use the SimpleSpanProcessor. */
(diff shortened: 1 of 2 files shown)


breedx-splk (approved):
Thanks for contributing this....this change will make logging/debugging much nicer because developers can see spans quickly instead of waiting for a batch.  This looks good to me.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java
```diff
@@ -11,15 +11,24 @@
...
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.ServiceLoader;
+import java.util.Set;
 
 final class TracerProviderConfiguration {
 
+  /** A set of the exporter names that should use the SimpleSpanProcessor. */
+  private static final Set<String> SPAN_EXPORTER_USING_SIMPLE_SPAN_PROCESSOR =
+      new HashSet<>(Collections.singletonList("logging"));
```
jkwatson, line 30: this can just all be replaced with `Collections.singleton("logging")`, I think

jkwatson: But, I think we can probably just dispense with the set, and just look for the one string directly for now. I don't know if we will ever have more than one hard-coded simple span processor exporter.

reggiemcdonald: Ahh gotcha - I wasn't sure what the future plans were. I'll just make it a string

anuraaga (approved):
Thanks!

jkwatson (approved):

PR #2833, opened 2021-02-17
[Breaking Change] Rename SpanLimits methods to be closer to the env vars

jkwatson:
Please do not merge unless we agree we can make this change before 1.0.0

### DIFF
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java
@@ -85,17 +85,17 @@ final class TracerProviderConfiguration {
 
     Integer maxAttrs = config.getInt("otel.span.attribute.count.limit");
     if (maxAttrs != null) {
-      builder.setMaxNumberOfAttributes(maxAttrs);
+      builder.setSpanAttributeLimit(maxAttrs);
     }
 
     Integer maxEvents = config.getInt("otel.span.event.count.limit");
     if (maxEvents != null) {
-      builder.setMaxNumberOfEvents(maxEvents);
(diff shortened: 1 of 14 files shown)


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SpanLimits.java
```diff
@@ -44,66 +44,66 @@ public static SpanLimitsBuilder builder() {
...
-  public abstract int getMaxNumberOfAttributes();
+  public abstract int getSpanAttributeLimit();
 
   /**
    * Returns the global default max number of events per {@link Span}.
    *
    * @return the global default max number of events per {@code Span}.
    */
-  public abstract int getMaxNumberOfEvents();
+  public abstract int getSpanEventLimit();
```
jkwatson, line 74: The word "Span" in these methods is probably redundant, since we're in the SpanLimits class.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java
```diff
@@ -85,17 +85,17 @@ static SpanLimits configureSpanLimits(ConfigProperties config) {
 
     Integer maxAttrs = config.getInt("otel.span.attribute.count.limit");
     if (maxAttrs != null) {
-      builder.setMaxNumberOfAttributes(maxAttrs);
+      builder.setSpanAttributeLimit(maxAttrs);
```
trask, line 88: I'd vote for `setAttributeCountLimit()` etc in order to match the env var naming (and agree with leaving off prefix otel.span as redundant)

anuraaga:
We discussed and felt that no name is ideal for these, and the current names seem fine so we won't do a rename. We will enhance our documentation to explain which methods correspond to which env vars.

jkwatson:
closing. we decided against making this change.

PR #2835, opened 2021-02-17
Reduce some NPE in API.

anuraaga:


api/all/src/main/java/io/opentelemetry/api/trace/ArrayBasedTraceState.java
```diff
@@ -21,6 +21,9 @@
   @Override
   @Nullable
   public String get(String key) {
+    if (key == null) {
```
jkwatson, line 24: Is this one really necessary? Looks like it'll just fall through and return null as currently implemented, no?

api/all/src/test/java/io/opentelemetry/api/trace/DefaultTracerTest.java
```diff
@@ -91,4 +78,7 @@ void testSpanContextPropagation_fromContextThenNoParent() {
     Span span = defaultTracer.spanBuilder(SPAN_NAME).setParent(context).setNoParent().startSpan();
     assertThat(span.getSpanContext()).isEqualTo(SpanContext.getInvalid());
   }
+
+  @Test
+  void doesNotCrash() {}
```
jkwatson, line 83: ?

api/all/src/main/java/io/opentelemetry/api/trace/ArrayBasedTraceState.java
anuraaga: Not really necessary, nah, but it makes it more obvious that we gracefully handle null instead of having to verify the implementation

api/all/src/test/java/io/opentelemetry/api/trace/DefaultTracerTest.java
anuraaga: Well - it didn't, right? :D

jkwatson (approved):

PR #2837, opened 2021-02-17
[Breaking Change] Move ResourceProvider SPI to autoconfigure module

### LINKED ISSUE #2796: Consider moving the ResourceProvider SPI to the autoconfigure module
This is the only SPI left in the SDK. All others are managed in the autoconfigure module. It would be less confusing to move it now, before 1.0, if we have the appetite for it.

anuraaga:
- Moves `ResourceProvider` SPI
- Removes `Resource.getTelemetrySdk`
- Adds `OpenTelemtrySdkAutoconfiguration.getResource`
- Replaces existing `ResourceProvider` implementations with static factories of `Resource` for use in manual merging
- Implements with compileOnly new `ResourceProvider` implementations that provide it

Fixes #2796


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
```diff
@@ -50,5 +61,26 @@ private static void configureMeterProvider(Resource resource, ConfigProperties c
     MetricExporterConfiguration.configureExporter(exporterName, config, meterProvider);
   }
 
+  private static Resource buildResource() {
+    ConfigProperties config = ConfigProperties.get();
+    Resource result = Resource.getDefault();
+
+    // TODO(anuraaga): We use a hyphen only once in this artifact, for
```
anuraaga, line 68: I punted on a cleaner way of this for now

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
```diff
@@ -50,5 +61,26 @@ private static void configureMeterProvider(Resource resource, ConfigProperties c
...
+    Set<String> disabledProviders =
+        new HashSet<>(config.getCommaSeparatedValues("otel.java.disabled.resource.providers"));
+    for (ResourceProvider resourceProvider : ServiceLoader.load(ResourceProvider.class)) {
+      if (disabledProviders.contains(resourceProvider.getClass().getName())) {
+        continue;
+      }
+      result = result.merge(resourceProvider.createResource(config));
+    }
+
+    result = result.merge(EnvironmentResource.getInstance());
```
mateuszrzeszutek, line 80: `EnvironmentResource` does not really need a static final instance anymore, so if you just make it a simple factory class you'll be able to reuse `config` and avoid constructing another `ConfigProperties` inside the `EnvironmentResource` class

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/OsResource.java
```diff
@@ -7,25 +7,33 @@
...
 import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.common.AttributesBuilder;
-import io.opentelemetry.sdk.resources.ResourceProvider;
+import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.semconv.resource.attributes.ResourceAttributes;
 import javax.annotation.Nullable;
 
-/** {@link ResourceProvider} which provides information about the current operating system. */
-public final class OsResource extends ResourceProvider {
+/** {@link Resource} which provides information about the current operating system. */
```
mateuszrzeszutek, line 14: Technically, this class isn't a resource, it just contains a method that returns one.

How about sth like
```suggestion
/** Provides a {@link Resource} containing information about the current operating system. */
```
instead?

mateuszrzeszutek (approved):

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/OsResource.java
anuraaga: Yeah it was a bit weird - went with factory to use a noun like is usual for class javadoc. With the change, I liked the `getnstance` name even less than I did before so renamed it to just `get`

mateuszrzeszutek (approved):

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/EnvironmentResource.java
```diff
@@ -7,21 +7,23 @@
 
 import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.common.AttributesBuilder;
-import io.opentelemetry.sdk.resources.ResourceProvider;
+import io.opentelemetry.sdk.resources.Resource;
 
-public class EnvironmentResource extends ResourceProvider {
+final class EnvironmentResource {
```
jkwatson, line 12: shouldn't this be public so that users who want to do everything manually can get a handle to it?

jkwatson: (just to make it like the other resources available in the resource module)

bogdandrutu: We can do that later

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource/EksResource.java
```diff
@@ -21,7 +21,14 @@
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-public final class EksResource extends ResourceProvider {
+/**
+ * A factory of a {@link Resource} which provides information about the current ECS container if
```
jkwatson, line 25: I might change all of these to say "A factory for a ..."

jkwatson: this cleanup can be done post-merge also

jkwatson (approved):
This API seems good to me. Thanks! Will create an issue to expose the EnvironmentResource as a separate (non-breaking) addition.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/EnvironmentResource.java
bogdandrutu: If this will be left public, add final please :)

jkwatson: #2858

PR #2838, opened 2021-02-17
[Breaking Change] Remove BaggageBuilder.setParent / setNoParent

### LINKED ISSUE #2784: Behavior of `Baggage.toBuilder` seems tricky to understand
The current behavior of the `Baggage.toBuilder` seems a bit tricky when it is combine with a parent. I am not sure based on the javadoc if I can understand the behavior (without looking at the code).

So I suggest to improve the javadoc for this method and clearly explain what would be the result when a parent is set, how conflicts (duplicate keys are resolved).

PS: @jkwatson there is always the option to remove this method, used only in tests and benchmarks :)))

anuraaga:
Instead, expect `Baggage.current().toBuilder()`

Fixes #2784

### DIFF
--- a/api/all/src/jmh/java/io/opentelemetry/api/baggage/BaggageBenchmark.java
+++ b/api/all/src/jmh/java/io/opentelemetry/api/baggage/BaggageBenchmark.java
@@ -5,7 +5,6 @@
 
 package io.opentelemetry.api.baggage;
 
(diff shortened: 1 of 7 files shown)


jkwatson (approved):
👍

bogdandrutu (approved):

PR #2839, opened 2021-02-17
[Breaking Change] Make TraceId.getTraceIdRandomPart private

### LINKED ISSUE #2785: Update documentation for getTraceIdRandomPart
See https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/trace/TraceId.java#L119

This is a legacy method since the TraceId was opaque and we could not access this unless asking for bytes or hex (extra allocations). 

Some problems that I think we need to understand:
* The logic of TraceIdRatioBasedSampler may change and may use a different part of the traceID or may compute a hash, specs does not define the algorithm.
* The specs says that the trace-id conforms with the w3c definition, which indeed suggest that the right part of the trace id should be random (https://www.w3.org/TR/trace-context/#randomness-of-trace-id). But does not define the size of that part.

I think we have two options:
* Keep this method and document exactly what it does, and how this does or not confirm with specs/w3c definitions. Probably we should remove the reference to the TraceIdRatioBasedSampler anyway from there since we don't know if that is how it will be implemented.
* @jkwatson favorite approach, remove the method for the moment. That means we can move the code of hex to long from API (is package protected now) to the sampler implementation and use it directly there.

anuraaga:
Fixes #2785


bogdandrutu (approved):
As it seems @anuraaga has a bit of a stronger opinion for this, I am fine. We can always destroy internal package 🗡️

bogdandrutu:
Merging this as we can always improve things.

PR #2842, opened 2021-02-17
[Breaking Change] Rename API getDefault to noop!

anuraaga:
Removes `Tracer.getDefault` and `Meter.getDefault` since they're easy to get from the providers.

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/DefaultOpenTelemetry.java
+++ b/api/all/src/main/java/io/opentelemetry/api/DefaultOpenTelemetry.java
@@ -33,7 +33,7 @@ final class DefaultOpenTelemetry implements OpenTelemetry {
 
   @Override
   public TracerProvider getTracerProvider() {
-    return TracerProvider.getDefault();
+    return TracerProvider.noop();
   }
 
   @Override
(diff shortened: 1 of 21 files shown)


api/all/src/main/java/io/opentelemetry/api/DefaultOpenTelemetry.java
```diff
@@ -17,11 +17,11 @@
 final class DefaultOpenTelemetry implements OpenTelemetry {
   private static final OpenTelemetry NO_OP = new DefaultOpenTelemetry(ContextPropagators.noop());
 
-  static OpenTelemetry noop() {
+  static OpenTelemetry getNoop() {
```
anuraaga, line 20: Had to rename these due to errorprone, which is presumably overstrict on these private APIs

```
/Users/aanuraag/git/opentelemetry-java/api/all/src/main/java/io/opentelemetry/api/DefaultOpenTelemetry.java:26: warning: [FunctionalInterfaceClash] When passing lambda arguments to this function, callers will need a cast to disambiguate with:
  static OpenTelemetry propagating(ContextPropagators propagators) {
                       ^
      OpenTelemetry.propagating(ContextPropagators)
    (see https://errorprone.info/bugpattern/FunctionalInterfaceClash)
error: warnings found and -Werror specified
```

bogdandrutu: Internal API, I don't care :)

bogdandrutu (approved):

carlosalberto (approved):

jkwatson (approved):
noop noop!

PR #2843, opened 2021-02-17
[Breaking Change] Remove TraceId.getTraceIdRandomPart

### LINKED ISSUE #2785: Update documentation for getTraceIdRandomPart

bogdandrutu:
Alternative to https://github.com/open-telemetry/opentelemetry-java/pull/2839, less than 10 lines of duplicate code, but avoid exposing Bigendian in internal.

Fixes #2785

### DIFF
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/TraceIdLong.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/TraceIdLong.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.trace.samplers;
+
+import java.util.Arrays;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+final class TraceIdLong {
+  private static final String ALPHABET = "0123456789abcdef";
+  private static final int ASCII_CHARACTERS = 128;
+  private static final byte[] DECODING = buildDecodingArray();
+
+  private static byte[] buildDecodingArray() {
+    byte[] decoding = new byte[ASCII_CHARACTERS];
+    Arrays.fill(decoding, (byte) -1);
+    for (int i = 0; i < ALPHABET.length(); i++) {
+      char c = ALPHABET.charAt(i);
+      decoding[c] = (byte) i;
+    }
+    return decoding;
+  }
+
+  /**
+   * Returns the {@code long} value whose base16 representation is stored in the first 16 chars of
+   * {@code chars} starting from the {@code offset}.
+   *
+   * @param chars the base16 representation of the {@code long}.
+   * @param offset the starting offset in the {@code CharSequence}.
+   */
+  static long longFromBase16String(CharSequence chars, int offset) {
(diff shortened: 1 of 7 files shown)


carlosalberto (approved):

jkwatson:
I prefer this version, even if it adds a tiny bit of duplicate code. We can always decide to make the BigEndianEncoding class more public later. For now, hiding it completely seems a bit safer.

jkwatson (approved):

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/TraceIdLong.java
```diff
@@ -0,0 +1,51 @@
...
+package io.opentelemetry.sdk.trace.samplers;
+
+import java.util.Arrays;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+final class TraceIdLong {
+  private static final String ALPHABET = "0123456789abcdef";
+  private static final int ASCII_CHARACTERS = 128;
+  private static final byte[] DECODING = buildDecodingArray();
```
anuraaga, line 15: These are pretty significant 10 lines of code :P We have a policy for our internal package, I don't see a reason to avoid it at the expense of duplicating state.

bogdandrutu: Because the functions that we would make public look very interesting for users to call unfortunately

anuraaga: I agree with the sentiment, but we could rename to `OpenTelemetryEncoding` or something I think which would help? `Utils.checkArgument` also seems pretty interesting, and I think to reduce memory usage / improve performance of String building in various places, I'd expect TemporaryBuffers going there at some point - it seems to me like having to think about the deliciousness of classes in the internal package is limiting / not so scalable as a general practice.

bogdandrutu: >  it seems to me like having to think about the deliciousness of classes in the internal package is limiting / not so scalable as a general practice.

Not sure I understand this

Definitely not completely against the other PR. Was an alternative.

bogdandrutu:
@anuraaga had a stronger preference for the other version. closing this

PR #2847, opened 2021-02-17
Mark classes final, or hide ctor for testing classes

bogdandrutu:

### DIFF
--- a/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/OpenTelemetryAssertions.java
+++ b/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/OpenTelemetryAssertions.java
@@ -14,7 +14,7 @@ import org.assertj.core.api.Assertions;
  * static import any {@code assertThat} method in this class instead of {@code
  * Assertions.assertThat}.
  */
-public class OpenTelemetryAssertions extends Assertions {
+public final class OpenTelemetryAssertions extends Assertions {
 
   /** Returns an assertion for {@link Attributes}. */
   public static AttributesAssert assertThat(Attributes attributes) {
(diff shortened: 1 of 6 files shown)


sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/OpenTelemetryAssertions.java
```diff
@@ -26,5 +26,5 @@ public static SpanDataAssert assertThat(SpanData spanData) {
     return new SpanDataAssert(spanData);
   }
 
-  protected OpenTelemetryAssertions() {}
+  private OpenTelemetryAssertions() {}
```
jkwatson, line 29: I feel like there must have been a good reason for this to be `protected` explicitly. Is this being extended somewhere like the instrumentation project.

bogdandrutu: Not used https://github.com/open-telemetry/opentelemetry-java-instrumentation/search?q=OpenTelemetryAssertions

jkwatson: If it turns out there is a good reason for this (maybe in one of the custom agent distributions?), we can very easily open it back up. 👍

jkwatson (approved):

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/OpenTelemetryAssertions.java
anuraaga, line 29: @bogdandrutu @jkwatson Yeah it's normal practice for the entrypoint into assertions to be extendable - it's how a user would be able to add their own custom assertions while using these too.

jkwatson: I'm totally fine opening these back up. It's testing code, not production code, and I think we should make it as easy as possible for our users to write tests for otel and extend as needed.

PR #2863, opened 2021-02-17
[Breaking Change] Remove B3Propagator.Builder, add static methods to access possible instances

bogdandrutu:

### DIFF
--- a/extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/B3Propagator.java
+++ b/extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/B3Propagator.java
@@ -27,23 +27,23 @@ import javax.annotation.concurrent.Immutable;
  * href=https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md#b3-requirements>B3
  * Requirements</a>
  *
- * <p>To register the default B3 propagator, which injects a single header, use the default instance
+ * <p>To register the default B3 propagator, which injects a single header, use:
  *
  * <pre>{@code
  * OpenTelemetry.setPropagators(
  *   DefaultContextPropagators
  *     .builder()
- *     .addTextMapPropagator(B3Propagator.getInstance())
+ *     .addTextMapPropagator(B3Propagator.injectingSingleHeader())
  *     .build());
  * }</pre>
  *
- * <p>To register a B3 propagator that injects multiple headers, use the builder
+ * <p>To register a B3 propagator that injects multiple headers, use:
  *
  * <pre>{@code
  * OpenTelemetry.setPropagators(
  *   DefaultContextPropagators
  *     .builder()
- *     .addTextMapPropagator(B3Propagator.builder().injectMultipleHeaders().build())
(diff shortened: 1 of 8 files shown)


carlosalberto (approved):

extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/B3Propagator.java
```diff
@@ -33,7 +33,7 @@
  * OpenTelemetry.setPropagators(
  *   DefaultContextPropagators
  *     .builder()
- *     .addTextMapPropagator(B3Propagator.getInstance())
+ *     .addTextMapPropagator(B3Propagator.getSingleHeaderInstance())
```
jkwatson, line 36: I think we should still call the single-header option `.getInstance()`. In the specs, single-header is supposed to be the default, with multi-header as configurable.

bogdandrutu: Then what about:

```
getDefault() // Returns the default instance
getMultiHeaders()
```

I think specs are written to a specific implementation in mind that there is a Builder/Factory method that without any options returns the single header version. But as with this implementation we don't have any of these patterns it is hard to reason about default.

There are other places where we can enforce "default", for example "b3" string for env vars and java properties should be the default for single header.

jkwatson: I think that's fair. Maybe the default here is only important for auto-configuration, and not the programmatic ways of accessing the instances. I don't feel overly strongly...I just don't want to do something counter-spec and have to release another breaking change before 1.0. ;)

bogdandrutu: We can always add another method call getDefault on top of the two that returns the single header

extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/B3Propagator.java
```diff
@@ -81,41 +84,27 @@ private B3Propagator(B3PropagatorInjector b3PropagatorInjector) {
...
-   * B3Propagator}.
+   * Returns an instance of the {@link B3Propagator} that injects multi headers format.
+   *
+   * <p>This instance extracts both formats, in the order: single header, multi header.
+   *
+   * @return an instance of the {@link B3Propagator} that injects multi headers format.
    */
-  public static B3Propagator.Builder builder() {
-    return new Builder();
+  public static B3Propagator getMultiHeadersInstance() {
```
anuraaga, line 93: Is `injectingMultiHeaders` a usable name? I think we need `inject` in the name or its very easy to miss that both versions behave the same on extraction.

bogdandrutu: Sounds reasonable to me.

anuraaga (approved):

jkwatson (approved):

PR #2869, opened 2021-02-18
Return null instead of throwing exceptions for TraceFlags.fromHex

bogdandrutu:

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/trace/TraceFlags.java
+++ b/api/all/src/main/java/io/opentelemetry/api/trace/TraceFlags.java
@@ -5,6 +5,8 @@
 
 package io.opentelemetry.api.trace;
 
+import io.opentelemetry.api.internal.OtelEncodingUtils;
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
 /**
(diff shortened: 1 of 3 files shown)


anuraaga (approved):

jkwatson:
Why not return TraceFlags.default in this case?

api/all/src/main/java/io/opentelemetry/api/trace/TraceFlags.java
```diff
@@ -45,16 +47,21 @@ static TraceFlags getSampled() {
...
-   * @throws IllegalArgumentException if invalid characters in the {@code src}.
    */
+  @Nullable
   static TraceFlags fromHex(CharSequence src, int srcOffset) {
+    // TODO: Avoid calling `subSequence`.
+    if (src == null
+        || src.length() < srcOffset + 2
+        || !OtelEncodingUtils.isValidBase16String(src.subSequence(srcOffset, srcOffset + 2))) {
+      return null;
+    }
```
trask, line 64: alternative, just for consideration: return TraceFlags.getInvalid(), similar to SpanContext.getInvalid(), callers can check `isValid()`, or they can forget and it will serialize as "00"

api/all/src/main/java/io/opentelemetry/api/trace/TraceFlags.java
```diff
@@ -45,16 +47,21 @@ static TraceFlags getSampled() {
...
-   * @throws IndexOutOfBoundsException if {@code src} is too short.
-   * @throws IllegalArgumentException if invalid characters in the {@code src}.
    */
+  @Nullable
   static TraceFlags fromHex(CharSequence src, int srcOffset) {
+    // TODO: Avoid calling `subSequence`.
+    if (src == null
+        || src.length() < srcOffset + 2
+        || !OtelEncodingUtils.isValidBase16String(src.subSequence(srcOffset, srcOffset + 2))) {
+      return null;
```
anuraaga, line 63: Do you think we can return a special implementation of `TraceFlags` with a `isValid` method that returns false? It'd be nice not to return null if possible, almost everything in `api/**` tries hard to prevent NPE in calling code in the first place.

anuraaga: Ah - my suggestion is the same as @trask's I think we wrote it at the same time :)

api/all/src/main/java/io/opentelemetry/api/trace/TraceFlags.java
bogdandrutu: The 00 flag is actually a valid representation that has a meaningful interpretation

trask: Yes I know that

trask: Not sure why that's a problem for this proposal?

bogdandrutu: I try to encapsulate somehow the fact that there is an error and leave the caller to decide about default vs not creating the SpanContext

trask: just store a `boolean invalid` field in ImmutableTraceFlags

trask: caller can check `getInvalid()` if they want to know, same as SpanContext

bogdandrutu: I see... Let me play with that tomorrow since release already happened

bogdandrutu:
Too late for this probably :)

PR #2871, opened 2021-02-18
Remove Nonnull annotations, we use package default Nonnull

bogdandrutu:

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/common/AttributeKey.java
+++ b/api/all/src/main/java/io/opentelemetry/api/common/AttributeKey.java
@@ -6,7 +6,6 @@
 package io.opentelemetry.api.common;
 
 import java.util.List;
-import javax.annotation.Nonnull;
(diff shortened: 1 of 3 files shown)


anuraaga (approved):

api/all/src/main/java/io/opentelemetry/api/common/AttributeKey.java
```diff
@@ -24,7 +23,6 @@
   String getKey();
 
   /** Returns the type of attribute for this key. Useful for building switch statements. */
-  @Nonnull
```
Oberon00, line 27: We don't default return values to Nonnull, do we? Is that even possible?

jkwatson: Hmm. this is a good question, and I don't know the answer. We definitely *want* to have this be the default (non-null return values).

PR #2873, opened 2021-02-18
Add tests for resources with securitymanager enabled.

anuraaga:
Saw these in coverage reports and have never actually used a security manager before so was a good opportunity to try it out.


jkwatson (approved):
Thanks!

PR #2881, opened 2021-02-18
Document expected behavior for null params

bogdandrutu:
This follows the way how Java documents null params for interfaces. The API needs to document the possible behaviors not what happens in our SDK (like adding throws annotations). If we want to ensure that no implementation can throw an exception (which I think we shouldn't) we can change the text to say that.

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/trace/Span.java
+++ b/api/all/src/main/java/io/opentelemetry/api/trace/Span.java
@@ -22,6 +22,12 @@ import javax.annotation.concurrent.ThreadSafe;
  *
  * <p>Spans are created by the {@link SpanBuilder#startSpan} method.
  *
+ * <p>When inserting attributes, {@code null} keys and values are prohibited, and in that case the
+ * behavior is undefined.
+ *
+ * <p>When adding events, {@code null} names, attributes, units and timestamps, and in that case the
+ * behavior is undefined.
+ *
  * <p>{@code Span} <b>must</b> be ended by calling {@link #end()}.
  */
 @ThreadSafe
@@ -274,7 +280,8 @@ public interface Span extends ImplicitContextKeyed {
    * <p>Only the value of the last call will be recorded, and implementations are free to ignore
    * previous calls.
    *
-   * @param statusCode the {@link StatusCode} to set.
+   * @param statusCode the {@link StatusCode} to set. {@code null} is prohibited, and in that case
+   *     the behavior is undefined.
    * @return this.
    */
   default Span setStatus(StatusCode statusCode) {
@@ -290,8 +297,10 @@ public interface Span extends ImplicitContextKeyed {
    * <p>Only the value of the last call will be recorded, and implementations are free to ignore
    * previous calls.
    *
-   * @param statusCode the {@link StatusCode} to set.
-   * @param description the description of the {@code Status}.
+   * @param statusCode the {@link StatusCode} to set. {@code null} is prohibited, and in that case
+   *     the behavior is undefined.
+   * @param description the description of the {@code Status}. {@code null} is prohibited, and in
+   *     that case the behavior is undefined.
    * @return this.
    */
   Span setStatus(StatusCode statusCode, String description);
@@ -303,7 +312,8 @@ public interface Span extends ImplicitContextKeyed {
    * this function. You should record this attribute manually using {@link
    * #recordException(Throwable, Attributes)} if you know that an exception is escaping.
    *
-   * @param exception the {@link Throwable} to record.
+   * @param exception the {@link Throwable} to record. {@code null} is prohibited, and in that case
+   *     the behavior is undefined.
    * @return this.
    */
   default Span recordException(Throwable exception) {
@@ -313,8 +323,10 @@ public interface Span extends ImplicitContextKeyed {
   /**
    * Records information about the {@link Throwable} to the {@link Span}.
    *
-   * @param exception the {@link Throwable} to record.
-   * @param additionalAttributes the additional {@link Attributes} to record.
+   * @param exception the {@link Throwable} to record. {@code null} is prohibited, and in that case
+   *     the behavior is undefined.
+   * @param additionalAttributes the additional {@link Attributes} to record. {@code null} is
+   *     prohibited, and in that case the behavior is undefined.
    * @return this.
    */
   Span recordException(Throwable exception, Attributes additionalAttributes);
@@ -327,7 +339,8 @@ public interface Span extends ImplicitContextKeyed {
    * <p>Upon this update, any sampling behavior based on {@code Span} name will depend on the
    * implementation.
    *
-   * @param name the {@code Span} name.
+   * @param name the {@code Span} name. {@code null} is prohibited, and in that case the behavior is
+   *     undefined.
    * @return this.
    */
   Span updateName(String name);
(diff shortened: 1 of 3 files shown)


api/all/src/main/java/io/opentelemetry/api/trace/Span.java
```diff
@@ -22,6 +22,14 @@
  *
  * <p>Spans are created by the {@link SpanBuilder#startSpan} method.
  *
+ * <p>When inserting attributes, some implementations may prohibit {@code null} keys and values, and
```
Oberon00, line 25: "some implementations may prohibit" is wrong IMHO. The API should already prohibit this (as does the spec). It might be better to say "null keys and values are not allowed and the behavior is undefined. Some implementations might be able to cope with them, e.g. by ignoring them while others might throw exceptions (e.g. NullPointerExceptions) or an error might occur later in an exporter."

jkwatson: This brings up an excellent point, @Oberon00 . If the spec "prohibits" this, but the spec also requires that mis-use of the APIs should never throw exceptions, how are we supposed to implement these two conflicting requirements, in general?

Oberon00: The compliance matrix has an entry "`null` values documented as invalid/undefined"

I don't see these requirements as conflicting: On the API level, the behavior is undefined. But for the two concrete implementations in the spec (SDK and Noop-implementation) the spec places the additional requirement on them to not throw errors / return null. However, we should not advertise this to users as behavior they can rely on.

anuraaga: I like the teminology `"null keys and values are not allowed and the behavior is undefined.` at this top level. And I don't think we need any other docs since we don't have any `@Nulable` annotations on the methods themselves - passing `null` would mean they ignored the docs anyways :)

bogdandrutu: If I understood correctly the suggestion this should be done.

api/all/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java
```diff
@@ -265,16 +264,16 @@
    *
    * <p>Important this is NOT equivalent with System.nanoTime().
    *
-   * @param startTimestamp the explicit start timestamp from the epoch of the newly created {@code
-   *     Span}.
+   * @param timestamp the explicit start timestamp from the epoch of the newly created {@code Span}.
+   *     {@code null} indicates calling time of {@link #startSpan()} should be used.
```
bogdandrutu, line 268: Careful about this change.

api/all/src/main/java/io/opentelemetry/api/trace/Span.java
```diff
@@ -364,7 +378,7 @@ default Span recordException(Throwable exception) {
    * <p>Use this method for specifying explicit end options, such as end {@code Timestamp}. When no
    * explicit values are required, use {@link #end()}.
    *
-   * @param timestamp the explicit timestamp from the epoch, for this {@code Span}. {@code 0}
+   * @param timestamp the explicit timestamp from the epoch, for this {@code Span}. {@code null}
    *     indicates current time should be used.
```
bogdandrutu, line 382: Careful about this change.

api/all/src/main/java/io/opentelemetry/api/trace/Span.java
```diff
@@ -22,6 +22,12 @@
  *
  * <p>Spans are created by the {@link SpanBuilder#startSpan} method.
  *
+ * <p>When inserting attributes, {@code null} keys and values are prohibited, and in that case the
```
anuraaga, line 25:
```suggestion
 * <p>When inserting attributes, {@code null} keys and values are prohibited, and if provided, the
```

Think this reads a bit better

api/all/src/main/java/io/opentelemetry/api/trace/Span.java
```diff
@@ -274,7 +280,8 @@ default Span addEvent(String name, Attributes attributes, Instant timestamp) {
    * <p>Only the value of the last call will be recorded, and implementations are free to ignore
    * previous calls.
    *
-   * @param statusCode the {@link StatusCode} to set.
+   * @param statusCode the {@link StatusCode} to set. {@code null} is prohibited, and in that case
```
anuraaga, line 283: Instead of these, can we generify the above even further `{@code null} is prohibited for any method argument in this class, and if provided, the behavior is undefined.`? Well maybe we'll add a `@Nullable` parameter later, but we may not either so we could address it later if that happens.

api/all/src/main/java/io/opentelemetry/api/trace/Span.java
anuraaga: This was about the special behavior for the value `0` which we need to keep. If following the above suggestion, we don't need to add a special note about `null` here (behavior is undefined, so "current time" is a fine behavior).

anuraaga (commented):
On a second thinking, I realized this is mostly about SDKs implementing the API - can we expect such users to be advanced enough to understand the `@Nullable` annotations without documenting this on every single interface? If it's possible, it's nice to optimize these docs to end users of the API instead of implementors.

carlosalberto (approved):

bogdandrutu:
> On a second thinking, I realized this is mostly about SDKs implementing the API - can we expect such users to be advanced enough to understand the @Nullable annotations without documenting this on every single interface? If it's possible, it's nice to optimize these docs to end users of the API instead of implementors.

That is not my motivation :), I think these clarification are for the users of the API, and they need to understand that calling these methods with null with different implementation (not our SDK or noop) may result in NPE or other exceptions. The API cannot guarantee that, so they have to understand this may happen and is not a bug of that implementation.

api/all/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java
jkwatson: this seems reasonable to me.

api/all/src/main/java/io/opentelemetry/api/trace/Span.java
bogdandrutu: Why is 0 value special?

jkwatson: `0` is only special for the method above. Not an issue for this method which takes an `Instant`, I don't think.

api/all/src/main/java/io/opentelemetry/api/trace/Span.java
jkwatson: It seems reasonable to me to put this up at the class level, yes.

anuraaga:
> The API cannot guarantee that, so they have to understand this may happen and is not a bug of that implementation.

Hmm - I guess this is true of many of the methods in the `trace` package, and later `metrics` and `logs`. We could add it to every class, but it does decrease the signal/noise ratio a lot and I don't know how much value there is. Can't we push the burden on documenting behavior differences with our SDK to the hypothetical custom SDK?

anuraaga:
Discussed with @bogdandrutu that we'll go with package-level documentation instead of class-level. For now we'll close this PR

PR #2888, opened 2021-02-18
mark the proto modules as alpha, since they contain metrics and logs

jkwatson:


bogdandrutu (approved):

jkwatson:
@anuraaga whoops! Merged this without your 👀  on it. Can revert if you don't think it's the right move.

anuraaga (commented):
Hmm - I think it'd be a corner case, but since this isn't in our normal BOM anymore by being alpha, I think there could be some chance of version conflict?

jkwatson:
> Hmm - I think it'd be a corner case, but since this isn't in our normal BOM anymore by being alpha, I think there could be some chance of version conflict?

You mean if they have one version of the stable bom and a different version of the alpha-bom (or for some reason are depending on the alpha proto directly)? 

Another, more involved solution, would be to shade/relocated the dependencies into the stable things that need them. I'm not sure how much work that might be.

anuraaga:
> You mean if they have one version of the stable bom and a different version of the alpha-bom (or for some reason are depending on the alpha proto directly)?

I guess the situation I'm imagining is more they're depending on a library which depends on the alpha proto directly, rather than the app doing so. Such a library should also be alpha, and the user would be able to use the alpha bom, but this may not happen.

> Another, more involved solution, would be to shade/relocated the dependencies into the stable things that need them. I'm not sure how much work that might be.

I think it should be possible to split out `proto-alpha` artifact, there could be a more elegant way, but otherwise just calling `Delete` on the folders not needed for each project would do the job in a pinch. If it sounds ok I can do that

PR #2890, opened 2021-02-19
Plugin the custom marshaller to otlp:trace span exporter

bogdandrutu:


jkwatson:
How confident are we that this is ready for prime time? [I have no opinion or knowledge...just asking the question]

jkwatson:
@anuraaga @bogdandrutu I think we're resurrecting this now, but it'll probably be easier to just have a fresh PR, rather than trying to port this one to the current state of affairs. Please re-open if I am wrong in that assessment.

PR #2891, opened 2021-02-19
Allow updates that may require additional POM updates on minor updates.

anuraaga:
I think these are generally allowed by semver since a user needs to update the POM anyways when updating, but we want to make sure they don't need to update code / retest everything.

### DIFF
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -41,8 +41,8 @@ new artifact which requires adding the new artifact to dependency declarations.
 
 As a user, if you always depend on the latest version of the BOM for a given `MAJOR` version, and
 you do not use classes in the `internal` package (which you MUST NOT do), you can be assured that
-your app will always function and have the latest features of OpenTelemetry without needing any
-changes to code.
+your app will always function and have access tothe latest features of OpenTelemetry without needing 
+any changes to code.
 
 ## API vs SDK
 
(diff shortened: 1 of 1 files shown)


jkwatson (approved):

VERSIONING.md
```diff
@@ -35,9 +35,14 @@ Backwards incompatible changes to `internal` packages are expected. Versions of
...
 of versions.
 
+Changes may be made that require changes to the an app's dependency declarations aside from just
+incrementing the version on `MINOR` version updates. For example, code may be separated out to a
+new artifact which requires adding the new artifact to dependency declarations.
+
 As a user, if you always depend on the latest version of the BOM for a given `MAJOR` version, and
 you do not use classes in the `internal` package (which you MUST NOT do), you can be assured that
-your app will always function and have the latest features of OpenTelemetry.
+your app will always function and have the latest features of OpenTelemetry without needing any
```
bogdandrutu, line 44: It is too much this I think. It means that if you add a new feature like a new sampler people still have to do code changes. I am worried about having the latest features

bogdandrutu: Maybe the latest updates for the features you use

anuraaga: I think I meant have access to the new features, reworded a bit

anuraaga: The idea being that you're never gated from new features due to dependency hell

bogdandrutu (approved):

VERSIONING.md
bogdandrutu: Yes, I understand; It sounded more like we promise any new feature will be enabled with just gradle work or system properties :)

PR #2893, opened 2021-02-19
Move otproto to otlp:common, so can be marked as stable

bogdandrutu:
The artifact is used by couple of our other artifacts and we likely prefer to have this in the bom. By renaming it to internal and moving all the code to an internal package this artifact now follows the rules of internal packages in our stable artifacts, which means it can break the API as long as using the bom will still work.


exporters/logging-otlp/build.gradle.kts
```diff
@@ -12,7 +12,7 @@ dependencies {
     compileOnly(project(":sdk:trace"))
     compileOnly(project(":sdk:metrics"))
 
-    implementation(project(":sdk-extensions:otproto"))
+    implementation(project(":exporters:otlp:internal"))
```
anuraaga, line 15: I think we can make this sort of change in the future if we run into a dependency conflict due to the artifact not being in BOM - currently we don't leak this artifact since it's `implementation` here not `api`. And since we transitively depend on the alpha `proto` anyways I don't think we run away from the BOM issue completely anyways with this change.

bogdandrutu: I agree, was just trying to improve the problem somehow. We can run from the proto problem by shading it in this artifact as well in the future.

But independent of solving that problem. It feels strange to have an otlp helper as an sdk-extension, should be a helper in the otlp exporter artifact.

anuraaga: Yeah that's a good point that the new naming seems better

anuraaga (approved):

exporters/logging-otlp/build.gradle.kts
bogdandrutu: Don't have to merge this since the package is alpha now so we can think more about this

bogdandrutu: @jkwatson what do you think?

jkwatson: It's a bit strange to publish an "internal" module to maven central, but it does seem better than publishing something that we never intend to be stable and supported.

jkwatson (approved):

exporters/logging-otlp/build.gradle.kts
bogdandrutu: What about we call it common then and only have an internal directory for the moment.

jkwatson: I guess the question is: will we eventually make this something that is supported and stable, once otlp is stable? 
If yes, then I think that `common` is a better way to go, since we do intend it to eventually be stable and supported, not just internal.

bogdandrutu: I think once the proto fields are stable, I would offer this as a helper.

jkwatson: yes, so then `common` would be better so we don't have more module naming thrash in maven central.

anuraaga: SGTM

bogdandrutu: Done.

PR #2895, opened 2021-02-19
[Breaking Change] Delete old ResourceProvider class

anuraaga:
Doh forgot this last step in the other PR


bogdandrutu (approved):
Is this a breaking change?

anuraaga:
It technically is - removes a class that had no functionality and which we already called out as removed in the release notes. I think it's still good to do a patch release for it though.

jkwatson (approved):

Issue #2736, opened 2021-02-06
Usage of `@Nullable` in the TextMapPropagator seems random


bogdandrutu:
Example:

Why in some APIs carrier is nullable and in others non-null?
```java
  interface Getter<C> {

    /**
     * Returns all the keys in the given carrier.
     *
     * @param carrier carrier of propagation fields, such as an http request.
     * @since 0.10.0
     */
    Iterable<String> keys(C carrier);  // ** Why not here? **

    /**
     * Returns the first value of the given propagation {@code key} or returns {@code null}.
     *
     * @param carrier carrier of propagation fields, such as an http request.
     * @param key the key of the field.
     * @return the first value of the given propagation {@code key} or returns {@code null}.
     */
    @Nullable
    String get(@Nullable C carrier, String key);  // ** Why here? **
  }
```

Why is `carrier` nullable? Can we document if there is a good reason?
```java
  <C> void inject(Context context, @Nullable C carrier, Setter<C> setter);
  <C> Context extract(Context context, @Nullable C carrier, Getter<C> getter);
```

jkwatson:
This was introduced at the request of @Oberon00 to support lambda usages. Additional documentation as to 'why' would be good.

bogdandrutu:
Then I would add that on the `Iterable<String> keys(C carrier);` as well because that seems forgotten

bogdandrutu:
Documentation and maybe change the `keys` API to also support nullable can be done after GA, I think adding the annotation is ABI compatible.

jkwatson:
Looking at this carefully, the `@Nullable` annotation on the `Getter` might not make sense any more, since the `Getter` is no longer a single-method interface, and hence can't be implemented as a lambda.

I actually think that the decision to make the parameter `@Nullable` has resulted in a slightly confusing API. It might be easier just to remove them all from these APIs and let the instrumentation implementation that is doing the propagation work out how it will work best for them. If the instrumentation author knows that they will pass in a null carrier then they can deal with the compiler warnings in their specific cases, and leave this API simpler for the 99% use-case.

bogdandrutu:
@Oberon00 can you please give details if this is still needed?

Oberon00:
[...]

PR #2904, opened 2021-02-22
Don't throw on null in propagator implementations.

anuraaga:
While it's not quite as bad as other parts of the API throwing since adding propagators is explicit, we still want to prevent exceptions in apps as per the otel's error handling rules.

Started looking into these files as part of #2736 and realized that it doesn't seem so bad to just remove the `@Nullable` given all these checks anyways.

### DIFF
--- a/extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/JaegerPropagator.java
+++ b/extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/JaegerPropagator.java
@@ -21,7 +21,6 @@ import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Objects;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.Nullable;
@@ -81,9 +80,13 @@ public final class JaegerPropagator implements TextMapPropagator {
   }
 
   @Override
-  public <C> void inject(Context context, C carrier, TextMapSetter<C> setter) {
-    Objects.requireNonNull(context, "context");
-    Objects.requireNonNull(setter, "setter");
+  public <C> void inject(Context context, @Nullable C carrier, TextMapSetter<C> setter) {
+    if (context == null) {
+      return;
+    }
+    if (setter == null) {
+      return;
+    }
 
     SpanContext spanContext = Span.fromContext(context).getSpanContext();
     if (spanContext.isValid()) {
@@ -93,7 +96,8 @@ public final class JaegerPropagator implements TextMapPropagator {
     injectBaggage(Baggage.fromContext(context), carrier, setter);
   }
 
-  private static <C> void injectSpan(SpanContext spanContext, C carrier, TextMapSetter<C> setter) {
+  private static <C> void injectSpan(
+      SpanContext spanContext, @Nullable C carrier, TextMapSetter<C> setter) {
 
     char[] chars = new char[PROPAGATION_HEADER_SIZE];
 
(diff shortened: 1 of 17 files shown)


carlosalberto (approved):

jkwatson (approved):
Thanks!

extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/JaegerPropagator.java
```diff
@@ -115,14 +119,20 @@ public static JaegerPropagator getInstance() {
...
+      Baggage baggage, @Nullable C carrier, TextMapSetter<C> setter) {
     baggage.forEach(
         (key, baggageEntry) -> setter.set(carrier, BAGGAGE_PREFIX + key, baggageEntry.getValue()));
   }
 
   @Override
   public <C> Context extract(Context context, @Nullable C carrier, TextMapGetter<C> getter) {
-    Objects.requireNonNull(getter, "getter");
+    if (context == null) {
+      return Context.root();
```
bogdandrutu, line 131: should this return `null`? Not sure what is the right behavior here.

jkwatson: not sure either, but not returning null seems like a good thing in general. And, I can definitely understand "you passed in no Context, you get an empty one by default" as a reasonable fallback.

trask: +1 on not returning null

bogdandrutu: But you passed null... it is not like we return null from nowhere

trask: methods that never return null are very nice, because it's one less thing callers need to worry about

if you do want to return null, even in a small case like this, the method return type should be annotated with `@Nullable`

anuraaga: As with all our safety checks, the behavior is undefined so we can pick anything really. I'd lean towards root context since it seems to have the least chance of problems.

bogdandrutu: I think by not returning the same instance of the context we don't respect one of the contract (maybe was just in my mind) that if the propagator fails to extract will return the same instance of the context so we can build a multi propagator that calls individual propagators until one succeeds

bogdandrutu: It looks like our documentation does not suggest that, will check the specs as well

bogdandrutu: Yes, that contract comes from the specs which we don't follow https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#extract

bogdandrutu: I will think a bit about this, but it may be fine if we document just that if the passed context implementation may return `Root`, so we can still implement the logic I said.

But we definitely need to fix our extract contract.

anuraaga: I think by returning `root` we are currently satisfying that spec

`If a value can not be parsed from the carrier, for a cross-cutting concern, the implementation MUST NOT throw an exception and MUST NOT store a new value in the Context, in order to preserve any previously existing valid value.`

We don't throw an exception nor store a new value in the Context since we're returning root. Semantically we can just consider null context and root context the same.

Is the concern addressed by adding `or empty context` in the behavior describing the unparsable here?

https://github.com/open-telemetry/opentelemetry-java/blob/main/api/context/src/main/java/io/opentelemetry/context/propagation/TextMapPropagator.java#L114

jkwatson:
I'm going to merge this as-is. If we decide that returning the root context is not an appropriate fallback, we can change the implementations to match what we decide is the more correct behavior. This is definitely better than throwing exceptions.

PR #2905, opened 2021-02-22
Use super wildcards for forEach consumer.

anuraaga:
I randomly noticed that our `forEach` didn't follow the pattern of Java of consumers accepting `? super`. I'm not sure how big of a deal this is in practice since our types aren't extendible - but it does mean we wouldn't ever be able to do `AttributesMap extends HashMap` which is what I was trying in the first place to run into this.

I believe this is ABI compatible because generics are erased - it's the reason that I couldn't extend HashMap. But appreciate a second opinion. It's API compatible because the bounds are made less, not more, restrictive.

### DIFF
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/AttributesMap.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/AttributesMap.java
@@ -69,7 +69,7 @@ final class AttributesMap implements Attributes {
   @Override
   public void forEach(BiConsumer<? super AttributeKey<?>, ? super Object> consumer) {
     for (Map.Entry<AttributeKey<?>, Object> entry : data.entrySet()) {
-      AttributeKey key = entry.getKey();
+      AttributeKey<?> key = entry.getKey();
       Object value = entry.getValue();
       consumer.accept((AttributeKey<?>) key, value);
     }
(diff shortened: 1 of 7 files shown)


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/AttributesMap.java
```diff
@@ -67,11 +67,11 @@ public boolean isEmpty() {
 
   @SuppressWarnings({"rawtypes", "unchecked"})
   @Override
-  public void forEach(BiConsumer<AttributeKey<?>, Object> consumer) {
+  public void forEach(BiConsumer<? super AttributeKey<?>, ? super Object> consumer) {
     for (Map.Entry<AttributeKey<?>, Object> entry : data.entrySet()) {
       AttributeKey key = entry.getKey();
```
Oberon00, line 72: Why do we use a raw type here?

```suggestion
      AttributeKey<?> key = entry.getKey();
```

jkwatson:
although we *could* make this change now, we don't need it now, and I'd rather make it when we actually need it. Keeping it more restrictive now doesn't seem to have any downsides, from my perspective.

jkwatson:
Important: the `? super String` on the standard acceptors of consumers etc. isn't about allowing extensions to the classes...it's allowing implementations of the consumer to be able to accept *super* classes of String, etc. So, you can write a `BiConsumer<Object, Object>` and it will work with `String, BaggageEntry` (for instance), because both `String` and `BaggageEntry` are also `Object`s. 

So, if we wanted to (for some reason) change our Baggage interface to have a method `forEach(BiConsumer<String, BetterBaggageEntry>)`, we'd have to put the `? super` in there so that existing implementations that were just `BiConsumer<String, BaggageEntry>` to continue to work.

I'm not opposed categorically to this change, and it does make our APIs look more like the built-in ones in the JDK, but I also don't think it's urgent, either.

PR #2906, opened 2021-02-22
Have AttributesMap extend HashMap instead of contain LinkedHashMap

anuraaga:
Contains #2905 

The ordering requirement of attributes was removed by @jkwatson so no need for `Linked`. Extension instead of composition (for this private class) saves us a few more bytes too, though we'd generally avoid that for public classes.

After
```
Benchmark                                                          Mode  Cnt     Score    Error   Units
FillSpanBenchmark.setFourAttributes                               thrpt   60  2106.729 ± 14.989  ops/ms
FillSpanBenchmark.setFourAttributes:·gc.alloc.rate                thrpt   60   791.987 ±  5.626  MB/sec
FillSpanBenchmark.setFourAttributes:·gc.alloc.rate.norm           thrpt   60   592.000 ±  0.001    B/op
FillSpanBenchmark.setFourAttributes:·gc.churn.G1_Eden_Space       thrpt   60   793.685 ± 15.437  MB/sec
FillSpanBenchmark.setFourAttributes:·gc.churn.G1_Eden_Space.norm  thrpt   60   593.209 ± 10.138    B/op
FillSpanBenchmark.setFourAttributes:·gc.churn.G1_Old_Gen          thrpt   60     0.002 ±  0.001  MB/sec
FillSpanBenchmark.setFourAttributes:·gc.churn.G1_Old_Gen.norm     thrpt   60     0.002 ±  0.001    B/op
FillSpanBenchmark.setFourAttributes:·gc.count                     thrpt   60   472.000           counts
FillSpanBenchmark.setFourAttributes:·gc.time                      thrpt   60   180.000               ms
```

Before
```
Benchmark                                                          Mode  Cnt     Score    Error   Units
FillSpanBenchmark.setFourAttributes                               thrpt   60  1936.862 ± 18.334  ops/ms
FillSpanBenchmark.setFourAttributes:·gc.alloc.rate                thrpt   60   806.646 ±  7.654  MB/sec
FillSpanBenchmark.setFourAttributes:·gc.alloc.rate.norm           thrpt   60   656.000 ±  0.001    B/op
FillSpanBenchmark.setFourAttributes:·gc.churn.G1_Eden_Space       thrpt   60   808.658 ± 15.555  MB/sec
FillSpanBenchmark.setFourAttributes:·gc.churn.G1_Eden_Space.norm  thrpt   60   657.776 ± 12.582    B/op
FillSpanBenchmark.setFourAttributes:·gc.churn.G1_Old_Gen          thrpt   60     0.002 ±  0.001  MB/sec
FillSpanBenchmark.setFourAttributes:·gc.churn.G1_Old_Gen.norm     thrpt   60     0.002 ±  0.001    B/op
FillSpanBenchmark.setFourAttributes:·gc.count                     thrpt   60   481.000           counts
FillSpanBenchmark.setFourAttributes:·gc.time                      thrpt   60   181.000               ms
```


build.gradle.kts
```diff
@@ -127,20 +127,22 @@ subprojects {
...
-                        // We suppress the "processing" warning as suggested in
-                        // https://groups.google.com/forum/#!topic/bazel-discuss/_R3A9TJSoPM
-                        "-Xlint:-processing",
-                        // We suppress the "options" warning because it prevents compilation on modern JDKs
-                        "-Xlint:-options",
-
-                        // Fail build on any warning
-                        "-Werror"
-                ))
+                if (name != "jmhCompileGeneratedClasses") {
```
anuraaga, line 130: Needed some maintenance to get JMH working

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/AttributesMap.java
```diff
@@ -52,32 +44,22 @@ int getTotalAddedValues() {
...
       Object value = entry.getValue();
-      consumer.accept(key, value);
+      consumer.accept((AttributeKey<?>) key, value);
     }
   }
 
   @Override
   public Map<AttributeKey<?>, Object> asMap() {
-    return Collections.unmodifiableMap(data);
+    return Collections.unmodifiableMap(this);
```
anuraaga, line 62: IntelliJ flags this as "redundant usage of unmodifiableMap wrapper" but I can't find anything wrong with it, without the wrapper the user would be able to mutate this map

sdk/trace/src/test/java/io/opentelemetry/sdk/trace/SdkSpanBuilderTest.java
```diff
@@ -875,8 +875,7 @@ void spanDataToString() {
                 + "kind=INTERNAL, "
                 + "startEpochNanos=[0-9]+, "
                 + "endEpochNanos=[0-9]+, "
-                + "attributes=AttributesMap\\{data=\\{http.status_code=500, "
-                + "http.url=https://opentelemetry.io}, capacity=128, totalAddedValues=2}, "
+                + "attributes=AttributesMap\\{data=\\{[^}]*}, capacity=128, totalAddedValues=2}, "
```
anuraaga, line 878: Looks the same but ordering is arbitrary now

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/AttributesMap.java
Oberon00: `AttributesMap` implements `Attributes` which is marked `@Immutable`. If that were true, the mapper would indeed be redundant.

anuraaga: Ah thanks for explaining! Yeah this is an implementation detail to support `Attributes.builder().putAll(...)` and not something that leaks into the API, so I guess I just need to add a comment.

Oberon00: Looking at https://errorprone.info/bugpattern/Immutable, `@Immutable` on an interface is supposed to mean that all classes implementing it are `@Immutable`. It does not only mean that mutation through the interface is impossible. So I think we need to remove that annotation from `Attributes`.

anuraaga: Well, we also say explicitly

```
Implementations of this interface *must* be immutable and have well-defined value-based
 * equals/hashCode implementations. If an implementation does not strictly conform to these
 * requirements, behavior of the OpenTelemetry APIs and default SDK cannot be guaranteed.
```

but ignore our own advice because it's an implementation detail that worked ok for us. If it's too weird, we could remove `implements Attributes` and reimplement `putAll` instead, I guess it's not all that different.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/AttributesMap.java
```diff
@@ -9,40 +9,32 @@
...
-/**
- * A map with a fixed capacity that drops attributes when the map gets full.
- *
- * <p>Note: this doesn't implement the Map interface, but behaves very similarly to one.
- */
+/** A map with a fixed capacity that drops attributes when the map gets full. */
 @SuppressWarnings({"rawtypes", "unchecked"})
-final class AttributesMap implements Attributes {
-  private final Map<AttributeKey<?>, Object> data;
+final class AttributesMap extends HashMap<AttributeKey<?>, Object> implements Attributes {
```
jkwatson, line 18: I definitely agree with not using the LinkedHashMap as the implementation any more. But, are we sure that a single allocation per span is worth introducing this antipattern? I know it's not a public class, but it feels like it would be very easy for us to start abusing this wide-open API, leading to much more confusing code.

anuraaga: I'd generally say abuse anything if it's private :) The allocations are not paid by us, but by the apps, so happy to do what I can, to me it feels free. I'm also ok with reimplementing proto serialization if it helps for the same reason, it's in some sense an antipattern but it's all private code and helps apps.

anuraaga: But yeah I think we can copy in putall to this class so in this case I think we can reduce some hackiness without losing performance so that is ideal will try it.

anuraaga: Ok remember why we extend `Attributes` here - since in `toSpanData` we don't actually make an immutable copy if the span has ended already. So we'd need #2905 I think to be able to extend `HashMap` here - it seems nice to skip an allocation where we can though not the end of the world, but if it brings us more inline with JDK (and I guess best practice? I know I often forget about the `? super` deal but have a feeling it's generally what you're supposed to do) may as well?

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/AttributesMap.java
jkwatson: ok, Agrawal. I'll let this one through, but I'm watching you like a hawk for abuses of this antipattern.  👁️ 👁️  🐦

jkwatson (approved):
🐦 👀

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/AttributesMap.java
jkwatson: ok, Agrawal. I'll let this one through, but I'm watching you like a hawk for abuses of this antipattern. 👁️ 👁️ 🐦

PR #2912, opened 2021-02-22
Remove throws statements from the API until we decide how to document it

bogdandrutu:
Before https://github.com/open-telemetry/opentelemetry-java/pull/2881 is merged I think in order to not have "behavior" changes after 1.0 is better to remove all throws statements.


jkwatson (approved):
👍

Issue #2917, opened 2021-02-23
Trace SDK has api dependency on experimental semconv artifact.


Oberon00:
See https://github.com/open-telemetry/opentelemetry-java/blob/v0.17.1/sdk/trace/build.gradle.kts#L14.

In my opinion an API dependency is not acceptable here if we really want to keep the ability to break semconv (which I think we should). Users will accidentally use this experimental package via the transitive dependency of the stable tracing SDK without being aware they are using experimental functionality.

This is more severe than but related to #2914.

jkwatson:
Do we actually expose anything from the semantic conventions in our trace sdk APIs? If not, we can just change this to `implementation`, I think.

Oberon00:
The question is what happens when the user uses an incompatible version of the semantic conventions. Is this prevented by the alpha BOM? I.e. does the alpha BOM include the stable BOM or does one need to declare versions of the stable and alpha BOM independently?

jkwatson:
We do imply in our versioning doc that if you update your SDK, you might also have to update your *runtime* API dependency. This falls firmly into that camp. Our guarantee is that we won't break API users by updates in the SDK. If an API user is using an alpha dependency (like semconv), they should expect to have to update it from time to time, since we don't guarantee backwards compatibility of non-alpha artifacts.

We state this clearly:
```
Changes may be made that require changes to the an app's dependency declarations aside from just
incrementing the version on `MINOR` version updates. 
```

The alpha BOM doesn't not include the main BOM, but artifacts in the alpha bom certainly depend on artifacts in the stable bom, and vice versa.

Issue #2914, opened 2021-02-23
Is it OK for stable trace SDK to have an implementation dependency on metric API?


Oberon00:
The trace SDK is stable but depends on the metric API which is in alpha: https://github.com/open-telemetry/opentelemetry-java/blob/v0.17.1/sdk/trace/build.gradle.kts#L17

It does this to support some internal metrics for the BatchSpanProcessor.

Note that the exposure of internal metrics is of course also not specified yet, see open-telemetry/opentelemetry-specification#959.

I would vote for removing any metric dependencies from the tracing SDK for now. But if not, I think at least a comment is warranted in the gradle file.

anuraaga:
I was going to say it's ok since we require the bom but realized the bom doesn't help with alpha modules.

I suggested removing them before but @jkwatson and @bogdandrutu definitely wanted to keep them and I do understand they're pretty useful metrics.

Oberon00:
Tangential question: If someone uses only the release BOM but not the alpha BOM, how will gradle resolve the version of the metrics API?

anuraaga:
It'll be what's in the pom like any transitive dependency. But since it's not dependency management if another dependency also uses metrics transitively it can end up a different version.

jkwatson:
I think this is 100% acceptable. We're not exposing anything related to metrics on the APIs of the SDK. This is just like if we had a dependency on any library that didn't have the same stability guarantees as our own SDK. changing the internal usage of metrics will not break ABI or API compatibility with anything in the SDK. It is only a potential dependency management issue, which is fine, IMO.

Oberon00:
Since it's only an implementation dependency, I think I agree. Would it still be worth a comment in the gradle file? Like `// Implementation-only dependencies on experimental APIs are OK`.

PR #2918, opened 2021-02-23
Move semconv dependency as implementation, not exposed in the API

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-java/issues/2917


Oberon00:
If there is actually no api dependency here, I think this resolves #2917 completely (according to what we discussed in #2914).

bogdandrutu:
@Oberon00 for sure no "API" dependency, if you want we can resolve that issue, and continue the discussion about implementation dependency on a alpha artifact in the other one.

jkwatson (approved):

PR #2919, opened 2021-02-23
Fix more api dependencies which should be implementation

bogdandrutu:

### DIFF
--- a/integration-tests/build.gradle.kts
+++ b/integration-tests/build.gradle.kts
@@ -8,6 +8,7 @@ extra["moduleName"] = "io.opentelemetry.integration.tests"
 dependencies {
     implementation(project(":sdk:all"))
     implementation(project(":exporters:jaeger"))
+    implementation(project(":semconv"))
 
     implementation("io.grpc:grpc-protobuf")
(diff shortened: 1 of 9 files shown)


sdk/common/build.gradle.kts
```diff
@@ -18,7 +18,8 @@ testSets {
 
 dependencies {
     api(project(":api:all"))
-    api(project(":semconv"))
+
+    implementation(project(":semconv"))
```
Oberon00, line 22: Do we have a dependency here at all?

jkwatson (approved):

integration-tests/build.gradle.kts
```diff
@@ -9,6 +9,7 @@ dependencies {
     implementation(project(":sdk:all"))
     implementation(project(":exporters:jaeger"))
 
+    implementation(project(":semconv"))
```
Oberon00, line 12:
```suggestion
    implementation(project(":semconv"))

```

Oberon00 (approved):

integration-tests/build.gradle.kts
bogdandrutu: Done.

PR #2923, opened 2021-02-24
Histogram Aggregation - core SDK - data

beanliu:
under the suggestion of @bogdandrutu, this PR is splitted from #2581 and adds histogram-related metric data types into the core SDK.

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleHistogramPointData.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleHistogramPointData.java
@@ -20,13 +20,6 @@ import javax.annotation.concurrent.Immutable;
 @AutoValue
 public abstract class DoubleHistogramPointData implements PointData {
   /**
-   * Functional interface for consuming bucket boundaries and counts as a sequence of pair values.
-   */
-  public interface BucketConsumer {
-    void accept(double upperBound, long count);
-  }
-
-  /**
    * Creates a DoubleHistogramPointData.
    *
    * @return a DoubleHistogramPointData.
@@ -80,14 +73,4 @@ public abstract class DoubleHistogramPointData implements PointData {
    * @return the read-only counts in each bucket. <b>do not mutate</b> the returned object.
    */
   public abstract List<Long> getCounts();
-
-  /** Iterates over all the bucket boundaries and counts in this histogram. */
-  public void forEach(BucketConsumer action) {
-    List<Double> boundaries = getBoundaries();
-    List<Long> counts = getCounts();
-    for (int i = 0; i < boundaries.size(); ++i) {
-      action.accept(boundaries.get(i), counts.get(i));
-    }
-    action.accept(Double.POSITIVE_INFINITY, counts.get(boundaries.size()));
-  }
 }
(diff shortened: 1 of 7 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleHistogramPointData.java
```diff
@@ -0,0 +1,93 @@
...
+public abstract class DoubleHistogramPointData implements PointData {
+  /**
+   * Functional interface for consuming bucket boundaries and counts as a sequence of pair values.
+   */
+  public interface BucketConsumer {
+    void accept(double upperBound, long count);
+  }
+
+  /**
+   * Creates a DoubleHistogramPointData.
```
jkwatson, line 30: Can you add documentation here about the relative sizes of the boundaries and counts? And, we should definitely have some validation on creation that the sizes are valid, I think.

jkwatson:
Looks good, aside from one bit of docs & some validation. Thanks for splitting this out!

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleHistogramPointData.java
```diff
@@ -0,0 +1,93 @@
...
+
+  /**
+   * The counts in each bucket.
+   *
+   * @return the read-only counts in each bucket. <b>do not mutate</b> the returned object.
+   */
+  public abstract List<Long> getCounts();
+
+  /** Iterates over all the bucket boundaries and counts in this histogram. */
+  public void forEach(BucketConsumer action) {
```
bogdandrutu, line 85: I think this is unnecessary helper at this moment.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleHistogramPointData.java
```diff
@@ -0,0 +1,93 @@
...
+   */
+  public abstract long getCount();
+
+  /**
+   * The bucket boundaries. For a Histogram with N defined boundaries, e.g, [x, y, z]. There are N+1
+   * counts: [-inf, x), [x, y), [y, z), [z, +inf].
+   *
+   * @return the read-only bucket boundaries in increasing order. <b>do not mutate</b> the returned
+   *     object.
+   */
```
bogdandrutu, line 74: I think we should make this consistent with the latest change in the histogram in the proto https://github.com/open-telemetry/opentelemetry-proto/pull/262

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleHistogramPointData.java
beanliu: make sense, added.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleHistogramPointData.java
beanliu: I wasn't aware of that. thanks for pointing it out. and the comment is updated now.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleHistogramPointData.java
beanliu: removed and will probably add back in the exporter PR.

bogdandrutu (approved):

PR #2924, opened 2021-02-24
Histogram Aggregation - core SDK - aggregator

beanliu:
under the suggestion of @bogdandrutu, this PR is splitted from #2581 and adds histogram aggregator into the core metrics SDK.

benchmark data of the aggregators from my laptop:

JMH version: 1.27
VM version: JDK 13.0.2, OpenJDK 64-Bit Server VM, 13.0.2+8
VM invoker: /Library/Java/JavaVirtualMachines/adoptopenjdk-13.jdk/Contents/Home/bin/java
VM options: -Xmx4G
JMH blackhole mode: full blackhole + dont-inline hint
Warmup: 5 iterations, 1 s each
Measurement: 10 iterations, 1 s each
Timeout: 10 min per iteration
Benchmark mode: Average time, time/op

|Benchmark                                          |Mode | Cnt  |  Score   | Error | Units|
|-------------------                                          |---- | ---  |  ----   | --- | -----|
|DoubleHistogramBenchmark.aggregate_1Threads      |  avgt  | 10 |  29.178 |±  3.746 | ns/op|
|DoubleMinMaxSumCountBenchmark.aggregate_1Threads  | avgt  | 10  | 30.218 |±  1.824  |ns/op|
|LongMinMaxSumCountBenchmark.aggregate_1Threads   |  avgt  | 10 |  26.643| ±  0.571 | ns/op|
|DoubleHistogramBenchmark.aggregate_5Threads      |  avgt  | 10 | 190.119 |± 17.777  |ns/op|
|DoubleMinMaxSumCountBenchmark.aggregate_5Threads   |avgt  | 10 | 226.926| ±  8.637  |ns/op|
|LongMinMaxSumCountBenchmark.aggregate_5Threads  |   avgt  | 10 | 199.217| ±  9.998|  ns/op|
|DoubleHistogramBenchmark.aggregate_10Threads       |avgt |  10 | 342.492| ±  6.906 | ns/op|
|DoubleMinMaxSumCountBenchmark.aggregate_10Threads | avgt  | 10|  457.424 |± 37.492 | ns/op|
|LongMinMaxSumCountBenchmark.aggregate_10Threads  |  avgt  | 10 | 489.890| ±  1.476 | ns/op|


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
```diff
@@ -0,0 +1,153 @@
...
+    return HistogramAccumulation.create(value, ImmutableLongArray.of(1));
+  }
+
+  @Override
+  public HistogramAccumulation accumulateLong(long value) {
+    return HistogramAccumulation.create(value, ImmutableLongArray.of(1));
+  }
+
+  static final class Handle extends AggregatorHandle<HistogramAccumulation> {
+    private final ImmutableDoubleArray boundaries;
```
bogdandrutu, line 80: This is not needed, because it is only internal you can use directly an "unmodifiable" list.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
```diff
@@ -0,0 +1,153 @@
...
+    Handle(ImmutableDoubleArray boundaries) {
+      this.boundaries = boundaries;
+      this.current = new State(this.boundaries.length() + 1);
+    }
+
+    // Benchmark shows that linear search performs better than binary search with ordinary
+    // buckets.
+    private int findBucketIndex(double value) {
+      for (int i = 0; i < boundaries.length(); ++i) {
+        if (Double.compare(value, boundaries.get(i)) <= 0) {
```
bogdandrutu, line 96: I think this needs to be updated after the buckets interval change?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
```diff
@@ -0,0 +1,153 @@
...
+    long[] mergedCounts = new long[x.getCounts().length()];
+    for (int i = 0; i < x.getCounts().length(); ++i) {
+      mergedCounts[i] = x.getCounts().get(i) + y.getCounts().get(i);
+    }
+    return HistogramAccumulation.create(
+        x.getSum() + y.getSum(), ImmutableLongArray.copyOf(mergedCounts));
+  }
+
+  @Override
+  public final MetricData toMetricData(
```
bogdandrutu, line 54: This can be implemented now correct?

beanliu: yep, rebased the branch and added the implementation.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
beanliu: this should conform to the latest definition. the change was amended to this PR before the data one got merged. :)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
beanliu: `boundaries` would be heavily queried. not sure if the (un)boxing overhead could be ignored. benchmark on my machine shows `List` is probably worse than `ImmutableDoubleArray`.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/AggregatorFactory.java
```diff
@@ -77,6 +77,18 @@ static AggregatorFactory minMaxSumCount() {
     return MinMaxSumCountAggregatorFactory.INSTANCE;
   }
 
+  /**
+   * Returns an {@code AggregatorFactory} that calculates an approximation of the distribution of
+   * the measurements taken.
+   *
+   * @param stateful configures if the aggregator is stateful.
```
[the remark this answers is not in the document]
beanliu: `stateful` indicates whether the aggregator is kept after the collection cycle by the `InstrumentProcessor` class. agreed that `temporality` is probably a better option here to make the interface consistent.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
bogdandrutu: Then keep a 'double[]' internally :)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
```diff
@@ -0,0 +1,158 @@
...
+   * merging accumulations with different boundaries.
+   */
+  @Override
+  public final HistogramAccumulation merge(HistogramAccumulation x, HistogramAccumulation y) {
+    long[] mergedCounts = new long[x.getCounts().length()];
+    for (int i = 0; i < x.getCounts().length(); ++i) {
+      mergedCounts[i] = x.getCounts().get(i) + y.getCounts().get(i);
+    }
+    return HistogramAccumulation.create(
+        x.getSum() + y.getSum(), ImmutableLongArray.copyOf(mergedCounts));
```
bogdandrutu, line 50: Same here, as I arguing for removing ImmutableDoubleArray. Since this is internal we can use directly the array (list you pointed is too slow), because we don't expose this publicly so we can ensure that it is immutable.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
beanliu: removed. the plan was to expose public ImmutableArray and use them across data & aggregator package. now guess they are not needed.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
beanliu: removed.

sdk/metrics/src/jmh/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramBenchmark.java
```diff
@@ -0,0 +1,75 @@
...
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Threads;
+import org.openjdk.jmh.annotations.Warmup;
+
+@State(Scope.Benchmark)
+public class DoubleHistogramBenchmark {
```
bogdandrutu, line 27: Nice benchmarks, can you post the results in the PR description?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
```diff
@@ -0,0 +1,166 @@
...
+  @Override
+  public final MetricData toMetricData(
+      Map<Labels, HistogramAccumulation> accumulationByLabels,
+      long startEpochNanos,
+      long lastCollectionEpoch,
+      long epochNanos) {
+    List<Double> boundaries = new ArrayList<>(this.boundaries.length);
+    for (double v : this.boundaries) {
+      boundaries.add(v);
+    }
```
bogdandrutu, line 65: You can cache this in the class.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
```diff
@@ -0,0 +1,166 @@
...
+    this.boundaries = boundaries;
+  }
+
+  @Override
+  public AggregatorHandle<HistogramAccumulation> createHandle() {
+    return new Handle(this.boundaries);
+  }
+
+  /**
+   * Return the result of the merge of two histogram accumulations. As long as one Aggregator
```
bogdandrutu, line 43: Good comment, but needs tests :D see the bug I mentioned in the `accumulateDouble` :). I agree we should not verify that, but we need to test that all Accumulation are the same expected size.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/HistogramAggregatorFactory.java
```diff
@@ -0,0 +1,53 @@
...
+import io.opentelemetry.sdk.metrics.common.InstrumentDescriptor;
+import io.opentelemetry.sdk.metrics.data.AggregationTemporality;
+import io.opentelemetry.sdk.resources.Resource;
+
+final class HistogramAggregatorFactory implements AggregatorFactory {
+  private final double[] boundaries;
+  private final AggregationTemporality temporality;
+
+  HistogramAggregatorFactory(double[] boundaries, AggregationTemporality temporality) {
+    this.boundaries = boundaries;
```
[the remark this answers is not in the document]
beanliu: good catch, I missed that while removing the ImmutableArray :) will fix.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
```diff
@@ -0,0 +1,166 @@
...
+            MetricDataUtils.toDoubleHistogramPointList(
+                accumulationByLabels,
+                isStateful() ? startEpochNanos : lastCollectionEpoch,
+                epochNanos,
+                boundaries)));
+  }
+
+  @Override
+  public HistogramAccumulation accumulateDouble(double value) {
+    return HistogramAccumulation.create(value, countsOfOne);
```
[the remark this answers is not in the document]
beanliu: the decision was an arbitrary one because I'm not sure if histograms would work with AsyncInstruments/Observer and there are not any builders for that if I remember correctly. 

> I mean if the boundaries are {1, 3} and a record of 2 then we need {0, 1, 0} counts correct?

I think this is a better default. but do we really need to implement `accumulateDouble`/`accumulateLong` at all?

sdk/metrics/src/jmh/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramBenchmark.java
beanliu: updated in the desc

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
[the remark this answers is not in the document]
bogdandrutu: That is a good question, but probably will need a small design improvement to allow Aggregator to implement/support only the sync instruments.

I think before solving that we should do the correct thing.

sdk/metrics/src/jmh/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramBenchmark.java
```diff
@@ -0,0 +1,75 @@
...
+    aggregatorHandle = aggregator.createHandle();
+  }
+
+  @Benchmark
+  @Fork(1)
+  @Warmup(iterations = 5, time = 1)
+  @Measurement(iterations = 10, time = 1)
+  @OutputTimeUnit(TimeUnit.MILLISECONDS)
+  @Threads(value = 10)
+  public void aggregate_10Threads() {
```
bogdandrutu, line 52: Sorry, but I am very bad of reading results in ops/ms. Can we use everywhere:

```java
  @BenchmarkMode({Mode.AverageTime})
  @OutputTimeUnit(TimeUnit.NANOSECONDS)
```

This will print `ns/op` :)

bogdandrutu: And if possible update the results :)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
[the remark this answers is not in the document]
beanliu: sounds good. updated.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
beanliu: test added.

sdk/metrics/src/jmh/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramBenchmark.java
beanliu: updated to use `ns/op`.

bogdandrutu: Not really "Benchmark mode: Throughput, ops/time" :)

beanliu: oh, I missed that. 🤦  probably need a cup of ☕ . will update soon

bogdandrutu (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/MetricDataUtils.java
```diff
@@ -44,4 +45,23 @@ private MetricDataUtils() {}
...
     return points;
   }
+
+  static List<DoubleHistogramPointData> toDoubleHistogramPointList(
+      Map<Labels, HistogramAccumulation> accumulationMap,
+      long startEpochNanos,
+      long epochNanos,
+      List<Double> boundaries) {
+    List<DoubleHistogramPointData> points = new ArrayList<>(accumulationMap.size());
+    accumulationMap.forEach(
```
bogdandrutu, line 55: As an exercise can you write this without lambda :)))

beanliu: seems that all `MetricDataUtils.toXyzPointList` use the same `forEach` pattern with lambdas. I think this may break the consistency if just rewrite one of them?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/AggregatorFactory.java
```diff
@@ -77,6 +77,18 @@ static AggregatorFactory minMaxSumCount() {
...
 
+  /**
+   * Returns an {@code AggregatorFactory} that calculates an approximation of the distribution of
+   * the measurements taken.
+   *
+   * @param temporality configures what temporality to be produced for the Histogram metrics.
+   * @param boundaries configures the fixed bucket boundaries.
+   * @return an {@code AggregationFactory} that calculates histogram of recorded measurements.
+   */
+  static AggregatorFactory histogram(double[] boundaries, AggregationTemporality temporality) {
```
bogdandrutu, line 88: Was thinking about this API, probably better to accept `List<Double>` here, and internally use the same array. Idea being that List is a bit more natural as an input param than double[]. What do you think? @jkwatson what do you think?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
```diff
@@ -0,0 +1,158 @@
...
+  @Override
+  public HistogramAccumulation accumulateLong(long value) {
+    return accumulateDouble((double) value);
+  }
+
+  // Benchmark shows that linear search performs better than binary search with ordinary
+  // buckets.
+  private static int findBucketIndex(double[] boundaries, double value) {
+    for (int i = 0; i < boundaries.length; ++i) {
+      if (Double.compare(value, boundaries[i]) <= 0) {
```
anuraaga, line 101:
```suggestion
      if (value <= boundaries[i]) {
```

Do we have to handle NAN? If so, I'd handle it specially, it's not clear to me what it means in the case of a histogram like this. I suspect we'd want the `MeterProvider` to filter out NANs and we use simple comparisons here though.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
```diff
@@ -0,0 +1,158 @@
...
+    // read-only
+    private final double[] boundaries;
+
+    @GuardedBy("lock")
+    private double sum;
+
+    @GuardedBy("lock")
+    private final long[] counts;
+
+    private final ReentrantLock lock = new ReentrantLock();
```
anuraaga, line 118:
```suggestion
    private final Object mutex = new Object();
```

anuraaga: Can you update below to use `synchronized(mutex)`? It's the pattern we use everywhere else so we should be consistent, and I think it might be a bit faster. We'd only use `ReentrantLock` if we used any of the methods other than `lock` / `unlock`

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/HistogramAggregatorFactory.java
```diff
@@ -0,0 +1,54 @@
...
+final class HistogramAggregatorFactory implements AggregatorFactory {
+  private final double[] boundaries;
+  private final AggregationTemporality temporality;
+
+  HistogramAggregatorFactory(double[] boundaries, AggregationTemporality temporality) {
+    this.boundaries = Arrays.copyOf(boundaries, boundaries.length);
+    this.temporality = temporality;
+
+    for (int i = 1; i < this.boundaries.length; ++i) {
+      if (Double.compare(this.boundaries[i - 1], this.boundaries[i]) >= 0) {
```
anuraaga, line 23:
```suggestion
      if (this.boundaries[i - 1] >= this.boundaries[i]) {
```

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/HistogramAggregatorFactory.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Arrays;
+
+final class HistogramAggregatorFactory implements AggregatorFactory {
+  private final double[] boundaries;
+  private final AggregationTemporality temporality;
+
+  HistogramAggregatorFactory(double[] boundaries, AggregationTemporality temporality) {
+    this.boundaries = Arrays.copyOf(boundaries, boundaries.length);
+    this.temporality = temporality;
+
```
anuraaga, line 21: Could validate that no argument is NAN

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/MetricDataUtils.java
anuraaga: `forEach` is nice for maps, like here, due to the tedium of entrySet. We should never use it with lists though since normal for is basically the same.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/aggregator/AggregatorFactoryTest.java
```diff
@@ -123,4 +124,73 @@ void getSumAggregatorFactory() {
...
+                    InstrumentDescriptor.create(
+                        "name",
+                        "description",
+                        "unit",
+                        InstrumentType.VALUE_RECORDER,
+                        InstrumentValueType.DOUBLE))
+                .isStateful())
+        .isTrue();
+
+    Assertions.assertThrows(
```
anuraaga, line 182: Oh looks like we have a few left in `MetricDataTest` but we've switched to just using assertj, can you use `assertThatThrownBy(() ->).isInstanceOf()`? And we like to follow up with `hasMessage` or `hasMessageContaining` with a light assertion on the message.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregatorTest.java
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  @Test
+  void testMultithreadedUpdates() throws Exception {
+    final AggregatorHandle<HistogramAccumulation> aggregatorHandle = aggregator.createHandle();
+    final Histogram summarizer = new Histogram();
+    int numberOfThreads = 10;
+    final long[] updates = new long[] {1, 2, 3, 5, 7, 11, 13, 17, 19, 23};
+    final int numberOfUpdates = 1000;
+    final CountDownLatch startingGun = new CountDownLatch(numberOfThreads);
```
anuraaga, line 121:
```suggestion
    final CountDownLatch starter = new CountDownLatch(numberOfThreads);
```

Let's avoid words like gun

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregatorTest.java
```diff
@@ -0,0 +1,181 @@
...
+    List<Thread> workers = new ArrayList<>();
```
anuraaga, line 122: Not a big deal but I'd generally use a threadpool executor instead of managing threads myself, should be a bit simpler code.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregatorTest.java
```diff
@@ -0,0 +1,181 @@
...
+    }
+    // make sure everything gets merged when all the aggregation is done.
+    summarizer.process(aggregatorHandle.accumulateThenReset());
+
+    assertThat(summarizer.accumulation)
+        .isEqualTo(HistogramAccumulation.create(101000, new long[] {5000, 5000, 0, 0}));
+  }
+
+  private static final class Histogram {
+    private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
```
anuraaga, line 159: Simple mutex here too (we're not taking a readlock anywhere so have no advantage of using this special lock)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
anuraaga: To clarify, by "everywhere else" I generally refer to non-metrics since I haven't looked at the metrics code so much. But we should get it closer to our tracing code in these aspects

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/AggregatorFactory.java
jkwatson: No strong opinion on this one. Either way, we should absolutely make a defensive copy of the data. Certainly I have no objection to using a List<Double> for the input boundaries here.

bogdandrutu: We do the defense copy already :)

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregatorTest.java
bogdandrutu: https://www.loc.gov/law/help/usconlaw/second-amendment.php#:~:text=The%20Second%20Amendment%2C%20one%20of,self%2Devident%2C%20and%20has%20given

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
beanliu: I don't see any contracts about `NaN` in the specifications so just being defensive here.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/DoubleHistogramAggregator.java
beanliu: mutex benchmark:
|Benchmark                                          |Mode  |Cnt     |Score     |Error  |Units|
|---                                                       |---      |---       |---          |---     |---|
|DoubleHistogramBenchmark.aggregate_10Threads       |avgt   |10  |1210.696 |±  50.972  |ns/op|
|DoubleHistogramBenchmark.aggregate_1Threads       | avgt  | 10     |9.997 |±   0.448  |ns/op|
|DoubleHistogramBenchmark.aggregate_5Threads        |avgt  | 10  | 696.606| ± 123.858  |ns/op|

lock benchmark:
|Benchmark                                          |Mode  |Cnt     |Score     |Error  |Units|
|---                                                       |---      |---       |---          |---     |---|
|DoubleHistogramBenchmark.aggregate_10Threads       |avgt   |10  |  371.369 |± 8.39 | ns/op|
|DoubleHistogramBenchmark.aggregate_1Threads       |avgt   |10  |   24.045 |± 0.648|  ns/op|
|DoubleHistogramBenchmark.aggregate_5Threads       |avgt   |10  | 187.410| ± 0.863 | ns/op|

the `mutext` approach works better only in a single-threaded environment. not sure if we should use `mutex` here.

anuraaga: Hmm - so many articles say they should have at least the same performance but obviously there are some traps too. `ReentrantLock` seems to make sense here

anuraaga (approved):

jkwatson (approved):
Thanks!

PR #2927, opened 2021-02-24
Add event assertion, full attributes assertion, and fuzzy attribute v…

anuraaga:
…alue assertion.

While comparing parity between the Groovy assertions we use in instrumentation now, I found we need at least

- Event assertion
- Assertion of all attributes (i.e., to assert there are not more attributes than expected)
- Assertion of attribute value as a `Consumer`, e.g., to very an attribute appears to be a thread ID even though we wouldn't assert the actual value


jkwatson (approved):
Cool. Thanks!

PR #2928, opened 2021-02-24
Defend against null in a few more places.

anuraaga:


api/all/src/main/java/io/opentelemetry/api/trace/Span.java
```diff
@@ -51,6 +54,9 @@ static Span fromContext(Context context) {
    */
   @Nullable
   static Span fromContextOrNull(Context context) {
```
Oberon00, line 56: Huh, should this be a public function?

jkwatson: This was discussed in our detailed API examination over the past couple weeks, and it was discovered it was being used by people as a way to detect if there was any span at all in the context, which the other method doesn't provide a way to do. So, yes, this was deemed necessary.  Thanks for asking the question, though!

jkwatson (approved):
Thanks!

PR #2934, opened 2021-02-25
add environment variables to ConfigProperties last

cmuelhau:
switching env variables with system properties gives them precedence and allows for more general configurability

### DIFF
--- a/sdk-extensions/autoconfigure/build.gradle.kts
+++ b/sdk-extensions/autoconfigure/build.gradle.kts
@@ -117,12 +117,12 @@ tasks {
 
     val testResourceDisabledByProperty by existing(Test::class) {
         jvmArgs("-Dotel.java.disabled.resource-providers=io.opentelemetry.sdk.extension.resources.OsResourceProvider,io.opentelemetry.sdk.extension.resources.ProcessResourceProvider")
-        // Properties win, this is ignored.
-        environment("OTEL_JAVA_DISABLED_RESOURCE_PROVIDERS", "io.opentelemetry.sdk.extension.resources.ProcessRuntimeResourceProvider")
     }
 
     val testResourceDisabledByEnv by existing(Test::class) {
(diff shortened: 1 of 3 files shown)


jkwatson:
Hi @cmuelhau . We had very deliberately chosen this order to match the behavior of the otel javaagent. There are arguments for each choice being better, so we stuck with the one that the javaagent had chosen.

cmuelhau:
> Hi @cmuelhau . We had very deliberately chosen this order to match the behavior of the otel javaagent. There are arguments for each choice being better, so we stuck with the one that the javaagent had chosen.

Hi @jkwatson  Thanks for taking the time to reply ;) I understand, just for some background: Our use case are java applications in docker images, preconfigured with jvm flags, so that the otel java agent is attached and traces/metrics exporter set to `none`. That would allow for specific env variables per otel agent configuration (where env variables are platform-agnostic).
As I see it, my option now would be to use JAVA_TOOL_OPTIONS to set all jvm flags (attach the javaagent and set system properties) or would you have another approach how this could be done? One can always hope ;)

jkwatson:
Definitely. This choice is definitely non-optimal for docker users who want to bake command line args into images and override via env vars. The autoconfiguration module is still marked as `alpha`, so there is definitely a chance to revisit this decision, but we would want to keep it the same between the javaagent and the SDK autoconfiguration mechanisms. 

Can you create an issue or discussion in the opentelemetry-java-instrumentation project to raise this issue? I do think there's still time to adjust if general agreement could be had.

For what it's worth, the reason we decided to go with this precedence order was to match the way that spring-boot does it: https://docs.spring.io/spring-boot/docs/2.1.18.RELEASE/reference/html/boot-features-external-config.html, which seemed as good a reason as any.

cmuelhau:
Alright, I'll follow up with an issue on opentelemetry-java-instrumentation, thanks!

cmuelhau:
While writing the issue, I came up with the following solution (I will also describe it in the issue, as this may or may not be feasible for everyone)
Basically, the agent is still attached with the entrypoint, but instead of using system properties to preconfigure, I'm using the Dockerfile ENV instruction to preset the respective environment variables. Those can then be overridden by the runtime environment variables.

PR #2936, opened 2021-02-26
Gracefully shut down exporter after it exports the remaining metrics

shaochengwang:
**Description:**
Currently inside `IntervalMetricReader.java`, when `shutdown()` is called, it is supposed to [*stop the scheduled task and call export one more time.*](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java#L35-L49). However, [`exporter.run()`](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java#L83-L105) is an asynchronous method, it will return without waiting for all the metrics are exported. Then [`exporter.shutdown()`](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java#L107-L109) will be called and shutdown the channel. As a result, an exception will be thrown and the remaining metrics are not exported.

I try to fix this issue by adding a `CountDownLatch` to ensure the last `export` is complete when exporter is being shutdown.

**Tests:**
Tested with [aws-otel-collector](https://github.com/aws-observability/aws-otel-collector/)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
```diff
@@ -82,6 +84,7 @@ private Exporter(InternalState internalState) {
     @SuppressWarnings("BooleanParameter")
     public void run() {
```
anuraaga, line 85: It looks like it'd be better to refactor to introduce `CompletableResultCode doRun` and call `join` when shutting down. I suspect much would change when introducing `forceFlush` to metrics but in the meantime that seems like a reasonable workaround.

jkwatson: I agree, I think we should use CompletableResultCode here to match the behavior on the tracing side.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
```diff
@@ -37,7 +38,10 @@ public void shutdown() {
     scheduler.shutdown();
     try {
       scheduler.awaitTermination(5, TimeUnit.SECONDS);
-      exporter.run();
+      CompletableResultCode result = exporter.forceFlush();
+      if (result != null) {
```
jkwatson, line 42: Can we make this return value non-null, so we don't have to do this check? Also, I'd like to have the `IntervalMetricReader.shutdown()` method also return `CompleteableResultCode` as well.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
```diff
@@ -104,6 +111,11 @@ public void run() {
       }
     }
 
+    CompletableResultCode forceFlush() {
+      run();
+      return exportResult.get();
```
jkwatson, line 116: there's a race here. `run()` could finish before the result is accessed, in which case this will return a null, which I think is a bad API. Take a look at the implementation in the BatchSpanProcessor for a clean way to deal with this race.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
shaochengwang: Thanks. will make the change.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
jkwatson: @anuraaga your suggestion was to create a `doRun` method that returned `CompleteableResultCode` so that the `shutdown` implementation can join that result code, right? Otherwise, this `run` method is async if the exporter is async and a proper orderly shutdown couldn't be implemented.

shaochengwang: I see what you mean here. Thanks for the explanation. :)

jkwatson: Can't this `run()` method just call `doRun()` and ignore the result, rather than duplicating all the code?

jkwatson: or maybe not ignore the result, but `join` it, so that this remains synchronous, I guess.

shaochengwang: I will let this `run()` directly call `doRun()` and ignore the result so that `run()` could keep asynchronous. I will let `shutdown()` call `doRun()` and `join()` the result to make sure it's an orderly shutdown.

jkwatson (approved):
Thanks!

jkwatson:
@anuraaga anything else you want to see here, or are we good to merge?

anuraaga (approved):
Oops thanks!

PR #2937, opened 2021-02-26
Returns invalid traceflags for a bad string parse instead of throwing…

anuraaga:
… an exception.

Supercedes #2869 

This removes the one and only place in the API that still declares it throws an exception. It avoids returning `null`, as in #2869, to reduce the chance of problems with usage of the API, an invalid traceflags should be handled by a propagator. But in the chance it's not handled, it will act as an empty traceflags and not cause crashes.

This is a behavior change which I agree with @bogdandrutu we need to make now or never since, i.e. a propagator that relied on the exceptions to handle invalid would not do so anymore if we removed them later. As `TraceFlags` presumably has no real use case outside of the w3c propagator, I'm comfortable with squeezing this into the release but open to thoughts.

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/trace/SpanContext.java
+++ b/api/all/src/main/java/io/opentelemetry/api/trace/SpanContext.java
@@ -135,7 +135,9 @@ public interface SpanContext {
    * @return {@code true} if this {@code SpanContext} is valid.
    */
   default boolean isValid() {
-    return TraceId.isValid(getTraceId()) && SpanId.isValid(getSpanId());
+    return TraceId.isValid(getTraceId())
+        && SpanId.isValid(getSpanId())
+        && getTraceFlags().isValid();
   }
 
   /**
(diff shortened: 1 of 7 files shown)


api/all/src/main/java/io/opentelemetry/api/trace/SpanContext.java
```diff
@@ -135,7 +135,9 @@ default boolean isSampled() {
    * @return {@code true} if this {@code SpanContext} is valid.
    */
   default boolean isValid() {
-    return TraceId.isValid(getTraceId()) && SpanId.isValid(getSpanId());
+    return TraceId.isValid(getTraceId())
+        && SpanId.isValid(getSpanId())
+        && getTraceFlags().isValid();
```
jkwatson, line 140: this is a change from the specification, I believe, which says that the SpanContext is valid iff the spanid and traceid are valid, but doesn't mention trace flags.

Oberon00:
I think introducing the concept of invalid trace flags is something that should go through the spec. Compare also https://github.com/open-telemetry/opentelemetry-specification/issues/753

PR #2939, opened 2021-02-26
Handle nulls in noop / multi

### LINKED ISSUE #2921: Verify/update all propagators to make sure they do not throw exceptions on bad inputs.
Ideally, all of the propagators should have the same behavior in the face of bad inputs.

extract: return the root Context if null inputs are provided for Context or Setter
inject: return immediately if null inputs are provided for Context or Getter

- [x] B3Propagators
- [x] W3CBaggagePropagator
- [x] W3CTraceContextPropagator
- [x] JaegerPropagator
- [x] OtTracePropagator
- [x] AwsXrayPropagator
- [ ] MultiTextMapPropagator
- [ ] NoOpTextMapPropagator

anuraaga:
Fixes #2921

This may not be a good idea for Multi but went with consistency for now, let me know thoughts.


jkwatson (approved):
Thanks!

api/context/src/main/java/io/opentelemetry/context/propagation/MultiTextMapPropagator.java
```diff
@@ -45,13 +45,22 @@
 
   @Override
   public <C> void inject(Context context, @Nullable C carrier, TextMapSetter<C> setter) {
+    if (context == null || setter == null) {
```
Oberon00, line 48: I think letting the null bubble through to the inner propagators would be better for inject.

api/context/src/main/java/io/opentelemetry/context/propagation/MultiTextMapPropagator.java
```diff
@@ -45,13 +45,22 @@
...
   public <C> Context extract(Context context, @Nullable C carrier, TextMapGetter<C> getter) {
+    if (context == null) {
+      return Context.root();
+    }
+    if (getter == null) {
+      return context;
+    }
     for (TextMapPropagator textPropagator : textPropagators) {
       context = textPropagator.extract(context, carrier, getter);
     }
```
Oberon00, line 66: As said [above](https://github.com/open-telemetry/opentelemetry-java/pull/2939/files#r583761460), I would not guard against null before this loop also here in extract, but only ensure that we don't return `null` even if `null` was passed in as `context` (and `textPropagators` was possibly empty).

api/context/src/main/java/io/opentelemetry/context/propagation/MultiTextMapPropagator.java
jkwatson: why is this better?

Oberon00: Mostly just a gut feeling, and principle of least surprise. Also the Javadoc in https://github.com/open-telemetry/opentelemetry-java/blob/v0.17.1/api/context/src/main/java/io/opentelemetry/context/propagation/TextMapPropagator.java#L46-L53 `TextMapPropagator.composite()` states that it "simply delegates injection and extraction", so that's what I would do.

jkwatson: Javadoc can certainly be updated. My opinion is that this is the safest approach, so that propagators that are not maintained by us will also be guarded against these bad inputs if they are wrapped up in a composite propagator with other "official" propagators.

bogdandrutu (dismissed):
Per offline discussion with @jkwatson block this until the release

api/context/src/main/java/io/opentelemetry/context/propagation/MultiTextMapPropagator.java
Oberon00: 👍 I think that makes sense. If a user wants the other behavior, they can implement their own "NullNaiveMultiPropagator".

jkwatson:
Let's continue the discussion after we release 1.0.0. Whatever we decide can go into the next minor release.

jkwatson:
> Per offline discussion with @jkwatson block this until the release

Can we unblock this now?

jkwatson:
> > Per offline discussion with @jkwatson block this until the release
> 
> Can we unblock this now?

How about now, @bogdandrutu ?

jkwatson:
@anuraaga looks like this needs a rebase.

PR #2943, opened 2021-02-26
Changelog updates for the 1.0.0 release.

jkwatson:


CHANGELOG.md
```diff
@@ -2,6 +2,27 @@
 
 ## Unreleased:
 
+## Version 1.0.0 - 2021-02-26
+
+### General
```
bogdandrutu, line 7: Would be good to maybe have a link to the stable packages?

jkwatson: good idea, yes... can't really add that until we do a publish and we can point at a published bom/pom, though, i think.

bogdandrutu: Ok, probably we can do it in the release notes. Or we can have an entry to the stable bom in the readme :) Can be done later.

bogdandrutu (approved):

CHANGELOG.md
jkwatson: I added it here.

Issue #2948, opened 2021-02-26
ImmutableBaggage.isKeyValid allows non-zero-length strings that contain nothing but spaces and/or tabs


jimshowalter:
It currently is:

`  private static boolean isKeyValid(String name) {
    return name != null && !name.isEmpty() && StringUtils.isPrintableString(name);
  }`

It needs to be:
`  private static boolean isKeyValid(String name) {
    return name != null && !name.trim().isEmpty() && StringUtils.isPrintableString(name);
  }`

Or it needs to use something like apache StringUtils isBlank.

akhatami:
@jimshowalter @jkwatson 

May I start working on this?

bogdandrutu:
@jkwatson I think this should be clarified in the specification because we want to have a consistent behavior across libraries

jkwatson:
> @jkwatson I think this should be clarified in the specification because we want to have a consistent behavior across libraries

Isn't key validity defined by the w3c spec already?

jimshowalter:
https://www.w3.org/TR/baggage/#key

"Leading and trailing whitespaces (OWS) are allowed but MUST be trimmed when converting the header into a data structure."

If you trim leading and trailing whitespace on a string consisting of nothing but whitespace, you get an empty string.

So !name.trim().isEmpty() is correct, and the current code is not correct.

bogdandrutu:
If we apply the w3c rules then we need to trim the keys always, for example " too" should overwrite "too". I want to clarify in the specs that we follow w3c rules which I think we don't say at the moment

jimshowalter:
The proposed change in my first post in this bug report was to change !name.isEmpty() to !name.trim().isEmpty(), which always trims the keys.

jkwatson:
@bogdandrutu Can you shepherd a tweak to the specs so we can get #2950 merged?

andresbeckruiz:
Has this issue been resolved? If not I would like to take a stab at it.

jkwatson:
> Has this issue been resolved? If not I would like to take a stab at it.
> 
> cc: @alolita

I think we're still waiting on the specification issue to get resolved.

aniketrb-github:
any updates on the spec. ?

MounikaNandyala:
Is it still open? 
If yes, we can try this way.
 
 private static boolean isKeyValid(String name) { 
// Trim leading and trailing whitespace
  name = name.trim();

// Check if the name is valid (non-empty after trimming)
[...]

PR #2950, opened 2021-02-26
Prevent ImmutableBaggage.isKeyValid from allowing non-zero-length strings that contain nothing but spaces and/or tabs (#2948)

akhatami:
Prevent ImmutableBaggage.isKeyValid from allowing non-zero-length strings that contain nothing but spaces and/or tabs (#2948)

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/baggage/ImmutableBaggage.java
+++ b/api/all/src/main/java/io/opentelemetry/api/baggage/ImmutableBaggage.java
@@ -94,7 +94,7 @@ final class ImmutableBaggage extends ImmutableKeyValuePairs<String, BaggageEntry
    * @return whether the name is valid.
    */
   private static boolean isKeyValid(String name) {
-    return name != null && !name.isEmpty() && StringUtils.isPrintableString(name);
+    return name != null && !name.trim().isEmpty() && StringUtils.isPrintableString(name);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


api/all/src/main/java/io/opentelemetry/api/baggage/ImmutableBaggage.java
```diff
@@ -94,7 +94,7 @@ public Baggage build() {
    * @return whether the name is valid.
    */
   private static boolean isKeyValid(String name) {
-    return name != null && !name.isEmpty() && StringUtils.isPrintableString(name);
+    return name != null && !name.trim().isEmpty() && StringUtils.isPrintableString(name);
```
jkwatson, line 97: can you add a unit test for this? Thanks!

akhatami: @jkwatson 

I added a unit test for it, but not sure if it's OK (it's my first collaboration on this project). Please tell me if I should do anything else. Thanks!

bogdandrutu: this is an unnecessary allocation with "trim". Can we avoid that?

bogdandrutu: If "   " is invalid, then "  t" means is equal with "t" ?

bogdandrutu: I think we will adopt the w3c standard and will be https://w3c.github.io/baggage/#key

jkwatson:
I'm going to close this as the author has not signed the CLA. Please re-open if/when appropriate.

Issue #1810, opened 2020-10-15
MetricsProcessor


as-polyakov:
**Is your feature request related to a problem? Please describe.**
Similar to SpanProcessor, we need a hook for metrics which would allow users to intercept metrics lifecycle eventы with their own logic.

**Describe the solution you'd like**

```
public interface MetricsProcessor {

  /**
   * Called when bind() method is called. Allows to manipulate labels which this
   * instrument is bound to. Particular use case includes enriching lables and/or adding more labels
   * depending on the Context
   * @param ctx context of the operation
   * @param descriptor instrument descriptor
   * @param labels immutable labels. When processors are chained output labels of the previous one is passed as
   * an input to the next one. Last labels returned by a chain of processors are used for bind() operation.
   * @return labels to be used as an input to the next processor in chain or bind() operation if this is the last processor
   */
    Labels onLabelsBound(Context ctx, InstrumentDescriptor descriptor, Labels labels);

  /**
   *  Called when .record() method is called. Allows to manipulate recorded value. When chained input of the next
   *  call is the output of the previous call. Final output is recorded
   * @param ctx context of the operation
   * @param descriptor instrument descriptor
   * @param labels immutable labels. When processors are chained output labels of the previous one is passed as
   * @param value recorded value
   * @return value to be used as an input to the next processor in chain or record() operation if this is the last processor
   */

    long onLongMeasurement(Context ctx, InstrumentDescriptor descriptor, Labels labels, long value);

    double onDoubleMeasurement(Context ctx, InstrumentDescriptor descriptor, Labels labels, double value);
}
```

**Describe alternatives you've considered**
Main problem I am trying to address - is to be able to add certain labels to metrics depending on Context (i.e. dynamically). One alternative can be to just do that and make it a standalone feature aka "sticky labels". Had a simple PR for this [here](https://github.com/parallelstream/opentelemetry-java/pull/1). However MetricsProcessor is a more generic solution which can address other use cases.
As for the implementation, instead of using immutable Labels we can pass ReadWriteLables object, this would make it follow the same patter as in SpanProcessor

**Additional context**
There's a problem with this approach - statically bound instruments will not be able to dynamically add labels during record operation. Since we only intercept onLabelsBound() in case one statically binds an instrument and then uses that to record multiple measurements with different contexts, only first bind call will be intercepted

jkwatson:
[...]

PR #2956, opened 2021-03-01
introducing view

parallelstream:
- View class as an overarching concept for AggregationFactory. In the future we can add more specifics into it
- The grand idea here is basically based around OpenCensus Views. We want to be able to register custom projections (views) on the metrics which would allow us to use different aggregations and pre-processing techniques.
- This is the very first step where we just introduce a view and wrap existing AggregatorFacotry
- Immediately next we would like to have LabelsProcessors added to the views which are going to be a separate PR, but at a glance - same thing we discussed in https://github.com/open-telemetry/opentelemetry-java/issues/1810. Whole PR might be found here https://github.com/open-telemetry/opentelemetry-java/pull/2800

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
@@ -95,11 +95,11 @@ public final class SdkMeterProvider implements MeterProvider, MetricProducer {
    *   .buildInstrument();
    *
    * // create a specification of how you want the metrics aggregated:
-   * AggregationFactory aggregationFactory = AggregationFactory.minMaxSumCount();
+   * AggregatorFactory aggregatorFactory = AggregatorFactory.minMaxSumCount();
    *
    * //register the view with the MeterSdkProvider
    * meterProvider.registerView(instrumentSelector, View.builder()
-   *   .setAggregatorFactory(factory).build());
+   *   .setAggregatorFactory(aggregatorFactory).build());
    * }</pre>
    */
   public void registerView(InstrumentSelector selector, View view) {
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
@@ -9,10 +9,10 @@ import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.api.metrics.MeterProvider;
 import io.opentelemetry.sdk.common.Clock;
 import io.opentelemetry.sdk.internal.ComponentRegistry;
-import io.opentelemetry.sdk.metrics.aggregator.AggregatorFactory;
 import io.opentelemetry.sdk.metrics.data.MetricData;
 import io.opentelemetry.sdk.metrics.export.MetricProducer;
 import io.opentelemetry.sdk.metrics.view.InstrumentSelector;
+import io.opentelemetry.sdk.metrics.view.View;
 import io.opentelemetry.sdk.resources.Resource;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -98,10 +98,11 @@ public final class SdkMeterProvider implements MeterProvider, MetricProducer {
    * AggregationFactory aggregationFactory = AggregationFactory.minMaxSumCount();
    *
    * //register the view with the MeterSdkProvider
-   * meterProvider.registerView(instrumentSelector, aggregationFactory);
+   * meterProvider.registerView(instrumentSelector, View.builder()
+   *   .setAggregatorFactory(factory).build());
    * }</pre>
    */
-  public void registerView(InstrumentSelector selector, AggregatorFactory aggregatorFactory) {
-    sharedState.getViewRegistry().registerView(selector, aggregatorFactory);
+  public void registerView(InstrumentSelector selector, View view) {
+    sharedState.getViewRegistry().registerView(selector, view);
   }
 }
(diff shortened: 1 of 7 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/View.java
```diff
@@ -0,0 +1,27 @@
...
+
+package io.opentelemetry.sdk.metrics.view;
+
+import com.google.auto.value.AutoValue;
+import io.opentelemetry.sdk.metrics.aggregator.AggregatorFactory;
+import javax.annotation.concurrent.Immutable;
+
+@AutoValue
+@Immutable
+public abstract class View {
```
jkwatson, line 14: If we decide to go with this structure, this class definitely needs javadoc, even if only placeholder javadoc for now.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/ViewRegistry.java
```diff
@@ -43,10 +46,10 @@
     configuration.put(InstrumentType.VALUE_OBSERVER, EMPTY_CONFIG);
   }
 
-  void registerView(InstrumentSelector selector, AggregatorFactory aggregatorFactory) {
+  void registerView(InstrumentSelector selector, View aggregatorFactory) {
```
bogdandrutu, line 49: s/aggregatorFactory/view

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/View.java
parallelstream: makes sense. Any ideas as to what mention there? My proposal
/**
* Encompasses aggregation and processing logic related to a specific view.
**/

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/View.java
```diff
@@ -0,0 +1,27 @@
...
+  public abstract AggregatorFactory getAggregatorFactory();
+
+  public static Builder builder() {
+    return new AutoValue_View.Builder();
+  }
+
+  @AutoValue.Builder
+  public abstract static class Builder {
```
bogdandrutu, line 22: We made an internal rule to not expose public inner classes. So this should be ViewBuilder :)

bogdandrutu, line 22: Hide the ctor for this so only View.builder() allows to create instances.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
```diff
@@ -98,10 +98,11 @@ public static SdkMeterProviderBuilder builder() {
    * AggregationFactory aggregationFactory = AggregationFactory.minMaxSumCount();
    *
    * //register the view with the MeterSdkProvider
-   * meterProvider.registerView(instrumentSelector, aggregationFactory);
+   * meterProvider.registerView(instrumentSelector, View.builder()
+   *   .setAggregatorFactory(factory).build());
```
bogdandrutu, line 102: s/factory/aggregationFactory? To make example work.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/ViewRegistry.java
parallelstream: done

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/View.java
parallelstream: done

parallelstream: done

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
parallelstream: done

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/View.java
jkwatson: That seems a little recursive. What's a view?

parallelstream: alright, I looked how OpenCensus defined it and here's what we can do:

```
/**
 * Views are the coupling of an AggregatorFactory applied to an Instrument.
 * Views are the connection to Accumulator and used to apply various processing to instruments before
 * storing the value in Accumulator.
 * */
```

jkwatson: Hmm. Thinking on this more, I think we should probably just leave this as a todo until we have an actual specification for Views.

parallelstream: done

jkwatson:
Can you give any hints about what you intend to add to Views? Maybe a design document with some ideas? I'm all on board with supporting opencensus style views, but I'd like to understand where we're going with the approach in this PR.

bogdandrutu:
@jkwatson I think the next steps are in #1810

bogdandrutu (approved):

jkwatson:
> @jkwatson I think the next steps are in #1810

We should just be clear at this point that this is all very prototype work. This isn't in any specification yet, and is extremely likely to be re-worked before metrics get to 1.0.

PR #2957, opened 2021-03-02
Don't fail on timeout / interrupted since the result isn't really don…

### LINKED ISSUE #2942: CompletableResultCode.join() suprisingly fail()s on timeout.
This is different from the JDK's [CompletableFuture.get(long, TimeUnit)](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#get-long-java.util.concurrent.TimeUnit-) and IMHO quite surprising. It also reduces the flexibility of the method as calling `fail()` in the `catch(TimeoutException)` is easier than re-implementing `join` on top of `whenComplete`.

Code reference:

https://github.com/open-telemetry/opentelemetry-java/blob/v0.17.1/sdk/common/src/main/java/io/opentelemetry/sdk/common/CompletableResultCode.java#L150-L157

I suggest either changing the behavior (silent break) or renaming the method to `joinOrFail`.

anuraaga:
…e yet.

If code checks `isSuccess` after join, there will be no change as it still returns `false` before or after this change. If code checks `isDone` after join, now it will more precisely return `false` on timeout - you don't check `isDone` if you don't have code to properly handle the non-done case, so this should be a safe (even beneficial) change.

Fixes #2942

### DIFF
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/common/CompletableResultCode.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/common/CompletableResultCode.java
@@ -136,7 +136,9 @@ public final class CompletableResultCode {
   }
 
   /**
-   * Waits for the specified amount of time for this {@link CompletableResultCode} to complete.
+   * Waits up to the specified amount of time for this {@link CompletableResultCode} to complete.
+   * Even after this method returns, the result may not be complete yet - you should always check
+   * {@link #isSuccess()} or {@link #isDone()} after calling this method to determine the result.
    *
    * @return this {@link CompletableResultCode}
    */
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/common/CompletableResultCode.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/common/CompletableResultCode.java
@@ -136,8 +136,7 @@ public final class CompletableResultCode {
   }
 
   /**
-   * Waits for the specified amount of time for this {@link CompletableResultCode} to complete. If
-   * it times out or is interrupted, the {@link CompletableResultCode} is failed.
+   * Waits for the specified amount of time for this {@link CompletableResultCode} to complete.
    *
    * @return this {@link CompletableResultCode}
    */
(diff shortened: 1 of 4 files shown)


sdk/trace/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorTest.java
```diff
@@ -317,52 +329,47 @@ void exporterThrowsException() {
 
   @Test
   @Timeout(5)
-  public void exporterTimesOut() throws InterruptedException {
```
anuraaga, line 320: I don't think this test was actually checking what it wanted to check, or at least I couldn't reason about it well. Tried rewriting it.

Oberon00:
One more thing that changes: join() will no longer force completion callbacks to run.

anuraaga:
@Oberon00 Yeah I think that's what exposed what I found to be a strange test. Calling callbacks because of join is a bug I think and we can fix it with extremely low chance of bad effects I guess.

sdk/common/src/main/java/io/opentelemetry/sdk/common/CompletableResultCode.java
```diff
@@ -136,8 +136,7 @@ public boolean isDone() {
   }
 
   /**
-   * Waits for the specified amount of time for this {@link CompletableResultCode} to complete. If
-   * it times out or is interrupted, the {@link CompletableResultCode} is failed.
+   * Waits for the specified amount of time for this {@link CompletableResultCode} to complete.
```
jkwatson, line 139: I think it might be useful to also have a note in here about the fact that the timeout *doesn't* change the state of the result at all.

jkwatson (approved):

sdk/common/src/main/java/io/opentelemetry/sdk/common/CompletableResultCode.java
Oberon00: Yes, I think we should add a big fat warning that after the join you need to manually check isDone/isSuccess.

Oberon00 (approved):

PR #2958, opened 2021-03-02
Fix flaky test

anuraaga:
If `shutdown` runs synchronously due to thread timing, it will NPE on the mock's result code if not mocked.

### DIFF
--- a/sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfigurationTest.java
+++ b/sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfigurationTest.java
@@ -7,8 +7,10 @@ package io.opentelemetry.sdk.autoconfigure;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.Mockito.when;
 
 import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.sdk.common.CompletableResultCode;
 import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.sdk.trace.SdkTracerProvider;
 import io.opentelemetry.sdk.trace.SpanLimits;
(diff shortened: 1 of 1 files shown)


sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfigurationTest.java
```diff
@@ -21,21 +23,30 @@
...
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.junit.jupiter.MockitoSettings;
+import org.mockito.quality.Strictness;
 
 // NB: We use AssertJ extracting to reflectively access implementation details to test configuration
 // because the use of BatchSpanProcessor makes it difficult to verify values through public means.
 @ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.LENIENT)
```
jkwatson, line 37: why not just put the lenient on the individual item, rather than the whole class?

anuraaga: Hmm - I've never really thought about it. I feel there are two patterns with mockito though, either you use strict stubs and never write `verify`, or you use lenient and always write `verify`. Don't think it's ever a great idea to try to mix them in the same class.

anuraaga: In this case it could be strict if I called `join` on the shutdown calls I think. That slows the tests down just a tiny bit, so wasn't sure it's worth it, but what do you think?

jkwatson: The other possibility is to just initialize the mock in the tests that need it, rather than have it be at the class level at all (this is almost always what I prefer, to keep the test cases more clearly self-contained).

jkwatson (approved):

PR #2959, opened 2021-03-02
Move api/context to context.

anuraaga:
Reflects package name


jkwatson:
gah! why does github make this look like a ton of new classes, rather than just moves?

jkwatson:
> gah! why does github make this look like a ton of new classes, rather than just moves?

and now github has sorted it out and makes it easy. 😕

jkwatson:
I'm totally fine with this, but I'd also like to make sure that @bogdandrutu is ok, so we don't move it again. :)

jkwatson (approved):

bogdandrutu:
I think the maven artifact is ":context" so probably we can stick with this path.

bogdandrutu (approved):

PR #2960, opened 2021-03-02
Ignore classpath error when otlp is set as metrics exporter but is no…

### LINKED ISSUE #2941: Bad user experience if metrics is not on the classpath
If you use the auto-configure module, but don't want metrics, and don't have it on your classpath, you get a big, ugly error when you start up: 

```
Exception in thread "main" io.opentelemetry.sdk.autoconfigure.ConfigurationException: OTLP Metrics Exporter enabled but opentelemetry-exporter-otlp-metrics not found on classpath. Make sure to add it as a dependency to enable this feature.
	at io.opentelemetry.sdk.autoconfigure.ClasspathUtil.checkClassExists(ClasspathUtil.java:15)
	at io.opentelemetry.sdk.autoconfigure.MetricExporterConfiguration.configureOtlpMetrics(MetricExporterConfiguration.java:53)
	at io.opentelemetry.sdk.autoconfigure.MetricExporterConfiguration.configureExporter(MetricExporterConfiguration.java:27)
	at io.opentelemetry.sdk.autoconfigure.OpenTelemetrySdkAutoConfiguration.configureMeterProvider(OpenTelemetrySdkAutoConfiguration.java:61)
	at io.opentelemetry.sdk.autoconfigure.OpenTelemetrySdkAutoConfiguration.initialize(OpenTelemetrySdkAutoConfiguration.java:42)
```

We shouldn't fail to start up the app if the user doesn't have any intention to have metrics enabled. Until metrics is stable, our auto-configuration module probably shouldn't default to `otlp` as the exporter.

Current workaround is: `-Dotel.metrics.exporter=none`, but it's a hassle to make the user do that.

anuraaga:
…t on the classpath since the default value is otlp but metrics is alpha.

Also changes error message to encourage users to use `-otlp`, not `-otlp-trace` artifact which I think we prefer.

Fixes #2941

### DIFF
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
@@ -58,7 +58,7 @@ final class MetricExporterConfiguration {
           "OTLP Metrics Exporter",
           "opentelemetry-exporter-otlp-metrics");
     } catch (ConfigurationException e) {
-      // Squash this for now, once metrics is stable and included in the `exporter-otlp` artifact
+      // Squash this for now, until metrics are stable and included in the `exporter-otlp` artifact
       // by default,
       return null;
     }
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
@@ -17,6 +17,7 @@ import io.prometheus.client.exporter.HTTPServer;
 import java.io.IOException;
 import java.time.Duration;
 import java.util.Collections;
+import javax.annotation.Nullable;
 
 final class MetricExporterConfiguration {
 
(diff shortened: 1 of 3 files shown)


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
```diff
@@ -48,12 +49,19 @@ private static void configureLoggingMetrics(
...
-        "io.opentelemetry.exporter.otlp.metrics.OtlpGrpcMetricExporter",
-        "OTLP Metrics Exporter",
-        "opentelemetry-exporter-otlp-metrics");
+    try {
+      ClasspathUtil.checkClassExists(
+          "io.opentelemetry.exporter.otlp.metrics.OtlpGrpcMetricExporter",
+          "OTLP Metrics Exporter",
+          "opentelemetry-exporter-otlp-metrics");
+    } catch (ConfigurationException e) {
+      // Squash this for now, once metrics is stable and included in the `exporter-otlp` artifact
```
iNikem, line 61:
```suggestion
      // Squash this for now, until metrics are stable and included in the `exporter-otlp` artifact
```

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/SpanExporterConfiguration.java
```diff
@@ -67,7 +67,7 @@ static OtlpGrpcSpanExporter configureOtlpSpans(ConfigProperties config) {
     ClasspathUtil.checkClassExists(
         "io.opentelemetry.exporter.otlp.trace.OtlpGrpcSpanExporter",
         "OTLP Trace Exporter",
-        "opentelemetry-exporter-otlp-trace");
+        "opentelemetry-exporter-otlp");
```
iNikem, line 70: Why this change?

anuraaga: Oops forgot to add to description, updated

jkwatson (approved):
Thanks!

bogdandrutu (approved):

PR #2962, opened 2021-03-02
Add a new module for extensions to the tracing APIs.

jkwatson:
Includes one simple possible example.

Everything about this PR is up for modification. This is more of a proposal for some possible convenience APIs and a new module to house them.

I also realize that we could very easily add some kotlin extension functions to Tracer that use this, and that would probably be a nice enhancement for kotlin users.


api/all/src/main/java/io/opentelemetry/api/internal/ReadOnlyArrayMap.java
```diff
@@ -32,7 +32,7 @@
 
 /** A read-only view of an array of key-value pairs. */
 @SuppressWarnings("unchecked")
-public final class ReadOnlyArrayMap<K, V> implements Map<K, V> {
+public final class ReadOnlyArrayMap<K, V> extends AbstractMap<K, V> {
```
jkwatson, line 35: this gets us valid Map equals/hashcode, which we didn't have before.

extensions/tracing/src/main/java/io/opentelemetry/extension/tracing/Spanalyzer.java
```diff
@@ -0,0 +1,67 @@
...
+import io.opentelemetry.context.Scope;
+import java.util.concurrent.Callable;
+
+/**
+ * Provides easy mechanisms for wrapping standard Java constructs with an OpenTelemetry Span.
+ *
+ * <p>Todo: figure out a better name for this.
+ */
+@SuppressWarnings("InconsistentOverloads")
+public class Spanalyzer {
```
jkwatson, line 19: obviously, this is a ridiculous name. Suggestions extremely welcome.

kenfinnigan: I know it was used in a different way in OpenTracing, but `SpanWrapper` kind of fits what you're doing here

jkwatson: we also already have an internal class called `SpanWrapper` that could lead to confusion, but yes, this is one that I had considered.

api/all/src/main/java/io/opentelemetry/api/internal/ReadOnlyArrayMap.java
bogdandrutu: Should this be sent as a standalone change anyway?

jkwatson: yes, I was thinking that as well. will do, for sure.

jkwatson: https://github.com/open-telemetry/opentelemetry-java/pull/2963

extensions/tracing/src/main/java/io/opentelemetry/extension/tracing/Spanalyzer.java
anuraaga: `ExtendedTracer` since presumably if moving these into core we'd add to `Tracer`?

anuraaga (commented):
This seems reasonable to me

extensions/tracing/src/main/java/io/opentelemetry/extension/tracing/Spanalyzer.java
jkwatson: not bad. Or maybe `TracerExtension`?

anuraaga: If kotlin, it'd definitely be `TracerExtensions` :) I guess per Java we'd want to name this what it is, which is an "extended tracer".

extensions/tracing/src/main/java/io/opentelemetry/extension/tracing/ExtendedTracer.java
```diff
@@ -0,0 +1,69 @@
...
+package io.opentelemetry.extension.tracing;
+
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.context.Scope;
+import java.util.concurrent.Callable;
+
+/** Provides easy mechanisms for wrapping standard Java constructs with an OpenTelemetry Span. */
+@SuppressWarnings("InconsistentOverloads")
+public class ExtendedTracer {
```
bogdandrutu, line 15: I think it would be nice to make this "implement" the Tracer interface as well for users to not have to store 2 of them.

bogdandrutu, line 15: "final" :)

extensions/tracing/src/main/java/io/opentelemetry/extension/tracing/ExtendedTracer.java
```diff
@@ -0,0 +1,69 @@
...
+      try (Scope scope = span.makeCurrent()) {
+        runnable.run();
+      } catch (Throwable e) {
+        span.recordException(e);
+        throw e;
+      } finally {
+        span.end();
+      }
+    };
+  }
```
bogdandrutu, line 45: If you do the implementation of the Tracer then users who wants to use this will do:

```java
// For extended
private final ExtendedTracer tracer = ExtendedTracer.create(tracerProvider.get("foo"));

// Before
private final Tracer tracer = tracerProvider.get("foo");
```

So probably the static methods are not very useful.

anuraaga:
@jkwatson By the way, did we want to call this package `incubator` or do we intend not to move code into the core after getting feedback?

bogdandrutu:
My 2 cents here: I think we should not move everything in core, but instead give users a stable extension package. This package may include helpers for metrics as well.

extensions/tracing/src/main/java/io/opentelemetry/extension/tracing/ExtendedTracer.java
anuraaga: This seems pretty nice, can delegate the methods

extensions/tracing/src/main/java/io/opentelemetry/extension/tracing/ExtendedTracer.java
```diff
@@ -0,0 +1,69 @@
...
+/** Provides easy mechanisms for wrapping standard Java constructs with an OpenTelemetry Span. */
+@SuppressWarnings("InconsistentOverloads")
+public class ExtendedTracer {
+
+  private final Tracer tracer;
+
+  /** Create a new {@link ExtendedTracer} that wraps the provided Tracer. */
+  public ExtendedTracer(Tracer tracer) {
+    this.tracer = tracer;
+  }
```
anuraaga, line 22:
```suggestion
  /** Create a new {@link ExtendedTracer} that wraps the provided Tracer. */
  public ExtendedTracer create(Tracer tracer) {
    return new ExtendedTracer(tracer);
  }

  private ExtendedTracer(Tracer tracer) {
    this.tracer = tracer;
  }
```

extensions/tracing/src/main/java/io/opentelemetry/extension/tracing/ExtendedTracer.java
```diff
@@ -0,0 +1,69 @@
...
+  /** Wrap the provided {@link Runnable} with a {@link Span} with the provided name. */
+  public Runnable wrap(String spanName, Runnable runnable) {
+    return wrap(tracer, spanName, runnable);
+  }
+
+  /**
+   * Wrap the provided {@link Runnable} with a {@link Span} with the provided name. Uses the
+   * provided {@link Tracer} for the functionality.
+   */
+  public static Runnable wrap(Tracer tracer, String spanName, Runnable runnable) {
```
anuraaga, line 33: We can remove these statics by implementing `Tracer`, it's free to just use `ExtendedTracer` instead of `Tracer`

extensions/tracing/src/main/java/io/opentelemetry/extension/tracing/ExtendedTracer.java
```diff
@@ -0,0 +1,69 @@
...
+
+  private final Tracer tracer;
+
+  /** Create a new {@link ExtendedTracer} that wraps the provided Tracer. */
+  public ExtendedTracer(Tracer tracer) {
+    this.tracer = tracer;
+  }
+
+  /** Wrap the provided {@link Runnable} with a {@link Span} with the provided name. */
+  public Runnable wrap(String spanName, Runnable runnable) {
```
anuraaga, line 25: Is `wrap` that useful? What sort of patterns is it good for? I sort of expect most people to want to `run` (similar to what `WithSpan` does when using the agent).

anuraaga (commented):
Span creation is basically the whole point of OpenTelemetry - so I'm not a fan of splitting our own ecosystem if we keep this in an extension. "Are you using the extended tracer?", either when answering a support request or recommending how to use OpenTelemetry, isn't a question I'd like to have to ask in the long term. So for these, I would go with experimenting with the goal of getting them into core.

This isn't true of less-core features, for example an `AttributeFilteringExporter` has been asked for a few times I think - it has enough demand that it could probably be in core, but since it's a totally different concept than anything we have, it doesn't feel like a split to have it in an extension too.

extensions/tracing/src/main/java/io/opentelemetry/extension/tracing/ExtendedTracer.java
jkwatson: will do! good idea!

extensions/tracing/src/main/java/io/opentelemetry/extension/tracing/ExtendedTracer.java
jkwatson: It seems like it would be useful to use with frameworks that provide Runnables, and/or require them to be passed into other parts of the framework. Although, that also begs the question of how those frameworks might require context propagation work as well.

It's a fair question, for sure. I'm not sure how people might want to use these kinds of wrappers. If you think that just providing `run` and `call` methods would be less risky and/or more useful, happy to change to that for starters.

extensions/tracing/src/main/java/io/opentelemetry/extension/tracing/ExtendedTracer.java
jkwatson: ok, will remove them for now.

extensions/tracing/src/main/java/io/opentelemetry/extension/tracing/ExtendedTracer.java
jkwatson: done

extensions/tracing/src/main/java/io/opentelemetry/extension/tracing/ExtendedTracer.java
jkwatson: simplified with just `run` and `call`. I think this makes sense as a first set of methods to provide.

anuraaga (approved):
Code LGTM - still wondering about extension vs incubator / long-term separation vs migrating into core. I do prefer incubator, but we could still do renames of the artifact after merging if that's better

jkwatson:
> Code LGTM - still wondering about extension vs incubator / long-term separation vs migrating into core. I do prefer incubator, but we could still do renames of the artifact after merging if that's better

oh, somehow I missed a rename to `incubator` was desired. I can do that now if you'd like.

anuraaga:
@jkwatson Yeah I think that'd be nice

jkwatson:
> @jkwatson Yeah I think that'd be nice

Do we want just `incubator` or `tracing-incubator`?

anuraaga:
@jkwatson Since API isn't meant to be split like the SDK is, `incubator` seems ok

anuraaga (approved):

jkwatson:
Anything left on here to do? Would be nice to have this released as a part of 1.1.0 in a couple of weeks.

extensions/incubator/src/main/java/io/opentelemetry/extension/tracing/ExtendedTracer.java
```diff
@@ -0,0 +1,58 @@
...
+import io.opentelemetry.context.Scope;
+import java.util.concurrent.Callable;
+
+/** Provides easy mechanisms for wrapping standard Java constructs with an OpenTelemetry Span. */
+public final class ExtendedTracer implements Tracer {
+
+  private final Tracer delegate;
+
+  /** Create a new {@link ExtendedTracer} that wraps the provided Tracer. */
+  public static ExtendedTracer create(Tracer delegate) {
```
bogdandrutu, line 20: personal taste :) 

s/delegate/base?

extensions/incubator/src/main/java/io/opentelemetry/extension/tracing/ExtendedTracer.java
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.extension.tracing;
```
bogdandrutu, line 6: please please `io.opentelemetry.extension.trace` to be consistent with api 🗡️

extensions/incubator/src/main/java/io/opentelemetry/extension/tracing/ExtendedTracer.java
jkwatson: hm. I personally prefer `delegate` since this is the GoF delegate pattern, but I don't feel overly strongly about it.

extensions/incubator/src/main/java/io/opentelemetry/extension/tracing/ExtendedTracer.java
jkwatson: will do.

extensions/incubator/src/main/java/io/opentelemetry/extension/tracing/ExtendedTracer.java
bogdandrutu: Leave it delegate, for me was more that this is the `base` tracer and extended :) leave it as delegate.

extensions/incubator/src/main/java/io/opentelemetry/extension/tracing/ExtendedTracer.java
jkwatson: wait. we can't do this, because we'll be putting the code in the same package as another module, which will possibly break users of the Java Module System.

jkwatson: never mind, it's in the `extension` sub-package so it's all good.

jkwatson: done

extensions/incubator/src/main/java/io/opentelemetry/extension/trace/ExtendedTracer.java
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.extension.trace;
```
anuraaga, line 6: extension.incubator.trace?

anuraaga (approved):

extensions/incubator/src/main/java/io/opentelemetry/extension/trace/ExtendedTracer.java
jkwatson: No opinion. @bogdandrutu what do you think?

carlosalberto (approved):

extensions/incubator/src/main/java/io/opentelemetry/extension/trace/ExtendedTracer.java
anuraaga: Since `moduleName` is `extra["moduleName"] = "io.opentelemetry.extension.incubator"` is it questionable?

jkwatson: Do we have a general rule of having the package match the module? I'm fine with that; I just hadn't considered it. :)

anuraaga: I thought so at least :) I think so far all our module names are prefixes of the packages in that artifact.

jkwatson: done

anuraaga (approved):

PR #2969, opened 2021-03-03
Add config option for queue poll wait time

sbandadd:


anuraaga:
I realized we probably do want to tweak the polling strategy. Can you share the code you've been experimenting with for signalling?

sbandadd:
@anuraaga This is what I was playing with https://github.com/sbandadd/opentelemetry-java/pull/1/files.

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfigurationTest.java
```diff
@@ -40,6 +40,7 @@
   void configureTracerProvider() {
     Map<String, String> properties = new HashMap<>();
     properties.put("otel.bsp.schedule.delay", "100000");
+    properties.put("otel.bsp.queue.poll.wait.time", "100");
```
bogdandrutu, line 43: This probably needs to go in specs?

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -188,7 +193,7 @@ public void run() {
         }
 
         try {
-          ReadableSpan lastElement = queue.poll(100, TimeUnit.MILLISECONDS);
+          ReadableSpan lastElement = queue.poll(queuePollWaitTimeNanos, TimeUnit.NANOSECONDS);
```
bogdandrutu, line 196: Can you document the motivation? We don't want to add configuration for everything unless there is a good use-case to configure that.

sbandadd:
We are looking for better alternatives. I don't think this is needed anyway.

PR #2971, opened 2021-03-03
Don't poll more than needed in BSP.

anuraaga:
Just whipped this up to compare against #2969. @sbandadd sorry if this overlaps with any approach you've tried.

### DIFF
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
@@ -127,10 +127,12 @@ public final class BatchSpanProcessor implements SpanProcessor {
     private long nextExportTime;
 
     private final BlockingQueue<ReadableSpan> queue;
+    private final BlockingQueue<Object> signal;
 
     private final AtomicReference<CompletableResultCode> flushRequested = new AtomicReference<>();
     private volatile boolean continueWork = true;
-    private final ArrayList<SpanData> batch;
+    private final ArrayList<ReadableSpan> batch;
+    private final ArrayList<SpanData> batchData;
 
     private Worker(
         SpanExporter spanExporter,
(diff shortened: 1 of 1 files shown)


jkwatson:
would love some benchmarking results along with these changes so we can compare apples to apples and not just guess at what might be better. :)

anuraaga:
Yup - it'll be a bit tricky since I don't think we have any overhead benchmarks where we'd measure e.g. CPU usage of a change. Might work on it but wanted to throw this out there first

jkwatson:
Along with any of these benchmarks of the BSP, we should also include metrics output so we can see how many spans we're dropping because the queue is full. I suspect this is a confounding factor that might be skewing results.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -261,9 +272,13 @@ private void exportCurrentBatch() {
         return;
       }
 
+      for (ReadableSpan span : batch) {
+        batchData.add(span.toSpanData());
```
jkwatson, line 276: For good or ill, moving this down here will make the CPU usage a bit more spiky, since we'll only do this conversion all at once when the queue is full, or when the timer expires. The current implementation will spread out the conversion smoothly, since it happens as each span is added to the queue. Which way is better? I'm not sure, honestly, but it's something to keep in mind.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -127,10 +127,12 @@ public CompletableResultCode forceFlush() {
     private long nextExportTime;
 
     private final BlockingQueue<ReadableSpan> queue;
```
sbandadd, line 129: Using ArrayDequeue would be better when draining in batches similar to https://github.com/sbandadd/opentelemetry-java/pull/1/files. But I think the most important thing is to come up with a CPU benchmark and compare different versions.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
anuraaga: Yeah there's some tradeoff - I guess for a thread with a defined "interval", I'd sort of expect it to wake, work, sleep. We have a bit of a pathological behavior right now where it's fairly conceivable for this thread to be sleeping and waking in a thrashing manner if a span is recorded every 101ms for example - sleeping/waking is very expensive and I suspect we didn't intend that to be possible.

anuraaga:
Superceded by #2983

PR #2979, opened 2021-03-04
Don't copy events toSpanData if ended.

anuraaga:
We have this check for attributes but seems we can have it for events too.


jkwatson (approved):
Thanks!

PR #2982, opened 2021-03-04
Store SpanShim/Baggage in the Context.

carlosalberto:
Store both the `SpanShim` and `Baggage` no make the OT Shim Specification compliant:

* `SpanShim` is stored in the `Context` to avoid creating disposable `SpanShim` objects when returning the active OT `Span`.
* `Baggage` of the `Span` is stored in the `Context` as well, if there's any defined for it.


opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/ScopeManagerShim.java
```diff
@@ -18,22 +22,31 @@ public ScopeManagerShim(TelemetryInfo telemetryInfo) {
   @Override
   @SuppressWarnings("ReturnMissingNullable")
   public Span activeSpan() {
+    Context context = Context.current();
+    io.opentelemetry.api.trace.Span span = io.opentelemetry.api.trace.Span.fromContext(context);
+    SpanShim spanShim = context.get(SPAN_SHIM_KEY);
```
jkwatson, line 27: Any thoughts to making `SpanShim` implement `ImplicitContextKeyed`, or do you not want to expose that API on the shim?

carlosalberto: Yes, definitely. Left it out to make it easier for Yuri (or whoever reviews that Spec PR) to review.

opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/ScopeManagerShim.java
```diff
@@ -18,22 +22,31 @@ public ScopeManagerShim(TelemetryInfo telemetryInfo) {
...
+    // Span was activated from outside the Shim layer unfortunately.
     return new SpanShim(telemetryInfo(), span);
   }
 
   @Override
   @SuppressWarnings("MustBeClosedChecker")
   public Scope activate(Span span) {
     io.opentelemetry.api.trace.Span actualSpan = getActualSpan(span);
-    return new ScopeShim(actualSpan.makeCurrent());
+    Context context = Context.current().with(actualSpan).with(SPAN_SHIM_KEY, (SpanShim) span);
```
jkwatson, line 48: If you implement `ImplicitContextKeyed`, I think you could have the implementation do the work of putting the span into the context as well.

opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/ScopeManagerShim.java
anuraaga: Do you think we can have `SpanShim` just implement both OTel and OT `span` and not have to keep track of two objects? It worked ok in OC so I suspect it'll work for OT.

https://github.com/open-telemetry/opentelemetry-java/blob/main/opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetrySpanImpl.java#L53

carlosalberto: I can take a stab at such optimization once the OT Spec portion is merged ;)

anuraaga: From the perspective of maintainence here, if both shims operated very similarly that would be quite great. Is it possible to block work on the OT shim until the spec is merged? Or alternatively, do you have to wait for the spec? I think the only way to spec out the shims is to implement them first, as @jsuereth did with OC I believe.

carlosalberto:
@open-telemetry/java-approvers Please review again, as this is ready to go in ;)

PS - I decided to _not_ try to make the `SpanShim` object implement both OT/OTel interfaces for now, in order to expedite a full, correct implementation of the OT Shim portion of the spec. Such potential change will be done in a follow up ;)

anuraaga (approved):
Curious what people think about moving OpenTracing / OpenCensus shims into auto instrumentation and not having actual libraries.

jkwatson:
> Curious what people think about moving OpenTracing / OpenCensus shims into auto instrumentation and not having actual libraries.

Aren't there non-autoinstrumentation users who might want these as well?

jkwatson (approved):

anuraaga:
> Aren't there non-autoinstrumentation users who might want these as well?

I dunno - I recall hearing OpenCensus Shim not being a priority anymore, and I wonder if it's because of the ease of auto instrumentation for most users.

trask:
@anuraaga and I are also a bit concerned about two layers of bridging at once if someone uses the OpenTracing shim (bridge) + the Javaagent (bridge)

carlosalberto:
@anuraaga Also, most OT users used to be manual instrumentation users, and the ones who used the OT SpecialAgent have so far simply moved directly to OTel auto-instrumentation, no need for the OT Shim.

anuraaga:
@carlosalberto I won't try to make any recommendation for OT I just want to make sure all the options are taken into account. In this case, I recommend thinking not only of the manual/auto axis but OT/OTel axis. By making our "OT shim" be a shim introduced by the java agent, we will have a good story of getting users onto OTel, in fact any instrumentation supported by OTel would be using it with the shim for custom frameworks instrumentation (this requires work on the auto instrumentation side of course to suppress overlapping instrumentation if we happen to go with it). Or if a user loves library instrumentation they can make the code change - it's what I'd probably do for my own frameworks and my understanding is this is where OC is going where they've found that library users tend to not mind just switching over, both agents and shims are off limits for this sort of persona.

It doesn't mean it applies for OT but I don't think we've actually done user research to know the shim needs to live outside a java agent.

carlosalberto:
Actually doing some OT research is not a bad idea. As I said, what I've seen is manual instrumentation for OT mostly, but it would be interesting to see the experience of other companies. I will open an issue in this repo for now, so we don't forget about that possibility.

PR #2983, opened 2021-03-04
Optimize batch span processor

sbandadd:
Description:
Batch span processor currently is aggressive in the sense that any new spans are sent to the exporter,
this involves lots of overhead from signaling under heavy load and overhead from constant polling by exporter thread
under less load. This PR makes exporter thread wait for maxExportBatchSize to avoid busy polling of the queue.

# Benchmark results
BatchSpanProcessorMultiThreadBenchmark.java result
image

BatchSpanProcessorCpuBenchmark.java result
image

### DIFF
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
@@ -130,9 +130,9 @@ public final class BatchSpanProcessor implements SpanProcessor {
     private final long exporterTimeoutNanos;
 
     private long nextExportTime;
-
     private final BlockingQueue<ReadableSpan> queue;
-
+    private final AtomicBoolean needSignal = new AtomicBoolean(false);
+    private final BlockingQueue<Boolean> signal;
     private final AtomicReference<CompletableResultCode> flushRequested = new AtomicReference<>();
     private volatile boolean continueWork = true;
     private final ArrayList<SpanData> batch;
@@ -148,6 +148,7 @@ public final class BatchSpanProcessor implements SpanProcessor {
       this.maxExportBatchSize = maxExportBatchSize;
       this.exporterTimeoutNanos = exporterTimeoutNanos;
       this.queue = queue;
+      this.signal = new ArrayBlockingQueue<>(1);
       Meter meter = GlobalMetricsProvider.getMeter("io.opentelemetry.sdk.trace");
       meter
           .longValueObserverBuilder("queueSize")
@@ -180,6 +181,10 @@ public final class BatchSpanProcessor implements SpanProcessor {
     private void addSpan(ReadableSpan span) {
       if (!queue.offer(span)) {
         droppedSpans.add(1);
+      } else {
+        if (queue.size() >= maxExportBatchSize && needSignal.get()) {
+          signal.offer(true);
+        }
       }
     }
 
@@ -191,20 +196,25 @@ public final class BatchSpanProcessor implements SpanProcessor {
         if (flushRequested.get() != null) {
           flush();
         }
-
-        try {
-          ReadableSpan lastElement = queue.poll(100, TimeUnit.MILLISECONDS);
-          if (lastElement != null) {
-            batch.add(lastElement.toSpanData());
-          }
-        } catch (InterruptedException e) {
-          Thread.currentThread().interrupt();
-          return;
+        while (!queue.isEmpty() && batch.size() < maxExportBatchSize) {
+          batch.add(queue.poll().toSpanData());
         }
-
         if (batch.size() >= maxExportBatchSize || System.nanoTime() >= nextExportTime) {
           exportCurrentBatch();
           updateNextExportTime();
+        }
+        if (queue.isEmpty()) {
+          try {
+            long pollWaitTime = nextExportTime - System.nanoTime();
+            if (pollWaitTime > 0) {
+              needSignal.set(true);
+              signal.poll(pollWaitTime, TimeUnit.NANOSECONDS);
+              needSignal.set(false);
+            }
+          } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            return;
+          }
         }
       }
     }
@@ -252,8 +262,10 @@ public final class BatchSpanProcessor implements SpanProcessor {
 
     private CompletableResultCode forceFlush() {
       CompletableResultCode flushResult = new CompletableResultCode();
-      // we set the atomic here to trigger the worker loop to do a flush on its next iteration.
-      flushRequested.compareAndSet(null, flushResult);
+      // we set the atomic here to trigger the worker loop to do a flush of the entire queue.
+      if (flushRequested.compareAndSet(null, flushResult)) {
+        signal.offer(true);
+      }
       CompletableResultCode possibleResult = flushRequested.get();
       // there's a race here where the flush happening in the worker loop could complete before we
       // get what's in the atomic. In that case, just return success, since we know it succeeded in
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
@@ -23,6 +23,7 @@ import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -131,7 +132,13 @@ public final class BatchSpanProcessor implements SpanProcessor {
 
     private long nextExportTime;
     private final BlockingQueue<ReadableSpan> queue;
-    private final AtomicBoolean needSignal = new AtomicBoolean(false);
+    // When waiting on the spans queue, exporter thread sets this atomic to the number of more
+    // spans it needs before doing an export. Writer threads would then wait for the queue to reach
+    // spansNeeded size before notifying the exporter thread about new entries.
+    // Integer.MAX_VALUE is used to imply that exporter thread is not expecting any signal. Since
+    // exporter thread doesn't expect any signal initially, this value is initialized to
+    // Integer.MAX_VALUE.
+    private final AtomicInteger spansNeeded = new AtomicInteger(Integer.MAX_VALUE);
     private final BlockingQueue<Boolean> signal;
     private final AtomicReference<CompletableResultCode> flushRequested = new AtomicReference<>();
     private volatile boolean continueWork = true;
@@ -182,7 +189,7 @@ public final class BatchSpanProcessor implements SpanProcessor {
       if (!queue.offer(span)) {
         droppedSpans.add(1);
       } else {
-        if (queue.size() >= maxExportBatchSize && needSignal.get()) {
+        if (queue.size() >= spansNeeded.get()) {
           signal.offer(true);
         }
       }
@@ -207,9 +214,9 @@ public final class BatchSpanProcessor implements SpanProcessor {
           try {
             long pollWaitTime = nextExportTime - System.nanoTime();
             if (pollWaitTime > 0) {
-              needSignal.set(true);
+              spansNeeded.set(maxExportBatchSize - batch.size());
               signal.poll(pollWaitTime, TimeUnit.NANOSECONDS);
-              needSignal.set(false);
+              spansNeeded.set(Integer.MAX_VALUE);
             }
           } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
--- a/sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfigurationTest.java
+++ b/sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfigurationTest.java
@@ -7,8 +7,10 @@ package io.opentelemetry.sdk.autoconfigure;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.Mockito.when;
 
 import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.sdk.common.CompletableResultCode;
 import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.sdk.trace.SdkTracerProvider;
 import io.opentelemetry.sdk.trace.SpanLimits;
@@ -21,20 +23,29 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.TimeUnit;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.junit.jupiter.MockitoSettings;
+import org.mockito.quality.Strictness;
 
 // NB: We use AssertJ extracting to reflectively access implementation details to test configuration
 // because the use of BatchSpanProcessor makes it difficult to verify values through public means.
 @ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.LENIENT)
 class TracerProviderConfigurationTest {
 
   private static final ConfigProperties EMPTY =
       ConfigProperties.createForTest(Collections.emptyMap());
 
-  @Mock private SpanExporter exporter;
+  @Mock private SpanExporter mockSpanExporter;
+
+  @BeforeEach
+  void setUp() {
+    when(mockSpanExporter.shutdown()).thenReturn(CompletableResultCode.ofSuccess());
+  }
 
   @Test
   void configureTracerProvider() {
@@ -69,7 +80,7 @@ class TracerProviderConfigurationTest {
   @Test
   void configureSpanProcessor_empty() {
     BatchSpanProcessor processor =
-        TracerProviderConfiguration.configureSpanProcessor(EMPTY, exporter);
+        TracerProviderConfiguration.configureSpanProcessor(EMPTY, mockSpanExporter);
 
     try {
       assertThat(processor)
@@ -88,7 +99,7 @@ class TracerProviderConfigurationTest {
                     .isInstanceOfSatisfying(
                         ArrayBlockingQueue.class,
                         queue -> assertThat(queue.remainingCapacity()).isEqualTo(2048));
-                assertThat(worker).extracting("spanExporter").isEqualTo(exporter);
+                assertThat(worker).extracting("spanExporter").isEqualTo(mockSpanExporter);
               });
     } finally {
       processor.shutdown();
@@ -105,7 +116,7 @@ class TracerProviderConfigurationTest {
 
     BatchSpanProcessor processor =
         TracerProviderConfiguration.configureSpanProcessor(
-            ConfigProperties.createForTest(properties), exporter);
+            ConfigProperties.createForTest(properties), mockSpanExporter);
 
     try {
       assertThat(processor)
@@ -124,7 +135,7 @@ class TracerProviderConfigurationTest {
                     .isInstanceOfSatisfying(
                         ArrayBlockingQueue.class,
                         queue -> assertThat(queue.remainingCapacity()).isEqualTo(2));
-                assertThat(worker).extracting("spanExporter").isEqualTo(exporter);
+                assertThat(worker).extracting("spanExporter").isEqualTo(mockSpanExporter);
               });
     } finally {
       processor.shutdown();
--- a/sdk/trace/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorTest.java
+++ b/sdk/trace/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorTest.java
@@ -197,12 +197,19 @@ class BatchSpanProcessorTest {
             .build();
 
     sdkTracerProvider = SdkTracerProvider.builder().addSpanProcessor(batchSpanProcessor).build();
-    for (int i = 0; i < 100; i++) {
+    for (int i = 0; i < 50; i++) {
       createEndedSpan("notExported");
     }
     List<SpanData> exported = waitingSpanExporter.waitForExport();
     assertThat(exported).isNotNull();
-    assertThat(exported.size()).isEqualTo(98);
+    assertThat(exported.size()).isEqualTo(49);
+
+    for (int i = 0; i < 50; i++) {
+      createEndedSpan("notExported");
+    }
+    exported = waitingSpanExporter.waitForExport();
+    assertThat(exported).isNotNull();
+    assertThat(exported.size()).isEqualTo(49);
 
     batchSpanProcessor.forceFlush().join(10, TimeUnit.SECONDS);
     exported = waitingSpanExporter.getExported();


sdk/trace/build.gradle.kts
```diff
@@ -45,6 +45,7 @@ dependencies {
     jmh("io.grpc:grpc-api")
     jmh("io.grpc:grpc-netty-shaded")
     jmh("org.testcontainers:testcontainers") // testContainer for OTLP collector
+    implementation("org.jctools:jctools-core:3.2.0")
```
jkwatson, line 48: We do not want to take additional external dependencies like this for our SDK.

jkwatson: If we want to use this, we'll need to shade it into our project, rather than have it be exposed as a transitive dependency.

jkwatson:
Can you include the benchmarks, before and after in this PR description, rather than just in the middle of a very long discussion on the related issue?

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -18,14 +18,15 @@
...
-import java.util.concurrent.BlockingQueue;
+import java.util.Collection;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import org.jctools.queues.MpscArrayQueue;
```
bogdandrutu, line 29: Is this better than `disruptor`? Can we run the same benchmark for https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk-extensions/async-processor ?

sbandadd: @bogdandrutu Is there an existing jmh benchmark with DisruptorAsyncSpanProcessor?

jkwatson:
Can you also include the `BatchSpanProcessorBenchmark` before & after as well?

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
bogdandrutu: Unfortunately not, but I think you can copy-paste the same one that you have and only initialize that SpanProcessor.

sbandadd:
> Can you also include the `BatchSpanProcessorBenchmark` before & after as well?

This is something I couldn't run locally (using docker-desktop on mac). I might be missing some setup required for jmh. Can you run this through the BatchSpanProcessorBenchmark and post the results?

dependencyManagement/build.gradle.kts
```diff
@@ -92,6 +92,7 @@ val DEPENDENCIES = listOf(
         "org.awaitility:awaitility:4.0.3",
         "org.codehaus.mojo:animal-sniffer-annotations:1.20",
         "org.curioswitch.curiostack:protobuf-jackson:1.2.0",
+        "org.jctools:jctools-core:3.2.0",
```
anuraaga, line 95: From looking the benchmarks, I think making the general change of signaling instead of polling makes sense to me. Let's not add this dependency in this PR though, we'll need to figure out how to vendor in only the `MpscQueue` to keep the size down if we find it to be significant (it seems like we might be but let's think about it separately)

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -128,26 +129,27 @@ public CompletableResultCode forceFlush() {
...
         int maxExportBatchSize,
         long exporterTimeoutNanos,
-        BlockingQueue<ReadableSpan> queue) {
+        MpscArrayQueue<ReadableSpan> queue) {
       this.spanExporter = spanExporter;
       this.scheduleDelayNanos = scheduleDelayNanos;
       this.maxExportBatchSize = maxExportBatchSize;
       this.exporterTimeoutNanos = exporterTimeoutNanos;
       this.queue = queue;
+      this.lock = new ReentrantLock();
```
bogdandrutu, line 151: What does the lock protect? Add `GuardedBy` annotations if any needed

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -252,8 +261,15 @@ private CompletableResultCode shutdown() {
...
-      flushRequested.compareAndSet(null, flushResult);
+      // we set the atomic here to trigger the worker loop to do a flush of the entire queue.
+      if (flushRequested.compareAndSet(null, flushResult)) {
+        lock.lock();
+        try {
+          needExport.signal();
+        } finally {
+          lock.unlock();
+        }
+      }
```
bogdandrutu, line 272: now that you grab the lock, I think it will be a big simplification if the flushRequested is protected by the lock.

sbandadd: could you clarify what would it simplify? Are you saying we can get rid of the AtomicReference? Then the exporter thread then have to take locks inside flush() as well as checking the flushRequested flag inside the lock.

jkwatson:
Benchmark results for the `BatchSpanProcessorBenchmarks`: 

Before (main branch):
image

After (this branch):
image

I would say that there really isn't a clear winner here. The zero-delay results are within the error bars, and the non-zero-delay results are mixed, at best.

sbandadd:
I don't understand the benchmark `BatchSpanProcessorBenchmark `. Why is it trying to test with forceFlush() on every iteration. Do we expect to forceFlush() after every iteration? How is this benchmark different from `BatchSpanProcessorFlushBenchmark`. Diffing the files shows only the class name is different.


```
diff BatchSpanProcessorBenchmark.java BatchSpanProcessorFlushBenchmark.java
34c34
< public class BatchSpanProcessorBenchmark {
---
> public class BatchSpanProcessorFlushBenchmark {

```

jkwatson:
> I don't understand the benchmark `BatchSpanProcessorBenchmark `. Why is it trying to test with forceFlush() on every iteration. Do we expect to forceFlush() after every iteration? How is this benchmark different from `BatchSpanProcessorFlushBenchmark`. Diffing the files shows only the class name is different.
> 
> ```
> diff BatchSpanProcessorBenchmark.java BatchSpanProcessorFlushBenchmark.java
> 34c34
> < public class BatchSpanProcessorBenchmark {
> ---
> > public class BatchSpanProcessorFlushBenchmark {
> ```
> 
> And of course forceFlush() is a major cost in that benchmark which we don't expect any improvements from this change.

It only does a `forceFlush()` after it has added all the spans have been sent through the processor to make sure that the BSP has actually handled them and they aren't stuck in the queue at the end of the process. 

I have no idea why we have that duplicated benchmark. It should probably be deleted.

sbandadd:
> It only does a forceFlush() after it has added all the spans.
Still it does do a forceFlush() on every iteration of the export(). The benchmark is not testing real scenarios.

```
  @Benchmark
  @Fork(1)
  @Threads(5)
  @Warmup(iterations = 5, time = 1)
  @Measurement(**iterations** = 10, time = 1)
  @OutputTimeUnit(TimeUnit.SECONDS)
  public void export() {
```

jkwatson:
> > It only does a forceFlush() after it has added all the spans.
> 
> Still it does do a forceFlush() on every iteration of the export(). The benchmark is not testing real scenarios.
> 
> ```
>   @Benchmark
>   @Fork(1)
>   @Threads(5)
>   @Warmup(iterations = 5, time = 1)
>   @Measurement(**iterations** = 10, time = 1)
>   @OutputTimeUnit(TimeUnit.SECONDS)
>   public void export() {
> ```

One `forceFlush()` per 1000/2000/5000 spans doesn't seem all that significant to me. But, please, by all means, write some better benchmarks and contribute them if you feel ours are not working for you.

dependencyManagement/build.gradle.kts
sbandadd: Yeah, lets get the MpscQueue in a different PR. @anuraaga Any help would be great !

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -128,26 +147,32 @@ public CompletableResultCode forceFlush() {
...
-
+    private final MpscArrayQueue<ReadableSpan> queue;
+    private final AtomicLong addedSpansCounter = new AtomicLong(0);
     private final AtomicReference<CompletableResultCode> flushRequested = new AtomicReference<>();
     private volatile boolean continueWork = true;
-    private final ArrayList<SpanData> batch;
+    private final Collection<SpanData> batch;
+    private final ReentrantLock lock;
+    public AtomicLong droppedSpansCounter = new AtomicLong(0);
+    public AtomicLong exportedSpansCounter = new AtomicLong(0);
```
sbandadd, line 158: This is a temporary add to show the benchmark results as I couldn't get the metric to work correctly.

sbandadd:
Added a benchmark with different thread configuration. Ran the BatchSpanProcessor under different versions. The benchmark clearly shows the overhead of locks/signal/polling.

- BlockingQueue: Current version. This is the baseline.
- ConcurrentQueue: Using java inbuilt ConcurrentLinkedQueue.
-  ConcurrentQueue with batch signal: Using java's inbuilt ConcurrentLinkedQueue, but sending signal to exporter every 512 adds to the queue.
- Disruptor with BlockingWait: Using Disruptor with TimeOutBlockingWaitStrategy with timeout set to scheduleDelayNanos
- Disruptor with SleepingWait: Using Disruptor with SleepingWaitStrategy() with default disruptor options.
- MPSCQueue with batch signal: Using MPSCQueue and sending signal to exporter every 512 adds to the queue.

image

All the benchmarks use the following configuration:

- maxQueueSize: 2048
- maxExportBatchSize: 512
- exportTimeOut: 5000ms
- Delay: 0ms. This is intentionally set low to surface the bottleneck, higher delays won't reveal the bottleneck since exporter is too slow.

jkwatson:
Is a 0ms exporter delay a "real-life" configuration? Sure, if export is free, then we can improve the BSP CPU usage, but if export takes any time at all, is the BSP CPU usage swamped by the export, in which case these tweaks are of limited value?

sbandadd:
I think you are missing the difference between a throughput benchmark and a cpu one. As mentioned in my previous comment, 0ms is meant for throughput benchmark and clearly shows the lock/signal/polling overhead. It also shows that forceFlush() impact on benchnmark.

As mentioned in my previous comment that I am following up with a cpu benchmark. This reflects "real world" scenario. JMH doesn't support measuring CPU time of exporter thread. So one has to use a profiler, in my case I used YJP

# Raw Data of CPU (user + kernel) time in ms
image

# Chart
image

# Throughput stays the same
image

jkwatson:
So, if I read your graphs correctly,

* ConcurrentQueue with batch signal: Using java's inbuilt ConcurrentLinkedQueue, but sending signal to exporter every 512 adds to the queue.

Is the best option that doesn't involve us depending on a 3rd party library. Do you have a PR with that implementation? Does it also do time-based exports so it will work correctly in low-throughput situations?

sbandadd:
Yes, it is java's inbuilt ConcurrentLinkedQueue. I have a local branch that use the ConcurrentLinkedQueue and yes it does time based export. We can go with this approach for now and pull in MPSCQueue later since it makes the implementation really clean and most efficient.

sdk/trace/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorBenchmark.java
```diff
@@ -33,13 +33,13 @@
 @State(Scope.Benchmark)
 public class BatchSpanProcessorBenchmark {
 
-  private static class DelayingSpanExporter implements SpanExporter {
+  public static class DelayingSpanExporter implements SpanExporter {
```
jkwatson, line 36: if we want to re-use this, please pull it up out of this class to the top level, so it's clear it's a reusable component for benchmarking.

sdk/trace/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorCpuBenchmark.java
```diff
@@ -0,0 +1,165 @@
...
+  @Warmup(iterations = 1, time = 1)
+  @Measurement(iterations = 5, time = 5)
+  @BenchmarkMode(Mode.Throughput)
+  @OutputTimeUnit(TimeUnit.SECONDS)
+  public void export_01Thread(
+      BenchmarkState benchmarkState, @SuppressWarnings("unused") ThreadState threadState) {
+    benchmarkState.numThreads = 1;
+    benchmarkState.processor.onEnd(
+        (ReadableSpan) benchmarkState.tracer.spanBuilder("span").startSpan());
+    LockSupport.parkNanos(1000_00);
```
jkwatson, line 103: let's format these like `100_000` to stick with the style in the rest of the codebase.

sdk/trace/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorCpuBenchmark.java
```diff
@@ -0,0 +1,165 @@
...
+  @Fork(1)
+  @Threads(1)
+  @Warmup(iterations = 1, time = 1)
+  @Measurement(iterations = 5, time = 5)
+  @BenchmarkMode(Mode.Throughput)
+  @OutputTimeUnit(TimeUnit.SECONDS)
+  public void export_01Thread(
+      BenchmarkState benchmarkState, @SuppressWarnings("unused") ThreadState threadState) {
+    benchmarkState.numThreads = 1;
+    benchmarkState.processor.onEnd(
```
jkwatson, line 101: so, if I read this correctly, each iteration of the benchmark only processes a single span per thread? It also includes the creation of the span by the span builder, which is going to muddy the waters about the actual BSP performance.

sdk/trace/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorCpuBenchmark.java
jkwatson: also, why are we parking for 100_000 nanos in every iteration?

sbandadd: If we don't park, then the benchmark is going to consumer all the CPU given the nature of the outer loop jmh has. This helps in measuring apples to apples.

sdk/trace/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorCpuBenchmark.java
sbandadd: I am confused on what you mean by iteration here. Do you mean jmh iterations in which case each iteration (in total 5 iterations) is running for 5 seconds and this would create as many spans as possible in that time. If you mean iteration in the sense of invocation of this logic, then yes this is going to create 1 span for BSP to process. The overhead of span creation is not a big deal as we are clearly seeing high throughput.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -128,26 +133,34 @@ public CompletableResultCode forceFlush() {
...
-    private final BlockingQueue<ReadableSpan> queue;
-
+    private final ConcurrentLinkedQueue<ReadableSpan> queue;
+    private final int maxQueueSize;
+    private final AtomicInteger queueSize = new AtomicInteger(0);
+    private final AtomicLong addedSpansCounter = new AtomicLong(0);
     private final AtomicReference<CompletableResultCode> flushRequested = new AtomicReference<>();
     private volatile boolean continueWork = true;
-    private final ArrayList<SpanData> batch;
+    private final Collection<SpanData> batch;
```
jkwatson, line 143: why change the type here? any particular reason?

sbandadd: This is a stray change from rebasing. Not necessary.

sdk/trace/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorCpuBenchmark.java
```diff
@@ -0,0 +1,165 @@
...
+        (ReadableSpan) benchmarkState.tracer.spanBuilder("span").startSpan());
+    LockSupport.parkNanos(100_000);
```
jkwatson, line 103: what is the purpose of these parks?

jkwatson: sorry, I see you answer this, but my comment disappeared so I couldn't see in-line. Please put a comment in the code about why you need these.

sbandadd: If we don't park, then the benchmark is going to consumer all the CPU given the nature of the outer loop jmh has and the results are not meaningful. This park helps maintain an expected rps, so we can compare the real CPU usage of BSP.

jkwatson:
I'm not sure how to evaluate this. The resulting code is significantly more complex, and hence harder to evaluate for correctness. I think I'd prefer having this implementation be something that is in the sdk-extensions module (tracing-incubator, probably), so that it could be used side-by-side in real production environments, rather than making these changes to the default implementations and have to revert and do a patch release if issues are found.

amitgud-doordash:
> I'm not sure how to evaluate this. The resulting code is significantly more complex, and hence harder to evaluate for correctness. I think I'd prefer having this implementation be something that is in the sdk-extensions module (tracing-incubator, probably), so that it could be used side-by-side in real production environments, rather than making these changes to the default implementations and have to revert and do a patch release if issues are found.

Is there a suggestion on adding more unit tests for assessing correctness? This PR is improving the performance by magnitudes. What can we do further evaluate the correctness other than what is already included in the PR? The current cpu overhead from the exporter is a production blocker.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -178,44 +191,58 @@ private Worker(
     }
 
     private void addSpan(ReadableSpan span) {
-      if (!queue.offer(span)) {
+      if (queueSize.get() >= maxQueueSize) {
         droppedSpans.add(1);
+      } else {
+        queue.offer(span);
+        queueSize.incrementAndGet();
+        if (addedSpansCounter.incrementAndGet() % maxExportBatchSize == 0) {
```
jkwatson, line 199: can't we use the queueSize here, rather than having to keep another counter in sync?

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -178,44 +191,58 @@ private Worker(
...
     private void flush() {
-      int spansToFlush = queue.size();
-      while (spansToFlush > 0) {
-        ReadableSpan span = queue.poll();
-        assert span != null;
-        batch.add(span.toSpanData());
-        spansToFlush--;
+      while (!queue.isEmpty()) {
+        batch.add(queue.poll().toSpanData());
+        queueSize.decrementAndGet();
```
jkwatson, line 245: Is there a reason why ithe polling and the decrementing the queuesize are done in the opposite order in this method than in the main run loop?

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
sbandadd: queue.size() is linear in complexity, so it is really bad. This is the reason I was proposing to use MPSCQueue which is much cleaner

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
sbandadd: The order doesn't matter here since queueSize is atomic and will be consistent

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
jkwatson: no, I was saying your `queueSize` variable which you're using to track the size of the queue, and could have kept the value of on the line above.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
jkwatson: then please put them in the same order, or extract a method that does both operations so it's not confusing

jkwatson:
> > I'm not sure how to evaluate this. The resulting code is significantly more complex, and hence harder to evaluate for correctness. I think I'd prefer having this implementation be something that is in the sdk-extensions module (tracing-incubator, probably), so that it could be used side-by-side in real production environments, rather than making these changes to the default implementations and have to revert and do a patch release if issues are found.
> 
> Is there a suggestion on adding more unit tests for assessing correctness? This PR is improving the performance by magnitudes. What can we do further evaluate the correctness other than what is already included in the PR? The current cpu overhead from the exporter is a production blocker.

Have you run this implementation in your production environment to compare? You say it's improving performance by magnitudes, but our existing benchmarks don't show any more throughput with your implementation than the one we already have. Yes, you created some new benchmarks that show that yours is better, but they also require running an external profiler to verify your assertions.

sbandadd:
Yes, we have run experiments in prod environment to verify the improvements. Your existing benchmark is doing a forceFlush() which makes the throughput meaningless, also the existing benchmarks are duplicate (BatchSpanProcessorBenchmark and BatchSpanProcessorFlushBenchmark) as I mentioned in my previous comments. Please try BatchSpanProcessorMultiThreadBenchmark I wrote and you could clearly see the throughput improvement.

Again throughput and CPU and two different things JMH consumes as much CPU as possible in the benchmark if allowed. So I wrote BatchSpanProcessorCpuBenchmark which makes things run in a steady state so one can verify cpu usage of exporter thread.

amitgud-doordash:
> > > I'm not sure how to evaluate this. The resulting code is significantly more complex, and hence harder to evaluate for correctness. I think I'd prefer having this implementation be something that is in the sdk-extensions module (tracing-incubator, probably), so that it could be used side-by-side in real production environments, rather than making these changes to the default implementations and have to revert and do a patch release if issues are found.
> > 
> > 
> > Is there a suggestion on adding more unit tests for assessing correctness? This PR is improving the performance by magnitudes. What can we do further evaluate the correctness other than what is already included in the PR? The current cpu overhead from the exporter is a production blocker.
> 
> Have you run this implementation in your production environment to compare? You say it's improving performance by magnitudes, but our existing benchmarks don't show any more throughput with your implementation than the one we already have. Yes, you created some new benchmarks that show that yours is better, but they also require running an external profiler to verify your assertions.

We are running this internally and our numbers concur with the data in comment https://github.com/open-telemetry/opentelemetry-java/pull/2983#issuecomment-794198484.

Existing benchmark (BatchSpanProcessorBechmark) uses `forceFlush()`, which negates the effect of the PR. Even at 5k spans, the cost adds up for a high RPS application generating multiple spans per request.

jkwatson:
I understand all of what you've said. Why are you opposed to having this as a separate incubator implementation for others to try out before we make it into the main default implementation?

sbandadd:
Also please note that throughput alone is a bad metric in the benchmark. One has to really look at the exportedSpans. For example I can write code that simply returns without adding to the queue and the throughput would be super high. So the existing BatchSpanProcessorBenchmark is even more meaningless TBH

jkwatson:
> Also please note that throughput alone is a bad metric in the benchmark. One has to really look at the exportedSpans. For example I can write code that simply returns without adding to the queue and the throughput would be super high. So the existing BatchSpanProcessorBenchmark is even more meaningless TBH

If you feel so strongly that our benchmark is meaningless (btw, please watch your language and be aware that there are human beings who wrote this code), then let's do something else. Please put in a PR that makes the benchmarks meaningful as a separate PR from also changing the implementations.

sdk/trace/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorCpuBenchmark.java
jkwatson: Please put a comment in the code about this.

sbandadd:
I am in no way talking about people who contributed to the code. We really appreciate all the work of the people who contributed here. I am only pointing out that the existing benchmark is not helpful for multiple reasons. 

1. BatchSpanProcessorBenchmark and BatchSpanProcessorFlushBenchmark are actually duplicate.
2. A forceFlush() on every loop creates a bottleneck and not creating stress on the BSP
3. Measuring only throughput is not helpful since number of spans getting exported is more appropriate. Measuring CPU usage of exporter thread is also important.

And we have contributed two benchmarks that addresses the above issues. I will look into the incubator implementation, thanks for pointing it up.

jkwatson:
> I am in no way talking about people who contributed to the code. We really appreciate all the work of the people who contributed here. I am only pointing out that the existing benchmark is not helpful for multiple reasons.
> 
> 1. BatchSpanProcessorBenchmark and BatchSpanProcessorFlushBenchmark are actually duplicate.
> 2. A forceFlush() on every loop creates a bottleneck and not creating stress on the BSP
> 3. Measuring only throughput is not helpful since number of spans getting exported is more appropriate. Measuring CPU usage of exporter thread is also important.
> 
> And we have contributed two benchmarks that addresses the above issues. I will look into the incubator implementation, thanks for pointing it up.

So, it will be very helpful to have the work separated. Let's have the benchmarks updated to be better, and have general agreement from the maintainers that it is better and more useful. Let's run the existing implementation against them once that is merged. Then, as a separate step, let's have a PR that changes the implementation and shows the change in the output of the benchmarks.

sbandadd:
Separate PR for benchmarks https://github.com/open-telemetry/opentelemetry-java/pull/3017/files

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -128,26 +133,34 @@ public CompletableResultCode forceFlush() {
...
+    @GuardedBy("lock")
+    private final Condition needExport;
 
     private Worker(
         SpanExporter spanExporter,
         long scheduleDelayNanos,
         int maxExportBatchSize,
         long exporterTimeoutNanos,
-        BlockingQueue<ReadableSpan> queue) {
+        ConcurrentLinkedQueue<ReadableSpan> queue,
```
anuraaga, line 154: I'm trying to get up to speed on thread - did you compare with `ArrayBlockingQueue`? I would be surprised if the linked list, with the allocations it needs to do would perform better and anyways we want to avoid the complicated counter increment. There is almost no work in the blocking section so even with contention it should not really put threads to sleep I guess.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -178,44 +191,58 @@ private Worker(
...
+          batch.add(queue.poll().toSpanData());
+          queueSize.decrementAndGet();
         }
-
         if (batch.size() >= maxExportBatchSize || System.nanoTime() >= nextExportTime) {
           exportCurrentBatch();
           updateNextExportTime();
         }
+        if (queue.isEmpty()) {
+          lock.lock();
```
anuraaga, line 226: Can we use a blocking queue for the signalling too? I think a one element queue is mostly equivalent but means we don't have to manage the locks ourselves, simplifying the code drastically.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
anuraaga: I'm particularly worried about the allocations since more garbage means unrelated app code ends up being affected by increased GC

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
sbandadd: queueSize changes because of exporter reading from the queue, so we cannot use this to determine whether a signal is needed. What we really want is a signal for the exporter every  maxExportBatchSize so exporter sends the entries as soon as possible.

An example with maxExportBatchSize=100 to show why queueSize is not great.

- Let's say we have 90 entries in the queue added. queueSize is 90.
- Exporter in parallel reads all the 90 entries. Now queueSize is 90.
- Exporter goes to sleep.
- We have 10 more new entries in the queue. queueSize is 10.
- If we use queueSize here, then exporter would wait for 90 more entries which introduces delay.

A simple counter that tracks total added spans is good enough.

jkwatson: We should trigger an export in only 3 cases.
1) The size of the queue has been reached (`queueSize` is tracking that here)
2) The timer has expired without an export
3) flush() is called.

I don't see how using some other number (the total number of things added to the queue throughout its entire history) meets any of these 3 criteria. That number has very little to do with whether any of those 3 criteria have been met.

sbandadd: Agreed with the three criteria except criteria (1) is not based on `queueSize` but on `batchSize`. Batch size tracks the entries in the buffer ready for export.

Triggering an export in the exporter thread and signaling exporter thread about new entries in the queue are two separate concerns. The three criteria you mentioned are being met in the exporter thread (run() function)

1. We check if the batch size reached maxExportBatchSize.
  `batch.size() >= maxExportBatchSize`
2. The timer has expired is verified in the same line as well. 
  `System.nanoTime() >= nextExportTime`
    Note that exporter thread only sleep to a max of scheduleDelayNanos to satisfy this criteria.
3. forceFlush() is sending a signal to the exporter thread with a flag, so it does flush() irrespective of the above 2 conditions.

Now coming back to the queue writer here is how criteria (1) is satisfied. Every maxExportBatchSize adds to the queue, we send a signal to the exporter thread about new entries. Exporter thread when woke up has the guarantee that it will read up maxExportBatchSize entries and trigger the export. We cannot simply use `queueSize == maxExportBatchSize` here because of the above example I pointed out with maxExportBatchSize=100.

- Let's say we have 90 entries in the queue added. queueSize is 90.
- Exporter in parallel reads all the 90 entries. Now queueSize is 90.
- Exporter goes to sleep.
- We have 10 more new entries in the queue. queueSize is 10.

If we use queueSize here, then exporter would wait for 90 more entries which introduces delay. So using a counter and signaling every maxExportBatchSize ensures criteria (1) is met asap.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
sbandadd: could you pease elaborate on what you meant by compare with `ArrayBlockingQueue`? My benchmarks took `ArrayBlockingQueue` the current implementation as the baseline and this approach showed clear improvement on throughput and CPU. Do you mean using `ArrayBlockingQueue` with batched signal? If so, yes I did ran the benchmark but it is much less throughput than using concurrent queue though there are CPU overhead improvements.

Regarding the GC, yes there would be some increased allocations. So ideally we should be pulling in MPSCQueue.

anuraaga: > Do you mean using ArrayBlockingQueue with batched signal? I

Yeah I meant this - I don't see it in the graphs, it would be nice to see the numbers to know what the real effect is. If it means simplifying the code (meaning not keeping track of the span count, for example) with a small difference, then it's still worth it especially given we may try to replace with MPSCQueue.

sbandadd: Here are the benchmark results. Throughput suffer with BlockingQueue, but exporter CPU overhead is less with batched signal. ConcurrentQueue is better overall though.

image

image

I don't see any significant gc.alloc.rate increase. The increase is well within the error boundary.

# With blocking queue batch signal
image

# With concurrent queue batch signal
image

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
anuraaga: @jkwatson This reminds me of an issue I filed a long long time ago

https://github.com/open-telemetry/opentelemetry-specification/issues/849

Never got clarification on whether we're supposed to eagerly export when we have maxExportBatchSize items. But it's true that what are current code is doing so we should preserve it for now which I think is the intent of this line.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -178,44 +210,47 @@ private Worker(
     }
 
     private void addSpan(ReadableSpan span) {
-      if (!queue.offer(span)) {
+      if (queue.size() >= maxQueueSize) {
         droppedSpans.add(1);
+      } else {
+        queue.offer(span);
```
anuraaga, line 216: I'm not comfortable with the thread-safety issues of this, it seems the queue can easily go over maxQueueSize which is an invariant we need to preserve. Since `ArrayBlockingQueue` is a queue with max capacity, let's just use it, it's defined for our job here - throughput with 0ms delay is sort of interesting theoretically but not an actual use case for this class, we can enjoy the CPU improvements when idle though thanks to the signalling which is what matters.

sbandadd: Agreed queue can go over maxQueueSize constraint temporarily. How about optimistically trying to insert into queue and backing off if maxQueueSize is reached?

```
      if (size.incrementAndGet() >= maxQueueSize) {
        size.decrementAndGet();
        return false;
      }
      queue.offer(span);
      return true
``` 

concurrentQueue does have better behavior under contention with reduced exporter cpu overhead and writer thread overhead.

If MPSCQueue is the way forward as an extension, then I feel we are good with a blocking queue and batch signaling. Is it realistic to assume a batch span processor based on MPSCQueue could be supported soon?

jkwatson:
@sbandadd would it be possible for you to attend one of our SIG meetings to discuss this all in more detail? Or jump onto CNCF slack and discuss in more real-time?

sbandadd:
Sure, I can join the SIG meetings. I joined the CNCF opentelemtety channel as well. Thanks !

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
anuraaga: @sbandadd Discussed a bit with @jkwatson and let's go ahead and split this into two parts in this order

1) Replacing constant polling with signaling
2) Improve concurrency of the queuing

We don't need to bundle them into a single PR. So can we update this one to use signaling but stick with `ArrayBlockingQueue`, which you'd demonstrated does improve CPU significantly? And in a followup we can address concurrency, most likely by vendoring in `MPSCQueue` as I don't see that being a big problem.

sbandadd: Sure thing. This sounds like a solid plan to me. Thanks for reaching a consensus !

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -191,21 +197,24 @@ public void run() {
...
+        if (queue.isEmpty()) {
+          try {
+            long pollWaitTime = nextExportTime - System.nanoTime();
+            if (pollWaitTime > 0) {
+              signal.poll(pollWaitTime, TimeUnit.NANOSECONDS);
+            }
+          } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            return;
+          }
```
anuraaga, line 216: Does this produce the same result using drain? Then I think it lets us use `queue.size() >= maxExportBatchSize()` in `addSpan` instead of maintaining a counter.

```java
while(queue.size() >= maxExportBatchSize)) {
  batch.clear();  // Can extract method for these three lines.
  queue.drainTo(batch);
  export(batch);
}
if (!queue.isEmpty() && System.nanoTime() >= nextExportTime) {
  batch.clear();
  queue.drainTo(batch);
  export(batch);
  updateNextExportTime();
} else {
  long pollWaitTime = nextExportTime - System.nanoTime();
  if (pollWaitTime > 0) {
    signal.clear();
    // Small race, queue might reach maxExportBatchSize between these two methods and stop increasing.
    // This is rare in practice and safe since after the poll interval completes the batch will be exported. This
    // behavior is not different from a processor that does not receive spans fast enough to reach the batch
    // size, which is basically what is happening during this interval of time.
    signal.poll(pollWaitTime, TimeUnit.NANOSECONDS);
  }
}
```

anuraaga: I guess the race would go away anyways if removing `signal.clear()` from that suggestion - the worst is just that it evaluates the `while` and `if` one more time before sleeping.

sbandadd: The logic looks really complicated imho. Repeatedly copying queue into the batch is bit confusing as well as the race. Using a counter we are essentially amortizing the cost of signal, it is intuitive to think that a signal every maxExportBatchSize essentially implies the exporter thread when receiving the signal is guaranteed to do the export.

sbandadd: In you suggestion, isn't export(batch) already clearing the batch? Also in the if condition `!queue.isEmpty() && System.nanoTime() >= nextExportTime`, it is exporting the whole batch at once instead of in maxExportBatchSize chunks.

anuraaga: >  it is intuitive to think that a signal every maxExportBatchSize essentially implies the exporter thread when receiving the signal is guaranteed to do the export.

But can't the signal be missed, since it's already exporting or something like that?

> In you suggestion, isn't export(batch) already clearing the batch? 

Yeah probably, it's somewhat pseudocode

> Also in the if condition !queue.isEmpty() && System.nanoTime() >= nextExportTime, it is exporting the whole batch at once instead of in maxExportBatchSize chunks.

We've already exported maxExportBatchSize chunks. However, every time the interval passed, we are supposed to clear the queue fully, so this is where remaining spans get exported - it won't export more than maxExportBatchSize even if the queue suddenly had many spans added, because `drainTo` won't fill more than `batch.size()` it's ok.

I think the loop, then if pattern here actually captures the complexity more directly. We have two conditions

- Eagerly export as many spans as possible when over maxExportBatchSize
- If interval has passed, make sure to export entire queue

So these are handled with two conditionals any time the thread wakes up, since there are actually two conditions. Does it make sense?

Let's not worry about the race, if we never clear the signal queue we're good and I think that will work well.

sbandadd: > But can't the signal be missed, since it's already exporting or something like that?

Not really, if exporter thread is doing IO then on the next iteration of the loop it copies over any span entries to batch and polls the signal queue, then eventually goes to sleep in the next iteration after eagerly copying over span entries. The guarantee the atomic counter provides is that if exporter thread is sleeping, then the next signal would guarantee that batch size would cross over matchExportBatchSize triggering IO.

> because drainTo won't fill more than batch.size() it's ok
oh so your pseudocode would really be queue.drainTo(batch, matchExportBatchSize)?

> However, every time the interval passed, we are supposed to clear the queue fully
My understanding is that current code always exports in chunks of maxExportBatchSize even when the interval is passed. That makes sense given the naming of the config options `maxExportBatchSize`


```
        try {
          ReadableSpan lastElement = queue.poll(100, TimeUnit.MILLISECONDS);
          ..

        // only one element is added to the batch in the loop, so this condition either means batch size is equal to maxExportBatchSize or  
        if (batch.size() >= maxExportBatchSize || System.nanoTime() >= nextExportTime) {
          exportCurrentBatch();
          updateNextExportTime();
        }
```


In practice queues either run close to full or close to empty. So if queue is running full at sizes >= maxExportBatchSize because exporter is slow to process IO or something, then we are unnecessarily sending signals. This won't be the case when we signal every maxExportBatchSize, it unties how fast or slow exporter really is.

sbandadd: The current code is handling the 2 conditions directly you mentioned anyway with the outer loop `while(continueWork)` and inside it is eagerly copying span entries and exporting with the condition `if (batch.size() >= maxExportBatchSize || System.nanoTime() >= nextExportTime)`.

anuraaga: > oh so your pseudocode would really be queue.drainTo(batch, matchExportBatchSize)?

Oops - yeah this is what I meant, sorry

> My understanding is that current code always exports in chunks of maxExportBatchSize even when the interval is passed.

Yeah - so the above comment hopefully clarifies, I meant we never drain more than batch size. But to clarify, it's a bit incorrect to say "always exports in chunks of maxExportBatchSize even when the interval is passed" - it could be a chunk of less than maxExportBatchSize too if interval has passed or else it's possible spans are just never exported.

> then we are unnecessarily sending signals.

I guess I'd still err on oversignaling - perhaps the above point that it's possible to send chunks with size less than the batch size clarifies my worry, that this `%` can get out-of-sync with reality when smaller chunks are sent and it becomes unclear what's going on. If we find the oversignalling to be much, we can follow up with a CAS guard too?

> inside it is eagerly copying span entries and exporting with the condition

I guess I'm mostly worried that the check is on the queue being empty `if (queue.isEmpty()) {`

If the queue is never empty, the BSP thread never cooperatively sleeps for other threads. Maybe it's ok but draining seems like it could reduce the overall time the BSP thread is awake.

sbandadd: > chunk of less than maxExportBatchSize
True, I didn't clarify the "or less" part. 

> this % can get out-of-sync with reality when smaller chunks are sent and it becomes unclear what's going on
The current code handles sending chunks with size less than  maxExportBatchSize when interval is passed. I am not sure sure what becomes unclear here?

> If the queue is never empty, the BSP thread never cooperatively sleeps for other threads.
I don't think this would be a concern for two reasons. 1. Exporter thread would either do the IO and get scheduled out by the kernel 2. If exporter is fast then queue would be empty for sure.

> Maybe it's ok but draining seems like it could reduce the overall time the BSP thread is awake.
With drain we would need another array of type ArrayList<ReadableSpan> which makes things bit more cumbersome.

It feels like we would need some experiment data to really see if there are any material impact on the state of exporter thread.

anuraaga: > The current code handles sending chunks with size less than maxExportBatchSize when interval is passed. I am not sure what becomes unclear here?

If I'm not mistaken, then if a chunk is sent that isn't `maxExportBatchSize` then the `%` becomes out of sync. The intention of this `%` is to send chunks with size `maxExportBatchSize` but if any batch is sent with a different size due to the interval check, then when the signal is sent, we will have less or more than `maxExportBatchSize` in an unclear way. For example if batch size is 5, and 4 spans are sent due to the interval check, then the signal will happen with only 1 span in the queue. I don't think that's our intention with that signal.

sbandadd: I updated the PR to just use an atomic boolean and queue size. I think it will make things much more clear though I didn't see any change in benchmark results.

drainTo() holds up queue lock for a long time which negatively impacts the writer threads.

jkwatson: I like this latest iteration quite a bit. Unfortunately, it looks like `BatchSpanProcessorTest.forceExport()` is quite flaky with this implementation. It failed about 1/5 times for me locally. I'm not sure what's going on.

sbandadd: @jkwatson could you paste the error? I ran the test 10 times and couldn't reproduce locally.

jkwatson: It just fails the first assertion with a '0', rather than a '49'.

anuraaga (approved):
This seems like a relatively small change with a significant improvement in signaling strength. Thanks!

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -131,7 +132,7 @@ public CompletableResultCode forceFlush() {
 
     private long nextExportTime;
     private final BlockingQueue<ReadableSpan> queue;
-    private final AtomicBoolean needSignal = new AtomicBoolean(false);
+    private final AtomicLong spansNeeded = new AtomicLong(Long.MAX_VALUE);
```
jkwatson, line 135: should this be initialized to `maxExportBatchSize`?

sbandadd: I am using MAX_VALUE to imply that exporter thread is not expecting any signal. When it needs a signal, it is updating the value to number of spans it requires before doing the IO. Since in the beginning exporter thread is not expecting any signal, setting it to MAX_VALUE seems reasonable.

jkwatson: cool. can you add a comment in the code about that, so when we look at this in 3 months, we don't have to dig through PRs to figure out why?  :)

sbandadd:
I made a minor update to use a AtomicInteger instead of AtomicLong.

jkwatson (approved):
Thanks for the hard work on this!

anuraaga:
Thanks @sbandadd!

PR #2984, opened 2021-03-04
Implement a leak-detecting SpanProcessor using the WeakConcurrentMap …

### LINKED ISSUE #1084: Create a span processor that detects unfinished spans and warns about them.
See #1074

jkwatson:
…from the context internals.

Resolves #1084

### DIFF
--- a/sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/LeakDetectingSpanProcessor.java
+++ b/sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/LeakDetectingSpanProcessor.java
@@ -1,0 +1,156 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.extension.incubator.trace;
+
+import static java.lang.Thread.currentThread;
+
+import io.opentelemetry.api.trace.SpanContext;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.internal.shaded.WeakConcurrentMap;
+import io.opentelemetry.sdk.trace.ReadWriteSpan;
+import io.opentelemetry.sdk.trace.ReadableSpan;
+import io.opentelemetry.sdk.trace.SpanProcessor;
+import java.lang.ref.Reference;
+import java.util.Arrays;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.BiConsumer;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * A {@link SpanProcessor} which will detect spans that are never ended. It will both detect spans
+ * that are garbage collected without ever having `end()` called on them.
+ *
+ * <p>Note: using this SpanProcessor will definitely impact the performance of your application. It
+ * is not recommended for production use, as it uses additional memory for each span to track
+ * whether the span has been ended.
+ */
+public class LeakDetectingSpanProcessor implements SpanProcessor {
+  private static final Logger logger = Logger.getLogger(LeakDetectingSpanProcessor.class.getName());
+
+  private final PendingSpans pendingSpans;
+
+  /**
+   * Create a new {@link LeakDetectingSpanProcessor} that will report any un-ended spans that get
+   * garbage collected.
+   */
+  public static LeakDetectingSpanProcessor create() {
+    return new LeakDetectingSpanProcessor();
+  }
+
+  private LeakDetectingSpanProcessor() {
+    this(
+        (s, throwable) ->
+            logger.log(
+                Level.WARNING,
+                "Span garbage collected before being ended.",
+                callerError((CallerStackTrace) throwable)));
+  }
+
+  // Visible for testing
+  LeakDetectingSpanProcessor(BiConsumer<String, Throwable> reporter) {
+    pendingSpans = PendingSpans.create(reporter);
+  }
+
+  @Override
+  public void onStart(Context parentContext, ReadWriteSpan span) {
+    CallerStackTrace caller = new CallerStackTrace(span);
+    StackTraceElement[] stackTrace = caller.getStackTrace();
+
+    // take off the first 3 stack frames, as they are from the SDK itself.
+    caller.setStackTrace(
+        Arrays.copyOfRange(stackTrace, Math.min(3, stackTrace.length), stackTrace.length));
+
+    pendingSpans.put(span.getSpanContext(), caller);
+  }
+
+  @Override
+  public boolean isStartRequired() {
+    return true;
+  }
+
+  @Override
+  public void onEnd(ReadableSpan span) {
+    pendingSpans.remove(span.getSpanContext()).ended = true;
+  }
+
+  @Override
+  public boolean isEndRequired() {
+    return true;
+  }
+
+  private static class PendingSpans extends WeakConcurrentMap<SpanContext, CallerStackTrace> {
+
+    private final ConcurrentHashMap<WeakKey<SpanContext>, CallerStackTrace> map;
+    private final BiConsumer<String, Throwable> reporter;
+
+    @SuppressWarnings("ThreadPriorityCheck")
+    private static PendingSpans create(BiConsumer<String, Throwable> reporter) {
+      PendingSpans pendingSpans = new PendingSpans(new ConcurrentHashMap<>(), reporter);
+      // Start cleaner thread ourselves to make sure it runs after initializing our fields.
+      Thread thread = new Thread(pendingSpans);
+      thread.setName("weak-ref-cleaner-leakingspandetector");
+      thread.setPriority(Thread.MIN_PRIORITY);
+      thread.setDaemon(true);
+      thread.start();
+      return pendingSpans;
+    }
+
+    private PendingSpans(
+        ConcurrentHashMap<WeakKey<SpanContext>, CallerStackTrace> map,
+        BiConsumer<String, Throwable> reporter) {
+      super(/* cleanerThread= */ false, /* reuseKeys= */ false, map);
+      this.map = map;
+      this.reporter = reporter;
+    }
+
+    // Called by cleaner thread.
+    @Override
+    public void run() {
+      try {
+        while (!Thread.interrupted()) {
+          // call blocks until something is GC'd.
+          Reference<? extends SpanContext> gcdReference = remove();
+          CallerStackTrace caller = map.remove(gcdReference);
+          if (caller != null && !caller.ended) {
+            reporter.accept("Span garbage collected before being ended.", callerError(caller));
+          }
+        }
+      } catch (InterruptedException ignored) {
+        // do nothing
+      }
+    }
+  }
+
+  private static class CallerStackTrace extends Throwable {
+
+    private static final long serialVersionUID = 1234567896L;
+
+    final String threadName = currentThread().getName();
+    final String spanInformation;
+
+    volatile boolean ended;
+
+    CallerStackTrace(ReadableSpan span) {
+      super("Thread [" + currentThread().getName() + "] started span : " + span + " here:");
+      this.spanInformation = span.getName() + " [" + span.getSpanContext() + "]";
+    }
+  }
+
+  private static AssertionError callerError(CallerStackTrace caller) {
+    // Sometimes unit test runners truncate the cause of the exception.
+    // This flattens the exception as the caller of end() isn't important vs the one that leaked
+    AssertionError toThrow =
+        new AssertionError(
+            "Span garbage collected before being ended. Thread: ["
+                + caller.threadName
+                + "] started span : "
+                + caller.spanInformation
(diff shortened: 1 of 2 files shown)


sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/LeakDetectingSpanProcessor.java
```diff
@@ -0,0 +1,153 @@
...
+import io.opentelemetry.sdk.trace.SpanProcessor;
+import java.lang.ref.Reference;
+import java.util.Arrays;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.BiConsumer;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * A {@link SpanProcessor} which will detect spans that are never ended. It will both detect spans
```
anuraaga, line 24:
```suggestion
 * A {@link SpanProcessor} which will detect spans that are never ended. It will detect spans
```

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/LeakDetectingSpanProcessor.java
```diff
@@ -0,0 +1,153 @@
...
+ * that are garbage collected without ever having `end()` called on them.
+ *
+ * <p>Note: using this SpanProcessor will definitely impact the performance of your application. It
+ * is not recommended for production use, as it uses additional memory for each span to track
+ * whether the span has been ended.
```
anuraaga, line 29:
```suggestion
 * where a leaked span was created.
```

The reality is `WeakConcurrentMap` is pretty low overhead - the main overhead of this / context debugger is tracking the stack traces.

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/LeakDetectingSpanProcessor.java
```diff
@@ -0,0 +1,153 @@
...
+ */
+public class LeakDetectingSpanProcessor implements SpanProcessor {
```
anuraaga, line 31:
```suggestion
public final class LeakDetectingSpanProcessor implements SpanProcessor {
```

sdk-extensions/tracing-incubator/src/test/java/io/opentelemetry/sdk/extension/incubator/trace/LeakDetectingSpanProcessorTest.java
```diff
@@ -0,0 +1,48 @@
...
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+public class LeakDetectingSpanProcessorTest {
+
+  @Test
+  void garbageCollectedScope() {
+    List<Throwable> logs = new ArrayList<>();
```
anuraaga, line 22: Could use `logunit`?

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/LeakDetectingSpanProcessor.java
```diff
@@ -0,0 +1,153 @@
...
+
+    CallerStackTrace(ReadableSpan span) {
+      super("Thread [" + currentThread().getName() + "] started span : " + span + " here:");
+      this.spanInformation = span.getName() + " [" + span.getSpanContext() + "]";
+    }
+  }
+
+  private static AssertionError callerError(CallerStackTrace caller) {
+    // Sometimes unit test runners truncate the cause of the exception.
+    // This flattens the exception as the caller of end() isn't important vs the one that leaked
```
anuraaga, line 142: This comment doesn't apply, `end()` isn't actually called at all - the context detection has the two duties of checking leaked scopes, and checking scopes closed on the wrong thread.

anuraaga (commented):
I was definitely hoping the weak map would be used for this so this is sweet :)

anuraaga:
https://app.codecov.io/gh/open-telemetry/opentelemetry-java/compare/2984/tree/sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/LeakDetectingSpanProcessor.java

Looks like getting coverage on `isStart/EndRequired` mostly for the heck of it, and testing nothing is logged for a GC'd span that was ended, would be nice

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/LeakDetectingSpanProcessor.java
jkwatson: yeah, but those stack traces definitely add a lot of memory overhead per span, unfortunately.

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/LeakDetectingSpanProcessor.java
jkwatson: will remove; thanks

sdk-extensions/tracing-incubator/src/test/java/io/opentelemetry/sdk/extension/incubator/trace/LeakDetectingSpanProcessorTest.java
jkwatson: I originally had logunit and @breedx-splk suggested doing it this way. I think the idea was that we could easily make providing the consumer part of the public API for use-cases that didn't want to just log, but wanted to gather the raw data and do something other than logging with it.

Honestly, I'm not sure it ends up mattering all that much, but will switch back to making assertions about logs if you think that's better.

jkwatson:
> https://app.codecov.io/gh/open-telemetry/opentelemetry-java/compare/2984/tree/sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/LeakDetectingSpanProcessor.java
> 
> Looks like getting coverage on `isStart/EndRequired` mostly for the heck of it, and testing nothing is logged for a GC'd span that was ended, would be nice

Do you have any ideas on how to test for the negative case without requiring waiting a long time to make sure nothing gets through?

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/LeakDetectingSpanProcessor.java
```diff
@@ -0,0 +1,151 @@
...
+
+/**
+ * A {@link SpanProcessor} which will detect spans that are never ended. It will detect spans that
+ * are garbage collected without ever having `end()` called on them.
+ *
+ * <p>Note: using this SpanProcessor will definitely impact the performance of your application. It
+ * is not recommended for production use, as it uses additional memory for each span to track where
+ * a leaked span was created.
+ */
+public final class LeakDetectingSpanProcessor implements SpanProcessor {
```
bogdandrutu, line 31: Could then this be part of the "testing" artifact? maybe for the moment just have it under a testing package?

anuraaga:
> Do you have any ideas on how to test for the negative case without requiring waiting a long time to make sure nothing gets through?

I think you can use guava-testlib's `GcFinalization` 

```java
WeakReference<Span> spanRef = new WeakReference<>(tracer.spanBuilder("life is short").startSpan());
GcFinalization.awaitClear(spanRef);
```

If there's a bug we'd be guaranteed to have had a log statement after that I believe.

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/LeakDetectingSpanProcessor.java
anuraaga: Since this is in the incubator, an artifact which will never have it's `-alpha` removed, I think it's fine here.

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/LeakDetectingSpanProcessor.java
```diff
@@ -0,0 +1,151 @@
...
+  public void onEnd(ReadableSpan span) {
+    pendingSpans.remove(span.getSpanContext()).ended = true;
+  }
+
+  @Override
+  public boolean isEndRequired() {
+    return true;
+  }
+
+  private static class PendingSpans extends WeakConcurrentMap<SpanContext, CallerStackTrace> {
```
anuraaga, line 82: Just noticed - I think this should be `<Span, CallerStackTrace>`. It'd be rare but `SpanContext` could stick around longer than `Span` which is what actually matters, so may as well be precise. If we wanted to implement something like Zipkin's "flush on GC" mode in the future, we'd need to store a reference to `SpanContext` in `CallerStackTrace` and `WeakConcurrentMap<SpanContext, SomethingWithSpanContext>` will prevent GC of the spancontext forever.

jkwatson: Good point. I couldn't decide what the right level to point to here. I chose spancontext because it references fewer things, but that wasn't a strong reason. 

I'll adjust it.

jkwatson: Hmm. A problem with this is that `ReadableSpan` is not a `Span`. I'd have to cast to `Span` to handle `onEnd()` properly. Does that seem ok?

jkwatson: I guess I can use `ReadableSpan` as the type for the reference, as the top of the interface hierarchy.

jkwatson: done

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/LeakDetectingSpanProcessor.java
```diff
@@ -0,0 +1,150 @@
...
+
+  /**
+   * Create a new {@link LeakDetectingSpanProcessor} that will report any un-ended spans that get
+   * garbage collected.
+   */
+  public static LeakDetectingSpanProcessor create() {
+    return new LeakDetectingSpanProcessor();
+  }
+
+  private LeakDetectingSpanProcessor() {
```
anuraaga, line 43: Ah - don't think we need this constructor since we have the factory :)

sdk-extensions/tracing-incubator/src/test/java/io/opentelemetry/sdk/extension/incubator/trace/LeakDetectingSpanProcessorTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+    Span testSpan = tracer.spanBuilder("testSpan").startSpan();
+    WeakReference<Span> spanRef = new WeakReference<>(testSpan);
+
+    testSpan.end();
+    testSpan = null;
+
+    GcFinalization.awaitClear(spanRef);
+
+    System.gc();
```
anuraaga, line 77: Don't think you should need this, `GcFinalization` is essentially calling this over and over until `spanRef.get() == null`

anuraaga (approved):

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/LeakDetectingSpanProcessor.java
jkwatson: fair enough!

jkwatson: inlined

sdk-extensions/tracing-incubator/src/test/java/io/opentelemetry/sdk/extension/incubator/trace/LeakDetectingSpanProcessorTest.java
jkwatson: removed

PR #2985, opened 2021-03-05
Ignore null / invalid spancontext and handle null attributes for links.

### LINKED ISSUE #2955: Adding an invalid span as a link
**Describe the bug**
What should happen when you add an invalid span (traceId=0, spanId=0) as a link?
Currently such a link is added to the span and even exported (I haven't tested it, just read the code: jaeger & otlp exporters do not filter links at all, nor does SDK). That's quite different to how we treat parent contexts - where we check if the context contains a valid span (in `SdkSpanBuilder#startSpan()`).
Is the current behavior correct?

anuraaga:
Fixes #2955

Spec PR :https://github.com/open-telemetry/opentelemetry-specification/pull/1492

### DIFF
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkSpanBuilder.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkSpanBuilder.java
@@ -83,12 +83,24 @@ final class SdkSpanBuilder implements SpanBuilder {
 
   @Override
   public SpanBuilder addLink(SpanContext spanContext) {
+    if (spanContext == null || !spanContext.isValid()) {
+      return this;
+    }
     addLink(LinkData.create(spanContext));
     return this;
   }
 
   @Override
   public SpanBuilder addLink(SpanContext spanContext, Attributes attributes) {
+    if (spanContext == null || !spanContext.isValid()) {
+      return this;
+    }
+    if (!spanContext.isValid()) {
+      return this;
+    }
+    if (attributes == null) {
+      attributes = Attributes.empty();
+    }
     int totalAttributeCount = attributes.size();
     addLink(
         LinkData.create(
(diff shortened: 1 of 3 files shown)


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkSpanBuilder.java
```diff
@@ -83,12 +83,24 @@ public SpanBuilder setSpanKind(SpanKind spanKind) {
...
     addLink(LinkData.create(spanContext));
     return this;
   }
 
   @Override
   public SpanBuilder addLink(SpanContext spanContext, Attributes attributes) {
+    if (spanContext == null || !spanContext.isValid()) {
+      return this;
+    }
+    if (!spanContext.isValid()) {
```
jkwatson, line 98: this is redundant with the check above, correct?

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkSpanBuilder.java
```diff
@@ -83,12 +83,24 @@ public SpanBuilder setSpanKind(SpanKind spanKind) {
...
   public SpanBuilder addLink(SpanContext spanContext) {
+    if (spanContext == null || !spanContext.isValid()) {
+      return this;
+    }
     addLink(LinkData.create(spanContext));
     return this;
   }
 
   @Override
   public SpanBuilder addLink(SpanContext spanContext, Attributes attributes) {
```
bogdandrutu, line 94: Should we also document in the API, same as you din in the specs:

"Implementation MAY ignore ...."

anuraaga: Sure

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkSpanBuilder.java
anuraaga: Oops

bogdandrutu (approved):

jkwatson (approved):

PR #2988, opened 2021-03-05
Add static wrappers for Executor / ExecutorService using current cont…

anuraaga:
…ext at invocation time.

Some asynchronous instrumentation requires registering an `ExecutorService` statically which propagates context from invocations to executions. For example, to properly trace `OkHttpClient` to support `enqueue`, we need to do something like this

```java
client
  .dispatcher(new Dispatcher(Context.currentContextWrapping(new Dispatcher().executorService())))
  .addInterceptor(OkHttpTracing.create(openTelemetry).newInterceptor())
```

I thought of names like `currentContextWrapping`, `wrapWithCurrentContext` but don't know if I like either name so suggestions appreciated.

### DIFF
--- a/context/src/main/java/io/opentelemetry/context/Context.java
+++ b/context/src/main/java/io/opentelemetry/context/Context.java
@@ -109,7 +109,7 @@ public interface Context {
    * dbExecutor = Context.wrapTasks(threadPool)} to ensure calls like {@code dbExecutor.execute(()
    * -> database.query())} have {@link Context} available on the thread executing database queries.
    */
-  static Executor wrapTasks(Executor executor) {
+  static Executor taskWrapping(Executor executor) {
     return command -> executor.execute(Context.current().wrap(command));
   }
 
@@ -125,7 +125,7 @@ public interface Context {
    * dbExecutor.execute(() -> database.query())} have {@link Context} available on the thread
    * executing database queries.
    */
-  static ExecutorService wrapTasks(ExecutorService executorService) {
+  static ExecutorService taskWrapping(ExecutorService executorService) {
     return new CurrentContextExecutorService(executorService);
   }
 
--- a/context/src/main/java/io/opentelemetry/context/Context.java
+++ b/context/src/main/java/io/opentelemetry/context/Context.java
@@ -99,6 +99,19 @@ public interface Context {
     return ArrayBasedContext.root();
   }
 
+  static Executor currentContextWrapping(Executor executor) {
+    return command -> executor.execute(Context.current().wrap(command));
+  }
+
+  /**
+   * Returns an {@link ExecutorService} which delegates to the provided {@code executorService},
+   * wrapping all invocations with the {@linkplain Context#current() current context} at the time of
+   * invocation.
+   */
+  static ExecutorService currentContextWrapping(ExecutorService executorService) {
+    return new CurrentContextExecutorService(executorService);
+  }
+
   /**
    * Returns the value stored in this {@link Context} for the given {@link ContextKey}, or {@code
    * null} if there is no value for the key in this context.
--- a/context/src/main/java/io/opentelemetry/context/Context.java
+++ b/context/src/main/java/io/opentelemetry/context/Context.java
@@ -101,18 +101,31 @@ public interface Context {
 
   /**
    * Returns an {@link Executor} which delegates to the provided {@code executor}, wrapping all
-   * invocations with the {@linkplain Context#current() current context} at the time of invocation.
+   * invocations of {@link Executor#execute(Runnable)} with the {@linkplain Context#current()
+   * current context} at the time of invocation.
+   *
+   * <p>This is generally used to create an {@link Executor} which will forward the {@link Context}
+   * during an invocation to another thread. For example, you may use something like {@code Executor
+   * dbExecutor = Context.wrapTasks(threadPool)} to ensure calls like {@code dbExecutor.execute(()
+   * -> database.query())} have {@link Context} available on the thread executing database queries.
    */
-  static Executor currentContextWrapping(Executor executor) {
+  static Executor wrapTasks(Executor executor) {
     return command -> executor.execute(Context.current().wrap(command));
   }
 
   /**
    * Returns an {@link ExecutorService} which delegates to the provided {@code executorService},
-   * wrapping all invocations with the {@linkplain Context#current() current context} at the time of
-   * invocation.
+   * wrapping all invocations of {@link ExecutorService} methods such as {@link
+   * ExecutorService#execute(Runnable)} or {@link ExecutorService#submit(Runnable)} with the
+   * {@linkplain Context#current() current context} at the time of invocation.
+   *
+   * <p>This is generally used to create an {@link ExecutorService} which will forward the {@link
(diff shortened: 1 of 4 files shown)


jkwatson (approved):
This seems like a reasonable addition to the API to me. @bogdandrutu any thoughts?

kenfinnigan:
Looked over this without any knowledge of prior discussions, apologies if I'm missing context.

My initial thought is there's a lot of nearly identical code and whether there's another way to approach it. For instance, could `ForwardingExectorService` actually be a `ContextWrappedExecutorService` where `Context` is added to every method? Granted this creates lots of `protected` methods, which may not be ideal either. `ContextExecutorService` passes in the context from the instance, and `CurrentContextExecutorService` passes in `Context.current()` to each call?

An alternative to `currentContextWrapping()` could be `executeWithCurrentContext()` or `executeInsideCurrentContext()`?

anuraaga:
> For instance, could ForwardingExectorService actually be a ContextWrappedExecutorService where Context is added to every method?

I'm not sure I follow the suggestion but I think this may not work mostly because it's not the methods that need context, it's the callbacks that need to be wrapped. But if there are ways to reduce duplication in these internal classes it'd be great to follow up on it :-)

> executeWithCurrentContext()

I think the problem I had with method names with current context on the right side is it sounded even more so that the context used is what's current when calling this method itself, not what's current when the b executor is invoked.

Another idea that came to mind which I like now is `executingWithInvocationContext` - it's clear at the expense of a new concept, invocation context, but that's needed to understand what's happening anyways I guess.

kenfinnigan:
> > For instance, could ForwardingExectorService actually be a ContextWrappedExecutorService where Context is added to every method?
> 
> I'm not sure I follow the suggestion but I think this may not work mostly because it's not the methods that need context, it's the callbacks that need to be wrapped. But if there are ways to reduce duplication in these internal classes it'd be great to follow up on it :-)

I can certainly put together a PR with the idea once this is in place.

> > executeWithCurrentContext()
> 
> I think the problem I had with method names with current context on the right side is it sounded even more so that the context used is what's current when calling this method itself, not what's current when the b executor is invoked.

I think I'm misunderstanding something, as I thought what you describe is what it does. The current context is passed in such a way that it's active during the callback on the Executor.

Any help in understanding where I went wrong would be appreciated.

> Another idea that came to mind which I like now is `executingWithInvocationContext` - it's clear at the expense of a new concept, invocation context, but that's needed to understand what's happening anyways I guess.

`InvocationContext` often refers to an object that was used to invoke some code, which doesn't appear to fit. It's also not obvious from the method name it's utilizing the current context.

anuraaga:
> I think I'm misunderstanding something, as I thought what you describe is what it does. The current context is passed in such a way that it's active during the callback on the Executor.

The context that should be used isn't the one that is current when `executeWithCurrentContext` is called - it should be one that is current when e.g., `ExecutorService.execute`, `ExecutorService.submit` are called. But this method name feels like it is talking about the current context when `executeWithCurrentContext` is called, at least to me.

kenfinnigan:
Thanks @anuraaga, I'd missed that.

I agree that the names I've suggested don't really fit the delayed execution meaning. I think one of your original ideas works, `wrapWithCurrentContext`.

context/src/main/java/io/opentelemetry/context/CurrentContextExecutorService.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+class CurrentContextExecutorService extends ForwardingExecutorService {
```
bogdandrutu, line 17: final :)

context/src/main/java/io/opentelemetry/context/ForwardingExecutorService.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
+
+/** A {@link ExecutorService} that implements methods that don't need {@link Context}. */
+abstract class ForwardingExecutorService implements ExecutorService {
```
bogdandrutu, line 16: Not sure if this is needed, I mean it is internal class so no strong opinion, but I think is overeng :)

context/src/main/java/io/opentelemetry/context/Context.java
```diff
@@ -99,6 +99,23 @@ static Context root() {
     return ArrayBasedContext.root();
   }
 
+  /**
+   * Returns an {@link Executor} which delegates to the provided {@code executor}, wrapping all
+   * invocations with the {@linkplain Context#current() current context} at the time of invocation.
+   */
+  static Executor currentContextWrapping(Executor executor) {
```
bogdandrutu, line 106: Naming is the hardest problem in eng, maybe consider `wrapCurrent` just in case we will add equivalent for wrapping with current Runnable/Callable and would be also consistent with current `wrap` helpers?

context/src/main/java/io/opentelemetry/context/Context.java
```diff
@@ -99,6 +99,23 @@ static Context root() {
     return ArrayBasedContext.root();
   }
 
+  /**
```
bogdandrutu, line 102: Maybe worth documenting the difference between this and the other wrap equivalent.

context/src/main/java/io/opentelemetry/context/Context.java
anuraaga: > maybe consider wrapCurrent just in case we will add equivalent for wrapping with current Runnable/Callable and would be also consistent with current wrap helpers?

I don't think we'll be adding those since it's very easy to call `Context.current().wrap` but even if we did, I precisely don't want consistency with it :) This method uses the current context when calling ExecutorService.execute, not when calling this method so it's quite different.

anuraaga:
@kenfinnigan Any thoughts on this other idea I just came up with, `wrapTasks`? We're wrapping tasks given to the executor with the current context.

context/src/main/java/io/opentelemetry/context/Context.java
anuraaga: Added some more detailed docs

trask:
I'm good with `wrapTasks`, just throwing out a couple others as options:

* `Context.propagating(Executor)`
* `Context.wrapping(Executor)`
* `Context.taskWrapping(Executor)`

anuraaga:
@trask Thanks, liked `taskWrapping` best so far.

kenfinnigan:
Ooh, I really like `taskWrapping`

anuraaga:
@bogdandrutu This look ok? Will go ahead and merge soon

bogdandrutu:
Throwing more ideas:

```
Context.forwarding()
```

I am also ok with `taskWrapping`

anuraaga:
Sounds like `taskWrapping` has multiple votes so going to go with it. Thanks!

PR #2989, opened 2021-03-05
Have SimpleSpanProcessor keep track of pending exports and wait for t…

anuraaga:
…hem when flushed.

### DIFF
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/SimpleSpanProcessor.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/SimpleSpanProcessor.java
@@ -15,6 +15,8 @@ import io.opentelemetry.sdk.trace.SpanProcessor;
 import io.opentelemetry.sdk.trace.data.SpanData;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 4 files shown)


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/SimpleSpanProcessor.java
```diff
@@ -31,6 +33,7 @@
 
   private final SpanExporter spanExporter;
   private final boolean sampled;
+  private final Map<CompletableResultCode, Boolean> pendingExports = new ConcurrentHashMap<>();
```
jkwatson, line 36: this could be `Map<CompletableResultCode, Void>`, right?

jkwatson (approved):
one teeny optimization could be made, but looks good to me.

bogdandrutu (commented):
I think this exporter becomes very expensive, and may worth documenting that in the javadoc. I would not recommend this exporter to be used in real production environment, probably even on serverless env is better to use batch + flush.

Not blocking PR, just want some clarification on the javadoc to explain to the users.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/SimpleSpanProcessor.java
anuraaga: I had forgotten about `newSetFromMap`, oops

anuraaga:
@bogdandrutu I don't think this processor has gotten much more expensive with this PR, a `ConcurrentHashMap` is generally considered "mostly free" in Java. ZPages use it and some atomic structures but I don't think we would discourage it in production for performance reasons.

Unrelated to this PR though, I agree that it will perform very poorly for almost all exporters due to backends not doing well without at least some batching. So I added some docs related to that, does it look ok?

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/SimpleSpanProcessor.java
anuraaga: For reference, `newSetFromMap` uses `Map<?, Boolean>` too - it's because `Void` can only be `null`, but Maps behave differently in their support for `null` values, in particular our best friend CHM does not support them.

bogdandrutu (approved):

PR #2990, opened 2021-03-05
Document depending on SDK in README instead of API.

anuraaga:
I have a hunch that more end users will be reading this than instrumenters, at least initially, so we should document depending on the SDK. It might not be a correct hunch though so no need to merge this but interested in what people think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ dependency versions in sync.
   <dependencies>
     <dependency>
       <groupId>io.opentelemetry</groupId>
-      <artifactId>opentelemetry-api</artifactId>
+      <artifactId>opentelemetry-sdk</artifactId>
     </dependency>
   </dependencies>
 </project>
@@ -83,7 +83,7 @@ dependency versions in sync.
 ```groovy
 dependencies {
   implementation platform("io.opentelemetry:opentelemetry-bom:1.0.0")
-  implementation('io.opentelemetry:opentelemetry-api')
+  implementation('io.opentelemetry:opentelemetry-sdk')
 }
 ```
 
@@ -116,7 +116,7 @@ We strongly recommend using our published BOM to keep all dependency versions in
   <dependencies>
     <dependency>
       <groupId>io.opentelemetry</groupId>
-      <artifactId>opentelemetry-api</artifactId>
+      <artifactId>opentelemetry-sdk</artifactId>
     </dependency>
   </dependencies>
 </project>
@@ -131,12 +131,11 @@ repositories {
 
 dependencies {
   implementation platform("io.opentelemetry:opentelemetry-bom:1.1.0-SNAPSHOT")
-  implementation('io.opentelemetry:opentelemetry-api')
+  implementation('io.opentelemetry:opentelemetry-sdk')
 }
 ```
 
-Libraries will usually only need `opentelemetry-api`, while applications
-may want to use `opentelemetry-sdk`.
+Libraries will usually only need `opentelemetry-api`, so should not depend on `opentelemetry-sdk`.
 
 ## Releases
 


jkwatson:
I'm definitely torn on this one. I don't have a good idea of who will be using the API/SDK on its own, actually. I would hope that most people would care about the auto-instrumentation agent, and not have to worry about this low-level library at all. I wonder if we should also include a strong recommendation to head over to the instrumentation repo, rather than try to make this repo the main landing point for users.

README.md
```diff
@@ -131,12 +131,11 @@ repositories {
...
 dependencies {
   implementation platform("io.opentelemetry:opentelemetry-bom:1.1.0-SNAPSHOT")
-  implementation('io.opentelemetry:opentelemetry-api')
+  implementation('io.opentelemetry:opentelemetry-sdk')
 }
 ```
 
-Libraries will usually only need `opentelemetry-api`, while applications
-may want to use `opentelemetry-sdk`.
+Libraries will usually only need `opentelemetry-api`, so should not depend on `opentelemetry-sdk`.
```
bogdandrutu, line 138: I like this picture https://github.com/open-telemetry/opentelemetry-specification/blob/main/internal/img/long-term-support.png maybe we can reuse it to show when to depend on which artifact

bogdandrutu (approved):

bogdandrutu:
@jkwatson 

>  I don't have a good idea of who will be using the API/SDK on its own, actually.

Integration like spring/grpc will definitely use this directly via API. Everyone who does not want to use auto-instrumentation will use them correct?

jkwatson:
> > I don't have a good idea of who will be using the API/SDK on its own, actually.
> 
> Integration like spring/grpc will definitely use this directly via API. Everyone who does not want to use auto-instrumentation will use them correct?

I think so, yes. But, I do see a lot of questions coming from people writing custom instrumentation, so I'm not sure we have a great handle on the user personas yet.

iNikem:
I am against this change. The majority of users should depend on API. @anuraaga What are these "instrumenters" who should depend on SDK?

anuraaga:
> What are these "instrumenters" who should depend on SDK?

Opposite - when you instrument a library, you only depend on an API. But if you're developing an app as an end user, you need the SDK or you don't get any traces at all.

anuraaga:
One thing to point out - with 1.0 we're starting to get lots of feedback from users. I think we've seen them all to be app developers that need the SDK rather than the API.

> I wonder if we should also include a strong recommendation to head over to the instrumentation repo, rather than try to make this repo the main landing point for users.

This seems like a nice recommendation to add indeed but I'm not sure it's enough - another issue I found is a user only concerned with tracing may depend on `-sdk-trace`. This is valid for some use cases but not really documented by us and isn't really what we want to recommend - we want almost all users to use the whole SDK.

iNikem:
Well... maybe you are right :) But I am concerned a little, that this change makes it seems that the "main" artifact of this repo is SDK. Is it the case? For me it always was API, but may be I am "instrumenter" :)

anuraaga:
> For me it always was API, but may be I am "instrumenter" :)

Not "may be" :P

I'm going to try to rework this to describe all the use cases in more detail instead of just assuming a default, prbably borrowing some terms from @tedsuo's circle of life.

anuraaga:
I coincidentally looked through opentelemetry instrumentation built into vertx and couchbase and had a thought - actually instrumenters should not be using the opentelemetry-api! (as a direct dependency in their build). Unless we provide a stable instrumentation-api, there is no way we can ensure following OTel semantic conventions - it means that long term instrumenters should depend on instrumentation-api, and this repo is really just for the SDK. Any thoughts on that?

iNikem:
> I coincidentally looked through opentelemetry instrumentation built into vertx and couchbase and had a thought - actually instrumenters should not be using the opentelemetry-api! (as a direct dependency in their build). Unless we provide a stable instrumentation-api, there is no way we can ensure following OTel semantic conventions - it means that long term instrumenters should depend on instrumentation-api, and this repo is really just for the SDK. Any thoughts on that?

This may be an ideal situation, but I think it will be hard to enforce. We have to provide API. And we will not be able to block people from using it.

anuraaga:
> And we will not be able to block people from using it.

It's not about blocking but recommending - if we had a stable instrumentation-api, I can't think of anyone we'd recommend to depend on the opentelemetry-api other than agent users, in which case we'd want to recommend that in agent docs, not here. Does it make sense?

jkwatson:
@anuraaga Are you planning on pushing for this, or is this PR just dead on the vine?

jkwatson:
> @anuraaga Are you planning on pushing for this, or is this PR just dead on the vine?

ping @anuraaga

anuraaga:
Yeah think I'll leave this one

PR #2993, opened 2021-03-05
point examples at the snapshot version of stable modules

### LINKED ISSUE #1686: Examples should be built in CI to ensure they aren't broken when users try them
Since examples are in a separate Gradle project, we don't build them on CI. We should since if they're broken, users will be surprised.

/cc @codeboten

jkwatson:
This will highlight any breaking changes before they are released.

Resolves #1686

### DIFF
--- a/examples/build.gradle
+++ b/examples/build.gradle
@@ -15,7 +15,7 @@ subprojects {
     version = "0.1.0-SNAPSHOT"
 
     ext {
-        openTelemetryVersion = "1.0.0"
+        openTelemetryVersion = "1.1.0-SNAPSHOT"
         openTelemetryAlphaVersion = "1.0.0-alpha"
         grpcVersion = '1.34.1'
         protobufVersion = '3.11.4'


examples/build.gradle
```diff
@@ -15,7 +15,7 @@ subprojects {
     version = "0.1.0-SNAPSHOT"
 
     ext {
-        openTelemetryVersion = "1.0.0"
+        openTelemetryVersion = "1.1.0-SNAPSHOT"
```
bogdandrutu, line 18: I thought about this in the past, downside is that users will copy paste and we probably want to use the released version there.

One option is to make this configurable, so we keep it here as the stable release but when building for testing we force the snapshot version.

jkwatson: I think I know how to make this parameterized, and have the CI jobs run against the snapshot, but the examples themselves use the stable release. Thanks for the suggestion!

jkwatson: ok, check it out. defined these versions as project properties, so can easily be different between CI and copy/pasting for examples.

anuraaga (approved):

PR #2995, opened 2021-03-05
Log clearer messages when the collector responds with UNIMPLEMENTED

### LINKED ISSUE #2571: Improve error message for connecting to collector without metrics endpoint
In open-telemetry/opentelemetry-java-instrumentation#1111 we found that all of our engineers misunderstood the error message for when connecting to a collector without a metrics endpoint. Given we default to tracing and metrics being enabled, I think this will be pretty common for users to run into. We should make the error message better.

jkwatson:
Resolves #2571

### DIFF
--- a/exporters/otlp/metrics/src/main/java/io/opentelemetry/exporter/otlp/metrics/OtlpGrpcMetricExporter.java
+++ b/exporters/otlp/metrics/src/main/java/io/opentelemetry/exporter/otlp/metrics/OtlpGrpcMetricExporter.java
@@ -85,7 +85,7 @@ public final class OtlpGrpcMetricExporter implements MetricExporter {
               switch (status.getCode()) {
                 case UNIMPLEMENTED:
                   logger.log(
-                      Level.WARNING,
+                      Level.SEVERE,
                       "Failed to export metrics. Server responded with UNIMPLEMENTED. "
                           + "This usually means that your collector is not configured with an otlp "
                           + "receiver in the \"pipelines\" section of the configuration. "
@@ -95,7 +95,8 @@ public final class OtlpGrpcMetricExporter implements MetricExporter {
                 case UNAVAILABLE:
                   logger.log(
                       Level.SEVERE,
-                      "Failed to export metrics. Server is UNAVAILABLE.",
+                      "Failed to export metrics. Server is UNAVAILABLE. "
+                          + "Make sure your collector is running and reachable from this network.",
                       t.getMessage());
                   break;
                 default:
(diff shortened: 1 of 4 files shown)


kenfinnigan (approved):

exporters/otlp/metrics/src/main/java/io/opentelemetry/exporter/otlp/metrics/OtlpGrpcMetricExporter.java
```diff
@@ -78,7 +78,19 @@ public void onSuccess(@Nullable ExportMetricsServiceResponse response) {
 
           @Override
           public void onFailure(Throwable t) {
-            logger.log(Level.WARNING, "Failed to export metrics", t);
+            if (t.getMessage().contains("UNIMPLEMENTED")) {
```
anuraaga, line 81: I think you can use `Status.fromThrowable` for more structured access. Is it also a good time to add a message for UNAVAILABLE, the most common exception we get asked about now adays?

anuraaga (commented):
Let's add some unit tests

exporters/otlp/metrics/src/main/java/io/opentelemetry/exporter/otlp/metrics/OtlpGrpcMetricExporter.java
jkwatson: when does UNAVAILABLE happen? Very happy to add it, if I know a good message to provide.

anuraaga: It's generally a connectivity issue. The collector being down, the endpoint being set wrong, network config blocking the requests. A message along the lines of making sure collector is running and reachable by network

jkwatson: ok, check it out now.

jkwatson:
> Let's add some unit tests

done.

jkwatson:
@anuraaga anything else to address here?

exporters/otlp/metrics/src/main/java/io/opentelemetry/exporter/otlp/metrics/OtlpGrpcMetricExporter.java
```diff
@@ -78,7 +80,35 @@ public void onSuccess(@Nullable ExportMetricsServiceResponse response) {
 
           @Override
           public void onFailure(Throwable t) {
-            logger.log(Level.WARNING, "Failed to export metrics", t);
+            if (t instanceof StatusRuntimeException) {
```
anuraaga, line 83: Can't you use fromThrowable? I think it handles both runtime and checked versions.

https://grpc.github.io/grpc-java/javadoc/io/grpc/Status.html#fromThrowable-java.lang.Throwable-

jkwatson: aha! I missed that. I'll update (although functionally it'll be the same, I think)

jkwatson: done

jkwatson:
@anuraaga anything further you want on this PR?

anuraaga (approved):

PR #3000, opened 2021-03-08
Support OTEL_EXPORTER_OTLP_METRICS/TRACES_ENDPOINT

### LINKED ISSUE #2911: Support OTEL_EXPORTER_OTLP_METRICS/TRACES_ENDPOINT
It seems currently only `OTEL_EXPORTER_OTLP_ENDPOINT` is supported, and not `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` nor `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`.

malafeev:
fix #2911

### DIFF
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/SpanExporterConfiguration.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/SpanExporterConfiguration.java
@@ -70,7 +70,10 @@ final class SpanExporterConfiguration {
         "opentelemetry-exporter-otlp");
     OtlpGrpcSpanExporterBuilder builder = OtlpGrpcSpanExporter.builder();
 
-    String endpoint = config.getString("otel.exporter.otlp.endpoint");
+    String endpoint = config.getString("otel.exporter.otlp.traces.endpoint");
+    if (endpoint == null) {
+      endpoint = config.getString("otel.exporter.otlp.endpoint");
+    }
     if (endpoint != null) {
       builder.setEndpoint(endpoint);
     }
--- a/sdk-extensions/autoconfigure/src/testFullConfig/java/io/opentelemetry/sdk/autoconfigure/EndpointConfigurationTest.java
+++ b/sdk-extensions/autoconfigure/src/testFullConfig/java/io/opentelemetry/sdk/autoconfigure/EndpointConfigurationTest.java
@@ -76,6 +76,7 @@ class EndpointConfigurationTest {
   void setUp() {
     otlpTraceRequests.clear();
     otlpMetricsRequests.clear();
+    GlobalOpenTelemetry.resetForTest();
   }
 
   @Test
--- a/sdk-extensions/autoconfigure/src/testFullConfig/java/io/opentelemetry/sdk/autoconfigure/EndpointConfigurationTest.java
+++ b/sdk-extensions/autoconfigure/src/testFullConfig/java/io/opentelemetry/sdk/autoconfigure/EndpointConfigurationTest.java
@@ -38,6 +38,7 @@ import io.opentelemetry.sdk.trace.export.SpanExporter;
 import java.util.Collections;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingDeque;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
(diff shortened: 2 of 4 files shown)


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/SpanExporterConfiguration.java
```diff
@@ -70,7 +70,10 @@ static OtlpGrpcSpanExporter configureOtlpSpans(ConfigProperties config) {
         "opentelemetry-exporter-otlp");
     OtlpGrpcSpanExporterBuilder builder = OtlpGrpcSpanExporter.builder();
 
-    String endpoint = config.getString("otel.exporter.otlp.endpoint");
+    String endpoint = config.getString("otel.exporter.otlp.traces.endpoint");
+    if (endpoint == null) {
+      endpoint = config.getString("otel.exporter.otlp.endpoint");
+    }
```
jkwatson, line 76: can you add some unit tests for these cases? Looks like we don't have any at the moment.

sdk-extensions/autoconfigure/src/testFullConfig/java/io/opentelemetry/sdk/autoconfigure/EndpointConfigurationTest.java
```diff
@@ -0,0 +1,110 @@
...
+    System.setProperty(
+        "otel.exporter.otlp.traces.endpoint", "http://localhost:" + server.httpPort());
+
+    // Point "otel.exporter.otlp.metrics.endpoint" to correct endpoint
+    System.setProperty(
+        "otel.exporter.otlp.metrics.endpoint", "http://localhost:" + server.httpPort());
+
+    System.setProperty("otel.exporter.otlp.timeout", "10000");
+
+    GlobalOpenTelemetry.get().getTracer("test").spanBuilder("test").startSpan().end();
```
jkwatson, line 97: Do we need to use the global instance for these tests?

malafeev: I guess one of the goal of `autoconfigure` module is to set global instance based on system properties/env variables.
So in the tests we set system properties and verify that global instance is configured properly.

jkwatson: We should also confirm that the instance returned by the autoconfigure module is also configured properly, then. That should be the primary thing that we verify, and we can additionally verify the global as a secondary assertion, IMO.

malafeev: added test for non global configuration

sdk-extensions/autoconfigure/src/testFullConfig/java/io/opentelemetry/sdk/autoconfigure/EndpointConfigurationTest.java
```diff
@@ -0,0 +1,184 @@
...
+                  .useBlockingTaskExecutor(true)
+                  .build());
+        }
+      };
+
+  @BeforeEach
+  void setUp() {
+    otlpTraceRequests.clear();
+    otlpMetricsRequests.clear();
+    GlobalOpenTelemetry.resetForTest();
```
jkwatson, line 96: Does it makes sense to do this in the `tearDown` rather than `setUp`? Or maybe both places would be safest?

jkwatson (approved):
Thanks!

sdk-extensions/autoconfigure/src/testFullConfig/java/io/opentelemetry/sdk/autoconfigure/EndpointConfigurationTest.java
malafeev: added

joneschmitz:
Can you please confirm that the environment variable is supposed to be OTEL_EXPORTER_OTLP_METRICS_ENDPOINT? When ran using this value, as listed in the sdk-extensions/autoconfigure/README.md, the 1.22.1 javaagent ignores this environment value and uses the default for localhost. Looking at the test file included in this PR, it seems that the configured environment variable is otel.exporter.otlp.metrics.endpoint instead and when I set it as an environment variable the logs mention its value.

jack-berg:
Yup that environment variable is correct. Documentation is available [here](https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk-extensions/autoconfigure#otlp-exporter-span-metric-and-log-exporters) BTW.

I'd recommend opening a separate issue, probably in the `opentelemetry-java-instrumentation` repo if you believe the behavior is specific to something the agent is doing during initialization. I think the first thing they'll say is to upgrade to a more recent version. But with that said, I don't remember anything in 1.22.1 that would have caused the issue you're describing.

Issue #2999, opened 2021-03-08
Add an example that uses auto-configuration


mateuszrzeszutek:
**Is your feature request related to a problem? Please describe.**
The [examples/](https://github.com/open-telemetry/opentelemetry-java/tree/main/examples) directory is missing the simplest example possible of using the Java SDK with auto-configuration, just using the defaults (with maybe logging exporter) to show how easy is to set it up.

**Describe the solution you'd like**
It could be a single-file Java app that just calls `OpenTelemetrySdkAutoConfiguration.initialize()` and starts a span - it could be as simple as the [.NET 5 minute example](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/trace/getting-started).

jkwatson:
Yes, this would be great!

mateuszrzeszutek:
OK, please assign me then -- I'll prepare something by tomorrow morning.

PR #3003, opened 2021-03-09
Add an example that uses auto-configuration

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java/issues/2999

### DIFF
--- a/examples/autoconfigure/build.gradle
+++ b/examples/autoconfigure/build.gradle
@@ -1,0 +1,15 @@
+plugins {
+    id 'java'
+}
+
+description = "OpenTelemetry Examples for SDK autoconfiguration"
+ext.moduleName = "io.opentelemetry.examples.autoconfigure"
+
+dependencies {
+    implementation("io.opentelemetry:opentelemetry-api")
+    implementation("io.opentelemetry:opentelemetry-exporter-logging")
+    implementation("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure")
+
+    //alpha modules
+    implementation("io.opentelemetry:opentelemetry-api-metrics")
+}
--- a/examples/autoconfigure/build.gradle
+++ b/examples/autoconfigure/build.gradle
@@ -9,7 +9,4 @@ dependencies {
     implementation("io.opentelemetry:opentelemetry-api")
     implementation("io.opentelemetry:opentelemetry-exporter-logging")
     implementation("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure")
-
-    //alpha modules
-    implementation("io.opentelemetry:opentelemetry-api-metrics")
 }
--- a/examples/autoconfigure/src/main/java/io/opentelemetry/example/autoconfigure/AutoConfigExample.java
+++ b/examples/autoconfigure/src/main/java/io/opentelemetry/example/autoconfigure/AutoConfigExample.java
@@ -1,0 +1,59 @@
+package io.opentelemetry.example.autoconfigure;
+
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.api.metrics.GlobalMetricsProvider;
+import io.opentelemetry.api.metrics.LongCounter;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.Tracer;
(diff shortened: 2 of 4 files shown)


examples/autoconfigure/src/main/java/io/opentelemetry/example/autoconfigure/AutoConfigExample.java
```diff
@@ -0,0 +1,59 @@
...
+
+    // Flush out the metrics that have not yet been exported
+    Thread.sleep(1000);
+  }
+
+  private final Tracer tracer;
+  private final LongCounter counter;
+
+  public AutoConfigExample() {
+    this.tracer = GlobalOpenTelemetry.get().getTracer(INSTRUMENTATION_NAME);
```
jkwatson, line 40: I think we'd greatly prefer that our examples don't use the global instance, if at all possible. Usage of the global should only be for extreme circumstances, or for use alongside the agent. For manual instrumentation, all of our examples should use instances of OpenTelemetry and MetricsProvider.

jkwatson: You will have to get it from the global, but that can be done separately. Since autoconfigure doesn't give you a MeterProvider, should it even be in this example?

mateuszrzeszutek: Hmm, good point. Removed metrics from this example -- we can add them back once autoconfigure (or `OpenTelemetry`) returns the created `MetricsProvider`.

examples/autoconfigure/build.gradle
```diff
@@ -0,0 +1,15 @@
...
+description = "OpenTelemetry Examples for SDK autoconfiguration"
+ext.moduleName = "io.opentelemetry.examples.autoconfigure"
+
+dependencies {
+    implementation("io.opentelemetry:opentelemetry-api")
+    implementation("io.opentelemetry:opentelemetry-exporter-logging")
+    implementation("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure")
+
+    //alpha modules
+    implementation("io.opentelemetry:opentelemetry-api-metrics")
```
jkwatson, line 14: This isn't needed now that metrics is out, correct/

anuraaga: Shouldn't be after it points to snapshot :)

mateuszrzeszutek: Good catch, removed.

examples/autoconfigure/src/main/java/io/opentelemetry/example/autoconfigure/AutoConfigExample.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * An example of using {@link io.opentelemetry.sdk.autoconfigure.OpenTelemetrySdkAutoConfiguration}
+ * and logging exporter: {@link io.opentelemetry.exporter.logging.LoggingSpanExporter}.
+ */
+public final class AutoConfigExample {
+  private static final String INSTRUMENTATION_NAME = AutoConfigExample.class.getName();
+
+  public static void main(String[] args) throws InterruptedException {
+    // Turn off metrics for this example
+    System.setProperty("otel.metrics.exporter", "none");
```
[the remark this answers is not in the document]
mateuszrzeszutek: Unfortunately it is - autoconfiguration tries to use OTLP by default:

```
Exception in thread "main" io.opentelemetry.sdk.autoconfigure.ConfigurationException: OTLP Metrics Exporter enabled but opentelemetry-exporter-otlp-metrics not found on classpath. Make sure to add it as a dependency to enable this feature.
	at io.opentelemetry.sdk.autoconfigure.ClasspathUtil.checkClassExists(ClasspathUtil.java:15)
	at io.opentelemetry.sdk.autoconfigure.MetricExporterConfiguration.configureOtlpMetrics(MetricExporterConfiguration.java:53)
```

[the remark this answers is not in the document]
mateuszrzeszutek: Fixed, thanks!

jkwatson (approved):
Thanks!

PR #3008, opened 2021-03-11
Fix AWS detectors to initialize singleton after statics

anuraaga:
Currently some will NPE, others just don't work. The order of static initialization is something that continues to surprise / bite me :(


jkwatson (approved):
Thanks! Is there any way to unit test this?

anuraaga:
Don't think so, we unit test to the best we can outside AWS environment but I think these real singletons are sort of stuck

PR #3012, opened 2021-03-11
Fix unit test for kotlin context propagation.

anuraaga:
`GlobalScope.launch` isn't supposed to inherit context

https://kotlinlang.org/docs/coroutine-context-and-dispatchers.html#children-of-a-coroutine

But the test didn't fail since actually we weren't awaiting on the result anyways. Switched to launching a child coroutine the correct way (I think) and awaiting

### DIFF
--- a/extensions/kotlin/src/test/kotlin/io/opentelemetry/extension/kotlin/KotlinCoroutinesTest.kt
+++ b/extensions/kotlin/src/test/kotlin/io/opentelemetry/extension/kotlin/KotlinCoroutinesTest.kt
@@ -33,10 +33,10 @@ class KotlinCoroutinesTest {
 
             assertThat(Context.current().get(ANIMAL)).isEqualTo("cat")
 
-            GlobalScope.launch {
+            async(Dispatchers.IO) {
                 // Child coroutine inherits context automatically.
                 assertThat(Context.current().get(ANIMAL)).isEqualTo("cat")
-            }
+            }.await()
         }
     }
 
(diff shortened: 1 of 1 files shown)


trask (approved):

asarkar:
Instead of multiple `async-await`, which is unnecessary since you're not interested in the deferred result, you could use `launch-join`, which acts like a void method. Furthermore, if you launch all of them inside a `coroutineScope`, by virtue of structured concurrency, you don't need to call join on each one, just the parent returned by `coroutineScope`. See https://dzone.com/articles/waiting-for-coroutines for examples.

anuraaga:
`launch-join` didn't raise the `AssertionError` on failure unfortunately which is why I went with `async-await`.

asarkar:
> launch-join didn't raise the AssertionError

Are you sure? I'm not seeing any `join` previously.

```
val job = GlobalScope.launch {
     // Child coroutine inherits context automatically.
    assertThat(Context.current().get(ANIMAL)).isEqualTo("cat")
}
```

`runBlocking { job.join() }` must wait.

anuraaga:
@asarkar Right - previously it was completely screwed :P To fix, I first tried `join` but it didn't have the behavior I wanted of rethrowing the exception thrown inside so switched to `await`.

asarkar:
I see, you want the exception. Exception in child `launch` will cancel the parent, so there's another way of testing this (`parent.isCancelled=true && parent.isCompleted=true`), but you've picked your poison :) Thanks.

jkwatson (approved):

asarkar:
@anuraaga I found that `runBlocking` implicitly calls `await`.

```
try {
    deferred = runBlocking { async(exceptionHandler) { throw ArithmeticException() } }
    Assertions.fail("Shouldn't be here")
} catch (ex: Exception) {}
finally {
    Assertions.assertNull(deferred)
}
```

PR #3017, opened 2021-03-11
Add batch span processor benchmarks

sbandadd:
Description:
This PR adds two benchmarks.

1. Current benchmark executs forceFlush() on every loop and creates a bottleneck which results in not stressing batch span processor. Current benchmark only
measures throughput which is not helpful on its own since number of spans getting exported is also important. BatchSpanProcessorMultiThreadBenchmark is created to address this issue.
2. Measuring CPU usage of exporter thread is also important, but the current benchmarks consumes as much CPU as possible which makes the measurement not meaningful.
To maintain a steady state, this PR creates a benchmark that generates 10k spans per second per thread. One would need to attach a profiler such as yourkit or JProfiler
to the benchmark run to understand the processor's CPU usage. BatchSpanProcessorCpuBenchmark is created for this purpose.

This PR also fixes a bug in calculating number of dropped/export spans. Earlier dropped and exported spans were actually counting the other one.

### DIFF
--- a/sdk/trace/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorCpuBenchmark.java
+++ b/sdk/trace/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorCpuBenchmark.java
@@ -1,0 +1,164 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.trace.export;
+
+import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.sdk.metrics.SdkMeterProvider;
+import io.opentelemetry.sdk.trace.ReadableSpan;
+import io.opentelemetry.sdk.trace.SdkTracerProvider;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.LockSupport;
+import org.openjdk.jmh.annotations.AuxCounters;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.annotations.Threads;
+import org.openjdk.jmh.annotations.Warmup;
+
+/*
+ * Run this along with a profiler to measure the CPU usage of BatchSpanProcessor's exporter thread.
+ */
+public class BatchSpanProcessorCpuBenchmark {
+  @State(Scope.Benchmark)
+  public static class BenchmarkState {
+    private SdkMeterProvider sdkMeterProvider;
+    private BatchSpanProcessor processor;
+    private Tracer tracer;
+    private int numThreads = 1;
+
+    @Param({"1"})
+    private int delayMs;
+
+    private long exportedSpans;
+    private long droppedSpans;
+
+    @Setup(Level.Iteration)
+    public final void setup() {
+      sdkMeterProvider = SdkMeterProvider.builder().buildAndRegisterGlobal();
+      SpanExporter exporter = new DelayingSpanExporter(delayMs);
+      processor = BatchSpanProcessor.builder(exporter).build();
+      tracer =
+          SdkTracerProvider.builder().addSpanProcessor(processor).build().get("benchmarkTracer");
+    }
+
+    @TearDown(Level.Iteration)
+    public final void recordMetrics() {
+      BatchSpanProcessorMetrics metrics =
+          new BatchSpanProcessorMetrics(sdkMeterProvider.collectAllMetrics(), numThreads);
+      exportedSpans = metrics.exportedSpans();
+      droppedSpans = metrics.droppedSpans();
+    }
+
+    @TearDown(Level.Iteration)
+    public final void tearDown() {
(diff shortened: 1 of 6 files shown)


sdk/trace/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorCpuBenchmark.java
```diff
@@ -0,0 +1,164 @@
...
+
+    public long droppedSpans() {
+      return benchmarkState.droppedSpans;
+    }
+  }
+
+  private static void doWork(BenchmarkState benchmarkState) {
+    benchmarkState.processor.onEnd(
+        (ReadableSpan) benchmarkState.tracer.spanBuilder("span").startSpan());
+    // This sleep is essential to maintain a steady state of the benchmark run by generating 10k
```
anuraaga, line 94: Thanks for helping with benchmarks - how is the number decided here? 

For overhead benchmarks, I usually avoid jmh because of this but if there's a way to simulate fixed load with jmh that's pretty cool.

sdk/trace/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorMetrics.java
```diff
@@ -0,0 +1,52 @@
...
+  private long getMetric(boolean dropped) {
+    String labelValue = String.valueOf(dropped);
+    Optional<Long> value =
+        allMetrics.stream()
+            .filter(metricData -> metricData.getName().equals("processedSpans"))
+            .filter(metricData -> !metricData.isEmpty())
+            .map(metricData -> metricData.getLongSumData().getPoints())
+            .flatMap(Collection::stream)
+            .filter(point -> labelValue.equals(point.getLabels().get("dropped")))
+            .map(LongPointData::getValue)
```
anuraaga, line 48: mapToLong

sdk/trace/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorMetrics.java
```diff
@@ -0,0 +1,52 @@
...
+    return getMetric(false) / numThreads;
+  }
+
+  public long droppedSpans() {
+    return getMetric(true) / numThreads;
+  }
+
+  private long getMetric(boolean dropped) {
+    String labelValue = String.valueOf(dropped);
+    Optional<Long> value =
```
anuraaga, line 41: I have a feeling two loops will be very similar code while more readable

sdk/trace/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorMultiThreadBenchmark.java
```diff
@@ -0,0 +1,158 @@
...
+    @Param({"0"})
+    private int delayMs;
+
+    private long exportedSpans;
+    private long droppedSpans;
+
+    @Setup(Level.Iteration)
+    public final void setup() {
+      sdkMeterProvider = SdkMeterProvider.builder().buildAndRegisterGlobal();
+      GlobalMetricsProvider.set(sdkMeterProvider);
```
anuraaga, line 49: You call register global do don't need this

sdk/trace/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorCpuBenchmark.java
sbandadd: This is hand picked. 10k sounds reasonable to me for testing

sdk/trace/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorMultiThreadBenchmark.java
sbandadd: Nice. Will remove.

sdk/trace/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorMetrics.java
sbandadd: The existing code is two loop and it took me a good amount of time to debug why dropped/exported spans metric are invalid and got a bit confused on what the two loops are doing. Functional would make it easy to see what are getting filtered and how the data is getting mapped.

That said, I would leave this to the maintainers.

sdk/trace/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorCpuBenchmark.java
anuraaga: We need to know how it was hand picked, especially since these will need to be maintained by other folks, possibly adding / tweaking benchmarks. Is it just a large amount of time to dwarf the time spent on the rest of the work?

sbandadd: Yes, it is a large amount of time. As shown in my runs, this clearly surfaced CPU overhead.

# Chart
image

anuraaga: So let's leave a comment that the time spent in the rest of the work is almost nothing, allowing this sleep to control the rate span generation.

sbandadd: Added

anuraaga (approved):
Having fixed-throughput overhead benchmarks is very helpful, thanks!

jkwatson:
I finally got to spend some dedicated time on this today, and got jmh hooked up with the async-profiler profiling option, and I do see some significant overhead introduced by the usage of the ArrayBlockingQueue (although it's been tough to get really consistent results).

thanks for putting in the work on this. Now, to agree on the best solution to the issue. :)

anuraaga:
@jkwatson Approved of these benchmarks offline so will go ahead and merge. Benchmarks are alive so I'm sure we'll continue to tweak but this is an improvement. Thanks @sbandadd

PR #3019, opened 2021-03-12
Update description on otel.exporter.otlp.headers

topher1120:
The description for configuring otlp headers is that they should be separated using semicolons.  However, tracing the code and testing with a vendor that uses multiple headers as configuration items shows that headers should be separated using commas instead.  
You can see this by viewing the source for either SpanExporterConfiguration or MetricExporterConfiguration and reviewing the line as given:
`config.getCommaSeparatedMap("otel.exporter.otlp.headers").forEach(builder::addHeader);`

Tracing back from this point shows that the header values are separated using commas.


anuraaga (approved):
Thanks!

jkwatson (approved):

PR #3026, opened 2021-03-12
Don't auto-start the IntervalMetricReader in the constructor.

jkwatson:
This changes the IntervalMetricReaderBuilder.build() method to not auto-start the IMR. There is a new method on the builder `buildAndStart()` which will do both for you.  

Adds a method to the IntervalMetricReader to start the background thread, but this is not called as a part of the constructor.

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
@@ -76,9 +76,15 @@ public final class IntervalMetricReader {
   }
 
   IntervalMetricReader(InternalState internalState) {
+    this(
+        internalState,
+        Executors.newScheduledThreadPool(1, new DaemonThreadFactory("IntervalMetricReader")));
+  }
+
+  // visible for testing
+  IntervalMetricReader(InternalState internalState, ScheduledExecutorService intervalMetricReader) {
     this.exporter = new Exporter(internalState);
-    this.scheduler =
-        Executors.newScheduledThreadPool(1, new DaemonThreadFactory("IntervalMetricReader"));
+    this.scheduler = intervalMetricReader;
   }
 
   /**
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
@@ -15,6 +15,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
@@ -31,6 +32,7 @@ public final class IntervalMetricReader {
 
   private final Exporter exporter;
   private final ScheduledExecutorService scheduler;
+  private volatile ScheduledFuture<?> scheduledFuture;
 
   /** Stops the scheduled task and calls export one more time. */
   public CompletableResultCode shutdown() {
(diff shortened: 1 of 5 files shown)


bogdandrutu (approved):

breedx-splk (approved):
Yeah this is solid...side effects of building something are bad!

carlosalberto (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
```diff
@@ -62,11 +64,24 @@ public static IntervalMetricReaderBuilder builder() {
...
-        TimeUnit.MILLISECONDS);
+  }
+
+  /**
+   * Starts this {@link IntervalMetricReader} to report to the configured exporter.
+   *
+   * @return this for fluent usage along with the builder.
+   */
+  public IntervalMetricReader start() {
+    if (scheduledFuture != null) {
```
anuraaga, line 75: Should we go ahead and use `AtomicReference` + CAS for the off chance of parallel start?

jkwatson: I can do that, yes.

jkwatson: actually...not sure how to do this, since you can't do a lazy populator for the atomic value. What I really want is an API that looks like:

`atomicReference.compareAndSet(null, ()-> schedule() )` but we don't have that API.  :(

I could do standard double-checked locking for this, or we could just punt and warn in the javadoc about this case. Thoughts?

jkwatson: Or, since it's not a throughput-sensitive call, simple synchronization on a lock would probably be fine, rather than adding the complication of double-checked locking.

anuraaga (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
```diff
@@ -57,16 +64,30 @@ public static IntervalMetricReaderBuilder builder() {
...
+  }
+
+  /**
+   * Starts this {@link IntervalMetricReader} to report to the configured exporter.
+   *
+   * @return this for fluent usage along with the builder.
+   */
+  public IntervalMetricReader start() {
+    synchronized (lock) {
+      if (scheduledFuture != null) {
```
anuraaga, line 80: Can you add a unit test for double-start?

anuraaga (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
jkwatson: will do, good call.

jkwatson: done

jkwatson: hmm. I wrote a test that passed locally, and failed when we allowed more than one scheduled job. But, it fails on CI. Do you have some advice on how to write a test like this that asserts a negative state in a multi-threaded situation that will work on CI? :)

anuraaga: Hmm - one idea is to record the export time in the test's `MetricExporter` - if any two consecutive exports have a delta less than the export interval. Since it's a single thread I think this invariant holds reliably.

jkwatson: I'll give it a shot

jkwatson: After many tries at various things (and learning that the scheduled executor timing is pretty darn sloppy at the millisecond level), I took the advice of @breedx-splk and passed in a mock executor service and just verified it was only called once.

PR #3029, opened 2021-03-15
Add ConfigurableSamplerProvider to allow custom samplers in agent, etc

anuraaga:


carlosalberto (approved):

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java
```diff
@@ -115,6 +119,14 @@ static SpanLimits configureSpanLimits(ConfigProperties config) {
...
   // Visible for testing
   static Sampler configureSampler(String sampler, ConfigProperties config) {
+    Map<String, Sampler> spiSamplers =
+        StreamSupport.stream(
+                ServiceLoader.load(ConfigurableSamplerProvider.class).spliterator(), false)
+            .collect(
+                Collectors.toMap(
+                    ConfigurableSamplerProvider::getName,
+                    provider -> provider.createSampler(config)));
+
```
trask, line 129: could move this into the `default` branch and only instantiate the requested sampler

trask (approved):

jkwatson (approved):
Thanks! Feel free to merge as-is, or address Trask's suggestion then merge.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java
anuraaga: This is our current pattern will stick with it for this PR but seems like a good improvement for all of em :)

PR #3031, opened 2021-03-16
Remove jcenter and google repositories from build.

anuraaga:
Former is sunset and flaky, latter we don't need since no Android build

### DIFF
--- a/examples/build.gradle
+++ b/examples/build.gradle
@@ -28,7 +28,6 @@ subprojects {
 
     repositories {
         mavenCentral()
-        jcenter()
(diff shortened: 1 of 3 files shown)


trask:
is this is a smaller version of #2387 that may have a better chance?

trask:
oops, i need to check which repo i'm in, sorry 😂

anuraaga:
For reference, build still fails though since Gradle Plugin Portal is just JCenter. /cry

jkwatson (approved):

PR #3034, opened 2021-03-17
Replace ArrayBlockingQueue with jctools queue.

anuraaga:
The minimized trace-shaded-deps jar is ~30K which is pretty small and probably worth it since it's only in the SDK.

Benchmarks at https://gist.github.com/anuraaga/1de9e2526a159b4e932d011c2dfb58e2 for throughput. I didn't check CPU overhead since I think that requires a profiler with our current setup, but there really shouldn't be any real change.

ArrayBlockingQueue caps off at around 3M ops/sec on my Macbook Pro 16, while the JCTools queue gets up to 8.5M. I also tried compound queue (shard queue into # of CPUs chunks) but didn't find a significant improvement so using the simpler one.


jkwatson:
I'll take a look and put it on the profiler tomorrow.

sdk/trace/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorMultiThreadBenchmark.java
```diff
@@ -42,7 +42,7 @@
     private long exportedSpans;
     private long droppedSpans;
 
-    @Setup(Level.Iteration)
+    @Setup(Level.Trial)
```
sbandadd, line 45: I think this will mess up with the exporter metrics..

sdk/trace-shaded-deps/src/main/java/io/opentelemetry/sdk/trace/internal/JcTools.java
```diff
@@ -0,0 +1,31 @@
...
+  public static <T> Queue<T> newMpscArrayQueue(int capacity) {
+    return new MpscArrayQueue<>(capacity);
+  }
+
+  /**
+   * Returns the capacity of the {@link Queue}, which must be a JcTools queue. We cast to the
+   * implementation so callers do not need to use the shaded classes.
+   */
+  public static long capacity(Queue<?> queue) {
+    return ((MessagePassingQueue<?>) queue).capacity();
```
sbandadd, line 27: you can use MpscArrayQueue right? It does have capacity()

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -73,7 +75,7 @@ public static BatchSpanProcessorBuilder builder(SpanExporter spanExporter) {
             scheduleDelayNanos,
             maxExportBatchSize,
             exporterTimeoutNanos,
-            new ArrayBlockingQueue<>(maxQueueSize));
+            JcTools.newMpscArrayQueue(maxQueueSize));
```
sbandadd, line 78: Note that MpscArrayQueue rounds the queue size to power to 2 for various perf reasons. In my opinion it is better to enforce this so users know what the actual memory that is getting allocated.

sdk/trace-shaded-deps/src/main/java/io/opentelemetry/sdk/trace/internal/JcTools.java
anuraaga: We're shading it and using this in a test in a different artifact where we don't want to have to reference the shaded class. I could cast to `MpscArrayQueue` here too but may as well stick with the iterface.

sdk/trace/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorMultiThreadBenchmark.java
anuraaga: Since when collecting metrics it clears the current values I think it might be ok. Either way, either this needs to be `Trial` or `tearDown` needs to be `Iteration` generally we'd only want to start one BSP per trial though I think.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
anuraaga: Do you mean falling back to `ArrayBlockingQueue` if size isn't power of 2? I don't think we can require this for the BSP setting instead since it's too tricky to use.

sbandadd: I meant enforcing the maxQueueSize to be a power of 2.

anuraaga: That we can't do we don't want to lose usability (adding restrictions that can only be conveyed through documentation or error messages) here. Would like to hear more thoughts on whether we should fallback if it's not power-of-2

sdk/trace/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorMultiThreadBenchmark.java
sbandadd: could you please run this benchmark and verify the exportedSpans metric is valid, i.e; current values do get cleared after collecting them at the end of each iteration.

anuraaga: It does keep on going up now - I don't think we should need to restart the BSP for it though. @jkwatson Do you know a nice way to aggregate the metrics into a rate for reporting in the JMH benchmark?

jkwatson: I don't understand the question. Is something not working about the way things are before this change? Also, should we be doing a `forceFlush()` at the iteration `teardown` ?

And, honestly, I'm not sure I understand the purpose of this benchmark in the first place. I ran this on the main branch, and for the 20-thread case, we drop almost all of the spans. Is the goal to see if we can make the BSP drop fewer spans if we can improve things?

jkwatson: BTW, I think we definitely start starving the export thread when we get to a lot of input threads:

image

jkwatson: (note: the jctools version is much, much better than the ABQ version on `main`):
image

anuraaga: For the metric, I think reporting dropped spans per second would be most consistent with how other numbers are also per s. I suspect otel metrics are supposed to provide a way to get such a metric :) The goal of the bench would be to export more spans per second hence wanting that (I guess "not dropping" is similar but given BSP's job is to export spans, I guess exported spans is the number more interesting).

anuraaga: Though if it's annoying to export a rate, I guess total spans isn't so bad either since we know the duration of the benchmarks are fixed (just need to tweak the aux counters definition I think so jmh doesn't report it with that /s).

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
sbandadd: Falling back is not great really since it is not an efficient solution. How about calling it out in the documentation that queue size is rounded to the next power of 2?

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfigurationTest.java
```diff
@@ -97,8 +98,7 @@ void configureSpanProcessor_empty() {
                 assertThat(worker)
                     .extracting("queue")
                     .isInstanceOfSatisfying(
-                        ArrayBlockingQueue.class,
-                        queue -> assertThat(queue.remainingCapacity()).isEqualTo(2048));
+                        Queue.class, queue -> assertThat(JcTools.capacity(queue)).isEqualTo(2048));
```
sbandadd, line 101: The existing logic is verifying the remainingCapacity(). Do you want to do the same?
```
     (JcTools.capacity(queue) - JcTools.size(queue)).isEqualTo(2048)
```

sdk/trace/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorMultiThreadBenchmark.java
sbandadd: > The goal of the bench would be to export more spans per second hence wanting that (I guess "not dropping" is similar but given BSP's job is to export spans

I think we should stick with exports / sec. It is natural and conveys directly what we care about. Dropped / sec is the reverse of it.

jkwatson: It seems that jmh is doing the right thing here to make sure that all of the "/s" values are correct (at least they add up roughly correctly before the change here), so I think sticking with what we have now is good.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -17,8 +17,10 @@
 import io.opentelemetry.sdk.trace.ReadableSpan;
 import io.opentelemetry.sdk.trace.SpanProcessor;
 import io.opentelemetry.sdk.trace.data.SpanData;
+import io.opentelemetry.sdk.trace.internal.JcTools;
```
sbandadd, line 20: Now this comment https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java#L40 is not relevant any more !

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfigurationTest.java
anuraaga: Nah, it only used tha tmethod since the JDK only provides that one, but capacity is what we're checking

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
anuraaga: I added a note to the builder that some more memory may be allocated, without going too much into implementation detail.

anuraaga (commented):
@jkwatson Can you take another look at this? Thanks!

sdk/trace/src/jmh/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorMultiThreadBenchmark.java
jkwatson: I think this change still needs to be reverted @anuraaga

anuraaga: Ok I went ahead and changed the shutdown to be per-iteration too then, need one of the changes to make sure threads are closed or JMH complains (at least to me). I don't think we actually wanted to initialize a whole BSP (worker thread, etc) per iteration though.

jkwatson: What complaints did you get from jmh?

anuraaga: Something about threads not all being shut down, waiting XXX seconds for them to shutdown. Which is definitely true since currently we don't call shutdown on every created BSP.

jkwatson: Ah yes. I had been wondering about that. Good find!

jkwatson (approved):
Let's try it.

sbandadd (approved):
Looks good!

PR #3036, opened 2021-03-17
BatchSpanProcessor that uses provided ScheduledExecutorService

### LINKED ISSUE #2980: Reduce number of threads: BatchSpanProcessor based on ScheduledExecutorService?
It might be beneficial to implement BatchSpanProcessor with a ScheduledExecutorService that can optionally be passed in from the outside so that it could share a thread(pool) with e.g. IntervalMetricsReader and maybe also user's background tasks. The IntervalMetricReader should also allow for passing in the ScheduledExecutorService it already uses from the outside.

Related: #2968 "Reduce CPU usage of batch span processor"

piotr-sumo:
Fixes #2980


Oberon00:
Please fill in the PR description, e.g. "Fixes #2980" if that is the case.

sdk-extensions/executor-service-processor/src/main/java/io/opentelemetry/sdk/extension/trace/export/ExecutorServiceSpanProcessor.java
```diff
@@ -0,0 +1,184 @@
...
+        new Worker(
+            spanExporter,
+            scheduleDelayNanos,
+            maxExportBatchSize,
+            exporterTimeoutNanos,
+            new ArrayBlockingQueue<>(maxQueueSize));
+    this.ownsExecutorService = ownsExecutorService;
+    this.executorService = executorService;
+    this.future =
+        executorService.scheduleAtFixedRate(
```
Oberon00, line 66: I think it would be better to use scheduleWithFixedDelay to avoid getting in a busy loop if export takes too long. That may mean dropped spans, but this is IMHO usually preferable to taking too much CPU time away.

jkwatson:
Since we're getting so many different variant PRs for updates to the BSP (I think we're up to 4!), it might make sense to put them all into an incubator module so people have options and can try them all out to see what works best.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/WorkerBase.java
```diff
@@ -0,0 +1,185 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public abstract class WorkerBase implements Runnable {
```
jkwatson, line 25: I don't think I'm comfortable making this a part of our public SDK APIs at this point.

piotr-sumo: @jkwatson I will try to put this class in package scope. Would that be OK?

piotr-sumo:
@jkwatson I have the question about the incubator module. Is something I should do in this PR?

jkwatson:
> @jkwatson I have the question about the incubator module. Is something I should do in this PR?

If it's a general question about it, a Discussion might be better so it's more visible.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/WorkerBase.java
jkwatson: I'm not sure that will help, since the Java module system doesn't let us have the same package in two jars loaded into the same classloader.

piotr-sumo: Ok, so I will remove this class. There will be some code duplication but at least we will avoid other problems.

jkwatson: If we put this class into an incubator module, and all the BSP options are in that module, in the same package, then that would work.

piotr-sumo: @jkwatson do you mean `sdk-extensions/tracing-incubator` gradle module?

jkwatson: yes, that seems like the right place to incubate sdk bits, like optional span processors.

piotr-sumo: Ok, I've moved the code to the incubator and reverted changes I've made to the original `BatchSpanProcessor`

sdk-extensions/tracing-incubator/build.gradle.kts
```diff
@@ -13,6 +13,7 @@ extra["moduleName"] = "io.opentelemetry.sdk.extension.trace.incubator"
 dependencies {
     api(project(":api:all"))
     api(project(":sdk:all"))
+    api(project(":sdk:metrics"))
```
jkwatson, line 16: I think we want to leave `metrics` as an implementation detail for now, like we have in the core sdk trace module.

jkwatson:
I'm interested to see how the various benchmarks run against this variant. Would you mind also copying over the benchmarks for the BatchSpanProcessor as well? And, if you could run them against this variant and compare to running them against the BSP on `main` right now?

sdk-extensions/tracing-incubator/build.gradle.kts
piotr-sumo: 👍

piotr-sumo:
@jkwatson sure, I will run benchmarks and post results here.

piotr-sumo:
@jkwatson I've tried to run `jmh` but unfortunately I've encountered https://github.com/melix/jmh-gradle-plugin/issues/175
```
Execution failed for task ':sdk:all:jmh'.
> A failure occurred while executing me.champeau.gradle.IsolatedRunner
   > Error while instantiating tests: unable to set 'list' on Runner. This plugin version doesn't seem to be compatible with JMH 1.25. Please report to the plugin authors at https://github.com/melix/jmh-gradle-plugin/.
```
I have JDK 15 installed on my machine.
How do you run microbenchmarks from your machine?

piotr-sumo:
@jkwatson It turns out that one needs JDK 11 to run benchmarks. I've run them and it turns out that the performance of this variant is a little bit worse than the original `BatchSpanProcessor`

piotr-sumo:
Screenshot 2021-03-22 at 15 11 47
Screenshot 2021-03-22 at 15 12 07

piotr-sumo:
Screenshot 2021-03-22 at 15 12 59
Screenshot 2021-03-22 at 15 13 16

piotr-sumo:
Screenshot 2021-03-22 at 15 13 49
Screenshot 2021-03-22 at 15 14 15

piotr-sumo:
Screenshot 2021-03-22 at 15 15 16
Screenshot 2021-03-22 at 15 15 30

piotr-sumo:
Screenshot 2021-03-22 at 15 16 02
Screenshot 2021-03-22 at 15 16 15

piotr-sumo:
Screenshot 2021-03-22 at 15 16 48
Screenshot 2021-03-22 at 15 17 04

piotr-sumo:
@jkwatson could you re-run tests? I don't have permissions to do this myself and the failure is not related to my changes:
```
Run gradle/wrapper-validation-action@v1
Error: connect ETIMEDOUT 104.18.190.9:443
```

piotr-sumo:
@jkwatson I've merged `main` to trigger test re-run. Could you now take a look at benchmark results and my PR?

jkwatson:
> @jkwatson I've merged `main` to trigger test re-run. Could you now take a look at benchmark results and my PR?

I'm on vacation this week. I'll look next week if @anuraaga  can't get to it.

piotr-sumo:
@anuraaga could take a look at this PR?

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/ExecutorServiceSpanProcessorBuilder.java
```diff
@@ -0,0 +1,185 @@
...
+    return this;
+  }
+
+  /**
+   * Sets the delay interval between two consecutive runs of the worker job. If unset, defaults to
+   * {@value WORKER_SCHEDULE_INTERVAL}ms.
+   */
+  public ExecutorServiceSpanProcessorBuilder setWorkerScheduleInterval(Duration interval) {
+    requireNonNull(interval, "interval");
+    return setWorkerScheduleInterval(interval.toMillis(), TimeUnit.MILLISECONDS);
```
anuraaga, line 142: May as well use nanos

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/WorkerBase.java
```diff
@@ -0,0 +1,186 @@
...
+
+  protected void exportCurrentBatch() {
+    Collection<SpanData> batch = getBatch();
+    if (batch.isEmpty()) {
+      return;
+    }
+
+    try {
+      final CompletableResultCode result = spanExporter.export(new ArrayList<>(batch));
+      result.join(exporterTimeoutNanos, TimeUnit.NANOSECONDS);
```
anuraaga, line 113: Sorry that it took reading code to come up with this basic question - is there a use case for sharing an `ExecutorService` if we're doing blocking I/O? For example, if we try to share a single thread between span processor and interval metric reader, than while we're waiting for spans to be exported, this thread is asleep doing nothing while metrics can't be processed. Or if we use two thread executor, then we may as well have used separate single-thread executors instead of opening up a can of worms for blocking I/O to affect other processes. @Oberon00 what do you think?

We had an asynchronous BSP at one point and for such a processor, sharing an executor makes a lot of sense.

Oberon00: > while we're waiting for spans to be exported, this thread is asleep doing nothing while metrics can't be processed

The current implementation of the IntervalMetricReader does metric collection and export on the same thread, so this can really be a problem. But if it did metric export and collection on separate threads, it could share only the metric exporting thread, not the collection thread.

> We had an asynchronous BSP at one point

I think the `join` will need to be removed and the BSP needs to use the `whenComplete` mechanism for this implementation to become more robust (to maintain the timeout, maybe additionally use `schedule` to signal the whenComplete callback to cancel the whenComplete).

This will also mean that we can no longer use `scheduleWithFixedDelay` but need to `schedule()` the next export ourselves whenever the last `whenComplete` is done.

@anuraaga Does that sound reasonable?

anuraaga: > it could share only the metric exporting thread, not the collection thread.

I guess even with this, if we're not doing the I/O simultaneously it could cause issues with the export throughput. Pretty sure that's what you're thinking too just confirming :)

> This will also mean that we can no longer use scheduleWithFixedDelay but need to schedule() the next export ourselves whenever the last whenComplete is done.

Yeah this seems ok to me - unless I'm missing something I think it's a prerequisite for having exporters share threads.

Oberon00: > it could cause issues with the export throughput

I'm not sure how serious this is. Even if metrics and spans have to wait for each other, it depends on the intervals if this is a problem. In the worst case where the thread is always "busy" (at least waiting for I/O to complete) and metrics and spans cause about the same traffic, even when using two threads (one for metrics, one for spans) you get at most 2x improvement.

piotr-sumo: @anuraaga @Oberon00 thank you for the detailed explanation. I will rework my SpanProcessor.

jkwatson:
FYI on the benchmarks...the CPU-based benchmarks should be unchanged, and this is intentional, as they really aren't supposed to be useful without a profiler hooked up.

sdk-extensions/tracing-incubator/src/jmh/java/io/opentelemetry/sdk/extension/incubator/trace/BatchSpanProcessorMetrics.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+package io.opentelemetry.sdk.extension.incubator.trace;
+
+import io.opentelemetry.sdk.metrics.data.LongPointData;
+import io.opentelemetry.sdk.metrics.data.MetricData;
+import java.util.Collection;
+import java.util.OptionalLong;
+
+public class BatchSpanProcessorMetrics {
```
jkwatson, line 13: does this class need to be public? I don't think we'd want to expose it as a part of the API, even in the incubator.

jkwatson: never mind...it's in the jmh package. 🤦

jkwatson: It would be really cool if we had a separate module for the jmh benchmarks, so we could share all of this code and not make copies of it. Can you create an issue to track that work?

piotr-sumo: I've created an issue https://github.com/open-telemetry/opentelemetry-java/issues/3083

anuraaga (approved):
This seems reasonable to me, thanks!

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/ExecutorServiceSpanProcessor.java
```diff
@@ -0,0 +1,210 @@
...
+
+    private void updateNextExportTime() {
+      nextExportTime.set(System.nanoTime() + scheduleDelayNanos);
+    }
+
+    @Override
+    public void run() {
+      // nextExportTime is set for the first time in the constructor
+
+      continueWork.set(true);
```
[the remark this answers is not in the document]
piotr-sumo: @Oberon00 I don't know what would happen if the next run of the worker `Runnable` will be scheduled on a different thread than the previous one. If it is ok to have a local boolean variable, I will update the code.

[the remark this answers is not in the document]
Oberon00: But it looks like the boolean is never used outside this method, except in shutdown, and there it seems to change nothing as you already check isShutdown too. But I may have misread the logic.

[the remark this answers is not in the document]
piotr-sumo: `isShutdown` - should the SpanProcessor shutdown. If true, then no more work should be done.
`continueWork` - should the current run (loop) continue or not. If false, then at some point in the future the loop will be running again.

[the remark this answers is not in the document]
Oberon00: The question is: If `continueWork.set(false);` is done in WorkerBase.shutdown, is `isShutdown` also already true? If so, then that access to continueWork could be removed. And then we have the situation that at the beginning of `run` continueWork is always set before it is read and continueWork is never used outside the method. In that case it can be transformed without change of behavior to a local boolean.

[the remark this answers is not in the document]
piotr-sumo: @Oberon00 you are correct. I've converted `continueWork` to local variable.

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/WorkerBase.java
```diff
@@ -0,0 +1,205 @@
...
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+abstract class WorkerBase implements Runnable {
```
jkwatson, line 28: It's probably ok since this is just in the incubator, but this class is confusing to me. It's unclear how one would use it in general to implement a Batching SpanProcessor. The responsibilities between the abstract class and a concrete implementation are not particularly well defined, as far as I can tell. 

I'd also prefer it if we can figure out a solution that doesn't involve inheritance. Can the functionality in this class be something that is either a) usable directly, without extending it or b) this class is concrete and takes an implementation of some strategy interface in order to do its work?

At the absolute minimum, this class would need detailed documentation on how to use it properly, as I suspect that there are many gotchas that have to be done just right for it to work correctly.

piotr-sumo: This class can reduce the amount of copied code from `BatchSpanProcessor` worker.
Let me think about other solutions you've proposed above.

piotr-sumo: @jkwatson I've removed `WorkerBase` class. I've introduced `WorkerExporter` which contains all methods useful for exporting a batch of spans.

piotr-sumo: I think that `WorkerExported` might be useful compared to `WorkerBase`.

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/ExecutorServiceSpanProcessor.java
```diff
@@ -0,0 +1,280 @@
...
+        long workerScheduleIntervalNanos) {
+      Meter meter = GlobalMeterProvider.getMeter("io.opentelemetry.sdk.trace");
+      meter
+          .longValueObserverBuilder("queueSize")
+          .setDescription("The number of spans queued")
+          .setUnit("1")
+          .setUpdater(
+              result ->
+                  result.observe(
+                      queue.size(), Labels.of(spanProcessorTypeLabel, spanProcessorTypeValue)))
```
jkwatson, line 159: might as well make these labels statically allocated, so we don't have to recreate them every collection cycle.

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/ExecutorServiceSpanProcessor.java
```diff
@@ -0,0 +1,280 @@
...
+    private final SpanExporter spanExporter;
+    private final ScheduledExecutorService executorService;
+
+    private Worker(
+        SpanExporter spanExporter,
+        long scheduleDelayNanos,
+        int maxExportBatchSize,
+        long exporterTimeoutNanos,
+        BlockingQueue<ReadableSpan> queue,
+        String spanProcessorTypeLabel,
```
jkwatson, line 146: Is there a good reason to have these part of the constructor? Aren't they always the same? I'd rather just have them be constants, probably in the enclosing class.

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/ExecutorServiceSpanProcessor.java
```diff
@@ -0,0 +1,280 @@
...
+        if (flushRequested.get() != null) {
+          workerExporter.flush(batch, queue);
+        }
+
+        try {
+          ReadableSpan lastElement = queue.peek();
+          if (lastElement != null) {
+            batch.add(lastElement.toSpanData());
+            // drain queue
+            queue.take();
```
jkwatson, line 214: Rather than `peek()` then later a `take()`, why not just `poll()` initially  and skip the extra step?

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/ExecutorServiceSpanProcessor.java
```diff
@@ -0,0 +1,280 @@
...
+      while (continueWork && !isShutdown.get()) {
+        if (flushRequested.get() != null) {
+          workerExporter.flush(batch, queue);
+        }
+
+        try {
+          ReadableSpan lastElement = queue.peek();
+          if (lastElement != null) {
+            batch.add(lastElement.toSpanData());
+            // drain queue
```
jkwatson, line 213: this comment doesn't seem accurate. You're not draining the queue, you're just removing the element you just `peek()`ed at, correct? 
Also, in the CPU-optimization work that was done for the main BSP, we definitely found a significant overhead in working with an ArrayBlockingQueue this way. Any thoughts to using the same signalling technique and the jctools queue here?

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/WorkerExporter.java
```diff
@@ -0,0 +1,117 @@
...
+
+  private final SpanExporter spanExporter;
+  private final ScheduledExecutorService executorService;
+  private final Logger logger;
+  private final long exporterTimeoutNanos;
+  private final BoundLongCounter exportedSpans;
+  private final AtomicReference<CompletableResultCode> flushRequested;
+  private final int maxExportBatchSize;
+
+  WorkerExporter(
```
jkwatson, line 34: Seems weird to have a public class with all public methods with a non-public constructor. Does this class need to be public right now?

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/WorkerExporter.java
```diff
@@ -0,0 +1,117 @@
...
+      SpanExporter spanExporter,
+      ScheduledExecutorService executorService,
+      Logger logger,
+      long exporterTimeoutNanos,
+      BoundLongCounter exportedSpans,
```
jkwatson, line 39: rename to "exportedSpanCounter"

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/WorkerExporter.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.logging.Logger;
+
+public class WorkerExporter {
+
+  private final SpanExporter spanExporter;
+  private final ScheduledExecutorService executorService;
+  private final Logger logger;
+  private final long exporterTimeoutNanos;
+  private final BoundLongCounter exportedSpans;
+  private final AtomicReference<CompletableResultCode> flushRequested;
```
jkwatson, line 31: The contract on what this is for and how it's used needs to be very carefully documented if we're going to have this class be public. I might start by renaming it to something like "flushSignal" or something that makes it clear that it's a 2-way communication pathway between the span processor and this worker.

piotr-sumo:
@jkwatson thank you for code review. I will address your comments as soon as I have time at work.

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/WorkerExporter.java
piotr-sumo: nope, it can be in package scope

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/ExecutorServiceSpanProcessor.java
piotr-sumo: I'll try to plug in jctools queue.

piotr-sumo: all tests pass 👍

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/ExecutorServiceSpanProcessor.java
piotr-sumo: I didn't want the thread to be blocked when the queue is empty. With `jctools` I can `poll()` and skip the extra step.

piotr-sumo:
@jkwatson I've addressed all comments. Could you re-review?

jkwatson (approved):
Let's give it a try in the incubator!

PR #3051, opened 2021-03-20
Introduce a ThrottlingLogger wrapper and use for error logging in exporters

### LINKED ISSUE #3013: Otel Collector excessive logging
**Describe the bug**
```[opentelemetry.auto.trace 2021-03-11 06:31:57:789 +0000] [BatchSpanProcessor_WorkerThread-1] WARN io.opentelemetry.exporter.jaeger.thrift.JaegerThriftSpanExporter - Failed to export spans
io.jaegertracing.internal.exceptions.SenderException: Could not send 2 spans, response 400: invalid character '\f' looking for beginning of value

	at io.jaegertracing.thrift.internal.senders.HttpSender.send(HttpSender.java:86)
	at io.opentelemetry.exporter.jaeger.thrift.JaegerThriftSpanExporter.lambda$export$2(JaegerThriftSpanExporter.java:102)
	at java.util.HashMap.forEach(HashMap.java:1289)
	at io.opentelemetry.exporter.jaeger.thrift.JaegerThriftSpanExporter.export(JaegerThriftSpanExporter.java:96)
	at io.opentelemetry.sdk.trace.export.BatchSpanProcessor$Worker.exportCurrentBatch(BatchSpanProcessor.java:262)
	at io.opentelemetry.sdk.trace.export.BatchSpanProcessor$Worker.run(BatchSpanProcessor.java:203)
	at java.lang.Thread.run(Thread.java:748)```

If the collector crash loops, it fills up the logs with this.  Making the logs pretty unreadable.

### PR DESCRIPTION
**Pull request by jkwatson**
Resolve #3013

### CODE REVIEW

File: exporters/otlp/metrics/src/main/java/io/opentelemetry/exporter/otlp/metrics/OtlpGrpcMetricExporter.java [hunk 1]
```diff
@@ -94,8 +96,8 @@ public void onFailure(Throwable t) {
                 logger.log(
                     Level.SEVERE,
                     "Failed to export metrics. Server is UNAVAILABLE. "
-                        + "Make sure your collector is running and reachable from this network.",
-                    t.getMessage());
+                        + "Make sure your collector is running and reachable from this network."
```
jkwatson, line 99: note: this was actually a bug

sdk/common/src/main/java/io/opentelemetry/sdk/internal/ThrottlingLogger.java
```diff
@@ -0,0 +1,86 @@
...
+  private static final int RATE_LIMIT = 5;
+  private static final int THROTTLED_RATE_LIMIT = 1;
+  private static final TimeUnit rateTimeUnit = MINUTES;
+
+  private final Logger delegate;
+  private final Clock clock;
+
+  // this window tracks the number of logs messages per time unit. It holds the times of `log`
+  // calls.
+  private final Deque<Long> window = new ConcurrentLinkedDeque<>();
```
anuraaga, line 32: This looks something like a leaky bucket sampler. Can we just move the rate limiter from jaeger remote sampler package and use it? It doesn't need a queue.

jkwatson: The rate limiter from the jaeger sampler has a bit of a different goal. This isn't really a rate limiter that I've implemented. Instead, it's a throttle that gets invoked if the rate exceeds a value. The throttle is never released, though, unlike the Jaeger rate limiter which will let a steady number through for all time.

My goal here was to have an exporter log noisily initially (up to the 5 per minute limit), but if it's a known noisy logger, then have it just be clamped down to 1 per minute after that. That's a bit different than a rate limiter.

anuraaga: I think the jaeger rate limiter, since its leaky bucket, is similar - if maxBalance is relatively hard and creditsPerSecond is set to one per minute, it would allow a lot of messages at first and then switch to 1 per minute

https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/RateLimiter.java#L21

At least I think so :)

The main difference is that the jaeger limiter could recover if the exporter stops being noisy. But that's ok I guess?

jkwatson: Yeah, I wasn't sure if it was ok or not. If you've got a known-noisy logger, do you really want to let it back in to noisy mode?

 I personally think the queue-based implementation is fairly elegant, as you can simply adjust the length of the queue in order to raise or lower the current limit. But, I'll take another look at that RateLimiter and see if I can generate similar behavior to this one.

anuraaga: But I don't like math :P So for the current throttled_rate_limit of 1 per minute, `creditsPerSecond = 1 / 60`. Then just need to decide how many log messages to allow before throttling, `maxBalance = 5` would log 5 consecutive messages without throttling and messages after that would be throttled to once per minute - unless logs stopped happening in which case after 5 minutes the rate limiter would be back to the initial state.

This might not be exactly what you have here but it seems like a reasonable strategy and I think it's close. The queue seems like it could be some decent overhead for something we want to actually be a no-op

jkwatson: I'm definitely not worried about overhead in the logging wrapper. The whole point of this thing is to reduce the amount of logging which is happening on the exporter (background) threads. The overhead is really irrelevant here, IMO.

jkwatson: ok, I re-implemented, using the Jaeger RateLimiter, and moved that class into the sdk-common internal package.

sdk/common/src/main/java/io/opentelemetry/sdk/internal/ThrottlingLogger.java
```diff
@@ -0,0 +1,89 @@
...
+  private void doLog(Level level, String message, @Nullable Throwable throwable) {
+    if (throwable != null) {
+      delegate.log(level, message, throwable);
+    } else {
+      delegate.log(level, message);
+    }
+  }
+
+  /**
+   * Check if the current wrapped logger is set to log at the given level.
```
anuraaga, line 82:
```suggestion
   * Returns whether the current wrapped logger is set to log at the given level.
```

anuraaga (approved):
Thanks!

PR #3057, opened 2021-03-24
Remove redundant build script code.

Oberon00:
We already set the release compiler option, so no need to set the target/source compatibility.
According to <https://docs.gradle.org/current/dsl/org.gradle.api.tasks.compile.CompileOptions.html#org.gradle.api.tasks.compile.CompileOptions:release>,

> If set, it will take precedences over the `AbstractCompile.getSourceCompatibility()` and
`AbstractCompile.getTargetCompatibility()` settings.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -8,7 +8,6 @@ import net.ltgt.gradle.errorprone.ErrorProneOptions
 import net.ltgt.gradle.errorprone.ErrorPronePlugin
 import org.gradle.api.plugins.JavaPlugin.*
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat
-import ru.vyarus.gradle.plugin.animalsniffer.AnimalSniffer
 import ru.vyarus.gradle.plugin.animalsniffer.AnimalSnifferExtension
 import ru.vyarus.gradle.plugin.animalsniffer.AnimalSnifferPlugin
 import java.time.Duration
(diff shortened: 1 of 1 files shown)


anuraaga (commented):
I think we tried this once and had to add back because Intellij doesn't respect when only the release flag is set. I remember you commenting on that fix PR too :)

Oberon00:
I'm using IntelliJ myself and had no issues though. I even have a slightly more complex build script where some compile tasks (main) are set to use Java 7 and others (test) to use Java 8, using only the release option, and IntelliJ correctly detects that. Or in other words: I cannot reproduce that issue. Maybe it only affects older IntelliJ versions?

anuraaga:
@Oberon00 I needed to use a completely clean workspace to reproduce it before using latest Intellij as of 10/2020, did you try a clean workspace?

Comment is this:
https://github.com/open-telemetry/opentelemetry-java/pull/1829#discussion_r512896424

If it's fixed in newer IntelliJ that's nice though :D

Oberon00:
I will try with a fresh checkout.

Oberon00:
I tried just now and I observed the following: The projects setting show the Java version as 11 (my JDK version):

image

However, any modules that actually contain code have the language level set to Java 8:

image

So probably the modules without any compile task get the target/source compatibility assigned.

anuraaga:
As long as the Java code all have the correct language level, this seems good. Thanks for confirming!

anuraaga (approved):

Issue #2975, opened 2021-03-03
Helper for merging a Resource with one or two key/value pairs


kenfinnigan:
**Is your feature request related to a problem? Please describe.**
In defining a `Resource` on `SdkTracerProviderBuilder` I have code that does:

```
builder.setResource(
        Resource.getDefault()
                .merge(
                        Resource.create(
                                Attributes.of(
                                        ResourceAttributes.SERVICE_NAME, serviceName,
                                        ResourceAttributes.SERVICE_VERSION, serviceVersion))));
```

It's fairly verbose for adding two key/value pairs to the default resource.

**Describe the solution you'd like**
A convenience method would be nice to remove the intermediate steps:

```
builder.setResource(
        Resource.getDefault()
                .merge(
                        ResourceAttributes.SERVICE_NAME, serviceName,
                        ResourceAttributes.SERVICE_VERSION, serviceVersion);
```

While it's a nice convenience method, I don't think it should go overboard in the number of key/value pairs it can accept.

**Additional context**
As `Resource.merge()` extracts the attributes from the resource to merge them,
creating the intermediary objects is unnecessary.

jkwatson:
I'm not sure what the right API for this is, but I do agree that it's pretty verbose at the moment.

jkwatson:
It would be good to provide an sdk-extension for this convenience function (and others, obviously) that we could consider for eventual inclusion in the SDK proper.

kenfinnigan:
Would that just be a static utility that returns the created `Resource`?

Or is there another example that does something similar?

jkwatson:
> Would that just be a static utility that returns the created `Resource`?
> 
> Or is there another example that does something similar?

yeah... similar to what I started for the API here: #2962

anuraaga:
I've found this pretty annoying already too and think it deserves a relatively low or no time in incubation if possible. 

I've always been wishing for `Resource.builder` since it always feels like I'm working on an `Attributes`.

```java
Resource.getDefault().toBuilder()
  .put(Keys.CAT, "meow")
  .put(Keys.DOG, "bark")
  .putAll(OsResource.get())
  .putAll(MY_ATTRIBUTES)
  .build()
```

piotr-sumo:
@anuraaga I've implemented what you've proposed above in https://github.com/open-telemetry/opentelemetry-java/pull/3065

PR #3065, opened 2021-03-26
Resource.Builder for convenient work with Resource class

piotr-sumo:
Fixes https://github.com/open-telemetry/opentelemetry-java/issues/2975

### DIFF
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
@@ -169,88 +169,11 @@ public abstract class Resource {
     return !name.getKey().isEmpty() && isValid(name.getKey());
   }
 
-  public static Resource.Builder builder() {
-    return new Resource.Builder();
+  public static ResourceBuilder builder() {
+    return new ResourceBuilder();
   }
 
-  public Builder toBuilder() {
+  public ResourceBuilder toBuilder() {
     return builder().putAll(this);
-  }
-
-  public static class Builder {
-
-    private final AttributesBuilder attributesBuilder = Attributes.builder();
-
-    public Builder put(String key, String value) {
-      attributesBuilder.put(key, value);
-      return this;
-    }
-
-    public Builder put(String key, long value) {
-      attributesBuilder.put(key, value);
-      return this;
-    }
-
-    public Builder put(String key, double value) {
-      attributesBuilder.put(key, value);
-      return this;
-    }
-
-    public Builder put(String key, boolean value) {
-      attributesBuilder.put(key, value);
-      return this;
-    }
-
-    public Builder put(String key, String... values) {
-      attributesBuilder.put(key, values);
-      return this;
-    }
-
-    public Builder put(String key, long... values) {
-      attributesBuilder.put(key, values);
-      return this;
-    }
-
-    public Builder put(String key, double... values) {
-      attributesBuilder.put(key, values);
-      return this;
-    }
-
-    public Builder put(String key, boolean... values) {
-      attributesBuilder.put(key, values);
-      return this;
-    }
-
-    public <T> Builder put(AttributeKey<T> key, T value) {
-      attributesBuilder.put(key, value);
-      return this;
-    }
-
-    public <T> Builder put(AttributeKey<Long> key, int value) {
-      attributesBuilder.put(key, value);
-      return this;
-    }
-
-    public Builder putAll(Attributes attributes) {
-      attributesBuilder.putAll(attributes);
-      return this;
-    }
-
-    /**
-     * Adds all attributes from given {@link Resource}, if it is not null.
-     *
-     * @param resource {@link Resource} whoose attributes will be copied
-     * @return this Builder
-     */
-    public Builder putAll(@Nullable Resource resource) {
-      if (resource != null) {
-        attributesBuilder.putAll(resource.getAttributes());
-      }
-      return this;
-    }
-
-    public Resource build() {
-      return new AutoValue_Resource(attributesBuilder.build());
-    }
   }
 }
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
@@ -169,10 +169,15 @@ public abstract class Resource {
     return !name.getKey().isEmpty() && isValid(name.getKey());
   }
 
+  /** Returns a new {@link ResourceBuilder} instance for creating arbitrary {@link Resource}. */
   public static ResourceBuilder builder() {
     return new ResourceBuilder();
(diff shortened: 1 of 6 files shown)


sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -168,4 +168,89 @@ private static boolean isValid(String name) {
...
+
+  public static Resource.Builder builder() {
+    return new Resource.Builder();
+  }
+
+  public Builder toBuilder() {
+    return builder().putAll(this);
+  }
+
+  public static class Builder {
```
anuraaga, line 180: We don't use inner classes, can you move to `ResourceBuilder`?

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -168,4 +168,89 @@ private static boolean isValid(String name) {
   private static boolean isValidAndNotEmpty(AttributeKey<?> name) {
     return !name.getKey().isEmpty() && isValid(name.getKey());
   }
+
+  public static Resource.Builder builder() {
```
anuraaga, line 172: Let's add javadoc, you can probably mostly copy/paste from Attributes

sdk/common/src/test/java/io/opentelemetry/sdk/resources/ResourceTest.java
```diff
@@ -193,4 +193,33 @@ void testDefaultResources() {
...
+    // given
+    Resource.Builder builder = Resource.getDefault().toBuilder();
+
+    // when
+    builder.put("dog says what?", "woof");
+
+    // then
+    Resource resource = builder.build();
+    // yes, it's supposed to be the reference check
+    assertThat(resource != Resource.getDefault()).isTrue();
```
anuraaga, line 222: We probably don't really need this assertion, but if including, `assertThat(resource).isNotSameAs(...)`

sdk/common/src/test/java/io/opentelemetry/sdk/resources/ResourceTest.java
```diff
@@ -193,4 +193,33 @@ void testDefaultResources() {
...
+    builder.putAll((Resource) null);
+
+    // then no exception is thrown
+    // and
+    assertThat(builder.build().getAttributes().get(ResourceAttributes.SERVICE_NAME))
+        .isEqualTo("unknown_service:java");
+  }
+
+  @Test
+  void shouldBuilderCopyResource() {
```
anuraaga, line 212: We'll need to add test coverage of all the new methods

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -168,4 +168,89 @@ private static boolean isValid(String name) {
...
+      return this;
+    }
+
+    /**
+     * Adds all attributes from given {@link Resource}, if it is not null.
+     *
+     * @param resource {@link Resource} whoose attributes will be copied
+     * @return this Builder
+     */
+    public Builder putAll(@Nullable Resource resource) {
```
anuraaga, line 245: We should have defensive null checks for all the methods but we don't need to accept `@Nullable` here I think

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -168,4 +168,89 @@ private static boolean isValid(String name) {
   private static boolean isValidAndNotEmpty(AttributeKey<?> name) {
     return !name.getKey().isEmpty() && isValid(name.getKey());
   }
+
+  public static Resource.Builder builder() {
+    return new Resource.Builder();
+  }
+
+  public Builder toBuilder() {
```
anuraaga, line 176: If we add this to incubator first, we would add `ResourceBuilder.from(Resource` and not add these methods yet. This class seems genuinely useful though and maybe we can go with it without incubator. @jkwatson what do you think?

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
piotr-sumo: Sure, will do.

piotr-sumo:
@anuraaga I've implemented all requested changes. I'm waiting for the info about the incubator - shall I move the code there or not.

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -168,4 +168,12 @@ private static boolean isValid(String name) {
   private static boolean isValidAndNotEmpty(AttributeKey<?> name) {
     return !name.getKey().isEmpty() && isValid(name.getKey());
   }
+
+  public static ResourceBuilder builder() {
```
jkwatson, line 172: these methods need javadoc if we're going to make them public.

sdk/common/src/main/java/io/opentelemetry/sdk/resources/ResourceBuilder.java
```diff
@@ -0,0 +1,173 @@
...
+ */
+
+package io.opentelemetry.sdk.resources;
+
+import io.opentelemetry.api.common.AttributeKey;
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.common.AttributesBuilder;
+
+@SuppressWarnings("AutoValueSubclassLeaked")
+public class ResourceBuilder {
```
jkwatson, line 13: This class needs javadoc.

jkwatson:
My personal opinion is that this isn't a big enough change to warrant going through the incubation process. @bogdandrutu since this is an API addition, your opinion would be valuable.

piotr-sumo:
@jkwatson I've addressed your comments. 
@bogdandrutu could you review this PR?

anuraaga (approved):

anuraaga:
Thanks @piotr-sumo

jkwatson (approved):

sdk/common/src/main/java/io/opentelemetry/sdk/resources/ResourceBuilder.java
```diff
@@ -0,0 +1,177 @@
...
+  public ResourceBuilder putAll(Resource resource) {
+    if (resource != null) {
+      attributesBuilder.putAll(resource.getAttributes());
+    }
+    return this;
+  }
+
+  /** Create the {@link Resource} from this. */
+  public Resource build() {
+    return new AutoValue_Resource(attributesBuilder.build());
```
jkwatson, line 175: should this just call `Resource.create()` rather than having an extra reference to the autovalue implementation ?

anuraaga:
Thanks @piotr-sumo !

PR #3073, opened 2021-03-29
Create some integration tests for the various B3 propagator options.

jkwatson:


integration-tests/src/test/java/io/opentelemetry/B3PropagationIntegrationTest.java
```diff
@@ -0,0 +1,245 @@
...
+      try (Response r = call.execute()) {
+        ResponseBody body = r.body();
+        assertThat(body.string()).isEqualTo("OK");
+      }
+    } finally {
+      span.end();
+    }
+
+    List<SpanData> finishedSpanItems = spanExporter.getFinishedSpanItems();
+    assertThat(finishedSpanItems).hasSize(3);
```
jamesylgan, line 93: Using a static int variable (with a descriptive name) might make more sense here - otherwise it is unclear why # span items should be `3`

jamesylgan (approved):
These ITs look great!

Small suggestion - it would be great if these ITs could leverage the classes:
`SdkTracerProviderConfigurer` and `ConfigurablePropagatorProvider` (so we can detect if the system of loading in custom code has changed)

integration-tests/src/test/java/io/opentelemetry/B3PropagationIntegrationTest.java
jkwatson: I can add a comment. I don't think having a static variable would help with clarity here.

jkwatson: comments added. good suggestion

integration-tests/src/test/java/io/opentelemetry/B3PropagationIntegrationTest.java
```diff
@@ -0,0 +1,247 @@
...
+import spark.Service;
+
+/** Integration tests for the B3 propagators, in various configurations. */
+public class B3PropagationIntegrationTest {
+
+  private static final int server1Port = new Random().nextInt(40000) + 1024;
+  private static final int server2Port = new Random().nextInt(40000) + 1024;
+  private static final InMemorySpanExporter spanExporter = InMemorySpanExporter.create();
+
+  private OpenTelemetrySdk clientSdk;
```
anuraaga, line 44: Maybe we don't need these fields (just came to mind since so many SDKs here is a bit hard to follow)

anuraaga (approved):

integration-tests/src/test/java/io/opentelemetry/B3PropagationIntegrationTest.java
```diff
@@ -0,0 +1,247 @@
...
+      server2.stop();
+    }
+    if (server1 != null) {
+      server1.stop();
+    }
+    spanExporter.reset();
+  }
+
+  @ParameterizedTest
+  @MethodSource("propagators")
```
anuraaga, line 73: BTW personally prefer ArgumentSource for compile safety and to avoid the suppresswarnings

jkwatson:
> These ITs look great!
> 
> Small suggestion - it would be great if these ITs could leverage the classes:
> `SdkTracerProviderConfigurer` and `ConfigurablePropagatorProvider` (so we can detect if the system of loading in custom code has changed)

I don't understand how adding these components to these tests would help you.

integration-tests/src/test/java/io/opentelemetry/B3PropagationIntegrationTest.java
jkwatson: :thu

integration-tests/src/test/java/io/opentelemetry/B3PropagationIntegrationTest.java
jkwatson: good idea. I can streamline this a bit.

PR #3079, opened 2021-03-30
Add autoconfiguration SdkMeterProviderConfigurer SPI. (#3060)

### LINKED ISSUE #3060: Autoconfiguration: provide option for configuring SdkMeterProvider
The autoconfiguration module doesn't provide any way configure the `SdkMeterProvider` other than specifying the resource associated with it. I want to be able to configure the views for the meter provider for `opentelemetry-java-instrumentation`, which relies on autoconfiguration.

It would be really helpful to be able to have configurability around view registry normally configured via `SdkMeterProvider.registerView(InstrumentSelector, AggregatorFactory)`.

Having an SPI hook for metrics analogous to the existing `SdkTracerProviderConfigurer` seems natural and a good starting point, i.e. `SdkMeterProviderConfigurer`. However, the `SdkMeterProviderBuilder` currently doesn't provide any configurability around the views, so the builder would need to be adjusted in some fashion.

Could have a set of env / system properties based configuration as well.

jack-berg:
Adds `SdkMeterProviderConfigurer` SPI to the autoconfiguration module. 

Closes #3060.

### DIFF
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/SdkMeterProviderConfigurer.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/SdkMeterProviderConfigurer.java
@@ -1,0 +1,19 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.autoconfigure.spi;
+
+import io.opentelemetry.sdk.metrics.SdkMeterProviderBuilder;
+
+/**
+ * A service provider interface (SPI) for performing additional programmatic configuration of a
+ * {@link SdkMeterProviderBuilder} during initialization. When using auto-configuration, you should
+ * prefer to use system properties or environment variables for configuration, but this may be
+ * useful to register components that are not part of the SDK such as registering views.
+ */
+public interface SdkMeterProviderConfigurer {
+  /** Configures the {@link SdkMeterProviderBuilder}. */
+  void configure(SdkMeterProviderBuilder meterProviderBuilder);
+}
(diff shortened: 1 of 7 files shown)


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/SdkMeterProviderConfigurer.java
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * A service provider interface (SPI) for performing additional programmatic configuration of a
+ * {@link SdkMeterProviderBuilder} during initialization. When using auto-configuration, you should
+ * prefer to use system properties or environment variables for configuration, but this may be
+ * useful to register components that are not part of the SDK such as registering views.
+ */
+public interface SdkMeterProviderConfigurer {
+
+  void configure(SdkMeterProviderBuilder meterProviderBuilder);
```
jkwatson, line 18: we should have some javadoc on the method as well.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProviderBuilder.java
```diff
@@ -74,7 +74,8 @@ public SdkMeterProviderBuilder setResource(@Nonnull Resource resource) {
    *   .setAggregatorFactory(aggregatorFactory).build());
    * }</pre>
    */
-  public SdkMeterProviderBuilder registerView(InstrumentSelector selector, View view) {
+  public SdkMeterProviderBuilder registerView(
+      @Nonnull InstrumentSelector selector, @Nonnull View view) {
```
jkwatson, line 78: is this necessary? I thought we had `@ParametersAreNonNullByDefault` everywhere.

jack-berg:
The build was failing because of an apparent bug in the `AsynchronousInstrumentAccumulator` which isn't applying registered `LabelProcessorFactory`. I've added a commit that fixes that bug to this PR. Let me know if I should break that out into its own PR.

jack-berg:
@jkwatson added a separate issue for the `AsynchronousInstrumentAccumulator` fix #3085, and proposed a fix in #3086.

jkwatson:
Looks like this PR shows the same failure as is happening on the main branch, post merging of #3086

jack-berg:
Rebased this on the main branch and it looks like everything is good now. 👍

jkwatson (approved):
Thanks!

PR #3080, opened 2021-03-30
Add automation for website docs updates

austinlparker:
This is a GitHub action that will open a PR with your `website_docs` updates to the opentelemetry.io repository on a published release (or manually).

PTAL, I assume the `secrets.GITHUB_TOKEN` is a personal access token with appropriate permissions on the org level? 

cc @jkwatson


jkwatson:
I think we'd only want to trigger this manually, since we usually update our docs after the release has proven successful (due to the vagaries of getting successful releases pushed out to maven central reliably).

austinlparker:
Updated to remove the automation so it's a manual trigger only.

jkwatson (approved):
Let's try this out after the docs are updated to be current

.github/workflows/docs-update.yml
```diff
@@ -0,0 +1,36 @@
...
+          DESTINATION_PATH: content/en/docs/java
+          # Source path should be 'website_docs', all files and folders are copied from here to dest
+          SOURCE_PATH: website_docs
+        run: |
+          TARGET_DIR=$(mktemp -d)
+          export GITHUB_TOKEN=$API_TOKEN_GITHUB
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git clone "https://$API_TOKEN_GITHUB@github.com/$DESTINATION_REPO.git" "$TARGET_DIR"
+          cp -r $SOURCE_PATH/* "$TARGET_DIR/$DESTINATION_PATH"
```
anuraaga, line 28: I think we need to use rsync to make sure deleted files are deleted.

.github/workflows/docs-update.yml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  update-docs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: make-pr
+        env:
+          API_TOKEN_GITHUB: ${{secrets.GITHUB_TOKEN}}
```
anuraaga, line 15: `GITHUB_TOKEN` is the repo's automated secret, not a PAT. I guess it won't be able to push to the destination repo. Is it possible to work with open-telemetry admins to have an org-scoped secret with a PAT for docs pushes that is exposed to our (and other) repo?

anuraaga:
> I think we'd only want to trigger this manually, since we usually update our docs after the release has proven successful (due to the vagaries of getting successful releases pushed out to maven central reliably).

Do we have a reason not to update any time on main branch push that affects the docs folder?

jkwatson:
> Do we have a reason not to update any time on main branch push that affects the docs folder?

That sounds like a good idea!

.github/workflows/docs-update.yml
austinlparker: hm, yeah, that's a good point. I assumed that since it was a public repo in the same org the permissions would be fine. Let me open an issue in the community repo to get a token.

austinlparker:
> Do we have a reason not to update any time on main branch push that affects the docs folder?

One point I'd make here is that I'm not sure you'd want the docs to update every time you push main, as it seems likely that there'll be a drift between `main` and whatever the most recent release is.

jkwatson:
> > Do we have a reason not to update any time on main branch push that affects the docs folder?
> 
> One point I'd make here is that I'm not sure you'd want the docs to update every time you push docs updates to main, as it seems likely that there'll be a drift between `main` and whatever the most recent release is.

We update the docs based on the latest release, not what is on the main branch.

austinlparker:
ah, well, whatever works for y'all.

jkwatson:
I assume this isn't ready to merge until the org-scoped token is ready to go?

austinlparker:
> I assume this isn't ready to merge until the org-scoped token is ready to go?

Yeah.

austinlparker:
@jkwatson per the issue I opened in community, it looks like the decision for this was to have SIG maintainers create a PAT from one of their accounts. Since the website is a public repo, I don't think it needs any special permissions? I have public_repo, write:org, read:org. I think you could create one and assign it to a new secret (I just updated the secret name to `GITHUB_PAT_TOKEN`) and that'd work?

.github/workflows/docs-update.yml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  update-docs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: make-pr
+        env:
+          API_TOKEN_GITHUB: ${{secrets.GITHUB_PAT_TOKEN}}
```
jkwatson, line 15: for unknown reasons, it looks like this is not a valid secret name. How about `DOC_UPDATE_TOKEN` ?

.github/workflows/docs-update.yml
```diff
@@ -0,0 +1,36 @@
...
+          # Destination repo should always be 'open-telemetry/opentelemetry.io'
+          DESTINATION_REPO: open-telemetry/opentelemetry.io
```
anuraaga, line 17: I don't think any of the maintainers here will be able to push a branch to this repo - it'd probably need to itself be a personal fork. I think it's a bit too complicated, let me comment on the issue to see if we can improve it.

austinlparker: Maybe it should be my PAT?

austinlparker:
crap, i screwed up the rebase. let me open a new PR

PR #3086, opened 2021-03-31
Fix bug where AsynchronousInstrumentAccumulator does not call configured LabelProcessorFactory. (#3085)

jack-berg:
Fixes #3085.

PR #3087, opened 2021-04-01
X-Ray traceid extraction - 64bit support

### LINKED ISSUE #3075: Enhance X-Ray propagator to support 64 bits traces extraction
Current X-Ray propagator (https://github.com/open-telemetry/opentelemetry-java/blob/main/extensions/aws/src/main/java/io/opentelemetry/extension/aws/AwsXrayPropagator.java) is very strict, requiring full 128 bit trace Id for extraction. In particular, second part of the ID has to be 8 chars long. The problem arises with 64 bit traces, padded with 0s, that get compressed by AWS. 

Consider following scenario:
- B3 configured as propagator for a service
- 64 bit traced id coming with request (`795bf8769fe8b9c0`), extracted, padded with 0s - OTEL internal 128 bit representation
- trace Id injected into outgoing AWS SDK request (say - SQS), second part (8 chars) - all 0s - `Root=1-00000000-00000000795bf8769fe8b9c0;Parent=7a339e56d8f86bb1;Sampled=1`
- trace id extracted upon consuming SQS message - first part (0s) compressed by AWS to a single 0 - `Root=1-0-00000000795bf8769fe8b9c0;Parent=7a339e56d8f86bb1;Sampled=1`
- AwsXrayPropagator unable to recognize the trace id upon injection, new trace id is created - propagation broken

Enhance AwsXrayPropagator to recognize "compression" of the second part as done by AWS. In such case the ID should stil be recognized as valid and left-padded with 0s.

Filing issue towards AWS to disable "compression" of leading 0s.

**Additional context**
Tested with current agent and AWS infra, leading 0s are always compressed (removed). 

Tagging @anuraaga for AWS visibility :) If the proposal makes sense, should be relatively easy to implement.

kubawach:
Closes #3075
Now epoch part of the X-Ray can be between 1 and 8 chars, which allows 64bits with truncated leading 0s to be extracted successfully.


extensions/aws/src/main/java/io/opentelemetry/extension/aws/AwsXrayPropagator.java
```diff
@@ -210,16 +211,20 @@ private static String parseTraceId(String xrayTraceId) {
     }
 
     // Check delimiters
-    if (xrayTraceId.charAt(TRACE_ID_DELIMITER_INDEX_1) != TRACE_ID_DELIMITER
-        || xrayTraceId.charAt(TRACE_ID_DELIMITER_INDEX_2) != TRACE_ID_DELIMITER) {
+    int firstDelimiter = xrayTraceId.indexOf(TRACE_ID_DELIMITER);
```
anuraaga, line 214: Can we keep the original code and add a parseSlow for the short case? It's against spec so not fair all parsing gets a bit slower :)

kubawach: Definitely! Please have a look at the updated version.

extensions/aws/src/test/java/io/opentelemetry/extension/aws/AwsXrayPropagatorTest.java
```diff
@@ -276,6 +276,85 @@ void extract_nullGetter() {
...
+        .isEqualTo(
+            SpanContext.createFromRemoteParent(
+                "00000000d188f8fa79d48a391a778fa6",
+                SPAN_ID,
+                TraceFlags.getSampled(),
+                TraceState.getDefault()));
+  }
+
+  @Test
+  void extract_epochParTwoChars() {
```
anuraaga, line 296: Think you meant part

anuraaga (approved):
Thanks!

extensions/aws/src/test/java/io/opentelemetry/extension/aws/AwsXrayPropagatorTest.java
kubawach: Indeed :D

jkwatson (approved):
Would just like an scare-comment in the StringUtils about it being internal, then good to merge.

api/all/src/main/java/io/opentelemetry/api/internal/StringUtils.java
```diff
@@ -5,12 +5,73 @@
 
 package io.opentelemetry.api.internal;
 
+import java.util.Objects;
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
-/** Internal utility methods for working with attribute keys, attribute values, and metric names. */
+/** Utilities for working with strings. */
```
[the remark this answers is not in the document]
kubawach: Good idea, done.

jkwatson:
Looks like this is being bitten by whatever is causing the build failures from #3086  as well.

kubawach:
> Looks like this is being bitten by whatever is causing the build failures from #3086 as well.

Yep - and locally the same test runs quite OK.

jkwatson:
> > Looks like this is being bitten by whatever is causing the build failures from #3086 as well.
> 
> Yep - and locally the same test runs quite OK.

It looks like the failure is only on ubuntu, and everything is 100% ok on macos.

kubawach:
> > > Looks like this is being bitten by whatever is causing the build failures from #3086 as well.
> > 
> > 
> > Yep - and locally the same test runs quite OK.
> 
> It looks like the failure is only on ubuntu, and everything is 100% ok on macos.

willing to take a risk and merge? 🤔 😀

jkwatson:
> > > > Looks like this is being bitten by whatever is causing the build failures from #3086 as well.
> > > 
> > > 
> > > Yep - and locally the same test runs quite OK.
> > 
> > 
> > It looks like the failure is only on ubuntu, and everything is 100% ok on macos.
> 
> willing to take a risk and merge? 🤔 😀

I don't want to merge anything further to the main branch until we've got the CI issue fixed.

kubawach:
> #3086

I see that it has been fixed, @jkwatson can you please kick rebuild on this PR?

jkwatson:
> > #3086
> 
> I see that it has been fixed, @jkwatson can you please kick rebuild on this PR?

Where do you see it's been fixed?

kubawach:
> > > #3086
> > 
> > 
> > I see that it has been fixed, @jkwatson can you please kick rebuild on this PR?
> 
> Where do you see it's been fixed?

I saw your comment about #3086 and somehow assumed this has been reverted. Yet I see it hasn't been yet (neither reverted nor fixed). So I'll wait then 🤷‍♂️

jkwatson:
ok, now the breaking bit has been reverted. :)

api/all/src/main/java/io/opentelemetry/api/internal/StringUtils.java
```diff
@@ -5,12 +5,78 @@
...
 
+import java.util.Objects;
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
-/** Internal utility methods for working with attribute keys, attribute values, and metric names. */
+/**
+ * Utilities for working with strings.
+ *
+ * <p>This class is internal and is hence not for public use. Its APIs are unstable and can change *
```
anuraaga, line 15:
```suggestion
 * <p>This class is internal and is hence not for public use. Its APIs are unstable and can change
```

PR #2964, opened 2021-03-02
labels processor

parallelstream:
* Introducing Labels Processor. The idea that this is a interception mechanism to manipulate labels during instrument binding or value recording.
* Couple of representative use cases:
* * I would like to add certain labels extracted from baggage/context to all measurements recorded
* * I would like to strip certain labels from measurements (useful in case pre-instrumented frameworks)

It is an implementation for https://github.com/open-telemetry/opentelemetry-java/issues/1810. Based on https://github.com/open-telemetry/opentelemetry-java/pull/2956 and whole idea is sketched in this PR https://github.com/open-telemetry/opentelemetry-java/pull/2800

PR #3088, opened 2021-04-01
Revert "Fix bug where AsynchronousInstrumentAccumulator does not call configured LabelProcessorFactory.

jkwatson:
Testing to see if this broke the main branch build in some unexplainable way


jkwatson:
@jack-berg I can't figure out why, but it seems like your bugfix has broken the main branch build for autoconfiguration. Do you have time to figure out what's going on?

jkwatson:
Note: the build seems to only be failing on ubuntu, not on the macos build; not sure if that's a clue or a red herring.

jack-berg:
I can try to look into it.

jack-berg:
@jkwatson is it possible its a fluke? The failure in the test is from traces failing to show up. Hard to imagine that this seemingly benign change to metrics could impact traces.

jack-berg:
I suppose several builds have failed for the same reason so its likely not a fluke.

jkwatson:
> I suppose several builds have failed for the same reason so its likely not a fluke.

And, I've run the build on this branch several times and it passes every time. It's related, but I don't understand why yet, either.

jack-berg:
@jkwatson the [javadoc](https://github.com/open-telemetry/opentelemetry-java/pull/2964/files#diff-52808450af51a5923c93ea4522d1c6c3551f3df2babe7b1a19220ddf242456bbR16-R21) on `LabelsProcessor#onLabelsBound` suggests that the concept is only suitable for synchronous instruments. Is this "bug fix" I've done conceptually flawed?

And if label processing is limited to synchronous instruments, then is `ViewRegistry#registerView(InstrumentSelector, View)` flawed? `InstrumentSelector` forces you to select an instrument, synchronous or asynchronous, and `View` forces you to specify a `LabelsProcessorFactory`, which may not make sense for asynchronous instruments?

jack-berg:
I think its probably something to do with `Context.current()` evaluation inside the asynchronous observation, but am not familiar enough with context and its complexities to figure it out. 

I think its probably best if we merge this revert.

jkwatson:
I've been working on this as well, and it looks like the issue might be that the OtlpExporter(s) aren't being cleanly shutdown in the FullConfigTest...when I run the build at --info I can clearly see messages yelling about this.

jkwatson:
ok, let's merge this so we can at least get the `main` branch building again while we try to sort out what's up with the autoconfigure test. @anuraaga might have some ideas, since it looks like the armeria server is what is having troubles starting up properly.

bogdandrutu (approved):

jkwatson:
looks like reverting this did *not* fix the CI issues, so I'll put in a PR to revert this revert.

PR #3094, opened 2021-04-02
Add docker build for tracecontext testsuite.

anuraaga:
It'll be nice to run this with testcontainers.


jkwatson (approved):
seems fine. What are you going to use this for?

anuraaga:
@jkwatson When you sent your cleanup PR to this I noticed it's annoying to run the test right now - in fact, we should avoid scripts in our repo that execute pip install commands potentially into someone's global environment :)

PR #3096, opened 2021-04-02
Migrate tracecontext test app to Armeria.

anuraaga:
This simplifies the code a bit by having classes available more convenient than just the servlet API. Will follow up with seeing if it can make it easier to pass more tests. I confirmed that this fails the same tests as current code.

### DIFF
--- a/integration-tests/tracecontext/src/main/java/io/opentelemetry/Application.java
+++ b/integration-tests/tracecontext/src/main/java/io/opentelemetry/Application.java
@@ -67,7 +67,52 @@ public final class Application {
       if (carrier == null) {
         return;
       }
-      carrier.add(key, value);
+      carrier.set(key, value);
+    }
+  }
+
+  private static class Service {
+    private final WebClient client = WebClient.of();
+
+    @Post("/verify-tracecontext")
+    @Blocking
+    public String serve(RequestHeaders headers, List<Request> requests) {
+      Context context =
+          openTelemetry
+              .getPropagators()
+              .getTextMapPropagator()
+              .extract(Context.current(), headers, ArmeriaGetter.INSTANCE);
+
+      for (io.opentelemetry.Request req : requests) {
+        Span span =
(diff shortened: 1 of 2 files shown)


integration-tests/tracecontext/src/main/java/io/opentelemetry/Application.java
```diff
@@ -34,99 +41,93 @@
...
+
+  private enum ArmeriaSetter implements TextMapSetter<RequestHeadersBuilder> {
+    INSTANCE;
+
+    @Override
+    public void set(@Nullable RequestHeadersBuilder carrier, String key, String value) {
+      if (carrier == null) {
+        return;
+      }
+      carrier.add(key, value);
```
kubawach, line 70: I recall there is an issue filled to use set() instead of add() for propagation headers in order to avoid duplicates (eg when reusing a request instance). I know this is only test code, so not influencing a real instrumentation, but just to stay aligned with "best practice", why not?

integration-tests/tracecontext/src/main/java/io/opentelemetry/Application.java
```diff
@@ -34,99 +41,93 @@
...
-        });
-
-    Spark.awaitInitialization();
+    WebClient client = WebClient.of();
+
+    Server server =
+        Server.builder()
+            .http(5000)
+            .annotatedService(
+                new Object() {
```
kubawach, line 84: I believe that splitting this into a separate class (even private) would improve readability a bit.

kubawach (approved):

integration-tests/tracecontext/src/main/java/io/opentelemetry/Application.java
anuraaga: Sure

jkwatson (approved):
Seems like a good cleanup! Much nicer than working with bare servlet APIs.

PR #3098, opened 2021-04-02
Update the CHANGELOG as of APR-02-2021 and add @since tags for new APIs.

jkwatson:


opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/OpenTracingPropagators.java
```diff
@@ -10,6 +10,8 @@
 /**
  * Container for {@link io.opentracing.propagation.Format.Builtin#TEXT_MAP} and {@link
  * io.opentracing.propagation.Format.Builtin#HTTP_HEADERS} format propagators.
+ *
+ * @since 1.1.0
```
anuraaga, line 14: Doh! Thanks for finding these, need to remember to check during PRs

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -169,14 +169,20 @@ private static boolean isValidAndNotEmpty(AttributeKey<?> name) {
     return !name.getKey().isEmpty() && isValid(name.getKey());
```
anuraaga, line 169: Believe we need to add to `ResourceBuilder` too

anuraaga (approved):

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
jkwatson: let's do that separately. :)

jkwatson: Although, it does get called when the builder is used, just not until you actually try to build the Resources. I guess it might be nicer to fail earlier, but in practice I think it would get caught properly. Please create an issue if you think it should fail earlier when using the builder.

anuraaga: I think my comment hit the wrong line - I meant we should add the since tag to that class too not just these methods.

jkwatson: oh! sorry about that. Yes, true true. :)

PR #3103, opened 2021-04-05
Run w3c tracecontext tests as part of build.

anuraaga:
Due to many failures can't actually assert anything yet but it's still worth having logs of where we are.


jkwatson (approved):
Thanks! This is a good start!

PR #3104, opened 2021-04-05
Don't crash on empty tracestate value

anuraaga:


jkwatson (approved):
Good fix

anuraaga:
@Oberon00 Doesn't the `nblk-chr` after it enforce there's a character?

Oberon00:
You are right 🙈  Sorry for the false alarm.

PR #3106, opened 2021-04-05
Map x-ray key/value pairs to baggage.

anuraaga:


anuraaga:
@willarmiros Please review too :)

extensions/aws/src/main/java/io/opentelemetry/extension/aws/AwsXrayPropagator.java
```diff
@@ -169,7 +179,7 @@ public static AwsXrayPropagator getInstance() {
             "Error parsing X-Ray trace header. Invalid key value pair: "
                 + part
                 + " Returning INVALID span context.");
-        return SpanContext.getInvalid();
+        return context;
```
jkwatson, line 182: ah, good fix in here, too

extensions/aws/src/main/java/io/opentelemetry/extension/aws/AwsXrayPropagator.java
```diff
@@ -107,19 +109,30 @@ public static AwsXrayPropagator getInstance() {
...
+        .append(KV_DELIMITER)
+        .append(parentId)
+        .append(TRACE_HEADER_DELIMITER)
+        .append(SAMPLED_FLAG_KEY)
+        .append(KV_DELIMITER)
+        .append(samplingFlag);
+
+    Baggage baggage = Baggage.fromContext(context);
+    baggage.forEach(
+        (key, value) ->
```
jkwatson, line 128: any worries here about blowing out the header if there's a lot shoved into the baggage?

extensions/aws/src/test/java/io/opentelemetry/extension/aws/AwsXrayPropagatorTest.java
```diff
@@ -168,11 +189,12 @@ void extract_AdditionalFields() {
...
         "Root=1-8a3c60f7-d188f8fa79d48a391a778fa6;Parent=53995c3f42cd8ad8;Sampled=1;Foo=Bar");
 
-    // TODO: assert additional fields when the propagator supports it
-    assertThat(getSpanContext(xrayPropagator.extract(Context.current(), carrier, getter)))
+    Context context = xrayPropagator.extract(Context.current(), carrier, getter);
+    assertThat(getSpanContext(context))
         .isEqualTo(
             SpanContext.createFromRemoteParent(
                 TRACE_ID, SPAN_ID, TraceFlags.getSampled(), TraceState.getDefault()));
+    assertThat(Baggage.fromContext(context).getEntryValue("Foo")).isEqualTo("Bar");
```
jkwatson, line 197: any worries with people putting your keys into the Baggage by accident? So, a baggage entry called `Root` or `Sampled` or `Parent` ?

extensions/aws/src/main/java/io/opentelemetry/extension/aws/AwsXrayPropagator.java
anuraaga: I think similar to how tracestate is handled by tracestatebuilder, for better or worse, baggage would similar be constrained at that logic layer instead of propagation.

extensions/aws/src/test/java/io/opentelemetry/extension/aws/AwsXrayPropagatorTest.java
anuraaga: Ah yeah we'd want to ignore it (I don't think the x-ray sdk is so sophisticated though...). I think that's what happens will add unit test.

jkwatson (approved):

extensions/aws/src/main/java/io/opentelemetry/extension/aws/AwsXrayPropagator.java
```diff
@@ -190,14 +210,21 @@ public static AwsXrayPropagator getInstance() {
               + "' with value "
               + traceHeader
               + "'. Returning INVALID span context.");
```
willarmiros, line 212: Is this log message still correct to say we're returning INVALID context? It is returning *an* invalid context, but is it returning *the* invalid context?

extensions/aws/src/main/java/io/opentelemetry/extension/aws/AwsXrayPropagator.java
```diff
@@ -180,8 +196,12 @@ public static AwsXrayPropagator getInstance() {
         spanId = parseSpanId(value);
       } else if (trimmedPart.startsWith(SAMPLED_FLAG_KEY)) {
         isSampled = parseTraceFlag(value);
+      } else {
+        if (baggage == null) {
+          baggage = Baggage.builder();
+        }
+        baggage.put(trimmedPart.substring(0, equalsIndex), value.trim());
```
willarmiros, line 203: So interestingly the X-Ray Trace header spec has a limit of 256 bytes on incoming arbitrary key-value pairs after the defined fields (baggage). This is not publicly documented, and is only implemented in the [Node SDK](https://github.com/aws/aws-xray-sdk-node/blob/b88909ef98c1b29c82691ced4153d1693c872c8f/packages/core/lib/utils.js#L194). So clearly it's not critical to matching the behavior of the X-Ray SDKs, but I think it's still worth calling out since a limit on the baggage size is a safe thing to do to prevent malicious headers from stalling a program. Would be interested to hear your thoughts/if there are any existing considerations for this in OTel.

willarmiros (approved):
Generally looks good & compatible with how X-Ray treats baggage today. But wondering if we can or should raise the bar and implement a piece of the X-Ray spec. In addition we'd actually document this behavior publicly which it is not today.

extensions/aws/src/main/java/io/opentelemetry/extension/aws/AwsXrayPropagator.java
anuraaga: Thanks for calling this out! The interesting point is to compare to w3c baggage

https://w3c.github.io/baggage/#limits

It's limit is 8192 - it's different but I don't see a big reason for that to affect how xray propagator itself works given it's a different spec. And smaller can be increased to larger later if we ever needed to. Will add the limit.

extensions/aws/src/main/java/io/opentelemetry/extension/aws/AwsXrayPropagator.java
anuraaga: Thanks, just removed that bit since it's sort of an implementation detail, logging the actual problem is what matters.

anuraaga (commented):
Added limit

extensions/aws/src/main/java/io/opentelemetry/extension/aws/AwsXrayPropagator.java
willarmiros: Awesome! Good to know about the w3c limits. LGTM.

PR #3120, opened 2021-04-06
Fix more tracestate testsuite cases.

anuraaga:
Got down to just 6 failures of the testsuite, all requiring support for handling multiple instances of a header which OpenTelemetry doesn't support yet.

Biggest changes were

- Trace state discarded if any item fails (I don't see this in the spec but it's being tested by test suite)
- Load SDK since trace ID generation is tested

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/trace/ArrayBasedTraceStateBuilder.java
+++ b/api/all/src/main/java/io/opentelemetry/api/trace/ArrayBasedTraceStateBuilder.java
@@ -126,8 +126,9 @@ final class ArrayBasedTraceStateBuilder implements TraceStateBuilder {
         if (i > MAX_TENANT_ID_SIZE) {
           return false;
         }
-        // vendor id (the part to the right of the '@' sign) must be 13 characters or less
-        if (key.length() - i > MAX_VENDOR_ID_SIZE) {
+        // vendor id (the part to the right of the '@' sign) must be 1-13 characters long
+        int remainingKeyChars = key.length() - i - 1;
+        if (remainingKeyChars > MAX_VENDOR_ID_SIZE || remainingKeyChars == 0) {
           return false;
         }
       }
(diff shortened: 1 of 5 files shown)


api/all/src/main/java/io/opentelemetry/api/trace/ArrayBasedTraceStateBuilder.java
```diff
@@ -126,8 +126,9 @@ private static boolean isKeyValid(@Nullable String key) {
         if (i > MAX_TENANT_ID_SIZE) {
           return false;
         }
-        // vendor id (the part to the right of the '@' sign) must be 13 characters or less
-        if (key.length() - i > MAX_VENDOR_ID_SIZE) {
+        // vendor id (the part to the right of the '@' sign) must be 1-13 characters long
+        int remainingKeyChars = key.length() - i - 1;
```
anuraaga, line 130: Had off-by-one error, if length is 4, the i for the last char is 3

api/all/src/main/java/io/opentelemetry/api/trace/propagation/W3CTraceContextPropagator.java
```diff
@@ -245,6 +245,11 @@ private static TraceState extractTraceState(String traceStateHeader) {
       checkArgument(index != -1, "Invalid TraceState list-member format.");
       traceStateBuilder.put(listMember.substring(0, index), listMember.substring(index + 1));
     }
-    return traceStateBuilder.build();
+    TraceState traceState = traceStateBuilder.build();
+    if (traceState.size() != listMembers.length) {
+      // Validation failure, drop the tracestate
```
Oberon00, line 250: Foes this happen only with duplicate keys?

anuraaga: This happens for any sort of validation error, if a key or value doesn't validate we don't populate it into TraceState

Oberon00: Ah, because `build` silently discards it? Makes sense.

jkwatson (approved):
Thanks!

PR #3121, opened 2021-04-06
Remove credential info from url in example 

### LINKED ISSUE #3118: HTTP Span Attributes: `http.url` must not contain username / password
## Why

As is stated in the recent [specification change](https://github.com/open-telemetry/opentelemetry-specification/pull/1502) :

> `http.url` MUST NOT contain credentials passed via URL in form of `https://username:password@www.example.com/`. In such case the attribute's value should be `https://www.example.com/`

## What

[This example](https://github.com/open-telemetry/opentelemetry-java/blob/04fbc2e6c52100eff0ada86d6df61adbc8d09569/examples/http/src/main/java/io/opentelemetry/example/http/HttpClient.java#L62 ) could be reviewed (and fixed) to make sure that others do not copy-paste potentially unsafe code.

Hangzhi:
Fix #3118

### DIFF
--- a/examples/http/src/main/java/io/opentelemetry/example/http/HttpClient.java
+++ b/examples/http/src/main/java/io/opentelemetry/example/http/HttpClient.java
@@ -60,12 +60,11 @@ public final class HttpClient {
          - http.scheme, peer.ip, peer.port, http.target
       */
 
-      if (url.getUserInfo() == null || url.getUserInfo().isEmpty()){
+      if (url.getUserInfo() == null || url.getUserInfo().isEmpty()) {
         span.setAttribute(SemanticAttributes.HTTP_URL, url.toString());
-      }else{
+      } else {
         span.setAttribute(url.toString().replace(url.getUserInfo() + '@', ""));
       }
-
 
       // Inject the request with the current Context/Span.
       textMapPropagator.inject(Context.current(), con, setter);
--- a/examples/http/src/main/java/io/opentelemetry/example/http/HttpClient.java
+++ b/examples/http/src/main/java/io/opentelemetry/example/http/HttpClient.java
@@ -61,7 +61,7 @@ public final class HttpClient {
          - http.scheme, peer.hostname, peer.port, http.target
          - http.scheme, peer.ip, peer.port, http.target
       */
-      uri = url.toURI();
+      URI uri = url.toURI();
       url =
           new URI(
                   uri.getScheme(),
--- a/examples/http/src/main/java/io/opentelemetry/example/http/HttpClient.java
+++ b/examples/http/src/main/java/io/opentelemetry/example/http/HttpClient.java
@@ -53,13 +53,19 @@ public final class HttpClient {
       span.setAttribute(SemanticAttributes.HTTP_METHOD, "GET");
       span.setAttribute("component", "http");
       /*
-       Only one of the following is required:
+       Only one of the following is required, and url must not contain credential in url.
          - http.url
          - http.scheme, http.host, http.target
          - http.scheme, peer.hostname, peer.port, http.target
          - http.scheme, peer.ip, peer.port, http.target
       */
-      span.setAttribute(SemanticAttributes.HTTP_URL, url.toString());
+
+      if (url.getUserInfo() == null || url.getUserInfo().isEmpty()){
+        span.setAttribute(SemanticAttributes.HTTP_URL, url.toString());
+      }else{
+        span.setAttribute(url.toString().replace(url.getUserInfo() + '@', ""));
+      }
+
 
       // Inject the request with the current Context/Span.
       textMapPropagator.inject(Context.current(), con, setter);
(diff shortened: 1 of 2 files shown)


examples/logging/src/main/java/io/opentelemetry/example/logging/LoggingExporterExample.java
```diff
@@ -19,7 +19,7 @@
   public LoggingExporterExample(OpenTelemetry openTelemetry) {
     tracer = openTelemetry.getTracer(INSTRUMENTATION_NAME);
     counter =
-        GlobalMetricsProvider.getMeter(INSTRUMENTATION_NAME)
+        GlobalMeterProvider.getMeter(INSTRUMENTATION_NAME)
```
Hangzhi: Apart from removing the user info in the example above. 
There appears "Cannot find symbol" in this logging example. I did minor modifications base on my understanding.

examples/logging/src/main/java/io/opentelemetry/example/logging/LoggingExporterExample.java
```diff
@@ -19,9 +19,7 @@
   public LoggingExporterExample(OpenTelemetry openTelemetry) {
     tracer = openTelemetry.getTracer(INSTRUMENTATION_NAME);
     counter =
-        GlobalMetricsProvider.getMeter(INSTRUMENTATION_NAME)
-            .longCounterBuilder("work_done")
-            .build();
+        GlobalMeterProvider.getMeter(INSTRUMENTATION_NAME).longCounterBuilder("work_done").build();
```
jkwatson, line 22: Our examples run against the latest release, not against the main branch, so this will get updated after we release 1.1.0. For now, it needs to stay as it was.

jkwatson: In fact, I'm quite concerned that the build seems to be passing with this... something might be amiss with our examples check as a part of CI.

jkwatson: Hm. This is actually annoyingly tricky to get right. We really want our examples to do 2 things:
1) Verify that we haven't broken anything in the stable APIs and
2) Provide good examples that work against the latest official release (not SNAPSHOT builds).

Unfortunately, I don't think it's possible to meet both of these goals for all the examples, since the stable SDK depends on the unstable metrics, so pulling when we build against snapshots for the stable stuff (to verify that we haven't broken anything), we pull in the unstable metrics version as well. 

Note: this is not your problem to solve in this PR...I'm just clarifying for future us. I'll open an issue with these details so we can make this less confusing. :)

Hangzhi: But I think the PR cannot be merged without passing all the tests. What can I do, besides leaving the logging example as what it was and removing the user info in the URL?  It seems that further discussion is needed for this conflict to be resolved.

jkwatson: I'd like to get #3122 fixed before we merge this, so if you're interested in working on that, that would be great.

jkwatson: Note: Now that we've released 1.1.0, this won't be an issue for this PR. Once #3127 has been merged, you can rebase and this PR will be good to go.

jkwatson: (#3122 is still an issue, long term, it's just not a blocker for this PR any more. :)

Hangzhi: Good! I'll do that after #3127 being merged

jkwatson: Merged now! Thanks!

examples/http/src/main/java/io/opentelemetry/example/http/HttpClient.java
```diff
@@ -53,13 +53,19 @@ private void makeRequest() throws IOException {
...
          - http.scheme, peer.hostname, peer.port, http.target
          - http.scheme, peer.ip, peer.port, http.target
       */
-      span.setAttribute(SemanticAttributes.HTTP_URL, url.toString());
+
+      if (url.getUserInfo() == null || url.getUserInfo().isEmpty()) {
+        span.setAttribute(SemanticAttributes.HTTP_URL, url.toString());
+      } else {
+        span.setAttribute(
+            SemanticAttributes.HTTP_URL, url.toString().replace(url.getUserInfo() + '@', ""));
```
kubawach, line 67: In the spirit of code safety, I'd apply here same logic as suggested for changes in the instrumentation module: 

```
new URI(
    url.getScheme(),
    null,
    url.getHost(),
    url.getPort(),
    url.getPath(),
    url.getQuery(),
    url.getFragment()).toUrl();
```

kubawach (approved):

examples/http/src/main/java/io/opentelemetry/example/http/HttpClient.java
jkwatson: Yes, this is definitely a safer way to do this work. @Hangzhi Can you make this change as well? With this, you can also get rid of the if/else and just do it once using the URI to remove the user info and format correctly.

jkwatson (approved):
Thanks!

PR #3126, opened 2021-04-07
Implement OpenTracing Shim error logging translation

malafeev:
Spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/opentracing.md#log

### DIFF
--- a/opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
+++ b/opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
@@ -274,7 +274,7 @@ final class SpanShim extends BaseShimObject implements Span {
     Throwable throwable;
     Attributes attributes;
 
-    static ExtractedFields build(Map<String, ?> fields) {
+    private static ExtractedFields build(Map<String, ?> fields) {
       ExtractedFields extractedFields = new ExtractedFields();
       extractedFields.name = getEventNameFromFields(fields);
       extractedFields.throwable = null;
--- a/opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
+++ b/opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
@@ -10,6 +10,7 @@ import static io.opentelemetry.api.common.AttributeKey.doubleKey;
 import static io.opentelemetry.api.common.AttributeKey.longKey;
 import static io.opentelemetry.api.common.AttributeKey.stringKey;
 
+import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.common.AttributesBuilder;
 import io.opentelemetry.api.trace.StatusCode;
@@ -235,23 +236,21 @@ final class SpanShim extends BaseShimObject implements Span {
       } else if (value instanceof Boolean) {
         attributesBuilder.put(booleanKey(key), (Boolean) value);
       } else {
+        AttributeKey<String> attributeKey = null;
         if (isExceptionEvent) {
-          switch (key) {
-            case Fields.ERROR_KIND:
-              key = SemanticAttributes.EXCEPTION_TYPE.getKey();
-              break;
-            case Fields.MESSAGE:
-              key = SemanticAttributes.EXCEPTION_MESSAGE.getKey();
-              break;
-            case Fields.STACK:
-              key = SemanticAttributes.EXCEPTION_STACKTRACE.getKey();
-              break;
-            default:
-              break;
+          if (key.equals(Fields.ERROR_KIND)) {
+            attributeKey = SemanticAttributes.EXCEPTION_TYPE;
+          } else if (key.equals(Fields.MESSAGE)) {
+            attributeKey = SemanticAttributes.EXCEPTION_MESSAGE;
+          } else if (key.equals(Fields.STACK)) {
+            attributeKey = SemanticAttributes.EXCEPTION_STACKTRACE;
           }
         }
         if (!isErrorEvent || !key.equals(Fields.ERROR_OBJECT) || !(value instanceof Throwable)) {
-          attributesBuilder.put(stringKey(key), value.toString());
+          if (attributeKey == null) {
+            attributeKey = stringKey(key);
+          }
+          attributesBuilder.put(attributeKey, value.toString());
         }
       }
     }
--- a/opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
+++ b/opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
@@ -119,7 +119,7 @@ final class SpanShim extends BaseShimObject implements Span {
 
   @Override
   public Span log(Map<String, ?> fields) {
-    ExtractedFields extractedFields = ExtractedFields.build(fields);
+    ExtractedFields extractedFields = new ExtractedFields(fields);
 
     if (extractedFields.throwable != null) {
       span.recordException(extractedFields.throwable, extractedFields.attributes);
@@ -132,7 +132,7 @@ final class SpanShim extends BaseShimObject implements Span {
 
   @Override
   public Span log(long timestampMicroseconds, Map<String, ?> fields) {
-    ExtractedFields extractedFields = ExtractedFields.build(fields);
+    ExtractedFields extractedFields = new ExtractedFields(fields);
 
     if (extractedFields.throwable != null) {
       // timestamp is not recorded
@@ -274,20 +274,18 @@ final class SpanShim extends BaseShimObject implements Span {
     private final Throwable throwable;
     private final Attributes attributes;
 
-    private static ExtractedFields build(Map<String, ?> fields) {
-      ExtractedFields extractedFields = new ExtractedFields();
-      extractedFields.name = getEventNameFromFields(fields);
-      extractedFields.throwable = null;
+    private ExtractedFields(Map<String, ?> fields) {
+      name = getEventNameFromFields(fields);
       boolean isExceptionEvent = false;
-      if (extractedFields.name.equals("error")) {
-        extractedFields.throwable = findThrowable(fields);
-      } else if (extractedFields.name.equals("exception")) {
+      if (name.equals("error")) {
+        throwable = findThrowable(fields);
+      } else if (name.equals("exception")) {
         isExceptionEvent = true;
+        throwable = null;
+      } else {
+        throwable = null;
       }
-      extractedFields.attributes =
-          convertToAttributes(fields, extractedFields.throwable != null, isExceptionEvent);
-
-      return extractedFields;
+      attributes = convertToAttributes(fields, throwable != null, isExceptionEvent);
     }
   }
 }
--- a/opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
+++ b/opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
@@ -264,13 +264,9 @@ final class SpanShim extends BaseShimObject implements Span {
 
   @Nullable
   private static Throwable findThrowable(Map<String, ?> fields) {
-    for (Map.Entry<String, ?> entry : fields.entrySet()) {
-      String key = entry.getKey();
-      Object value = entry.getValue();
-
-      if (key.equals(Fields.ERROR_OBJECT) && value instanceof Throwable) {
-        return (Throwable) value;
-      }
+    Object value = fields.get(Fields.ERROR_OBJECT);
+    if (value instanceof Throwable) {
+      return (Throwable) value;
     }
     return null;
   }
--- a/opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
+++ b/opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
@@ -13,6 +13,7 @@ import static io.opentelemetry.api.common.AttributeKey.stringKey;
 import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.common.AttributesBuilder;
 import io.opentelemetry.api.trace.StatusCode;
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
 import io.opentracing.Span;
 import io.opentracing.SpanContext;
 import io.opentracing.log.Fields;
@@ -34,6 +35,8 @@ import javax.annotation.Nullable;
  */
 final class SpanShim extends BaseShimObject implements Span {
   private static final String DEFAULT_EVENT_NAME = "log";
+  private static final String ERROR = "error";
+  private static final String EXCEPTION = "exception";
 
   private final io.opentelemetry.api.trace.Span span;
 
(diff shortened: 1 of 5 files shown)


opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
```diff
@@ -214,10 +230,64 @@ static Attributes convertToAttributes(Map<String, ?> fields) {
...
+        return (Throwable) value;
+      }
+    }
+    return null;
+  }
+
+  static class ExtractedFields {
+    String name;
+    Throwable throwable;
+    Attributes attributes;
```
jkwatson, line 275:
```suggestion
  private static class ExtractedFields {
    private final String name;
    private final Throwable throwable;
    private final Attributes attributes;
```

opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
```diff
@@ -214,10 +230,64 @@ static Attributes convertToAttributes(Map<String, ?> fields) {
...
+
+    static ExtractedFields build(Map<String, ?> fields) {
```
jkwatson, line 277:
```suggestion
    private static ExtractedFields build(Map<String, ?> fields) {
```

opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
```diff
@@ -214,10 +230,64 @@ static Attributes convertToAttributes(Map<String, ?> fields) {
...
+        if (!isErrorEvent || !key.equals(Fields.ERROR_OBJECT) || !(value instanceof Throwable)) {
+          attributesBuilder.put(stringKey(key), value.toString());
+        }
       }
     }
 
     return attributesBuilder.build();
   }
+
+  static Throwable findThrowable(Map<String, ?> fields) {
```
jkwatson, line 257:
```suggestion
  private static Throwable findThrowable(Map<String, ?> fields) {
```

jkwatson: In fact, all the helper methods in here need to be made private.

opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
```diff
@@ -214,10 +230,62 @@ static Attributes convertToAttributes(Map<String, ?> fields) {
...
 
     return attributesBuilder.build();
   }
+
+  private static Throwable findThrowable(Map<String, ?> fields) {
+    for (Map.Entry<String, ?> entry : fields.entrySet()) {
+      String key = entry.getKey();
+      Object value = entry.getValue();
+
+      if (value == null) {
```
anuraaga, line 262: Don't think you need this since you have `value instanceof Throwable`

opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
```diff
@@ -119,17 +119,32 @@ public Span setTag(String key, Number value) {
 
   @Override
   public Span log(Map<String, ?> fields) {
-    span.addEvent(getEventNameFromFields(fields), convertToAttributes(fields));
+    ExtractedFields extractedFields = new ExtractedFields(fields);
```
anuraaga, line 122: Do we need this object instead of a helper method with the logic currently in this object's constructor?

opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
malafeev: removed

opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
malafeev: yes, we need it, because I call it in 2 places, second is `log(long timestampMicroseconds, Map<String, ?> fields)`

anuraaga: Yeah - can't you have a helper method which treats `-1` specially, for example? Just trying to avoid the allocation here which doesn't seem to be doing that much.

malafeev: done

opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
```diff
@@ -214,10 +232,39 @@ static Attributes convertToAttributes(Map<String, ?> fields) {
       } else if (value instanceof Boolean) {
         attributesBuilder.put(booleanKey(key), (Boolean) value);
       } else {
-        attributesBuilder.put(stringKey(key), value.toString());
+        if (isExceptionEvent) {
+          switch (key) {
+            case Fields.ERROR_KIND:
+              key = "exception.type";
```
jkwatson, line 238: we should definitely make these three keys be statically initialized (the corresponding AttributeKey instances) to avoid extra allocations in here.

jkwatson: Or, better yet, use the ones already defined in the SemanticAttributes class in the semconv module.

malafeev: `SemanticAttributes` is not available in this module, should I add new dependency?

jkwatson: My opinion: definitely, we should use the semantic conventions in our shim. @carlosalberto what do you think? It would mean that the shim depends on an alpha module (of course, the shim is also alpha, so that seems fine).

carlosalberto: Sounds good for now! Hopefully we don't end up reverting that in case the OT shim becomes stable _and_ semantic conventions are not ;)

jkwatson: If that happens, we can easily copy the needed keys over to the shim in the interim.

malafeev: done

opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
```diff
@@ -214,10 +235,39 @@ static Attributes convertToAttributes(Map<String, ?> fields) {
...
+              break;
+            case Fields.STACK:
+              key = SemanticAttributes.EXCEPTION_STACKTRACE.getKey();
+              break;
+            default:
+              break;
+          }
+        }
+        if (!isErrorEvent || !key.equals(Fields.ERROR_OBJECT) || !(value instanceof Throwable)) {
+          attributesBuilder.put(stringKey(key), value.toString());
```
jkwatson, line 254: This doesn't solve the problem of extra allocations. You're still allocating a new key here, when you already have the key above for the given cases. Also, updating the variable you're switching on inside the switch block is pretty sketchy. Please refactor this code to have another variable for the `AttributeKey` instance that will be used here, so you can get rid of the extra allocations, and make the code cleaner.

malafeev: refactored

opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
```diff
@@ -214,10 +236,37 @@ static Attributes convertToAttributes(Map<String, ?> fields) {
...
+          }
+          attributesBuilder.put(attributeKey, value.toString());
+        }
       }
     }
 
     return attributesBuilder.build();
   }
+
+  private static Throwable findThrowable(Map<String, ?> fields) {
```
anuraaga, line 261:
```suggestion
  @Nullable
  private static Throwable findThrowable(Map<String, ?> fields) {
```

opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
```diff
@@ -183,7 +183,28 @@ public void finish(long finishMicros) {
...
   }
 
-  static String getEventNameFromFields(Map<String, ?> fields) {
+  private void logInternal(long timestampMicroseconds, Map<String, ?> fields) {
+    String name = getEventNameFromFields(fields);
+    boolean isExceptionEvent = false;
+    Throwable throwable = null;
+    if (name.equals(ERROR)) {
+      throwable = findThrowable(fields);
+    } else if (name.equals(EXCEPTION)) {
```
anuraaga, line 192: I can't find this name `exception` in the spec

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/opentracing.md#log

As far as I can tell, we only have "error event", and that error event may or may not have a throwable.

So

```java
boolean isError;
boolean recordedException;
if (getEventNameForFields == "error") {
  isError = true;
  if (throwable != null) {
    recordException()
    recordedException = true;
  }

...

// Filling attributes
if (recordedException) {
  if (key.equals(ERROR_OBJECT)) {
    continue;
  }
  // Don't see anything in spec to skip ERROR_KIND, etc for recordedException case like current code does, should we make clearer in the spec if we think it should be skipped?
}
if (isError) {
  if (key.equals(ERROR_KIND)) {
    key = EXCEPTION_TYPE;
  }
  ...
}
```

opentracing-shim/src/test/java/io/opentelemetry/opentracingshim/SpanShimTest.java
```diff
@@ -104,6 +109,48 @@ void finish_micros() {
     assertThat(spanData.getEndEpochNanos()).isEqualTo(micros * 1000L);
   }
 
+  @Test
```
anuraaga, line 112: The code has a ton of if statements so I suspect just these tests isn't covering it enough :) Didn't do a full look, but for example, if an event has both exception and error fields. Please flesh out all corner cases.

opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
malafeev: Next paragraph is confusing, but it says that name being set to `exception`:
 
> Else, a call to AddEvent is performed with name being set to exception, along the specified key/value pair set as additional event attributes, including mapping of the following key/value pairs:

anuraaga: Yeah that's the name of the OpenTelemetry event, but is still in the paragraph for `If pair set contains a event=error entry`

malafeev: updated

opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
```diff
@@ -183,7 +183,30 @@ public void finish(long finishMicros) {
...
-  static String getEventNameFromFields(Map<String, ?> fields) {
+  private void logInternal(long timestampMicroseconds, Map<String, ?> fields) {
+    String name = getEventNameFromFields(fields);
+    Throwable throwable = null;
+    boolean isError = false;
+    if (name.equals(ERROR)) {
+      throwable = findThrowable(fields);
+      isError = true;
+      if (throwable == null) {
+        name = EXCEPTION;
```
anuraaga, line 194:
```suggestion
        name = SemanticAttributes.EXCEPTION_EVENT_NAME;
```

opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
```diff
@@ -34,6 +36,8 @@
  */
 final class SpanShim extends BaseShimObject implements Span {
   private static final String DEFAULT_EVENT_NAME = "log";
+  private static final String ERROR = "error";
+  private static final String EXCEPTION = "exception";
```
anuraaga, line 40:
```suggestion
```

opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
```diff
@@ -214,10 +238,40 @@ static Attributes convertToAttributes(Map<String, ?> fields) {
...
+        }
       }
     }
 
     return attributesBuilder.build();
   }
+
+  @Nullable
+  private static Throwable findThrowable(Map<String, ?> fields) {
+    for (Map.Entry<String, ?> entry : fields.entrySet()) {
```
anuraaga, line 267: Instead of looping can just be `fields.get(Fields.ERROR_OBJECT)` right?

opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
```diff
@@ -214,10 +238,40 @@ static Attributes convertToAttributes(Map<String, ?> fields) {
...
+        if (isError && !isRecordingException) {
+          if (key.equals(Fields.ERROR_KIND)) {
+            attributeKey = SemanticAttributes.EXCEPTION_TYPE;
+          } else if (key.equals(Fields.MESSAGE)) {
+            attributeKey = SemanticAttributes.EXCEPTION_MESSAGE;
+          } else if (key.equals(Fields.STACK)) {
+            attributeKey = SemanticAttributes.EXCEPTION_STACKTRACE;
+          }
+        }
+        if (!isRecordingException
```
anuraaga, line 251: Let's make the code match the spec closer

`along the rest of the specified key/value pair`

This appears to say that we record exception with the value of ERROR_OBJECT, and everything else ("the rest") is copied as is.

So I think the logic can be added above the complex if statement as just

```java
if (isRecordingException && key.equals(Fields.ERROR_OBJECT) {
  // Already recorded as the exception itself so don't add as attribute.
  continue;
}
```

Then I think this is just `if (attributeKey == null)`

malafeev: updated

opentracing-shim/src/test/java/io/opentelemetry/opentracingshim/SpanShimTest.java
jkwatson: yeah, looks like we don't have any coverage for anything but string values at the moment. Would be really good to add some coverage for numeric and boolean keys as well.

opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java
malafeev: done

opentracing-shim/src/test/java/io/opentelemetry/opentracingshim/SpanShimTest.java
malafeev: added

jkwatson (approved):
Looks good, thanks!

anuraaga (approved):
Thanks!

Issue #3084, opened 2021-03-31
Simplify OpenTracingShim creational patterns


malafeev:
Introduce builder or
instead of 4 leave only two by removing `createTracerShim(OpenTelemetry openTelemetry)` and `createTracerShim()` from `OpenTracingShim` class

jkwatson:
I might rename this to just "simplify OpenTracingShim creational patterns" because we might want to go with a builder.

IrisTuntun:
Hi, can I take this issue?

cc @alolita

anuraaga:
Thanks @IrisTuntun for your interest in this issue! I think we should hold off temporarily until getting some input on this spec issue I filed

https://github.com/open-telemetry/opentelemetry-specification/issues/1608

I am hoping we can reduce to just a single factory `create(OpenTelemetry openTelemetry)` to simplify both the usage but also the concepts.

jack-berg:
Resolved in #5110.

PR #3129, opened 2021-04-07
Add OpenTracingShimBuilder class and test 

IrisTuntun:
Add `OpenTracingShimBuilder` class to simplify OpenTracingShim creational patterns. Future use cases should not depend on `OpenTracingShim` any more.

resolve https://github.com/open-telemetry/opentelemetry-java/issues/3084


opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/OpenTracingShimBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+/** 
+ * Builder for creating an OpenTracing {@link io.opentracing.Tracer} that is implemented using the
+ * OpenTelemetry APIs.
+ */
+public class OpenTracingShimBuilder{
+    private Tracer tracer = getTracer(GlobalOpenTelemetry.getTracerProvider());
+    private OpenTracingPropagators propagators = OpenTracingPropagators.builder().build();
+
+    /** Returns a new {@code OpenTracingShimBuilder} instance. */
+    public static OpenTracingShimBuilder builder(){
```
jkwatson, line 23: Our pattern for this is to have this method live in the `OpenTracingShim` class, rather than here.

opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/OpenTracingShim.java
```diff
@@ -11,6 +11,9 @@
 import io.opentelemetry.api.trace.TracerProvider;
 
 /**
+ * This class is deprecated. Please Use {@link OpenTracingShimBuilder} instead to create an
```
jkwatson, line 14: Hmm. I wonder if it would be good to keep this class, just with the `builder()` method on it, to keep the patterns the same as they are in the rest of the code. It does make this class a bit anemic, but it was already that way. :)

jkwatson:
@IrisTuntun are you planning on picking this up again at some point?

anuraaga:
@jkwatson I had originally suggested holding off on this until the spec issue, which hasn't gotten any input on yet

https://github.com/open-telemetry/opentelemetry-java/issues/3084#issuecomment-815435479

PR #3140, opened 2021-04-08
adding implementation-version to manifest files

### LINKED ISSUE #3032: Make it possible to detect OpenTelemetry API version at runtime
I just had a look at the manifest of 1.0.0 JARs.

API

    Manifest-Version: 1.0
    Automatic-Module-Name: io.opentelemetry.api

SDK

    Manifest-Version: 1.0
    Automatic-Module-Name: io.opentelemetry.sdk

There is no longer any information about the version of those JARs in manifest. Was this removed on purpose?

Previously manifest looked like this:

    Manifest-Version: 1.0
    Implementation-Title: all
    Implementation-Version: 0.14.1
    Built-By: runner
    Built-JDK: 11.0.9
    Source-Compatibility: 1.8
    Target-Compatibility: 1.8
    Automatic-Module-Name: io.opentelemetry.api

SDK contains io/opentelemetry/sdk/version.properties with content:

    sdk.version=1.0.0

But I did not find a good way to detect API version at runtime. 

In our Java agent we want to detect which version of OpenTelemetry API is used by our monitored applications.

kubawach:
Resolves #3032 

Once implementation version is included in manifest files, it can be easily read in the runtime.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -237,7 +237,9 @@ subprojects {
                     inputs.property("moduleName", moduleName)
 
                     manifest {
-                        attributes("Automatic-Module-Name" to moduleName)
+                        attributes(
+                                "Automatic-Module-Name" to moduleName,
+                                "Implementation-Version" to "${project.version}")
                     }
                 }
             }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -120,8 +120,9 @@ subprojects {
                 }
             }
 
+            val releaseVersion = 8
             withType(JavaCompile::class) {
-                options.release.set(8)
+                options.release.set(releaseVersion)
 
                 if (name != "jmhCompileGeneratedClasses") {
                     options.compilerArgs.addAll(listOf(
@@ -239,7 +240,11 @@ subprojects {
                     manifest {
                         attributes(
                                 "Automatic-Module-Name" to moduleName,
-                                "Implementation-Version" to "${project.version}")
+                                "Built-By" to System.getProperty("user.name"),
+                                "Built-JDK" to System.getProperty("java.version"),
+                                "Implementation-Title" to project.name,
+                                "Implementation-Version" to project.version,
+                                "Release-Version" to releaseVersion)
                     }
                 }
             }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -120,9 +120,8 @@ subprojects {
                 }
             }
 
-            val releaseVersion = 8
             withType(JavaCompile::class) {
-                options.release.set(releaseVersion)
+                options.release.set(8)
 
                 if (name != "jmhCompileGeneratedClasses") {
                     options.compilerArgs.addAll(listOf(
@@ -243,8 +242,7 @@ subprojects {
                                 "Built-By" to System.getProperty("user.name"),
                                 "Built-JDK" to System.getProperty("java.version"),
                                 "Implementation-Title" to project.name,
-                                "Implementation-Version" to project.version,
-                                "Release-Version" to releaseVersion)
+                                "Implementation-Version" to project.version)
                     }
                 }
             }


build.gradle.kts
```diff
@@ -237,7 +237,9 @@ subprojects {
                     inputs.property("moduleName", moduleName)
 
                     manifest {
-                        attributes("Automatic-Module-Name" to moduleName)
+                        attributes(
+                                "Automatic-Module-Name" to moduleName,
+                                "Implementation-Version" to "${project.version}")
```
Oberon00, line 242: Can we add all the attributes from https://github.com/open-telemetry/opentelemetry-java/blob/v0.14.1/build.gradle#L151-L156 back? Especially implementation-title.

kubawach: Certainly! I initially wanted to close the gap mentioned in #3032 but rest of the attributes makes sense as well.

build.gradle.kts
```diff
@@ -237,7 +238,13 @@ subprojects {
...
 
                     manifest {
-                        attributes("Automatic-Module-Name" to moduleName)
+                        attributes(
+                                "Automatic-Module-Name" to moduleName,
+                                "Built-By" to System.getProperty("user.name"),
+                                "Built-JDK" to System.getProperty("java.version"),
+                                "Implementation-Title" to project.name,
+                                "Implementation-Version" to project.version,
+                                "Release-Version" to releaseVersion)
```
jkwatson, line 247: I'm unable to find any references to this attribute. Do you have a reference to a doc that describes what this is supposed to represent?

kubawach: releaseVersion? It's set to 8 (constant for the build).

jkwatson: but is this a standard jar-manifest attribute? I don't think we should use manifest attributes that aren't pretty standard.

Oberon00: See also https://docs.oracle.com/javase/9/docs/specs/jar/jar.html#main_attributesmain-attributes

kubawach: Previously (prior to kotlin migration we had: 
```
Source-Compatibility: 1.8
Target-Compatibility: 1.8
```
reflecting actual gradle build settings.
However currently we use javac `--release-version` which takes precedence (overwrites) source/target-compatibility settings (if any). Therefore I decide to replace the old attributes with the new one, reflecting current build settings (to avoid any confusion). 

If we want to stay aligned with the manifest spec, we shall get rid of this completely - but I believe this does bring additional value :)

Oberon00: I don't know. If we invented these on our own, then I doubt any tool will read them. If you can find a reference to this attribute somewhere (specifications, documentation, 3rd party source code) we can keep them, otherwise I don't think it makes any sense.

kubawach: Done :)

Oberon00 (approved):
Thank you!

build.gradle.kts
```diff
@@ -237,7 +237,12 @@ subprojects {
                     inputs.property("moduleName", moduleName)
 
                     manifest {
-                        attributes("Automatic-Module-Name" to moduleName)
+                        attributes(
+                                "Automatic-Module-Name" to moduleName,
+                                "Built-By" to System.getProperty("user.name"),
+                                "Built-JDK" to System.getProperty("java.version"),
+                                "Implementation-Title" to project.name,
```
Oberon00, line 244: Just realized: I think project.name will not be very helpful as it will only be the last part, like `all` or `common`. Something based on `archivesBaseName` would be better IMHO.

kubawach: Hah, same! :) Tried that, however comparing title from old artifacts (pre-kotlin) with what yields name vs archivesBaseName I realised that it was "name" previously, so used it for backwards compatibility. Coupled with "Automatic-Module-Name" it gives full data.

jkwatson (approved):

PR #3142, opened 2021-04-08
Improve key validity code documentation for trace state

jamesylgan:
- [x] Switch from multiple single line comments to multi-line javadoc format
- [x] Add documentation onto the public method so that users will know how to construct valid keys 
- [x] Addressed code style issues raised by CI

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/trace/ArrayBasedTraceStateBuilder.java
+++ b/api/all/src/main/java/io/opentelemetry/api/trace/ArrayBasedTraceStateBuilder.java
@@ -40,6 +40,14 @@ final class ArrayBasedTraceStateBuilder implements TraceStateBuilder {
     this.parent = parent;
   }
 
+   /** 
+   * @param key is opaque string up to 256 characters printable. It MUST begin with a lowercase letter, and
+   * can only contain lowercase letters a-z, digits 0-9, underscores _, dashes -, asterisks *, and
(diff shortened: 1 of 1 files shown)


jkwatson:
Thanks for this doc cleanup! Looks like it still needs some tweaking to get checkstyle to agree.

jamesylgan:
> Thanks for this doc cleanup! Looks like it still needs some tweaking to get checkstyle to agree.

Whoops :)

jamesylgan:
@jkwatson any way the CI checks could be changed to include `actual` as well as `expected`? If this is an external CI tool, then nevermind, but if it's changeable, would help here

`[Task :api:all:checkstyleMain FAILED] [JavadocTagContinuationIndentation] Line continuation have incorrect indentation level, expected level should be 4.`

jkwatson:
> @jkwatson any way the CI checks could be changed to include `actual` as well as `expected`? If this is an external CI tool, then nevermind, but if it's changeable, would help here - it seems to complain with 4 spaces before the asterisk, so not entirely sure what it is looking for
> 
> `[Task :api:all:checkstyleMain FAILED] [JavadocTagContinuationIndentation] Line continuation have incorrect indentation level, expected level should be 4.`

Have you run `./gradlew spotlessApply` ? That should reformat the javadoc to the expected format. I don't think we have any power to change how checkstyle/spotless formats its errors.

jkwatson (approved):
Thanks!

PR #3128, opened 2021-04-07
Modification of enum semantic attributes 

harshita19244:
Resolves https://github.com/open-telemetry/opentelemetry-java/issues/3006

PR #3047, opened 2021-03-19
Update semantic convention constants generator script and template

arminru:
Fixes #3042.

PR #3144, opened 2021-04-09
Semantic Conventions: use only a compatible version of the tools image

jkwatson:
related: #3047 and #3128


anuraaga (approved):

PiratkaXx:
Hi! Just a little question. What kind and whoose is that number we choose? Sorry, I think it looks like stupid.but I feel a little looser in this little counry. As I new born baby)) Not shore I can speak as usual in English as you all. I ll try do it all) I give you my word)

PiratkaXx:
Nice to meet you too) So nice to be a friend. just so little time))) Thanks.

PR #3146, opened 2021-04-12
SDK arch test - ensuring that API implmenetation returns only API types

### LINKED ISSUE #1718: Tooling to prevent returning private type from public method
We currently return a private type, `TracerSdk` from a public method in a public class. This specific issue needs to be fixed, but we should have tooling to prevent it in the first place since it's never a good idea. I've seen broken releases before because unit tests passed but the end user couldn't even use the new API if they tried.

Not sure but I think even the javadoc tool can confirm this and we're probably missing it since we disable javadoc warnings.

kubawach:
Resolves #1718

There are numerous possible approaches to that. I decided to implement a weak guardrails, limiting the scope to known issues only. Basically the goal of the test is to ensure that public SDK methods, that override an API method will always return API type, java type or void. I intentionally decided against simple check if method returned by a SDK method is exactly the same as returned by corresponding API method to allow covariant returns within API scope.

### DIFF
--- a/all/src/test/java/io/opentelemetry/all/SdkDesignTest.java
+++ b/all/src/test/java/io/opentelemetry/all/SdkDesignTest.java
@@ -43,7 +43,7 @@ public class SdkDesignTest {
             .arePublic()
             .and()
             .arePublic()
-            .and(implmentOrOverride())
+            .and(implementOrOverride())
             .should()
             .haveRawReturnType(
                 inPackage("io.opentelemetry.api..", "io.opentelemetry.context..", "java.."))
--- a/all/src/test/java/io/opentelemetry/all/SdkDesignTest.java
+++ b/all/src/test/java/io/opentelemetry/all/SdkDesignTest.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.all;
+
+import com.tngtech.archunit.base.DescribedPredicate;
+import com.tngtech.archunit.base.Optional;
+import com.tngtech.archunit.base.PackageMatcher;
+import com.tngtech.archunit.core.domain.JavaClass;
+import com.tngtech.archunit.core.domain.JavaClassList;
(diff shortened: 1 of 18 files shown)


all/src/test/java/io/opentelemetry/all/SdkDesignTest.java
```diff
@@ -0,0 +1,95 @@
...
+import com.tngtech.archunit.lang.syntax.ArchRuleDefinition;
+import com.tngtech.archunit.lang.syntax.elements.MethodsShouldConjunction;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+public class SdkDesignTest {
+
+  // TODO: limited to trace as currently metrics fails the test (SdkMeterProvider returns SdkMeter)
```
jkwatson, line 25: can we fix this small bit and include metrics as well?

kubawach: Gladly! :)

kubawach:
Updated the code to match requested SDK change.

jkwatson (approved):
Thanks! The meter stuff ended up being a bit bigger than I thought it would. 👍

carlosalberto (approved):

all/src/test/java/io/opentelemetry/all/SdkDesignTest.java
```diff
@@ -0,0 +1,94 @@
...
+            .should()
+            .haveRawReturnType(
+                inPackage("io.opentelemetry.api..", "io.opentelemetry.context..", "java.."))
+            .orShould()
+            .haveRawReturnType("void");
+
+    covariantReturnRule.check(SDK_OTEL_CLASSES);
+  }
+
+  static DescribedPredicate<? super JavaMethod> implmentOrOverride() {
```
anuraaga, line 56:
```suggestion
  static DescribedPredicate<? super JavaMethod> impelmentOrOverride() {
```

anuraaga (approved):
Nice!

PR #3149, opened 2021-04-13
Make the ViewRegistry immutable and remove the mutating deprecated method.

jkwatson:


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/ViewRegistryBuilder.java
```diff
@@ -0,0 +1,49 @@
...
+      new EnumMap<>(InstrumentType.class);
+  private static final LinkedHashMap<Pattern, View> EMPTY_CONFIG = new LinkedHashMap<>();
+
+  ViewRegistryBuilder() {
+    configuration.put(InstrumentType.COUNTER, EMPTY_CONFIG);
+    configuration.put(InstrumentType.UP_DOWN_COUNTER, EMPTY_CONFIG);
+    configuration.put(InstrumentType.VALUE_RECORDER, EMPTY_CONFIG);
+    configuration.put(InstrumentType.SUM_OBSERVER, EMPTY_CONFIG);
+    configuration.put(InstrumentType.UP_DOWN_SUM_OBSERVER, EMPTY_CONFIG);
+    configuration.put(InstrumentType.VALUE_OBSERVER, EMPTY_CONFIG);
```
jack-berg, line 26: Consider:
```
    for (InstrumentType type : InstrumentType.values()) {
      configuration.put(type, EMPTY_CONFIG);
    }
```

jack-berg (approved):
I'm not an approver, but for what its worth this looks good to me!

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/ViewRegistryBuilder.java
jkwatson: good call!

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/ViewRegistryBuilder.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  ViewRegistry build() {
+    return new ViewRegistry(configuration);
+  }
+
+  ViewRegistryBuilder addView(InstrumentSelector selector, View view) {
```
kubawach, line 33: Completely and totally NIT: by convention builder method prefix is "with" - ie "withView".

kubawach (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/ViewRegistryBuilder.java
jkwatson: is it?  Our `SdkTracerProviderBuilder` uses "addSpanProcessor", and this is consistent with that.

kubawach: Typically yes, otherwise I'd not have suggested that. I believe it's usually to have naming separation between POJO (set, add) and builder (with). Additionally most builders are "fluent", and "with" sounds better when chaining methods. 
However there is no strict rule in Java world about builder naming, most things are quite "to personal taste", so as I said - completely and totally NIT :D

jkwatson: we use `set` and `add` in our Builders in this project, so we should stick with that pattern.

kubawach: Definitely consistency is a good thing :)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/ViewRegistry.java
```diff
@@ -8,59 +8,38 @@
...
 final class ViewRegistry {
-  private static final LinkedHashMap<Pattern, View> EMPTY_CONFIG = new LinkedHashMap<>();
   static final View CUMULATIVE_SUM =
       View.builder().setAggregatorFactory(AggregatorFactory.sum(true)).build();
   static final View SUMMARY =
       View.builder().setAggregatorFactory(AggregatorFactory.minMaxSumCount()).build();
   static final View LAST_VALUE =
       View.builder().setAggregatorFactory(AggregatorFactory.lastValue()).build();
 
-  // The lock is used to ensure only one updated to the configuration happens at any moment.
```
anuraaga, line 35: 🎉

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/ViewRegistry.java
```diff
@@ -8,59 +8,38 @@
...
-  ViewRegistry() {
+  ViewRegistry(EnumMap<InstrumentType, LinkedHashMap<Pattern, View>> configuration) {
     this.configuration = new EnumMap<>(InstrumentType.class);
-    configuration.put(InstrumentType.COUNTER, EMPTY_CONFIG);
-    configuration.put(InstrumentType.UP_DOWN_COUNTER, EMPTY_CONFIG);
-    configuration.put(InstrumentType.VALUE_RECORDER, EMPTY_CONFIG);
-    configuration.put(InstrumentType.SUM_OBSERVER, EMPTY_CONFIG);
-    configuration.put(InstrumentType.UP_DOWN_SUM_OBSERVER, EMPTY_CONFIG);
-    configuration.put(InstrumentType.VALUE_OBSERVER, EMPTY_CONFIG);
+    // make a copy for safety
```
anuraaga, line 35: I would generally not do this for private constructors, but no big deal either way

anuraaga (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/ViewRegistry.java
jkwatson: I probably wouldn't normally, either, but since it's a one-time configuration cost to make the defensive copy, this will protect us a bit if somehow this class ends up more public in the future.

PR #3153, opened 2021-04-14
Don't default the tracer name.

jkwatson:
And, make sure that the standard exporters can handle a null InstrumentationLibraryInfo name.

The spec changed in v1.2.0 to say we should no longer have a default: https://github.com/open-telemetry/opentelemetry-specification/pull/1534

### DIFF
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/common/InstrumentationLibraryInfo.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/common/InstrumentationLibraryInfo.java
@@ -5,8 +5,6 @@
 
 package io.opentelemetry.sdk.common;
 
-import static java.util.Objects.requireNonNull;
-
 import com.google.auto.value.AutoValue;
 import io.opentelemetry.api.trace.Tracer;
 import javax.annotation.Nullable;
@@ -30,7 +28,6 @@ public abstract class InstrumentationLibraryInfo {
    * @return the new instance
    */
   public static InstrumentationLibraryInfo create(String name, @Nullable String version) {
-    requireNonNull(name, "name");
     return new AutoValue_InstrumentationLibraryInfo(name, version);
   }
 
@@ -48,6 +45,7 @@ public abstract class InstrumentationLibraryInfo {
    *
    * @return the name of the instrumentation library.
    */
+  @Nullable
   public abstract String getName();
 
   /**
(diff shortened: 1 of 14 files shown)


sdk/common/src/main/java/io/opentelemetry/sdk/common/InstrumentationLibraryInfo.java
```diff
@@ -48,6 +45,7 @@ public static InstrumentationLibraryInfo empty() {
    *
    * @return the name of the instrumentation library.
    */
+  @Nullable
```
anuraaga, line 48: Hmm - going from non-null to Nullable isn't technically a safe change, as can be seen with the amount of null checks we had to add everywhere (any custom exporter out there would have the same problem on a minor version bump). Since the spec says SHOULD preserve original value, but we know preserving null in Java is generally a bad idea, can we just skip this change?

jkwatson: It's a fine question. @bogdandrutu was the one who pushed for this in the specs. Opinions?

Honestly, I also think this is a bad change. I would much prefer to have a fallback name, so that exporters don't all have to guard against this case. 

Can we maybe use empty string as the default, rather than keeping nulls?

anuraaga: Empty string is still a better default than keeping nulls indeed.

kubawach: I'd definitely vote against setting this nullable as the value will propagate, causing lots of extra null-checks (as documented in this PR).

jkwatson: Please your opinions to https://github.com/open-telemetry/opentelemetry-specification/issues/1619  :)

kubawach: Added :)

jkwatson:
Turned this into a draft, pending the outcome of the related spec issue.

Sifat1998 (approved):

jkwatson:
closing, as we're going with #3170

Issue #3157, opened 2021-04-15
why host.name can't be a default attribute in resource


tydhot:
I have found host.name isn't in any default ResourceProvider. Why it can't be?

jkwatson:
I think it could be. Someone would need to write a ResourceProvider for this. Currently, the EC2Resource does provide this, but nothing else does at the moment.

tydhot:
> I think it could be. Someone would need to write a ResourceProvider for this. Currently, the EC2Resource does provide this, but nothing else does at the moment.

Oh, I am willing to write a ResourceProvider to make it be the default attribute. But I am not sure how should the ResourceProvider be named and what other attributes should be added.

iNikem:
`HostResourceProvider`? :)

tydhot:
> `HostResourceProvider`? :)

I think the name is appropriate, I will try to write one :)

PR #3162, opened 2021-04-16
add host.name as default attribute by add HostResourceProvider 

tydhot:
issue #3157

### DIFF
--- a/sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/HostResource.java
+++ b/sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/HostResource.java
@@ -9,32 +9,39 @@ import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.common.AttributesBuilder;
 import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.semconv.resource.attributes.ResourceAttributes;
-
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 
 /** Factory of a {@link Resource} which provides information about the host info. */
 public final class HostResource {
 
-    private static final Resource INSTANCE = buildResource();
+  private static final Resource INSTANCE = buildResource();
 
-    /**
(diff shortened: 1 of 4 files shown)


sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/HostResource.java
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * Returns a factory for a {@link Resource} which provides information about host.
+     */
+    public static Resource get() {
+        return INSTANCE;
+    }
+
+    // Visible for testing
+    static Resource buildResource() {
```
anuraaga, line 29: Do you think we can add the `host.arch` attribute too?

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/host.md#host

sdk-extensions/resources/src/test/java/io/opentelemetry/sdk/extension/resources/HostResourceTest.java
```diff
@@ -0,0 +1,33 @@
...
+        Attributes attributes = HostResource.buildResource().getAttributes();
+
+        // then
+        assertThat(attributes.get(ResourceAttributes.HOST_NAME)).isNotBlank();
+    }
+
+    @Nested
+    @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+    @ExtendWith(SecurityManagerExtension.class)
+    static class SecurityManagerEnabled {
```
anuraaga, line 25: Don't need this test since security manager doesn't affect it I guess.

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/HostResource.java
tydhot: I think is a good idea.I will add it.

anuraaga (approved):
Thanks!

jkwatson (approved):
Thanks!

PR #3168, opened 2021-04-19
Add @SpanAttribute annotation for adding attributes to spans created through instrumentation

HaloFour:
Adds a new annotated that when combined with the `@WithSpan` annotation will add the value of the annotated parameter as an attribute to the newly-created `Span`:

```java
@WithSpan("method")
public void someMethod(@SpanAttribute("name") String someName) {
    // do stuff
}
```

For the above example the OTel auto-instrumentation will create a new `Span` with the name "method" and will take the value of the parameter `someName` and add it as a String attribute with the name "name".

For the attribute names there is the caveat that, by default, the Java compiler does not include the parameter name metadata in the compiled classes.  You can opt-into this by passing the `-parameters` compiler argument.  Without this the instrumentation cannot derive a useful name for the parameter.  This could be an argument to make the `value` element of the annotation not have a default value.

See: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3188

### DIFF
--- a/extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/SpanAttribute.java
+++ b/extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/SpanAttribute.java
@@ -17,7 +17,7 @@ import java.lang.annotation.Target;
  * within the body of the method.
  *
  * <p>Application developers can use this annotation to signal OpenTelemetry auto-instrumentation
- * that a new span should be created whenever marked method is executed.
+ * that a new span attribute should be added to a span created when the parent method is executed.
  *
  * <p>If you are a library developer, then probably you should NOT use this annotation, because it
  * is non-functional without the OpenTelemetry auto-instrumentation agent, or some other annotation
--- a/extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/SpanAttribute.java
+++ b/extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/SpanAttribute.java
@@ -13,6 +13,8 @@ import java.lang.annotation.Target;
 /**
  * This annotation marks that a parameter of a method annotated by the {@link WithSpan} annotation
  * should be added as an attribute to the newly created {@link io.opentelemetry.api.trace.Span}.
+ * Using this annotation is equivalent to calling {@code Span.currentSpan().setAttribute(...)}
+ * within the body of the method.
  *
  * <p>Application developers can use this annotation to signal OpenTelemetry auto-instrumentation
  * that a new span should be created whenever marked method is executed.
--- a/extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/SpanAttribute.java
+++ b/extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/SpanAttribute.java
@@ -21,22 +21,18 @@ import java.lang.annotation.Target;
  * is non-functional without the OpenTelemetry auto-instrumentation agent, or some other annotation
  * processor.
  *
- * @see <a href="https://github.com/open-telemetry/opentelemetry-auto-instr-java">OpenTelemetry
- *     Auto-Instrumentation</a>
+ * @see <a href="https://github.com/open-telemetry/opentelemetry-java-instrumentation">OpenTelemetry
+ *     OpenTelemetry Instrumentation for Java</a>
  */
 @Target(ElementType.PARAMETER)
 @Retention(RetentionPolicy.RUNTIME)
-public @interface WithSpanAttribute {
+public @interface SpanAttribute {
   /**
    * Optional name of the attribute.
    *
-   * <p>If not specified, an appropriate default name should be created by auto-instrumentation.
-   * E.g. the name of the parameter if formal parameter names are stored in the class file on
-   * compilation using the {@code -parameters} compiler option.
-   *
-   * @see <a
-   *     href="https://docs.oracle.com/javase/tutorial/reflect/member/methodparameterreflection.html">Obtaining
-   *     Names of Method Parameters</a>
+   * <p>If not specified and the code is compiled using the `{@code -parameters}` argument to
+   * `javac`, the parameter name will be used instead. If the parameter name is not available, e.g.,
+   * because the code was not compiled with that flag, the attribute will be ignored.
    */
   String value() default "";
 }
--- a/extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/SpanAttribute.java
+++ b/extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/SpanAttribute.java
@@ -25,7 +25,7 @@ import java.lang.annotation.Target;
  *
  * @see <a href="https://github.com/open-telemetry/opentelemetry-java-instrumentation">OpenTelemetry
  *     OpenTelemetry Instrumentation for Java</a>
- * @since 1.2.0
+ * @since 1.4.0
  */
 @Target(ElementType.PARAMETER)
 @Retention(RetentionPolicy.RUNTIME)
--- a/extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/SpanAttribute.java
+++ b/extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/SpanAttribute.java
@@ -25,6 +25,7 @@ import java.lang.annotation.Target;
  *
  * @see <a href="https://github.com/open-telemetry/opentelemetry-java-instrumentation">OpenTelemetry
  *     OpenTelemetry Instrumentation for Java</a>
+ * @since 1.2.0
  */
 @Target(ElementType.PARAMETER)
 @Retention(RetentionPolicy.RUNTIME)
--- a/extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/WithSpanAttribute.java
+++ b/extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/WithSpanAttribute.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.extension.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This annotation marks that a parameter of a method annotated by the {@link WithSpan} annotation
+ * should be added as an attribute to the newly created {@link io.opentelemetry.api.trace.Span}.
+ *
+ * <p>Application developers can use this annotation to signal OpenTelemetry auto-instrumentation
+ * that a new span should be created whenever marked method is executed.
+ *
+ * <p>If you are a library developer, then probably you should NOT use this annotation, because it
+ * is non-functional without the OpenTelemetry auto-instrumentation agent, or some other annotation
+ * processor.
+ *
+ * @see <a href="https://github.com/open-telemetry/opentelemetry-auto-instr-java">OpenTelemetry
+ *     Auto-Instrumentation</a>
+ */
+@Target(ElementType.PARAMETER)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface WithSpanAttribute {
+  /**
+   * Optional name of the attribute.
+   *
+   * <p>If not specified, an appropriate default name should be created by auto-instrumentation.
+   * E.g. {@code "className"."method"}
+   */
+  String value() default "";
+}
--- a/extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/WithSpanAttribute.java
+++ b/extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/WithSpanAttribute.java
@@ -31,7 +31,12 @@ public @interface WithSpanAttribute {
    * Optional name of the attribute.
    *
    * <p>If not specified, an appropriate default name should be created by auto-instrumentation.
-   * E.g. {@code "className"."method"}
+   * E.g. the name of the parameter if formal parameter names are stored in the class file on
+   * compilation using the {@code -parameters} compiler option.
+   *
+   * @see <a
+   *     href="https://docs.oracle.com/javase/tutorial/reflect/member/methodparameterreflection.html">Obtaining
+   *     Names of Method Parameters</a>
    */
   String value() default "";
 }
--- a/extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/WithSpan.java
+++ b/extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/WithSpan.java
@@ -22,8 +22,8 @@ import java.lang.annotation.Target;
  * is non-functional without the OpenTelemetry auto-instrumentation agent, or some other annotation
  * processor.
  *
- * @see <a href="https://github.com/open-telemetry/opentelemetry-auto-instr-java">OpenTelemetry
- *     Auto-Instrumentation</a>
+ * @see <a href="https://github.com/open-telemetry/opentelemetry-java-instrumentation">OpenTelemetry
+ *     OpenTelemetry Instrumentation for Java</a>
  */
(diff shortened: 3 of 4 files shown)


extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/WithSpanAttribute.java
```diff
@@ -0,0 +1,37 @@
...
+ * <p>If you are a library developer, then probably you should NOT use this annotation, because it
+ * is non-functional without the OpenTelemetry auto-instrumentation agent, or some other annotation
+ * processor.
+ *
+ * @see <a href="https://github.com/open-telemetry/opentelemetry-auto-instr-java">OpenTelemetry
+ *     Auto-Instrumentation</a>
+ */
+@Target(ElementType.PARAMETER)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface WithSpanAttribute {
```
Oberon00, line 29: Some more bikeshedding suggestions: `CaptureAsSpanAttribute`, `AsSpanAttribute`, `AddToSpan`

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/WithSpanAttribute.java
```diff
@@ -0,0 +1,37 @@
...
+  /**
+   * Optional name of the attribute.
+   *
+   * <p>If not specified, an appropriate default name should be created by auto-instrumentation.
```
Oberon00, line 33: Since this can be used on multiple parameters, wouldn't that be more `classname.method.argname`? Or should this be a required parameter?

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/WithSpanAttribute.java
```diff
@@ -0,0 +1,37 @@
...
+   * E.g. {@code "className"."method"}
```
Oberon00, line 34:
```suggestion
   * E.g. {@code className.method}
```

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/WithSpanAttribute.java
HaloFour: oops, that is total copypasta on my part.  I would think that the parameter name itself would be sufficient, although note the caveat mentioned in above that this would require that the code be compiled with `-parameters` which should definitely be mentioned in this javadoc comment.

jkwatson:
Why is this a better approach than just calling `Span.current().setAttribute(...)` ? The advantage of `@WithSpan` is that it hides a bunch of boilerplate try-with-resources/finally code. This, in my opinion, just adds some hard-to-read annotation gunk, which clutters up method signatures, rather than providing any real value.

jkwatson:
> Why is this a better approach than just calling `Span.current().setAttribute(...)` ? The advantage of `@WithSpan` is that it hides a bunch of boilerplate try-with-resources/finally code. This, in my opinion, just adds some hard-to-read annotation gunk, which clutters up method signatures, rather than providing any real value.

Just adding to this, if a user chose to use these annotations, they would be requiring that the implementation do the non-optimal thing of creating a new instance of the AttributeKey for every call, rather than allocating them up front (or, even better, using the semantic convention attributes).

HaloFour:
@jkwatson 

> Why is this a better approach than just calling `Span.current().setAttribute(...)` ?

It's a minor point but the instrumentation can choose to not create a span and that would have to be something that the code within the method then attempts to detect.  But that is probably a fairly rare circumstance.  I think the only time it comes up now is if you try to create a `SERVER` or `CLIENT` span when already in the context of another span of the same kind.

> Just adding to this, if a user chose to use these annotations, they would be requiring that the implementation do the non-optimal thing of creating a new instance of the AttributeKey for every call, rather than allocating them up front (or, even better, using the semantic convention attributes).

This is a good point.  The implementation I have in my own aspect memoizes the "bindings" to the method parameters so the `AttributeKey<?>` instances are created only once per method.  I've not gotten that far with the draft PR to OTel Instrumentation.  But you couldn't use the shared semconv attribute key instances.  I don't know that this is a terribly big deal as the cases where I've used this kind of functionality has been for more custom attributes that I wouldn't expect to live in the semconv module.

As for the general usefulness of this feature, I leave that to the group to decide.  There seemed to be some interest when I posted a [discussion](https://github.com/open-telemetry/opentelemetry-java-instrumentation/discussions/2037) that mentioned it.  But it is admittedly a pretty minor feature.

Oberon00:
Another advantage of annotations is that you only need a `compileOnly` dependency on them. So if you don't need the otel-api otherwise, you have one dependency fewer.

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/WithSpanAttribute.java
jkwatson: If we go with this addition, I'd just have it be `@SpanAttribute`. It's going to be super noisy in the source code already, so we should strive to make it as concise as is reasonable.

anuraaga: +1 to `SpanAttribute`

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/WithSpanAttribute.java
anuraaga: We should phrase our javadoc for the end user, not instrumentation authors, in general I think (if notes for instrumentation are appropriate, I'd rather have them in a .md file than javadoc). So this would be something like

```
If not specified and the code is compiled using the `{@code -parameters}` argument to `javac`, the parameter name will be used instead. If the parameter name is not available, e.g., because the code was not compiled with that flag, the attribute will be ignored.
```

anuraaga (commented):
Now I've posted on the linked discussion, and the Slack, and here - hope things didn't fragment too much :)

As I posted in the discussion, I find supporting this sort of annotation idiomatic vs frameworks like spring. I think the performance issue is a non-issue since as posted in slack, there are cardinality guarantees that mean we should just cache the attribute keys (with bytecode manipulation, technically it doesn't even need to be a Map, the keys could be added to a static initializer and referenced directly. We have no tooling for such coolness though and would probably just use a Map at first).

> Another advantage of annotations is that you only need a compileOnly dependency on them. So if you don't need the otel-api otherwise, you have one dependency fewer.

In user apps this may commonly be the case, but let's not forget our own troubles with `compileOnly` and `GuardedBy` :) In principle, we should assume anything that's not `SOURCE` retention does require a proper dependency.

jkwatson:
I don't love this, because I feel that annotation-oriented-programming is not a great approach to writing readable code, but if the other maintainers are on board, I won't stand in the way.

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/WithSpanAttribute.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * <p>Application developers can use this annotation to signal OpenTelemetry auto-instrumentation
+ * that a new span should be created whenever marked method is executed.
+ *
+ * <p>If you are a library developer, then probably you should NOT use this annotation, because it
+ * is non-functional without the OpenTelemetry auto-instrumentation agent, or some other annotation
+ * processor.
+ *
+ * @see <a href="https://github.com/open-telemetry/opentelemetry-auto-instr-java">OpenTelemetry
+ *     Auto-Instrumentation</a>
```
mateuszrzeszutek, line 25: We shouldn't use the old java instrumentation repo URL:

```suggestion
 * @see <a href="https://github.com/open-telemetry/opentelemetry-java-instrumentation">OpenTelemetry
 *     OpenTelemetry Instrumentation for Java</a>
```

trask:
I think this is a nice option for some uses. I think it's worth putting the alternative `Span.current().setAttribute(...)` into the `@SpanAttribute` javadoc, since it's not uncommon to want to pull out a nested field from a parameter. Also that should prevent people from getting the idea that this is the only want to capture span attributes, and then adding method parameters just so that they can be captured with `@SpanAttribute` 😄.

I'm not sure about making the attribute name optional, it seems too easy to break something, e.g. by compiling the class differently, or by refactoring and renaming the parameter. Are there examples of other annotations that do this (e.g. JAX-RS, etc)?

HaloFour:
@trask 

> I'm not sure about making the attribute name optional, it seems too easy to break something, e.g. by compiling the class differently, or by refactoring and renaming the parameter. Are there examples of other annotations that do this (e.g. JAX-RS, etc)?

The Spring parameter mapping annotations (`@PathVariable`, `@RequestParam`) have optional names and will infer the name from the parameter.

AFAICT the JAXRS annotations all require the name.

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/WithSpanAttribute.java
HaloFour: Oops, also a copypasta.  I'll update the link in the javadoc for `@WithSpan` while I'm at it.

kubawach:
> @trask
> 
> > I'm not sure about making the attribute name optional, it seems too easy to break something, e.g. by compiling the class differently, or by refactoring and renaming the parameter. Are there examples of other annotations that do this (e.g. JAX-RS, etc)?
> 
> The Spring parameter mapping annotations (`@PathVariable`, `@RequestParam`) have optional names and will infer the name from the parameter.
> 
> AFAICT the JAXRS annotations all require the name.

I think that optional names strike a good compromise between necessity to have a constant attribute name (cause eg automation relies on it) and moving fast with the implementation on the customer side (names of attributes same as parameter - less friction).

kubawach:
From my point of view this is an interesting feature, worth implementing. Coming from enterprise software I often saw that annotation driven approach to implementing aspects of an application not directly related to the core business functionality (such as logging, telemetry, security...) is very welcomed. Therefore this kind of feature can increase / speed-up adoption of OTEL :)

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/SpanAttribute.java
```diff
@@ -0,0 +1,40 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This annotation marks that a parameter of a method annotated by the {@link WithSpan} annotation
+ * should be added as an attribute to the newly created {@link io.opentelemetry.api.trace.Span}.
+ * Using this annotation is equivalent to calling {@code Span.currentSpan().setAttribute(...)}
+ * within the body of the method.
+ *
+ * <p>Application developers can use this annotation to signal OpenTelemetry auto-instrumentation
```
kubawach, line 19: This seems a copy-paste :) Probably should read: 
```
* <p>Application developers can use this annotation to signal OpenTelemetry auto-instrumentation
*  that a new span attribute should be added to a span created when the parent method is executed.
```

kubawach (changes requested):

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/SpanAttribute.java
HaloFour: Good catch, indeed a copypasta fail!

anuraaga (approved):
We discussed at SIG today and decided on optional params. The main motivation is that we found Spring has many mechanisms for resolving names, including a free one for kotlin users. We're hoping you'll help us implement these :)

HaloFour:
Sounds good.  I'm going to try to put some effort into the instrumentation side of things this weekend.  Based on the conversations here I want to ensure that the implementation keeps the cost down as much as possible, e.g. ensuring that the `AttributeKey<T>` is allocated exactly one.

I think there's a conversation around how flexible the instrumentation should attempt to be when coercing a parameter type to an attribute where it would require allocations or copying.  For example, to me it feels reasonable to support parameters of type `long[]`, but there are no free ways to get that into a `List<Long>` compatible with the OTel API.

trask (approved):

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/SpanAttribute.java
```diff
@@ -0,0 +1,40 @@
...
+ * <p>If you are a library developer, then probably you should NOT use this annotation, because it
+ * is non-functional without the OpenTelemetry auto-instrumentation agent, or some other annotation
+ * processor.
+ *
+ * @see <a href="https://github.com/open-telemetry/opentelemetry-java-instrumentation">OpenTelemetry
+ *     OpenTelemetry Instrumentation for Java</a>
+ */
+@Target(ElementType.PARAMETER)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface SpanAttribute {
```
jkwatson, line 31: We need a `@since 1.2.0` on this class javadoc as well, if we're going to add it.

jkwatson:
@bogdandrutu since this is effectively an API addition, if only in the annotation package, your opinion here matters as well.

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/SpanAttribute.java
anuraaga: @HaloFour Once you add this, we'll merge :)

HaloFour: Sorry about that, been a bit swamped recently.  Speaking of, with this merged is there an expected turnaround time for instrumentation to match this annotation?  I've been working on some changes to my WIP PR to try to address some of the concerns brought up here but it has some more to go and I've not had a lot of opportunity to look at it this week.

trask: > Speaking of, with this merged is there an expected turnaround time for instrumentation to match this annotation?

that's a good thought. I hadn't thought about it, but it would be ideal for this annotation to land in the same release cycle as the support for it in the java agent to avoid user confusion (this repo usually releases first week of the month, and instrumentation repo usually releases the week after that).

makes sense to hold off on merging this PR for now then?

jkwatson: I'm happy to hold off until after the release. Of course, I'd be happy to wait even longer. ;)

HaloFour: @trask 

> makes sense to hold off on merging this PR for now then?

That works for me.

@jkwatson 

>  Of course, I'd be happy to wait even longer. ;)

You should see the projects I've been working on lately.  Spring WebFlux + Swagger/OpenAPI.  I think we have twice as many annotations as we have code.  😁

jkwatson: Swagger is the worst. It's the example I always cite about annotation abuse.

jkwatson:
@HaloFour I think this needs a rebase to pick up the changes to our CI settings to do a build on 20.04.

HaloFour:
Sorry about the holdup, life got in the way.  I've rebased this PR and I've also pushed up a draft PR of the implementation with the ByteBuddy instrumentation attempting to simplify the approach and take into consideration some of the concerns brought up here and on Slack:  https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3188

jkwatson:
Since we're planning on a release next week, should we hold off until just after that release, so the instrumentation project has a chance to integrate?

HaloFour:
@jkwatson 

> Since we're planning on a release next week, should we hold off until just after that release, so the instrumentation project has a chance to integrate?

Agreed.  I'd want to ensure that the instrumentation PR gets plenty of careful consideration and while I've been making progress I've only recently marked it ready for review.  I'd be happy to wait until that PR has been approved (or as close as it can be without this annotation being ready).

jkwatson (approved):
sigh

anuraaga (approved):

anuraaga:
@HaloFour Don't mind @jkwatson :P Thanks!

PR #3170, opened 2021-04-20
Default the tracer name to an empty string

jkwatson:
An alternative to #3153, assuming that the spec gets updated to default to an empty string, rather than null.


anuraaga (approved):

kubawach:
Seems empty string has been approved (https://github.com/open-telemetry/opentelemetry-specification/issues/1619) but not added to spec yet.

kubawach (approved):

jkwatson:
ok, looks like this is now good to go. I'll close #3153

carlosalberto (approved):

PR #3172, opened 2021-04-21
update the jaeger-client dependency to the latest

jkwatson:


exporters/jaeger-thrift/src/main/java/io/opentelemetry/exporter/jaeger/thrift/JaegerThriftSpanExporterBuilder.java
```diff
@@ -46,7 +47,11 @@ public JaegerThriftSpanExporterBuilder setEndpoint(String endpoint) {
    */
   public JaegerThriftSpanExporter build() {
     if (thriftSender == null) {
-      thriftSender = new HttpSender.Builder(endpoint).build();
+      try {
+        thriftSender = new HttpSender.Builder(endpoint).build();
+      } catch (TTransportException e) {
```
jkwatson, line 52: this was an unfortunate breaking change in the library's APIs. :(

anuraaga (approved):
Any reason to only update this one? Though it is nice that we got the code change done outside of the normal update-all-dependencies PR which might have been the reason.

jkwatson:
> Any reason to only update this one? Though it is nice that we got the code change done outside of the normal update-all-dependencies PR which might have been the reason.

Only that I saw that jaeger was updated for a security vulnerability in lib-thrift and it might be good to make sure we were upgraded, even though the vulnerability only impacted the server side of jaeger. We should surely upgrade other deps as well.

PR #3175, opened 2021-04-23
Fetch Jaeger configuration sampling when using Jaeger Collector

bbarin:


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java
```diff
@@ -152,6 +154,20 @@ static Sampler configureSampler(String sampler, ConfigProperties config) {
           }
           return Sampler.parentBased(Sampler.traceIdRatioBased(ratio));
         }
+      case "jaeger":
```
jkwatson, line 157: Is this an official option in the opentelemetry specification? I don't see it here: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#general-sdk-configuration

jkwatson:
Hi @bbarin . I think this is a good addition, but before we can accept it, we need 2 things.

1) You will need to sign the CNCF CLA. You should see the information in this pull request.
2) This will need to be an official part of the OpenTelemetry specification. I don't see this as a standard configuration option here:  https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#general-sdk-configuration . If you could work with the spec process to get this added, that will expedite getting this feature added.

Thanks!

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java
bbarin: It's not. The proposal is to create a new option `jaeger` to be able to fetch the sampling configuration from the Jaeger Collector. Is there a problem to create a new option?

jkwatson: We would want to have it in the official specification before implementing in java specifically, as we want to make sure that all languages use the same options, formats, names, etc.

bbarin: Makes sense. I will close this PR

PR #3176, opened 2021-04-23
Auto-detect Android and use an Android-friendly IdGenerator instance.

jkwatson:


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/AndroidFriendlyRandomIdGenerator.java
```diff
@@ -0,0 +1,52 @@
...
+    long idLo;
+    Random random = getRandom();
+    do {
+      idHi = random.nextLong();
+      idLo = random.nextLong();
+    } while (idHi == INVALID_ID && idLo == INVALID_ID);
+    return TraceId.fromLongs(idHi, idLo);
+  }
+
+  private static Random getRandom() {
```
kubawach, line 44: This can be replaced with ThreadLocal.initialValue() (https://docs.oracle.com/javase/7/docs/api/java/lang/ThreadLocal.html#initialValue()) which was intended for this kind of initialisation.

kubawach (approved):

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/AndroidFriendlyRandomIdGenerator.java
jkwatson: I knew there was an API for this...I looked and couldn't find it, but obviously I missed what was staring me in the face. BTW, there's an even easier way now:

`ThreadLocal.withInitial(Random::new)`

jkwatson: Of course, that can't be used on Android API 21. 🤦🏼

kubawach: Nice, love the supplier version. Easy way to make this whole class generic, ie allowing supplier to be injected via constructor. However now probably an overkill due to lack of useful use-cases for that.

jkwatson: I did consider that as well, but this is a small enough amount of code that I think it's fine. Since this is an internal API, we can always change it in the future if we need a more flexible id generator setup.

jkwatson: Update to use the `initialValue()` override.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/AndroidFriendlyRandomIdGenerator.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * {@link IdGenerator} instance that doesn't use {@link java.util.concurrent.ThreadLocalRandom},
+ * which is broken on most versions of Android (it uses the same seed everytime it starts up).
+ */
+enum AndroidFriendlyRandomIdGenerator implements IdGenerator {
+  INSTANCE;
+
+  // note: we can't use ThreadLocal.withInitial(..) on Android API 21
+  private static final ThreadLocal<Random> threadLocalRandom =
```
anuraaga, line 20: Maybe better to avoid the thread local and just go with a static final, I don't think we'd expect such contention for it to win over normal access on Android

anuraaga (approved):

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/AndroidFriendlyRandomIdGenerator.java
jkwatson: Hm. Interesting thought. You're probably right in normal use-cases. I wish there were a good way to measure that outside an actual Android VM.

jkwatson: I'll at least test this out locally on an Android app just to see if there is anything obviously bad about just using a static final.

jkwatson: seems to work fine in my limited testing; I'll change to the simpler implementation. Thanks for the suggestion!

anuraaga (approved):

PR #3180, opened 2021-04-27
baggage parsing improvements

kubawach:
Initially it was though to simply resolve parser's TODO: single pass. 
However, after closer look ended up implementing parser for key and value in accordance with W3C spec (https://w3c.github.io/baggage/) and RFC (https://www.ietf.org/rfc/rfc2616.txt, 2.2).
Added also a bunch of unit tests to cover various corner cases.


jkwatson:
Do you have benchmarks that show this is better than what we currently have? I'm pretty hesitant to add 500 lines of parsing code that we have to maintain without strong evidence that it is significantly better than the current implementation.

kubawach:
> Do you have benchmarks that show this is better than what we currently have? I'm pretty hesitant to add 500 lines of parsing code that we have to maintain without strong evidence that it is significantly better than the current implementation.

The TODO comment asked for a single pass implementation (instead of multiple split / trim operations). Intuitively this one should be much faster (no trimming for key / val, single pass). Yet you're right, it would be good to provide some strong evidence :)

kubawach:
> Do you have benchmarks that show this is better than what we currently have? I'm pretty hesitant to add 500 lines of parsing code that we have to maintain without strong evidence that it is significantly better than the current implementation.

OK, so I run pretty basic JMH benchmark:
* 100 elements (key, value, metadata) in a header, pre-allocated, no whitespaces 
* 5 iterations warmup, 15 iterations test

Results are as follows:
```
Result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.oldPropagatorExtractBenchmark":
  19195.220 ±(99.9%) 2818.884 ns/op [Average]
  (min, avg, max) = (15703.962, 19195.220, 24009.204), stdev = 2636.786
  CI (99.9%): [16376.336, 22014.104] (assumes normal distribution)

Secondary result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.oldPropagatorExtractBenchmark:·gc.alloc.rate":
  2023.475 ±(99.9%) 294.097 MB/sec [Average]
  (min, avg, max) = (1588.065, 2023.475, 2434.695), stdev = 275.099
  CI (99.9%): [1729.377, 2317.572] (assumes normal distribution)

Secondary result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.oldPropagatorExtractBenchmark:·gc.time":
  65.000 ±(99.9%) 0.001 ms [Sum]
  (min, avg, max) = (2.000, 4.333, 6.000), stdev = 0.976
  CI (99.9%): [65.000, 65.000] (assumes normal distribution)

  ---

  Result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.propagatorExtractBenchmark":
  8985.815 ±(99.9%) 659.728 ns/op [Average]
  (min, avg, max) = (8476.136, 8985.815, 10685.210), stdev = 617.110
  CI (99.9%): [8326.087, 9645.544] (assumes normal distribution)

Secondary result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.propagatorExtractBenchmark:·gc.alloc.rate":
  545.161 ±(99.9%) 36.037 MB/sec [Average]
  (min, avg, max) = (457.316, 545.161, 575.719), stdev = 33.709
  CI (99.9%): [509.124, 581.199] (assumes normal distribution)

Secondary result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.propagatorExtractBenchmark:·gc.time":
  28.000 ±(99.9%) 0.001 ms [Sum]
  (min, avg, max) = (1.000, 1.867, 5.000), stdev = 1.187
  CI (99.9%): [28.000, 28.000] (assumes normal distribution)
```

Numbers look pretty convincing, however, as stated, the test was pretty basic. I can run more (with whitespaces, with errors) if needed. Let me know :)

jkwatson:
Can you also compare memory allocations between the two implementations, for completeness?

kubawach:
@jkwatson decided to include the benchmark in the codebase, along with the "jdk string based" one, so we can easily compare and also see what are we comparing :) 
By the numbers, memory also looks solid:

```
Result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.defaultPropagatorExtractBenchmark":
  36732.359 ±(99.9%) 2436.325 ns/op [Average]
  (min, avg, max) = (32732.662, 36732.359, 39531.125), stdev = 2278.940
  CI (99.9%): [34296.034, 39168.684] (assumes normal distribution)

Secondary result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.defaultPropagatorExtractBenchmark:·gc.alloc.rate":
  628.739 ±(99.9%) 42.770 MB/sec [Average]
  (min, avg, max) = (581.688, 628.739, 702.670), stdev = 40.007
  CI (99.9%): [585.969, 671.509] (assumes normal distribution)

Secondary result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.defaultPropagatorExtractBenchmark:·gc.alloc.rate.norm":
  36232.040 ±(99.9%) 0.103 B/op [Average]
  (min, avg, max) = (36232.013, 36232.040, 36232.390), stdev = 0.097
  CI (99.9%): [36231.937, 36232.144] (assumes normal distribution)

Secondary result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.defaultPropagatorExtractBenchmark:·gc.churn.G1_Eden_Space":
  632.431 ±(99.9%) 75.278 MB/sec [Average]
  (min, avg, max) = (604.334, 632.431, 806.477), stdev = 70.415
  CI (99.9%): [557.153, 707.709] (assumes normal distribution)

Secondary result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.defaultPropagatorExtractBenchmark:·gc.churn.G1_Eden_Space.norm":
  36487.220 ±(99.9%) 3889.119 B/op [Average]
  (min, avg, max) = (31401.501, 36487.220, 46149.640), stdev = 3637.884
  CI (99.9%): [32598.101, 40376.339] (assumes normal distribution)

Secondary result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.defaultPropagatorExtractBenchmark:·gc.churn.G1_Old_Gen":
  0.006 ±(99.9%) 0.013 MB/sec [Average]
  (min, avg, max) = (≈ 0, 0.006, 0.047), stdev = 0.012
  CI (99.9%): [≈ 0, 0.019] (assumes normal distribution)

Secondary result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.defaultPropagatorExtractBenchmark:·gc.churn.G1_Old_Gen.norm":
  0.372 ±(99.9%) 0.775 B/op [Average]
  (min, avg, max) = (≈ 0, 0.372, 2.880), stdev = 0.725
  CI (99.9%): [≈ 0, 1.147] (assumes normal distribution)

Secondary result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.defaultPropagatorExtractBenchmark:·gc.count":
  47.000 ±(99.9%) 0.001 counts [Sum]
  (min, avg, max) = (3.000, 3.133, 4.000), stdev = 0.352
  CI (99.9%): [47.000, 47.000] (assumes normal distribution)

Secondary result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.defaultPropagatorExtractBenchmark:·gc.time":
  21.000 ±(99.9%) 0.001 ms [Sum]
  (min, avg, max) = (1.000, 1.400, 2.000), stdev = 0.507
  CI (99.9%): [21.000, 21.000] (assumes normal distribution)


  ---

  Result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.jdkStringOpsBasedPropagatorExtractBenchmark":
  50519.369 ±(99.9%) 2961.672 ns/op [Average]
  (min, avg, max) = (48066.130, 50519.369, 57574.272), stdev = 2770.350
  CI (99.9%): [47557.697, 53481.041] (assumes normal distribution)

Secondary result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.jdkStringOpsBasedPropagatorExtractBenchmark:·gc.alloc.rate":
  1183.830 ±(99.9%) 65.079 MB/sec [Average]
  (min, avg, max) = (1035.209, 1183.830, 1239.366), stdev = 60.875
  CI (99.9%): [1118.750, 1248.909] (assumes normal distribution)

Secondary result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.jdkStringOpsBasedPropagatorExtractBenchmark:·gc.alloc.rate.norm":
  93943.930 ±(99.9%) 0.589 B/op [Average]
  (min, avg, max) = (93942.029, 93943.930, 93944.654), stdev = 0.551
  CI (99.9%): [93943.341, 93944.519] (assumes normal distribution)

Secondary result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.jdkStringOpsBasedPropagatorExtractBenchmark:·gc.churn.G1_Eden_Space":
  1183.900 ±(99.9%) 76.463 MB/sec [Average]
  (min, avg, max) = (1007.710, 1183.900, 1213.103), stdev = 71.524
  CI (99.9%): [1107.437, 1260.363] (assumes normal distribution)

Secondary result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.jdkStringOpsBasedPropagatorExtractBenchmark:·gc.churn.G1_Eden_Space.norm":
  93968.744 ±(99.9%) 4126.131 B/op [Average]
  (min, avg, max) = (85902.529, 93968.744, 102480.985), stdev = 3859.585
  CI (99.9%): [89842.613, 98094.874] (assumes normal distribution)

Secondary result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.jdkStringOpsBasedPropagatorExtractBenchmark:·gc.churn.G1_Old_Gen":
  0.027 ±(99.9%) 0.013 MB/sec [Average]
  (min, avg, max) = (0.001, 0.027, 0.051), stdev = 0.012
  CI (99.9%): [0.014, 0.040] (assumes normal distribution)

Secondary result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.jdkStringOpsBasedPropagatorExtractBenchmark:·gc.churn.G1_Old_Gen.norm":
  2.145 ±(99.9%) 1.013 B/op [Average]
  (min, avg, max) = (0.097, 2.145, 3.947), stdev = 0.947
  CI (99.9%): [1.132, 3.158] (assumes normal distribution)

Secondary result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.jdkStringOpsBasedPropagatorExtractBenchmark:·gc.count":
  88.000 ±(99.9%) 0.001 counts [Sum]
  (min, avg, max) = (5.000, 5.867, 6.000), stdev = 0.352
  CI (99.9%): [88.000, 88.000] (assumes normal distribution)

Secondary result "io.opentelemetry.api.baggage.propagation.W3CBaggagePropagatorBenchmark.jdkStringOpsBasedPropagatorExtractBenchmark:·gc.time":
  44.000 ±(99.9%) 0.001 ms [Sum]
  (min, avg, max) = (2.000, 2.933, 4.000), stdev = 0.458
  CI (99.9%): [44.000, 44.000] (assumes normal distribution)
```

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/Element.java
```diff
@@ -0,0 +1,100 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.api.baggage.propagation;
+
+import java.util.BitSet;
+
+class Element {
```
jkwatson, line 10: even though this isn't public, it would be very helpful to have at least some description about what this class is about.

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/Element.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  private boolean tryNextTokenChar(int i) {
+    if (leadingSpace) {
+      markStart(i);
+    }
+    return !trailingSpace;
+  }
+
+  void markStart(int start) {
```
jkwatson, line 88: I think this can be private

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/Parser.java
```diff
@@ -0,0 +1,145 @@
...
+
+  private boolean skipToNext;
+
+  public Parser(String baggageHeader) {
+    this.baggageHeader = baggageHeader;
+    reset(0);
+  }
+
+  public void reset(int i) {
+
```
jkwatson, line 48: there's some odd extra whitespace in here.  Also, please go through this class and make everything private that can be private.

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/Parser.java
```diff
@@ -0,0 +1,145 @@
...
+  private int metaStart;
+
+  private boolean skipToNext;
+
+  public Parser(String baggageHeader) {
+    this.baggageHeader = baggageHeader;
+    reset(0);
+  }
+
+  public void reset(int i) {
```
jkwatson, line 47: needs a better parameter name than 'i', and maybe a bit of doc about what this method does.

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/Parser.java
```diff
@@ -0,0 +1,145 @@
...
+    this.meta = "";
+    this.metaStart = 0;
+  }
+
+  public void setState(State state, int start) {
+    this.state = state;
+    this.metaStart = start;
+  }
+
+  public void parseInto(BaggageBuilder baggageBuilder) {
```
jkwatson, line 64: I think this method should be up right below the constructor, to make it clear that it's the useful API for this class.

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/Element.java
```diff
@@ -0,0 +1,100 @@
...
+    this.end = end;
+    this.readingValue = false;
+    trailingSpace = true;
+  }
+
+  private void setValue(String header) {
+    this.value = header.substring(this.start, this.end);
+  }
+
+  boolean tryNextChar(char character, int i) {
```
jkwatson, line 63: this method is always called with a `!` in front of it. It might be better to invert the logic and rename the method appropriately.

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/Element.java
```diff
@@ -0,0 +1,100 @@
...
+
+  private boolean leadingSpace;
+  private boolean readingValue;
+  private boolean trailingSpace;
+  private int start;
+  private int end;
+  private String value;
+
+  Element(char[] excludedChars) {
+    for (char separator : excludedChars) {
```
jkwatson, line 22: odd calling this loop variable `separator`. It's not always a separator is it?

jkwatson:
This seems like a reasonable approach. Needs a little cleanup, but it seems to do the job. Thanks!

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/Element.java
kubawach: Will do.

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/Element.java
kubawach: Initially I did exactly as you suggest, but then the method returns "false" if it completes properly (so "try" is successful).  I found this highly misleading, so changed to what we have now. All of the usages read now: "if try is unsuccessful, then...". Makes sense?

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/Parser.java
kubawach: Will do.

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/Parser.java
kubawach: Will do.

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/Parser.java
kubawach: Will do.

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/Element.java
kubawach: Will do.

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/Element.java
kubawach: Right, forgot to change this after refactoring.

kubawach:
@jkwatson all resolved, please have a look :)

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/Element.java
jkwatson: you could rename it to something like "rejectNextChar"?

api/all/src/jmh/java/io/opentelemetry/api/baggage/propagation/JdkStringOperationsBasedPropagator.java
```diff
@@ -0,0 +1,108 @@
...
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.propagation.TextMapGetter;
+import io.opentelemetry.context.propagation.TextMapPropagator;
+import io.opentelemetry.context.propagation.TextMapSetter;
+import java.util.Collection;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/** JDK string-ops based W3C baggage propagator. For comparison purposes. */
+class JdkStringOperationsBasedPropagator implements TextMapPropagator {
```
jkwatson, line 22: This should be deleted before this PR will be ready to merge. There's no reason to keep it around if we're going with the new approach.

kubawach: I was thinking about leaving it just as a "reference" implementation that could be compared against in case of a newer implementation. Yet if you say it should be deleted, it will be :)

jkwatson: yeah, we don't want to leave dead code around. :)

jkwatson (approved):
thanks! @anuraaga are you interested in taking a look before we merge?

kubawach:
@anuraaga if convenient please have a look :)

carlosalberto (approved):

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/Parser.java
```diff
@@ -0,0 +1,148 @@
...
+
+  private boolean skipToNext;
+
+  public Parser(String baggageHeader) {
+    this.baggageHeader = baggageHeader;
+    reset(0);
+  }
+
+  void parseInto(BaggageBuilder baggageBuilder) {
+    char[] chars = baggageHeader.toCharArray();
```
anuraaga, line 48: Can we just use `charAt` instead of the copy?

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/Parser.java
```diff
@@ -0,0 +1,148 @@
...
+            break;
+          }
+        case ',':
+          {
+            switch (state) {
+              case VALUE:
+                value.tryTerminating(i, baggageHeader);
+                break;
+              case META:
+                meta = baggageHeader.substring(metaStart, i).trim();
```
anuraaga, line 86: Not a huge deal but is it possible to avoid the trim? Can be a followup anyways

anuraaga (commented):
Sorry for late response, still feel half in vacation mode though :) Small comment

jkwatson:
@kubawach this needs a rebase since we've updated to ubuntu 20.04 for CI

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/Parser.java
kubawach: Done. Somehow I missed that :)

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/Parser.java
kubawach: I'd leave that as a followup. According to the spec (https://w3c.github.io/baggage/#baggage-string) OWS has to be trimmed from the meta upon parsing. Therefore to avoid explicit trim() the parser would need to keep meta in a separate, tracked data structure (just as it does for key and value).

kubawach:
Rebased to main && all comments resolved.

kubawach:
Thanks @jkwatson for merging this 👍

PR #3182, opened 2021-04-27
Add spotbugs at reportLevel high; fix or exclude found items.

johnbley:
I've found spotbugs with reportLevel=high to generally have a good signal:noise ratio for a linting tool.  `opentelemetry-java-instrumentation` uses it as well.  This PR suggests adding it to the default `check` target as a build failure, fixes some suboptimal `equals` implementations it found, and adds exclusion rules to make the build pass.

### DIFF
--- a/buildscripts/spotbugs-exclude.xml
+++ b/buildscripts/spotbugs-exclude.xml
@@ -1,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<FindBugsFilter>
+  <!-- Nullable-related stuff seem busted against this repo for whatever reason -->
+  <Match>
+    <Bug pattern="NP_NONNULL_PARAM_VIOLATION"/>
+  </Match>
+  <Match>
+    <Bug pattern="NP_PARAMETER_MUST_BE_NONNULL_BUT_MARKED_AS_NULLABLE"/>
+  </Match>
+  <Match>
+  <!-- This an an oddly specific one... -->
+    <Bug pattern="SC_START_IN_CTOR"/>
+  </Match>
+  <Match>
+    <!-- https://github.com/spotbugs/spotbugs/issues/573 kotlin is not well supported (yet) -->
+    <Source name="~.*\.kt"/>
+  </Match>
+  <Match>
+    <Source name="~.*(Test|Benchmark).*"/>
+  </Match>
+  <Match>
+    <!-- Another broken null detector -->
+    <Class name="io.opentelemetry.api.common.AttributesBuilder"/>
+    <Bug pattern="NP_NULL_PARAM_DEREF"/>
+  </Match>
+  <Match>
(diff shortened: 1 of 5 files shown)


context/src/main/java/io/opentelemetry/context/internal/shaded/AbstractWeakConcurrentMap.java
```diff
@@ -283,7 +283,7 @@ public boolean equals(Object other) {
       if (other instanceof WeakKey<?>) {
         return ((WeakKey<?>) other).get() == get();
       } else {
-        return other.equals(this);
+        return other == null ? false : other.equals(this);
```
Oberon00, line 286:
```suggestion
        return other != null && other.equals(this);
```

I prefer not using `?:` in purely boolean expressions.

buildscripts/spotbugs-exclude.xml
```diff
@@ -0,0 +1,36 @@
...
+  <!-- Nullable-related stuff seem busted against this repo for whatever reason -->
+  <Match>
+    <Bug pattern="NP_NONNULL_PARAM_VIOLATION"/>
+  </Match>
+  <Match>
+    <Bug pattern="NP_PARAMETER_MUST_BE_NONNULL_BUT_MARKED_AS_NULLABLE"/>
+  </Match>
+  <Match>
+  <!-- This an an oddly specific one... -->
+    <Bug pattern="SC_START_IN_CTOR"/>
```
jkwatson, line 12: can you add a comment about what this check is that we're ignoring? Maybe it's something we should fix?

context/src/main/java/io/opentelemetry/context/internal/shaded/AbstractWeakConcurrentMap.java
```diff
@@ -283,7 +283,7 @@ public boolean equals(Object other) {
       if (other instanceof WeakKey<?>) {
         return ((WeakKey<?>) other).get() == get();
       } else {
-        return other.equals(this);
+        return other != null && other.equals(this);
```
anuraaga, line 286: We excluded these two files from checks since they're vendored code and we'd prefer not to diverge for maintenance reasons. Does that make sense?

buildscripts/spotbugs-exclude.xml
johnbley: It's a check for "starting a thread in a constructor" - the cleanerThread in `WeakConcurrentMap`.  I'll adjust the comment.

context/src/main/java/io/opentelemetry/context/internal/shaded/AbstractWeakConcurrentMap.java
johnbley: Ah, ok, this makes sense then.  I will drop this PR per your other comment.

PR #3183, opened 2021-04-27
Add a task to compare public APIs for every stable module.

### LINKED ISSUE #2692: Automatic Binary Compatibility checking of API
It'd be nice to get a report of binary-compatibility breakages in the API from release to release.

PR contributions that have accidental binary-compatibility-breaking changes in stable portions of the code get an automatic check/failure.

We could, instead, rely on unit tests and careful review of semantics/API changes from Approvers/Maintainers.

**Additional context**
There are a few options for automatic checkers, some with gradle support.  Listed in the order I think to be best suited for this project:

- [JAPI compare](https://github.com/siom79/japicmp) based on the old semantic-versioning maven plugin
- [Mima](https://github.com/lightbend/mima) ([Gradle Plugin](https://github.com/borissmidt/gradle-mima-plugin)) - The "Scala" checker that should work for java.
- [JAPI compliance checker](https://github.com/lvc/japi-compliance-checker)

jkwatson:
This adds a `japicmp` task that will generate a diff of public APIs. By default, it will generate the current local version vs. the latest released version. You can also specify specific versions to compare and generate diffs for with two gradle properties: 

For example
`./gradlew japicmp -PapiBaseVersion=1.0.0 -PapiNewVersion=1.1.0` 
will generate diffs between 1.1.0 and 1.0.0. 

I would like every PR to include updates to the diff files (if any), but I'm not sure the best way to enforce that.

I would also like, before every release, a maintainer generate the diffs vs. the previous release and make sure that there are no surprise API changes or breakages.

As a baseline, I have generated diffs between 1.1.0 and 1.0.0, for historical completeness, as well as the current diffs vs. the latest.

This code is pretty hacky, so suggestions for improvement are very welcome!

resolves #2692


exporters/zipkin/docs/api_diff_1.1.0_vs_1.0.0.txt
```diff
@@ -0,0 +1,4 @@
+Comparing source compatibility of  against 
+***  MODIFIED CLASS: PUBLIC FINAL io.opentelemetry.exporter.zipkin.ZipkinSpanExporter  (not serializable)
+	===  CLASS FILE FORMAT VERSION: 52.0 <- 52.0
+	+++  NEW FIELD: PUBLIC(+) STATIC(+) FINAL(+) java.util.logging.Logger baseLogger
```
jkwatson, line 4: this is a bummer that I found while working on this. And, why I think we should make something like this check a part of our normal process.

api/all/docs/api_diff_1.1.0_vs_1.0.0.txt
```diff
@@ -0,0 +1,2 @@
+Comparing source compatibility of  against
```
jkwatson, line 1: Note: the lack of versions in the file appears to be due to this bug: https://github.com/melix/japicmp-gradle-plugin/issues/31

jkwatson:
Also note that this picked up the API change from #2799 after I rebased. See commit https://github.com/open-telemetry/opentelemetry-java/pull/3183/commits/891b1adcf63dda3467b44e492634a319855534fd

build.gradle.kts
```diff
@@ -422,9 +424,86 @@ subprojects {
         }
     }
 
+    plugins.withId("me.champeau.gradle.japicmp") {
+        afterEvaluate {
+            // hack to find the current released version of the project
+            val temp: Configuration = project.configurations.create("tempConfig")
+            project.dependencies.add("tempConfig", "io.opentelemetry:opentelemetry-bom:latest.release")
+            val moduleVersion = project.configurations["tempConfig"].resolvedConfiguration.firstLevelModuleDependencies.elementAt(0).moduleVersion
```
anuraaga, line 432: I believe you can use `latest.release` below without resolving this separately.

build.gradle.kts
```diff
@@ -422,9 +424,86 @@ subprojects {
...
+            project.configurations.remove(temp)
+
+            val userRequestedBase = project.properties["apiBaseVersion"]
+            val baselineVersion = userRequestedBase ?: moduleVersion
+            val newVersion : String? = project.properties["apiNewVersion"] as String?
+            val baselineArtifact: File?
+            var newArtifact: File? = null
+            val oldGroup = project.group
+            try {
+                //this dance is needed due to gradle not allowing more than one version of a dependency.
```
anuraaga, line 442: Nit Let's add space after `//`

build.gradle.kts
```diff
@@ -422,9 +424,86 @@ subprojects {
...
+                project.group = "virtual_baseline_for_japicmp"
```
anuraaga, line 443: You mention a dance but I'm still confused, what's the reason for having to set project.group? Don't think it's related to dependency resolution in any way, only publishing

jkwatson: I cribbed from here:  https://android.googlesource.com/platform/external/grpc-grpc-java/+/refs/heads/android10-d4-release/build.gradle

I think gradle will only let a module have a single resolved version of a given artifact.

build.gradle.kts
jkwatson: Not sure what you mean

build.gradle.kts
jkwatson: (note, I found that code by following several issues about how to implement this particular kind of check in the japicmp project)

jkwatson: this issue: https://github.com/melix/japicmp-gradle-plugin/issues/16

build.gradle.kts
```diff
@@ -422,9 +424,86 @@ subprojects {
         }
     }
 
+    plugins.withId("me.champeau.gradle.japicmp") {
+        afterEvaluate {
```
anuraaga, line 428: Looks like we need to move this logic into a task (`CreateJapiJars`?). Otherwise it all happens on any build, which when cached may not be that bad, is still quite slow.

build.gradle.kts
```diff
@@ -422,9 +424,86 @@ subprojects {
...
+            // hack to find the current released version of the project
+            val temp: Configuration = project.configurations.create("tempConfig")
+            project.dependencies.add("tempConfig", "io.opentelemetry:opentelemetry-bom:latest.release")
+            val moduleVersion = project.configurations["tempConfig"].resolvedConfiguration.firstLevelModuleDependencies.elementAt(0).moduleVersion
+            project.configurations.remove(temp)
+
+            val userRequestedBase = project.properties["apiBaseVersion"]
+            val baselineVersion = userRequestedBase ?: moduleVersion
```
anuraaga, line 436:
```suggestion
            val baselineVersion = userRequestedBase ?: "latest.release"
```

And don't need the previous logic, I think?

build.gradle.kts
anuraaga: I see - I think the comment can be something like "Temporarily change the group name because we want to fetch an artifact with the same Maven coordinates as the project, which Gradle would not allow otherwise."

build.gradle.kts
```diff
@@ -422,9 +424,86 @@ subprojects {
...
+                    project.group = oldGroup
+                }
+            }
+
+            tasks {
+                val jar = getByName("jar") as Jar
+                if (newArtifact == null) {
+                    newArtifact = file(jar.archiveFile)
+                }
+                create<JapicmpTask>("japicmp") {
```
anuraaga, line 476:
```suggestion
                val japicmp by registering<JapicmpTask> {
```

build.gradle.kts
```diff
@@ -422,9 +424,86 @@ subprojects {
...
+                    packageExcludes = listOf("*.internal", "io.opentelemetry.internal.shaded.jctools.*")
+                    if (newVersion == null) {
+                        val baseVersionString = if (userRequestedBase == null) "latest" else baselineVersion
+                        txtOutputFile = file("$projectDir/docs/api_diff_current_vs_${baseVersionString}.txt")
+                    } else {
+                        txtOutputFile = file("$projectDir/docs/api_diff_${newVersion}_vs_${baselineVersion}.txt")
+                    }
+                }
+                //have the build task depend on the api comparison task, to make it more likely it will get used.
+                getByName("build").dependsOn("japicmp")
```
anuraaga, line 495:
```suggestion
                named("build") {
                  dependsOn(japicmp)
                }
```

build.gradle.kts
```diff
@@ -422,9 +424,86 @@ subprojects {
...
+            }
+        }
+    }
+
     plugins.withId("maven-publish") {
+        //generate the api diff report for any module that is stable and publishes a jar.
+        if (!project.hasProperty("otel.release")
+                && project.tasks.map { it.name }.contains("jar")) {
```
anuraaga, line 503: Do you think it's simpler to just do something like `project.name.startsWith("bom")` instead of the jar task check? While in practice we would never do it, but I try to avoid ordering dependencies where possible, checking for `jar` means the java plugin must have been applied before the maven-publish plugin in the build script, fine in practice, but if can be avoided it's sort of nice.

api/all/docs/api_diff_1.1.0_vs_1.0.0.txt
anuraaga, line 1: I think it'd be better to aggregate all these into top level, `docs/${newVersion}/${archivesBaseName}` to get an idea of all the changes in a version more easily.

build.gradle.kts
```diff
@@ -422,9 +424,86 @@ subprojects {
...
+            }
+
+            tasks {
+                val jar = getByName("jar") as Jar
+                if (newArtifact == null) {
+                    newArtifact = file(jar.archiveFile)
+                }
+                create<JapicmpTask>("japicmp") {
+                    dependsOn("jar")
+                    oldClasspath = files(baselineArtifact)
```
anuraaga, line 478: Just wondering whether you considered aggregating all jars into one task instead of per-project? I can see pros and cons of having a single file vs. many files

build.gradle.kts
jkwatson: I'll try it. not sure if it'll properly resolve or not

build.gradle.kts
jkwatson: where do I get `registering` from? IDEA doesn't seem to know it's a thing

api/all/docs/api_diff_1.1.0_vs_1.0.0.txt
jkwatson: yeah, at one point I had this implemented and decided it would be better to keep the diffs close to the source, but I don't feel super strongly about that.

Also, I did want to make sure that we saw diffs vs. the latest release. what directory would that end up in, do you think?  Or, would we just put the concrete ones up at the top level for the historical records, and put the incremental snapshot changes inside the modules?

build.gradle.kts
jkwatson: It is slower than I expected, yes. How will moving it to a task help, though, if `build` depends on it?

build.gradle.kts
jkwatson: or, rather, IDEA sees two of these `registering` delegate functions, but neither of them apply

build.gradle.kts
jkwatson: I didn't think of making only one task, but mostly because I barely got this approach working. :) I definitely think we can evolve this over time.

build.gradle.kts
jkwatson: good question. I don't have a strong opinion on this at all. I was just trying to get something working. :)

build.gradle.kts
jkwatson: hmm. the issue is that we don't have a way to name the jar file down on these lines:
``` 
                val depModule = "io.opentelemetry:${base.archivesBaseName}:$baselineVersion@jar"
                val depJar = "${base.archivesBaseName}-${baselineVersion}.jar"
```
There may be another way to do it, but we'd probably have to start hacking on the gradle plugin (which I want to do to fix some bugs anyway)

build.gradle.kts
jkwatson: yep, this works fine and is definitely cleaner.

build.gradle.kts
jkwatson: It would probably be cleaner to have it in a separate task, though, I agree.

build.gradle.kts
jkwatson: aha. I figured it out. It takes the task class as a parameter.

build.gradle.kts
jkwatson: One thing that would make this maybe less good is that there is information that would have to be communicated from CreateApiJars task down into the japicmp task. It could be done via putting stuff into externals, but that's a little brittle as well.

jkwatson: for now, I think it makes sense to put all the logic inside the one task. I'll update to make it so all of that is in the task, so that non-`build` tasks don't need to have this stuff in it.

jkwatson: I refactored all of this to put all the logic into the one task, as well as extract some helper functions which I think makes it much easier to understand what's going on.

context/src/main/java/io/opentelemetry/context/Context.java
```diff
@@ -108,6 +108,8 @@ static Context root() {
    * during an invocation to another thread. For example, you may use something like {@code Executor
    * dbExecutor = Context.wrapTasks(threadPool)} to ensure calls like {@code dbExecutor.execute(()
    * -> database.query())} have {@link Context} available on the thread executing database queries.
+   *
+   * @since 1.1.0
```
kubawach, line 112: Nice! Shows that the tool can be very useful also for documentation purposes.

exporters/zipkin/docs/api_diff_1.1.0_vs_1.0.0.txt
kubawach: Totally agree with above statement.

kubawach (approved):
Great stuff, I'd really like to see that in instrumentation as well.

ocelotl:
This is cool :sunglasses: We are doing something similar in [Python](https://github.com/open-telemetry/opentelemetry-python/pull/1816).

api/all/docs/api_diff_1.1.0_vs_1.0.0.txt
anuraaga: I think I'd still be find with diff vs current (`docs/current/`) to be aggregated, and we get to see all the new API in the next version very easily.

build.gradle.kts
```diff
@@ -24,6 +25,43 @@ plugins {
...
     id("net.ltgt.errorprone") apply false
     id("ru.vyarus.animalsniffer") apply false
+    id("me.champeau.gradle.japicmp") apply false
+}
+
+
+/**
+ * Locate the project's artifact of a particular version.
+ */
+fun Project.findArtifact(version: String) : File {
```
anuraaga, line 35: Fun with kotlin? :P

api/all/docs/api_diff_1.1.0_vs_1.0.0.txt
anuraaga: Both of my suggestions should have been `docs/apidiffs` by the way :)

anuraaga: I guess in the meantime it wouldn't be that hard for us to overwrite the file with a line that has an appropriate version if we want to. But no big deal

build.gradle.kts
```diff
@@ -422,9 +460,54 @@ subprojects {
...
     }
 
+    plugins.withId("me.champeau.gradle.japicmp") {
+        afterEvaluate {
+            tasks {
+                val jApiCmp by registering(JapicmpTask::class) {
+                    dependsOn("jar")
+                    // the japicmp "old" version is either the user-specified one, or the latest release.
+                    val userRequestedBase = project.properties["apiBaseVersion"] as String?
+                    val baselineVersion: String = userRequestedBase ?: latestReleasedVersion
```
anuraaga, line 470: I've never seen this lazy pattern in any other language so far - that's pretty sweet

anuraaga (approved):
I'm ok with this as-is, but do have a strongish preference to collecting the docs into one if not in this PR then a follow up

api/all/docs/api_diff_1.1.0_vs_1.0.0.txt
jkwatson: cool. I'll update this to put everything under docs/apidiffs:

`docs/apidiffs/current/<module>.txt`
`docs/apidiffs/v1.1.0_v1.0.0/<module>.txt`

jkwatson: definitely something we could enhance ourselves with a little sed or something.

jkwatson: ok, moved and regenerated. please take a look.

PR #3186, opened 2021-04-28
Coerce null AttributeKey to empty string

anuraaga:
`AttributeKey.getKey()` does not have the `@Nullable` annotation, but can return `null` which isn't really allowed. We hide this from users in almost all practical APIs, for example by filtering them out when building `Attributes` but still, I think we should make our `@Nullable` aspect correct (well, I am trying to add Nullaway to the build and found this, which Nullaway doesn't like correctly. Most Nullaway fixes will be trivial but split this one out since it's a logic change). 

The spec forbids both null and empty string keys
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attributes

We end up creating them to avoid crashing apps. But I think it means that coercing null to empty does not make us any less spec compliant and fixes the nullness issue.


jkwatson (approved):
Seems reasonable to me.

PR #3193, opened 2021-05-04
Fix otlp metric proto summary quantile conversion bug.

### LINKED ISSUE #3192: Conversion to SummaryDataPoint protos has wrong percentile
**Describe the bug**
The [MetricAdapter](https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MetricAdapter.java#L232-L233) conversion of summary data points to OTLP protos sets the max quantile to be `100.0` instead of `1.0`. 

The  [ValueAtPercentile](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/ValueAtPercentile.java#L21) class defines percentiles as `[0.0, 100.0]` while the [metrics proto](https://github.com/open-telemetry/opentelemetry-proto/blob/a17f202fdae65e828ac29fb663b5bc5b64b13290/opentelemetry/proto/metrics/v1/metrics.proto#L609) defines percentiles to be `[0.0, 1.0]`, so a conversion is needed between the java and metrics proto representations. 

**Steps to reproduce**
Generate summary metrics, export them over OTLP, and view the quantiles being reported in the logs. 
If possible, provide a recipe for reproducing the error.

The summary max quantile exported as `1.0`. 

**What did you see instead?**
The summary max quantile exported as `100.0`. 

Latest.

jack-berg:
Closes #3192.

### DIFF
--- a/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MetricAdapter.java
+++ b/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MetricAdapter.java
@@ -229,7 +229,7 @@ public final class MetricAdapter {
         for (ValueAtPercentile valueAtPercentile : doubleSummaryPoint.getPercentileValues()) {
           builder.addQuantileValues(
               DoubleSummaryDataPoint.ValueAtQuantile.newBuilder()
-                  .setQuantile(valueAtPercentile.getPercentile())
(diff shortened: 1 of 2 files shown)


jkwatson (approved):
Thanks! good catch!

PR #3200, opened 2021-05-05
Make the EnvironmentResource public.

### LINKED ISSUE #2858: Expose the EnvironmentResource in the auto-configuration module
This resource should also be available for direct usage for people who want to carefully manage their Resource construction process.

jkwatson:
resolves #2858 

Open question: is this in a good package?


anuraaga (approved):

anuraaga (commented):
Hmm - I can't imagine another resource like this in autoconfigure so new package seems like overkill. But it could happen so making a `resources` package is probably slightly better

jkwatson:
> Hmm - I can't imagine another resource like this in autoconfigure so new package seems like overkill. But it could happen so making a `resources` package is probably slightly better

So, if we move this out of the core autoconfigure package, then we would be forced to either a) expose `ConfigProperties.get()` or just copy the sysprop/env var handling code into this class. Any thoughts about that?

anuraaga:
Hmm good point - in that case let's just keep it where it is given the low chance of finding a new resource in this package.

PR #3201, opened 2021-05-05
Add dropped attribute/event counts to jaeger and zipkin exporters.

### LINKED ISSUE #3197: Record dropped attribute/event counts in non-OTLP exporters
See https://github.com/open-telemetry/opentelemetry-specification/pull/1662

jkwatson:
Resolves #3197

### DIFF
--- a/exporters/jaeger-thrift/src/main/java/io/opentelemetry/exporter/jaeger/thrift/Adapter.java
+++ b/exporters/jaeger-thrift/src/main/java/io/opentelemetry/exporter/jaeger/thrift/Adapter.java
@@ -80,19 +80,15 @@ final class Adapter {
         TimeUnit.NANOSECONDS.toMicros(span.getEndEpochNanos() - span.getStartEpochNanos()));
 
     List<Tag> tags = toTags(span.getAttributes());
-    int numberOfAttributes = span.getAttributes().size();
-    if (numberOfAttributes != span.getTotalAttributeCount()) {
-      tags.add(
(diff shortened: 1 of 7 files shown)


exporters/jaeger-thrift/src/main/java/io/opentelemetry/exporter/jaeger/thrift/Adapter.java
```diff
@@ -78,8 +80,21 @@ static Span toJaeger(SpanData span) {
         TimeUnit.NANOSECONDS.toMicros(span.getEndEpochNanos() - span.getStartEpochNanos()));
 
     List<Tag> tags = toTags(span.getAttributes());
+    int numberOfAttributes = span.getAttributes().size();
```
anuraaga, line 83: Possible minor readability improvement throughout

```java
int droppedOf* = size() - count();
if (droppedOf != 0) {
}
```

Reads the same as what we want, if there are dropped, record it

anuraaga (approved):

exporters/jaeger-thrift/src/main/java/io/opentelemetry/exporter/jaeger/thrift/Adapter.java
jkwatson: that's totally fair; I can make that change tomorrow.

jkwatson: done!

PR #3204, opened 2021-05-06
Add a truly no-op OpenTelemetry API as an extension.

anuraaga:
This is a strawmans proposal for a completely no-op implementation of the OpenTelemetry API. The current default OpenTelemetry, while low overhead, still introduces allocations for `SpanBuilder`, `PropagatedSpan`, and `Context`, and mutates `ThreadLocal`, an operation notorious for having reasonable overhead.

If a framework is already used by hundreds of services and generating billions of dollars, the owners will always be wary of any change that could introduce negative effects. In terms of instrumentation, I've generally seen this manifest in a way that the framework wants control of new code, so instead of using the OpenTelemetry API directly, they introduce their own instrumentation API and implement that with OpenTelemetry, or whatever. So many OpenTelemetry's in the world...

I have a hypothesis that this can be mitigated by having a truly no-op API - no allocations, no ThreadLocal. It is obvious by looking at the relatively small amount of code here that it is safe and will not introduce negative effects if used to instrument such a library. The framework can default `OpenTelemetry` to this, and if a framework user then decides to enable remote propagation, they pass to the framework `OpenTelemetry.propagating`, or if they want to do actual tracing, they will add the OpenTelemetry SDK themselves and pass it in. This makes the decision of adding any overhead on the user, not the framework.

That being said, there are a couple of issues, but I think they're still OK, possibly with extra documentation.

- There is already `OpenTelemetry.noop()`.This one is even more no-op :) Because these aren't in the same artifact I think it's survivable
- `Context.current()` does not equal `Context.root()`. In practice, this should not cause problems but is a big difference in behavior. We can add `root` to `ContextStorageProvider` if this is a big deal (I think I had it in one early iteration, removed without seeing the use case but now I found one :).

Happy to hear any thoughts.


Oberon00:
I think the bundled API implementation should be the most minimal possible. If the spec currently does not allow that, we should push for changes there.

anuraaga:
@Oberon00 I think I agree but am not sure about the feasibility 1) context propagation has no concept of API vs implementation right now and 2) it seems highly breaking

Would you be ok to include this as `-alpha` in the meantime?

extensions/noop-api/src/main/java/io/opentelemetry/extension/noopapi/NoopTracerProvider.java
```diff
@@ -0,0 +1,104 @@
...
+  enum NoopTracer implements Tracer {
+    INSTANCE;
+
+    @Override
+    public SpanBuilder spanBuilder(String spanName) {
+      return NoopSpanBuilde.INSTANCE;
+    }
+  }
+
+  enum NoopSpanBuilde implements SpanBuilder {
```
jkwatson, line 41: missing an `r` at the end of the enum name

jkwatson:
How close would it be if we just had our existing API with no propagators and this no-op context storage installed? I *think* it's functionally equivalent, and might be a slightly easier solution?

anuraaga:
@jkwatson It'd remove a lot of the overhead related to context, but still have the allocation of `SpanBuilder` and `PropagatedSpan`. It's pretty good but does remove the nice story of "no overhead at all" when adding OpenTelemetry to a framework.

jkwatson:
> @jkwatson It'd remove a lot of the overhead related to context, but still have the allocation of `SpanBuilder` and `PropagatedSpan`. It's pretty good but does remove the nice story of "no overhead at all" when adding OpenTelemetry to a framework.

I think that PropagatedSpan actually is no-alloc since the Span.wrap() gives you a singleton invalid one if the SpanContext is invalid.  SpanBuilder is a good point, though.

anuraaga:
@jkwatson Do you think we can merge this?

jkwatson (approved):
I'm ok with this as an alpha extension. I want to understand more about how/if/when it will be used and by whom. :)

anuraaga:
@jkwatson Yup me too. Going to link to the merged folder in some discussions which is nicer than the open PR :)

Issue #1564, opened 2020-08-20
Use Java 9+ APIs for better precision in ProcessResource


anuraaga:
Currently, we use de-facto, but not well defined APIs for extracting process information and are limited in what we can get. We should use reflection to access Java 9+ APIs when available to provide better information when running on modern JVM.

From https://github.com/open-telemetry/opentelemetry-java/pull/1555/files#diff-a802c9460480a3f888b7e6fa984c5686R32

anuraaga:
Also at the same time it would be good to switch the current platform calls to reflection as well to unbreak Android support (though it'd be quite limited still).

iNikem:
Anybody thinks that multi-release jar is a good way to do that?

anuraaga:
Yeah MRJar would work well too :)

piotr-sumo:
@anuraaga can I contribute to this issue?

anuraaga:
Thanks @piotr-sumo!

piotr-sumo:
@anuraaga Basically I have the reflection version ready. 

However, if we want to go with multi-release jar I'll change my code.

PR #3212, opened 2021-05-07
Use JDK 10 API to fetch process pid

piotr-sumo:
Resolves https://github.com/open-telemetry/opentelemetry-java/issues/1564
`ManagementFactory.getRuntimeMXBean().getPid()` is available since JDK 10, but I've come across https://bugs.openjdk.java.net/browse/JDK-8209058 when compiling. I've decided to go with JDK 11 because the code compiles and it is LTS release anyway.

### DIFF
--- a/sdk-extensions/resources/build.gradle.kts
+++ b/sdk-extensions/resources/build.gradle.kts
@@ -8,6 +8,8 @@ plugins {
 description = "OpenTelemetry SDK Resource Providers"
 extra["moduleName"] = "io.opentelemetry.sdk.extension.resources"
 
+val mrJarVersions = listOf(11)
+
 dependencies {
     api(project(":sdk:common"))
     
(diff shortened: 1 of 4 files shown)


piotr-sumo:
@anuraaga do you have any hints how to pass `codecov/patch` check? I don't want to introduce any fake tests just to increase code coverage.

jkwatson:
> @anuraaga do you have any hints how to pass `codecov/patch` check? I don't want to introduce any fake tests just to increase code coverage.

Can the strategy used here work? https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/all/src/test/java/io/opentelemetry/sdk/internal/SystemClockTest.java

anuraaga (approved):
I don't think we merge coverage across jvms. I don't think we have to worry about diff coverage, in this case though should be ok

jkwatson (approved):
Thanks!

PR #3226, opened 2021-05-12
Move TemporaryBuffers to internal and use it everywhere.

anuraaga:
Also replaces loop over `String` with `getChars` which is hotspottable.

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/internal/TemporaryBuffers.java
+++ b/api/all/src/main/java/io/opentelemetry/api/internal/TemporaryBuffers.java
@@ -11,6 +11,9 @@ package io.opentelemetry.api.internal;
  * multiple derived objects at the same time because the same memory will be used. In general, you
  * should get a temporary buffer, fill it with data, and finish by converting into the derived
  * object within the same method to avoid multiple usages of the same buffer.
+ *
+ * <p>This class is internal and is hence not for public use. Its APIs are unstable and can change
+ * at any time.
  */
 public final class TemporaryBuffers {
 
--- a/api/all/src/main/java/io/opentelemetry/api/internal/TemporaryBuffers.java
+++ b/api/all/src/main/java/io/opentelemetry/api/internal/TemporaryBuffers.java
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package io.opentelemetry.api.trace.propagation;
+package io.opentelemetry.api.internal;
 
 /**
  * {@link ThreadLocal} buffers for use when creating new derived objects such as {@link String}s.
@@ -12,7 +12,7 @@ package io.opentelemetry.api.trace.propagation;
  * should get a temporary buffer, fill it with data, and finish by converting into the derived
  * object within the same method to avoid multiple usages of the same buffer.
  */
-final class TemporaryBuffers {
+public final class TemporaryBuffers {
 
   private static final ThreadLocal<char[]> CHAR_ARRAY = new ThreadLocal<>();
 
@@ -22,7 +22,7 @@ final class TemporaryBuffers {
    * not be zeroed and may be larger than the requested size, you must make sure to fill the entire
    * content to the desired value and set the length explicitly when converting to a {@link String}.
    */
-  static char[] chars(int len) {
+  public static char[] chars(int len) {
     char[] buffer = CHAR_ARRAY.get();
     if (buffer == null || buffer.length < len) {
       buffer = new char[len];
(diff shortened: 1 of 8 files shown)


api/all/src/main/java/io/opentelemetry/api/internal/TemporaryBuffers.java
```diff
@@ -12,7 +12,7 @@
  * should get a temporary buffer, fill it with data, and finish by converting into the derived
  * object within the same method to avoid multiple usages of the same buffer.
  */
-final class TemporaryBuffers {
+public final class TemporaryBuffers {
```
jkwatson, line 15: We should add the boilerplate about internal classes to this javadoc, I think.

jkwatson (approved):
Thanks! Any observable performance improvements in any of the benchmarks from the jittability changes?

breedx-splk (approved):
This is a nice little optimization!
Seems like in a few cases we would end up hanging on to more memory in the long term, but the kinds of buffers/strings used here are pretty small...so this is probably better (in terms of saving the allocations).

api/all/src/main/java/io/opentelemetry/api/internal/TemporaryBuffers.java
anuraaga: Yup added

anuraaga (commented):
> Thanks! Any observable performance improvements in any of the benchmarks from the jittability changes?

Was a bit lazy to compare those but let me try. Not expecting a whole lot, I also liked the less LoC :)

> Seems like in a few cases we would end up hanging on to more memory in the long term, but the kinds of buffers/strings used here are pretty small

Luckily, now that there's only one class, we'll end up with less memory left around than before since we only have one per thread, instead of three (w3c, trace ID, span ID) :D

anuraaga:
Improvement with `getChars` is more than I was expecting, ~20%. W3C propagator only changed to use `getChars`

After
```
Benchmark                                                                                Mode  Cnt     Score    Error   Units
W3CTraceContextPropagatorInjectBenchmark.measureInject                                   avgt   15    40.443 ±  0.357   ns/op
W3CTraceContextPropagatorInjectBenchmark.measureInject:·gc.alloc.rate                    avgt   15  1507.718 ± 13.883  MB/sec
W3CTraceContextPropagatorInjectBenchmark.measureInject:·gc.alloc.rate.norm               avgt   15    96.011 ±  0.001    B/op
W3CTraceContextPropagatorInjectBenchmark.measureInject:·gc.churn.G1_Eden_Space           avgt   15  1503.231 ± 87.734  MB/sec
W3CTraceContextPropagatorInjectBenchmark.measureInject:·gc.churn.G1_Eden_Space.norm      avgt   15    95.712 ±  5.268    B/op
W3CTraceContextPropagatorInjectBenchmark.measureInject:·gc.churn.G1_Survivor_Space       avgt   15     0.008 ±  0.003  MB/sec
W3CTraceContextPropagatorInjectBenchmark.measureInject:·gc.churn.G1_Survivor_Space.norm  avgt   15     0.001 ±  0.001    B/op
W3CTraceContextPropagatorInjectBenchmark.measureInject:·gc.count                         avgt   15   153.000           counts
W3CTraceContextPropagatorInjectBenchmark.measureInject:·gc.time                          avgt   15   100.000               ms
```

Before
```
Benchmark                                                                                Mode  Cnt     Score    Error   Units
W3CTraceContextPropagatorInjectBenchmark.measureInject                                   avgt   15    60.323 ±  0.657   ns/op
W3CTraceContextPropagatorInjectBenchmark.measureInject:·gc.alloc.rate                    avgt   15  1010.462 ± 10.567  MB/sec
W3CTraceContextPropagatorInjectBenchmark.measureInject:·gc.alloc.rate.norm               avgt   15    96.013 ±  0.001    B/op
W3CTraceContextPropagatorInjectBenchmark.measureInject:·gc.churn.G1_Eden_Space           avgt   15  1009.866 ± 59.877  MB/sec
W3CTraceContextPropagatorInjectBenchmark.measureInject:·gc.churn.G1_Eden_Space.norm      avgt   15    95.961 ±  5.689    B/op
W3CTraceContextPropagatorInjectBenchmark.measureInject:·gc.churn.G1_Survivor_Space       avgt   15     0.005 ±  0.003  MB/sec
W3CTraceContextPropagatorInjectBenchmark.measureInject:·gc.churn.G1_Survivor_Space.norm  avgt   15    ≈ 10⁻³             B/op
W3CTraceContextPropagatorInjectBenchmark.measureInject:·gc.count                         avgt   15   124.000           counts
W3CTraceContextPropagatorInjectBenchmark.measureInject:·gc.time                          avgt   15    75.000               ms
```

anuraaga:
Ah just in case people are curious, I'm fairly sure the `getChars` performance improvement will be mostly 0 on Java 8, the intrinsics were added in Java 9+ if I'm not mistaken, on Java 8 the loop should be the same exact bytecode.

PR #3228, opened 2021-05-12
Upgrade to Gradle 7.0

anuraaga:
All the changes were required due to breakages, many plugins didn't bother maintaining support for both 7+ and pre-7 in the same version


build.gradle.kts
```diff
@@ -361,14 +361,19 @@ subprojects {
...
-                if (isCanBeResolved && !isCanBeConsumed
-                        // Older ones (like JMH) may not, so check the name as well.
-                        // Kotlin compiler classpaths don't support BOM nor need it.
-                        || name.endsWith("Classpath") && !name.startsWith("kotlin")) {
-                    add(name, platform(project(":dependencyManagement")))
+            add(dependencyManagement.name, platform(project(":dependencyManagement")))
+            afterEvaluate {
+                configurations.configureEach {
+                    if (isCanBeResolved && !isCanBeConsumed) {
+                        extendsFrom(dependencyManagement)
```
anuraaga, line 375: This pattern from https://github.com/gradle/gradle/issues/12016#issuecomment-579177392

Somehow I landed on the previous code based on this comment, probably in an attempt to not add another configuration. But looking back at it, this version seems better anyways.

docs/apidiffs/current_vs_latest/opentelemetry-extension-kotlin.txt
```diff
@@ -1,2 +1,10 @@
 Comparing source compatibility of  against 
-No changes.
\ No newline at end of file
+===  UNCHANGED CLASS: PUBLIC FINAL io.opentelemetry.extension.kotlin.ContextExtensionsKt  (not serializable)
```
anuraaga, line 2: Wonder if we need to disable jcmp for kotlin somehow, no clue what's happening here

jkwatson: hrm. yeah, looks like jcmp doesn't know what to make of kotlin classes. :) I'd be ok with disabling it.

jkwatson (approved):
Thanks!

build.gradle.kts
breedx-splk: Agreed.

breedx-splk (approved):
Looks like a solid and worthwhile slog.  Thanks for taking that on and keeping things current.

PR #3229, opened 2021-05-12
Update example dependencies to 1.2.0

### LINKED ISSUE #3122: Figure out a better plan for the examples builds
We really have wanted our examples to do 2 things:
1) Verify that we haven't broken anything in the stable APIs and
2) Provide good examples that work against the latest official release (not SNAPSHOT builds).

Unfortunately, I don't think it's possible to meet both of these goals for all the examples, since the stable SDK depends on the unstable metrics, so pulling when we build against snapshots for the stable stuff (to verify that we haven't broken anything), we pull in the unstable metrics version as well. 

So, we need to sort this out; I suggest we probably need to have another module which is the "don't break the stable APIs" module (which depends on the latest snapshots), separate from the user-facing examples which should document how to use the code based on the latest release.

jkwatson:
Also, remove example build vs. snapshots
Resolves #3122

### DIFF
--- a/.github/workflows/pr-examples-build.yml
+++ b/.github/workflows/pr-examples-build.yml
@@ -30,7 +30,6 @@ jobs:
       - uses: burrunan/gradle-cache-action@v1.10
         with:
           remote-build-cache-proxy-enabled: false
-          ## todo: how to make sure these are kept up to date, so we always build against the latest snapshots for stable modules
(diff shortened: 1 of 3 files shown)


breedx-splk (approved):
Looks good to me.

anuraaga (approved):

PR #3230, opened 2021-05-12
Allow root Context to be overridden

wesleyhillyext:
In our codebase, I've created a custom implementation of the `Context` interface, and a custom `ContextStorage` implementation to go with it, to better fit in with our specific environment.

One issue I encountered is that the `Context.root()` method is hard-coded to return an `ArrayBasedContext`, and by extension, `Context.current()` may also return the `ArrayBasedContext` instead of the desired custom implementation of `Context`.

The change I propose adds a `ContextStorage.root()` method so that a custom context storage can also provide a suitable custom root context. I made it a `default` method so the current behavior is preserved when not overridden. I wasn't sure what the next release would be, so I didn't add a `@since` to the JavaDoc.


jkwatson:
Hi @wesleyhillyext . I think this is probably a fine enhancement (I'll let @anuraaga comment as well, since he's definitely our context expert).  Before we can accept this, however, we'd need you to sign the CLA which should be easy to do via the EasyCLA links above.

Also, this is an API change, but I don't see the corresponding update to the current apidiffs in this PR. Did you run `./gradlew build` and make sure to add any files in the docs/apidiffs/current_vs_latest directory to your commits?

anuraaga (commented):
Thanks yeah this change is needed.

Can you use the new method in the tests that return a custom context implementation? https://github.com/open-telemetry/opentelemetry-java/tree/main/context/src

anuraaga:
By the way, while waiting for this to be released, you can return your context instead of `null` from the storage, it doesn't really harm anything. That would allow `current()` to always have your implementation which is more common than `root()` in many cases.

wesleyhillyext:
Thanks for the quick feedback! I amended the pull request accordingly.

- Added the `current_vs_latest/opentelemetry-context.txt` change.
- Changed the three test implementations of `Context` found so that their accompanying `ContextStorage` overrides the new `root()` method. Hopefully these are right; I'm familiar with the gRPC context but not Brave.

I had seen the CLA bot message, and I'm working to find someone to be a CLA manager at my company.

And thanks, I am indeed using the workaround of not returning `null` from my `ContextStorage.current()`.

anuraaga (approved):

wesleyhillyext:
Hi all; sorry it took a while, but the CLA process is now complete for this pull request. Let me know if you need anything else from my end.

jkwatson:
@wesleyhillyext I think you need to rebase to pick up build changes that happened since you last worked on this. Thanks!

context/src/main/java/io/opentelemetry/context/ContextStorage.java
```diff
@@ -102,4 +102,14 @@ static void addWrapper(Function<? super ContextStorage, ? extends ContextStorage
    */
   @Nullable
   Context current();
+
+  /**
+   * Returns the root {@link Context} which all other {@link Context} are derived from.
+   *
+   * <p>The default implementation returns the root {@link ArrayBasedContext}, but subclasses can
```
anuraaga, line 109:
```suggestion
   * <p>The default implementation returns the root {@code ArrayBasedContext}, but subclasses can
```

ArrayBasedContext isn't public so can't link to it

anuraaga (approved):
Noticed just one nit

kubawach (approved):
For completeness you could also implement root() for NoopContextStorageProvider

jkwatson:
> For completeness you could also implement root() for NoopContextStorageProvider

This can be a follow-on PR if desired. I'd like to get this merged, since it's been open for a very long time.

PR #3244, opened 2021-05-21
Remove null checks from SdkSpanBuilder

anuraaga:


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/AttributesMap.java
```diff
@@ -27,9 +27,6 @@
   }
 
   <T> void put(AttributeKey<T> key, T value) {
-    if (key == null || key.getKey() == null || value == null) {
```
anuraaga, line 30: Checked by all callers, this is an internal class

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -251,7 +251,7 @@ Clock getClock() {
 
   @Override
   public <T> ReadWriteSpan setAttribute(AttributeKey<T> key, T value) {
-    if (key == null || key.getKey() == null || key.getKey().length() == 0 || value == null) {
```
anuraaga, line 254: Removing the unnecessary key null check (null key mapped to empty string now) is originally what had me find the other ones we never removed

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkSpanBuilder.java
```diff
@@ -109,7 +112,6 @@ public SpanBuilder addLink(SpanContext spanContext, Attributes attributes) {
   }
 
   private void addLink(LinkData link) {
-    Objects.requireNonNull(link, "link");
```
anuraaga, line 112: Private method

jkwatson (approved):
Thanks!

carlosalberto (approved):

PR #3256, opened 2021-05-26
Add a prototype PassThroughPropagator

anuraaga:
For https://github.com/open-telemetry/opentelemetry-specification/issues/1717

About 3x performance for passing through vs parse/serialize.

```
Benchmark                                                                 Mode  Cnt      Score     Error   Units
PassThroughPropagatorBenchmark.parse                                     thrpt   60   5705.975 ±  75.797  ops/ms
PassThroughPropagatorBenchmark.parse:·gc.alloc.rate                      thrpt   60   2174.161 ±  28.934  MB/sec
PassThroughPropagatorBenchmark.parse:·gc.alloc.rate.norm                 thrpt   60    600.000 ±   0.001    B/op
PassThroughPropagatorBenchmark.parse:·gc.churn.G1_Eden_Space             thrpt   60   2178.084 ±  43.372  MB/sec
PassThroughPropagatorBenchmark.parse:·gc.churn.G1_Eden_Space.norm        thrpt   60    601.123 ±   9.464    B/op
PassThroughPropagatorBenchmark.parse:·gc.churn.G1_Old_Gen                thrpt   60      0.002 ±   0.001  MB/sec
PassThroughPropagatorBenchmark.parse:·gc.churn.G1_Old_Gen.norm           thrpt   60      0.001 ±   0.001    B/op
PassThroughPropagatorBenchmark.parse:·gc.count                           thrpt   60    792.000            counts
PassThroughPropagatorBenchmark.parse:·gc.time                            thrpt   60    336.000                ms
PassThroughPropagatorBenchmark.passthrough                               thrpt   60  14528.893 ± 158.380  ops/ms
PassThroughPropagatorBenchmark.passthrough:·gc.alloc.rate                thrpt   60   2878.115 ±  31.488  MB/sec
PassThroughPropagatorBenchmark.passthrough:·gc.alloc.rate.norm           thrpt   60    312.000 ±   0.001    B/op
PassThroughPropagatorBenchmark.passthrough:·gc.churn.G1_Eden_Space       thrpt   60   2886.064 ±  51.659  MB/sec
PassThroughPropagatorBenchmark.passthrough:·gc.churn.G1_Eden_Space.norm  thrpt   60    312.864 ±   4.436    B/op
PassThroughPropagatorBenchmark.passthrough:·gc.churn.G1_Old_Gen          thrpt   60      0.002 ±   0.001  MB/sec
PassThroughPropagatorBenchmark.passthrough:·gc.churn.G1_Old_Gen.norm     thrpt   60     ≈ 10⁻⁴              B/op
PassThroughPropagatorBenchmark.passthrough:·gc.count                     thrpt   60    807.000            counts
PassThroughPropagatorBenchmark.passthrough:·gc.time                      thrpt   60    388.000                ms
```

### DIFF
--- a/extensions/incubator/src/main/java/io/opentelemetry/extension/incubator/propagation/PassThroughPropagator.java
+++ b/extensions/incubator/src/main/java/io/opentelemetry/extension/incubator/propagation/PassThroughPropagator.java
@@ -38,6 +38,12 @@ public final class PassThroughPropagator implements TextMapPropagator {
   private static final ContextKey<List<String>> EXTRACTED_KEY_VALUES =
       ContextKey.named("passthroughpropagator-keyvalues");
 
+  private final List<String> fields;
+
+  private PassThroughPropagator(List<String> fields) {
+    this.fields = Collections.unmodifiableList(fields);
+  }
+
   /**
    * Returns a {@link TextMapPropagator} which will propagate the given {@code fields} from
    * extraction to injection.
@@ -61,12 +67,6 @@ public final class PassThroughPropagator implements TextMapPropagator {
       return TextMapPropagator.noop();
     }
     return new PassThroughPropagator(fieldsList);
-  }
-
-  private final List<String> fields;
-
-  private PassThroughPropagator(List<String> fields) {
-    this.fields = Collections.unmodifiableList(fields);
   }
 
   @Override
--- a/extensions/incubator/src/main/java/io/opentelemetry/extension/incubator/propagation/PassThroughPropagator.java
+++ b/extensions/incubator/src/main/java/io/opentelemetry/extension/incubator/propagation/PassThroughPropagator.java
@@ -33,7 +33,7 @@ import javax.annotation.Nullable;
  * io.opentelemetry.api.OpenTelemetry} that only propagates. Similarly, you will never need this
  * when using the OpenTelemetry SDK to enable telemetry.
  */
-public class PassThroughPropagator implements TextMapPropagator {
+public final class PassThroughPropagator implements TextMapPropagator {
 
   private static final ContextKey<List<String>> EXTRACTED_KEY_VALUES =
       ContextKey.named("passthroughpropagator-keyvalues");
--- a/extensions/incubator/src/main/java/io/opentelemetry/extension/incubator/propagation/PassThroughPropagator.java
+++ b/extensions/incubator/src/main/java/io/opentelemetry/extension/incubator/propagation/PassThroughPropagator.java
@@ -39,24 +39,27 @@ public final class PassThroughPropagator implements TextMapPropagator {
       ContextKey.named("passthroughpropagator-keyvalues");
 
   /**
-   * Returns a {@link PassThroughPropagator} which will propagate the given {@code fields} from
+   * Returns a {@link TextMapPropagator} which will propagate the given {@code fields} from
    * extraction to injection.
    */
-  public static PassThroughPropagator create(String... fields) {
+  public static TextMapPropagator create(String... fields) {
     requireNonNull(fields, "fields");
     return create(Arrays.asList(fields));
   }
 
   /**
-   * Returns a {@link PassThroughPropagator} which will propagate the given {@code fields} from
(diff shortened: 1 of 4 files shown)


extensions/incubator/src/main/java/io/opentelemetry/extension/incubator/propagation/PassThroughPropagator.java
```diff
@@ -0,0 +1,99 @@
...
+      String value = getter.get(carrier, field);
+      if (value != null) {
+        if (extracted == null) {
+          extracted = new ArrayList<>();
+        }
+        extracted.add(field);
+        extracted.add(value);
+      }
+    }
+    return extracted != null ? context.with(EXTRACTED_KEY_VALUES, extracted) : context;
```
anuraaga, line 97: Does context key being private provide enough safety to not wrap `extracted` with unmodifiable?

carlosalberto: Yes, it should be fine IMHO

carlosalberto:
Interesting results, thanks for the prototype. Saw you specify `fields` during the propagator creation time, which is not enough for name-variable formats such as the Jaeger one (although it should be possible to add support to it by adding an additional, optional `prefix` parameter or so).

One interesting thing to consider, even if not now, is how will we expose this functionality as env vars, i.e. `pt-tracecontext`, specially if you want to passthrough combined formats, e.g. W3C's TraceContext + Baggage or Zipkin + Baggage.

Overall the numbers look good!

jkwatson:
@anuraaga Is this something we want to merge, or is it good enough to have it as a PR for now?

anuraaga:
> although it should be possible to add support to it by adding an additional, optional prefix parameter or so

Yup I think we can follow up with this as an option since most propagators probably wouldn't need it

> One interesting thing to consider, even if not now, is how will we expose this functionality as env vars

Yeah, I guess one consistent / generic way is similar to OTEL_SAMPLER_ARG, OTEL_PROPAGATOR_ARG, otherwise we'd probably just make a custom variable for it, `OTEL_PASSTHROUGH_PROPAGATOR_FORMATS=tracecontext,b3`. Someone may ask for the ability to specify the fields themselves on env var, but I think it's rare enough that wouldn't support that in the first version.

> Is this something we want to merge, or is it good enough to have it as a PR for now?

If it was merged and released, I would use it even in alpha state, for what that's worth :)

extensions/incubator/src/main/java/io/opentelemetry/extension/incubator/propagation/PassThroughPropagator.java
```diff
@@ -0,0 +1,102 @@
...
+        StreamSupport.stream(fields.spliterator(), false)
+            .map(field -> requireNonNull(field, "field"))
+            .collect(Collectors.toList());
+    if (fieldsList.isEmpty()) {
+      return TextMapPropagator.noop();
+    }
+    return new PassThroughPropagator(fieldsList);
+  }
+
+  private final List<String> fields;
```
jkwatson, line 66: I thought we agreed this goes up top, not below the static methods. 😞 
It's even in the CONTRIBUTING.md :)

extensions/incubator/src/main/java/io/opentelemetry/extension/incubator/propagation/PassThroughPropagator.java
```diff
@@ -0,0 +1,102 @@
...
+
+  private PassThroughPropagator(List<String> fields) {
```
jkwatson, line 68: constructor goes up top, too

extensions/incubator/src/main/java/io/opentelemetry/extension/incubator/propagation/PassThroughPropagator.java
anuraaga: Ah I think it was less agree and more just outvote :P Think I'll continue to forget this in the future

jkwatson (approved):

PR #3267, opened 2021-05-30
Failed to export spans. Error message: CANCELLED: io.grpc.Context was cancelled without error

donbeave:
Hello!

Sometimes some of my microservices are losing spans due to the gRPC error: "CANCELLED: io.grpc.Context was cancelled without error":

catched_error

console_log

I found a solution how to fix this behaviour by forking the current gRPC context and injecting it during the export spans request.

Versions what I'm using:

`opentelemetry-java`: 1.2.0
`opentelemetry-java-instrumentation`: 1.2.0

### DIFF
--- a/exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporter.java
+++ b/exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporter.java
@@ -9,6 +9,7 @@ import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.MoreExecutors;
 import io.grpc.ConnectivityState;
+import io.grpc.Context;
 import io.grpc.ManagedChannel;
 import io.grpc.Status;
 import io.opentelemetry.api.metrics.BoundLongCounter;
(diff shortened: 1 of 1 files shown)


jkwatson:
Could this be an issue with the grpc instrumentation in the agent? I wonder if this is the best way to fix the issue you're seeing. @anuraaga any ideas?

anuraaga:
@donbeave Are you using the Java agent? If not, are you using the BatchSpanProcessor? We never call `Context.cancel` in instrumentation, and if using the BatchSpanProcessor, the context attached when exporting spans should be the root context. The only case I can think of for cancel to affect export is using the SimpleSpanProcessor and cancelling the request handler's context - in that case, the cancellation being propagated seems like correct behavior. But could be missing another case for cancellation that I didn't come up with - @donbeave do you know more details of why this cancellation is happening?

donbeave:
@anuraaga No, I'm not using Java agent, I'm using some libraries from `opentelemetry-java-instrumentation`. For example in this case it's `opentelemetry-grpc-1.6` (https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/grpc-1.6/library) and manually add client and server interceptors. That's all because I'm planning to build my application with the GraalVM native image feature, as I understand Java agent will not work with the native image. And yes, you are right, I'm using `SimpleSpanProcessor`, here is my SDK configuration:

```java
@Factory
public class OpenTelemetryConfig {

    @Bean
    @Singleton
    public SpanExporter otelSpanExporter() {
        return OtlpGrpcSpanExporter.builder().build();
    }

    @Bean
    @Singleton
    public SpanProcessor otelSpanProcessor(SpanExporter spanExporter) {
        return SimpleSpanProcessor.create(spanExporter);
    }

    @Inject
    @Bean
    @Singleton
    public SdkTracerProvider otelSdkTraceProvider(SpanProcessor spanProcessor, ApplicationContext applicationContext) {
        String applicationName = applicationContext.getProperty("micronaut.application.name", String.class)
                .orElse(io.micronaut.context.env.Environment.DEFAULT_NAME);

        return SdkTracerProvider.builder()
                .addSpanProcessor(spanProcessor)
                .setResource(Resource.getDefault().toBuilder().put(SERVICE_NAME, applicationName).build())
                .build();
    }

    @Bean
    @Singleton
    public OpenTelemetry openTelemetry(SdkTracerProvider sdkTracerProvider) {
        return OpenTelemetrySdk.builder()
                .setTracerProvider(sdkTracerProvider)
                .setPropagators(ContextPropagators.create(W3CTraceContextPropagator.getInstance()))
                .buildAndRegisterGlobal();
    }

    @Bean
    @Singleton
    public Tracer otelTracer(OpenTelemetry openTelemetry) {
        return openTelemetry.getTracer("micronaut-opentelemetry", "1.0.0");
    }

}
```

But I can't find any place where a request was cancelled from the called application. It seems not a case for me, if somehow I can proof please let me know how to check that cancellation. But what I see in the called application I'm getting the correct gRPC response. So probably you mean it's somewhere in the receiver application something cancelling the span exporting?

anuraaga:
I'm not sure but wouldn't be surprised if a stub deadline propagates into the context like you're seeing. Either way the exporter is behaving correctly - if we change to fork there will be no way to control cancelation of the export from outside, a loss in functionality.

For production loads, we strongly recommend the batch span processor - any reason to use the simple one? If so, it's probably best to implement your own span processor that forks the context within the processor, it shouldn't be too much code.

donbeave:
@anuraaga I also have tried to use `TraceServiceBlockingStub` instead of `TraceServiceFutureStub` and it works fine as well. The problem only with `TraceServiceFutureStub`, which actually also works in blocking mode here as I understand, because below is using `MoreExecutors.directExecutor()` provided by Guava library.

Yeah, this is not a production configuration, this is what I'm using for running some functional tests, which uses OpenTelemetry. That's why I'm using `SimpleSpanProcessor`. Maybe add `fork` functionality as an option? Not make it the default behavior?

donbeave:
Or maybe use blocking stub here?

Something like this:

```java
        spansSeen.add(spans.size());
        ExportTraceServiceRequest exportTraceServiceRequest =
                ExportTraceServiceRequest.newBuilder()
                        .addAllResourceSpans(SpanAdapter.toProtoResourceSpans(spans))
                        .build();

        TraceServiceGrpc.TraceServiceBlockingStub exporter;
        if (timeoutNanos > 0) {
            exporter = traceService.withDeadlineAfter(timeoutNanos, TimeUnit.NANOSECONDS);
        } else {
            exporter = traceService;
        }

        try {
            ExportTraceServiceResponse response = exporter.export(exportTraceServiceRequest);

            spansExportedSuccess.add(spans.size());
            result.succeed();
        } catch (Throwable t) {
            spansExportedFailure.add(spans.size());
            Status status = Status.fromThrowable(t);
            switch (status.getCode()) {
                case UNIMPLEMENTED:
                    logger.log(
                            Level.SEVERE,
                            "Failed to export spans. Server responded with UNIMPLEMENTED. "
                                    + "Full error message: "
                                    + t.getMessage());
                    break;
                case UNAVAILABLE:
                    logger.log(
                            Level.SEVERE,
                            "Failed to export spans. Server is UNAVAILABLE. "
                                    + "Make sure your collector is running and reachable from this network. "
                                    + "Full error message:"
                                    + t.getMessage());
                    break;
                default:
                    logger.log(
                            Level.WARNING, "Failed to export spans. Error message: " + t.getMessage());
                    break;
            }
            if (logger.isLoggable(Level.FINEST)) {
                logger.log(Level.FINEST, "Failed to export spans. Details follow: " + t);
            }
            result.fail();
        }

        return result;
    }
```

anuraaga:
Even for a non-production case you should use Batch span processor for export, simple span processor's only real use case is logging.

The simple span processor is intended not to block, but if you really need it to you can call forceFlush in your code on the processor. The direct executor only controls the callback thread but doesn't make it a blocking call.

We could add an option to fork context in the simple processor but we don't want anyone to use the simple processor with grpc exporters anyways so lean against it. Can you use the batch processor? We can probably strengthen our docs to discourage simple more for these exporters.

donbeave:
@anuraaga Yes, the batch processor is working fine for me. This configuration works almost same as how simple processor works `BatchSpanProcessor.builder(spanExporter).setMaxQueueSize(1).setMaxExportBatchSize(1).build()`. In my case, I need to flush almost instantly, just right after receiving any event, before the executed test finished and before the application was closed. So, I'm not sure if anyone else needs a fork context option here.

anuraaga:
@donbeave Good to hear the batch processor is working. Can we go ahead and close this PR then? At least for now we don't see a reason to fork context in our processors.

donbeave:
@anuraaga yes, can close this PR. Thank you for your help!

donbeave:
@anuraaga Sorry, I'm back again. I found that `BatchSpanProcessor` sometimes doesn't work in my scenario. I'm using OpenTelemetry in integration tests and sometimes it shutdown the application early than all spans was uploaded, even if my configuration is like this:

```java
    @Bean
    @Singleton
    public SpanExporter otelSpanExporter() {
        return OtlpGrpcSpanExporter.builder()
                .setTimeout(Duration.ZERO)
                .build();
    }

    @Bean
    @Singleton
    public SpanProcessor otelSpanProcessor(SpanExporter spanExporter) {
        return BatchSpanProcessor.builder(spanExporter)
                .setScheduleDelay(Duration.ZERO)
                .setMaxExportBatchSize(1)
                .build();
    }
```

But it works well with `SimpleSpanProcessor`, all spans are exported right after the end of a particular span.

donbeave:
Forgot to add, before closing the application I have tried to force flush spans, but it doesn't solve the problem:

```java
    @PreDestroy
    public void preDestroy(SpanProcessor spanProcessor) {
        spanProcessor.forceFlush();
        spanProcessor.shutdown();
    }
```

Oberon00:
forceFlush and shutdown both return a CompletableResultCode which you should probably `join` to wait for them to actually complete.

donbeave:
@Oberon00 you are right, now it works, sorry for bothering again.

anuraaga:
Thanks @donbeave - just wondering, in case others run into it too, we define a `close` method which also calls shutdown and join, which I thought Spring automatically calls on shutdown

https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/Bean.html#destroyMethod--

Do you know if this was happening for you or why it might not have? The intent was for Spring users to not have to worry about flushing on shutdown so if that's not working would be nice to understand.

donbeave:
@anuraaga yeah, thank you! I just didn't check carefully, didn't see this method. I'm using Micronaut, there is no support for OpenTelemetry yet, but I'm working on it right now, I will change to use `close` method instead.

PR #3273, opened 2021-06-03
Don't create Supplier when decoding byte

anuraaga:
I realized we should avoid this too given just how hot the code path is. Over many forks I can confirm there is a significant (not huge) change in the execution time, so I guess the GC profiler somehow doesn't keep track of lambda overhead.

I considered porting over Guava's `Preconditions` handling of `%s` but figured it's not worth it, the tried and trued if statement pattern doesn't seem like much of a regression to me.

After
```
Benchmark                                                    Mode  Cnt    Score    Error   Units
SpanIdBenchmark.getSpanIdBytes                               avgt   45   22.248 ±  0.098   ns/op
SpanIdBenchmark.getSpanIdBytes:·gc.alloc.rate                avgt   45  685.071 ±  3.113  MB/sec
SpanIdBenchmark.getSpanIdBytes:·gc.alloc.rate.norm           avgt   45   24.000 ±  0.001    B/op
SpanIdBenchmark.getSpanIdBytes:·gc.churn.G1_Eden_Space       avgt   45  682.211 ± 25.920  MB/sec
SpanIdBenchmark.getSpanIdBytes:·gc.churn.G1_Eden_Space.norm  avgt   45   23.901 ±  0.910    B/op
SpanIdBenchmark.getSpanIdBytes:·gc.churn.G1_Old_Gen          avgt   45    0.002 ±  0.001  MB/sec
SpanIdBenchmark.getSpanIdBytes:·gc.churn.G1_Old_Gen.norm     avgt   45   ≈ 10⁻⁴             B/op
SpanIdBenchmark.getSpanIdBytes:·gc.count                     avgt   45  287.000           counts
SpanIdBenchmark.getSpanIdBytes:·gc.time                      avgt   45  111.000               ms
```

Before
```
Benchmark                                                    Mode  Cnt    Score    Error   Units
SpanIdBenchmark.getSpanIdBytes                               avgt   45   26.344 ±  0.101   ns/op
SpanIdBenchmark.getSpanIdBytes:·gc.alloc.rate                avgt   45  578.526 ±  2.254  MB/sec
SpanIdBenchmark.getSpanIdBytes:·gc.alloc.rate.norm           avgt   45   24.000 ±  0.001    B/op
SpanIdBenchmark.getSpanIdBytes:·gc.churn.G1_Eden_Space       avgt   45  581.850 ± 25.282  MB/sec
SpanIdBenchmark.getSpanIdBytes:·gc.churn.G1_Eden_Space.norm  avgt   45   24.143 ±  1.074    B/op
SpanIdBenchmark.getSpanIdBytes:·gc.churn.G1_Old_Gen          avgt   45    0.002 ±  0.001  MB/sec
SpanIdBenchmark.getSpanIdBytes:·gc.churn.G1_Old_Gen.norm     avgt   45   ≈ 10⁻⁴             B/op
SpanIdBenchmark.getSpanIdBytes:·gc.count                     avgt   45  230.000           counts
SpanIdBenchmark.getSpanIdBytes:·gc.time                      avgt   45   94.000               ms
```

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/internal/OtelEncodingUtils.java
+++ b/api/all/src/main/java/io/opentelemetry/api/internal/OtelEncodingUtils.java
@@ -14,7 +14,7 @@ public final class OtelEncodingUtils {
   static final int BYTE_BASE16 = 2;
   static final int LONG_BASE16 = BYTE_BASE16 * LONG_BYTES;
   private static final String ALPHABET = "0123456789abcdef";
-  private static final int ASCII_CHARACTERS = 128;
+  private static final int NUM_ASCII_CHARACTERS = 128;
   private static final char[] ENCODING = buildEncodingArray();
   private static final byte[] DECODING = buildDecodingArray();
 
@@ -28,7 +28,7 @@ public final class OtelEncodingUtils {
   }
 
   private static byte[] buildDecodingArray() {
-    byte[] decoding = new byte[ASCII_CHARACTERS];
+    byte[] decoding = new byte[NUM_ASCII_CHARACTERS];
     Arrays.fill(decoding, (byte) -1);
     for (int i = 0; i < ALPHABET.length(); i++) {
       char c = ALPHABET.charAt(i);
@@ -110,10 +110,12 @@ public final class OtelEncodingUtils {
    * @return the resulting {@code byte}
    */
   public static byte byteFromBase16(char first, char second) {
-    Utils.checkArgument(
-        first < ASCII_CHARACTERS && DECODING[first] != -1, () -> "invalid character " + first);
-    Utils.checkArgument(
-        second < ASCII_CHARACTERS && DECODING[second] != -1, () -> "invalid character " + second);
+    if (first >= NUM_ASCII_CHARACTERS || DECODING[first] == -1) {
+      throw new IllegalArgumentException("invalid character " + first);
+    }
+    if (second >= NUM_ASCII_CHARACTERS || DECODING[second] == -1) {
+      throw new IllegalArgumentException("invalid character " + second);
+    }
     int decoded = DECODING[first] << 4 | DECODING[second];
     return (byte) decoded;
   }
--- a/api/all/src/jmh/java/io/opentelemetry/api/trace/SpanIdBenchmark.java
+++ b/api/all/src/jmh/java/io/opentelemetry/api/trace/SpanIdBenchmark.java
@@ -22,7 +22,7 @@ import org.openjdk.jmh.annotations.Warmup;
 @Warmup(iterations = 5, time = 1)
 @Measurement(iterations = 15, time = 1)
 @OutputTimeUnit(TimeUnit.NANOSECONDS)
-@Fork(1)
+@Fork(3)
 @Threads(1)
 public class SpanIdBenchmark {
 
--- a/api/all/src/main/java/io/opentelemetry/api/internal/Utils.java
+++ b/api/all/src/main/java/io/opentelemetry/api/internal/Utils.java
@@ -5,7 +5,6 @@
 
 package io.opentelemetry.api.internal;
 
-import java.util.function.Supplier;
 import javax.annotation.concurrent.Immutable;
 
 /** General internal utility methods. */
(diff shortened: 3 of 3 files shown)


jkwatson (approved):
I thought the lambda got turned into a static method in the bytecode. I don't think this makes much difference, but it also doesn't seem harmful.

anuraaga:
@jkwatson I think a lambda that references a variable, e.g., `first` can't be static since it has that state. Not sure about the GC numbers but good indeed to get some cycles back for each single ID byte

jkwatson:
It can if the field is a parameter to the static function.

api/all/src/main/java/io/opentelemetry/api/internal/OtelEncodingUtils.java
```diff
@@ -14,7 +14,7 @@
   static final int BYTE_BASE16 = 2;
   static final int LONG_BASE16 = BYTE_BASE16 * LONG_BYTES;
   private static final String ALPHABET = "0123456789abcdef";
-  private static final int ASCII_CHARACTERS = 128;
+  private static final int NUM_ASCII_CHARACTERS = 128;
```
kubawach, line 17: 👌

api/all/src/main/java/io/opentelemetry/api/internal/OtelEncodingUtils.java
```diff
@@ -110,10 +110,12 @@ public static void byteToBase16(byte value, char[] dest, int destOffset) {
    * @return the resulting {@code byte}
    */
   public static byte byteFromBase16(char first, char second) {
-    Utils.checkArgument(
-        first < ASCII_CHARACTERS && DECODING[first] != -1, () -> "invalid character " + first);
-    Utils.checkArgument(
-        second < ASCII_CHARACTERS && DECODING[second] != -1, () -> "invalid character " + second);
+    if (first >= NUM_ASCII_CHARACTERS || DECODING[first] == -1) {
```
kubawach, line 113: Simpler is better.

kubawach (approved):

jkwatson:
> It can if the value is a parameter to the static function.

Indeed, that's precisely how the bytecode looks: 

```
 // access flags 0x100A
  private static synthetic lambda$byteFromBase16$1(C)Ljava/lang/String;
   L0
    LINENUMBER 116 L0
    NEW java/lang/StringBuilder
    DUP
    INVOKESPECIAL java/lang/StringBuilder.<init> ()V
    LDC "invalid character "
    INVOKEVIRTUAL java/lang/StringBuilder.append (Ljava/lang/String;)Ljava/lang/StringBuilder;
    ILOAD 0
    INVOKEVIRTUAL java/lang/StringBuilder.append (C)Ljava/lang/StringBuilder;
    INVOKEVIRTUAL java/lang/StringBuilder.toString ()Ljava/lang/String;
    ARETURN
   L1
    LOCALVARIABLE second C L0 L1 0
    MAXSTACK = 2
    MAXLOCALS = 1

  // access flags 0x100A
  private static synthetic lambda$byteFromBase16$0(C)Ljava/lang/String;
   L0
    LINENUMBER 114 L0
    NEW java/lang/StringBuilder
    DUP
    INVOKESPECIAL java/lang/StringBuilder.<init> ()V
    LDC "invalid character "
    INVOKEVIRTUAL java/lang/StringBuilder.append (Ljava/lang/String;)Ljava/lang/StringBuilder;
    ILOAD 0
    INVOKEVIRTUAL java/lang/StringBuilder.append (C)Ljava/lang/StringBuilder;
    INVOKEVIRTUAL java/lang/StringBuilder.toString ()Ljava/lang/String;
    ARETURN
   L1
    LOCALVARIABLE first C L0 L1 0
    MAXSTACK = 2
    MAXLOCALS = 1
```

anuraaga:
Weird wonder where the value actually comes from - guess it must be able to reference outside stack frames never new that's possible.  Well, anyways seem to have pulled a few cycles from somewhere.

jkwatson:
> Weird wonder where the value actually comes from - guess it must be able to reference outside stack frames never new that's possible. Well, anyways seem to have pulled a few cycles from somewhere.

I personally don't understand how the callsite bytecode works in this case, so it might be in there.

PR #3275, opened 2021-06-03
Don't use AutoValue for TraceIdRatio sampler

anuraaga:
While examining this code, I found AutoValue seems to make it much more confusing than the code reduction benefit, it's just supposed to be one field, something our equals / hashcode can take advantage of to keep things small.


jkwatson (approved):
Seems fine to me!

PR #3282, opened 2021-06-03
Update the metrics autoconfigure code with support for duration units.

jkwatson:
Even though metrics aren't stable yet, it seems good to have some configuration consistency.

### DIFF
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
@@ -92,9 +92,9 @@ final class MetricExporterConfiguration {
         IntervalMetricReader.builder()
             .setMetricProducers(Collections.singleton(meterProvider))
             .setMetricExporter(exporter);
-    Duration exportIntervalMillis = config.getDuration("otel.imr.export.interval");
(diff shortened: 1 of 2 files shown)


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
```diff
@@ -92,9 +92,9 @@ private static void configureIntervalMetricReader(
         IntervalMetricReader.builder()
             .setMetricProducers(Collections.singleton(meterProvider))
             .setMetricExporter(exporter);
-    Long exportIntervalMillis = config.getLong("otel.imr.export.interval");
+    Duration exportIntervalMillis = config.getDuration("otel.imr.export.interval");
```
breedx-splk, line 95: probably rename that variable here too?

breedx-splk (changes requested):
Cool. I think that variable wants a little rename.

anuraaga (approved):
Just @breedx-splk's nit but LGTM

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
jkwatson: done

PR #3290, opened 2021-06-07
Use switch instead of enummap for status creation

anuraaga:
For a complete enum-switch, Java bytecode is a table lookup within the bytecode so always beats a memory lookup via EnumMap. More importantly, it's way less code

After (Ran benchmarks before setting to more appropriate AverageTime and too lazy to rerun for an obvious improvement, higher is better)
```
Benchmark                                                      Mode  Cnt   Score    Error   Units
StatusBenchmark.createWithoutDescription                      thrpt   30   0.384 ±  0.004  ops/ns
StatusBenchmark.createWithoutDescription:·gc.alloc.rate       thrpt   30  ≈ 10⁻⁴           MB/sec
StatusBenchmark.createWithoutDescription:·gc.alloc.rate.norm  thrpt   30  ≈ 10⁻⁶             B/op
StatusBenchmark.createWithoutDescription:·gc.count            thrpt   30     ≈ 0           counts
```

Before
```
Benchmark                                                      Mode  Cnt   Score    Error   Units
StatusBenchmark.createWithoutDescription                      thrpt   30   0.231 ±  0.003  ops/ns
StatusBenchmark.createWithoutDescription:·gc.alloc.rate       thrpt   30  ≈ 10⁻⁴           MB/sec
StatusBenchmark.createWithoutDescription:·gc.alloc.rate.norm  thrpt   30  ≈ 10⁻⁶             B/op
StatusBenchmark.createWithoutDescription:·gc.count            thrpt   30     ≈ 0           counts
```

### DIFF
--- a/sdk/trace/src/test/java/io/opentelemetry/sdk/trace/data/ImmutableStatusDataTest.java
+++ b/sdk/trace/src/test/java/io/opentelemetry/sdk/trace/data/ImmutableStatusDataTest.java
@@ -10,17 +10,26 @@ import static org.assertj.core.api.Assertions.assertThat;
 import io.opentelemetry.api.trace.StatusCode;
 import org.junit.jupiter.api.Test;
 
-/** Unit tests for {@link ImmutableStatusData}. */
 class ImmutableStatusDataTest {
   @Test
-  void defaultConstants() {
+  void statuses() {
     StatusCode[] codes = StatusCode.values();
-    assertThat(codes).hasSize(3);
-    assertThat(StatusData.unset().getStatusCode()).isEqualTo(StatusCode.UNSET);
-    assertThat(StatusData.unset().getDescription()).isEmpty();
-    assertThat(StatusData.ok().getStatusCode()).isEqualTo(StatusCode.OK);
-    assertThat(StatusData.ok().getDescription()).isEmpty();
-    assertThat(StatusData.error().getStatusCode()).isEqualTo(StatusCode.ERROR);
-    assertThat(StatusData.error().getDescription()).isEmpty();
+    for (StatusCode code : codes) {
+      StatusData status = ImmutableStatusData.create(code, "");
+      switch (code) {
+        case UNSET:
+          assertThat(status).isSameAs(StatusData.unset());
+          break;
+        case OK:
+          assertThat(status).isSameAs(StatusData.ok());
+          break;
+        case ERROR:
+          assertThat(status).isSameAs(StatusData.error());
+          break;
+      }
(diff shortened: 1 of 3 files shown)


sdk/trace/src/test/java/io/opentelemetry/sdk/trace/data/ImmutableStatusDataTest.java
```diff
@@ -23,15 +23,4 @@ void defaultConstants() {
     assertThat(StatusData.error().getStatusCode()).isEqualTo(StatusCode.ERROR);
     assertThat(StatusData.error().getDescription()).isEmpty();
   }
-
-  @Test
-  void generateCodeToStatus() {
```
anuraaga, line 28: We don't need this since errorprone will make sure the compile fails if a new enum value is added.

trask (approved):

kubawach (approved):
Wow, that's a huge improvement (over 66%) and an important lesson WRT Java performance.  Intuitively map is faster, but not with enums! :)

sdk/trace/src/test/java/io/opentelemetry/sdk/trace/data/ImmutableStatusDataTest.java
kubawach: Frankly I'd prefer to rely on our own tests rather than on a build tool configuration. But perhaps I'm too defensive :)

jkwatson: alternatively, we could *only* have this test method, which covers everything in the method above and wouldn't require adding more entries if we add another status.

jkwatson:
> Wow, that's a huge improvement (over 66%) and an important lesson WRT Java performance. Intuitively map is faster, but not with enums! :)

"huge". relatively, perhaps, but we're talking about < 10ns either way.

jkwatson (approved):

kubawach:
> > Wow, that's a huge improvement (over 66%) and an important lesson WRT Java performance. Intuitively map is faster, but not with enums! :)
> 
> "huge". relatively, perhaps, but we're talking about < 10ns either way.

Definitely, that's why I noted percentage :)

sdk/trace/src/test/java/io/opentelemetry/sdk/trace/data/ImmutableStatusDataTest.java
anuraaga: Sure combined into one test

PR #3291, opened 2021-06-07
Only check one part of trace ID is valid since it guarantees a value …

anuraaga:
…full ID.

Just small code reduction

### DIFF
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/AndroidFriendlyRandomIdGenerator.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/AndroidFriendlyRandomIdGenerator.java
@@ -31,11 +31,11 @@ enum AndroidFriendlyRandomIdGenerator implements IdGenerator {
 
   @Override
   public String generateTraceId() {
-    long idHi;
+    long idHi = random.nextLong();
+    long idLo;
     do {
-      idHi = random.nextLong();
-    } while (idHi == INVALID_ID);
-    long idLo = random.nextLong();
+      idLo = random.nextLong();
+    } while (idLo == INVALID_ID);
     return TraceId.fromLongs(idHi, idLo);
   }
 }
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/AndroidFriendlyRandomIdGenerator.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/AndroidFriendlyRandomIdGenerator.java
@@ -32,11 +32,10 @@ enum AndroidFriendlyRandomIdGenerator implements IdGenerator {
   @Override
   public String generateTraceId() {
     long idHi;
-    long idLo;
     do {
       idHi = random.nextLong();
-      idLo = random.nextLong();
-    } while (idHi == INVALID_ID && idLo == INVALID_ID);
+    } while (idHi == INVALID_ID);
+    long idLo = random.nextLong();
     return TraceId.fromLongs(idHi, idLo);
   }
 }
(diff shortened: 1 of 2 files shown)


trask (approved):

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RandomIdGenerator.java
```diff
@@ -27,12 +27,11 @@ public String generateSpanId() {
...
     long idHi;
-    long idLo;
     ThreadLocalRandom random = ThreadLocalRandom.current();
     do {
       idHi = random.nextLong();
-      idLo = random.nextLong();
-    } while (idHi == INVALID_ID && idLo == INVALID_ID);
+    } while (idHi == INVALID_ID);
+    long idLo = random.nextLong();
     return TraceId.fromLongs(idHi, idLo);
```
kubawach, line 35: Can be compacted even more now by TraceId.fromLongs(idHi, random.nextLong()), although this negatively impacts readability (same for android).

kubawach (approved):

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RandomIdGenerator.java
anuraaga: Hehe - yeah I'm mostly interested in "real code" (unnecessary check) vs "lines of code" (local variable name that provides some good semantic context too).

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/AndroidFriendlyRandomIdGenerator.java
```diff
@@ -32,11 +32,10 @@ public String generateSpanId() {
   @Override
   public String generateTraceId() {
     long idHi;
-    long idLo;
     do {
       idHi = random.nextLong();
-      idLo = random.nextLong();
-    } while (idHi == INVALID_ID && idLo == INVALID_ID);
+    } while (idHi == INVALID_ID);
```
jkwatson, line 37: Is this correct? I thought for 64-bit-traceid interop, that we required the low bytes to be non-zero in all cases, rather than the high bytes.

Oberon00:
I think this may have bad interactions with the TraceIdRatioBasedSampler. At lower sampler rates, you might have 0% chance of getting sampled with this change.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/AndroidFriendlyRandomIdGenerator.java
anuraaga: This makes sense to me - can't find anything in the spec but will check on that.

anuraaga:
@Oberon00 Currently this PR doesn't change the high part at all so I don't think it would interact. But I'm thinking of swapping high and lo - it would make the high part truly random, instead of random except missing the 0 value. Isn't it (theoretically but not a big deal in practice) more precise?

jkwatson:
> @Oberon00 Currently this PR doesn't change the high part at all so I don't think it would interact. But I'm thinking of swapping high and lo - it would make the high part truly random, instead of random except missing the 0 value. Isn't it (theoretically but not a big deal in practice) more precise?

If I read the `TraceIdRatioBasedSampler` code correctly, it uses the low bytes, not the high bytes for the sampling decision.

anuraaga:
Oops I read the `16` as a radix, not offset :) So switching to idLo check means this PR shouldn't affect that sampler for better or worse.

jkwatson (approved):

PR #3292, opened 2021-06-07
Small tweak to name of internal method.

anuraaga:
When reading code, randomly found this method which was a bit confusing since it doesn't always copy, only when it is ensuring a limit.


trask (approved):
another similar option: `applyAttributesLimit`

kubawach (approved):
I'd second Trask's proposal. "ensure" suggests more of a validation, wile "apply" means exactly what the method does.

jkwatson (approved):
I'm fine with either naming tweak, since this is a non-public method, either way.

PR #3293, opened 2021-06-07
Fix drift in SpanLimits comments.

anuraaga:


kubawach (approved):

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SpanLimitsBuilder.java
```diff
@@ -10,6 +10,9 @@
 
 /** Builder for {@link SpanLimits}. */
 public final class SpanLimitsBuilder {
+
+  // These values are the default values for all the limits.
```
jkwatson, line 14: does this even add any value since the fields have the word DEFAULT in them already?

jkwatson (approved):

PR #3294, opened 2021-06-07
Use autoconfigure SPI for trace-propagators extension.

anuraaga:
While I'm not exactly sure what @kenfinnigan will do with them, if there's a use case I don't see anything wrong with this since the autoconfigure dep isn't leaked at all.

https://github.com/open-telemetry/opentelemetry-java/discussions/3284

### DIFF
--- a/extensions/trace-propagators/build.gradle.kts
+++ b/extensions/trace-propagators/build.gradle.kts
@@ -12,6 +12,8 @@ extra["moduleName"] = "io.opentelemetry.extension.trace.propagation"
 dependencies {
     api(project(":api:all"))
 
+    compileOnly(project(":sdk-extensions:autoconfigure"))
+
     testImplementation("io.jaegertracing:jaeger-client")
     testImplementation("com.google.guava:guava")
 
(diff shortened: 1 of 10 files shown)


kubawach (approved):
Nice. Less "manual" configuration code is always better.

kenfinnigan:
Thanks @anuraaga, looks great.

The only comment is based on what you asked in my original question, about whether the SPI should be a separate artifact. Ideally, it would be separate, but it's not a showstopper.

anuraaga:
Thanks @kenfinnigan I think splitting SPI makes sense and is consistent with auto instrumentation too. Will look at it in another PR.

jkwatson (approved):
This seems fine, especially since it doesn't impact using the propagators directly without autoconfigure around.

jkwatson:
> Nice. Less "manual" configuration code is always better.

As long as we don't lose sight of the fact that there are 100% manual configuration use-cases that we must always support as well.

kenfinnigan (approved):
Thanks @anuraaga!

jkwatson:
@anuraaga looks like this didn't include the apidiffs changes introduced by this. I'll add it to my CHANGELOG updates before release.

PR #3298, opened 2021-06-08
Update the semconv to spec v1.4.0

jkwatson:


anuraaga:
@jkwatson Any thoughts on releasing SDK 1.3.0 before this or doesn't matter at all?

jkwatson:
> @jkwatson Any thoughts on releasing SDK 1.3.0 before this or doesn't matter at all? Since we're planning to anyways, it allows the versions to be aligned for a bit more.

I'm fine either way. We're always going to be at least one minor version behind the spec (and even if we merge this, there is still the schema stuff to get in).

anuraaga (approved):

kubawach (approved):

arminru (approved):

semconv/src/main/java/io/opentelemetry/semconv/trace/attributes/SemanticAttributes.java
```diff
@@ -638,6 +638,34 @@
...
+   * specify this, the value can be omitted.
+   */
+  public static final AttributeKey<String> RPC_JSONRPC_VERSION = stringKey("rpc.jsonrpc.version");
+
+  /**
+   * `method` property from request. Unlike `rpc.method`, this may not relate to the actual method
+   * being called. Useful for client-side traces since client does not know what will be called on
+   * the server.
+   */
+  public static final AttributeKey<String> RPC_JSONRPC_METHOD = stringKey("rpc.jsonrpc.method");
```
Oberon00, line 652: That new attribute is a bit unfortunate, I opened a spec PR: open-telemetry/opentelemetry-specification#1748 to remove it / clarify that rpc.method can be used instead.

Oberon00 (approved):

semconv/src/main/java/io/opentelemetry/semconv/trace/attributes/SemanticAttributes.java
jkwatson: @Oberon00 do you think we should just wait for that to settle before merging this, then?

Oberon00: IMHO it's better to have this one bad attribute than not to update at all.

jkwatson: I just cut the 1.3.0 release, so we'll get what we get in the next version. :)

jkwatson:
@anuraaga @Oberon00 I amended this PR to also populate a new field for the schema URL in each of the files. Please be sure to review again if you have feelings.

anuraaga (approved):

PR #3304, opened 2021-06-09
Separate out SPI for autoconfigure.

anuraaga:
Allows custom autoconfigure implementations like Quarkus


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/DefaultConfigProperties.java
```diff
@@ -84,6 +88,7 @@ public Integer getInt(String name) {
    * @return null if the property has not been configured.
    * @throws NumberFormatException if the property is not a valid long.
```
kubawach, line 89: @throws ConfigurationException if the property is not a valid long.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/DefaultConfigProperties.java
```diff
@@ -104,6 +109,7 @@ public Long getLong(String name) {
    * @return null if the property has not been configured.
    * @throws NumberFormatException if the property is not a valid double.
```
kubawach, line 110: @throws ConfigurationException if the property is not a valid double.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/DefaultConfigProperties.java
```diff
@@ -24,20 +25,21 @@
  * for both of these will be normalized to be all lower case, and underscores will be replaced with
  * periods.
  */
-public final class ConfigProperties {
+final class DefaultConfigProperties implements ConfigProperties {
```
kubawach, line 28: As in all exceptional cases this class throws standardised ConfigurationException - perhaps it would make sense to set this as a part of contract and move to ConfigProperties interface? Then of course all other implementation would be required to throw this exception in case of non-recoverable issues.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
```diff
@@ -70,7 +71,7 @@ private static void configureMeterProvider(Resource resource, ConfigProperties c
   }
 
   private static Resource buildResource() {
-    ConfigProperties config = ConfigProperties.get();
+    DefaultConfigProperties config = DefaultConfigProperties.get();
```
kubawach, line 74: Set ConfigProperties as the type here?

kubawach (approved):

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/DefaultConfigProperties.java
anuraaga: Yeah that was the other option, and really then we may as well move the class into the SPI instead of separating an interface for `ConfigProperties`. @kenfinnigan would you want to use this default configproperties or implement it yourself? If latter, I think not pushing the exception contract to SPI makes sense - the exception is used by our autoconfiguration but doesn't have to be used by other mechanisms.

kenfinnigan: @anuraaga Right now I think I'd be ok with using `ConfigProperties` directly, but I haven't directly played with it to know whether there would be issues I'm not thinking of

anuraaga: I think I'm leaning towards what we have now - an interface is an interface, even SPI. I don't think we should have any implementation in it.

sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/ConfigProperties.java
```diff
@@ -0,0 +1,85 @@
...
+
+package io.opentelemetry.sdk.autoconfigure.spi;
+
+import java.time.Duration;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/** Properties used for auto-configuration of the OpenTelemetry SDK components. */
+public interface ConfigProperties {
```
jkwatson, line 14: It doesn't feel to me like this belongs in an SPI package. This interface is useful for non-SPI-based configuration as well, isn't it? Are we really thinking that we should restrict this to being SPI only?

jkwatson:
@kenfinnigan can you explain how breaking out the SPI interfaces into a separate module will help? It seems like Quarkus can do whatever it wants and not use the rest of autoconfigure if it doesn't make sense. I'm confused about the need for this, and would prefer not to have to introduce yet another module if we don't have to.

kenfinnigan:
@jkwatson it assists the dead code elimination process by not having the classes in a dependency at all to begin with, as opposed to relying on Quarkus build or GraalVM native image process to recognize that classes aren't actually used and can be removed.

jkwatson:
> @jkwatson it assists the dead code elimination process by not having the classes in a dependency at all to begin with, as opposed to relying on Quarkus build or GraalVM native image process to recognize that classes aren't actually used and can be removed.

Hmm. ok, I understand that. I'm not sure it's enough of a justification to add another module, especially if we have to do something odd like move `ConfigProperties` out of the autoconfigure module itself, where it really belongs. How much pain do you incur by having to remove unused classes? That seems like a pretty easy build-time thing that you have to do anyway, isn't it? (he says naively, knowing exactly nothing about Quarkus/GraalVM build processes)

kenfinnigan:
In JVM mode there's probably not a lot we can do to actually remove the unused classes from the module, without shading contents of the dependency to do so. For GraalVM it increases the image build time as more classes means more execution paths to identify and determine what is needed for them.

Another concern I just realized is if the SPI is part of the autoconfigure module implementation, it's easy for a user to mistakenly think all the configuration settings provided with autoconfigure are available. In this situation, it would no longer be possible to dead code eliminate the classes as they'd be utilized, and whatever config they'd still will likely not be applied resulting in a confusing experience

jkwatson:
Are you planning on using the same implementations of the SPIs, like for resources, propagators, exporters, etc, just not use the auto-configure code to wire them up? So, like an alternate implementation of the auto-configure module that's narrowly oriented toward Quarkus/GraalVM? Is that something that would make sense to have in this repository as well, or is too specific for that use-case?

kenfinnigan:
I think so, but there could also be instances of resources, propagators, and exporters that are loaded based on CDI beans, and not just SPI implemented ones. But yes, it wouldn't use the autoconfigure implementation to wire them up.

I think it would be too specific to Quarkus as it relies on a mixture of what we call "build steps", which are run during the build, and "recorders", which generate bytecode.

sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/ConfigProperties.java
kenfinnigan: An alternative would be for the SPI classes to not take `ConfigProperties` as an object for any creation methods and instead have the implementation retrieve the config from `ConfigProperties` with `get()`.

`ConfigProperties` could also have an INSTANCE that is stored on "first use". I saw many usages of `ConfigProperties.get()` which essentially reloads the object and properties, but I'm not sure the benefits of the multiple creations.

jkwatson: @anuraaga thoughts?

mateuszrzeszutek:
This is partially related to this PR, but in the javaagent we have this hack to copy properties from additional property sources (SPI, config file) so that sdk-autoconfigure sees them: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java#L38

It'd be great if `ConfigProperties` (or whatever we end up with) was able to consume those additional properties directly - by making it an SPI, adding a `get(Map<String, String)` method or something else. Not necessarily in this PR, of course; I just wanted this one weird scenario not to get forgotten 😄

jkwatson:
> This is partially related to this PR, but in the javaagent we have this hack to copy properties from additional property sources (SPI, config file) so that sdk-autoconfigure sees them: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java#L38
> 
> It'd be great if `ConfigProperties` (or whatever we end up with) was able to consume those additional properties directly - by making it an SPI, adding a `get(Map<String, String)` method or something else. Not necessarily in this PR, of course; I just wanted this one weird scenario not to get forgotten 😄

I'm unclear on what feature you're looking for here. Can you create an issue to track it?

jkwatson:
I've been mulling over the move of the ConfigProperties interface to the new SPI module, and I still really don't like it. It doesn't feel like the right spot for that interface to live. Unfortunately, I don't know of a better place for it at the moment. We could put it into the SDK-common module, but we've been trying to keep that sort of configuration stuff out of there so far. The other option, to create yet another module to hold it and the implementation feels even more heavy-weight.

Any other ideas?

sdk-extensions/autoconfigure-spi/build.gradle.kts
```diff
@@ -0,0 +1,12 @@
+plugins {
+    `java-library`
+    `maven-publish`
+}
+
+description = "OpenTelemetry SDK Auto-configuration SPI"
```
jkwatson, line 6: this module needs a `gradle.properties` to mark it as `alpha`

kenfinnigan:
Other than what I suggested yesterday I don't have other ideas, but we need to hear what @anuraaga thinks

jkwatson:
> Other than what I suggested yesterday I don't have other ideas, but we need to hear what @anuraaga thinks

Yeah, I don't particularly like the hidden aspect of having all implementations have to know to call ConfigProperties.get() to get the normalized configuration data. I'm much more in favor of injecting them into the SPI interface method.

iNikem:
I admit I did not follow this discussion a lot, but after a quick glance it seems to me that instrumentation repo has solved this problem already, haven't we? We have `io.opentelemetry.instrumentation.api.config.Config` class that everybody can and should use and we have `io.opentelemetry.javaagent.spi.config.PropertySource` SPI to contribute to that central `Config` instance. Isn't this what this PR is about?

jkwatson:
> I admit I did not follow this discussion a lot, but after a quick glance it seems to me that instrumentation repo has solved this problem already, haven't we? We have `io.opentelemetry.instrumentation.api.config.Config` class that everybody can and should use and we have `io.opentelemetry.javaagent.spi.config.PropertySource` SPI to contribute to that central `Config` instance. Isn't this what this PR is about?

This PR is about making SPI a stand-alone module to separate it from the user of the SPI (this project's auto-configurtion, or some other usage in Quarkus or Graalvm). The configuration piece just ended up being a bit of a sticking point in that process.

PR #3307, opened 2021-06-10
Make sure the 2nd call to shutdown() on the GRPC exporters succeeds.

### LINKED ISSUE #3306: OtlpGrpcSpanExporter's shutdown can hang forever if the Channel is already shut down
**Describe the bug**
`OtlpGrpcSpanExporter`'s shutdown can hang forever if the underlying `Channel` is already shut down

**Steps to reproduce**
1. Create an `OtlpGrpcSpanExporter`    
2. Create a `BatchSpanProcessor` and feed in the exporter    
3. Create an `SdkTracer` and feed in the processor    
4. Create an `OpenTelemetrySdk` and set the tracer    
5. Report some spans (optional)    
6. Call `shutdown()` on the `SdkTracer` and wait for the callback    
7. Call `shutdown()` on the `BatchSpanProcessor` and wait for the callback    
8. Call `shutdown()` on the `OtlpGrpcSpanExporter` and wait for the callback    
9. Observe that the callback never completes

I expected the callback from `OtlpGrpcSpanExporter`'s shutdown to complete

**What did you see instead?**
The program hung indefinitely

```scala
libraryDependencies ++= Seq(
  "io.opentelemetry"  % "opentelemetry-exporter-otlp" % "1.2.0",
  "io.grpc" % "grpc-okhttp" % "1.38.0",
   // "io.grpc"           % "grpc-netty"                  % "1.38.0" // I also tried this
  "io.opentelemetry" % "opentelemetry-api" % "1.2.0",
  "io.opentelemetry" % "opentelemetry-sdk" % "1.2.0",
)
```

**Environment**
Compiler: Scala 2.13.4 and AdoptOpenJDK 11.0.3
OS: NixOS 20.09
Runtime: same JDK as above
Runtime OS: same as above

**Additional context**
I will follow up with a java example that demonstrates this

jkwatson:
Resolves #3306

### DIFF
--- a/exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporter.java
+++ b/exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporter.java
@@ -179,11 +179,11 @@ public final class OtlpGrpcSpanExporter implements SpanExporter {
    */
   @Override
   public CompletableResultCode shutdown() {
-    if (managedChannel.isShutdown()) {
-      return CompletableResultCode.ofSuccess();
-    }
     final CompletableResultCode result = new CompletableResultCode();
     managedChannel.notifyWhenStateChanged(ConnectivityState.SHUTDOWN, result::succeed);
+    if (managedChannel.isShutdown()) {
+      return result.succeed();
+    }
     managedChannel.shutdown();
     this.spansSeen.unbind();
     this.spansExportedSuccess.unbind();
--- a/exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporter.java
+++ b/exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporter.java
@@ -179,6 +179,9 @@ public final class OtlpGrpcSpanExporter implements SpanExporter {
    */
   @Override
   public CompletableResultCode shutdown() {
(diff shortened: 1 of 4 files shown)


exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporter.java
```diff
@@ -179,6 +179,9 @@ public static OtlpGrpcSpanExporter getDefault() {
    */
   @Override
   public CompletableResultCode shutdown() {
+    if (managedChannel.isShutdown()) {
+      return CompletableResultCode.ofSuccess();
+    }
```
jkwatson, line 184: There's definitely still a race condition in here, where the shutdown can complete in between this check and the addition of the 2nd notification listener. I'm not sure if there's a solution to that or not.

andimiller: can you add the listener first and do it like this?
```java
managedChannel.notifyWhenStateChanged(ConnectivityState.SHUTDOWN, result::succeed);
if (managedChannel.isShutdown()) {
  return result.succeed();
}
managedChannel.shutdown();
```

jkwatson: That does seem tighter. Depending on the grpc internals, there might still be a race, but this change will definitely make it less likely. Thanks for the suggestion!

anuraaga (approved):

PR #3299, opened 2021-06-08
Add the OpenTelemetry schema URL to the InstrumentationLibraryInfo

jkwatson:
and to the corresponding API calls.

PR #3309, opened 2021-06-11
Introduce usage of the OpenTelemetry schema with a Tracer/MeterBuilder

jkwatson:
This is an alternative approach to #3299 utilizing a builder for tracers/meters, rather than adding another overload with a confusing nullable parameter.


anuraaga (approved):
Lots of code, but the end result API seems cleaner than the other

jkwatson:
ok, let's go with this one, then. I think it's a little better, as well.

PR #3312, opened 2021-06-14
Reorder methods on SpanData for better AutoValue toString

anuraaga:
We commonly use the toString generated from the AutoValue TestData in test assertions, which look terrible because of the huge Resource that gets printed before some important stuff

### DIFF
--- a/docs/apidiffs/current_vs_latest/opentelemetry-extension-annotations.txt
+++ b/docs/apidiffs/current_vs_latest/opentelemetry-extension-annotations.txt
@@ -1,2 +1,10 @@
 Comparing source compatibility of  against 
-No changes.
++++  NEW ANNOTATION: PUBLIC(+) ABSTRACT(+) io.opentelemetry.extension.annotations.SpanAttribute  (not serializable)
+	+++  CLASS FILE FORMAT VERSION: 52.0 <- n.a.
+	+++  NEW INTERFACE: java.lang.annotation.Annotation
+	+++  NEW SUPERCLASS: java.lang.Object
+	+++  NEW METHOD: PUBLIC(+) ABSTRACT(+) java.lang.String value()
+	+++  NEW ANNOTATION: java.lang.annotation.Target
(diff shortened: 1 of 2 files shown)


kubawach (approved):

mateuszrzeszutek (commented):
❤️  the idea, scrolling over a huuuge resource can be annoying.

Isn't `toString()` implemented in `SpanWrapper` though?

docs/apidiffs/current_vs_latest/opentelemetry-extension-annotations.txt
```diff
@@ -1,2 +1,10 @@
 Comparing source compatibility of  against 
-No changes.
\ No newline at end of file
++++  NEW ANNOTATION: PUBLIC(+) ABSTRACT(+) io.opentelemetry.extension.annotations.SpanAttribute  (not serializable)
```
Oberon00, line 2: What's with the changes in these files? I'm getting the same diff locally if I do a totally unrelated change (presumably also if I just rebuild).

Probably this needs to go into a separate PR and was missing from some other PR.

anuraaga:
I guess most sdk-testing users are fine indeed - but we do provide TestSpanData here which has the problem. And we use it in the agent as you've probably found annoying :)

docs/apidiffs/current_vs_latest/opentelemetry-extension-annotations.txt
anuraaga: Yeah I've been noticing this. I think the two reasonable options are

1) Fail PR build if there are API diffs
2) Move diff computation and creation of a PR to a nightly job and remove it from normal `check`.

I like 2) better, but it requires a PAT which still can't be scoped 😿

I don't think requiring contributors to split their PRs or deal with merge conflicts because the tooling isn't ideal yet is a good idea though so wouldn't worry too much about it until doing one of those.

jkwatson (approved):

PR #3313, opened 2021-06-14
Fix missing apidiff from #3168.

Oberon00:
#3168 added a new annotation but the apidiff update is missing. As a result, the diff crops up in other PRs, e.g. https://github.com/open-telemetry/opentelemetry-java/pull/3312/files#diff-619e30119c066f6bb370538e7126a2c0acaa4316bafc55441dccaf3e3e438182


jkwatson (approved):
Thanks!

PR #3314, opened 2021-06-14
SdkSpanBuilder: Use Context.root() to simplify logic.

Oberon00:
I could also remove the now useless `setNoParent().setParent()` sequences from the unit test if you want.


kubawach (approved):
Nice, parentContext logic in startSpan is much clearer now.

Oberon00:
Approximately the same logic is already used in [`DefaultTracer.NoopSpanBuilder`](https://github.com/open-telemetry/opentelemetry-java/blob/v1.3.0/api/all/src/main/java/io/opentelemetry/api/trace/DefaultTracer.java#L60) by the way.

trask (approved):

jkwatson (approved):
seems ok to me

carlosalberto (approved):

anuraaga (approved):

PR #3316, opened 2021-06-15
Extract java conventions plugin.

anuraaga:
There is a relatively recent Gradle idiom to write shared build logic as convention plugins

https://docs.gradle.org/current/samples/sample_convention_plugins.html

This allows us to split our current giant build.gradle into separate plugins for each type of configuration. I will follow up with moving / extracting more files. There is another great improvement in being able to use the native DSL instead of things like `configure<>` in these plugins.

### DIFF
--- a/all/build.gradle.kts
+++ b/all/build.gradle.kts
@@ -1,5 +1,5 @@
 plugins {
-    java
+    id("otel.java-conventions")
 }
 
 description = "OpenTelemetry All"
--- a/api/all/build.gradle.kts
+++ b/api/all/build.gradle.kts
@@ -1,5 +1,5 @@
 plugins {
-    id("java-library")
+    id("otel.java-conventions")
     id("maven-publish")
 
     id("me.champeau.jmh")
--- a/api/metrics/build.gradle.kts
+++ b/api/metrics/build.gradle.kts
@@ -1,5 +1,5 @@
 plugins {
-    id("java-library")
+    id("otel.java-conventions")
     id("maven-publish")
 
     id("me.champeau.jmh")
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -25,12 +25,9 @@ plugins {
     id("de.marcphilipp.nexus-publish") apply false
     id("io.morethan.jmhreport") apply false
     id("me.champeau.jmh") apply false
-    id("net.ltgt.errorprone") apply false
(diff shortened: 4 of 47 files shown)


trask (approved):

kubawach (approved):

Oberon00:
I don't really see the advantage right now. git says `418 insertions(+), 385 deletions(-)`, so the code got a bit longer. Also, it introduces the new concept of `buildSrc` that needs to be understood to understand the project build. Since you say it is a *recent* idiom, I also don't think many people will recognize the pattern right away.

I noticed the instrumentation repo used `apply from` instead of a plugin to similarly make the root build.gradle smaller. That seems to be an easier to understand pattern to me.

> I will follow up with moving / extracting more files. 

The build.gradle is already cut down from 537 to 234 lines, so I wonder how much more could be extracted.

> There is another great improvement in being able to use the native DSL instead of things like configure<> in these plugins.

What does this mean? Is this used here yet? I see `withType<Javadoc>().configureEach` and similar in the plugin source code.

anuraaga:
> so the code got a bit longer

It's almost all the new build file, which will get amortized over extracting more script plugins.

> I noticed the instrumentation repo used apply from instead of a plugin to similarly make the root build.gradle smaller.

Unfortunately, this just does not work with Kotlin - scripts loaded like that aren't plugins and have no context, meaning there's no `project`, etc or the DSL introduced by applied plugins. I had resigned myself to our huge `build.gradle.kts` but found this pattern to make things smaller, which I know @jkwatson wanted too.

`buildSrc` itself is quite old, and while yes it's a newish idiom, it works well in allowing us to split this build file up and I think it's worth it.

> What does this mean? Is this used here yet? I see withType<Javadoc>().configureEach and similar in the plugin source code.

Yeah, things like `configure<BasePluginConvention>` are now `base`.

anuraaga:
> The build.gradle is already cut down from 537 to 234 lines, so I wonder how much more could be extracted.

I am going to enjoy moving all that errorprone stuff into a separate file too

jkwatson:
I absolutely am in favor of componentizing the build script, so if this is a step toward that, I approve whole-heartedly!

jkwatson (approved):
Seems like a good step!

Oberon00 (approved):

carlosalberto (approved):

PR #3317, opened 2021-06-16
Define an extension for otel-specific configuration of Java.

anuraaga:
Provides compile time safety instead of using the String, and allows lazy evaluation by using `Property<String>` so no need to worry about evaluation order

### DIFF
--- a/all/build.gradle.kts
+++ b/all/build.gradle.kts
@@ -3,7 +3,7 @@ plugins {
 }
 
 description = "OpenTelemetry All"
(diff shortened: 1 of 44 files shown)


jkwatson (approved):
If you don't set the module name for a given module, does the build fail?

anuraaga:
Currently it does, for any project - I'm thinking of moving the logic for it from this file to publish configuration so at least non published projects don't need to set it though.

PR #3318, opened 2021-06-16
Split out errorprone and jacoco build scripts.

anuraaga:


jkwatson (approved):
looks like a solid refactoring to me!

PR #3319, opened 2021-06-16
Add the OpenTelemetry schema URL to the Resource.

jkwatson:


exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MetricAdapter.java
```diff
@@ -60,11 +60,15 @@
...
       }
-      resourceMetrics.add(
+      ResourceMetrics.Builder resourceMetricsBuilder =
           ResourceMetrics.newBuilder()
               .setResource(ResourceAdapter.toProtoResource(entryResource.getKey()))
-              .addAllInstrumentationLibraryMetrics(instrumentationLibraryMetrics)
-              .build());
+              .addAllInstrumentationLibraryMetrics(instrumentationLibraryMetrics);
+      String schemaUrl = entryResource.getKey().getSchemaUrl();
+      if (schemaUrl != null) {
```
kubawach, line 68: I'd extract the builder creation to a separate helper method to keep the fluent builder flow.

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -135,7 +155,19 @@ public Resource merge(@Nullable Resource other) {
...
     attrBuilder.putAll(other.getAttributes());
-    return new AutoValue_Resource(attrBuilder.build());
+
+    if (other.getSchemaUrl() == null) {
+      return create(attrBuilder.build(), getSchemaUrl());
+    }
+    if (getSchemaUrl() == null) {
+      return create(attrBuilder.build(), other.getSchemaUrl());
+    }
+    if (!Objects.equals(other.getSchemaUrl(), getSchemaUrl())) {
```
kubawach, line 165: Perhaps we should log a warning here? Users can be surprised by missing schema knowing it was set before the merge.

kubawach (approved):

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
anuraaga: We can use normal `String.equals` since we checked both for nulls

sdk/common/src/main/java/io/opentelemetry/sdk/resources/ResourceBuilder.java
```diff
@@ -18,6 +18,7 @@
 public class ResourceBuilder {
 
   private final AttributesBuilder attributesBuilder = Attributes.builder();
+  private String schemaUrl;
```
anuraaga, line 21: `@Nullable`

anuraaga (approved):

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
jkwatson: I think it's reasonable to log here, until we have a strategy for merging them, yes. I'll add it.

sdk/common/src/main/java/io/opentelemetry/sdk/resources/ResourceBuilder.java
jkwatson: done

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
jkwatson: done

PR #3320, opened 2021-06-16
Support retrieving container.id for crio and docker environments

lo-jason:


sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,86 @@
...
+  /** Returns resource with container information. */
+  public static Resource get() {
+    ContainerResource factory = new ContainerResource(UNIQUE_HOST_NAME_FILE_NAME);
+    String containerId = factory.extractContainerId();
+
+    if (containerId == null) {
+      return Resource.empty();
+    }
+    return Resource.create(
+        Attributes.builder().put(ResourceAttributes.CONTAINER_ID, containerId).build());
```
jkwatson, line 43: this can be simplified to just `Attributes.of(...)`

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,86 @@
...
+              if (matcher.matches() && matcher.group(2) != null && !matcher.group(2).isEmpty()) {
+                return matcher.group(2);
+              }
+            }
+          }
+          return null;
+        }
+      } catch (FileNotFoundException e) {
+        return null;
+      } catch (IOException e) {
```
jkwatson, line 80: these two exceptions can be collapsed into one catch block, yes? Isn't `FileNotFoundException` already an `IOException`?

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,86 @@
...
+   * <p>We see this with CRI-O "crio-abcdef1234567890ABCDEF.freetext", then use {@linkplain
+   * ContainerResource#HEX_EXTRACTOR} to extract the container hex id
+   *
+   * @return containerId
+   */
+  @Nullable
+  @SuppressWarnings("DefaultCharset")
+  public String extractContainerId() {
+    File nameFile = new File(cgroupFilePath);
+    if (nameFile.exists() && nameFile.canRead()) {
```
jkwatson, line 60: if you flip this `if` to the opposite, you can return early and get rid of some of the nesting in here.

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,86 @@
...
+      try (FileReader fr = new FileReader(nameFile);
+          BufferedReader br = new BufferedReader(fr); ) {
+        String line;
+        while ((line = br.readLine()) != null) {
+          if (!line.isEmpty()) {
```
jkwatson, line 65: same with switching the logic here to return early and remove some levels of nesting.

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,86 @@
...
+            // This cgroup output line should have the container id in it
+            String[] sections = line.split(File.separator);
+            if (sections.length > 1) {
```
jkwatson, line 68: I think you can also do it here. Less nesting makes the code much easier to read. thanks!

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,86 @@
...
+   * ContainerResource#HEX_EXTRACTOR} to extract the container hex id
+   *
+   * @return containerId
+   */
+  @Nullable
+  @SuppressWarnings("DefaultCharset")
+  public String extractContainerId() {
+    File nameFile = new File(cgroupFilePath);
+    if (nameFile.exists() && nameFile.canRead()) {
+      try (FileReader fr = new FileReader(nameFile);
```
jkwatson, line 61: you don't need to have both of these variables separately..you can just call
`BufferedReader reader = new BufferedReader(new FileReader(nameFile))`

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,86 @@
...
+          BufferedReader br = new BufferedReader(fr); ) {
```
jkwatson, line 62: would prefer to have this variable named "reader" rather than abbreviating it.

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
lo-jason: done

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
lo-jason: done

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
lo-jason: done

PR #3321, opened 2021-06-16
#3308 Support adding container.id to resource metadata

lo-jason:
#3308


jkwatson:
I put a bunch of comments on the PR you just closed:  https://github.com/open-telemetry/opentelemetry-java/pull/3320

lo-jason:
> I put a bunch of comments on the PR you just closed: #3320

@jkwatson thank you for the comments. I have made the changes suggested.
Sorry about the duplicate PR i had the wrong submitter email and had to generate a new PR.

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,82 @@
...
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import javax.annotation.Nullable;
+
+/** Factory for {@link Resource} retrieving Container ID information. */
+public final class ContainerResource {
```
kubawach, line 22: Other "system" resource factories (eg OsResource) eagerly build and cache the resource for improved performance. I believe same could be applied here.

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,82 @@
...
+   * by either '.' or '-'
+   *
+   * <p>We see this with CRI-O "crio-abcdef1234567890ABCDEF.freetext", then use {@linkplain
+   * ContainerResource#HEX_EXTRACTOR} to extract the container hex id
+   *
+   * @return containerId
+   */
+  @Nullable
+  @SuppressWarnings("DefaultCharset")
+  public String extractContainerId() {
```
kubawach, line 59: Can be made package-protected.

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,82 @@
...
+    try (BufferedReader reader = new BufferedReader(new FileReader(cgroupFilePath))) {
```
kubawach, line 60: NIT: I'd prefer (but this may be a personal preference only) to first test file existence (File.exists(...)) instead of relying on exception (FileNotFound here).

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
lo-jason: done

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
lo-jason: done

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
lo-jason: done

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,90 @@
...
+      Pattern.compile("^([\\w]*?-)?([a-fA-F0-9]+)(\\.[\\w]*?)?$");
+
+  private final String cgroupFilePath;
+
+  // package private for testing purposes
+  ContainerResource(String cgroupFilePath) {
+    this.cgroupFilePath = cgroupFilePath;
+  }
+
+  private static final ContainerResource INSTANCE =
```
jkwatson, line 36: this field should be up with the rest of the static fields.

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,90 @@
...
+      new ContainerResource(UNIQUE_HOST_NAME_FILE_NAME);
+
+  /** Returns resource with container information. */
+  public static Resource get() {
+    String containerId = INSTANCE.extractContainerId();
+
+    if (containerId == null) {
+      return Resource.empty();
+    }
+    return Resource.create(Attributes.of(ResourceAttributes.CONTAINER_ID, containerId));
```
jkwatson, line 46: I'm confused by this bit. I would just have a private static create() method that did this work. Then you don't have to have any logic in here, nor create what is just a copy of the existing INSTANCE on return. 

Take a look at how the EcsResource does this for a good example of the pattern.

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,88 @@
...
+/** Factory for {@link Resource} retrieving Container ID information. */
+public final class ContainerResource {
+
+  private static final Logger logger = Logger.getLogger(ContainerResource.class.getName());
+  private static final String UNIQUE_HOST_NAME_FILE_NAME = "/proc/self/cgroup";
+  private static final Pattern HEX_EXTRACTOR =
+      Pattern.compile("^([\\w]*?-)?([a-fA-F0-9]+)(\\.[\\w]*?)?$");
+  private static final Resource INSTANCE = buildResource();
+
+  static Resource buildResource() {
```
jkwatson, line 30: looks good. this method can be private, yes?

lo-jason: done

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
lo-jason: written to make it in-line with other factories.

jkwatson:
The fullConfigTest has started failing on CI in more than one PR now. I'm investigating, but I don't know what's going on yet.

jkwatson:
@lo-jason sorry for the trouble, but you'll need to rebase against the main branch to get the fixes for the failing tests.

Thanks!

kubawach (approved):

jkwatson (approved):
Thanks!

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,88 @@
...
+        String lastSection = sections[sections.length - 1];
+        Matcher matcher = HEX_EXTRACTOR.matcher(lastSection);
+        if (matcher.matches() && matcher.group(2) != null && !matcher.group(2).isEmpty()) {
+          return matcher.group(2);
+        }
+      }
+    } catch (IOException e) {
+      logger.log(Level.WARNING, "Unable to read file: " + e.getMessage());
+    }
+    return null;
```
anuraaga, line 84: I think a similar snippet in `DockerHelper` is from before we switched to Java 8. Can we use `Files.lines`?

```java
Path namePath = Paths.get(cgroupFilePath);
if (!Files.exists(namePath) || !Files.isReadable(nameFile)) {
  return null;
}
try (Stream<String> s = Files.lines(namePath)) {
  s.filter(line -> !line.isEmpty())
  .map(line -> line.split('/'))
  .filter(sections -> sections.length == 0)
  .map(ContainerResource::getIdFromCgroupLine)
  .findFirst()
  .orElse(null);
}
```

Seems more modern and even if there's some overhead from a Stream it doesn't matter in a resource detector. findFirst should short circuit such that the whole file isn't read.

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import javax.annotation.Nullable;
+
+/** Factory for {@link Resource} retrieving Container ID information. */
+public final class ContainerResource {
+
+  private static final Logger logger = Logger.getLogger(ContainerResource.class.getName());
+  private static final String UNIQUE_HOST_NAME_FILE_NAME = "/proc/self/cgroup";
+  private static final Pattern HEX_EXTRACTOR =
```
anuraaga, line 26: This regex looks quite gnarly. Can we just iterate for the indexes of optional metadata?

```java
@Nullable
private static String getIdFromCgroupLine(String line) {
  // Start at the character after '-' if found or beginning of string (-1 + 1) if not.
  int startIndex = line.indexOf('-') + 1;
  int endIndex = line.lastIndexOf('.');
  if (endIndex < 0) {
    endIndex = line.length();
  }
  String id = line.substring(startIndex, endIndex);
  if (!OtelEncodingUtils.isValidBase16StringString(id)) {
    return null;
  }
  return id;
}
```

It's a few more lines of code then creating the pattern / matcher, but not that much and seems much more straightforward (an if/else instead of relying on `-1 + 1` is totally reasonable to make even more straightforward).

sdk-extensions/resources/src/main/resources/META-INF/services/io.opentelemetry.sdk.autoconfigure.spi.ResourceProvider
```diff
@@ -2,3 +2,4 @@ io.opentelemetry.sdk.extension.resources.HostResourceProvider
 io.opentelemetry.sdk.extension.resources.OsResourceProvider
 io.opentelemetry.sdk.extension.resources.ProcessResourceProvider
 io.opentelemetry.sdk.extension.resources.ProcessRuntimeResourceProvider
+io.opentelemetry.sdk.extension.resources.ContainerResourceProvider
```
anuraaga, line 5: Let's put at top for alphabetical order and keep the newline at end of file

sdk-extensions/resources/src/test/java/io/opentelemetry/sdk/extension/resources/ContainerResourceTest.java
```diff
@@ -0,0 +1,77 @@
...
+  private static final String EXPECTED_CGROUP_3 =
+      "d86d75589bf6cc254f3e2cc29debdf85dde404998aa128997a819ff991827356";
+
+  // with prefix
+  private static final String CGROUP_LINE_4 =
+      "13:name=systemd:/podruntime/docker/kubepods/docker-dc579f8a8319c8cf7d38e1adf263bc08d23";
+  private static final String EXPECTED_CGROUP_4 = "dc579f8a8319c8cf7d38e1adf263bc08d23";
+
+  @SuppressWarnings("DefaultCharset")
+  public File createCGroup(File file, String line) throws IOException {
```
anuraaga, line 46:
```suggestion
  private static File createCGroup(File file, String line) throws IOException {
```

anuraaga: I'd order this at bottom of file

anuraaga: Also change to accept `Path` instead and use `Files.write`

sdk-extensions/resources/src/test/java/io/opentelemetry/sdk/extension/resources/ContainerResourceTest.java
```diff
@@ -0,0 +1,77 @@
...
+        BufferedWriter bw = new BufferedWriter(os)) {
+      bw.write("");
+      bw.write(line);
+      bw.flush();
+    }
+    return file;
+  }
+
+  @Test
+  public void testInvalidContainer(@TempDir File tempFolder) throws IOException {
```
anuraaga, line 58: `@TempDir` works with `Path` too :)

sdk-extensions/resources/src/test/java/io/opentelemetry/sdk/extension/resources/ContainerResourceTest.java
```diff
@@ -0,0 +1,77 @@
...
+    File cgroup = createCGroup(new File(tempFolder, "cgroup1"), INVALID_CGROUP_LINE_1);
+    assertEquals(null, extractContainerId(cgroup.getPath()));
```
anuraaga, line 60: We use assertj,

`assertThat(extract).isNull()`

sdk-extensions/resources/src/test/java/io/opentelemetry/sdk/extension/resources/ContainerResourceTest.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  @Test
+  public void testContainer(@TempDir File tempFolder) throws IOException {
+    File cgroup = createCGroup(new File(tempFolder, "cgroup1"), CGROUP_LINE_1);
+    assertEquals(EXPECTED_CGROUP_1, extractContainerId(cgroup.getPath()));
```
anuraaga, line 66: `assertThat(extract).isEqualTo(`

sdk-extensions/resources/src/test/java/io/opentelemetry/sdk/extension/resources/ContainerResourceTest.java
```diff
@@ -0,0 +1,77 @@
...
+    try (FileWriter os = new FileWriter(file);
+        BufferedWriter bw = new BufferedWriter(os)) {
+      bw.write("");
+      bw.write(line);
+      bw.flush();
+    }
+    return file;
+  }
+
+  @Test
```
anuraaga, line 57: All the tests will be more readable if you inline the values instead of extracting constants for the cgroup line / expected cgroup.

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,88 @@
...
+      Pattern.compile("^([\\w]*?-)?([a-fA-F0-9]+)(\\.[\\w]*?)?$");
+  private static final Resource INSTANCE = buildResource();
+
+  private static Resource buildResource() {
```
anuraaga, line 30: Instead of testing `extractContainerId`, go ahead and make an overload of `buildResource` accepting the filename and test the actual resource. We're missing coverage of significant logic like line 33

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,88 @@
...
+      return null;
+    }
+    try (BufferedReader reader = new BufferedReader(new FileReader(nameFile))) {
+      String line;
+      while ((line = reader.readLine()) != null) {
+        if (line.isEmpty()) {
+          continue;
+        }
+        // This cgroup output line should have the container id in it
+        String[] sections = line.split(File.separator);
```
anuraaga, line 71: I think we should use `/` instead of `File.separator`, cgroup should just have one accepted separator

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
lo-jason: done

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
lo-jason: done

sdk-extensions/resources/src/test/java/io/opentelemetry/sdk/extension/resources/ContainerResourceTest.java
lo-jason: Done

lo-jason:
@anuraaga

sdk-extensions/resources/src/test/java/io/opentelemetry/sdk/extension/resources/ContainerResourceTest.java
lo-jason: prefer not to because magic string, length issues.

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,104 @@
...
+   * <p>We see this with CRI-O "crio-abcdef1234567890ABCDEF.freetext", then use {@linkplain
+   * ContainerResource#HEX_EXTRACTOR} to extract the container hex id
+   *
+   * <p>package private for testing purposes
+   *
+   * @return containerId
+   */
+  @IgnoreJRERequirement
+  @Nullable
+  @SuppressWarnings("DefaultCharset")
```
anuraaga, line 63: Shouldn't be needed anymore I think

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,104 @@
...
+  private static Optional<String> extractContainerId(Path cgroupFilePath) {
```
anuraaga, line 64: We don't really use Optional in this codebase. You already have Nullable annotation, return null instead of optional

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,104 @@
...
+          .map(line -> getIdFromLine(line))
+          .filter(Objects::nonNull)
+          .findFirst();
+    } catch (IOException e) {
+      logger.log(Level.WARNING, "Unable to read file: " + e.getMessage());
+    }
+    return Optional.empty();
+  }
+
+  @SuppressWarnings("SystemOut")
```
anuraaga, line 80: Remove

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,104 @@
...
+  private static String getIdFromLine(String line) {
+    // This cgroup output line should have the container id in it
+    System.out.println("Processing: " + line);
```
anuraaga, line 83: Remove

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,104 @@
...
+    String[] sections = line.split("/");
+    if (sections.length <= 1) {
```
anuraaga, line 85: I think this enforces `/` is always there - is it guaranteed?

Also can just use `lastIndexOf('/')` instead of splitting

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,104 @@
...
+      return null;
+    }
+
+    String lastSection = sections[sections.length - 1];
+    int startIdx = lastSection.indexOf("-");
+    int endIdx = lastSection.lastIndexOf(".");
+
+    Matcher matcher =
+        HEX_EXTRACTOR.matcher(
```
anuraaga, line 94: Shouldn't need HEX_EXTRACTOR, can you use OtelEncoding.isValidHexString?

sdk-extensions/resources/src/test/java/io/opentelemetry/sdk/extension/resources/ContainerResourceTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+    Path cgroup4 = createCGroup(tempFolder.resolve("cgroup4"), CGROUP_LINE_4);
+    assertThat(getContainerId(buildResource(cgroup4))).isEqualTo(EXPECTED_CGROUP_4);
+  }
+
+  private static String getContainerId(Resource resource) {
+    return resource.getAttributes().get(ResourceAttributes.CONTAINER_ID);
+  }
+
+  public static Path createCGroup(Path path, String line) throws IOException {
```
anuraaga, line 81:
```suggestion
  private static Path createCGroup(Path path, String line) throws IOException {
```

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
lo-jason: it appears that '/' is always there.

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
lo-jason: it looks a bit uglier w/o optional since we are using findFirst but point taken.

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
lo-jason: +1

jkwatson:
@lo-jason looks like this needs a rebase in order to pick up the latest changes to the build. Thanks!

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
```diff
@@ -0,0 +1,104 @@
...
+      if (value.isPresent()) {
+        return value.get();
+      }
+    } catch (IOException e) {
+      logger.log(Level.WARNING, "Unable to read file: " + e.getMessage());
+    }
+    return null;
+  }
+
+  private static String getIdFromLine(String line) {
```
anuraaga, line 82:
```suggestion
  @Nullable
  private static String getIdFromLine(String line) {
```

anuraaga (approved):
Just tiny nit, thanks!

sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
lo-jason: Done, thanks!

PR #3324, opened 2021-06-17
Extract JMH conventions plugin.

anuraaga:


buildSrc/src/main/kotlin/otel.jmh-conventions.gradle.kts
```diff
@@ -0,0 +1,36 @@
+plugins {
```
kubawach, line 1: Nice! Much shorter and with improved readability as compared to plugin use.

kubawach (approved):

jkwatson (approved):
love it! Thanks!

PR #3325, opened 2021-06-17
Update to Gradle 7.1

anuraaga:


jkwatson (approved):
Thanks for keeping on top of it!

PR #3329, opened 2021-06-18
Migrate protobuf logic to plugin

anuraaga:
Also remove IDE logic which seems to have been fixed in the protobuf plugin at some point


kubawach (approved):
Looks good!

build.gradle.kts
```diff
@@ -107,37 +104,6 @@ subprojects {
...
-            afterEvaluate {
-                // Classpath when compiling protos, we add dependency management directly
-                // since it doesn't follow Gradle conventions of naming / properties.
-                dependencies {
-                    add("compileProtoPath", platform(project(":dependencyManagement")))
-                    add("testCompileProtoPath", platform(project(":dependencyManagement")))
-                }
-            }
-        }
-
```
jkwatson, line 140: whoa. this stuff isn't needed any more?

jkwatson: oh wait...it's just moved to the other file. 🤦🏼

jkwatson (approved):
By the way, if you just do a project "build" in IDEA, does it work for you? For me, it partially fails to due some of the jmh dependencies not being resolved, even though IDEA doesn't show any errors in the files.

Issue #3328, opened 2021-06-17
AWS X-Ray Centralized Sampler


luneo7:
**Is your feature request related to a problem? Please describe.**
Today we already have a Jaeger remote sampler in the Java SDK, but we don't have one that targets AWS X-Ray, it would be nice to have one sampler for X-Ray.

**Describe the solution you'd like**
A sampler that takes the service name, a fallback sampler, and a endpoint parameter that can point to the X-Ray service or to the X-Ray receiver with proxy server enabled, that fetches the sampling config and does centralized sampling.

anuraaga:
Thanks for filing this @luneo7 - we're planning to implement support for the X-Ray sampler and hope to get it out soon

luneo7:
@anuraaga any news when we can get a version with everything set up? Thanks =]

jkwatson:
This is being implemented in the opentelemetry-java-contrib project.

PR #3331, opened 2021-06-18
Add support for fetching xray sampling rules.

anuraaga:
First step for #3328

### DIFF
--- a/sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/GetSamplingRulesResponse.java
+++ b/sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/GetSamplingRulesResponse.java
@@ -7,165 +7,104 @@ package io.opentelemetry.sdk.extension.aws.trace;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.google.auto.value.AutoValue;
 import java.util.List;
 import java.util.Map;
 import javax.annotation.Nullable;
 
 @AutoValue
-@JsonDeserialize(builder = GetSamplingRulesResponse.Builder.class)
 abstract class GetSamplingRulesResponse {
 
-  @JsonProperty("NextToken")
+  @JsonCreator
+  static GetSamplingRulesResponse create(
+      @JsonProperty("NextToken") String nextToken,
+      @JsonProperty("SamplingRuleRecords") List<SamplingRuleRecord> samplingRules) {
+    return new AutoValue_GetSamplingRulesResponse(nextToken, samplingRules);
+  }
+
   @Nullable
   abstract String getNextToken();
 
-  @JsonProperty("SamplingRuleRecords")
   abstract List<SamplingRuleRecord> getSamplingRules();
 
-  @AutoValue.Builder
-  abstract static class Builder {
-    @JsonCreator
-    static Builder builder() {
-      return new AutoValue_GetSamplingRulesResponse.Builder();
-    }
-
-    @JsonProperty("NextToken")
-    abstract Builder setNextToken(@Nullable String nextToken);
-
-    @JsonProperty("SamplingRuleRecords")
-    abstract Builder setSamplingRules(List<SamplingRuleRecord> samplingRules);
-
-    abstract GetSamplingRulesResponse build();
-  }
-
   @AutoValue
-  @JsonDeserialize(builder = SamplingRuleRecord.Builder.class)
   abstract static class SamplingRuleRecord {
 
-    @JsonProperty("CreatedAt")
+    @JsonCreator
+    static SamplingRuleRecord create(
+        @JsonProperty("CreatedAt") String createdAt,
+        @JsonProperty("ModifiedAt") String modifiedAt,
+        @JsonProperty("SamplingRule") SamplingRule rule) {
+      return new AutoValue_GetSamplingRulesResponse_SamplingRuleRecord(createdAt, modifiedAt, rule);
+    }
+
     abstract String getCreatedAt();
 
-    @JsonProperty("ModifiedAt")
     abstract String getModifiedAt();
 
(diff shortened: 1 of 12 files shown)


sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/GetSamplingRulesResponse.java
```diff
@@ -0,0 +1,173 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.google.auto.value.AutoValue;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+@AutoValue
+@JsonDeserialize(builder = GetSamplingRulesResponse.Builder.class)
```
anuraaga, line 17: Cribbed pattern for AutoValue + Jackson from https://github.com/artem-zinnatullin/AutoJackson

jkwatson: this stuff is a bit crazy, but I guess it works!

jkwatson (approved):
I'm going to mostly trust you on this one. Are all the builders really necessary? Could you just write some super simple java beans without autovalue and get jackson to map the data more easily?

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/GetSamplingRulesResponse.java
luneo7: Since this is the response it will mostly be only deserialized, and you can annotate with `@JsonDeserialize(builder = Whatever.Builder.class)` I think you don't need to annotate anything else with `@JsonProperty("property")` other than the the ones in the `Builder` class, unless we are going to serialize it then we would need to have in both... but we can also change the ObjectMapper

luneo7: For instance if using the ObjectMapper with `setPropertyNamingStrategy(PropertyNamingStrategies.UPPER_CAMEL_CASE)`, SamplingRule can just be 

```
@AutoValue
@JsonDeserialize(builder = AutoValue_SamplingRule.Builder.class)
public abstract class SamplingRule {

  public abstract Map<String, String> attributes();

  public abstract Double fixedRate();

  public abstract String httpMethod();

  public abstract String host();

  public abstract Integer priority();

  public abstract Integer reservoirSize();

  public abstract String resourceArn();

  public abstract String ruleArn();

  public abstract String ruleName();

  public abstract String serviceName();

  public abstract String serviceType();

  public abstract String urlPath();

  public abstract Integer version();

  SamplingRule() {}

  public static Builder builder() {
    return new AutoValue_SamplingRule.Builder();
  }

  @AutoValue.Builder
  @JsonPOJOBuilder(withPrefix = "")
  public abstract static class Builder {
    public abstract Builder attributes(Map<String, String> attributes);

    public abstract Builder fixedRate(Double fixedRate);

    @JsonProperty("HTTPMethod")
    public abstract Builder httpMethod(String httpMethod);

    public abstract Builder host(String host);

    public abstract Builder priority(Integer priority);

    public abstract Builder reservoirSize(Integer reservoirSize);

    @JsonProperty("ResourceARN")
    public abstract Builder resourceArn(String resourceArn);

    @JsonProperty("RuleARN")
    public abstract Builder ruleArn(String ruleArn);

    public abstract Builder ruleName(String ruleName);

    public abstract Builder serviceName(String serviceName);

    public abstract Builder serviceType(String serviceType);

    @JsonProperty("URLPath")
    public abstract Builder urlPath(String urlPath);

    public abstract Builder version(Integer version);

    public abstract SamplingRule build();
  }
}
```

luneo7: Also should we have package private constructors in the auto value?

luneo7:
Since `SamplingTargets` (`GetSamplingTargetsRequest`, `GetSamplingTargetsResult`, `SamplingStatisticsDocument`, `SamplingTargetDocument`) are also used in the X-Ray centralized sampler should we have them here as well?

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/GetSamplingRulesResponse.java
```diff
@@ -0,0 +1,173 @@
...
+      abstract Builder setPriority(int priority);
+
+      @JsonProperty("ReservoirSize")
+      abstract Builder setReservoirSize(int reservoirSize);
+
+      @JsonProperty("ResourceARN")
+      abstract Builder setResourceArn(String resourceArn);
+
+      @JsonProperty("RuleARN")
+      @Nullable
```
luneo7, line 151: You don't need to annotate the method with `@Nullable` since it returns the builder that won't be null

anuraaga:
> Could you just write some super simple java beans without autovalue and get jackson to map the data more easily?

I possibly could but probably would lack toString due to the tediousness and at some point we always end up wanting it I think (i e Intellij debug). I can see if a better pattern comes up though these are internal.

> should we have them here as well?

Yeah just keeping the size of the prs smaller

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/GetSamplingRulesResponse.java
anuraaga: Oops thanks!

anuraaga:
Went ahead with removing the builders in favor of factory methods. While we don't like such long factories usually, since it's only called from object mapper I think it's ok and reduces the code (and especially the bytecode) quite a bit.

PR #3333, opened 2021-06-19
*WIP* [Metrics]  Implement new specification

jsuereth:
This PR Is *NOT* ready for formal review.  This is still an early prototype allowing us to kick the tires on API decisions and make sure all (previous) functionality we rely on makes it away back into the API specification.   This not meant to be "clean" java code (yet), only architecturally close to what we may provide in the final SDK.    Naming / nitty-gritty comments will be ignored until architecture is stable (not sure which pieces of code will remain as we stabilize the SDK).  Also note:  I *do* plan to rebase and rewrite history for easier-to-consume chunks.    Right now this is my working branch, and it's not stable enough for anything more than high-level review or commentary.

`sdk/metrics/implementation_notes.md` Provides current TODOs/ideas/investigation areas.

Posting this by request of Java SiG for review / ongoing updates. 

A few components to call out:

- New "Storage"/"state" package to more clearly separate Aggregation from usage-of-storage + instruments.
- Exemplar Sampling + Wiring in all aggregators + the data model (not exporters yet)
- `MeasurementProcessor` implementation
- Deprecation (and follow-on build-tool hell) of `Summary` metric.

## TODOs

- [ ] Fix all unit tests/builds
- [ ] View API
- [ ] Detangle `Aggregator` from `InstrumentInfo`
- [ ] Clean up `*SharedState`
- [ ] Create "matcher" api for metrics data package in tests


jkwatson:
I know this isn't ready for review, @jsuereth , but I'm wondering how we are ever going to be able to tackle reviewing something this huge effectively. Do you have any thoughts on how to get this reviewed incrementally, eventually?

jsuereth:
@jkwatson Yeah, I'm going to rebase and fragment into composable/digestable commits for each section.   If necessary, I'll peel pieces off into separate PRs.

There will be one "big bang" PR for API/SDK replacement, but hoping you can review the components via commits.

jsuereth:
@jkwatson Did a first pass at git history rewrite for easier consumption.

Notes:
- [ef21ce3](https://github.com/open-telemetry/opentelemetry-java/pull/3333/commits/ef21ce35a55c01968bf3ed6530e013ad09ab28b3)  - This can probably be a separate PR if necessary.  Updates metric DataModel to match "stable" release + add assertj helpers
- [
20d2cfa](https://github.com/open-telemetry/opentelemetry-java/pull/3333/commits/20d2cfa88c329a38dc7f237efae4d00a68f5b4ed) Metrics API-only changes
- [50bec80](https://github.com/open-telemetry/opentelemetry-java/pull/3333/commits/50bec80e0f54f9238be8525d8ab645b1e33b6e08) Metrics SDK-only changes
- [42b22f4](https://github.com/open-telemetry/opentelemetry-java/pull/3333/commits/42b22f430508ed8340e55d2288a92939e931c1e1) Other SDK changes to account for the above commits

I'd still like to do a bit of cleanup + get views wired back in (as close to what I think the spec will be as possible), but let me know if that split makes sense.  Unforutnately the last 3 commits need to be consumed together.

api/metrics/src/main/java/io/opentelemetry/api/metrics/MeterProvider.java
```diff
@@ -5,43 +5,32 @@
...
-  }
-
-  /**
-   * Gets or creates a named meter instance.
-   *
-   * @param instrumentationName The name of the instrumentation library, not the name of the
-   *     instrument*ed* library.
-   * @return a tracer instance.
-   */
-  Meter get(String instrumentationName);
```
jsuereth, line 34: From SiG:  Bring this back for non-otel-instrumentation providers.

api/metrics/src/main/java/io/opentelemetry/api/metrics/BoundDoubleHistogram.java
```diff
@@ -0,0 +1,32 @@
...
+/** A hisgogram instrument that records {@code long} values with preassociated attributes. */
+@ThreadSafe
+public interface BoundDoubleHistogram {
+  /**
+   * Record a value with a pre-bound set of attributes.
+   *
+   * @param value The amount of the measurement.
+   * @param context The explicit context to associate with this measurement.
+   */
+  public void record(double value, Context context);
```
jkwatson, line 20: for when you're ready... interface methods inherit the access of the class, so shouldn't have additional modifiers on the methods.

api/metrics/src/main/java/io/opentelemetry/api/metrics/BoundDoubleCounter.java
```diff
@@ -5,20 +5,28 @@
...
 import javax.annotation.concurrent.ThreadSafe;
 
-/** A {@code Bound Instrument} for a {@link DoubleCounter}. */
+/** A counter instrument that records {@code double} values with preassociated attributes. */
 @ThreadSafe
-public interface BoundDoubleCounter extends BoundSynchronousInstrument {
+public interface BoundDoubleCounter {
   /**
-   * Adds the given {@code increment} to the current value. The values cannot be negative.
+   * Reecord a value with prebound attributes.
```
anuraaga, line 15: Perhaps not really a point for this PR and more for the spec (but I believe feedback loop is faster on language SIG PRs than spec issues :-D) but what's the use case for prebound attributes given we have `Resource`? We haven't written that many metrics in instrumentation yet, but so far I haven't seen any time I would want to use prebound (having the Resource is important of course to handle "static attributes")

api/metrics/src/main/java/io/opentelemetry/api/metrics/DoubleCounter.java
```diff
@@ -5,59 +5,44 @@
...
-
+public interface DoubleCounter extends Counter {
+  /**
+   * Record a value with a set of attributes.
+   *
+   * @param value The increment amount. MUST be non-negative.
+   * @param attributes A set of attributes to associate with the count.
+   * @param context The explicit context to associate with this measurement.
+   */
+  public void add(double value, Attributes attributes, Context context);
```
anuraaga, line 22: The API generally looks reasonable to me. The main point that I'd like to make sure is considered is making sure attributes that are recorded are decouplable from attributes that are exported. I see that there is a View AttributesProcessor - is it possible to use Views to model semantic conventions? For example this list of Attributes

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#attributes

It seems like having instrumentation (via the API) know what attributes go with what metric is too tedious, and somewhat inefficient, at least for request metrics. For example, we have this

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpServerMetrics.java#L97

I think we want to make sure the API supports the ability to have one Attributes be collected to all the metrics for a request, and the SDK takes care of "applying the semantic conventions" by picking the attributes based on this spec. If Views can be used for this, and we imagine HttpServerMetricsView or such, a view for each semantic convention, then it sounds good.

api/metrics/src/main/java/io/opentelemetry/api/metrics/BoundDoubleCounter.java
jsuereth: Just ran performance benchmarks.  Pre-bound synchronous instruments are 2x faster due to lacking 1 of the lock-and-access paths for the `ConcurrentMap`.

It's literally just a performance optimisation (and it's used inside the SDK)

api/metrics/src/main/java/io/opentelemetry/api/metrics/DoubleCounter.java
jsuereth: Views can be used for this, but I'm guessing it'll be a bit awkward, as views are SDK-only right now.

We should chat, as this entire prototype made me realize we need (efficient) mechanisms for dealing with `Attributes` in aggregate, similar to your use case.   Additionally, I don't know if your use case is well supported (or not) here.   We should talk through various options and see if we wind up with a missing abstraction in the API that could help here, or if it can be solved SDK-side...

bogdandrutu: @jsuereth one thing that I know from a "source", is that you can have very good performance if you aggregate using all the "Attributes" in the critical path, then "reduce the attributes" on the result (aggregation of 5sec for example).

api/metrics/src/main/java/io/opentelemetry/api/metrics/BoundDoubleCounter.java
anuraaga: Yeah I'm sure it will perform better to prebind attributes vs recording each time when the attributes are known. I'm wondering whether we have any examples of attributes that are known ahead of time, that aren't in the Resource?

Also, while writing a unit test for metrics, I found it a bit unnatural for bound instruments to have a different API then unbound. How about allowing attributes to be passed in even for bound instruments? In most cases maybe they won't be passed in, but API-wise I'm not sure of a reason to forbid it.

anuraaga: Nevermind, figured out how to use the bound counters.

PR #3334, opened 2021-06-21
Extract publishing and animalsniffer plugins.

anuraaga:
Also fixed a couple of deprecations

### DIFF
--- a/buildSrc/src/main/kotlin/otel.japicmp-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.japicmp-conventions.gradle.kts
@@ -42,35 +42,38 @@ fun findArtifact(version: String): File {
     }
 }
 
-afterEvaluate {
-    tasks {
-        val jApiCmp by registering(JapicmpTask::class) {
-            dependsOn("jar")
-            // the japicmp "old" version is either the user-specified one, or the latest release.
-            val apiBaseVersion: String? by project
-            val baselineVersion = apiBaseVersion ?: latestReleasedVersion
-            val baselineArtifact = findArtifact(baselineVersion)
-            oldClasspath = files(baselineArtifact)
+// generate the api diff report for any module that is stable and publishes a jar.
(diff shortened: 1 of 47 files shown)


buildSrc/src/main/kotlin/otel.publish-conventions.gradle.kts
```diff
@@ -0,0 +1,103 @@
...
+
+if (System.getenv("CI") != null) {
+    signing {
+        useInMemoryPgpKeys(System.getenv("GPG_PRIVATE_KEY"), System.getenv("GPG_PASSWORD"))
+        sign(publishing.publications["mavenPublication"])
+    }
+}
+
+tasks {
+    register("updateVersionInDocs") {
```
anuraaga, line 82: We previously added this to allprojects but it seems to only be needed for published ones

buildSrc/src/main/kotlin/otel.japicmp-conventions.gradle.kts
```diff
@@ -42,35 +42,38 @@ fun findArtifact(version: String): File {
...
-    tasks {
-        val jApiCmp by registering(JapicmpTask::class) {
-            dependsOn("jar")
-            // the japicmp "old" version is either the user-specified one, or the latest release.
-            val apiBaseVersion: String? by project
-            val baselineVersion = apiBaseVersion ?: latestReleasedVersion
-            val baselineArtifact = findArtifact(baselineVersion)
-            oldClasspath = files(baselineArtifact)
+// generate the api diff report for any module that is stable and publishes a jar.
+if (!project.hasProperty("otel.release") && !project.name.startsWith("bom")) {
```
anuraaga, line 46: This seems like a "feature" of the japicmp conventions, not publishing ones so didn't use the conditional plugin application like before

kubawach (approved):
Nice :)

jkwatson (approved):
Thanks!

PR #3335, opened 2021-06-21
Add support for fetching xray sampling targets.

anuraaga:
#3328

### DIFF
--- a/sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/XraySamplerClient.java
+++ b/sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/XraySamplerClient.java
@@ -3,37 +3,82 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+// Includes work from:
+
+/*
+ * Copyright 2010-2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Portions copyright 2006-2009 James Murty. Please see LICENSE.txt
+ * for applicable license terms and NOTICE.txt for applicable notices.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
 package io.opentelemetry.sdk.extension.aws.trace;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationContext;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import com.fasterxml.jackson.databind.module.SimpleModule;
 import io.opentelemetry.sdk.extension.aws.internal.JdkHttpClient;
+import java.io.IOException;
 import java.io.UncheckedIOException;
+import java.math.BigDecimal;
 import java.util.Collections;
+import java.util.Date;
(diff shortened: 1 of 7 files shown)


sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/XraySamplerClient.java
```diff
@@ -3,37 +3,82 @@
...
   private static final Map<String, String> JSON_CONTENT_TYPE =
       Collections.singletonMap("Content-Type", "application/json");
 
   private final String getSamplingRulesEndpoint;
+  private final String getSamplingTargetsEndpoint;
   private final JdkHttpClient httpClient;
 
   XraySamplerClient(String host) {
     this.getSamplingRulesEndpoint = host + "/GetSamplingRules";
+    this.getSamplingTargetsEndpoint = host + "/SamplingTargets";
```
kubawach, line 68: I see there is a discrepancy in naming of AWS API. Bot endpoints are used to "get" something but only one of them starts with "get"

sdk-extensions/aws/src/test/java/io/opentelemetry/sdk/extension/aws/trace/XraySamplerClientTest.java
```diff
@@ -10,45 +10,24 @@
...
+import java.nio.charset.StandardCharsets;
+import java.time.Instant;
+import java.util.Arrays;
+import java.util.Date;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
+import org.skyscreamer.jsonassert.JSONAssert;
 
 class XraySamplerClientTest {
```
kubawach, line 28: Shouldn't we also add a malformed response test? Just to verify that in such case an unchecked IO exception will be always thrown.

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/XraySamplerClient.java
```diff
@@ -3,37 +3,82 @@
...
 final class XraySamplerClient {
 
-  private static final ObjectMapper OBJECT_MAPPER =
-      new ObjectMapper()
-          .setSerializationInclusion(JsonInclude.Include.NON_EMPTY)
-          // In case API is extended with new fields.
-          .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, /* state= */ false);
+  private static final ObjectMapper OBJECT_MAPPER;
+
+  static {
```
kubawach, line 48: Out of curiosity - why separate static init block instead of in-place, as it as was previously?

kubawach (approved):

luneo7 (approved):

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/XraySamplerClient.java
```diff
@@ -42,13 +87,41 @@ GetSamplingRulesResponse getSamplingRules(GetSamplingRulesRequest request) {
...
+      return parseServiceSpecificDate(p.getText());
+    }
+
+    // Copied from AWS SDK
+    // https://github.com/aws/aws-sdk-java/blob/7b1e5b87b0bf03456df9e77716b14731adf9a7a7/aws-java-sdk-core/src/main/java/com/amazonaws/util/DateUtils.java#L239
+    /** Parses the given date string returned by the AWS service into a Date object. */
+    private static Date parseServiceSpecificDate(String dateString) {
+      try {
+        BigDecimal dateValue = new BigDecimal(dateString);
+        return new Date(dateValue.scaleByPowerOfTen(AWS_DATE_MILLI_SECOND_PRECISION).longValue());
```
jkwatson, line 121: I'm as big a fan of `BigDecimal` as the next person, but isn't this just multiplying by 1000? `Long.valueOf(dateString) * 1000` seems a little simpler?

anuraaga: Just for more context, even this brand new aws go sdk uses bigdecimal, so I guess it's the expected convention for aws api clients

https://github.com/aws/smithy-go/blob/main/time/time.go#L76

jkwatson: yep, I think keeping consistency is fine here.

jkwatson (approved):
👍🏼

PR #3341, opened 2021-06-22
Add logic for processing a xray remote SamplingRule

anuraaga:
#3328

Real logic, not just plumbing!

### DIFF
--- a/sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/SamplingRuleApplier.java
+++ b/sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/SamplingRuleApplier.java
@@ -1,0 +1,248 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.extension.aws.trace;
+
+import io.opentelemetry.api.common.AttributeKey;
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.trace.SpanKind;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.sdk.trace.data.LinkData;
+import io.opentelemetry.sdk.trace.samplers.Sampler;
+import io.opentelemetry.sdk.trace.samplers.SamplingResult;
+import io.opentelemetry.semconv.resource.attributes.ResourceAttributes;
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
+
+final class SamplingRuleApplier {
+
+  private final Sampler fixedRateSampler;
+
+  private final Map<String, Matcher> attributeMatchers;
+  private final Matcher urlPathMatcher;
+  private final Matcher serviceNameMatcher;
+  private final Matcher httpMethodMatcher;
+  private final Matcher hostMatcher;
+  private final Matcher serviceTypeMatcher;
+  private final Matcher resourceArnMatcher;
+
+  SamplingRuleApplier(GetSamplingRulesResponse.SamplingRule rule) {
+    fixedRateSampler = Sampler.parentBased(Sampler.traceIdRatioBased(rule.getFixedRate()));
+
+    if (rule.getAttributes().isEmpty()) {
+      attributeMatchers = Collections.emptyMap();
+    } else {
+      attributeMatchers =
+          rule.getAttributes().entrySet().stream()
+              .collect(Collectors.toMap(Map.Entry::getKey, e -> toMatcher(e.getValue())));
+    }
+
+    urlPathMatcher = toMatcher(rule.getUrlPath());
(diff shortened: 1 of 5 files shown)


sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/SamplingRuleApplier.java
```diff
@@ -0,0 +1,248 @@
...
+      } else {
+        if (tokenStart == -1) {
+          tokenStart = i;
+        }
+      }
+    }
+    if (tokenStart != -1) {
+      patternBuilder.append(Pattern.quote(globPattern.substring(tokenStart)));
+    }
+    return Pattern.compile(patternBuilder.toString());
```
kubawach, line 185: This may blow with `PatternSyntaxException`. I think this path should be protected and most probably log warning / return "FalseMatcher".

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/SamplingRuleApplier.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+    arn = resource.getAttributes().get(ResourceAttributes.FAAS_ID);
+    if (arn != null) {
+      return arn;
+    }
+    return attributes.get(ResourceAttributes.FAAS_ID);
+  }
+
+  @Nullable
+  private static String getServiceType(Resource resource) {
```
kubawach, line 124: A bit of theoretical question, but as this does not cover whole enum, wouldn't map be actually faster here?

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/SamplingRuleApplier.java
```diff
@@ -0,0 +1,248 @@
...
+        parentContext, traceId, name, spanKind, attributes, parentLinks);
+  }
+
+  @Nullable
+  private static String getArn(Attributes attributes, Resource resource) {
+    String arn = resource.getAttributes().get(ResourceAttributes.AWS_ECS_CONTAINER_ARN);
+    if (arn != null) {
+      return arn;
+    }
+    String cloudPlatform = resource.getAttributes().get(ResourceAttributes.CLOUD_PLATFORM);
```
kubawach, line 111: NIT: rest of the method could be factored out into a helper "getLambdaArn" call for improved readability.

kubawach (approved):

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/SamplingRuleApplier.java
anuraaga: Do you think it's possible even though we only add quoted strings and these simple wildcards?

kubawach: Good question! I thought it can be broken by nonsensical pattern (like a sequence of dots) but it seems that Java allows that without question. So now it should be always safe, at least as far as I know regexes ;)

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/SamplingRuleApplier.java
anuraaga: Would need to bench to confirm but Map sounds good. I think the key point is that hashcodes are cached for Strings in Java so in practice should only be computed around once per app.

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/SamplingRuleApplier.java
```diff
@@ -0,0 +1,254 @@
...
+    Map<String, String> xrayCloudPlatform = new HashMap<>();
+    xrayCloudPlatform.put(ResourceAttributes.CloudPlatformValues.AWS_EC2, "AWS::EC2::Instance");
+    xrayCloudPlatform.put(ResourceAttributes.CloudPlatformValues.AWS_ECS, "AWS::ECS::Container");
+    xrayCloudPlatform.put(ResourceAttributes.CloudPlatformValues.AWS_EKS, "AWS::EKS::Container");
+    xrayCloudPlatform.put(
+        ResourceAttributes.CloudPlatformValues.AWS_ELASTIC_BEANSTALK,
+        "AWS::ElasticBeanstalk::Environment");
+    xrayCloudPlatform.put(
+        ResourceAttributes.CloudPlatformValues.AWS_LAMBDA, "AWS::Lambda::Function");
+    XRAY_CLOUD_PLATFORM = Collections.unmodifiableMap(xrayCloudPlatform);
```
kubawach, line 40: Nice touch :D

jkwatson (approved):
This is way more complicated that I would have guessed. What's the performance like on this thing, I wonder?

luneo7 (approved):

anuraaga:
@jkwatson It's a pretty strangely designed system / API in general but is what it is. Performance tends to be OK in practice since most rules have one matcher with possibly one wildcard, and TrueMatcher for the rest. But yeah it'd be easy to create a disastrous rule too, don't think x-ray does anything to stop that.

PR #3343, opened 2021-06-24
Add the actual XRay remote sampler which polls rules and orders / app…

anuraaga:
…lies them.

This implements the core part of XRay remote sampling, which polls rules and applies a fixed rate based on matching rules. This is the most important part, but in followups I will also be adding support for XRay's "centralized reservoir" (where the server handles the reservoir quotas instead of being fixed per client, this is what the GetSamplingTargets RPC is used for). @jkwatson You aint seen complicated yet :P

### DIFF
--- a/sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/AwsXrayRemoteSampler.java
+++ b/sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/AwsXrayRemoteSampler.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.extension.aws.trace;
+
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.trace.SpanKind;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.sdk.internal.DaemonThreadFactory;
+import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.sdk.trace.data.LinkData;
+import io.opentelemetry.sdk.trace.samplers.Sampler;
+import io.opentelemetry.sdk.trace.samplers.SamplingResult;
+import java.io.Closeable;
+import java.util.List;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.annotation.Nullable;
+
+/** Remote sampler that gets sampling configuration from AWS X-Ray. */
+public final class AwsXrayRemoteSampler implements Sampler, Closeable {
+
+  private static final Logger logger = Logger.getLogger(AwsXrayRemoteSampler.class.getName());
+
+  private static final String WORKER_THREAD_NAME =
+      AwsXrayRemoteSampler.class.getSimpleName() + "_WorkerThread";
+
+  private final Resource resource;
+  private final Sampler initialSampler;
+  private final XraySamplerClient client;
+  private final ScheduledExecutorService executor;
+  private final ScheduledFuture<?> pollFuture;
+
+  @Nullable private volatile GetSamplingRulesResponse previousRulesResponse;
+  private volatile Sampler sampler;
+
+  /**
+   * Returns a {@link AwsXrayRemoteSamplerBuilder} with the given {@link Resource}. This {@link
+   * Resource} should be the same as what the {@linkplain io.opentelemetry.sdk.OpenTelemetrySdk
+   * OpenTelemetry SDK} is configured with.
+   */
+  // TODO(anuraaga): Deprecate after
+  // https://github.com/open-telemetry/opentelemetry-specification/issues/1588
+  public static AwsXrayRemoteSamplerBuilder newBuilder(Resource resource) {
+    return new AwsXrayRemoteSamplerBuilder(resource);
+  }
+
+  AwsXrayRemoteSampler(
+      Resource resource, String endpoint, Sampler initialSampler, long pollingIntervalNanos) {
+    this.resource = resource;
+    this.initialSampler = initialSampler;
+    client = new XraySamplerClient(endpoint);
+    executor =
+        Executors.newSingleThreadScheduledExecutor(new DaemonThreadFactory(WORKER_THREAD_NAME));
+
+    sampler = initialSampler;
+
+    pollFuture =
+        executor.scheduleAtFixedRate(
+            this::getAndUpdateSampler, 0, pollingIntervalNanos, TimeUnit.NANOSECONDS);
+  }
+
+  @Override
+  public SamplingResult shouldSample(
+      Context parentContext,
+      String traceId,
+      String name,
+      SpanKind spanKind,
+      Attributes attributes,
+      List<LinkData> parentLinks) {
+    return sampler.shouldSample(parentContext, traceId, name, spanKind, attributes, parentLinks);
+  }
+
+  @Override
+  public String getDescription() {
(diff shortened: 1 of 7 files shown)


sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/AwsXrayRemoteSampler.java
```diff
@@ -0,0 +1,105 @@
...
+   * Resource} should be the same as what the {@linkplain io.opentelemetry.sdk.OpenTelemetrySdk
+   * OpenTelemetry SDK} is configured with.
+   */
+  // TODO(anuraaga): Deprecate after
+  // https://github.com/open-telemetry/opentelemetry-specification/issues/1588
+  public static AwsXrayRemoteSamplerBuilder newBuilder(Resource resource) {
+    return new AwsXrayRemoteSamplerBuilder(resource);
+  }
+
+  AwsXrayRemoteSampler(
```
kubawach, line 54: NIT: this is my defensive thinking, but as the constructor is package-protected and thus accessible, I'd validate constructor parameters here.

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/AwsXrayRemoteSampler.java
```diff
@@ -0,0 +1,105 @@
...
+    this.initialSampler = initialSampler;
+    client = new XraySamplerClient(endpoint);
+    executor =
+        Executors.newSingleThreadScheduledExecutor(new DaemonThreadFactory(WORKER_THREAD_NAME));
+
+    sampler = initialSampler;
+
+    pollFuture =
+        executor.scheduleAtFixedRate(
+            this::getAndUpdateSampler, 0, pollingIntervalNanos, TimeUnit.NANOSECONDS);
```
kubawach, line 66: I believe that pooling interval should have a sanity check - that is if the value is outside of typical, expected values, a warning should be logged. For me it's easy to imagine sb setting interval to 1 (assuming it's one second) and degrading the performance with request each nanosecond ;)

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/SamplingRulesSampler.java
```diff
@@ -0,0 +1,73 @@
...
+    for (SamplingRuleApplier applier : ruleAppliers) {
+      if (applier.matches(name, attributes, resource)) {
+        return applier.shouldSample(
+            parentContext, traceId, name, spanKind, attributes, parentLinks);
+      }
+    }
+
+    // In practice, X-Ray always returns a Default rule that matches all requests so it is a bug in
+    // our code or X-Ray to reach here, fallback just in case.
+    logger.log(
```
kubawach, line 61: 👌

kubawach (approved):

jkwatson:
>  You aint seen complicated yet :P

Hm. If that's the case, I'm definitely starting to question if this repository is the right place to host the code. Can AWS publish a library with all this code in it that we can just depend on, rather than having it all live here (as well in wherever it lives for the old AWS java instrumentation)?

anuraaga:
I think code for otel components that interface with vendors generally go in an otel  contrib repo. We've been treating the sdk-extensions as a sort of contrib but maybe we should move to java-contrib? Not sure if it makes a difference in the actual scheme of things though - if the issue is review load a better option (though not mutually exclusive) is probably to add @kubawach to approvers finally :)

jkwatson:
> I think code for otel components that interface with vendors generally go in an otel contrib repo. We've been treating the sdk-extensions as a sort of contrib but maybe we should move to java-contrib? Not sure if it makes a difference in the actual scheme of things though - if the issue is review load a better option (though not mutually exclusive) is probably to add @kubawach to approvers finally :)

The issue isn't review load (but I'm also in favor of getting @kubawach added to reviewers), but potential maintenance & support load over the long term. If you are the only person who is really able to maintain the code, and debug issues with its interaction with AWS services, then that feels like the core implementation details probably don't belong here. A simple AWS propagator is one thing, but when we're talking about maintaining several thousand lines of non-trivial code for an AWS-specific sampler implementation, I'm probably against it.

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/AwsXrayRemoteSampler.java
```diff
@@ -0,0 +1,105 @@
...
+  public static AwsXrayRemoteSamplerBuilder newBuilder(Resource resource) {
+    return new AwsXrayRemoteSamplerBuilder(resource);
+  }
+
+  AwsXrayRemoteSampler(
+      Resource resource, String endpoint, Sampler initialSampler, long pollingIntervalNanos) {
+    this.resource = resource;
+    this.initialSampler = initialSampler;
+    client = new XraySamplerClient(endpoint);
+    executor =
```
jkwatson, line 59: personally, I'd just assign this in the declaration above, since it's final and doesn't depend on params.

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/AwsXrayRemoteSampler.java
jkwatson: it looks like all the validation is down in the builder.

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/AwsXrayRemoteSampler.java
jkwatson: I think they're all validated in the builder

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/SamplingRulesSampler.java
jkwatson: Hm. In the case this happens, it will totally spam the logs at WARNING for every span. I know it shouldn't ever happen, but should we maybe not be so aggressive with the logging here?

jkwatson (approved):
Thanks! Looks good enough to merge, then move over to the contrib repo.

jmacd (approved):
:eyes: 

Speaking for myself (and my employer), we have a strong interest in being able to approximately count sampled spans. I am also confident we can do this counting correctly and with a specification, as discussed in https://github.com/open-telemetry/oteps/pull/148. From this perspective, it's the "Priority", "FixedRate", and "ReservoirSize" that are involved in sampling probability. As discussed in OTEP 148 Samplers are capable of producing probabilities for fixed rate sampling, we'll just have to replace Leaky-bucket rate limiting with Varopt reservoir sampling.

I'd like to draw a non-blocking connection between the XRay sampler and the discussion about Views taking place in the Metrics SIG. Sampler configurations are like views on spans, and as long as we're able to count them, they are also views that can produce metrics. The big questions are whether this configuration syntax can become a standard for sampling in OTel's default SDKs, and if how is configuration is supplied.

anuraaga:
@jmacd Thanks - I agree that any rate limiting can be made compatible with the need for counting spans by tweaking the algorithm, perhaps to Varopt. In the meantime I'll be starting with the same as Jaeger but will be happy to change in the future.

>  The big question ia whether this configuration syntax can become a standard for sampling in OTel's default SDKs, and if so--how is configuration is supplied?

Configuration is basically just data, not that much unlike even spans - just the data that's sent is about configuring an app, not an event. I'd expect we define a proto and transmit it over gRPC just like OTLP, there doesn't seem to be a reason to break from convention. What's in the proto will be a finer nuance - I wouldn't expect to crib much from the X-Ray data model for sampling since it's not very intuitive / well designed I think. In particular, reasoning out the result of having both rate limiting and fixed rate possible at the same time is mind blowing. Adding attribute matching to the jaeger data model is probably a better start.

PR #3352, opened 2021-06-26
Migrate Metrics SDK "data' package to Attributes

jsuereth:
With the `Labels` removal in the new Metrics API, this is a first cut at changing the existing implementation in preparation for the new SDK.

What this PR brings:

- The `io.opentelmetry.sdk.metrics.data` package now uses the `io.opentelemetry.api.common.Attributes` type instead of the old `Labels` type.
- An AssertJ testing DSL for all the various types in the `*.metrics.data` package.  This is designed to:
   - Solve floating-point equality precision-in-testing concerns in one locations
   - Be resilient to order-concerns of points / attributes
   - Have easily readable expectations and error messages
-  Update all existing tests that either used `Labels` or relied on ordering/equality of `Labels` instances to use new Assertj library for metrics.

_Note:  `Summary` is effectively deprecated in the OTel Data model (for backwards compat primarily), so it remains for OpenCensus Shims (Summary is also deprecated in OpenCensus) and I did not provide convenience APIs for summary.  In the new SDK no instrument will output a summary by default._


exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/MetricAdapter.java
```diff
@@ -102,14 +101,18 @@ private static String cleanMetricName(String descriptorMetricName) {
...
-      List<String> labelNames = Collections.emptyList();
-      List<String> labelValues = Collections.emptyList();
-      Labels labels = pointData.getLabels();
-      if (labels.size() != 0) {
-        labelNames = new ArrayList<>(labels.size());
-        labelValues = new ArrayList<>(labels.size());
-
-        labels.forEach(new Consumer(labelNames, labelValues));
+      Attributes attributes = pointData.getAttributes();
+      final List<String> labelNames = new ArrayList<>(attributes.size());
```
anuraaga, line 105: Looks like we lost the (micro-)-optimization of not allocating lists on empty attributes. If intentional, I'd remove the `if (attribute.size() != 0)` check too as not needed anymore

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/MetricAdapter.java
```diff
@@ -102,14 +101,18 @@ private static String cleanMetricName(String descriptorMetricName) {
...
+      final List<String> labelValues = new ArrayList<>(attributes.size());
+      if (attributes.size() != 0) {
+        attributes.forEach(
+            (key, value) -> {
+              String sanitizedLabelName = sanitizer.apply(key.getKey());
+              labelNames.add(sanitizedLabelName);
+              // TODO: We want to create an error-log if there is overlap in toString of attribute
+              // values for the same key name.
+              labelValues.add(value == null ? "" : value.toString());
```
anuraaga, line 114: Don't believe we support null values anymore, think we may have at some point before

opencensus-shim/src/test/java/io/opentelemetry/opencensusshim/InteroperabilityTest.java
```diff
@@ -422,42 +419,49 @@ void testSupportedMetricsExportedCorrectly() {
...
-    assertThat(metric.getDescription()).isEqualTo("double gauge");
-    assertThat(metric.getUnit()).isEqualTo("ms");
-    assertThat(metric.getType()).isEqualTo(MetricDataType.DOUBLE_GAUGE);
-    assertThat(metric.getDoubleGaugeData().getPoints().size()).isEqualTo(1);
-    PointData pointData = metric.getDoubleGaugeData().getPoints().iterator().next();
-    assertThat(((DoublePointData) pointData).getValue()).isEqualTo(60);
-    assertThat(pointData.getLabels().size()).isEqualTo(0);
-
+    assertThat(metric)
+        .hasName("double_gauge")
```
anuraaga, line 423: Nice

opencensus-shim/src/test/java/io/opentelemetry/opencensusshim/InteroperabilityTest.java
```diff
@@ -422,42 +419,49 @@ void testSupportedMetricsExportedCorrectly() {
...
+        .hasDescription("double gauge")
+        .hasUnit("ms")
+        .hasDoubleGauge()
+        .points()
```
anuraaga, line 427: Tend to like not switching gears in an assertion (i.e., allowing a further assertion on the metric after this one) with something like `pointsSatisfyExactly(...)`. One reason is for the indentation to match the structure. I guess it's mostly just personal preference though so don't mind too much

sdk/metrics-testing/build.gradle.kts
```diff
@@ -0,0 +1,22 @@
+plugins {
+    id("otel.java-conventions")
+    id("otel.publish-conventions")
+}
+
+description = "OpenTelemetry SDK Testing utilities"
+otelJava.moduleName.set("io.opentelemetry.sdk.metrics-testing")
```
anuraaga, line 7: We've generally suffixed with `metrics` in all our packages (even had a gradle hack that matched the suffix before the more declarative `gradle.properties` :P). Since alpha and eventually will be merged into testing doesn't matter too much though

sdk/metrics-testing/src/main/java/io/opentelemetry/sdk/testing/assertj/metrics/MetricDataAssert.java
```diff
@@ -0,0 +1,183 @@
...
+import org.assertj.core.api.AbstractAssert;
+
+/** Test assertions for {@link MetricData}. */
+public class MetricDataAssert extends AbstractAssert<MetricDataAssert, MetricData> {
+  protected MetricDataAssert(MetricData actual) {
+    super(actual, MetricDataAssert.class);
+  }
+
+  /** Ensures the {@link Resource} associated with a metric matches the expected value. */
+  public MetricDataAssert hasResource(Resource resource) {
```
anuraaga, line 21: What do you think about just `hasResourceAttributes` and `hasResourceAttributesSatisfying`? Resource only has one field, attributes

sdk/metrics/build.gradle.kts
```diff
@@ -13,6 +15,7 @@ description = "OpenTelemetry SDK Metrics"
 otelJava.moduleName.set("io.opentelemetry.sdk.metrics")
 
 dependencies {
+    api(project(":context"))
```
anuraaga, line 18: Nit alphabetical order

sdk/metrics/build.gradle.kts
```diff
@@ -21,6 +24,7 @@ dependencies {
     testAnnotationProcessor("com.google.auto.value:auto-value")
 
     testImplementation(project(":sdk:testing"))
+    testImplementation(project(":sdk:metrics-testing"))
```
anuraaga, line 27: Nit alphabetical (unless renaming package to `metrics-testing` :P)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/MetricDataUtils.java
```diff
@@ -17,12 +19,21 @@
 final class MetricDataUtils {
   private MetricDataUtils() {}
 
+  // Temporary workaround while new API/SDK is implemented.
+  static Attributes toAttributes(Labels labels) {
+    AttributesBuilder result = Attributes.builder();
+    labels.forEach((k, v) -> result.put(k, v));
```
anuraaga, line 25:
```suggestion
    labels.forEach(result::put);
```

If you want

anuraaga (approved):
Just some small comments, no blockers. Thanks!

sdk/metrics-testing/src/main/java/io/opentelemetry/sdk/testing/assertj/metrics/MetricDataAssert.java
jsuereth: While I half-agree, in practice almost all the tests (and code) just pass `Resource` wholesale, and check for the one static instance that was given to an SDK.  Very rarely are folks checking resource detection/attribute scenarios in Metrics code.

I imagine you could add those two methods, but I don't think I'd wind up using them for most of the metrics tests.

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/MetricAdapter.java
jsuereth: should I remove the check or keep it for safety?

sdk/metrics-testing/build.gradle.kts
jsuereth: If possible, I'd like to avoid chaning the module name (and directory) because I have a slew of PRs lined up behind this one for the new metrics implementation and the least amount of major rebasing I need to do, the better.

If this were a fundamental naming issue (or something that wouldn't be resolved when metrics goes stable) I'd be happy to change it, but as things are my commits are starting to get unwieldy if I leave this two long (hence trying to piece out components like this one for review).

opencensus-shim/src/test/java/io/opentelemetry/opencensusshim/InteroperabilityTest.java
jsuereth: I hear you. I chose the methods to match existing patterns in Assertj, particularly around Iterables/Collections.  

The TL;DR: is this matches Assertj even if I'm more of a hamcrest fan myself (primarily for composability).

Want to call out two things:

1. The goal of this library is folks STOP using exact ordering where it's not important (you can choose when, but prefer `inAnyOrder` if you don't know whether order matters).   If you look at examples there's a lot of `.points().satisfiesExactlyInAnyOrder`.
2. There's a lot of Iterable matches I don't want to do a combinatorial explosion on providing, similar for Attributes.   I could take an approach here (similar to what I did for Attribtues) where you can go into a `.points` *or* use a `hasPoints` that implicitly uses "containsInAnyOrder".

opencensus-shim/src/test/java/io/opentelemetry/opencensusshim/InteroperabilityTest.java
jsuereth: Thanks!

sdk/metrics/build.gradle.kts
jsuereth: Fixed

sdk/metrics/build.gradle.kts
jsuereth: Fixed

pirgeo (commented):
There seems to be an empty file named `point.getAttributes())` in the repo root. Is this intentional?

jsuereth:
> There seems to be an empty file named `point.getAttributes())` in the repo root. Is this intentional?

@pirgeo No, that was a botched commit from windows on my part.  Must have done something dumb w/ copy-paste on the terminal.  Fixed.

sdk/metrics-testing/src/main/java/io/opentelemetry/sdk/testing/assertj/metrics/MetricDataAssert.java
jkwatson: just FYI, Resource also has `schemaUrl` now; not sure if that alters the fundamentals, though.

sdk/metrics-testing/build.gradle.kts
jkwatson: we can always go back and change module names at the end of the PR chain, if anyone feels strongly about it.

sdk/metrics-testing/src/main/java/io/opentelemetry/sdk/testing/assertj/metrics/AbstractPointDataAssert.java
```diff
@@ -0,0 +1,50 @@
...
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.sdk.metrics.data.PointData;
+import io.opentelemetry.sdk.testing.assertj.AttributesAssert;
+import io.opentelemetry.sdk.testing.assertj.OpenTelemetryAssertions;
+import org.assertj.core.api.AbstractAssert;
+import org.assertj.core.api.Assertions;
+
+/** Test assertions for {@link PointData}. */
+public class AbstractPointDataAssert<
+        PointAssertT extends AbstractPointDataAssert<PointAssertT, PointT>,
```
jkwatson, line 17: this style of type declaration isn't something we've used in the project up until now. I'm not fundamentally opposed to it, but it's definitely more verbose. @anuraaga any thoughts on this? Otherwise, I would think we'd want to go with `<A,P>` or something like that.

sdk/metrics/build.gradle.kts
```diff
@@ -14,12 +16,14 @@ otelJava.moduleName.set("io.opentelemetry.sdk.metrics")
 
 dependencies {
     api(project(":api:metrics"))
+    api(project(":context"))
```
jkwatson, line 19: do we need context at this point in the work?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/GaugeData.java
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.metrics.data;
+
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * A gauage metric point.
```
jkwatson, line 11:
```suggestion
 * A gauge metric point.
```

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/PointData.java
```diff
@@ -5,9 +5,15 @@
...
 package io.opentelemetry.sdk.metrics.data;
 
-import io.opentelemetry.api.metrics.common.Labels;
+import io.opentelemetry.api.common.Attributes;
 import javax.annotation.concurrent.Immutable;
 
+/**
+ * A point in the "Metric stream" data model.
+ *
+ * <p>This is distinguish from {@code Measurement} in that it may have aggregated data, and has its
```
jkwatson, line 14:
```suggestion
 * <p>This is distinguished from {@code Measurement} in that it may have aggregated data, and has its
```

jkwatson:
I will admit I did not look at every single converted assertion, but this looks pretty good. Thanks for all the effort! Just a couple tiny nits and one question about generic type naming.

sdk/metrics-testing/src/main/java/io/opentelemetry/sdk/testing/assertj/metrics/AbstractPointDataAssert.java
jsuereth: If you think A/P is more readable, happy to change.   I chose what I did because I kept messing up the letters, otherwise.

sdk/metrics/build.gradle.kts
jsuereth: No, think this is an artifact of how I fragmented the branch. Removed.

sdk/metrics-testing/src/main/java/io/opentelemetry/sdk/testing/assertj/metrics/AbstractPointDataAssert.java
jkwatson: let's not block this PR on it. It's a simple thing to change in the future.

jkwatson (approved):

PR #3353, opened 2021-06-27
Wire Exemplars into the metrics.data package

jsuereth:
Based on #3352

- Adds Exemplars to the data package
- Adds exemplars to OTLP export (if they exist)
- Adds exemplars to Prometheus export (if they exist)
- Add some tests and helper methods


exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/MetricAdapter.java
```diff
@@ -187,14 +203,20 @@ private static void addHistogramSamples(
     List<Long> counts = doubleHistogramPointData.getCounts();
     for (int i = 0; i < counts.size(); i++) {
       List<String> labelValuesWithLe = new ArrayList<>(labelValues.size() + 1);
+      double boundary = i < boundaries.size() ? boundaries.get(i) : Double.POSITIVE_INFINITY;
```
jkwatson, line 206:
```suggestion
      double boundary = (i < boundaries.size()) ? boundaries.get(i) : Double.POSITIVE_INFINITY;
```

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/MetricAdapter.java
```diff
@@ -224,5 +246,47 @@ private static int estimateNumSamples(int numPoints, MetricDataType type) {
...
+          exemplar.getRecordTimeNanos(),
+          "traceid",
+          exemplar.getTraceId(),
+          "spanid",
+          exemplar.getSpanId());
+    }
+    return new io.prometheus.client.exemplars.Exemplar(valueOf(exemplar));
+  }
+
+  private static Optional<Exemplar> filterExemplars(
```
jkwatson, line 274: what do you think about moving this directly below where it's called? took me a bit to track it down while reading here.

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/MetricAdapter.java
```diff
@@ -224,5 +246,47 @@ private static int estimateNumSamples(int numPoints, MetricDataType type) {
...
+  }
+
+  private static double valueOf(Exemplar exemplar) {
+    double value = 0;
+    if (exemplar instanceof LongExemplar) {
+      value = ((LongExemplar) exemplar).getValue();
+    } else if (exemplar instanceof DoubleExemplar) {
+      value = ((DoubleExemplar) exemplar).getValue();
+    }
+    return value;
```
jkwatson, line 288: this method makes me very sad. If only boxing into Number weren't an unacceptable cost. :(

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleExemplar.java
```diff
@@ -0,0 +1,31 @@
...
+import com.google.auto.value.AutoValue;
+import io.opentelemetry.api.common.Attributes;
+import javax.annotation.concurrent.Immutable;
+
+/** An {@link Exemplar} with {@code double} measurments. */
+@Immutable
+@AutoValue
+public abstract class DoubleExemplar implements Exemplar {
+
+  public static DoubleExemplar create(
```
jkwatson, line 17: we need some javadoc here, since it's a public method.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoublePointData.java
```diff
@@ -15,10 +17,20 @@
...
-public abstract class DoublePointData implements PointData {
+public abstract class DoublePointData implements SampledPointData {
+
   public static DoublePointData create(
       long startEpochNanos, long epochNanos, Attributes attributes, double value) {
-    return new AutoValue_DoublePointData(startEpochNanos, epochNanos, attributes, value);
+    return create(startEpochNanos, epochNanos, attributes, value, Collections.emptyList());
+  }
+
+  public static DoublePointData create(
```
jkwatson, line 27: javadoc needed here.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/Exemplar.java
```diff
@@ -0,0 +1,43 @@
...
+   */
+  Attributes getFilteredAttributes();
+
+  /** Returns the timestamp in nanos when measurement was collected. */
+  long getRecordTimeNanos();
+
+  /**
+   * (Optional) Span ID of the exemplar trace.
+   *
+   * <p>span_id may be missing if the measurement is not recorded inside a trace or the trace was
```
jkwatson, line 32: why the switch to snake_case here?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/Exemplar.java
```diff
@@ -0,0 +1,43 @@
...
+   * not sampled.
+   */
+  String getSpanId();
+  /**
+   * (Optional) Trace ID of the exemplar trace.
+   *
+   * <p>trace_id may be missing if the measurement is not recorded inside a trace or if the trace is
```
jkwatson, line 39:
```suggestion
   * <p>traceId may be missing if the measurement is not recorded inside a trace or if the trace is
```

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/LongExemplar.java
```diff
@@ -0,0 +1,30 @@
...
+import com.google.auto.value.AutoValue;
+import io.opentelemetry.api.common.Attributes;
+import javax.annotation.concurrent.Immutable;
+
+/** An {@link Exemplar} with {@code long} measurments. */
+@Immutable
+@AutoValue
+public abstract class LongExemplar implements Exemplar {
+
+  public static LongExemplar create(
```
jkwatson, line 17: javadoc needed here.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/LongPointData.java
```diff
@@ -30,6 +32,15 @@
 
   public static LongPointData create(
       long startEpochNanos, long epochNanos, Attributes attributes, long value) {
-    return new AutoValue_LongPointData(startEpochNanos, epochNanos, attributes, value);
+    return create(startEpochNanos, epochNanos, attributes, value, Collections.emptyList());
+  }
+
+  public static LongPointData create(
```
jkwatson, line 38: javadoc needed here (and on the method above, while we're at it)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/SampledPointData.java
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+package io.opentelemetry.sdk.metrics.data;
+
+import java.util.Collection;
+import javax.annotation.concurrent.Immutable;
+
+/** A point that includes (optional) exemplars which lead to its creation. */
+@Immutable
+public interface SampledPointData extends PointData {
```
jkwatson, line 13: is this necessary/desirable only because SummaryPointData isn't sampled? If so, I think we could just move the method up to PointData itself since the SummaryPointData is legacy anyway. Thoughts?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/Exemplar.java
jsuereth: Copy-pasta from the specification/proto.  I'll make it look java, wish I could deep link into the proto though.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/SampledPointData.java
jsuereth: Yeah, we could move it up.  I'm fine making that change, should I wait for more input or just do that now?

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/MetricAdapter.java
jsuereth: Agreed

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/MetricAdapter.java
jsuereth: Done

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleExemplar.java
jsuereth: Fixed

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoublePointData.java
jsuereth: Interesting, previous code did NOT have javadoc here, added to both.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/LongExemplar.java
jsuereth: fixed

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/LongPointData.java
jsuereth: done

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/SampledPointData.java
jkwatson: I'd go ahead and move it up. Makes a code a bit simpler all-in-all.

jkwatson:
Looks good to me! I don't know if the mapping to prometheus is good, bad or ugly, but it all looks pretty decent to me.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/SampledPointData.java
jsuereth: done

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/MetricAdapter.java
```diff
@@ -224,5 +263,30 @@ private static int estimateNumSamples(int numPoints, MetricDataType type) {
...
+        .map(e -> new Sample(name, labelNames, labelValues, value, toPrometheusExemplar(e)))
+        .orElseGet(() -> new Sample(name, labelNames, labelValues, value));
+  }
+
+  private static io.prometheus.client.exemplars.Exemplar toPrometheusExemplar(Exemplar exemplar) {
+    // TODO: Keep filtered attributes for prometheus?
+    if (exemplar.getSpanId() != null && exemplar.getTraceId() != null) {
+      return new io.prometheus.client.exemplars.Exemplar(
+          valueOf(exemplar),
+          exemplar.getRecordTimeNanos(),
```
brian-brazil, line 282: This should be in milliseconds

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/MetricAdapter.java
```diff
@@ -224,5 +263,30 @@ private static int estimateNumSamples(int numPoints, MetricDataType type) {
...
+          "traceid",
```
brian-brazil, line 283: trade_id and span_id are what client_java uses.

jsuereth: Is there a standard somewhere specifying this I can point to?  If not I'll just link against the Java source.

P.S. Thanks for the review!!!

sdk/metrics-testing/src/main/java/io/opentelemetry/sdk/testing/assertj/metrics/AbstractPointDataAssert.java
```diff
@@ -47,4 +49,24 @@ public AttributesAssert attributes() {
     isNotNull();
     return OpenTelemetryAssertions.assertThat(actual.getAttributes());
   }
+
+  /** Returns convenience API to assert against the {@code exemplars} field. */
+  public AbstractIterableAssert<?, ? extends Iterable<? extends Exemplar>, Exemplar, ?>
```
anuraaga, line 54: This goes back to my comment on the other PR, at the time it seemed to still be ok having only one, but now we have `attributes` and `exemplars` which can't both be called on the same point. This makes me strongly advocate for `hasExemplarsSatisfying` type of methods instead, or otherwise delete them for now instead of locking them into the API with the next release.

sdk/metrics-testing/src/test/java/io/opentelemetry/sdk/testing/assertj/metrics/MetricAssertionsTest.java
```diff
@@ -90,8 +93,14 @@
...
 
+  private static final DoubleExemplar DOUBLE_EXEMPLAR =
+      DoubleExemplar.create(Attributes.empty(), 0, "span", "trace", 1.0);
+
   private static final DoublePointData DOUBLE_POINT_DATA =
-      DoublePointData.create(1, 2, Attributes.empty(), 3.0);
+      DoublePointData.create(1, 2, Attributes.empty(), 3.0, Collections.emptyList());
+
+  private static final DoublePointData DOUBLE_POINT_DATA_WITH_EXEMPLAR =
+      DoublePointData.create(1, 2, Attributes.empty(), 3.0, Arrays.asList(DOUBLE_EXEMPLAR));
```
anuraaga, line 103: `Collections.singletonList` - not a big deal but may as well make intellij not mark this out

exporters/prometheus/src/test/java/io/opentelemetry/exporter/prometheus/MetricAdapterTest.java
```diff
@@ -384,4 +415,41 @@ void toMetricFamilySamples() {
...
+        if (sample.exemplar != null) {
+          StringBuilder sb = new StringBuilder(sample.toString());
+          sb.append(" Exemplar=").append(exemplarToString(sample.exemplar));
+          return sb.toString();
+        }
+      }
+      return object.toString();
+    }
+    /** Convert an exemplar into a human readable string. */
+    private static String exemplarToString(Exemplar exemplar) {
```
anuraaga, line 439: Sent https://github.com/prometheus/client_java/pull/671

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/Exemplar.java
```diff
@@ -0,0 +1,43 @@
...
+   */
+  Attributes getFilteredAttributes();
+
+  /** Returns the timestamp in nanos when measurement was collected. */
+  long getRecordTimeNanos();
+
+  /**
+   * (Optional) Span ID of the exemplar trace.
+   *
+   * <p>Span ID may be missing if the measurement is not recorded inside a trace or the trace was
```
anuraaga, line 32: `@Nullable` for both and `{@code null}` instead of `missing` in the javadocs

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/LongExemplar.java
```diff
@@ -0,0 +1,40 @@
...
+@AutoValue
+public abstract class LongExemplar implements Exemplar {
+
+  /**
+   * Construct a new exemplar.
+   *
+   * @param filteredAttributes The set of {@link Attributes} not already associated with the {@link
+   *     PointData}.
+   * @param recordTimeNanos The time when the sample qas recorded in nanoseconds.
+   * @param spanId (optional) The associated {@code SpanId}
```
anuraaga, line 23: Perhaps we can `compileOnly(":api:all")` to be able to `{@link SpanId}`. Otherwise, having this in `@code` doesn't seem particularly helpful

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/PointData.java
```diff
@@ -38,4 +39,8 @@
    * @return the attributes associated with this {@code Point}.
    */
   Attributes getAttributes();
+  /**
+   * (Optional) List of exemplars collected from measurements that were used to form the data point.
```
anuraaga, line 43:
```suggestion
   * List of exemplars collected from measurements that were used to form the data point.
```

Optional conflates with the ability to be null. We generally don't call out the fact that this could be empty but you can add wording using `empty` if you want

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/PointData.java
```diff
@@ -38,4 +39,8 @@
...
+   */
+  Collection<Exemplar> getExemplars();
```
anuraaga, line 45: We use `List` for similar types in `SpanData` let's go with it here, there doesn't seem to be much gain in the more generic type

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/Exemplar.java
```diff
@@ -0,0 +1,43 @@
...
+public interface Exemplar {
+  /**
+   * The set of key/value pairs that were filtered out by the aggregator, but recorded alongside the
+   * original measurement. Only key/value pairs that were filtered out by the aggregator should be
+   * included
+   */
+  Attributes getFilteredAttributes();
+
+  /** Returns the timestamp in nanos when measurement was collected. */
+  long getRecordTimeNanos();
```
anuraaga, line 27: How about `getEpochNanos()` to match `PointData`? This is semantically the same thing right?

anuraaga: Though renaming all to `getTimestampNanos()` would probably be good IMO as a separate discussion

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoublePointData.java
```diff
@@ -16,9 +18,38 @@
 @Immutable
 @AutoValue
 public abstract class DoublePointData implements PointData {
+
+  /**
+   * Creates a {@link DoublePointData}.
+   *
+   * @param startEpochNanos (optional) The starting time for the period where this point was
```
anuraaga, line 25: Is this really optional? If so we need to document something like `0 indicates no starting time.`

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/MetricAdapter.java
brian-brazil: Not that I know of, the Java client is the first one and that's what it's using. The OM spec uses trace_id in an example.

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/MetricAdapter.java
```diff
@@ -275,14 +276,14 @@ private static Sample createSample(
   }
 
   private static io.prometheus.client.exemplars.Exemplar toPrometheusExemplar(Exemplar exemplar) {
-    // TODO: Keep filtered attributes for prometheus?
     if (exemplar.getSpanId() != null && exemplar.getTraceId() != null) {
       return new io.prometheus.client.exemplars.Exemplar(
           valueOf(exemplar),
-          exemplar.getRecordTimeNanos(),
-          "traceid",
+          // Convert to ms for prometheus, truncate nanosecond preceision.
```
brian-brazil, line 282: May I ask how you're getting ns resolution timestamps? System.currentTimeMillis() is all that Java provides in this area.

anuraaga: `Instant.now` gets us microseconds on Java 9+, we use it here

https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/common/src/main/java9/io/opentelemetry/sdk/internal/Java9VersionSpecific.java#L39

We indeed don't have nanos for arbitrary wall time but do have the notion of "nanos anchored to the beginning of a Span"

https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/common/src/main/java/io/opentelemetry/sdk/internal/MonotonicClock.java

So OTel uses nanos for all its timestamps as convention. For exemplars with a span ID, it seems the timestamp could be derived from the span start time, but that could make it out of sync with the timestamp of the raw data recording so I guess it's not something we'd do.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/Exemplar.java
jsuereth: This field matches the OTLP proto field.  I think there's a good pairing between the two as It's easy to see how java is representing OTLP, and how the internal datamodel can be used for exports.  If you'd like to rename, perhaps discussing there would be the best option.  For now, I'd prefer to keep the `metrics.data` package in line with OTLP / OTel metrics data model.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoublePointData.java
jsuereth: According to the data model (which this is copy-pasted from) yes.

From and SDK standpoint, we should never export a point without a starting time from our own instruments.   The only place we expect this to happen is in the collector (when adapting other metric formats).   I can update this with that nuance.

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/MetricAdapter.java
jsuereth: We haven't actually figured out how to record the timestamp just yet.   OTLP / our metrics data model consistently wants it in `ns`, so that's how we plan to track it here.   As @anuraaga states, there are options in Otel for finer-grained resolution, particularly for exemplars as we sample those and COULD use the `MonotonicClock` of the associated `Span`.

This PR does NOT introduce sampling of exemplars, only the export from the data model.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/PointData.java
jsuereth: I was planning to use a highly concurrent-friendly collection type that is NOT a `List` here in the implementation.   If it's all the same, I don't see a big gain for this being `List`, but I see a big downside in choice of collection for sampling.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/PointData.java
jsuereth: The comment is directly copy-pasted from the spec/proto.  I can remove it if that's the convention in Java.

exporters/prometheus/src/test/java/io/opentelemetry/exporter/prometheus/MetricAdapterTest.java
```diff
@@ -347,28 +358,48 @@ void toSamples_HistogramPoints() {
...
+                    ImmutableList.of(
+                        LongExemplar.create(
+                            Attributes.empty(),
+                            /*recordTime=*/ 0,
+                            "other_span_id",
+                            "other_trace_id",
+                            /*value=*/ 0),
+                        LongExemplar.create(
+                            Attributes.empty(),
+                            /*recordTime=*/ 0,
```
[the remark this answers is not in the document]
jsuereth: done!  Missed this after fixing the earlier comment, thanks for calling it out.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/LongExemplar.java
jsuereth: Removed `@code` for now.  We can revisit if we want to document any specific string formatting concerns related to using SpanId/TraceId

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/MetricAdapter.java
```diff
@@ -187,17 +204,40 @@ private static void addHistogramSamples(
...
   }
 
+  private static Optional<Exemplar> filterExemplars(
+      Collection<Exemplar> exemplars, double min, double max) {
+    return exemplars.stream()
+        .filter(exemplar -> valueOf(exemplar) <= max && valueOf(exemplar) > min)
+        .findAny();
+  }
+
+  private static double valueOf(Exemplar exemplar) {
```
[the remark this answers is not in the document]
jsuereth: Good idea, done.

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/MetricAdapter.java
```diff
@@ -187,17 +204,40 @@ private static void addHistogramSamples(
...
+          createSample(
+              name + SAMPLE_SUFFIX_BUCKET,
+              labelNamesWithLe,
+              labelValuesWithLe,
+              cumulativeCount,
+              filterExemplars(doubleHistogramPointData.getExemplars(), lower, boundary)));
     }
   }
 
+  private static Optional<Exemplar> filterExemplars(
```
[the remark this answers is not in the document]
jsuereth: Why check the length?  That can be a rather expensive operation, depending on the collection used.   Updated to use private methods.   I think readability is debatable, but I can see avoiding stream overhead.

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/MetricAdapter.java
```diff
@@ -187,17 +204,40 @@ private static void addHistogramSamples(
...
       List<String> labelValuesWithLe = new ArrayList<>(labelValues.size() + 1);
+      double boundary = (i < boundaries.size()) ? boundaries.get(i) : Double.POSITIVE_INFINITY;
       labelValuesWithLe.addAll(labelValues);
-      labelValuesWithLe.add(
-          doubleToGoString(i < boundaries.size() ? boundaries.get(i) : Double.POSITIVE_INFINITY));
+      labelValuesWithLe.add(doubleToGoString(boundary));
 
       cumulativeCount += counts.get(i);
+      // TODO: We need to select our exemplar samples FROM the bucket counts and put them in here.
+      double lower = i > 0 ? boundaries.get(i - 1) : Double.NEGATIVE_INFINITY;
```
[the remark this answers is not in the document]
jsuereth: Added these methods directly to `DoubleHistogramDataPoint`, to just improve readability in general.   Agree this is likely an easily botched or hard-to-read portion of our data model.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MetricAdapter.java
```diff
@@ -269,10 +279,44 @@ private static AggregationTemporality mapToTemporality(
...
+    exemplar
+        .getFilteredAttributes()
+        .forEach(
+            (key, value) ->
+                builder.addFilteredAttributes(CommonAdapter.toProtoAttribute(key, value)));
+    if (exemplar instanceof LongExemplar) {
+      builder.setAsInt(((LongExemplar) exemplar).getValue());
+    } else if (exemplar instanceof DoubleExemplar) {
+      builder.setAsDouble(((DoubleExemplar) exemplar).getValue());
+    }
```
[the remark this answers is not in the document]
jsuereth: Logged for now.  It's *possible* we add a new type in an SDK and have SDK-exporter version mismatch.  Would rather not throw unless we absolutely need to.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/PointData.java
anuraaga: Yeah - optional in the spec generally refers to whether a method exists at all. Given the method is here, now the SDK has implemented an optional part of the spec. Spec should generally not be copied into the SDK since it's target audience is SDK authors, not end users.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/PointData.java
anuraaga: Can you describe the structure more to understand better? And is it that hard to present a List view of it on export (when presumably everything is mostly frozen?). Consistency within the SDK is better unless there is a big downside

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/Exemplar.java
anuraaga: Thanks for reminding of OTLP. So yeah `getEpochNanos` is correct I guess since it's how we translate OTLP `time_unix_nanos` for all the other data fields in this SDK I think

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleExemplar.java
```diff
@@ -0,0 +1,46 @@
...
+   * Construct a new exemplar.
+   *
+   * @param filteredAttributes The set of {@link Attributes} not already associated with the {@link
+   *     PointData}.
+   * @param recordTimeNanos The time when the sample qas recorded in nanoseconds.
+   * @param spanId (optional) The associated SpanId.
+   * @param traceId (optional) The associated TraceId.
+   * @param value The value recorded.
+   */
+  public static DoubleExemplar create(
```
jonatan-ivanov, line 27: I'm not sure what do I walk into with this but would not a generic `Exemplar<T extends Number>` work?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoublePointData.java
```diff
@@ -16,9 +18,40 @@
...
+   *
+   * @param startEpochNanos The starting time for the period where this point was sampled. Note:
+   *     While start time is optional in OTLP, all SDKs should produce it for all their metrics, so
+   *     it is required here.
+   * @param epochNanos The ending time for the period when this value was sampled.
+   * @param attributes The set of attributes associated with this point.
+   * @param value The value that was sampled.
+   * @param exemplars A collection of interesting sampled values from this time period.
+   */
+  public static DoublePointData create(
```
jonatan-ivanov, line 48: Where will this be called and how (i.e.: who will create the collection of exemplars)?

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MetricAdapter.java
```diff
@@ -199,6 +213,7 @@ private static AggregationTemporality mapToTemporality(
           .getAttributes()
           .forEach(
               (key, value) -> builder.addAttributes(CommonAdapter.toProtoAttribute(key, value)));
+      longPoint.getExemplars().forEach(e -> builder.addExemplars(toExemplar(e)));
```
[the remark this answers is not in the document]
jsuereth: I'm not as familiar with AutoValue (I'm a Scala guy), is this easy to do?   Do you have a doc link I could check out?

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MetricAdapter.java
```diff
@@ -269,10 +285,48 @@ private static AggregationTemporality mapToTemporality(
...
+    if (exemplar.getTraceId() != null) {
+      builder.setTraceId(convertTraceId(exemplar.getTraceId()));
+    }
+    exemplar
+        .getFilteredAttributes()
+        .forEach(
+            (key, value) ->
+                builder.addFilteredAttributes(CommonAdapter.toProtoAttribute(key, value)));
+    if (exemplar instanceof LongExemplar) {
+      builder.setAsInt(((LongExemplar) exemplar).getValue());
```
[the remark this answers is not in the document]
jsuereth: Two things:
1. I've been told this project is trying to avoid `.stream()` overhead, maybe @anuraaga or @jkwatson can comment?
2. This is a protocol buffer builder, I can't touch how it works (it's auto-generated code) :)

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MetricAdapter.java
```diff
@@ -269,10 +285,48 @@ private static AggregationTemporality mapToTemporality(
...
       result.add(builder.build());
     }
     return result;
   }
 
+  static Exemplar toExemplar(io.opentelemetry.sdk.metrics.data.Exemplar exemplar) {
+    // TODO - Use a thread local cache for spanid/traceid -> byte conversion.
+    Exemplar.Builder builder = Exemplar.newBuilder();
+    builder.setTimeUnixNano(exemplar.getRecordTimeNanos());
+    if (exemplar.getSpanId() != null) {
```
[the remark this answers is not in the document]
jsuereth: Besides protecting `convertSpan` from `null`, I can't remember off the top of my head the behavior for `null` strings in the proto libraries. I think it's ok.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleExemplar.java
jsuereth: Maybe @jkwatson can comment on this?   I tried to match the current shape of the Java code, and Long vs. Double staying unboxed is a theme across all the metrics, so I stuck with it.  It has a natural pairing with the protocol buffers we plan to send this data as.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoublePointData.java
jsuereth: That's a separate PR.   

Preview here: https://github.com/jsuereth/opentelemetry-java/blob/wip-metrics-api/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/MetricDataUtils.java#L85

This PR is ONLY the `DataModel -> Exporter` code (I.e. the stable portions of the specification).

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/PointData.java
jsuereth: So two things:

1. I actually want this to be `Iterable` (I forked this PR earlier than the prototype), which I'll update if you agree.
2. I'm using a linked-list that allows prepend-only compare-and-swap behavior for lock-less addition.   This *is* a list where order is reversed from expectations *and* the `get(index)` method would be highly inefficient to use.   The primary reason to use this list is to have a highly-concurrent append directly from the sampler and it's immutable so easy to share/export to OTLP.

You can see an earlier prototype here: https://github.com/jsuereth/opentelemetry-java/blob/wip-metrics-api/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/aggregator/ExemplarList.java

I can implement the List interface (with lots of exception throwing) but I feel it's pretty disingenuous to give people a "get" method that's highly inefficient.  Probably too much "Scala" collection design in me.   PTAL and let me know your thoughts here.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/Exemplar.java
jsuereth: done!

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MetricAdapter.java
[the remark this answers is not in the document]
jonatan-ivanov: Got it, could you please mark these comments as resolved (this and the previous two)?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleExemplar.java
jkwatson: We've shied away from using generic numbers for hot-path, high usage scenarios to avoid the extra object allocation required to store the information. Exemplars should be much lower volume, though, so this might be a case where we could gain readability/maintainability by eating an extra allocation to box the numbers.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/PointData.java
anuraaga: Yeah I wouldn't worry too much about deceptive get methods which have already been presented to us by Java in `LinkedList`.

But given metrics is a concurrent thing, while spans we expect generally not to be, I can see where using different collections can come in handy so a more generic type seems ok. Given the choice between `Collection` and `Iterable` I would pick `Collection` - Java doesn't even give easy conversion to `Stream` for `Iterable` so it's quite an anemic return type for a public API.

By the way, I don't know if you already looked at it, but if it's possible to use, we should add to our shading-in of JCTools a structure instead of reimplementing, this one seems similar to what you describe.

https://github.com/JCTools/JCTools/blob/master/jctools-core/src/main/java/org/jctools/queues/MpscLinkedQueue.java

Currently we shade into the trace artifact but no reason not to move that to common

https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk/trace-shaded-deps

sdk/metrics-testing/src/main/java/io/opentelemetry/sdk/testing/assertj/metrics/AbstractPointDataAssert.java
anuraaga: I had forgotten the metrics-testing artifact is still alpha, so there is no urgent need to delete methods

exporters/prometheus/src/test/java/io/opentelemetry/exporter/prometheus/MetricAdapterTest.java
```diff
@@ -384,4 +415,41 @@ void toMetricFamilySamples() {
...
+      }
+      if (object instanceof Sample) {
+        Sample sample = (Sample) object;
+        if (sample.exemplar != null) {
+          StringBuilder sb = new StringBuilder(sample.toString());
+          sb.append(" Exemplar=").append(exemplarToString(sample.exemplar));
+          return sb.toString();
+        }
+      }
+      return object.toString();
```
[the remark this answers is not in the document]
jsuereth: If they were collpased it means I thought I had fixed them OR got confused with github UI :)  Let me know what they were and I'll fix em up!

exporters/prometheus/src/test/java/io/opentelemetry/exporter/prometheus/MetricAdapterTest.java
jsuereth: YAY!

sdk/metrics-testing/src/main/java/io/opentelemetry/sdk/testing/assertj/metrics/AbstractPointDataAssert.java
jsuereth: TL;DR; I'd rather not detract from the important bits of the CL for cosmetic testing concerns, so if you really want to remove it, that's fine, but i"d ask you to wait for the new metrics SDK implementation to land. 

That said, here's my opinion:

The convention of having convenient mechanisms of altering the "Assertion" class is core to Assertj design right now, and we have to deal with.  E.g. AbstractIterable supports an "extracting" method that flips the thing you're asserting on in an incompatible way.  I've made use of this in some scenarios and it's highly convenient.   Even if you don't allow these methods in OTel-provided APIs, they still exist (and are expected) from upstream Assertj.  Do you plan to prevent their usage as well?

If this is just a matter of my choosing bad names, I'm used to dealing with my poor naming and am happy to fix it.  However the notion that you shouldn't flip the underlying assertion class.... Well yes it's not ideal but that ship has sailed in choosing Assertj.  In fact, sometimes it's really important convenient.  Specifically, you probably want to go take a second look here: https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics-testing/src/main/java/io/opentelemetry/sdk/testing/assertj/metrics/MetricDataAssert.java#L99

When it comes to readability, feel free to look at usages like this: https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/DoubleCounterSdkTest.java#L73

Again, happy to remove all these methods if you think they'll lead to major issues in practice, but I don't really think that they hurt readability or correctness.   They *do* have a wart if you want to look at two collections in the same assert, but then you can just go to the more verbose syntax.

sdk/metrics-testing/src/test/java/io/opentelemetry/sdk/testing/assertj/metrics/MetricAssertionsTest.java
jsuereth: Ah, I'm using vscode.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/PointData.java
jsuereth: Alright I'll go with List<> then for consistency and just document the hell out of concurrency/efficiency going forward.

Thanks for the link to the MPscLinkedQueue!  We don't really want a queue, and we may wind up with a "keep-latest-but-have-slots" array-of-atomic-references going forward.  Once the exemplar sampling is finalized, we'll have a beter idea what we need (see: https://github.com/open-telemetry/opentelemetry-specification/issues/1797#issuecomment-876431458)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleExemplar.java
jsuereth: Alright, going to stick with existing patterns.

jkwatson (approved):
Thanks!

anuraaga (approved):
Thanks!

PR #3355, opened 2021-06-28
Add reservoir borrowing and statistics collection to x-ray sampler.

anuraaga:
@jkwatson Hope it's ok to keep on sending PRs here until the contrib repo [permissions](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/40#issuecomment-868208319) issue is resolved (which you can help with during US time zone to get this code moved sooner :P).

@willarmiros Welcome back from vacation, if you can give a skim of the merged code as well would be great.


sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/SamplingRuleApplier.java
```diff
@@ -50,7 +57,21 @@
...
+  private final Statistics statistics;
+
+  SamplingRuleApplier(String clientId, GetSamplingRulesResponse.SamplingRule rule) {
+    this.clientId = clientId;
+    ruleName = rule.getRuleName();
+    if (rule.getReservoirSize() > 0) {
+      // Until calling GetSamplingTargets, the default is to borrow 1/s if reservoir size is
+      // positive.
+      reservoirSampler = new RateLimitingSampler(1);
+      borrowing = true;
```
anuraaga, line 69: In a separate (and hopefully final) PR, I will add support for "updating" a rule based on a sampling target - what X-Ray returns in real time based on the statistics collected from all client IDs (processes). That will be the case where reservoir sampler is set but borrowing is false.

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/SamplingRuleApplier.java
```diff
@@ -251,4 +305,14 @@ public String toString() {
       return pattern.toString();
     }
   }
+
+  // We keep track of sampling requests and decisions to report to X-Ray to allow it to allocate
+  // quota from the central reservoir. We do not lock around updates because sampling is called on
```
anuraaga, line 310: Different from x-ray SDK but IMO much better

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/AwsXrayRemoteSampler.java
```diff
@@ -31,6 +33,18 @@
   private static final String WORKER_THREAD_NAME =
       AwsXrayRemoteSampler.class.getSimpleName() + "_WorkerThread";
 
+  // Unique per-process client ID, generated as a random string.
+  private static final String CLIENT_ID;
+
+  static {
```
kubawach, line 39: NIT - I'd put it as a private static method. Too much logic for a static init block, looks like an orphaned code ;)

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/SamplingRuleApplier.java
```diff
@@ -40,7 +43,11 @@
     XRAY_CLOUD_PLATFORM = Collections.unmodifiableMap(xrayCloudPlatform);
   }
 
+  private final String clientId;
+  private final String ruleName;
```
kubawach, line 47: As far as I can tell ruleName is used only for statistics purposes. Therefore could be moved to that class (set in constructor). Then, in order to reduce some boilerplate builder code SamplingStatisticsDocument.Builder could be manufactured out of Statistics class.

kubawach (approved):

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/AwsXrayRemoteSampler.java
```diff
@@ -31,6 +33,18 @@
...
+    SecureRandom rand = new SecureRandom();
+    byte[] bytes = new byte[12];
+    rand.nextBytes(bytes);
+    char[] clientIdChars = new char[24];
+    OtelEncodingUtils.bytesToBase16(bytes, clientIdChars, 12);
```
jkwatson, line 44: you ok depending on internal here?

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/SamplingRuleApplier.java
```diff
@@ -114,8 +137,39 @@ SamplingResult shouldSample(
...
-    return fixedRateSampler.shouldSample(
-        parentContext, traceId, name, spanKind, attributes, parentLinks);
+    // Incrementing requests first ensures sample / borrow rate are positive.
+    statistics.requests.increment();
+    SamplingResult result =
+        reservoirSampler.shouldSample(
+            parentContext, traceId, name, spanKind, attributes, parentLinks);
+    if (result.getDecision() != SamplingDecision.DROP) {
+      // We use the result from the reservoir sampler if it worked.
+      if (borrowing) {
```
jkwatson, line 147: just curious...what does "borrowing" mean?

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/SamplingRuleApplier.java
```diff
@@ -114,8 +137,39 @@ SamplingResult shouldSample(
...
+        statistics.borrowed.increment();
+      }
+      statistics.sampled.increment();
+      return result;
+    }
+    result =
```
jkwatson, line 153: I'm confused (maybe it's the ridiculous heat in Portland right now). do I read this correctly that if the reservoir sampler says "DROP", you get a 2nd opinion from the fixedRateSampler?

willarmiros (commented):
Generally lgtm apart from the 5% rate default rule implementation, I can see there's quite a bit of prior work that I missed so I'm sure I'm just not seeing it.

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/AwsXrayRemoteSampler.java
anuraaga: Sure

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/SamplingRulesSampler.java
```diff
@@ -27,6 +27,7 @@
   private final SamplingRuleApplier[] ruleAppliers;
 
   SamplingRulesSampler(
```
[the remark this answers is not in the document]
anuraaga: The class isn't public so the name isn't that important. But agree it can be consistent with getDescription so went with that

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/SamplingRuleApplier.java
willarmiros: Just a nice to have & consistent with the current X-Ray SDK behavior AFAIK. Not blocking on this PR.

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/SamplingRuleApplier.java
```diff
@@ -50,7 +57,21 @@
...
   private final Matcher resourceArnMatcher;
 
-  SamplingRuleApplier(GetSamplingRulesResponse.SamplingRule rule) {
+  private final Statistics statistics;
+
+  SamplingRuleApplier(String clientId, GetSamplingRulesResponse.SamplingRule rule) {
+    this.clientId = clientId;
+    ruleName = rule.getRuleName();
+    if (rule.getReservoirSize() > 0) {
+      // Until calling GetSamplingTargets, the default is to borrow 1/s if reservoir size is
```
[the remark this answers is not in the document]
willarmiros: Ah I see - I forgot the server also returns the default rule. I was thinking about the localized default sampling rule, which is the same as a the default rule returned by the server, but is used as the initial sampler in X-Ray SDKs. This probably isn't too important since we should very quickly get the default rule from server, but we should err on the side of caution and still make the initial 0.05 as well (say a customer misconfigures their collector so it can't get rules from server, and is expecting a default 5% rate but gets 100% instead).

willarmiros:
One other thing from skimming the code - in the RemoteSamplerBuilder, you set the default polling interval to 300 seconds:

https://github.com/open-telemetry/opentelemetry-java/blob/f83a9531be3d57a5f71a10398e72f879977afc0c/sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/AwsXrayRemoteSamplerBuilder.java#L20

Two things: 1) we should have a 5 second jitter applied to avoid overwhelming the backend with concurrent requests, and 2) I'm sure you're going to address this in your next PR, but just as a heads up polling for sampling targets has a much shorter interval of 10 seconds, and a 100 ms jitter. See [X-Ray code](https://github.com/aws/aws-xray-sdk-java/blob/9b724620c126b05313f4c97d18853bbb469e9ba5/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/sampling/pollers/TargetPoller.java#L38).

willarmiros (approved):
LGTM for these changes, my other comments can be addressed in separate PRs.

jkwatson (approved):

anuraaga:
Yup will be adding jitter in next PR

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/SamplingRuleApplier.java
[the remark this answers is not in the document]
anuraaga: Thanks! I was trying to come up with where the default of 0.05 is surfaced to customers (if it's not in any doc than the number could be anything) but realized it's surfaced as the Default on the console so will definitely make sense to go with it.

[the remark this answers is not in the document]
willarmiros: For reference, here it is in the docs (spelled out instead of numerical, of course 🙃)

https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-sampling

Issue #3354, opened 2021-06-28
Add factories for SamplingResult


anuraaga:
We can currently only implement a Sampler by calling `SamplingResult.create`. For results without attributes, we do return singletons from this method so it doesn't have any real overhead. But it's somewhat annoying to have to call `create` with a parameter and not so intuitive - every time I do it I still ctrl-click through the method to confirm it's not creating a new object. The alternative of creating a `private static final` in the Sampler seems redundant vs us providing some factories

`SamplingResult.recordAndSample()`
`SamplingResult.recordOnly()`
`SamplingResult.drop()`

Rocksnake:
My understanding of your point of view is to divide the `oncreate` method into three to obtain the corresponding instance separately, and ensure that there is only one instance globally. To call `oncreate`with a parameter is implemented through these three split methods.

Is my understanding correct? If it is, I am very glad to do it. Could you can assign it to me. If not, can you give me some guidance.

anuraaga:
For backwards compatibility we can't remove the current `create` so would just add these new methods. We already have the singletons as returned by `create` so just have to return those from the factories.

Let's wait to hear what other maintainers think before making any change if we decide to go with it the help is appreciated :)

jkwatson:
This seems ok to me. Making it clear that we're not creating a new object with these calls is a good idea.

anuraaga:
Cool - @Rocksnake if you'd like to help that'd be great

Rocksnake:
I tried the following according to my understanding, and obtained them separately for different factories：
`SamplingResult.java`
<img width="819" alt="截屏2021-07-01 11 03 26" src="https://user-images.githubusercontent.com/44219737/124058621-932b7d00-d9de-11eb-9a67-f55d248e4032.png">
`ImmutableSamplingResult.java`
<img width="766" alt="截屏2021-07-01 11 03 59" src="https://user-images.githubusercontent.com/44219737/124058643-9fafd580-d9de-11eb-9283-ade0d3f7de0c.png">
and I wrote the unit test in SamplingResultTest without errors.
![image](https://user-images.githubusercontent.com/44219737/124059006-3f6d6380-d9df-11eb-861a-66e60d6e59cb.png)

Can you help me check if my understanding is correct? If it is correct, I will optimize the method of obtaining factories in the whole project.

anuraaga:
[...]

PR #3360, opened 2021-07-01
fix:Add factories for SamplingResult #3354

Rocksnake:
- Add three methods to get the factory to replace the original create

- Modify all the places where SamplingResult.create is used in the project

- Write unit test and pass the test

Remarks：since the create method with parameters is used in RateLimitingSampler, I keep the versions with Attributes


Rocksnake:
Could anybody help to approve running these workflows, great thanks!
This PR resolves #3354

Oberon00:
For the future: Please also mention the ticket number in the description as "Fixes #3354". GitHub does not link issues from the title.

> Could anybody help to approve running these workflows, great thanks!

I have an "Approve and run" button and clicked it and it seemed to have started but now it seems stuck. Probably an actual maintainer needs to approve the run.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/SamplingResult.java
```diff
@@ -23,48 +23,112 @@
...
+   *
+   * <p>Using {@link #recordAndSample()} instead of this method is slightly faster and
+   * shorter if you don't need attributes.
+   *
+   * @param attributes The attributes to return from {@link SamplingResult#getAttributes()}. A
+   *     different object instance with the same elements may be returned.
+   * @return A {@link SamplingResult} with the attributes equivalent to {@code attributes} and the
+   * {@link SamplingDecision#RECORD_AND_SAMPLE}
+   */
+  static SamplingResult recordAndSample(Attributes attributes) {
```
jkwatson, line 54: I wouldn't create these overloads with attributes. As @anuraaga mentioned, the existing method works fine for that, and actually does create a new object when being called.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/ImmutableSamplingResult.java
```diff
@@ -22,6 +22,18 @@
   static final SamplingResult EMPTY_RECORDED_SAMPLING_RESULT =
       ImmutableSamplingResult.createWithoutAttributes(SamplingDecision.RECORD_ONLY);
 
+  static SamplingResult getEmptyRecordedAndSampledSamplingResult() {
```
jkwatson, line 25: small thing, but I might just call this `recordAndSample()`. It's much easier to read and there's fewer words to parse through to get to what it's doing. I know this isn't public, but I think it would be more readable to have a shorter, less wordy method name.  Same for the other methods.

jkwatson: actually...do we even need these methods at all? The things that need them can just use the fields like they are today. I think we should stick with that.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/SamplingResult.java
```diff
@@ -23,48 +23,112 @@
...
+   * no attributes and {@link SamplingResult#getDecision()} returning {@code decision}.
+   *
+   * <p>Use {@link #drop(Attributes)} if you need attributes.
+   *
+   * <p>This is meant for use by custom {@link Sampler} implementations.
    *
-   * @param decision The decision made on the span.
    * @return A {@link SamplingResult} with empty attributes and the provided {@code decision}.
    */
-  static SamplingResult create(SamplingDecision decision) {
```
jkwatson, line 36: we can't delete this public method. That would be a breaking change.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/SamplingResult.java
```diff
@@ -23,48 +23,112 @@
...
    * shorter if you don't need attributes.
    *
-   * @param decision The decision made on the span.
    * @param attributes The attributes to return from {@link SamplingResult#getAttributes()}. A
    *     different object instance with the same elements may be returned.
    * @return A {@link SamplingResult} with the attributes equivalent to {@code attributes} and the
-   *     provided {@code decision}.
+   * {@link SamplingDecision#DROP}
    */
-  static SamplingResult create(SamplingDecision decision, Attributes attributes) {
```
jkwatson, line 63: we can't delete this method. that would be a breaking change.

jkwatson:
@Rocksnake please run the `./gradlew build` task and make sure to add any changes in the docs/apidiffs directory to this PR, since it is a change to the public APIs in the SDK. Thanks!

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/SamplingResult.java
Rocksnake: Thanks, but in the project, `RateLimitingSampler` calls the method with parameters. I think the method with parameters is necessary when we need specific sampling. I'm a first-time contributor, I hope you can give me some guidance.

anuraaga: @rocksnake Thanks for the help - as per the below comment, we won't be (can't) removing the `create` methods. We should continue to use `create` for creating the results with attributes

Rocksnake: Ok, thanks. And in my last commit, I have restored the original create method.

docs/apidiffs/current_vs_latest/opentelemetry-sdk-trace.txt
```diff
@@ -1,4 +1,12 @@
 Comparing source compatibility of  against 
+***! MODIFIED INTERFACE: PUBLIC ABSTRACT io.opentelemetry.sdk.trace.samplers.SamplingResult  (not serializable)
+	===  CLASS FILE FORMAT VERSION: 52.0 <- 52.0
+	+++! NEW METHOD: PUBLIC(+) STATIC(+) io.opentelemetry.sdk.trace.samplers.SamplingResult drop()
+	+++! NEW METHOD: PUBLIC(+) STATIC(+) io.opentelemetry.sdk.trace.samplers.SamplingResult drop(io.opentelemetry.api.common.Attributes)
```
jkwatson, line 5: we'd still like to get rid of these overloads that have attributes as a parameter. Thanks.

Rocksnake: image
I think this kind of call case is also necessary.

But I will try my best to think about how to replace these calls with attributes as parameter.

Rocksnake: Based on all the above, I feel that we should keep the old calls with attributes, and use the new short method in the future development process. I have achieved this idea and hope you approve.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/ImmutableSamplingResult.java
jkwatson: yes, let's get rid of these methods and just let the package access the package-protected constants like it has been.

Rocksnake:
I have resolve all the problems, could anyone help to approve running these workflows and merge in main branch, great thanks!

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/ImmutableSamplingResult.java
jkwatson: @Rocksnake please get rid of these methods. thanks!

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/SamplingResult.java
```diff
@@ -33,6 +33,7 @@
    * @param decision The decision made on the span.
    * @return A {@link SamplingResult} with empty attributes and the provided {@code decision}.
    */
+  @SuppressWarnings("UngroupedOverloads")
```
trask, line 36: I don't think you need this suppress warnings

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/SamplingResult.java
trask: is it too early to deprecate?

jkwatson: I think it's fine to keep it here. we could deprecate at some point, but it's not harmful or bad to use this method, just slightly confusing since it won't actually *create* anything.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/SamplingResult.java
```diff
@@ -60,13 +61,50 @@ static SamplingResult create(SamplingDecision decision) {
    * @return A {@link SamplingResult} with the attributes equivalent to {@code attributes} and the
    *     provided {@code decision}.
    */
+  @SuppressWarnings("UngroupedOverloads")
```
jkwatson, line 64: It would be better if we can make sure we group the overloads so we don't need this warning.

Rocksnake: I have removed them, could you please trigger the check for me again. Thanks~

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/SamplingResult.java
```diff
@@ -67,6 +67,42 @@ static SamplingResult create(SamplingDecision decision, Attributes attributes) {
         : ImmutableSamplingResult.createSamplingResult(decision, attributes);
   }
 
+  /**
+   * Returns a {@link SamplingResult} corresponding to {@link SamplingDecision#RECORD_AND_SAMPLE}
+   * with no attributes and {@link SamplingResult#getDecision()} returning {@code decision}.
+   *
+   * <p>This is meant for use by custom {@link Sampler} implementations.
+   *
+   * @return A {@link SamplingResult} with empty attributes and the provided {@code decision}.
```
anuraaga, line 76:
```suggestion
   * <p>This is meant for use by custom {@link Sampler} implementations and is equivalent to calling
   * {@code SamplingResult.create(SamplingDecision.RECORD_AND_SAMPLE)}.
   *
   * @return A {@link SamplingResult} with empty attributes and the provided {@code decision}.
```

Let's add something like this to the three methods.

anuraaga (commented):
Thanks! Just @jkwatson's coment for removing the unnecessary private methods and a small suggestion to the javadocs

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/SamplingResult.java
Rocksnake: I have added them, could you check for me and merge it please. Thanks.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/ImmutableSamplingResult.java
anuraaga: @Rocksnake I think you missed this comment

Rocksnake:
截屏2021-07-06 10 46 30

Does anyone know the reason for this problem, I need you help to solve it. The last check is no-error, I only modify three lines of annotation this time.

Rocksnake:
Could anyone merge this branch into main, I have solved all above problem, great thanks.

jkwatson:
> Could anyone merge this branch into main, I have solved all above problem, great thanks.

You still haven't removed the methods I asked you to. Please do so.

Rocksnake:
> > Could anyone merge this branch into main, I have solved all above problem, great thanks.
> 
> You still haven't removed the methods I asked you to. Please do so.

I haved removed these methods, please check again, thank you!

jkwatson:
This PR needs to be rebased against the main branch before we can merge it.

jkwatson:
@Rocksnake this needs a rebase. Will you be able to get to it soon?

Rocksnake:
> @Rocksnake this needs a rebase. Will you be able to get to it soon?

Ok, I will rebased it this day.

PR #3363, opened 2021-07-02
Allow users to access the default Clock and TestClock

### LINKED ISSUE #3359: Add Clock.getDefault returning SystemClock and move TestClock to sdk-testing as public
We can expect some external SDK extensions, perhaps a rate limiting sampler, to require a `Clock`. We define this interface but there doesn't seem to be a practical way for extensions outside this repo to use that interface because the implementations of it, `SystemClock` and `TestClock` are not exposed. While there does seem to be difficulty in "injecting" the Clock that is actually used in the SDK into an extension, at least a user should be able to do it, e.g.

```java
OpenTelemetrySdk.builder()
  .setSampler(CustomRateLimitingSampler.create()) // Implemented by delegating to constructor with the default Clock
  .build()
```

```java
OpenTelemetrySdk.builder()
  .setClock(CustomClock.getInstance())
  .setSampler(CustomRateLimitingSampler.create(CustomClock.getInstance())
  .build()
```

And it would also be our duty to help test this code by publishing TestClock in the testing package

anuraaga:
Fixes #3359 

Updates TestClock API to use time types now that it'll be public


docs/apidiffs/current_vs_latest/opentelemetry-sdk-testing.txt
```diff
@@ -1,2 +1,11 @@
 Comparing source compatibility of  against 
-No changes.
\ No newline at end of file
++++  NEW CLASS: PUBLIC(+) FINAL(+) io.opentelemetry.sdk.testing.time.TestClock  (not serializable)
+	+++  CLASS FILE FORMAT VERSION: 52.0 <- n.a.
+	+++  NEW SUPERCLASS: java.lang.Object
+	+++  NEW METHOD: PUBLIC(+) void advance(java.time.Duration)
```
anuraaga, line 5: API diffs reminded me to clean up this API :)

sdk/all/src/test/java/io/opentelemetry/sdk/internal/SystemClockTest.java
```diff
@@ -22,7 +23,7 @@ void millisPrecision() {
     // If we test many times, we can be fairly sure we didn't just get lucky with having a rounded
     // result on a higher than expected precision timestamp.
     for (int i = 0; i < 100; i++) {
-      long now = SystemClock.getInstance().now();
+      long now = Clock.getDefault().now();
```
jkwatson, line 26: it probably makes sense to keep calling `SystemClock.getInstance()` in the test of the SystemClock. :)

sdk/all/src/test/java/io/opentelemetry/sdk/internal/SystemClockTest.java
```diff
@@ -34,7 +35,7 @@ void microsPrecision() {
     // coincidentally rounded to millis precision.
     int numHasMicros = 0;
     for (int i = 0; i < 100; i++) {
-      long now = SystemClock.getInstance().now();
+      long now = Clock.getDefault().now();
```
jkwatson, line 38: ditto down here.

sdk/common/src/main/java/io/opentelemetry/sdk/common/Clock.java
```diff
@@ -10,6 +10,15 @@
 /** Interface for getting the current time. */
 @ThreadSafe
 public interface Clock {
+
+  /**
+   * Returns a default {@link Clock} which reads from {@linkplain System#currentTimeMillis() system
```
jkwatson, line 15: it also reads from `System.nanoTime()` as well, yes?

sdk/common/src/main/java/io/opentelemetry/sdk/common/Clock.java
```diff
@@ -10,6 +10,15 @@
...
+   * time}.
+   */
+  static Clock getDefault() {
```
jkwatson, line 18: any thoughts to exposing a `getDefaultMonotonicClock()` method as well?

jkwatson: In fact, I wonder if it might be better to *only* expose the monotonic clock, since the `now()` method on the SystemClock isn't appropriate for doing timings and will certainly be misused for that reason.

jkwatson: Thinking more on this, this is also probably not a good thing to expose (read the javadoc on the MonotonicClock), so I'm not sure what the best way to expose the safest things might be. Lots of docs, I guess?

anuraaga: Yeah the fact that `now()` is "duration safe in `MonotonicClock` is an implementation detail of `MonotonicClock` but not a feature of the `Clock` interface itself. I added javadoc to `now()` to clarify the pitfall, which would have applied before or after this PR.

jsuereth: +1 to exposing monotonic clock...

anuraaga: `MonotonicClock` is only used by the SDK (e.g., we have no user-facing builders that accept it as a parameter) so I don't think we need to add a factory method here for it. @jsuereth `MonotonicClock` is a public internal class so if you're asking for it for metrics, does it work for you already?

anuraaga: Ah but that's my #3372. If you already need it then I could stop moving to package-private, that's for the other PR I guess.

sdk/common/src/main/java/io/opentelemetry/sdk/common/Clock.java
```diff
@@ -10,19 +10,36 @@
...
+   * using {@link #nanoTime()}. For example, this usage is correct.
+   *
+   * <pre>{@code
+   * long startNanos = clock.nanoTime();
+   * // Spend time...
+   * long durationNanos = clock.nanoTime() - startNanos;
+   * }</pre>
    *
-   * @return the current epoch timestamp in nanos.
+   * <p>This usage is not correct.
```
jkwatson, line 30:
```suggestion
   * <p>This usage is NOT correct.
```

jkwatson (approved):
Thanks!

jkwatson:
I think this needs a rebase to pick up the PR check changes.

jkwatson:
This'll needed to be added to the changelog for 1.4.0

jsuereth:
FYI - I think there was a merge conflict with my Metric-Test Commit where some assertj migrations were reverted to "isEqualTo" or "containsExactly".   This actually breaks a lot of downstream metrics SDK rewrite code.

(a) Why was this not visible in the diff?
(b) Is there a way you could communicate those changes in the future, @anuraaga ?   It's not apparent in the current codebase, but the ordering of attributes breaks once the underlying aggregators move from Labels => Attributes which means when I rebased all my tests started failing again.

jsuereth:
Nevermind previous comment, was an unrelated conflict with something else....

PR #3365, opened 2021-07-02
Use property to specify test java version instead of adding tasks.

anuraaga:
As we add more java versions to our test set, adding tasks to the same build won't scale as well as going ahead and parallelizing on GHA. This makes us more consistent with instrumentation repo too.

BTW I found https://github.com/line/armeria/issues/3673

### DIFF
--- a/.github/workflows/pr-examples-build.yml
+++ b/.github/workflows/pr-examples-build.yml
@@ -15,10 +15,6 @@ jobs:
         os:
           - macos-latest
           - ubuntu-18.04
-        include:
-          - os: ubuntu-18.04
-            testAdditionalJavaVersions: true
-            coverage: true
     steps:
       - uses: actions/checkout@v2.3.4
       - id: setup-java-11
--- a/all/build.gradle.kts
+++ b/all/build.gradle.kts
@@ -14,8 +14,11 @@ tasks {
         options.release.set(11)
     }
 
-    named("testJava8") {
-        enabled = false
+    val testJavaVersion: String? by project
+    if (testJavaVersion == "8") {
+        test {
+            enabled = false
+        }
     }
 }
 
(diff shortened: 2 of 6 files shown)


.github/workflows/pr-examples-build.yml
```diff
@@ -15,10 +15,6 @@ jobs:
         os:
           - macos-latest
           - ubuntu-18.04
-        include:
```
anuraaga, line 18: We use neither flags in this build, they were just copy pasta artifacts I guess

Oberon00:
What about using a task name rule instead? https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:task_rules

That way, one could still (e.g. locally) execute tests for multiple Java versions in the same gradle invocation, like `./gradlew testJava8 testJava11 testJava16`, and they would be be parallelized.

anuraaga:
@Oberon00 While working on this I realized that we probably are not running extra tests added via test-sets on Java 8. Using a property to reconfigure `withType<Test>` is easy to reason about compared to adding tasks I guess, and I don't think it's common enough to want to run multiple JRE tests at the same time to try to dig into a probably complex solution to that.

trask (approved):
> This makes us more consistent with instrumentation repo too.

👍

all/build.gradle.kts
```diff
@@ -14,8 +14,11 @@ tasks {
         options.release.set(11)
     }
 
-    named("testJava8") {
-        enabled = false
+    val testJavaVersion: String? by project
+    if (testJavaVersion == "8") {
```
jkwatson, line 18: why is this block needed, exactly?

jkwatson (approved):
seems good to me. Go ahead and change the required checks and merge when ready.

all/build.gradle.kts
anuraaga: The comment block above is (possibly confusingly) explaining it, we don't need to run this on java 8 and it doesn't work because of jfr-events

PR #3366, opened 2021-07-02
Now `MultiTextMapPropagator` allows to select extractors and injectors separately

### LINKED ISSUE #3364: Composite propagators
Implementation of composite propagator, with ability to choose which should be used as injectors and which should be used as extractors per spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#create-a-composite-propagator

In my particular use case, I would love to support both w3c and b3 headers as "extractor", but I want to emit only w3c as trace headers.

slinkydeveloper:
Fix #3364

This PR includes:

* A patch for `MultiTextMapPropagator` to select extractors and injectors separately
* The builder `CompositeTextMapPropagatorBuilder`
* Unit test

Signed-off-by: Francesco Guardiani <francescoguard@gmail.com>

### DIFF
--- a/context/src/main/java/io/opentelemetry/context/propagation/CompositeTextMapPropagatorBuilder.java
+++ b/context/src/main/java/io/opentelemetry/context/propagation/CompositeTextMapPropagatorBuilder.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.context.propagation;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A builder for configuring an {@link TextMapPropagator} specifying which propagators should be used for extracting the context and which should be used for injecting the context.
+ */
+public final class CompositeTextMapPropagatorBuilder {
+
+  private final List<TextMapPropagator> extractors;
+  private final List<TextMapPropagator> injectors;
+
+  /**
+   * Package protected to disallow direct initialization.
+   *
+   * @see TextMapPropagator#builder()
+   */
+  CompositeTextMapPropagatorBuilder() {
+    this.extractors = new ArrayList<>();
+    this.injectors = new ArrayList<>();
+  }
+
+  /**
+   * Add a {@link TextMapPropagator} to be used only to extract the context.
+   */
+  public CompositeTextMapPropagatorBuilder extractor(TextMapPropagator propagator) {
+    this.extractors.add(propagator);
+    return this;
+  }
+
+  /**
+   * Add a {@link TextMapPropagator} to be used only to inject the context.
+   */
+  public CompositeTextMapPropagatorBuilder injector(TextMapPropagator propagator) {
+    this.injectors.add(propagator);
+    return this;
+  }
+
+  /**
+   * Add a {@link TextMapPropagator} to be used both to extract and inject the context.
+   */
+  public CompositeTextMapPropagatorBuilder propagator(TextMapPropagator propagator) {
+    this.injectors.add(propagator);
+    this.extractors.add(propagator);
+    return this;
+  }
+
+  /**
+   * Returns the built {@link TextMapPropagator}
+   *
+   * @see CompositeTextMapPropagatorBuilder
+   */
+  public TextMapPropagator build() {
+    if (this.injectors.isEmpty()) {
+      this.injectors.add(TextMapPropagator.noop());
+    }
+    if (this.extractors.isEmpty()) {
+      this.extractors.add(TextMapPropagator.noop());
+    }
+
+    return new MultiTextMapPropagator(this.extractors, this.injectors);
+  }
+}
--- a/context/src/main/java/io/opentelemetry/context/propagation/MultiTextMapPropagator.java
+++ b/context/src/main/java/io/opentelemetry/context/propagation/MultiTextMapPropagator.java
@@ -16,7 +16,8 @@ import java.util.Set;
 import javax.annotation.Nullable;
 
 final class MultiTextMapPropagator implements TextMapPropagator {
-  private final TextMapPropagator[] textPropagators;
+  private final TextMapPropagator[] extractors;
+  private final TextMapPropagator[] injectors;
   private final Collection<String> allFields;
 
   MultiTextMapPropagator(TextMapPropagator... textPropagators) {
@@ -24,9 +25,20 @@ final class MultiTextMapPropagator implements TextMapPropagator {
   }
 
   MultiTextMapPropagator(List<TextMapPropagator> textPropagators) {
-    this.textPropagators = new TextMapPropagator[textPropagators.size()];
-    textPropagators.toArray(this.textPropagators);
-    this.allFields = Collections.unmodifiableList(getAllFields(this.textPropagators));
+    this.extractors = new TextMapPropagator[textPropagators.size()];
+    textPropagators.toArray(this.extractors);
+    this.injectors = this.extractors; // No need to copy
+    this.allFields = Collections.unmodifiableList(getAllFields(this.injectors));
+  }
+
+  MultiTextMapPropagator(List<TextMapPropagator> extractors, List<TextMapPropagator> injectors) {
+    this.extractors = new TextMapPropagator[extractors.size()];
+    extractors.toArray(this.extractors);
+
+    this.injectors = new TextMapPropagator[injectors.size()];
+    injectors.toArray(this.injectors);
+
+    this.allFields = Collections.unmodifiableList(getAllFields(this.injectors));
   }
 
   @Override
@@ -48,7 +60,7 @@ final class MultiTextMapPropagator implements TextMapPropagator {
     if (context == null || setter == null) {
       return;
     }
-    for (TextMapPropagator textPropagator : textPropagators) {
+    for (TextMapPropagator textPropagator : injectors) {
       textPropagator.inject(context, carrier, setter);
     }
   }
@@ -61,7 +73,7 @@ final class MultiTextMapPropagator implements TextMapPropagator {
     if (getter == null) {
       return context;
     }
-    for (TextMapPropagator textPropagator : textPropagators) {
+    for (TextMapPropagator textPropagator : extractors) {
       context = textPropagator.extract(context, carrier, getter);
     }
     return context;
--- a/context/src/main/java/io/opentelemetry/context/propagation/TextMapPropagator.java
+++ b/context/src/main/java/io/opentelemetry/context/propagation/TextMapPropagator.java
@@ -44,6 +44,13 @@ import javax.annotation.concurrent.ThreadSafe;
 public interface TextMapPropagator {
 
   /**
+   * Returns a composite text map propagator builder that allows to specify specific propagators as extractors and specific propagators as injectors.
+   */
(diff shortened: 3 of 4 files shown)


jkwatson:
@slinkydeveloper this proposed change includes additions to stable public APIs. Can you run `./gradlew build` and include the changes in doc/apidiffs in this PR so we can examine and document what public API changes you are proposing? Thanks!

context/src/main/java/io/opentelemetry/context/propagation/CompositeTextMapPropagatorBuilder.java
```diff
@@ -0,0 +1,69 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.context.propagation;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A builder for configuring an {@link TextMapPropagator} specifying which propagators should be used for extracting the context and which should be used for injecting the context.
```
anuraaga, line 12:
```suggestion
 * A builder for configuring a {@link TextMapPropagator} specifying which propagators should be used for extracting the context and which should be used for injecting the context.
```

context/src/main/java/io/opentelemetry/context/propagation/CompositeTextMapPropagatorBuilder.java
```diff
@@ -0,0 +1,69 @@
...
+   */
+  CompositeTextMapPropagatorBuilder() {
+    this.extractors = new ArrayList<>();
+    this.injectors = new ArrayList<>();
+  }
+
+  /**
+   * Add a {@link TextMapPropagator} to be used only to extract the context.
+   */
+  public CompositeTextMapPropagatorBuilder extractor(TextMapPropagator propagator) {
```
anuraaga, line 33: For better or worse, we generally use prefixes like `get` `set` in this project. So let's use `addExtractor` `addInjector` for consistency

context/src/main/java/io/opentelemetry/context/propagation/CompositeTextMapPropagatorBuilder.java
```diff
@@ -0,0 +1,69 @@
...
+   *
+   * @see TextMapPropagator#builder()
+   */
+  CompositeTextMapPropagatorBuilder() {
+    this.extractors = new ArrayList<>();
+    this.injectors = new ArrayList<>();
+  }
+
+  /**
+   * Add a {@link TextMapPropagator} to be used only to extract the context.
```
anuraaga, line 30:
```suggestion
   * Adds a {@link TextMapPropagator} to be used only when extracting context.
```

context/src/main/java/io/opentelemetry/context/propagation/CompositeTextMapPropagatorBuilder.java
```diff
@@ -0,0 +1,69 @@
...
+   */
+  public CompositeTextMapPropagatorBuilder extractor(TextMapPropagator propagator) {
+    this.extractors.add(propagator);
+    return this;
+  }
+
+  /**
+   * Add a {@link TextMapPropagator} to be used only to inject the context.
```
anuraaga, line 38:
```suggestion
   * Adds a {@link TextMapPropagator} to be used only when injecting context.
```

context/src/main/java/io/opentelemetry/context/propagation/CompositeTextMapPropagatorBuilder.java
```diff
@@ -0,0 +1,69 @@
...
+   */
+  public CompositeTextMapPropagatorBuilder injector(TextMapPropagator propagator) {
+    this.injectors.add(propagator);
+    return this;
+  }
+
+  /**
+   * Add a {@link TextMapPropagator} to be used both to extract and inject the context.
```
anuraaga, line 46:
```suggestion
   * Adds a {@link TextMapPropagator} to be used both when extracting and injecting context.
```

context/src/main/java/io/opentelemetry/context/propagation/CompositeTextMapPropagatorBuilder.java
```diff
@@ -0,0 +1,69 @@
...
+  CompositeTextMapPropagatorBuilder() {
+    this.extractors = new ArrayList<>();
+    this.injectors = new ArrayList<>();
+  }
+
+  /**
+   * Add a {@link TextMapPropagator} to be used only to extract the context.
+   */
+  public CompositeTextMapPropagatorBuilder extractor(TextMapPropagator propagator) {
+    this.extractors.add(propagator);
```
anuraaga, line 33: Add `requireNonNull(propagator, "propagator")` to all the methods and unit tests verifying the null failure

context/src/main/java/io/opentelemetry/context/propagation/CompositeTextMapPropagatorBuilder.java
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  /**
+   * Returns the built {@link TextMapPropagator}
+   *
+   * @see CompositeTextMapPropagatorBuilder
+   */
+  public TextMapPropagator build() {
+    if (this.injectors.isEmpty()) {
+      this.injectors.add(TextMapPropagator.noop());
```
anuraaga, line 61: We don't need these blocks right?

context/src/main/java/io/opentelemetry/context/propagation/MultiTextMapPropagator.java
```diff
@@ -16,17 +16,29 @@
 import javax.annotation.Nullable;
 
 final class MultiTextMapPropagator implements TextMapPropagator {
```
anuraaga, line 18: I think it's appropriate to rename this to `CompositeTextMapPropagator` in this PR

context/src/main/java/io/opentelemetry/context/propagation/CompositeTextMapPropagatorBuilder.java
slinkydeveloper: You mean we don't need noop injectors and extractor? I see that `TextMapPropagator#composite` adds them

anuraaga: `TextMapPropagator#composite` replaces the `MultiTextMapPropagator` completely with no-op if there aren't any propagators. So yeah I guess we can do similar here, it is

```java
if (injectors.isEmpty() && extractors.isEmpty()) {
  return NoopTextMapPropagator.getInstance();
}
return new MultiTextMapPropagator(injectors, extractors);
```

slinkydeveloper: Makes sense!

slinkydeveloper:
Pushed the doc changes as well

anuraaga (approved):
Thanks!

context/src/main/java/io/opentelemetry/context/propagation/CompositeTextMapPropagator.java
```diff
@@ -15,18 +15,31 @@
...
+
+  CompositeTextMapPropagator(
+      List<TextMapPropagator> extractors, List<TextMapPropagator> injectors) {
+    this.extractors = new TextMapPropagator[extractors.size()];
+    extractors.toArray(this.extractors);
+
+    this.injectors = new TextMapPropagator[injectors.size()];
+    injectors.toArray(this.injectors);
+
+    this.allFields = Collections.unmodifiableList(getAllFields(this.injectors));
```
jkwatson, line 42: is this right? The fields are only from the injectors? I think this might need to be the union of the extractors and injectors, rather than just the injectors.

slinkydeveloper: That's not what I understood from the `fields()` javadoc, it rather seems to me that method is used only when writing the context back to the wire, and not while extracting it...

jkwatson: I think the javadoc provides that as a usage (to clear out headers if the header implementation is re-used), but I think `fields()` could also be used by extraction in some cases. I'm honestly not 100% sure what the behavior should be in this case when injection doesn't necessarily match extraction.

jkwatson: I created a spec issue for this, since I don't think I know what the answer should be:  https://github.com/open-telemetry/opentelemetry-specification/issues/1809

kubawach: I second @jkwatson with this - I know at least one case in instrumentation where `fields` are considered during extraction (see `io.opentelemetry.instrumentation.awslambda.v1_0.TracingRequestStreamHandler`)

slinkydeveloper: Let's take a concrete case I have in mind: I have 2 extractors, 1 injector, assuming fields returns the union of the header names used by extractors and injectors, if you use `fields` to preallocate the headers when writing back to the wire, you're gonna waste allocations because you're gonna allocate for those extractors as well. 

Perhaps you need 2 methods for getting the fields, one for injectors and one for extractors?

kubawach (approved):
Definition of `fields` with separate injectors / extractors needs clarification before merging.

context/src/main/java/io/opentelemetry/context/propagation/CompositeTextMapPropagator.java
jkwatson: Please comment on the spec issue. :)

kubawach: Weighting issues - sub-optimal performance vs actual bug I'd say that keeping all fields still wins. But yeah, spec issue is the proper place to discuss this.

yurishkuro: @kubawach please provide a permalink to your example of TracingRequestStreamHandler, I couldn't find it.

kubawach: @yurishkuro here you go: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/ApiGatewayProxyRequest.java

yurishkuro: @kubawach thanks. If I am reading this correctly, I wouldn't classify this usage as "on the extract path", and I probably would've implemented that check differently in the first place, the use of fields() to determine "noHttpPropagationNeeded" seems a bit round-about. Here's the exact usage link, for reference:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/1b5df6d78a6877dab38f573b44d0358927d1ab3d/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/ApiGatewayProxyRequest.java#L27

PR #3368, opened 2021-07-02
Make Jaeger remote sampler available via SPI

pavolloffay:
Signed-off-by: Pavol Loffay <p.loffay@gmail.com>

This PR makes `JaegerRemoteSampler` available via SPI. The ultimate goal is to make it available in the java auto-instrumentation (javaagent).  I have talked about this with @tedsuo and @yurishkuro that including it in the javaagent distro should be fine (Jaeger is considered as a special vendor). 

After this PR is merged I will create a PR to auto-instrumentation to package it in the final distribution. We should also document it in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#general-sdk-configuration

### DIFF
--- a/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProvider.java
+++ b/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProvider.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.extension.trace.jaeger.sampler;
+
+import io.opentelemetry.sdk.autoconfigure.ConfigProperties;
+import io.opentelemetry.sdk.autoconfigure.spi.ConfigurableSamplerProvider;
+import io.opentelemetry.sdk.trace.samplers.Sampler;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+public class JaegerRemoteSamplerProvider implements ConfigurableSamplerProvider {
+
+  private static final String ENDPOINT_KEY = "endpoint";
+  private static final String POLLING_INTERVAL = "pollingInterval";
+  private static final String INITIAL_SAMPLING_RATE = "initialSamplingRate";
+
+  @Override
+  public Sampler createSampler(ConfigProperties config) {
+    String serviceName = config.getString("otel.service.name");
+    if (serviceName == null) {
+      Map<String, String> resourceAttributes =
+          config.getCommaSeparatedMap("otel.resource.attributes");
+      serviceName = resourceAttributes.get("otel.service.name");
+    }
+
+    JaegerRemoteSamplerBuilder builder = JaegerRemoteSampler.builder().setServiceName(serviceName);
+    Map<String, String> params = config.getCommaSeparatedMap("otel.traces.sampler.arg");
+
+    // Optional configuration
+    String endpoint = params.get(ENDPOINT_KEY);
+    if (endpoint != null) {
+      builder.setEndpoint(endpoint);
+    }
+    String pollingInterval = params.get(POLLING_INTERVAL);
+    if (pollingInterval != null) {
+      builder.setPollingInterval(Integer.valueOf(pollingInterval), TimeUnit.MILLISECONDS);
+    }
+    String initialSamplingRate = params.get(INITIAL_SAMPLING_RATE);
+    if (initialSamplingRate != null) {
+      builder.setInitialSampler(Sampler.traceIdRatioBased(Double.valueOf(initialSamplingRate)));
+    }
+    return builder.build();
+  }
+
+  @Override
+  public String getName() {
+    return "jaeger_remote";
+  }
+}
--- a/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProvider.java
+++ b/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProvider.java
@@ -13,21 +13,24 @@ import java.util.concurrent.TimeUnit;
 
 public class JaegerRemoteSamplerProvider implements ConfigurableSamplerProvider {
 
+  private static final String ATTRIBUTE_PROPERTY = "otel.resource.attributes";
+  private static final String SERVICE_NAME_PROPERTY = "otel.service.name";
+  private static final String SAMPLER_ARG_PROPERTY = "otel.traces.sampler.arg";
+
   private static final String ENDPOINT_KEY = "endpoint";
   private static final String POLLING_INTERVAL = "pollingInterval";
   private static final String INITIAL_SAMPLING_RATE = "initialSamplingRate";
 
   @Override
   public Sampler createSampler(ConfigProperties config) {
-    String serviceName = config.getString("otel.service.name");
+    String serviceName = config.getString(SERVICE_NAME_PROPERTY);
     if (serviceName == null) {
-      Map<String, String> resourceAttributes =
-          config.getCommaSeparatedMap("otel.resource.attributes");
-      serviceName = resourceAttributes.get("otel.service.name");
+      Map<String, String> resourceAttributes = config.getCommaSeparatedMap(ATTRIBUTE_PROPERTY);
+      serviceName = resourceAttributes.get(SERVICE_NAME_PROPERTY);
     }
 
     JaegerRemoteSamplerBuilder builder = JaegerRemoteSampler.builder().setServiceName(serviceName);
-    Map<String, String> params = config.getCommaSeparatedMap("otel.traces.sampler.arg");
+    Map<String, String> params = config.getCommaSeparatedMap(SAMPLER_ARG_PROPERTY);
 
     // Optional configuration
     String endpoint = params.get(ENDPOINT_KEY);
--- a/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProvider.java
+++ b/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProvider.java
@@ -13,9 +13,10 @@ import java.util.concurrent.TimeUnit;
 
 public class JaegerRemoteSamplerProvider implements ConfigurableSamplerProvider {
 
-  private static final String ATTRIBUTE_PROPERTY = "otel.resource.attributes";
-  private static final String SERVICE_NAME_PROPERTY = "otel.service.name";
-  private static final String SAMPLER_ARG_PROPERTY = "otel.traces.sampler.arg";
+  // visible for testing
+  static final String ATTRIBUTE_PROPERTY = "otel.resource.attributes";
+  static final String SERVICE_NAME_PROPERTY = "otel.service.name";
+  static final String SAMPLER_ARG_PROPERTY = "otel.traces.sampler.arg";
 
   private static final String ENDPOINT_KEY = "endpoint";
   private static final String POLLING_INTERVAL = "pollingInterval";
@@ -43,7 +44,8 @@ public class JaegerRemoteSamplerProvider implements ConfigurableSamplerProvider 
     }
     String initialSamplingRate = params.get(INITIAL_SAMPLING_RATE);
     if (initialSamplingRate != null) {
-      builder.setInitialSampler(Sampler.traceIdRatioBased(Double.valueOf(initialSamplingRate)));
+      builder.setInitialSampler(
+          Sampler.parentBased(Sampler.traceIdRatioBased(Double.valueOf(initialSamplingRate))));
     }
     return builder.build();
   }
--- a/sdk-extensions/jaeger-remote-sampler/src/test/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProviderTest.java
+++ b/sdk-extensions/jaeger-remote-sampler/src/test/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProviderTest.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.extension.trace.jaeger.sampler;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.InstanceOfAssertFactories.type;
+
+import io.opentelemetry.sdk.autoconfigure.spi.ConfigurableSamplerProvider;
+import java.util.ServiceLoader;
+import org.junit.jupiter.api.Test;
+
+public class JaegerRemoteSamplerProviderTest {
+
+  @Test
+  void serviceProvider() {
+    ServiceLoader<ConfigurableSamplerProvider> samplerProviders =
+        ServiceLoader.load(ConfigurableSamplerProvider.class);
+    assertThat(samplerProviders).hasAtLeastOneElementOfType(JaegerRemoteSamplerProvider.class);
+    assertThat(samplerProviders)
+        .singleElement(type(JaegerRemoteSamplerProvider.class))
+        .satisfies(provider -> assertThat(provider.getName()).isEqualTo("jaeger_remote"));
+  }
+}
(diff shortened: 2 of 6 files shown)


jkwatson:
@pavolloffay please run `./gradle build` and include changes in the docs/apidiffs in this PR, since this adds a public class to a stable module. Thanks!

jkwatson:
@pavolloffay I'm pretty hesistant to add environment variable support for things that haven't been spec'd yet. I'm ok with adding this SPI, but without an official environment variable option, I think we shouldn't add it to the autoconfigure module yet.

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProvider.java
```diff
@@ -0,0 +1,41 @@
...
+import io.opentelemetry.sdk.autoconfigure.ConfigProperties;
+import io.opentelemetry.sdk.autoconfigure.spi.ConfigurableSamplerProvider;
+import io.opentelemetry.sdk.trace.samplers.Sampler;
+import java.util.concurrent.TimeUnit;
+
+public class JaegerRemoteSamplerProvider implements ConfigurableSamplerProvider {
+
+  @Override
+  public Sampler createSampler(ConfigProperties config) {
+    String serviceName = config.getString("jaeger.service");
```
yurishkuro, line 17: curious why this is "jaeger service" and not the default resource.service.name

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProvider.java
```diff
@@ -0,0 +1,41 @@
...
+    if (poolingInterval != null) {
+      builder.setPollingInterval(poolingInterval, TimeUnit.MILLISECONDS);
+    }
+
+    return builder.build();
+  }
+
+  @Override
+  public String getName() {
+    return "jaeger_remote_sampler";
```
anuraaga, line 39: How about moving this module to java-contrib? We're starting to use that repo more. We wouldn't need to be as strict about the spec there. Otherwise it does seem like we would want to add this Sampler to the spec.

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProvider.java
pavolloffay: I was thinking about this, I will rename it.

pavolloffay:
> @pavolloffay I'm pretty hesistant to add environment variable support for things that haven't been spec'd yet. I'm ok with adding this SPI, but without an official environment variable option, I think we shouldn't add it to the autoconfigure module yet.

Does the "without an official" mean that we can add the environment variables that are here with a note that are subject to a change or should I remove them? Note that the SPI without configuration does not make much sense. However, until the spec PR is merged I can use `OTEL_EXPORTER_JAEGER_ENDPOINT` to set the endpoint if we strictly want to use documented variables.

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProvider.java
pavolloffay: The goal is to include it in the official Java agent distribution. I would say modules from `java-contrib` do not make it that far. 

Note that Jaeger remote sampler has been here in the main java repository for a long time, but it's not included in the auto-instrumentation.

settings.gradle.kts
```diff
@@ -70,6 +70,7 @@ include(":sdk-extensions:tracing-incubator")
 include(":sdk-extensions:jaeger-remote-sampler")
 include(":sdk-extensions:jfr-events")
 include(":sdk-extensions:zpages")
+include(":sdk-extensions:jaeger-remote-sampler")
```
anuraaga, line 73: It's already included on line 70

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProvider.java
anuraaga: Ah I didn't know that's your plan - in that case, it's even more important to go through spec. The actual sampler implementation doesn't need to be added, just add the variables

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#general-sdk-configuration

I think it could be similar to the propagators, `"jaeger": Link to Jaeger remote sampler doc` and a section with its endpoint-specific variables.

Since this artifact is already a stable one, we shouldn't add something like this name that we may need to change after the spec so can't merge this until the spec is updated

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProvider.java
```diff
@@ -0,0 +1,41 @@
...
+  public Sampler createSampler(ConfigProperties config) {
+    String serviceName = config.getString("jaeger.service");
+    JaegerRemoteSamplerBuilder builder = JaegerRemoteSampler.builder().setServiceName(serviceName);
+
+    // Optional configuration
+    String endpoint = config.getString("jaeger.remote.sampler.endpoint");
+    if (endpoint != null) {
+      builder.setEndpoint(endpoint);
+    }
+    Double initialRate = config.getDouble("jaeger.remote.sampler.initial.rate");
```
anuraaga, line 25: I imagine we want this to be `initial_sampler` which can be set to any other possible value for `OTEL_SAMPLER`. Another good point we should go through the spec for

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProvider.java
pavolloffay: > Ah I didn't know that's your plan - in that case, it's even more important to go through spec. The actual sampler implementation doesn't need to be added, just add the variables

yeah it is in the PR description :)

pavolloffay:
I have updated the PR:
* added `opentelemetry-sdk-extension-jaeger-remote-sampler.txt`
* used only existing SDK configuration variables

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProvider.java
anuraaga: Yeah missed it - but while it's nice we could reuse many variables, still we've defined this new env variable value here. We need it in the spec - since you've talked to @tedsuo and @yurishkuro already presumably the spec PR can be merged relatively quickly since there already seems to be consensus for it though we'll all want to come up with a name (I wouldn't repeat the word `sampler` in this value)

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProvider.java
pavolloffay: removing this for now

pavolloffay:
Spec PR https://github.com/open-telemetry/opentelemetry-specification/pull/1791

jkwatson:
I think this will end up being fine, but let's hold off a bit until the spec gets approved, and not try to rush it into the 1.4.0 release this week.

anuraaga (approved):

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProvider.java
```diff
@@ -0,0 +1,64 @@
...
+  // visible for testing
+  // example value endpoint=localhost:14250,pollingIntervalMs=5000,initialSamplingRate=0.25
+  static Map<String, String> parseArg(String arg) {
+    Map<String, String> paramMap = new HashMap<>();
+    String[] params = arg.split(",");
+    for (int i = 0; i < params.length; i++) {
+      String param = params[i];
+      String[] keyVal = param.split("=");
+      if (keyVal.length != 2) {
+        throw new IllegalArgumentException(
```
pavolloffay, line 57: can `createSampler` throw an exception? The javadoc does not prohibit it.

pavolloffay: I was thinking about using `ConfigurationException` but that is not public.

pavolloffay:
CI is failing on 

```
Run actions/setup-java@v2
Trying to resolve the latest version from remote
Error: Could not find satisfied version for SemVer '8'. 
```

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProvider.java
jkwatson: didn't this end up being "jaeger_remote" in the spec PR that got merged?

sdk-extensions/jaeger-remote-sampler/src/test/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProviderTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+  @Test
+  void serviceProvider() {
+    ServiceLoader<ConfigurableSamplerProvider> samplerProviders =
+        ServiceLoader.load(ConfigurableSamplerProvider.class);
+    Iterator<ConfigurableSamplerProvider> iterator = samplerProviders.iterator();
+    iterator.forEachRemaining(
+        configurableSamplerProvider ->
+            assertThat(configurableSamplerProvider)
+                .isInstanceOf(JaegerRemoteSamplerProvider.class));
```
anuraaga, line 27:
```suggestion
    assertThat(samplerProviders).hasAtLeastOneElementOfType(JaegerRemoteSamplerProvider.class);
```

Found this while writing a similar test :)

anuraaga: Or you can check the name too with this pattern

https://github.com/open-telemetry/opentelemetry-java-contrib/pull/64/files#diff-e9cb51992b77bc5666278183358a03a9c3e5b0ec6a500277e9e5681086d3f8b4R20

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProvider.java
jkwatson: exceptions thrown at this point will bubble up to the caller of auto-configuration's `initialize()` method, which I think is fine for bad configuration, since there's no way to start up an SDK in that case.

jkwatson:
one small question, then looks good to me.

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProvider.java
```diff
@@ -0,0 +1,64 @@
...
+
+  private static final String ENDPOINT_KEY = "endpoint";
+  private static final String POLLING_INTERVAL = "pollingInterval";
+  private static final String INITIAL_SAMPLING_RATE = "initialSamplingRate";
+
+  @Override
+  public Sampler createSampler(ConfigProperties config) {
+    String serviceName = config.getString("otel.service.name");
+    JaegerRemoteSamplerBuilder builder = JaegerRemoteSampler.builder().setServiceName(serviceName);
+    Map<String, String> params = parseArg(config.getString("otel.traces.sampler.arg"));
```
anuraaga, line 25: Just noticed - can this be `config.getCommaSeparatedMap`?

Like https://github.com/open-telemetry/opentelemetry-java-contrib/pull/67/files

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProvider.java
```diff
@@ -0,0 +1,64 @@
...
+
+public class JaegerRemoteSamplerProvider implements ConfigurableSamplerProvider {
+
+  private static final String ENDPOINT_KEY = "endpoint";
+  private static final String POLLING_INTERVAL = "pollingInterval";
+  private static final String INITIAL_SAMPLING_RATE = "initialSamplingRate";
+
+  @Override
+  public Sampler createSampler(ConfigProperties config) {
+    String serviceName = config.getString("otel.service.name");
```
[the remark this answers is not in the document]
pavolloffay: good point

pavolloffay:
PR updated

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProvider.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+    String pollingInterval = params.get(POLLING_INTERVAL);
+    if (pollingInterval != null) {
+      builder.setPollingInterval(Integer.valueOf(pollingInterval), TimeUnit.MILLISECONDS);
+    }
+    String initialSamplingRate = params.get(INITIAL_SAMPLING_RATE);
+    if (initialSamplingRate != null) {
+      builder.setInitialSampler(Sampler.traceIdRatioBased(Double.valueOf(initialSamplingRate)));
+    }
+    return builder.build();
```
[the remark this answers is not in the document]
pavolloffay: yay I wanted to avoid getting to private fields in the test.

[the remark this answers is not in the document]
jkwatson: do you have another idea for how to test this code?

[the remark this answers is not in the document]
pavolloffay: nope, I have added some tests using assertj. It looks rather exotic but it works :)

pavolloffay:
I had to use `mock-maker-inline` due to final classes but then an existing test fails: 

```
> Task :sdk-extensions:jaeger-remote-sampler:test

JaegerRemoteSamplerTest > connectionWorks() FAILED
    Wanted but not invoked:
    samplingManagerImplBase.getSamplingStrategy(
        <Capturing argument>,
        <any>
    );
    -> at io.opentelemetry.sdk.extension.trace.jaeger.proto.api_v2.SamplingManagerGrpc$SamplingManagerImplBase.getSamplingStrategy(SamplingManagerGrpc.java:100)

    However, there was exactly 1 interaction with this mock:
    samplingManagerImplBase.bindService();
    -> at io.grpc.internal.ServerImplBuilder.addService(ServerImplBuilder.java:131)
        at io.opentelemetry.sdk.extension.trace.jaeger.proto.api_v2.SamplingManagerGrpc$SamplingManagerImplBase.getSamplingStrategy(SamplingManagerGrpc.java:100)
        at io.opentelemetry.sdk.extension.trace.jaeger.sampler.JaegerRemoteSamplerTest.connectionWorks(JaegerRemoteSamplerTest.java:114)


```

yurishkuro:
🎉 🎉 🎉 

@pavolloffay once this is officially released, are you aware of any other functional gaps vs. Jaeger Java SDK? It would be nice if we could put a banner on top of Jaeger SDK repo asking people to use OTEL SDK instead.

Issue #3311, opened 2021-06-12
Flaky test in CI


jkwatson:
This seems to be showing up quite a bit now:

```
> Task :context:strictContextEnabledTest

StrictContextEnabledTest > garbageCollectedScope() FAILED
    org.awaitility.core.ConditionTimeoutException: Assertion condition defined as a lambda expression in io.opentelemetry.context.StrictContextEnabledTest Contain the string <Scope garbage collected before being closed.> ==> None of the 0 captured log events matched the filter predicate within 30 seconds.
        at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:166)
        at org.awaitility.core.AssertionCondition.await(AssertionCondition.java:119)
        at org.awaitility.core.AssertionCondition.await(AssertionCondition.java:31)
        at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:939)
        at org.awaitility.core.ConditionFactory.untilAsserted(ConditionFactory.java:723)
        at io.opentelemetry.context.StrictContextEnabledTest.garbageCollectedScope(StrictContextEnabledTest.java:32)

        Caused by:
        org.opentest4j.AssertionFailedError: Contain the string <Scope garbage collected before being closed.> ==> None of the 0 captured log events matched the filter predicate
            at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:39)
            at org.junit.jupiter.api.Assertions.fail(Assertions.java:117)
            at io.github.netmikey.logunit.api.LogCapturer.lambda$assertContains$3(LogCapturer.java:174)
            at java.base/java.util.Optional.orElseGet(Optional.java:369)
            at io.github.netmikey.logunit.api.LogCapturer.assertContains(LogCapturer.java:173)
            at io.github.netmikey.logunit.api.LogCapturer.assertContains(LogCapturer.java:121)
            at io.opentelemetry.context.StrictContextEnabledTest.lambda$garbageCollectedScope$0(StrictContextEnabledTest.java:36)
```

Oberon00:
Also here in my PR: https://github.com/open-telemetry/opentelemetry-java/actions/runs/935680619

Rocksnake:
> Also here in my PR: https://github.com/open-telemetry/opentelemetry-java/actions/runs/935680619

In the action of this PR, it has run without errors. Has this issue been solved? If not, could you provide the recurring scene. Thanks!

jkwatson:
[...]

PR #3369, opened 2021-07-04
Fix:fix the error in flaky test

Rocksnake:
Link: #3311 

When I change the `log` source to `ThreadLocalContextStorage` and test in `StrictContextEnabledTest.java`，the error in the issue will be reproduced.

1. I think the error caused by scope
2. Scope should be closed when test is over
3. So I add the Scope.close when the when the current context should be restored to what it was before attaching toAttach.


jkwatson:
I don't think this is a solution to the flaky test, and it introduces logging where there shouldn't be any logging. 

The issue with the flaky test is in the testing methodology, not the production code. 

😕

anuraaga:
Thanks for looking into this @Rocksnake - but we haven't seen the flaky error in a long time so for now think it may not reproduce enough. I'm going to go ahead and close the issue and this PR for now, but if we see it again we can reopen and take a look then. Appreciate the help!

Issue #3339, opened 2021-06-22
Alternative queue implementation that does not use the `jdk.unsupported` module


brunojcm:
**Is your feature request related to a problem? Please describe.**
The current version of `opentelemetry-java` (1.3.0 as of this bug report) requires the `jdk.unsupported` JDK module to be included if one is using `jlink` to build a custom JDK. This is not a sustainable approach.

**Describe the solution you'd like**
Provide a config that allows a `BatchSpanProcessor` to be created using another `Queue<ReadableSpan>` implementation that does not rely on unsupported modules.

**Describe alternatives you've considered**
I'm currenlty adding `jdk.unsupported`, but I need to have a plan to remove those eventually. `opentelemetry-java` is my biggest dependency that requires `jdk.unsupported`.

anuraaga:
Hi @brunojcm - can you confirm that with jdk.unsupported not included, you get an exception from here?

https://github.com/JCTools/JCTools/blob/master/jctools-core/src/main/java/org/jctools/util/UnsafeAccess.java#L71

We could probably catch exceptions and fallback to `ArrayBlockingQueue` here instead of adding extra knobs if so

https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace-shaded-deps/src/main/java/io/opentelemetry/sdk/trace/internal/JcTools.java#L18

brunomed:
(brunojcm here, using my other account to reply)

That is the full stacktrace:
```
[otel.javaagent 2021-06-22 05:21:27:988 +0000] [main] INFO io.opentelemetry.javaagent.tooling.VersionLogger - opentelemetry-javaagent - version: 1.3.0
[otel.javaagent 2021-06-22 05:21:28:177 +0000] [main] INFO io.grpc.netty.shaded.io.netty.util.internal.PlatformDependent - Your platform does not provide complete low-level API for accessing direct buffers reliably. Unless explicitly requested, heap buffer will always be preferred to avoid potential system instability.
ERROR io.opentelemetry.javaagent.OpenTelemetryAgent
java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
[...]

PR #3371, opened 2021-07-05
Different pattern for embedding shaded dependency.

anuraaga:
Seeing https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3471 I thought of making another attempt to find a cleaner approach to embedding a relocated archive. This pattern is different - it's probably a little better

- Does not "side-step" Gradle by copying into the output Jar. So the classes are available in areas Gradle knows they should be (we see this with the removal of `jmh(":sdk:trace-shaded-deps")`). It doesn't remove the need of `compileOnly` though since it only adds classes to the output, not Java to the input (which don't exist in shaded form) which is needed for compilation
- No evaluation order dependency, always a benefit to reduce build fragility

Because shadow outputs straight to a jar though we still don't seem to have a way to avoid having the extractShadowJar, which is very similar to the reconfiguration of Jar we used to have.

@trask What do you think?

### DIFF
--- a/sdk/trace/build.gradle.kts
+++ b/sdk/trace/build.gradle.kts
@@ -1,3 +1,5 @@
+import ru.vyarus.gradle.plugin.animalsniffer.AnimalSniffer
+
 plugins {
     id("otel.java-conventions")
     id("otel.publish-conventions")
@@ -9,7 +11,13 @@ plugins {
 description = "OpenTelemetry SDK For Tracing"
 otelJava.moduleName.set("io.opentelemetry.sdk.trace")
 
-evaluationDependsOn(":sdk:trace-shaded-deps")
+
+sourceSets {
+    main {
+        val traceShadedDeps = project(":sdk:trace-shaded-deps")
+        output.dir(traceShadedDeps.file("build/extracted/shadow"), "builtBy" to ":sdk:trace-shaded-deps:extractShadowJar")
+    }
+}
 
 dependencies {
     api(project(":api:all"))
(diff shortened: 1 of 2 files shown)


sdk/trace/build.gradle.kts
```diff
@@ -69,10 +76,9 @@ tasks {
         }
     }
 
-    jar {
-        inputs.files(project(":sdk:trace-shaded-deps").file("src"))
-        val shadowJar = project(":sdk:trace-shaded-deps").tasks.named<Jar>("shadowJar")
-        from(zipTree(shadowJar.get().archiveFile))
-        dependsOn(shadowJar)
+    withType<AnimalSniffer>().configureEach {
```
anuraaga, line 79: @jkwatson We should have had animalsniffer plugin on trace-shaded-deps, and it would have helped catch #3339 early. Now with the classes added to the main sourceset, the plugin in this configuration is taking care of it though. Give trace-shaded-deps isn't published, and it's really the job of anything "using it" to add constraints like animalsniffer, I guess this matches reality a bit better.

mateuszrzeszutek (approved):

trask (approved):
👍

I'm not sure if it's applicable in this PR, but when applying this pattern to https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3471, I also added `includeEmptyDirs = false` to the copy task which avoided empty directories left-behind from the original package dirs

jkwatson (approved):
Thanks!

PR #3372, opened 2021-07-05
Make MonotonicClock not implement Clock and move to package private.

anuraaga:
Inspired by #3359 but separate because it can. While writing the javadoc for `Clock.now()` I realized this one doesn't implement `Clock.now()` (it has the caveat of getting out of sync with kernel time if not created per span). This seems to be an implementation detail of spans (at least for now, if we need it for Logs? we can move it again when the time comes, presumably the monotonic clock would need to be put in `Context` or similar for that to be possible anyways).


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -57,7 +56,7 @@
   // The kind of the span.
   private final SpanKind kind;
   // The clock used to get the time.
-  private final Clock clock;
+  private final MonotonicClock clock;
```
anuraaga, line 59: Given `RecordEventsReadableSpan` wouldn't work correctly with any other `Clock` this seems like an unexpectedly big improvement

Oberon00:
Some related thoughts: In practice, to use custom timestamps with things like end(), addEvent(), it may be helpful if the span had a `now()` function, or a `getClock()`. For example: I call `now`, then I collect some event attributes that may take a bit of time to collect, then I add the event with the previous timestamp. If I would use the SystemClock here, this would be a huge loss of accuracy.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/MonotonicClock.java
```diff
@@ -19,7 +19,7 @@
  * it is not recommended to use only one instance for a very long period of time.
  */
 @Immutable
-public final class MonotonicClock implements Clock {
```
jsuereth, line 22: Just an FYI - We may want to use this clock for the upcoming "Timer" instrument in Metrics.

anuraaga: Yeah I think we might move it to a public place again in some form. I want to give it more though at the time though, at least for now the biggest problem this PR is solving is that spans must have this clock (a rename to `TraceClock` might be warranted) and that it's not a `Clock` as per the SDK interface definition.

anuraaga: `MonotonicClock` exists to be anchored to a trace - just want to confirm you need `MonotonicClock` for the `Timer` and not just `Clock.nanoTime()`? (it is a bit weird that `Clock` exposes a monotonic time method in addition to wall time, it's effectively both a normal and monotonic clock, just not anchored to a trace). The only purpose of `MonotonicClock` is to provide "more accurate" wall timestamps, through its specialization of `now()`, I wouldn't expect it to be needed to capture a duration.

anuraaga:
I tried renaming to AnchoredClock since there might be confusion by the current name - MonotonicClock is not meant to be used to obtain monotonic time, it's only to obtain wall time (in a monotonic way)

sdk/common/src/main/java/io/opentelemetry/sdk/internal/MonotonicClock.java
```diff
@@ -1,59 +0,0 @@
...
-
-/**
- * This class provides a mechanism for calculating the epoch time using {@link System#nanoTime()}
- * and a reference epoch timestamp.
- *
- * <p>This is needed because Java has millisecond granularity for epoch times and tracing events are
- * recorded more often.
- *
- * <p>This clock needs to be re-created periodically in order to re-sync with the kernel clock, and
- * it is not recommended to use only one instance for a very long period of time.
```
jkwatson, line 19: do you think we should bring this comment over from here to the new class? I'm not sure I understand exactly what the problem is that's being called out here. I believe that `nanoTime()` is guaranteed to be monotonic, so I'm not sure what this is getting at, TBH.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/AnchoredClock.java
```diff
@@ -0,0 +1,54 @@
...
+ * monotonic time} to the anchor point.
+ *
+ * <p>This is needed because Java has lower granularity for epoch times and tracing events are
+ * recorded more often. There is also a performance improvement in avoiding referencing the system's
+ * wall time where possible. Instead of computing a true wall time for every timestamp within a
+ * trace, we compute it once at the local root and then anchor all descendant span timestamps to
+ * this root's timestamp.
+ */
+@Immutable
+final class AnchoredClock {
```
jkwatson, line 23: maybe "`TimingClock`"?

sdk/common/src/main/java/io/opentelemetry/sdk/internal/MonotonicClock.java
jkwatson: Or was this comment about the fact that `now()` and `nanoTime()` weren't really coordinated and consistent, maybe?

jkwatson (approved):
Not sure I love the name, but it's package-private so it could be tweaked.

sdk/common/src/main/java/io/opentelemetry/sdk/internal/MonotonicClock.java
anuraaga: It's about the fact that e.g., if you have a global MonotonicClock and just used that forever, if the user changed the time on their computer (or NTP did or whatever), the MonotonicClock wouldn't respect it since it doesn't access wall time after initialization. But since this is basically an implementation detail and we don't actually do anything special, we create a MonotonicClock on the start of the trace, figured it's not such a useful comment.

PR #3375, opened 2021-07-06
Sync .editorconfig with instrumentation repo

trask:
If this is acceptable, I can reformat all the `.gradle` and `.kts` files either in this PR or subsequent PR (to change indent from 4 to 2 spaces).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -485,9 +485,7 @@ ij_shell_redirect_followed_by_space = false
 ij_shell_switch_cases_indented = false
 
 [{*.gant, *.gradle, *.groovy, *.gson, *.gy}]
-indent_size = 4
-tab_width = 4
-ij_continuation_indent_size = 8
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -529,7 +527,7 @@ ij_groovy_call_parameters_wrap = off
 ij_groovy_catch_on_new_line = false
 ij_groovy_class_annotation_wrap = split_into_lines
 ij_groovy_class_brace_style = end_of_line
-ij_groovy_class_count_to_use_import_on_demand = 5
+ij_groovy_class_count_to_use_import_on_demand = 999
```
jkwatson, line 530: There's no groovy in this codebase!  ;)

jkwatson (approved):

.editorconfig
trask: 😭

anuraaga (approved):

PR #3377, opened 2021-07-08
Add vararg hasEvents assertion

anuraaga:
We have this for all the other Assert classes that operate on an item in a list, just not EventDataAssert. Noticed it can be helpful e.g., for asserting on a single event.


mateuszrzeszutek (approved):

jsuereth (approved):

sdk/testing/src/test/java/io/opentelemetry/sdk/testing/assertj/OpenTelemetryAssertionsTest.java
```diff
@@ -162,6 +162,8 @@ void passing() {
                       attributes -> assertThat(attributes).isEqualTo(Attributes.empty()))
                   .hasAttributesSatisfying(attributes -> assertThat(attributes).isEmpty());
             })
+        .hasEventsSatisfyingExactly(
```
jkwatson, line 165: do we want a negative test as well?

jkwatson (approved):
could use a negative test, but 👍🏼  otherwise

PR #3378, opened 2021-07-08
Add hasAttributes which takes a list of entries.

anuraaga:
For attributes it's very common to just assert the entries in the attributes, so this can reduce two indents

### DIFF
--- a/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/SpanDataAssert.java
+++ b/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/SpanDataAssert.java
@@ -7,7 +7,9 @@ package io.opentelemetry.sdk.testing.assertj;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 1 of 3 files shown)


mateuszrzeszutek (approved):

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/SpanDataAssert.java
```diff
@@ -210,6 +213,18 @@ public SpanDataAssert hasAttributes(Attributes attributes) {
     return this;
   }
 
+  /** Asserts the span has the given attributes. */
+  @SuppressWarnings({"rawtypes", "unchecked"})
+  @SafeVarargs
+  public final SpanDataAssert hasAttributes(Map.Entry<? extends AttributeKey<?>, ?>... entries) {
```
jkwatson, line 219: this ain't pretty, but I guess it works.

jkwatson (approved):

PR #3380, opened 2021-07-08
Provide an option for disabling errorprone for a smoother dev experience

jkwatson:
Just `-PdisableErrorProne=true` or put it in your personal gradle.properties file.

### DIFF
--- a/buildSrc/src/main/kotlin/otel.errorprone-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.errorprone-conventions.gradle.kts
@@ -9,10 +9,14 @@ plugins {
 
 val enableNullaway: String? by project
 
+val disableErrorProne = properties["disableErrorProne"]
+
 tasks {
     withType<JavaCompile>().configureEach {
         with(options) {
             errorprone {
+                enabled = (disableErrorProne != "true")
+
                 disableWarningsInGeneratedCode.set(true)
                 allDisabledChecksAsWarnings.set(true)
 
--- a/buildSrc/src/main/kotlin/otel.errorprone-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.errorprone-conventions.gradle.kts
@@ -9,13 +9,16 @@ plugins {
 
 val enableNullaway: String? by project
 
-val disableErrorProne = properties["disableErrorProne"]
+val disableErrorProne = properties["disableErrorProne"]?.toString()?.toBoolean() ?: false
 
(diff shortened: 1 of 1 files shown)


jsuereth (approved):

buildSrc/src/main/kotlin/otel.errorprone-conventions.gradle.kts
```diff
@@ -9,10 +9,14 @@ plugins {
 
 val enableNullaway: String? by project
 
+val disableErrorProne = properties["disableErrorProne"]
+
 tasks {
     withType<JavaCompile>().configureEach {
         with(options) {
             errorprone {
+                enabled = (disableErrorProne != "true")
```
anuraaga, line 18: We had an issue in build scripting once of comparing a boolean to the string `"true"` cause issues so want to stop this admittedly lazy pattern that you might have copied from the other repo. 

https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3530

jkwatson: what was the issue that PR solved? Trying to understand how it would end up a boolean and not as a string...

anuraaga: This one https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3340/files

anuraaga: I think the problem is properties are `Object`, not `String`

jkwatson: right. In this particular narrow case it probably doesn't matter much, since it's a string if you put it on the command line or in your gradle.properties, but there's no reason not to be extra careful about it anyway.

jkwatson: one difference between the impl over in the instrumentation repo is that you're pulling from `gradle.startParameters.projectProperties` and I'm pulling from `properties`. Yours is actually a String, whereas mine is an Object, so I don't have a `toBoolean()` method unless I cast it to a String. 

Why did you pull from `startParameters`, btw? When I used that pattern, it didn't seem to pick up my personal `gradle.properties` in the .gradle directory in my home directory, where as just using `properties` directly did.

kubawach (approved):

anuraaga (approved):

PR #3381, opened 2021-07-08
Update the semconv constants to v1.5.0

jkwatson:
Also, tweak our checkstyle settings to allow long markdown-style links.

### DIFF
--- a/buildscripts/checkstyle.xml
+++ b/buildscripts/checkstyle.xml
@@ -41,10 +41,6 @@
             <property name="allowByTailComment" value="true"/>
             <property name="allowNonPrintableEscapes" value="true"/>
         </module>
-        <module name="LineLength">
-            <property name="max" value="100"/>
-            <property name="ignorePattern" value="^package.*|^import.*|a href|href|http://|https://|ftp://|\[.*\]\(.*\)"/>
-        </module>
         <module name="AvoidStarImport"/>
         <module name="RedundantImport"/>
         <module name="OneTopLevelClass"/>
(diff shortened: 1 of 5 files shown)


carlosalberto (approved):

buildscripts/checkstyle.xml
```diff
@@ -43,7 +43,7 @@
         </module>
         <module name="LineLength">
```
anuraaga, line 44: Hmm - can we delete this? I think spotless already does the desired behavior, line length 100 except for long links

anuraaga (approved):

buildscripts/checkstyle.xml
anuraaga: As a principle I guess I'd remove any whitespace related checks from checkstyle since spotless should be handling

jkwatson: does it handle markdown-style links, though? Check out the diff in the ResourceAttributes here. Would it handle that, I wonder?

anuraaga: I think more precisely, what it does it break at 100, except when a single token is longer than 100 chars, so wouldn't matter what type of link it is. I guess it's pretty sane behavior since we generally don't have a token longer than 100 chars unless we really needed to have it.

jkwatson: I really wish that the semconv yaml didn't have markdown in it, because we're just getting that from the generator.

Oberon00: If someone has time, the generator could be adapted to render the markdown to HTML, fit for Javadoc.

jkwatson: @anuraaga I finally see what you are saying...why are we checkstyling when we're already spotlessing? Great question. I'd be happy to remove this check and let spotless just do our formatting for us.

PR #3384, opened 2021-07-09
Schedule IntervalMetricReader with fixed delay.

anuraaga:
Just randomly found this when looking at the class, I'm surprised we wouldn't use a fixed delay for an interval reader.

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
@@ -121,6 +121,7 @@ public final class IntervalMetricReader {
       doRun().join(internalState.getExportIntervalMillis(), TimeUnit.MILLISECONDS);
     }
 
+    // Make sure not to throw any exceptions.
     CompletableResultCode doRun() {
       final CompletableResultCode flushResult = new CompletableResultCode();
       if (exportAvailable.compareAndSet(true, false)) {
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
@@ -98,7 +98,7 @@ public final class IntervalMetricReader {
         return this;
       }
       scheduledFuture =
-          scheduler.scheduleAtFixedRate(
+          scheduler.scheduleWithFixedDelay(
               exporter,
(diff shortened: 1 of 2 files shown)


Oberon00:
I think there is a case for the fixed rate. If you view the data as e.g. a barchart where every minute is a bar, and you have an interval of one minute, you are more likely to get bars with twice the data and bars with no data if you use a fixed delay than if using the fixed rate.

That said, the fixed rate definitely has problems. E.g. if a slow synchronous export took longer than a few intervals, the timer will fire without delay a few times to "catch up".

anuraaga:
Do you have a recommendation? :P I think this interval is mostly about controlling load on a backend. If that's true, I suspect we want to ensure it is treated as a minimum rather than estimate.

kubawach (approved):

jkwatson:
I don't think there's a right answer on this one. Both potentially have issues, both have benefits. I think fixed rate is better, in general, though. If metrics export is taking longer than the configured interval, then something is very wrong somewhere.

The main consideration, I think, should be for the behavior of asynchronous instruments, since they are only recorded on export. How important is it that they are recorded on a regular interval?

Oberon00:
If we are considering mostly async "instant finish" exporters, I would definitely keep fixed rate, as then in any given time interval, the number of exports is about the same, whereas with fixed delay you may have more drift over time.

anuraaga:
Well I think drift is often considered good - if multiple servers start up at the same time they will together send bursts of traffic on the fixed rate forever.

kubawach:
Looking purely from backend point of view (worked a lot in backend-ish software development previously) I'd definitely appreciate fixed delay. We had for example scenario when "ingest" took a lot of time due to network latency (multiple proxies, sub-optimal geo distribution) and we couldn't do anything about it :)

jsuereth:
Backends I know either expect a fixed delay *or* coerce the data to look like it.   The closer we can remain to a fixed timer (and we won't ever be exact) the better it is on backends like this.

At a minimum, this should be a configuration option on `IntervalReader` (which is likely changing its configuration/setup significantly in the new metrics SDK), but at a maximum I'd say leave it as it is for better backend behavior.

jkwatson:
> Backends I know either expect a fixed delay _or_ coerce the data to look like it. The closer we can remain to a fixed timer (and we won't ever be exact) the better it is on backends like this.

Do you mean rate here, rather than delay?

jsuereth:
@jkwatson Yeah, I meant a fixed rate.  I.e. I'm not in favor of this change.

anuraaga:
Alrighty

PR #3385, opened 2021-07-09
Allow forceFlush of IntervalMetricReader, globally too for autoconfig…

anuraaga:
…ure / agent.

@jkwatson Sorry for this late change - it would be really helpful for some prototypes of metrics on AWS Lambda to have this ability to force flush, and I'm hoping the alpha status of the metrics artifacts makes it ok.

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
@@ -18,6 +18,7 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.concurrent.Immutable;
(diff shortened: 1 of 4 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/IntervalMetricReader.java
```diff
@@ -30,12 +31,31 @@
...
+  /**
+   * {@linkplain IntervalMetricReader#forceFlush() Force flushes} the globally registered {@link
+   * IntervalMetricReader} if available, or does nothing otherwise.
+   */
+  public static CompletableResultCode forceFlushGlobal() {
+    IntervalMetricReader intervalMetricReader = globalIntervalMetricReader.get();
+    if (intervalMetricReader != null) {
+      return intervalMetricReader.forceFlush();
+    }
+    return CompletableResultCode.ofSuccess();
```
kubawach, line 52: Just wondering - perhaps we should log (low level) debug/info here? As it's now possible to start without registering global, a customer may do exactly that and expect forceFlushGlobal to work (even though careful reading of JavaDoc explains this ;) ).

kubawach (approved):
Assuming that @jkwatson doesn't mind changes in API it looks good :)

jkwatson (approved):
With the upcoming multi-pipeline metrics SDK, this won't make sense, but in the interim, this seems fine by me.

PR #3386, opened 2021-07-09
Add TraceAssert.getSpan

anuraaga:
For the common pattern of `hasSpansSatisfyingExactly`, there isn't any way to verify the parent of a span since we don't have any trace data in scope, only a TraceAssert. Allowing access of spans through the TraceAssert enables this and is similar to what we have in instrumentation repo (this is probably the biggest gap between the two right now). Also added a shortcut `hasParent` which seems worth it.

### DIFF
--- a/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/SpanDataAssert.java
+++ b/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/SpanDataAssert.java
@@ -112,6 +112,11 @@ public final class SpanDataAssert extends AbstractAssert<SpanDataAssert, SpanDat
     return this;
   }
 
+  /** Asserts the span has the given parent {@link SpanData span}. */
+  public SpanDataAssert hasParent(SpanData parent) {
+    return hasParentSpanId(parent.getSpanId());
+  }
+
   /** Asserts the span has the given {@link Resource}. */
   public SpanDataAssert hasResource(Resource resource) {
     isNotNull();
--- a/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/SpanDataAssert.java
+++ b/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/SpanDataAssert.java
@@ -112,7 +112,11 @@ public final class SpanDataAssert extends AbstractAssert<SpanDataAssert, SpanDat
     return this;
   }
 
-  /** Asserts the span has the given parent {@link SpanData span}. */
+  /**
+   * Asserts the span has the given parent {@link SpanData span}.
+   *
+   * <p>Equivalent to {@code span.hasParentSpanId(parent.getSpanId())}.
+   */
   public SpanDataAssert hasParent(SpanData parent) {
     return hasParentSpanId(parent.getSpanId());
   }
--- a/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/TraceAssert.java
+++ b/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/TraceAssert.java
@@ -52,6 +52,14 @@ public final class TraceAssert
     return this;
   }
 
+  /**
+   * Returns the {@linkplain SpanData span} at the {@code index} within the trace. This can be
+   * useful for asserting the parent of a span.
+   */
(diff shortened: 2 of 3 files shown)


sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/TraceAssert.java
```diff
@@ -52,6 +52,14 @@ public final TraceAssert hasSpansSatisfyingExactly(Consumer<SpanDataAssert>... a
     return this;
   }
 
+  /**
+   * Returns the {@linkplain SpanData span} at the {@code index} within the trace. This can be
+   * useful for asserting the parent of a span.
+   */
+  public SpanData getSpan(int index) {
+    return actual.get(0);
```
kubawach, line 60: Shouldn't this be `actual.get(index)` ?

sdk/testing/src/test/java/io/opentelemetry/sdk/testing/junit5/OpenTelemetryExtensionTest.java
```diff
@@ -85,7 +85,12 @@ public void exportTraces() {
             trace ->
                 trace
                     .hasTraceId(traceId)
-                    .hasSpansSatisfyingExactly(s -> s.hasName("testa1"), s -> s.hasName("testa2"))
+                    .hasSpansSatisfyingExactly(
+                        s -> s.hasName("testa1"),
+                        s ->
+                            s.hasName("testa2")
+                                .hasParentSpanId(trace.getSpan(0).getSpanId())
```
kubawach, line 92: So with current "hasParent" these two check exactly the same (although the second one is a bit shorter). Was the intention here to demonstrate that? :)

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/SpanDataAssert.java
```diff
@@ -112,6 +112,11 @@ public SpanDataAssert hasParentSpanId(String parentSpanId) {
     return this;
   }
 
+  /** Asserts the span has the given parent {@link SpanData span}. */
```
kubawach, line 115: I believe that this method is slightly misleading, as it checks only if current span has parentId equal to argument's id. It does not check whole span (all attributes). It does yield shorter, more readable code so perhaps should have a different name to avoid any confusion (say `hasParentId` - overloaded method) ?

sdk/testing/src/test/java/io/opentelemetry/sdk/testing/junit5/OpenTelemetryExtensionTest.java
anuraaga: That was the intent indeed. And apparently I need to check with the non-0 index in the test...

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/SpanDataAssert.java
anuraaga: What does it mean to check the whole span in this case? I feel it's pretty accurate even if just checking id

kubawach: If I were an average Joe user I'd assume that hasParentId() will check if parent of the current span has the specified ID, while hasParent(span) will check that the parent is a particular span (all attributes matching). So in order to avoid confusion, I believe that this information (only ID is asserted) could be included either in name or in javadoc of this method.

anuraaga: I think the testing artifact isn't for average Joes so the APIs don't have to be as simple, there's a stronger emphasis on convenience to make sure people write instrumentation I think. But more docs is always good, I added the code snippet of equivalence does it help clear it up?

jkwatson (approved):

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/SpanDataAssert.java
kubawach: Definitely! Thanks :)

PR #3389, opened 2021-07-09
Update docs for 1.4.0

anuraaga:


jkwatson (approved):

jkwatson:
this also needs the apidiffs updated. I'll amend it and get them both committed.

jkwatson:
> this also needs the apidiffs updated. I'll amend it and get them both committed.

nm, looks like I don't have permission to tweak your PR branch

anuraaga:
@jkwatson Weird that shouldn't be the case but I'll confirm the setting next time. Incidentally I had a similar incident too before because I use the `gh` CLI which had a bug in a previous version preventing pushing to people's branches that was fixed by updating. Just in case that's possibly the case here too.

PR #3394, opened 2021-07-12
Tweak the semantic convention code generation.

jkwatson:
In particular, to better support links and html in the docs.


kubawach (approved):
👍

buildscripts/semantic-convention/templates/SemanticAttributes.java.j2
```diff
@@ -64,10 +64,10 @@ public final class {{class}} {
   {%- for attribute in attributes if attribute.is_local and not attribute.ref %}
 
   /**
-  * {% filter escape %}{{attribute.brief | to_doc_brief}}.{% endfilter %}
+  * {% filter safe %}{{attribute.brief | render_markdown}}{% endfilter %}
```
Oberon00, line 67:
```suggestion
  * {{attribute.brief | render_markdown | safe}}
```

buildscripts/semantic-convention/templates/SemanticAttributes.java.j2
```diff
@@ -64,10 +64,10 @@ public final class {{class}} {
...
     {%- if attribute.note %}
   *
-  * Note: {% filter escape %}{{attribute.note | to_doc_brief}}.{% endfilter %}
+  * <p>Notes: {% filter safe %}{{attribute.note | render_markdown}}{% endfilter %}
```
Oberon00, line 70:
```suggestion
  * <p>Notes: {{attribute.note | render_markdown | safe}}
```

Oberon00 (approved):
You could also try using `render_markdown(code="{{@code {0}}}")`. Pro: Will look a bit more idiomatic, con: Would probably break if there are `}` in the input.

Oberon00:
Unfortunately, there is no documentation for this one. 😬
* This is the parameter: https://github.com/open-telemetry/build-tools/blob/v0.4.1/semantic-conventions/src/opentelemetry/semconv/templating/code.py#L42
* Which is used here: https://github.com/open-telemetry/build-tools/blob/v0.4.1/semantic-conventions/src/opentelemetry/semconv/templating/code.py#L102
* This uses Python's `str.format` which uses the syntax documented here: https://docs.python.org/3/library/string.html#formatstrings

Oberon00:
So instead of `<code>sometext</code>` you will end up with `{@code sometext}`.

buildscripts/semantic-convention/templates/SemanticAttributes.java.j2
jkwatson: I don't even think it needs the `| safe` which I believe is a no-op.

jkwatson:
> Unfortunately, there is no documentation for this one. 😬
> 
> * This is the parameter: https://github.com/open-telemetry/build-tools/blob/v0.4.1/semantic-conventions/src/opentelemetry/semconv/templating/code.py#L42
> * Which is used here: https://github.com/open-telemetry/build-tools/blob/v0.4.1/semantic-conventions/src/opentelemetry/semconv/templating/code.py#L102
> * This uses Python's `str.format` which uses the syntax documented here: https://docs.python.org/3/library/string.html#formatstrings

Also unfortunately, it doesn't look like it works. :(

```
  File "/templates/SemanticAttributes.java.j2", line 67, in template
    * {{attribute.brief | render_markdown(code="{{@code {0}}}"}}
jinja2.exceptions.TemplateSyntaxError: unexpected '}', expected ')'
```

Oberon00:
> Also unfortunately, it doesn't look like it works. :(

Because you forgot the closing `)` between the `"` and the final `}}`. 😃

jkwatson:
> > Also unfortunately, it doesn't look like it works. :(
> 
> 
> 
> Because you forgot the closing `)` between the `"` and the final `}}`. 😃 

Dangit. This code ends up looking like perl.

Oberon00:
Yeah, too many nesting levels here with quotes and various brackets.

jkwatson:
Now I have the power! https://github.com/open-telemetry/opentelemetry-java/pull/3398

PR #3397, opened 2021-07-14
Populate labels in addition to attributes for now on metrics.

anuraaga:
We switched the SDK to populate attributes instead of labels in 1.3.0. However, the most popular OTLP backend, the OpenTelemetry Collector, still does not support the new format. We probably should have waited until that before changing our output, or alternatively output both.

It means though that metrics are mostly unusable (no ability to set any label) for 1.3.0+ (including 1.4) and the collector and I think we do need to release a patch to at least get this working again. I noticed the issue relatively soon after 1.3 in AWS integration tests but thought it was a compatibility issue with the tests themselves - I should trust the tests more ;)


exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MetricAdapter.java
```diff
@@ -328,5 +331,24 @@ private static ByteString convertSpanId(String id) {
         OtelEncodingUtils.bytesFromBase16(id, SpanId.getLength()));
   }
 
+  // Fill labels too until Collector supports attributes and users have had a chance to update.
```
kubawach, line 334: NIT: mark as TODO so that it stands-out as something to change in the near future

kubawach (approved):

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MetricAdapter.java
jkwatson: Not sure if it's exactly a TODO since it's not actionable until the collector updates. Any ideas on how best to track it and not forget?

jkwatson (approved):
👍🏼

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MetricAdapter.java
```diff
@@ -328,5 +331,24 @@ private static ByteString convertSpanId(String id) {
...
+  @SuppressWarnings("deprecation")
+  private static void fillAttributes(
+      Attributes attributes,
+      Consumer<KeyValue> attributeSetter,
+      Consumer<io.opentelemetry.proto.common.v1.StringKeyValue> labelSetter) {
+    attributes.forEach(
+        (key, value) -> {
+          attributeSetter.accept(CommonAdapter.toProtoAttribute(key, value));
+          if (key.getType() == AttributeType.STRING) {
```
jsuereth, line 343: Nit: You should be `.toString()` some of the other primitive types (like int, double, boolean).

jsuereth (approved):
Change LGTM, with one comment.  We haven't formalized the AttributeValue => String conversion e.g. for Prometheus exporter yet, but you should (as best you can) toString non-strin gattributes until we can drop the labels completely.

We also need to formalize this string mapping for the prometheus exporter, but we should NOT drop attributes.

jkwatson:
> Change LGTM, with one comment. We haven't formalized the AttributeValue => String conversion e.g. for Prometheus exporter yet, but you should (as best you can) toString non-strin gattributes until we can drop the labels completely.
> 
> We also need to formalize this string mapping for the prometheus exporter, but we should NOT drop attributes.

well, we won't drop attributes right now, since the API still only supports Labels, right?  ;)

jsuereth:
@jkwatson Ha!  Right.  I'll be sad on my branch where I started using `AttributeKey<boolean>`, but I can look into fixing that when I cause mischief there.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MetricAdapter.java
anuraaga: Ok went with toString without checking type

PR #3409, opened 2021-07-19
Add assertion for exception

anuraaga:
Exceptions are converted to data by our SDK so it makes sense to have an assertion matching the logic


jkwatson (approved):
Good add!

PR #3412, opened 2021-07-20
Add containsEntry overload that accepts int

anuraaga:
Similar to what we do for `put`

### DIFF
--- a/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/AttributesAssert.java
+++ b/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/AttributesAssert.java
@@ -38,6 +38,11 @@ public final class AttributesAssert extends AbstractAssert<AttributesAssert, Att
     return this;
   }
 
+  /** Asserts the attributes have the given key and value. */
+  public AttributesAssert containsEntry(AttributeKey<Long> key, int value) {
+    return containsEntry(key, (long) value);
+  }
+
   /** Asserts the attributes have the given key and string value. */
   public AttributesAssert containsEntry(String key, String value) {
     return containsEntry(AttributeKey.stringKey(key), value);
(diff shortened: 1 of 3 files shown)


kubawach (approved):
Looks good

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/AttributesAssert.java
```diff
@@ -38,6 +38,11 @@
     return this;
   }
 
+  /** Asserts the attributes have the given key and value. */
+  public AttributesAssert containsEntry(AttributeKey<Long> key, int value) {
```
jkwatson, line 42: why the `AttributeKey` parameter, rather than String `key` like the other methods?

anuraaga: This is the equivalent of the above method which uses AttributeKey. The String version doesn't need an overload since there's no type parameter match, int is automatically converted to long.

jkwatson: but why not just `(String, int)` like all the other specific methods?

anuraaga: Because it isn't needed - you can call `(String, long)` with an int, but you can't call `(Foo<T>, T)` with an int

anuraaga: And the reason that the `AttributeKey` form of the method is important is when using `SemanticAttributes.RESPONSE_LENGTH` for example

jkwatson (approved):

PR #3418, opened 2021-07-21
OTLP HTTP trace exporter

jack-berg:
Draft of a OTLP HTTP trace exporter implementation. Wanted to get some feedback on the direction before I spend time polishing.

### DIFF
--- a/exporters/logging-otlp/src/main/java/io/opentelemetry/exporter/logging/otlp/OtlpJsonLoggingSpanExporter.java
+++ b/exporters/logging-otlp/src/main/java/io/opentelemetry/exporter/logging/otlp/OtlpJsonLoggingSpanExporter.java
@@ -5,8 +5,10 @@
 
 package io.opentelemetry.exporter.logging.otlp;
 
+import com.fasterxml.jackson.core.JsonFactory;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.io.SegmentedStringWriter;
+import io.opentelemetry.exporter.otlp.internal.HexEncodingStringJsonGenerator;
 import io.opentelemetry.exporter.otlp.internal.SpanAdapter;
 import io.opentelemetry.proto.trace.v1.ResourceSpans;
 import io.opentelemetry.sdk.common.CompletableResultCode;
@@ -25,7 +27,8 @@ import org.curioswitch.common.protobuf.json.MessageMarshaller;
  */
 public final class OtlpJsonLoggingSpanExporter implements SpanExporter {
 
-  private static final MessageMarshaller marshaller =
+  private static final JsonFactory JSON_FACTORY = new JsonFactory();
+  private static final MessageMarshaller MARSHALLER =
       MessageMarshaller.builder()
           .register(ResourceSpans.class)
           .omittingInsignificantWhitespace(true)
@@ -45,11 +48,9 @@ public final class OtlpJsonLoggingSpanExporter implements SpanExporter {
   public CompletableResultCode export(Collection<SpanData> spans) {
     List<ResourceSpans> allResourceSpans = SpanAdapter.toProtoResourceSpans(spans);
     for (ResourceSpans resourceSpans : allResourceSpans) {
-      SegmentedStringWriter sw =
-          new SegmentedStringWriter(
-              HexEncodingStringJsonGenerator.JSON_FACTORY._getBufferRecycler());
-      try (JsonGenerator gen = HexEncodingStringJsonGenerator.create(sw)) {
-        marshaller.writeValue(resourceSpans, gen);
+      SegmentedStringWriter sw = new SegmentedStringWriter(JSON_FACTORY._getBufferRecycler());
+      try (JsonGenerator gen = HexEncodingStringJsonGenerator.create(sw, JSON_FACTORY)) {
+        MARSHALLER.writeValue(resourceSpans, gen);
       } catch (IOException e) {
         // Shouldn't happen in practice, just skip it.
         continue;
--- a/exporters/otlp/common/build.gradle.kts
+++ b/exporters/otlp/common/build.gradle.kts
@@ -15,6 +15,7 @@ dependencies {
     api(project(":sdk:all"))
     api(project(":sdk:metrics"))
 
+    api("org.curioswitch.curiostack:protobuf-jackson")
     implementation("com.google.protobuf:protobuf-java")
 
     testImplementation(project(":sdk:testing"))
--- a/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/HexEncodingStringJsonGenerator.java
+++ b/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/HexEncodingStringJsonGenerator.java
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package io.opentelemetry.exporter.logging.otlp;
+package io.opentelemetry.exporter.otlp.internal;
 
 import com.fasterxml.jackson.core.Base64Variant;
 import com.fasterxml.jackson.core.JsonFactory;
@@ -13,14 +13,19 @@ import com.fasterxml.jackson.core.util.JsonGeneratorDelegate;
 import io.opentelemetry.api.internal.TemporaryBuffers;
 import java.io.IOException;
 
-final class HexEncodingStringJsonGenerator extends JsonGeneratorDelegate {
+public final class HexEncodingStringJsonGenerator extends JsonGeneratorDelegate {
 
-  static final JsonFactory JSON_FACTORY = new JsonFactory();
-
-  static JsonGenerator create(SegmentedStringWriter stringWriter) {
+  /**
+   * Create a JSON generator which encodes byte arrays as case-insensitive hex-encoded strings.
+   *
+   * @param stringWriter the string write
+   * @param jsonFactory the json factory
+   * @return the generator
+   */
+  public static JsonGenerator create(SegmentedStringWriter stringWriter, JsonFactory jsonFactory) {
     final JsonGenerator delegate;
     try {
-      delegate = JSON_FACTORY.createGenerator(stringWriter);
+      delegate = jsonFactory.createGenerator(stringWriter);
     } catch (IOException e) {
       throw new IllegalStateException("Unable to create in-memory JsonGenerator, can't happen.", e);
     }
--- a/exporters/otlp/trace/build.gradle.kts
+++ b/exporters/otlp/trace/build.gradle.kts
@@ -28,6 +28,10 @@ dependencies {
     implementation("io.grpc:grpc-stub")
     implementation("com.google.protobuf:protobuf-java")
 
+    implementation("com.squareup.okhttp3:okhttp:4.9.0")
+    implementation("com.squareup.okhttp3:okhttp-tls:4.9.0")
+    implementation("com.google.protobuf:protobuf-java-util:3.17.3")
+
     testImplementation(project(":sdk:testing"))
 
     testImplementation("io.grpc:grpc-testing")
--- a/exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpHttpSpanExporter.java
+++ b/exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpHttpSpanExporter.java
@@ -1,0 +1,211 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.exporter.otlp.trace;
+
+import com.google.rpc.Status;
+import io.opentelemetry.api.metrics.BoundLongCounter;
+import io.opentelemetry.api.metrics.GlobalMeterProvider;
+import io.opentelemetry.api.metrics.LongCounter;
+import io.opentelemetry.api.metrics.Meter;
+import io.opentelemetry.api.metrics.common.Labels;
+import io.opentelemetry.exporter.otlp.internal.SpanAdapter;
+import io.opentelemetry.proto.collector.trace.v1.ExportTraceServiceRequest;
+import io.opentelemetry.sdk.common.CompletableResultCode;
+import io.opentelemetry.sdk.internal.ThrottlingLogger;
+import io.opentelemetry.sdk.trace.data.SpanData;
+import io.opentelemetry.sdk.trace.export.SpanExporter;
+import java.io.IOException;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.annotation.concurrent.ThreadSafe;
+import okhttp3.Call;
+import okhttp3.Callback;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.RequestBody;
+import okhttp3.Response;
+import okhttp3.ResponseBody;
+import org.jetbrains.annotations.NotNull;
+
+/** Exports spans using OTLP via HTTP, using OpenTelemetry's protobuf model. */
+@ThreadSafe
+public final class OtlpHttpSpanExporter implements SpanExporter {
+
+  private static final String EXPORTER_NAME = OtlpHttpSpanExporter.class.getSimpleName();
+  private static final Labels EXPORTER_NAME_LABELS = Labels.of("exporter", EXPORTER_NAME);
+  private static final Labels EXPORT_SUCCESS_LABELS =
+      Labels.of("exporter", EXPORTER_NAME, "success", "true");
+  private static final Labels EXPORT_FAILURE_LABELS =
+      Labels.of("exporter", EXPORTER_NAME, "success", "false");
+
+  private final ThrottlingLogger logger =
+      new ThrottlingLogger(Logger.getLogger(OtlpHttpSpanExporter.class.getName()));
+
+  private final BoundLongCounter spansSeen;
+  private final BoundLongCounter spansExportedSuccess;
+  private final BoundLongCounter spansExportedFailure;
+
+  private final OkHttpClient client;
+  private final String endpoint;
+  private final Map<String, List<String>> headers;
+  private final RequestResponseHandler requestResponseHandler;
+
+  OtlpHttpSpanExporter(
+      OkHttpClient client,
+      String endpoint,
+      Map<String, List<String>> headers,
+      RequestResponseHandler requestResponseHandler) {
+    // TODO: should this be io.opentelemetry.exporters.otlp with a label that indicates the
+    // protocol?
+    Meter meter = GlobalMeterProvider.getMeter("io.opentelemetry.exporters.otlp-http");
+    this.spansSeen =
+        meter.longCounterBuilder("spansSeenByExporter").build().bind(EXPORTER_NAME_LABELS);
+    LongCounter spansExportedCounter = meter.longCounterBuilder("spansExportedByExporter").build();
+    this.spansExportedSuccess = spansExportedCounter.bind(EXPORT_SUCCESS_LABELS);
+    this.spansExportedFailure = spansExportedCounter.bind(EXPORT_FAILURE_LABELS);
+
+    this.client = client;
+    this.endpoint = endpoint;
+    this.headers = headers;
+    this.requestResponseHandler = requestResponseHandler;
+  }
+
+  /**
+   * Submits all the given spans in a single batch to the OpenTelemetry collector.
+   *
+   * @param spans the list of sampled Spans to be exported.
+   * @return the result of the operation
+   */
+  @Override
+  public CompletableResultCode export(Collection<SpanData> spans) {
+    spansSeen.add(spans.size());
+    ExportTraceServiceRequest exportTraceServiceRequest =
+        ExportTraceServiceRequest.newBuilder()
+            .addAllResourceSpans(SpanAdapter.toProtoResourceSpans(spans))
+            .build();
+
+    Request.Builder requestBuilder =
+        new Request.Builder()
+            .url(endpoint)
+            .post(requestResponseHandler.build(exportTraceServiceRequest));
+    if (headers != null) {
+      headers.forEach(
+          (key, values) -> values.forEach(value -> requestBuilder.addHeader(key, value)));
+    }
+
+    CompletableResultCode result = new CompletableResultCode();
+
+    client
+        .newCall(requestBuilder.build())
+        .enqueue(
+            new Callback() {
+              @Override
+              public void onFailure(@NotNull Call call, @NotNull IOException e) {
+                spansExportedFailure.add(spans.size());
+                result.fail();
+                logger.log(
+                    Level.SEVERE,
+                    "Failed to export spans. The request could not be executed. Full error message: "
+                        + e.getMessage());
+              }
+
+              @Override
+              public void onResponse(@NotNull Call call, @NotNull Response response) {
+                if (response.isSuccessful()) {
+                  spansExportedSuccess.add(spans.size());
+                  result.succeed();
+                  return;
+                }
+
+                spansExportedFailure.add(spans.size());
+                result.fail();
+                int code = response.code();
+
+                Status status;
+                try {
+                  status = requestResponseHandler.extractErrorStatus(response.body());
+                } catch (IOException e) {
+                  status =
+                      Status.newBuilder()
+                          .setMessage(
+                              "Unable to extract error message from request: " + e.getMessage())
+                          .build();
+                }
+
+                logger.log(
+                    Level.SEVERE,
+                    "Failed to export spans. Server responded with code "
+                        + code
+                        + ". Error message: "
+                        + status.getMessage());
+              }
+            });
+
+    return result;
+  }
+
+  /**
+   * The OTLP exporter does not batch spans, so this method will immediately return with success.
+   *
+   * @return always Success
+   */
+  @Override
+  public CompletableResultCode flush() {
+    return CompletableResultCode.ofSuccess();
+  }
+
+  /**
+   * Returns a new builder instance for this exporter.
+   *
+   * @return a new builder instance for this exporter.
+   */
+  public static OtlpHttpSpanExporterBuilder builder() {
+    return new OtlpHttpSpanExporterBuilder();
+  }
+
+  /**
+   * Returns a new {@link OtlpHttpSpanExporter} using the default values.
+   *
+   * @return a new {@link OtlpHttpSpanExporter} instance.
+   */
+  public static OtlpHttpSpanExporter getDefault() {
+    return builder().build();
+  }
+
+  /** Shutdown the exporter. */
+  @Override
+  public CompletableResultCode shutdown() {
+    // TODO: should we update a boolean shutdown flag which prevents future exports?
+    final CompletableResultCode result = CompletableResultCode.ofSuccess();
+    this.spansSeen.unbind();
+    this.spansExportedSuccess.unbind();
+    this.spansExportedFailure.unbind();
+    return result;
+  }
+
+  interface RequestResponseHandler {
+
+    /**
+     * Build a request body for the export trace service request.
+     *
+     * @param exportTraceServiceRequest the export trace service request
+     * @return the request body
+     */
+    RequestBody build(ExportTraceServiceRequest exportTraceServiceRequest);
+
+    /**
+     * Extract the status from the response body. Called when response status code is 4XX and 5XX.
+     *
+     * @param responseBody the response body
+     * @return the status
+     * @throws IOException if unable to extract a status from the response body
+     */
+    Status extractErrorStatus(ResponseBody responseBody) throws IOException;
+  }
+}
--- a/exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpHttpSpanExporterBuilder.java
+++ b/exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpHttpSpanExporterBuilder.java
@@ -1,0 +1,171 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.exporter.otlp.trace;
+
+import static io.opentelemetry.api.internal.Utils.checkArgument;
+import static java.util.Objects.requireNonNull;
+
+import com.google.common.base.Preconditions;
+import com.google.common.util.concurrent.MoreExecutors;
+import io.opentelemetry.exporter.otlp.trace.OtlpHttpSpanExporter.RequestResponseHandler;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.security.cert.CertificateException;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
+import okhttp3.Dispatcher;
+import okhttp3.OkHttpClient;
+import okhttp3.tls.HandshakeCertificates;
+
+/** Builder utility for this exporter. */
+public final class OtlpHttpSpanExporterBuilder {
+
+  private static final String DEFAULT_ENDPOINT = "http://localhost:55681/v1/traces";
+  private static final Encoding DEFAULT_ENCODING = Encoding.PROTOBUF;
+  private static final long DEFAULT_TIMEOUT_SECS = 10;
+
+  private String endpoint = DEFAULT_ENDPOINT;
+  private Encoding encoding = DEFAULT_ENCODING;
+  private boolean isCompressionEnabled = false;
+  private long timeoutNanos = TimeUnit.SECONDS.toNanos(DEFAULT_TIMEOUT_SECS);
+  @Nullable private Map<String, List<String>> headers;
+  @Nullable private byte[] trustedCertificatesPem;
+
+  /**
+   * Sets the OTLP endpoint to connect to. If unset, defaults to {@value DEFAULT_ENDPOINT}. The
+   * endpoint must start with either http:// or https://, and should append the version and signal
+   * to the path (i.e. v1/traces).
+   */
+  public OtlpHttpSpanExporterBuilder setEndpoint(String endpoint) {
+    requireNonNull(endpoint, "endpoint");
+
+    URI uri;
+    try {
+      uri = new URI(endpoint);
+    } catch (URISyntaxException e) {
+      throw new IllegalArgumentException("Invalid endpoint, must be a URL: " + endpoint, e);
+    }
+
+    if (uri.getScheme() == null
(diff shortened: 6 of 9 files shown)


jack-berg:
I added the `OtlpHttpSpanExporter` to the `:exporters:otlp:trace` to defer thinking about where to put it, but it will have to be addressed. 

Keeping it in `:exporter:otlp:trace` avoids extra modules / artifacts. It also removes the need to rename or think about renaming existing modules / packages / artifacts to disambiguate between `otlp/grpc` and `otlp/http`. The current naming conventions imply that otlp = `otlp/grpc`. We can keep that, and have new modules / packages / artifacts use `otlp-http` to disambiguate, but this asymmetry isn't ideal. Keeping the exporter in `:expoter:otlp:trace` sidesteps this whole conversation. This doesn't feel wrong to me, since `otlp/grpc` and `otlp/http` with JSON format and Protobuf format are all different flavors of `otlp`. 

Another thing to think is whether to shade the new dependencies on `okhttp`, and whether it matters based on where the exporter lives.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/HexEncodingStringJsonGenerator.java
```diff
@@ -13,14 +13,19 @@
 import io.opentelemetry.api.internal.TemporaryBuffers;
 import java.io.IOException;
 
-final class HexEncodingStringJsonGenerator extends JsonGeneratorDelegate {
+public final class HexEncodingStringJsonGenerator extends JsonGeneratorDelegate {
```
jack-berg, line 16: Moved to `:exporters:otlp:common` and made `create(..)` public because now both the otlp logging exporter and the otlp http exporter with JSON format both need this special encoding.

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpHttpSpanExporter.java
```diff
@@ -0,0 +1,211 @@
...
+  private final Map<String, List<String>> headers;
+  private final RequestResponseHandler requestResponseHandler;
+
+  OtlpHttpSpanExporter(
+      OkHttpClient client,
+      String endpoint,
+      Map<String, List<String>> headers,
+      RequestResponseHandler requestResponseHandler) {
+    // TODO: should this be io.opentelemetry.exporters.otlp with a label that indicates the
+    // protocol?
```
jack-berg, line 65: Thoughts?

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpHttpSpanExporter.java
```diff
@@ -0,0 +1,211 @@
...
+  public CompletableResultCode shutdown() {
+    // TODO: should we update a boolean shutdown flag which prevents future exports?
+    final CompletableResultCode result = CompletableResultCode.ofSuccess();
+    this.spansSeen.unbind();
+    this.spansExportedSuccess.unbind();
+    this.spansExportedFailure.unbind();
+    return result;
+  }
+
+  interface RequestResponseHandler {
```
jack-berg, line 192: Not in love with the name here, but this contains the behavior that varies based on whether JSON or Profobuf format is used.

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpHttpSpanExporter.java
```diff
@@ -0,0 +1,211 @@
...
+   * @return a new {@link OtlpHttpSpanExporter} instance.
+   */
+  public static OtlpHttpSpanExporter getDefault() {
+    return builder().build();
+  }
+
+  /** Shutdown the exporter. */
+  @Override
+  public CompletableResultCode shutdown() {
+    // TODO: should we update a boolean shutdown flag which prevents future exports?
```
jack-berg, line 184: Thoughts?

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpHttpSpanExporterBuilder.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    this.endpoint = endpoint;
+    return this;
+  }
+
+  /** Sets the encoding of payloads. If unset, defaults to {@link #DEFAULT_ENCODING}. */
+  public OtlpHttpSpanExporterBuilder setEncoding(Encoding encoding) {
+    // TODO: if we we don't anticipate additional encodings, could replace this with
+    // setJsonEncoding()
```
jack-berg, line 70: Thoughts? I don't anticipate additional formats besides JSON and Protobuf, so maybe `setJsonEncoding()` and `setProtobufEncoding()` method are a better idea.

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpHttpSpanExporterBuilder.java
```diff
@@ -0,0 +1,171 @@
...
+    if (isCompressionEnabled) {
+      clientBuilder.addInterceptor(OtlpHttpUtil.GZIP_INTERCEPTOR);
+    }
+
+    if (trustedCertificatesPem != null) {
+      try {
+        HandshakeCertificates handshakeCertificates =
+            OtlpHttpUtil.toHandshakeCertificates(trustedCertificatesPem);
+        clientBuilder.sslSocketFactory(
+            handshakeCertificates.sslSocketFactory(), handshakeCertificates.trustManager());
```
jack-berg, line 148: This code can probably fix this [TODO](https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterBuilder.java#L183).

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpHttpSpanExporterBuilder.java
```diff
@@ -0,0 +1,171 @@
...
+  /**
+   * Constructs a new instance of the exporter based on the builder's values.
+   *
+   * @return a new exporter's instance
+   */
+  public OtlpHttpSpanExporter build() {
+    OkHttpClient.Builder clientBuilder =
+        new OkHttpClient.Builder()
+            // TODO: which thread should handle these calls?
+            .dispatcher(new Dispatcher(MoreExecutors.newDirectExecutorService()))
```
jack-berg, line 137: Noticed that the `OtlpGrpcSpanExporter` executes using `MoreExecutors.directExecutor()` and followed that pattern here. Other options would be to execute requests in a blocking fashion or execute them on another group of threads.

exporters/otlp/common/build.gradle.kts
```diff
@@ -15,6 +15,7 @@ dependencies {
     api(project(":sdk:all"))
     api(project(":sdk:metrics"))
 
+    api("org.curioswitch.curiostack:protobuf-jackson")
```
jkwatson, line 18: do we really need to expose this as an `api` dependency?

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpHttpSpanExporter.java
jkwatson: well, this is just the meter name, so having include the protocol seems correct, I think. I wouldn't imagine someone would end up having both exporters in the same JVM so breaking out by label doesn't seem to add much value.

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpHttpSpanExporter.java
jkwatson: I'm not sure we need to be that draconian. We could evict all the http connections from the okhttp connection pool, though, as a final cleanup step.

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpHttpSpanExporterBuilder.java
jkwatson: yeah, I think two methods are better; we also end up having to expose one less enum to the public API of this module. If we ever end up with more encodings we can think about making it more flexible.

jkwatson:
I think it would be unfortunate if otlp grpc users suddenly got an okhttp dependency thrust upon them, forcing them to exclude it to keep their artifact smaller (and avoid version clashes). so, if we want to put them in the same module, we should definitely shade-in okhttp.

exporters/otlp/common/build.gradle.kts
jack-berg: I moved `HexEncodingStringJsonGenerator` to `:exporters:otlp:common` because multiple classes depend on the ability to hex encode trace id and span id. That class contains a method `public static JsonGenerator create(...)` which returns a class `org.curioswitch.curiostack:protobuf-jackson`. 

So without `api` here, modules dependent on this code have to specify their own dependency on `org.cuioswitch.curiostack:protobuf-jackson`.

anuraaga: Didn't look at the code yet but highest level comment is that I wasn't thinking of supporting json here. I don't know of any use case for it, as far as I know the main reason we have the protocol is for browsers where protobuf can cause challenges. Perhaps someone will ask for it later (I don't expect it TBH) but we can do it then.

anuraaga:
Let's use a different module - I expect a very small minority of users to need this module so want to reduce it's discoveribility vs our "standard recommendation" grpc.

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpHttpSpanExporterBuilder.java
anuraaga: grpc executes user callbacks on the direct executor, not the requests themselves which are always async on netty event loop.

IIRC this okhttp knob actually controls what threads requests run on, which we definitely don't want on the hot path. So let's leave this at the default.

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpHttpUtil.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+    return certBuilder.build();
+  }
+
+  private static class ProtobufRequestResponseHandler implements RequestResponseHandler {
+
+    @Override
+    public RequestBody build(ExportTraceServiceRequest exportTraceServiceRequest) {
+      return RequestBody.create(
+          exportTraceServiceRequest.toByteArray(), MediaType.parse("application/x-protobuf"));
```
anuraaga, line 66: It can and probably should be a future PR but would definitely be good to use okio buffer and codedinputstream to skip the intermediate unpooled buffer

exporters/otlp/common/build.gradle.kts
jack-berg: Well that simplifies a lot of stuff!

jack-berg:
> Let's use a different module - I expect a very small minority of users to need this module so want to reduce it's discoveribility vs our "standard recommendation" grpc.

@anuraaga thanks for the great feedback! Very helpful. For the separate modules, how about:
`:exporters:otlp:trace-http` and `:exporters:otlp:metrics-http` (metrics to be added in a future PR)?

exporters/otlp/trace/build.gradle.kts
```diff
@@ -28,6 +28,10 @@ dependencies {
     implementation("io.grpc:grpc-stub")
     implementation("com.google.protobuf:protobuf-java")
 
+    implementation("com.squareup.okhttp3:okhttp:4.9.0")
```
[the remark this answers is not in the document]
jack-berg: Ok! I'm inclined to keep the surface area smaller and choosing to either shade or not shade. Can always publish additional artifacts later, but it's harder to stop publishing.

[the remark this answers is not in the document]
jack-berg: On second thought, maybe okhttp3 isn't such a great idea. Its only getting critical fixes [through December 31st, 2021](https://square.github.io/okhttp/#requirements). And since we're going with the separate module route, it doesn't seem as critical that the kotlin stdlib is included.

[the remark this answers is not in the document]
jkwatson: 3.12.x isn't getting updates, but the 4.x will be.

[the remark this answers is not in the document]
anuraaga: Ah I didn't know about the EoL then seems ok (indeed especially with the separate module)

exporters/logging-otlp/src/main/java/io/opentelemetry/exporter/logging/otlp/OtlpJsonLoggingSpanExporter.java
```diff
@@ -5,6 +5,8 @@
 
 package io.opentelemetry.exporter.logging.otlp;
 
+import static io.opentelemetry.exporter.logging.otlp.HexEncodingStringJsonGenerator.JSON_FACTORY;
```
anuraaga, line 8: Believe we can revert any json changes

exporters/otlp-http/trace/README.md
```diff
@@ -0,0 +1,8 @@
+# OpenTelemetry - OTLP Trace Exporter - HTTP
+
+[![Javadocs][javadoc-image]][javadoc-url]
+
+This is the OpenTelemetry exporter, sending span data to OpenTelemetry collector via HTTP.
```
anuraaga, line 5: HTTP without gRPC

exporters/otlp-http/trace/src/main/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporter.java
```diff
@@ -0,0 +1,222 @@
...
+        gzipSink.close();
+      }
+    };
+  }
+
+  private static Status extractErrorStatus(Response response) {
+    try {
+      ResponseBody responseBody = response.body();
+      if (responseBody == null) {
+        return Status.newBuilder()
```
anuraaga, line 171: I think we need to pick a non-zero code for this.

exporters/otlp-http/trace/src/main/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporterBuilder.java
```diff
@@ -0,0 +1,161 @@
...
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
+import okhttp3.Headers;
+import okhttp3.OkHttpClient;
+import okhttp3.tls.HandshakeCertificates;
+
+/** Builder utility for this exporter. */
```
anuraaga, line 25: `Builder for {@link OtlpHttpSpanExporter}.`

exporters/otlp-http/trace/src/main/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporterBuilder.java
```diff
@@ -0,0 +1,161 @@
...
+   * unset, defaults to {@value DEFAULT_TIMEOUT_SECS}s.
+   */
+  public OtlpHttpSpanExporterBuilder setTimeout(Duration timeout) {
+    requireNonNull(timeout, "timeout");
+    return setTimeout(timeout.toNanos(), TimeUnit.NANOSECONDS);
+  }
+
+  /**
+   * Sets the OTLP endpoint to connect to. If unset, defaults to {@value DEFAULT_ENDPOINT}. The
+   * endpoint must start with either http:// or https://, and should append the version and signal
```
anuraaga, line 59: Should it append or must it? Or how about "include the full HTTP path" in case endpoint doesn't use the same format as the collector?

exporters/otlp-http/trace/src/main/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporterBuilder.java
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  /**
+   * Sets the method used to compress payloads. If unset, compression is disabled. Currently the
+   * only supported compression method is "gzip".
+   */
+  public OtlpHttpSpanExporterBuilder setCompression(String compressionMethod) {
+    requireNonNull(compressionMethod, "compressionMethod");
+    Preconditions.checkArgument(
+        "gzip".equals(compressionMethod),
```
anuraaga, line 89: We prefer to use yoda order only when needed. We checked null so the variable can be on left side.

exporters/otlp-http/trace/src/test/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporterTest.java
```diff
@@ -0,0 +1,168 @@
...
+import okhttp3.mockwebserver.MockResponse;
+import org.junit.Rule;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+import org.slf4j.event.Level;
+import org.slf4j.event.LoggingEvent;
+
+class OtlpHttpSpanExporterTest {
+
+  @Rule private final OtlpHttpRule otlpHttp = new OtlpHttpRule();
```
anuraaga, line 36: There seems to be a mix of junit5 and junit4 - does it work? Anyways we'd prefer to stick with junit5 - can you use the armeria mock server we have in a couple other tests?

https://github.com/open-telemetry/opentelemetry-java/blob/c3614b6b2345361a3484f92961231ca19b558242/sdk-extensions/aws/src/test/java/io/opentelemetry/sdk/extension/aws/internal/JdkHttpClientTest.java

Not sure but don't think you should need a custom dispatcher either, the mock server takes care of allowing assertions etc

exporters/otlp-http/trace/src/main/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporterBuilder.java
jkwatson: I personally prefer "yoda" order. Have we codified this style choice anywhere?

anuraaga: Don't think so, we probably could. I would have probably commented if you did it somewhere :P `if` statements read terribly with yoda order IMO

exporters/otlp-http/trace/src/main/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporter.java
jack-berg: The [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp) says:
> This specification does not use Status.code field and the server MAY omit Status.code field. The clients are not expected to alter their behavior based on Status.code field but MAY record it for troubleshooting purposes.

The status code field isn't currently recorded anywhere for troubleshooting purposes. Still, probably best to to set it in case future maintainers decide to use it. I'll go with `UNKNOWN = 2`.

exporters/otlp-http/trace/src/main/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporterBuilder.java
jack-berg: Well the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#configuration-options) says:
> When using OTEL_EXPORTER_OTLP_ENDPOINT, exporters SHOULD follow the collector convention of appending the version and signal to the path (e.g. v1/traces or v1/metrics), if not present already. 

It looks like I misread it. Looking at it again, it seems like exporters "SHOULD" append the version and signal to the path when the `OTEL_EXPORTER_OTLP_ENDPOINT` env variable / system property is used. So the `autoconfigure` module will want to append the version and signal, but the exporter class itself doesn't have to. 

I like your phrasing 😄

exporters/otlp-http/trace/src/test/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporterTest.java
jack-berg: It did work, but yeah that was definitely an oversight. I've switched to the armeria mock server - haven't used it before but I like the interface! Definitely simplifies stuff nicely.

exporters/otlp-http/trace/src/test/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporterTest.java
```diff
@@ -0,0 +1,246 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+import org.slf4j.event.Level;
+import org.slf4j.event.LoggingEvent;
+
+class OtlpHttpSpanExporterTest {
+
+  private static final MediaType APPLICATION_PROTOBUF =
+      MediaType.create("application", "x-protobuf");
+  private static final HeldCertificate HELD_CERTIFICATE;
```
jack-berg, line 54: Note: I've used the the certificate utility from `okhttp-tls` instead of the `SelfSignedCertificateExtension` seen in other tests. I did this because the certificate from `SelfSignedCertificateExtension` doesn't include a subject alternative name and the okhttp client fails TLS validation without this field.

anuraaga: Ah yeah I noticed this too separately - quite bizarre okhttp restriction

exporters/otlp-http/trace/build.gradle.kts
```diff
@@ -0,0 +1,34 @@
...
+
+  testImplementation("com.linecorp.armeria:armeria-junit5")
+}
+
+tasks {
+  shadowJar {
+    minimize()
+
+    relocate("okhttp3", "io.opentelemetry.internal.shaded.okhttp3")
+    relocate("okio", "io.opentelemetry.internal.shaded.okio")
```
anuraaga, line 32: We also have to relocate kotlin as well unfortunately.

exporters/otlp-http/trace/build.gradle.kts
```diff
@@ -0,0 +1,34 @@
...
+  implementation("com.squareup.okhttp3:okhttp-tls")
+  implementation("com.squareup.okio:okio")
+
+  testImplementation(project(":sdk:testing"))
+
+  testImplementation("com.linecorp.armeria:armeria-junit5")
+}
+
+tasks {
+  shadowJar {
```
anuraaga, line 28: This is going to result in publishing two artifacts, one with a `-all` classifier. I think if we shade we should publish the non-shaded version as well, so this seems ok. The `-shaded` classifier is probably easier to understand and used by some projects like gRPC.

That being said, I think I would like to go ahead and leave this out until someone asks for it - it's not hard to add a shaded version later, but in the meantime, documenting the two is a bit painful.

exporters/otlp-http/trace/src/main/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporter.java
```diff
@@ -0,0 +1,225 @@
...
+      public void writeTo(BufferedSink bufferedSink) throws IOException {
+        BufferedSink gzipSink = Okio.buffer(new GzipSink(bufferedSink));
+        requestBody.writeTo(gzipSink);
+        gzipSink.close();
+      }
+    };
+  }
+
+  private static Status extractErrorStatus(Response response) {
+    try {
```
anuraaga, line 169: Move the `try` down to make clearer the only code that can throw an exception is `parseFrom`

exporters/otlp-http/trace/src/main/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporter.java
```diff
@@ -0,0 +1,225 @@
...
+      ResponseBody responseBody = response.body();
+      if (responseBody == null) {
+        return Status.newBuilder()
+            .setMessage("Unable to extract error message from empty response body.")
```
anuraaga, line 173: Let's use `response.getMessage()` for the fallbacks, I think using it as is or something like `Could not extract Status from response body, HTTP message: ` are both reasonable.

FWIW, I think it's relatively common for the response body to be not-parsable, any off-the-shelf auth or throttling proxy is a perfectly valid use case but will not have a Status, so I don't think we should sound too scary

exporters/otlp-http/trace/src/main/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporter.java
```diff
@@ -0,0 +1,225 @@
...
+            .setCode(Code.UNKNOWN.getNumber())
+            .build();
+      } else {
+        return Status.parseFrom(responseBody.bytes());
+      }
+    } catch (IOException e) {
+      return Status.newBuilder()
+          .setMessage("Unable to extract error message from response: " + e.getMessage())
```
anuraaga, line 181: In the context of my above comment, I think the response.getMessage() is fine for this and we can ignore e.getMessage() which is unlikely to be helpful for responses from off-the-shelf proxies.

exporters/otlp-http/trace/src/test/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporterTest.java
```diff
@@ -0,0 +1,246 @@
...
+      throw new IllegalStateException("Error building certificate.", e);
+    }
+  }
+
+  @RegisterExtension
+  static MockWebServerExtension server =
+      new MockWebServerExtension() {
+        @Override
+        protected void configureServer(ServerBuilder sb) {
+          sb.tlsSelfSigned(false);
```
anuraaga, line 73:
```suggestion
```

Shouldn't need this since you call `tls` below to "reset" the TLS

exporters/otlp-http/trace/src/test/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporterTest.java
```diff
@@ -0,0 +1,246 @@
...
+import okio.GzipSource;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+import org.slf4j.event.Level;
+import org.slf4j.event.LoggingEvent;
+
+class OtlpHttpSpanExporterTest {
+
+  private static final MediaType APPLICATION_PROTOBUF =
```
anuraaga, line 52: Can use the already existing `MediaType.PROTOBUF`

anuraaga (commented):
Thanks a lot, almost there

exporters/otlp-http/trace/src/test/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporterTest.java
jack-berg: No 😞 . The spec requires `application/x-protobuf` and `MediaType.PROTOBUF` is `application/protobuf`.

exporters/otlp-http/trace/src/main/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporter.java
jack-berg: Ok. I do think it will be helpful to have the status message be different based on whether the response body is empty, or present but failed to parse. Definitely agree that the status message is more likely to contain useful information than `e.getMessage()`.

exporters/otlp-http/trace/build.gradle.kts
jack-berg: Not applicable now that I incorporated the feedback not to publish a shadow jar.

exporters/otlp-http/trace/src/main/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporter.java
```diff
@@ -0,0 +1,224 @@
...
+            .addAllResourceSpans(SpanAdapter.toProtoResourceSpans(spans))
+            .build();
+
+    Request.Builder requestBuilder = new Request.Builder().url(endpoint);
+    if (headers != null) {
+      requestBuilder.headers(headers);
+    }
+    RequestBody requestBody =
+        RequestBody.create(
+            exportTraceServiceRequest.toByteArray(), MediaType.parse("application/x-protobuf"));
```
anuraaga, line 97: Can you extract a constant for this MediaType?

anuraaga (approved):
Thanks, just found one small point left. @jkwatson is on vacation but let's give him a chance to review before merging.

jkwatson:
Since this is a new module that isn't marked `alpha`, shouldn't there be apidiffs for it?

jack-berg:
@jkwatson actually could use a pointer on that. There's no apidiffs because I've commented out the `id("otel.publish-conventions")` in `build.gradle.kts`. And I've commented it out because I get the following error trying to run `:build` if its enabled:
```
Could not determine the dependencies of task ':exporters:otlp-http:trace:check'.
> Could not create task ':exporters:otlp-http:trace:jApiCmp'.
   > Could not resolve all files for configuration ':exporters:otlp-http:trace:detachedConfiguration1'.
      > Could not find io.opentelemetry:opentelemetry-exporter-otlp-http-trace:1.4.1.
        Searched in the following locations:
          - https://repo.maven.apache.org/maven2/io/opentelemetry/opentelemetry-exporter-otlp-http-trace/1.4.1/opentelemetry-exporter-otlp-http-trace-1.4.1.pom
``` 
Seems like a chicken and the egg issue: can't add publish conventions until there's an artifact out there that can be used as a comparison point for the apidiffs. 

I know the apidiff stuff is fairly new - is there a playbook on how to proceed adding a new module?

jkwatson:
> I know the apidiff stuff is fairly new - is there a playbook on how to proceed adding a new module?

I thought @anuraaga had fixed this issue so that it would generate a new file for a new module. 

Curious why you commented out publishing, though... shouldn't this be published?

jack-berg:
It should be published. I commented out publishing because the build fails with [this error](https://github.com/open-telemetry/opentelemetry-java/pull/3418#issuecomment-892131316) when publishing is enabled. 

Let me see if I can't dig into it a bit more and get to the bottom of why new modules are causing this issue.

jkwatson:
> It should be published. I commented out publishing because the build fails with [this error](https://github.com/open-telemetry/opentelemetry-java/pull/3418#issuecomment-892131316) when publishing is enabled.
> 
> Let me see if I can't dig into it a bit more and get to the bottom of why new modules are causing this issue.

hmm. interesting. If you can't sort this out, I can try to take a look (or perhaps @anuraaga will know the problem right away).

jack-berg:
Looking closer at the [otel.japicmp-conventions.gradle.kts](https://github.com/open-telemetry/opentelemetry-java/blob/main/buildSrc/src/main/kotlin/otel.japicmp-conventions.gradle.kts#L46) config, I don't think this is actually a solved problem:

- The `jApiCmp` task is added if the project does not have `otel.release` and does not start with `bom`. Because `otel.release` is reserved for alpha projects and this new project will not be alpha, it will get the `jApiCmp` task.
- If we have to have this task, then we need to configure it to compare the latest jar to nothing, since no previous version exists. I can't find a way to configure `jApiCmp` to do a "compare to nothing". The [oldClasspath](https://github.com/open-telemetry/opentelemetry-java/blob/main/buildSrc/src/main/kotlin/otel.japicmp-conventions.gradle.kts#L55) property is what specifies what your comparing the new jar against, and its required. 
- The `jApiCmp` code was initially added on May 6th, 2021 in PR #3183. If you look at the blame history of [settings.gradle.kts](https://github.com/open-telemetry/opentelemetry-java/blame/main/settings.gradle.kts), only one new project has been added since May 6th, `:sdk:metrics-testing`, and its exempt from the `jApiCmp` task because its marked as `otel.release=alpha`. 

One way we could solve this is by adding another property to new projects that haven't been published yet, such as `otel.initial-release`. Then configure `jApiCmp` task to skip projects with this flag, and remove the flag after the first publication.

jkwatson:
> One way we could solve this is by adding another property to new projects that haven't been published yet, such as `otel.initial-release`. Then configure `jApiCmp` task to skip projects with this flag, and remove the flag after the first publication.

Something like this seems totally reasonable to me. Or, we institute a policy where every new module has to go through at least one alpha release before it is made non-alpha, which would also solve the problem. What do you think about that, @anuraaga ?

anuraaga:
Do we need a new flag? I guess we can disable the task if oldClasspath is invalid, so that'd be pretty easy. If it's not obvious how to do it I can look into that. Let's merge this first though.

jkwatson (approved):
Thanks!

sumitm-iiit:
Can someone please specify this change is part of which open telemetry java agent? I have tried with Open Telemetry Java Agent 1.5.3 but it seems this change is not present.

Oberon00:
I think you should ask that in https://github.com/open-telemetry/opentelemetry-java-instrumentation. As displayed by GitHub, this change is included in version 1.5.0 of "core" opentelemetry-java.

sumitm-iiit:
@Oberon00 : Thank you for your quick response. Sure I will ask this question in htts://github.com/open-telemetry/opentelemetry-java-instrumentation..

Oberon00:
@sumitm-iiit BTW, under https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/README.md#otlp-exporter-both-span-and-metric-exporters there is `otel.experimental.exporter.otlp.metrics.protocol`, maybe that is what you are looking for?

sumitm-iiit:
I have tried that but it seems the agent itself does not packages OTLP over
HTTP exporter inside it.

PR #3419, opened 2021-07-21
Provide an option to not set the auto-configured SDK as the global.

jkwatson:


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
```diff
@@ -34,8 +35,11 @@ public static Resource getResource() {
   /**
    * Returns an {@link OpenTelemetrySdk} automatically initialized through recognized system
    * properties and environment variables.
+   *
+   * @param setResultAsGlobal Whether to automatically set the configured SDK as the {@link
```
anuraaga, line 39: Just wondering if it can ever hurt to set the global? Prevents classload or something?

anuraaga (approved):

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
jkwatson: It can hurt if (in my case) if the user isn't in complete control of the JVM that is being used. In my particular case, I'm trying to do some work in gradle, and this gets registered in the gradle daemon's JVM, and subsequent builds fail when they try to configure again.

PR #3423, opened 2021-07-24
Swap old Metrics API for currently specified Metrics API

jsuereth:
- Implement, as best as possible, the latest version of the [metrics API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md) specification, with feedback from Java SiG from the prototype.
- Wire the new API into all other SDK usages of the API.
- This PR *touches the SDK as little as possible*.   It does the *LEAST* amount of renaming, rewiring and reworking to get things working.   This is *only* meant to be an API swap, with a LOT of follow on PRs to come (based on the [prototype](https://github.com/open-telemetry/opentelemetry-java/pull/3333)).  This leaves a lot of wonky behavior such as:
  - The SDK references labels everywhere, including a `LabelsProcessor` for views.  This will be refactored to match the SDK specifcaiton in follow-up PR.
  - The SDK currently has "value observer" and "value recorder" instead of "gauge" and "histogram" instruments.   We keep those in the SDK to minimize the diff-set to view, but they will be cleaned up in a follow on PR.
  - The Histogram instrument is producing Summary metrics right now.  This will be fixed when the new SDK implementation lands.
  - The InstrumentDescriptor API has NOT BEEN touched, so instrument names are out of date.   This will also be fixed in a follow up PR.


api/metrics/src/main/java/io/opentelemetry/api/metrics/BoundDoubleCounter.java
```diff
@@ -5,20 +5,28 @@
...
-   * <p>The value added is associated with the current {@code Context}.
+   * <p>Note: This may use {@code Context.current()} to pull the context associated with this
+   * measurement.
    *
-   * @param increment the value to add.
+   * @param value The increment amount. MUST be non-negative.
    */
-  void add(double increment);
+  void add(double value);
+  /**
```
anuraaga, line 23: Empty line between methods. Seems to be a few spots in the PR

api/metrics/src/main/java/io/opentelemetry/api/metrics/BoundDoubleHistogram.java
```diff
@@ -0,0 +1,32 @@
...
+/** A hisgogram instrument that records {@code long} values with preassociated attributes. */
+@ThreadSafe
+public interface BoundDoubleHistogram {
+  /**
+   * Record a value with a pre-bound set of attributes.
+   *
+   * @param value The amount of the measurement.
+   * @param context The explicit context to associate with this measurement.
+   */
+  void record(double value, Context context);
```
anuraaga, line 20: Let's be consistent in the order of no-context, context

api/metrics/src/main/java/io/opentelemetry/api/metrics/Counter.java
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.api.metrics;
+
+/**
+ * A counter instrument.
+ *
+ * <p>This will have a specific subclass that recoords appropriate primitive values, e.g. {@link
```
anuraaga, line 11:
```suggestion
 * <p>Type-specific subclasses record appropriate primitive values, e.g. {@link
```

Present tense instead of future since it already does

api/metrics/src/main/java/io/opentelemetry/api/metrics/GlobalMeterProvider.java
```diff
@@ -5,81 +5,30 @@
...
-/**
- * IMPORTANT: This is a temporary class, and solution for the metrics package until it will be
- * marked as stable.
- */
-public final class GlobalMeterProvider {
-  private static final Object mutex = new Object();
-  private static final AtomicReference<MeterProvider> globalMeterProvider = new AtomicReference<>();
+/** This class is a temporary solution until metrics SDK is marked stable. */
+public class GlobalMeterProvider {
+  private static final AtomicReference<MeterProvider> globalMeterProvider =
```
anuraaga, line 13: Can this just be a normal `volatile` since we only set and get?

api/metrics/src/main/java/io/opentelemetry/api/metrics/BoundDoubleCounter.java
```diff
@@ -5,20 +5,28 @@
...
 import javax.annotation.concurrent.ThreadSafe;
 
-/** A {@code Bound Instrument} for a {@link DoubleCounter}. */
+/** A counter instrument that records {@code double} values with preassociated attributes. */
 @ThreadSafe
-public interface BoundDoubleCounter extends BoundSynchronousInstrument {
+public interface BoundDoubleCounter {
   /**
-   * Adds the given {@code increment} to the current value. The values cannot be negative.
+   * Reecord a value with prebound attributes.
```
anuraaga, line 15:
```suggestion
   * Records a value with prebound attributes.
```

For all the javadoc let's use present tense instead of command form

api/metrics/src/main/java/io/opentelemetry/api/metrics/LongCounterBuilder.java
```diff
@@ -5,14 +5,42 @@
...
-  @Override
+  /**
+   * Sets the description for this instrument.
+   *
+   * <p>Description stirngs should follw the instrument description rules:
+   * https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-description
+   */
+  LongCounterBuilder setDescription(String description);
+  /**
+   * Set the unit of measure for this instrument.
```
anuraaga, line 21: Careful with command form, let's use present tense throughout

api/metrics/src/main/java/io/opentelemetry/api/metrics/LongUpDownCounterBuilder.java
```diff
@@ -5,14 +5,41 @@
...
+import java.util.function.Consumer;
+
 /** Builder class for {@link LongUpDownCounter}. */
-public interface LongUpDownCounterBuilder extends SynchronousInstrumentBuilder {
-  @Override
+public interface LongUpDownCounterBuilder {
+  /**
+   * Sets the description for this instrument.
+   *
+   * <p>Description stirngs should follw the instrument description rules:
```
anuraaga, line 15:
```suggestion
   * <p>Description strings should follow the instrument description rules:
```

Don't think I'll catch all of these in review, would be good if you can run a spell check

api/metrics/src/main/java/io/opentelemetry/api/metrics/MeterProvider.java
```diff
@@ -5,43 +5,48 @@
...
-    return DefaultMeterProvider.getInstance();
-  }
-
-  /**
-   * Gets or creates a named meter instance.
+   * Gets or creates a named and versioned meter instance.
    *
    * @param instrumentationName The name of the instrumentation library, not the name of the
    *     instrument*ed* library.
-   * @return a tracer instance.
```
anuraaga, line 32: 😂

sdk/metrics-testing/src/main/java/io/opentelemetry/sdk/testing/assertj/metrics/MetricAssertions.java
```diff
@@ -33,11 +35,21 @@ public static DoubleHistogramAssert assertThat(DoubleHistogramData metric) {
...
+    return new DoubleSummaryDataAssert(metric);
+  }
+
   /** Returns an assertion for {@link DoubleHistogramPointData}. */
   public static DoubleHistogramPointDataAssert assertThat(DoubleHistogramPointData point) {
     return new DoubleHistogramPointDataAssert(point);
   }
 
+  /** Returns an assertion for {@link DoubleSummaryPointData}. */
+  public static DoubleSummaryPointDataAssert assertThat(DoubleSummaryPointData point) {
```
anuraaga, line 49: 🎉

anuraaga (commented):
Just nits, all-in-all looks good

api/metrics/src/main/java/io/opentelemetry/api/metrics/GlobalMeterProvider.java
jsuereth: The only difference between `volatile` and what I have is the *reader* pays the write-cache-flush+reload cost, while the writer can go on their merry way without a full sync.   Probably WAY to high on the micro-optimizatioon, just standard practice for concurrent programming on the JVM for when I used to do it often.

api/metrics/src/main/java/io/opentelemetry/api/metrics/BoundDoubleCounter.java
jsuereth: should be fixed.

api/metrics/src/main/java/io/opentelemetry/api/metrics/BoundDoubleHistogram.java
jsuereth: K, went with least-to-most arguments, assuming context-explicit usage to be the least common.

api/metrics/src/main/java/io/opentelemetry/api/metrics/BoundDoubleCounter.java
jsuereth: Should be updated.

api/metrics/src/main/java/io/opentelemetry/api/metrics/LongCounterBuilder.java
jsuereth: Understood, should be fixed.

api/metrics/src/main/java/io/opentelemetry/api/metrics/LongUpDownCounterBuilder.java
jsuereth: Didn't realize I was missing one!  Added to visual studio, re-ran through all the main API

api/metrics/src/main/java/io/opentelemetry/api/metrics/Counter.java
```diff
@@ -0,0 +1,14 @@
...
+
+package io.opentelemetry.api.metrics;
+
+/**
+ * A counter instrument.
+ *
+ * <p>This will have a specific subclass that records appropriate primitive values, e.g. {@link
+ * LongCounter} will record {@code long} values.
+ */
+public interface Counter extends SynchronousInstrument {}
```
bogdandrutu, line 14: Is this necessary? Consider to remove?

api/metrics/src/main/java/io/opentelemetry/api/metrics/DoubleGaugeBuilder.java
```diff
@@ -0,0 +1,39 @@
...
+  /**
+   * Sets the unit of measure for this instrument.
+   *
+   * <p>Unit strings should follow the instrument unit rules:
+   * https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-unit
+   */
+  DoubleGaugeBuilder setUnit(String unit);
+
+  /** Sets the gauge for recording {@code long} values. */
+  LongGaugeBuilder ofLongs();
```
bogdandrutu, line 29: I don't think this is a very "java" pattern.

api/metrics/src/main/java/io/opentelemetry/api/metrics/DoubleGaugeBuilder.java
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Builds this asynchronous instrument with the given callback.
+   *
+   * <p>The callback will only be called when the {@link Meter} is being observed.
+   *
+   * @param callback A state-capturing callback used to observe values on-demand.
+   */
+  void buildWithCallback(Consumer<ObservableDoubleMeasurement> callback);
```
bogdandrutu, line 38: Personally I don't like the mix of sync/async instruments in the same builder. May lead to confusions. I understand that the downside is some duplicate code, but I don't think is a big deal.

bogdandrutu (commented):
Some high-level comments.

api/metrics/src/main/java/io/opentelemetry/api/metrics/GlobalMeterProvider.java
anuraaga: Yeah - since there isn't really performance impact (if there is it's on the get side, not set) we should err on simpler code. In particular, we make it harder for future readers who will need to lookup javadoc to understand `lazySet`, a very uncommonly used method compared to `volatile` which is well understood. This code is going to be deleted anyways, which is probably more the reason to just keep things as simple as possible right now.

api/metrics/src/main/java/io/opentelemetry/api/metrics/DoubleGaugeBuilder.java
anuraaga: FWIU of the discussions, the user feedback is that it's complicated to understand the difference between a sync and async instrument, they record the same thing just one has a callback for retrieving values instead of direct add. There is no change in the actual data emitted by the SDK. So mixing means you first pick 1) what you want to record and then 2) how you want to record it rather than combining those into one thought process, which to me at least does seem confusing.

anuraaga: That being said, perhaps it gets simplified if gauge was the only asynchronous instrument. I'm sure that's been discussed though I haven't been following the spec discussions. But intuitively indeed I'm not too sure when I would use a callback with an updowncounter for example.

api/metrics/src/main/java/io/opentelemetry/api/metrics/DoubleGaugeBuilder.java
jsuereth: To be fair, I am a Scala guy, so it's not surprising.  However, a few Java-isms I based this on (again as a Scala outsider when investigating boxing + specialization):

- https://docs.oracle.com/javase/8/docs/api/java/util/PrimitiveIterator.OfLong.html
- https://beam.apache.org/releases/javadoc/2.0.0/org/apache/beam/sdk/transforms/Sum.html#ofLongs--

TL;DR: it was the JDK's chosen name when you specialize an interface for primitive inputs, so I chose it.  If you have an alternative suggestion, happy to hear it.

api/metrics/src/main/java/io/opentelemetry/api/metrics/DoubleGaugeBuilder.java
jsuereth: The switch to merge these together was an ask in the Java SiG meeting.   I agree with @anuraaga that picking the type of metric is likely to be primary concern.    Would be good if you could join the Java SiG (when @jkwatson is back from vacation) to discuss this specific choice.  I disagree on confusion as I think the opposite is true: Splitting these leads to greater confusion.   If this was purely about duplicated code, I agree it's not a big deal.   This design was chosen to simplify and reduce user confusion.

@anuraaga I think, likely, the Java instrumentation folk may need to be using UpDownCounter more often.    E.g. JVM memory usage SHOULD be an up-down-counter callback.

bogdandrutu: @jsuereth Please ping me when to join, would be happy to do it.

api/metrics/src/main/java/io/opentelemetry/api/metrics/GlobalMeterProvider.java
jsuereth: Sure, but so we're clear, the performance improvement is on the *set* side, not the get.  The get will have a memory barrier that flushes cache, `lazySet` vs. `set` (or volatile write) is cache-flush vs. not cache-flush, so this only impacts speed of writing the global, allowing (e.g.) the agent to consider processing its startup rather than flushing right away.

`lazySet` is pretty common thing for high performance concurrency.  e.g. look at JCTools: https://github.com/JCTools/JCTools/search?q=lazySet
I do *not* think we should shy away from it.  Java's `volatile` feature is pretty poor for modern architecture.    I'm ok migrating this specific code to `volatile`, but want to make sure we aren't preferring `volatile`, generally.

api/metrics/src/main/java/io/opentelemetry/api/metrics/DoubleGaugeBuilder.java
jsuereth: Thursday at 9am PDT, next week should have @jkwatson back, I'll be unable to make it but I'm ok accepting whatever you hash out there.  Alternatively, may be worth joining the 4pm PDT meeting (Thursday august 5th) w/ @jkwatson + @anuraaga for high bandwidth discussion.

api/metrics/src/main/java/io/opentelemetry/api/metrics/Counter.java
jsuereth: No, and removed.

api/metrics/src/main/java/io/opentelemetry/api/metrics/GlobalMeterProvider.java
jsuereth: This is updated, but leaving unresolved because I'm concerned about the`lazySet` here.  I think usage of `lazySet` is critical to most efficient concurrent solutions in Java and I don't want us shying away from it when it's warranted.

anuraaga: Yeah - any app sets a global provider (or OpenTelemetry) once at the beginning of the app initialization so the performance of set doesn't matter at all here. We might also trick something into thinking it does matter if we use `lazySet` here.

Where it does have an effect we would use it and probably add comments on the motivation - in many cases we probably would just end up using it via some other library like jctools though.

anuraaga (approved):

api/metrics/src/main/java/io/opentelemetry/api/metrics/GlobalMeterProvider.java
jsuereth: Sounds good.

api/metrics/src/main/java/io/opentelemetry/api/metrics/LongUpDownCounter.java
```diff
@@ -5,62 +5,46 @@
...
- * }</pre>
- */
+/** An up-down-counter instrument that records {@code long} values. */
 @ThreadSafe
-public interface LongUpDownCounter extends SynchronousInstrument<BoundLongUpDownCounter> {
-
+public interface LongUpDownCounter extends UpDownCounter {
   /**
-   * Adds the given {@code increment} to the current value.
+   * Records a value.
```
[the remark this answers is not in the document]
jsuereth: I went for consistency across instruments here.   There's an 'oddity' here where users can actually use aggregators to change what happens when recording values.   E.g. *by default* this would add an increment to current value, but you can use a `View` to convert this to a histogram or other behavior, which is why I went with `record a value` consistently across instruments.

api/metrics/src/main/java/io/opentelemetry/api/metrics/DoubleGaugeBuilder.java
hdost: I know this is related to DateTimes, but provides context for the conversation: https://docs.oracle.com/javase/tutorial/datetime/overview/naming.html

api/metrics/src/main/java/io/opentelemetry/api/metrics/DoubleGaugeBuilder.java
jkwatson: My opinion is that this is a good limitation of the API surface, but I can also see the point that it could be confusing to users who are new to the API. Hard call. I *personally* like this API better than having separate builders, but I'm only one user who is too deep in the weeds to know if this would be confusing. 

@kenfinnigan Any thoughts?

api/metrics/src/main/java/io/opentelemetry/api/metrics/BoundDoubleHistogram.java
```diff
@@ -0,0 +1,33 @@
...
+
+  /**
+   * Records a value with a pre-bound set of attributes.
+   *
+   * @param value The amount of the measurement.
+   * @param context The explicit context to associate with this measurement.
+   */
+  void record(double value, Context context);
+
+  void unbind();
```
jkwatson, line 32: javadoc needed here.

api/metrics/src/main/java/io/opentelemetry/api/metrics/BoundDoubleUpDownCounter.java
```diff
@@ -5,20 +5,29 @@
...
+  /**
+   * Records a value with a pre-bound attributes.
+   *
+   * @param value The increment amount. May be positive, negative or zero.
+   * @param context The explicit context to associate with this measurement.
+   */
+  void add(double value, Context context);
 
-  @Override
   void unbind();
```
jkwatson, line 32: I think all the `unbind()`s are going to need javadoc without the super interface.

api/metrics/src/main/java/io/opentelemetry/api/metrics/DoubleGaugeBuilder.java
jkwatson: wait... this particular case isn't a mix of sync/async is it? It's just async, since the gauge is only async, correct?

api/metrics/src/main/java/io/opentelemetry/api/metrics/DoubleUpDownCounterBuilder.java
```diff
@@ -5,14 +5,42 @@
...
   DoubleUpDownCounter build();
+
+  /**
+   * Builds this asynchronous insturment with the given callback.
+   *
+   * <p>The callback will only be called when the {@link Meter} is being observed.
+   *
+   * @param callback A state-capturing callback used to observe values on-demand.
+   */
+  void buildWithCallback(Consumer<ObservableDoubleMeasurement> callback);
```
jkwatson, line 45: here's the case with both sync and async in the same builder. :)

api/metrics/src/main/java/io/opentelemetry/api/metrics/DoubleUpDownCounterBuilder.java
```diff
@@ -5,14 +5,42 @@
...
+
+  /**
+   * Builds and returns a {@code DoubleUpDownCounter} with the desired options.
+   *
+   * @return a {@code DoubleUpDownCounter} with the desired options.
+   */
   DoubleUpDownCounter build();
+
+  /**
+   * Builds this asynchronous insturment with the given callback.
```
jkwatson, line 39:
```suggestion
   * Builds this asynchronous instrument with the given callback.
```

api/metrics/src/main/java/io/opentelemetry/api/metrics/GlobalMeterProvider.java
```diff
@@ -5,29 +5,17 @@
...
 
-/**
- * IMPORTANT: This is a temporary class, and solution for the metrics package until it will be
- * marked as stable.
- */
-public final class GlobalMeterProvider {
-  private static final Object mutex = new Object();
-  private static final AtomicReference<MeterProvider> globalMeterProvider = new AtomicReference<>();
+/** This class is a temporary solution until metrics SDK is marked stable. */
+public class GlobalMeterProvider {
```
jkwatson, line 11: this change seems unrelated to the rest of the PR. Is it needed for some reason?

api/metrics/src/main/java/io/opentelemetry/api/metrics/ObservableMeasurement.java
```diff
@@ -0,0 +1,13 @@
...
+ */
+
+package io.opentelemetry.api.metrics;
+
+/**
+ * A mechanism for observing measurments.
+ *
+ * <p>see {@link ObservableDoubleMeasurement} or {@link ObservableLongMeasurement}.
+ */
+public interface ObservableMeasurement {}
```
jkwatson, line 13: with no methods and no types does this have enough value to keep?

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporter.java
```diff
@@ -35,13 +36,15 @@
...
-
+  private static final AttributeKey<String> EXPORTER_KEY = AttributeKey.stringKey("exporter");
+  private static final AttributeKey<String> SUCCESS_KEY = AttributeKey.stringKey("success");
   private static final String EXPORTER_NAME = OtlpGrpcSpanExporter.class.getSimpleName();
-  private static final Labels EXPORTER_NAME_LABELS = Labels.of("exporter", EXPORTER_NAME);
-  private static final Labels EXPORT_SUCCESS_LABELS =
-      Labels.of("exporter", EXPORTER_NAME, "success", "true");
-  private static final Labels EXPORT_FAILURE_LABELS =
-      Labels.of("exporter", EXPORTER_NAME, "success", "false");
+  private static final Attributes EXPORTER_NAME_Attributes =
```
jkwatson, line 42: should probably CAPITALIZE the ATTRIBUTES here as well. :)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractInstrumentBuilder.java
```diff
@@ -0,0 +1,109 @@
...
+    return getThis();
+  }
+
+  public BuilderT setDescription(String description) {
+    this.description = description;
+    return getThis();
+  }
+
+  @FunctionalInterface
+  protected static interface SwapBuilder<T> {
```
jkwatson, line 50: this interface is hanging out in the middle of the class. I think our standards say to put it at the bottom.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/InstrumentSelector.java
```diff
@@ -12,8 +12,8 @@
 import javax.annotation.concurrent.Immutable;
 
 /**
- * Provides means for selecting one ore more {@link io.opentelemetry.api.metrics.Instrument}s. Used
- * for configuring aggregations for the specified instruments.
+ * Provides means for selecting one ore more instruments. Used for configuring aggregations for the
```
jkwatson, line 15:
```suggestion
 * Provides means for selecting one or more instruments. Used for configuring aggregations for the
```

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/SdkMeterRegistryTest.java
```diff
@@ -135,10 +133,6 @@ void suppliesDefaultMeterForNullName() {
     assertThat(meter.getInstrumentationLibraryInfo().getName())
         .isEqualTo(SdkMeterProvider.DEFAULT_METER_NAME);
 
-    meter = (SdkMeter) meterProvider.get(null, null);
-    assertThat(meter.getInstrumentationLibraryInfo().getName())
-        .isEqualTo(SdkMeterProvider.DEFAULT_METER_NAME);
```
jkwatson, line 140: why delete these tests?

jkwatson:
Just a few minor comments. Looks good! Thanks!

api/metrics/src/main/java/io/opentelemetry/api/metrics/GlobalMeterProvider.java
jsuereth: The previous implementation looked like a mistake / copy-paste with missing reason for existence.  I can remove this if it's contentious.

api/metrics/src/main/java/io/opentelemetry/api/metrics/ObservableMeasurement.java
jsuereth: It helped in the SDK implementation.   I can remove it, but you don't see the code which uses this yet.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/SdkMeterRegistryTest.java
jsuereth: the two-argument `get` method is gone.

api/metrics/src/main/java/io/opentelemetry/api/metrics/GlobalMeterProvider.java
jkwatson: nah, I don't think it's a big deal...just was puzzled when I saw it in amongst the API changes.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/SdkMeterRegistryTest.java
jkwatson: well ok then!

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractInstrumentBuilder.java
jsuereth: moved.  Debated adding ascii-art-lawn-chair to denote its chill attitude towards live (and high coupling to the constructors required for builders).

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporter.java
jsuereth: Ah, VS Code let me down here.  The "ignore capitalization" vs. "use capitalization" in find replace vs. "change symbol" hot keys always get me.  Fixed

api/metrics/src/main/java/io/opentelemetry/api/metrics/BoundDoubleUpDownCounter.java
jsuereth: Added, copying previous javadoc.   Kinda unhappy with both bind + unbind's documentation on why/how they exists.

Thinking about discussing how (theoretically) if you only use `BOUND` instruments you should never see memory allocations in the hot path of your applications as, AFAIK, this is a guarantee from the metrics SDK.   However, this is just the API so that guarantee isn't assured, only possible.    WDYT?

api/metrics/src/main/java/io/opentelemetry/api/metrics/BoundDoubleHistogram.java
jsuereth: done.

api/metrics/src/main/java/io/opentelemetry/api/metrics/BoundDoubleUpDownCounter.java
jkwatson: yeah, we can't guarantee this in the API javadoc, unfortunately.

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporter.java
jkwatson: maybe didn't push the fix?

jkwatson (approved):
Approved, pending API discussion and a tiny capitalization nit.

anuraaga:
@jkwatson I think it's ok to merge first and follow up on the API discussion given this is an alpha artifact and the PR is huge. What do you think?

api/metrics/src/main/java/io/opentelemetry/api/metrics/GlobalMeterProvider.java
```diff
@@ -36,50 +24,7 @@ public static MeterProvider get() {
...
-   *
-   * @param instrumentationName The name of the instrumentation library, not the name of the
-   *     instrument*ed* library.
-   * @return a MeterBuilder instance.
-   * @since 1.4.0
-   */
-  public static MeterBuilder meterBuilder(String instrumentationName) {
-    return get().meterBuilder(instrumentationName);
+  public static void set(MeterProvider provider) {
+    globalMeterProvider = (provider == null) ? NoopMeterProvider.getInstance() : provider;
```
jack-berg, line 28: Equating `null` = `NoopMeterProvider.getInstance()` is strange IMO. 

I think this should return or throw if null is passed. Throwing is inconsistent with current behavior, so I say return.

jkwatson:
> @jkwatson I think it's ok to merge first and follow up on the API discussion given this is an alpha artifact and the PR is huge. What do you think?

This actually needs a rebase, since there have been additional API usages added since it was last rebased against main.

api/metrics/src/main/java/io/opentelemetry/api/metrics/LongUpDownCounterBuilder.java
```diff
@@ -5,14 +5,42 @@
...
+
+  /**
+   * Builds and returns a {@code LongUpDownCounter} with the desired options.
+   *
+   * @return a {@code LongUpDownCounter} with the desired options.
+   */
   LongUpDownCounter build();
+
+  /**
+   * Builds this asynchronous insturment with the given callback.
```
jack-berg, line 39:
```suggestion
   * Builds this asynchronous instrument with the given callback.
```

api/metrics/src/main/java/io/opentelemetry/api/metrics/Meter.java
```diff
@@ -8,174 +8,53 @@
...
- *   <li>Record pre-defined aggregation data (or already aggregated data). This should be used to
- *       report cpu/memory usage, or simple metrics like "queue_length".
- * </ul>
- *
- * <p>TODO: Update comment.
+ * <p>A Meter is generally associated with an instrumentation library, e.g. "I monitor apache
+ * httpclient".
  */
 @ThreadSafe
 public interface Meter {
```
jack-berg, line 20: The async instruments are accessible via the `buildWithCallback` methods from the respective builders, rather than from the `Meter`.

Is this allowed according to the spec?

api/metrics/src/main/java/io/opentelemetry/api/metrics/LongUpDownCounterBuilder.java
```diff
@@ -5,14 +5,42 @@
...
+   *
+   * <p>The callback will only be called when the {@link Meter} is being observed.
+   *
+   * @param callback A state-capturing callback used to observe values on-demand.
+   */
+  void buildWithCallback(Consumer<ObservableLongMeasurement> callback);
```
jack-berg, line 45: I'd like to see "observable" somehow incorporated into the naming convention of methods that access async instruments. Its recommended in the spec and I think it better clarifies the async distinction than "callback". 

Maybe `buildObservable`?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
```diff
@@ -51,13 +51,6 @@ public Meter get(String instrumentationName) {
     return meterBuilder(instrumentationName).build();
   }
 
-  @Override
-  public Meter get(String instrumentationName, String instrumentationVersion) {
-    return meterBuilder(instrumentationName)
-        .setInstrumentationVersion(instrumentationVersion)
-        .build();
-  }
```
jack-berg, line 59: It looks like you dropped this because the default method in `MeterProvider` covers it. By that logic, I think you can drop `public Meter get(String instrumentationName)` as well.

api/metrics/src/main/java/io/opentelemetry/api/metrics/Meter.java
jkwatson: I would say so. You still get the builder via the `Meter`. The spec isn't supposed to precisely define the language-specific APIs, but the capabilities that must be provided.

api/metrics/src/main/java/io/opentelemetry/api/metrics/LongUpDownCounterBuilder.java
jsuereth: @jkwatson @anuraaga for comment.   This was a specific deviation to fit the java "builder" convention, so I defer to Java SiG here.

api/metrics/src/main/java/io/opentelemetry/api/metrics/Meter.java
jsuereth: This was deemed a Java-specific deviation to abide by the common builder pattern in Java.   I like the simplicity it acheived, but it is something we should discuss and can migrate to having separate builders by-async type instead of a builder per-metric-type and then having the async builder + sync builder coexist.

api/metrics/src/main/java/io/opentelemetry/api/metrics/LongUpDownCounterBuilder.java
jsuereth: Fixed.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
jsuereth: Fixed, good call out!

api/metrics/src/main/java/io/opentelemetry/api/metrics/LongUpDownCounterBuilder.java
jkwatson: The spec calls this an "observable" I think? In that case, probably migrating to that verbiage would be helpful. I don't want to hold up this PR for that change, though.

jkwatson:
I'm going to merge this, so we can build from it. Thanks, @jsuereth !

PR #3429, opened 2021-07-27
Add Iterable versions of hasTraces / Spans

anuraaga:
It's reasonable to need to set up assertions programatically, for example when testing a method that executes dozens of requests to verify concurrency. This way a List type of thing doesn't need to be converted to an array and is consistent with the other vararg methods we have in our APIs


mateuszrzeszutek (approved):
👍 
Thanks!

carlosalberto (approved):

jkwatson (approved):

PR #3432, opened 2021-07-28
Update nullaway and add a Contract for StringUtils.

anuraaga:
Also finishes fixing nullness in api:all


api/all/src/main/java/io/opentelemetry/api/internal/StringUtils.java
```diff
@@ -27,6 +27,7 @@
    * @param string a string reference to check
    * @return {@code true} if the string is null or is the empty string
    */
+  @Contract("null -> true")
```
jkwatson, line 30: Is the intention that we only use this on `internal` code? I'm a little nervous about owning a public `Contract` annotation in this project.

anuraaga: Nah we'd use it on any code not just this package - that's why it's in the `internal` folder, that it's only for us.

jkwatson: But if we put it into our own public APIs, then we expose it to end-users, and who knows what they're going to do with it that we might break.

anuraaga: Ah I see what you're saying. In that case I guess I'll say the intention is for internal code for now - more accurately I want to use it in ways to get nullaway checking our repo, for now this method is the only one I've found that needs it not sure what else will crop up later.

jkwatson: I wonder if japicmp will pick up public usages of annotations. Can you verify that it does?

anuraaga: It seems to detect if the `RetentionPolicy` is `RUNTIME` but not the current `CLASS`. There isn't really any use for it to be at runtime for us, but given we'll use it so infrequently I think the cost isn't a big deal either. Should I switch it to RUNTIME?

jkwatson: Oh. If it's not RUNTIME we don't expose it to the resulting class files, right? So that should be ok to use even in non-internal classes, maybe?

anuraaga: CLASS means only in the byte code at compile time (mostly annotation processor) but not when actually running the app on JVM. RUNTIME means also loaded during the running of the app, which has a tiny bit of memory usage. Both sort of expose the annotation to users, though CLASS is naturally much harder to actually use. I think either is fine here.

jkwatson: could we use SOURCE here and thus not have them get into the bytecode at all? I don't know enough about how nullaway works to know if that would be enough.

anuraaga: Oh duh - think that was my intention anyways but forgot while copying over from Jetbrains. SOURCE works and shouldn't have any caveats related to leaking out.

jkwatson (approved):

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/Parser.java
```diff
@@ -82,7 +82,11 @@ void parseInto(BaggageBuilder baggageBuilder) {
                 break;
               case KEY: // none
             }
-            baggageBuilder.put(key.getValue(), value.getValue(), BaggageEntryMetadata.create(meta));
+            String parsedKey = key.getValue();
```
Oberon00, line 85: That's quite a bit of code duplication, can't this be extracted to a common?

api/all/src/main/java/io/opentelemetry/api/internal/Contract.java
```diff
@@ -0,0 +1,100 @@
...
+ * exception if {@code true} is passed to it<br>
+ * {@code @Contract("_ -> this")} - the method always returns its qualifier (e.g. {@link
+ * StringBuilder#append(String)}). {@code @Contract("null -> fail; _ -> param1")} - the method
+ * throws an exception if the first argument is null, otherwise it returns the first argument (e.g.
+ * {@code Objects.requireNonNull}).<br>
+ * {@code @Contract("!null, _ -> param1; null, !null -> param2; null, null -> fail")} - the method
+ * returns the first non-null argument, or throws an exception if both arguments are null (e.g.
+ * {@code Objects.requireNonNullElse} in Java 9).<br>
+ *
+ * <p>This annotation is the same provided with Jetbrains annotations and used by Nullaway for
```
kubawach, line 88: Nice!

kubawach (approved):
LGTM :)

PR #3433, opened 2021-07-28
Add AttributesAssert.containsKey

anuraaga:
While it's more common to check both key and value, it may be appropriate to only check key. In particular, this is extremely important to add to `hasEntrySatisfying` (as in this PR) since otherwise in the case of a missing key, it fallsback to a null check with an unhelpful assertion message `Expecting actual to not be null`


mateuszrzeszutek (approved):

carlosalberto (approved):

jsuereth (approved):

iNikem (approved):

jkwatson (approved):
👍🏽

PR #3439, opened 2021-08-03
Add OTLP signal specific certificate autoconfiguration

### LINKED ISSUE #3428: Support OTLP signal specific certificate configuration
N/A

Currently, you can only specify `OTEL_EXPORTER_OTLP_CERTIFICATE `. It sets the certificates OTLP/gRPC traces, and doesn't do anything for OTLP/gRPC metrics. 

`OTEL_EXPORTER_OTLP_CERTIFICATE` should set certificates for metrics, and  `OTEL_EXPORTER_OTLP_TRACES_CERTIFICATE` and `OTEL_EXPORTER_OTLP_METRICS_CERTIFICATE`  should be supported as defined by the [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.5.0/specification/protocol/exporter.md).

N/A

**Additional context**
Discussed that this would be addressed in future [here](https://github.com/open-telemetry/opentelemetry-java/pull/2569#discussion_r563455694).

jack-berg:
Resolves #3428

A prerequisite of adding metric specific certificate autoconfiguration is that `OtlpGrpcMetricExporter` has TLS support added. This PR addresses that as well.


exporters/otlp/metrics/src/main/java/io/opentelemetry/exporter/otlp/metrics/OtlpGrpcMetricExporterBuilder.java
```diff
@@ -7,6 +7,7 @@
 
 import static io.grpc.Metadata.ASCII_STRING_MARSHALLER;
 import static io.opentelemetry.api.internal.Utils.checkArgument;
+import static io.opentelemetry.exporter.otlp.internal.SslUtil.setTrustedCertificatesPem;
```
jkwatson, line 10: I think I'd prefer to not have this method statically imported. I'd rather see the explicit call to `SslUtil.set...` for clarity at the callsite.

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterBuilder.java
```diff
@@ -7,31 +7,19 @@
 
 import static io.grpc.Metadata.ASCII_STRING_MARSHALLER;
 import static io.opentelemetry.api.internal.Utils.checkArgument;
+import static io.opentelemetry.exporter.otlp.internal.SslUtil.setTrustedCertificatesPem;
```
jkwatson, line 10: ditto on not statically importing this method

jkwatson (approved):
just a couple tiny nits. looks good to me otherwise!

anuraaga (approved):

anuraaga:
Thanks @jack-berg!

PR #3440, opened 2021-08-04
Add http/protobuf exporter for metrics

### LINKED ISSUE #3395: Add support for the http/protobuf export for OTLP
Recent spec changes have suggested that SDKs SHOULD support http/protobuf as well as grpc for OTLP exporters.

jack-berg:
Closes #3395.


exporters/otlp-http/metrics/src/main/java/io/opentelemetry/exporter/otlp/http/metric/OtlpHttpMetricExporter.java
```diff
@@ -0,0 +1,230 @@
...
+      String endpoint,
+      @Nullable Headers headers,
+      boolean isCompressionEnabled) {
+    Meter meter = GlobalMeterProvider.getMeter("io.opentelemetry.exporters.otlp-http");
+    this.metricsSeen =
+        meter.longCounterBuilder("metricsSeenByExporter").build().bind(EXPORTER_NAME_LABELS);
+    LongCounter metricsExportedCounter =
+        meter.longCounterBuilder("metricsExportedByExporter").build();
+    this.metricsExportedSuccess = metricsExportedCounter.bind(EXPORT_SUCCESS_LABELS);
+    this.metricsExportedFailure = metricsExportedCounter.bind(EXPORT_FAILURE_LABELS);
```
jkwatson, line 76: we don't record metrics for the grpc metric exporter, so I wouldn't do it here. I think it's redundant since you'd see the metrics in wherever you were exporting the metrics to. :)

jack-berg: Yeah I noticed that about the grpc metric exporter. Figured it was an "add later" omission like TLS support.

On one hand, with these types of metrics you can capture information about failed exports in a way that can't be observed by the server you're exporting to. Although, if the failed exports aren't intermittent, then the metrics are unlikely to make it to a sink where you can observe them.

jkwatson: If we want to add metrics to our metrics export, let's do it all at one time, rather than just here. Sound ok?

exporters/otlp-http/metrics/src/main/java/io/opentelemetry/exporter/otlp/http/metric/OtlpHttpMetricExporter.java
```diff
@@ -0,0 +1,199 @@
...
+
+  private static final MediaType PROTOBUF_MEDIA_TYPE = MediaType.parse("application/x-protobuf");
+
+  private final ThrottlingLogger logger =
+      new ThrottlingLogger(Logger.getLogger(OtlpHttpMetricExporter.class.getName()));
+
+  private final OkHttpClient client;
+  private final String endpoint;
+  @Nullable private final Headers headers;
+  private final boolean isCompressionEnabled;
```
jkwatson, line 47: how about just calling this `compressionEnabled` without the `is`?

exporters/otlp-http/metrics/src/main/java/io/opentelemetry/exporter/otlp/http/metric/OtlpHttpMetricExporterBuilder.java
```diff
@@ -0,0 +1,161 @@
...
+
+/** Builder utility for {@link OtlpHttpMetricExporter}. */
+public final class OtlpHttpMetricExporterBuilder {
+
+  private static final long DEFAULT_TIMEOUT_SECS = 10;
+  private static final String DEFAULT_ENDPOINT = "http://localhost:4317/v1/metrics";
+
+  private long timeoutNanos = TimeUnit.SECONDS.toNanos(DEFAULT_TIMEOUT_SECS);
+  private String endpoint = DEFAULT_ENDPOINT;
+  private boolean isCompressionEnabled = false;
```
jkwatson, line 33: same here. I'd prefer just `compressionEnabled` without the `is`

exporters/otlp-http/metrics/src/main/java/io/opentelemetry/exporter/otlp/http/metric/OtlpHttpMetricExporter.java
jack-berg: Yeah as long as there's no convention, I prefer that. Will do that to the equivalent fields over in `OtlpHttpSpanExporter` as well.

anuraaga (approved):

exporters/otlp-http/trace/src/main/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporter.java
```diff
@@ -112,11 +116,11 @@ public CompletableResultCode export(Collection<SpanData> spans) {
               @Override
               public void onFailure(Call call, IOException e) {
                 spansExportedFailure.add(spans.size());
-                result.fail();
                 logger.log(
                     Level.SEVERE,
                     "Failed to export spans. The request could not be executed. Full error message: "
                         + e.getMessage());
+                result.fail();
```
anuraaga, line 123: I guess this is to try to solve this flake "org.opentest4j.AssertionFailedError: Contain the string <Failed to export spans. Server responded with HTTP status code 500. Error message: Server error!> ==> None of the 0 captured log events matched the filter predicate" but seems to still be happening. Maybe need to just use awaitility for that assertion due to thread issues.

jack-berg: Nice! Thanks for the recommendation. I've been pulling my hair out over that - glad to see there's a pattern for dealing with log threading issues in tests.

jack-berg: Didn't fix my problem 😞 

This is pretty mysterious. Have tried a bunch of stuff to try to recreate it locally with no luck.

jkwatson (approved):
looks good! Thanks!

jack-berg:
Test flake that's been plaguing these builds has occurred on [main](https://github.com/open-telemetry/opentelemetry-java/runs/3256512354) now, so its definitely not isolated to code introduced here.

Not sure exactly how to fix it. Asserting based on logs is new to me and I don't have good intuition as to when its reliable / not reliable.

jkwatson:
to add to the fun, @jack-berg , this needs a rebase after we merged the API changes from @jsuereth

exporters/otlp-http/trace/src/main/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporter.java
anuraaga: @jack-berg Took a look at the code, while it doesn't explain that I can't repro a failure locally (I'd expect it to be more likely locally thanks to the CPU speed) but could it be because of the ThrottlingLogger? It allows 5 log messages in, so we wouldn't expect a problem but it's definitely a component that could be causing a log message to not get pushed out.

anuraaga: FWIW, I feel I've only seen this failure on our Mac OS Java 17 CI test. That doesn't help much, I tried Java 17 locally on a Mac to no problem, but recording for the record.

jack-berg:
Yeah, will rebase after PR #3446 is merged. `io.opentelemetry.sdk.extension.trace.jaeger.sampler.JaegerRemoteSampler` is spamming something like 10k+ logs / second, which I'm betting is the cause of the build reliability woes.

jkwatson:
> Yeah, will rebase after PR #3446 is merged. `io.opentelemetry.sdk.extension.trace.jaeger.sampler.JaegerRemoteSampler` is spamming something like 10k+ logs / second, which I'm betting is the cause of the build reliability woes.

done!

jkwatson:
merging. only one of the many builds had a flake

Issue #3326, opened 2021-06-17
Allow using custom configuration sources in sdk-autoconfigure


mateuszrzeszutek:
**Is your feature request related to a problem? Please describe.**
Currently the sdk-autoconfigure module only loads configuration from two sources: system properties and environment variable. Aside from those two, the javaagent also allows using the `PropertySource` SPI and a config file (`otel.javaagent.configuration-file`) to configure agent and SDK.

The ask here is to bridge that gap between the SDK and the javaagent and allow passing configuration from custom sources when using the autoconfigure module. Currently to solve that problem we use a [hack](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java#L38) and copy all sdk-related properties to system properties so that SDK picks them up. 

**Describe the solution you'd like**
I don't have any preferred solution, any of the following would do:
* enable passing a `Map<String, String>` that contains full configuration to `OpenTelemetrySdkAutoConfiguration#initialize()`;
* enable passing a `ConfigProperties` instance (assuming it's an interface after https://github.com/open-telemetry/opentelemetry-java/pull/3304 gets merged) to the `initialize()` method; the javaagent `Config` could simply implement that interface;
* porting the SPI class from the javaagent to sdk-autoconfigure;
* any other that I did not think of.

**Additional context**
The javaagent SPI: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/spi/config/PropertySource.java

jkwatson:
The 2nd option (add a method to `initialize()` that takes ConfigProperties as a parameter seems pretty reasonable to me.

trask:
@jkwatson @anuraaga what are your latest thoughts on this? would you like send a PR that converts `ConfigProperties` to an interface and adds an `initialize()` overload that takes `ConfigProperties` (instead of waiting on #3304)?

jkwatson:
Happy to have a PR to implement this.

mateuszrzeszutek:
Cool! Can you assign this to me? I should be able to get it done this week.

PR #3441, opened 2021-08-05
Allow using custom configuration sources in sdk-autoconfigure

mateuszrzeszutek:
Fixes https://github.com/open-telemetry/opentelemetry-java/issues/3326


sdk-extensions/autoconfigure/build.gradle.kts
```diff
@@ -117,22 +122,30 @@ tasks {
...
+    environment("OTEL_METRICS_EXPORTER", "none")
   }
 
+  val testResourceNotInitialized by existing(Test::class)
+
   val check by existing {
     dependsOn(
       testConfigError,
       testFullConfig,
+      testInitializeRegistersGlobal,
```
mateuszrzeszutek, line 141: Turns out those tests weren't even added to the `check` task - I've added and fixed them, they were pretty outdated

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
```diff
@@ -25,25 +25,30 @@
...
-  /** Returns the automatically configured {@link Resource}. */
+  /**
+   * Returns the automatically configured {@link Resource} after one of {@code initialize()} methods
+   * has been called. Prior to that, this method will always return an {@linkplain Resource#empty()
+   * empty resource}.
+   */
   public static Resource getResource() {
-    return RESOURCE;
+    return resource;
   }
```
mateuszrzeszutek, line 37: I really dislike this part of this PR: now that we allow using custom `ConfigProperties` impl there's a clear temporal dependency between `getResource()` and `initialize()` - which I guess sort of makes sense but I still don't like it.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
```diff
@@ -6,131 +6,55 @@
...
    */
   @Nullable
-  public String getString(String name) {
-    return config.get(name);
-  }
+  String getString(String name);
+
+  /**
+   * Returns a boolean-valued configuration property. Uses the same rules as {@link
+   * Boolean#parseBoolean(String)} for handling the values.
```
jack-berg, line 26: "Uses the same rules as {@link Boolean#parseBoolean(String)} for handling the values." is an implementation detail.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
```diff
@@ -6,131 +6,55 @@
...
+   * @return false if the property has not been configured.
+   */
+  boolean getBoolean(String name);
 
   /**
    * Returns a integer-valued configuration property.
    *
    * @return null if the property has not been configured.
-   * @throws NumberFormatException if the property is not a valid integer.
+   * @throws ConfigurationException if the property is not a valid integer.
```
jack-berg, line 36: I'm thinking we should be consistent and document all methods of the interface as having the potential to throw `ConfigurationException`.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
```diff
@@ -6,131 +6,55 @@
...
-  }
+  String getString(String name);
+
+  /**
+   * Returns a boolean-valued configuration property. Uses the same rules as {@link
+   * Boolean#parseBoolean(String)} for handling the values.
+   *
+   * @return false if the property has not been configured.
+   */
+  boolean getBoolean(String name);
```
jack-berg, line 30: Why does `getBoolean` return a primitive instead of `Boolean`? The behavior is inconsistent with the other methods that return null if the property is unset.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
```diff
@@ -154,117 +78,23 @@ public Double getDouble(String name) {
...
-      char c = rawValue.charAt(lastDigitIndex);
-      if (Character.isDigit(c)) {
-        break;
-      }
-      lastDigitIndex -= 1;
-    }
-    // Pull everything after the last digit.
-    return rawValue.substring(lastDigitIndex + 1);
-  }
+  Map<String, String> getCommaSeparatedMap(String name);
```
jack-berg, line 99: Don't like that `getCommaSeparatedValues` and `getCommaSeparatedMap` return empty list / map instead of null when not set. I know there's code that depends on that behavior, but if we're making a breaking change by changing this class to an interface, now might be a good time to ensure consistency.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
jack-berg: This is only the case because of the `getResource()` method. Get rid of this method and the `initialize` methods can operate without any side affects in `OpenTelemetrySdkAutoconfiguration`.

To me, `getResource` feels awkward here. Feels like it could live on `OpenTelemetrySdk` or `OpenTelemetry` instead.

jack-berg (commented):
Had a few thoughts on this.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
mateuszrzeszutek: I noticed it too, wasn't sure whether to change it in this PR - but you're right, we're making a breaking change so we might as well fix that one along with it.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
mateuszrzeszutek: We can't put it into `OpenTelemetry` because the resource is a part of SDK and not API, but `OpenTelemetrySdk` sounds good.
@anuraaga @jkwatson was there any reason for not adding it to the `OpenTelemetrySdk`?

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
mateuszrzeszutek: Hmm, I think we're probably returning nulls for Integers/Longs/etc because there's no "natural null object" for those - with a list or a map you can always return an empty one and avoid unneeded null checks.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
jkwatson: I don't agree. I think this is just shorthand for "read those other javadocs to see how this is parsed". The other option would be to copy the contents of that javadoc over here.

jkwatson: Oh, you're saying this probably doesn't belong on the interface javadoc. Hmm. Perhaps changing it to "Implementations should use the same rules as ..." would be the right thing to do here.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
jkwatson: I don't think we should return nulls. That's a bad API to expose, as all callers will need to do null checking. We should return the empty lists/maps as we do today, IMO.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
jkwatson: There is a need to be able to get a handle to the Resource that would be autoconfigured, then modify that and use it to build an SDK. So, it would be very cumbersome to have to build an SDK, then get the resource, modify it and rebuild a new SDK.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
jack-berg: I see where you're coming from, but having responses be nullable allows callers to differentiate between "empty" and "not set". 

Don't feel strongly about this.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
jkwatson: But, I agree that this API is not good, as-is. Perhaps we should split out the Resource autoconfiguration into it's own class. Something like `OpenTelemetryResource.initialize()`

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
mateuszrzeszutek: Hmm, even `getString()` or `getCommaSeparatedValues()` which cannot fail? Also, for `getBoolean()` that'd conflict with the `Implementations should use the same rules as Boolean.parseBoolean()` comment.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
mateuszrzeszutek: Done, sort of - I've added an `OpenTelemetryResourceAutoConfiguration` class that borrows a few patterns from `GlobalOpenTelemetry` and holds the global auto-configured resource.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
```diff
@@ -25,25 +22,20 @@
...
-  /** Returns the automatically configured {@link Resource}. */
-  public static Resource getResource() {
-    return RESOURCE;
-  }
-
   /**
-   * Returns an {@link OpenTelemetrySdk} automatically initialized through recognized system
-   * properties and environment variables.
+   * Returns an {@link OpenTelemetrySdk} automatically initialized through recognized properties
+   * contained in passed {@code config}.
```
jkwatson, line 27:
```suggestion
   * contained in the {@code config} parameter.
```

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetryResourceAutoConfiguration.java
```diff
@@ -0,0 +1,113 @@
...
+import io.opentelemetry.sdk.resources.Resource;
+import java.util.HashSet;
+import java.util.ServiceLoader;
+import java.util.Set;
+import javax.annotation.Nullable;
+
+/** Auto-configuration for the OpenTelemetry {@link Resource}. */
+public final class OpenTelemetryResourceAutoConfiguration {
+
+  @Nullable private static volatile Resource resource = null;
```
jkwatson, line 19: I wonder...how important is it that we keep a copy of this, making a global of the configured Resource? What would be the downside if we just always re-initialized it?

sdk-extensions/autoconfigure/src/testFullConfig/java/io/opentelemetry/sdk/autoconfigure/ResourceTest.java
```diff
@@ -15,7 +16,10 @@
 
   @Test
   void resource() {
-    Attributes attributes = OpenTelemetrySdkAutoConfiguration.getResource().getAttributes();
+    // make sure that the SDK is autoconfigured first regardless of test execution order
+    GlobalOpenTelemetry.get();
```
jkwatson, line 20: why is this necessary when you're only testing the Resource?

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetryResourceAutoConfiguration.java
mateuszrzeszutek: Hmm, there'd be a risk of using different `ConfigProperties` for each call. For example, initializing the SDK with a custom `ConfigProperties` implementation and then calling `getResource()` (that uses default config) could result in different resources. 
So, if somebody used the javaagent's `ConfigPropertySource` SPI to set some resource attributes they'd always be visible in the auto-configured SDK, but if you called `OpenTelemetryResource...#getResource()` without passing the  same `ConfigProperties` you'd get a different resource (constructed just from system properties and env vars).

I don't know it it's a problem though, if we document that behavior it should probably be fine. In case of the instrumentation javaagent the config's globally available, so it'd be fine to just ensure we pass it every time we call `getResource()`.

jkwatson: I think if we never try to claim that there's a single global Resource, (maybe "initialize..." isn't the right name for the method), that this would be fine. It's just a factory for Resource instances based on config.

@anuraaga any thoughts on this?

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ConfigProperties.java
jack-berg: If someone comes along later with an implementation that isn't based on parsing string values like in `DefaultConfigProperties` then javadoc that implies string parsing will seem strange. A semi-realistic example would be if the config source was YAML or JSON with support for different typed values. It would be appropriate for any of the methods to throw if there was a type mismatch.

mateuszrzeszutek: That's a great point -- I'll add this. Thanks!

sdk-extensions/autoconfigure/src/testFullConfig/java/io/opentelemetry/sdk/autoconfigure/ResourceTest.java
mateuszrzeszutek: No longer applicable, removed.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetryResourceAutoConfiguration.java
mateuszrzeszutek: Done - PTAL.

jkwatson: I think this looks good. Would like @anuraaga to chime in before merging.

jkwatson (approved):
Thanks!

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
```diff
@@ -61,6 +52,17 @@ public static OpenTelemetrySdk initialize(boolean setResultAsGlobal) {
...
   }
 
+  /**
+   * Returns an {@link OpenTelemetrySdk} automatically initialized through recognized system
+   * properties and environment variables.
+   *
+   * @param setResultAsGlobal Whether to automatically set the configured SDK as the {@link
+   *     io.opentelemetry.api.GlobalOpenTelemetry} instance.
+   */
+  public static OpenTelemetrySdk initialize(boolean setResultAsGlobal) {
```
anuraaga, line 62: Nit we generally order less argument methods over more argument ones

anuraaga (commented):
Sorry for coming to this late, last holidays for a while.

It seems like this approach will require a lot of duplication - it's still OK I think but was there a reason to not use the `pass in a Map` approach?

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/DefaultConfigProperties.java
```diff
@@ -0,0 +1,220 @@
...
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
+
+/**
+ * Properties to be used for auto-configuration of the OpenTelemetry SDK components. These
+ * properties will be a combination of system properties and environment variables. The properties
+ * for both of these will be normalized to be all lower case, and underscores will be replaced with
+ * periods.
+ */
+final class DefaultConfigProperties implements ConfigProperties {
```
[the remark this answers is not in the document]
mateuszrzeszutek: Most of it is already there - the javaagent `Config` already parses maps, lists, booleans. The only thing missing is the duration parsing.
So, we won't be copying a lot of new code into the javaagent - but `Config` will end up having the same logic as this class, so there'll be duplication in the end.

anuraaga (approved):
This change looks good since it unblocks separating out SPI and the java agent from providing the custom config by duplicating the logic. I think we can followup to see if it's worth deduping that with a simpler mechanism.

jkwatson:
I'm going to merge this so it gets into 1.5.0. Let's see how it goes with the agent using this, and we can adjust if we think we can make it simpler.

PR #3446, opened 2021-08-06
Allow Jaeger remote sampler to be closed.

anuraaga:
When debugging logs of CI, I noticed 90% of the log is spam from leaked jaeger pollers.

### DIFF
--- a/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerBuilder.java
+++ b/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerBuilder.java
@@ -26,6 +26,7 @@ public final class JaegerRemoteSamplerBuilder {
   private String serviceName;
   private Sampler initialSampler = INITIAL_SAMPLER;
   private int pollingIntervalMillis = DEFAULT_POLLING_INTERVAL_MILLIS;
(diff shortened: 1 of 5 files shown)


jack-berg (approved):
I've noticed this too. Nice change.

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerBuilder.java
```diff
@@ -53,6 +54,7 @@ public JaegerRemoteSamplerBuilder setEndpoint(String endpoint) {
   public JaegerRemoteSamplerBuilder setChannel(ManagedChannel channel) {
     requireNonNull(channel, "channel");
     this.channel = channel;
+    closeChannel = false;
```
jkwatson, line 57: I think this worth some javadoc explanation.

jkwatson (approved):
Just needs a bit of javadoc about the non-closing of custom channels, and looks good to me.

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerBuilder.java
jkwatson: let's add javadoc as a follow-up, so we can get rid of the build spam.

PR #3447, opened 2021-08-06
Don't fail fast on tests.

anuraaga:
@jkwatson While sorting through some of the very common CI failures like the HttpSpanExporter one, let's go ahead and use our judgment to merge PRs when a reasonable representative set of the jobs pass, if that's ok with you.

### DIFF
--- a/.github/workflows/main-build.yml
+++ b/.github/workflows/main-build.yml
@@ -10,6 +10,7 @@ jobs:
     name: Build
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


jkwatson:
> @jkwatson While sorting through some of the very common CI failures like the HttpSpanExporter one, let's go ahead and use our judgment to merge PRs when a reasonable representative set of the jobs pass, if that's ok with you.

Assuming that GH will let us merge. I thought we had it set up to require them all to pass, but I might be mis-remembering.

jkwatson (approved):
👍🏽

PR #3448, opened 2021-08-06
Use higher timeout for signal-specific timeout tests.

anuraaga:
The signal-specific tests go lower than our default to verify the flags work. Let's go higher instead - anecdotally, I've never considered a timeout less than 10s to be enough for stable CI.

This PR is because I saw this failure after 5s timeout. This change doesn't guarantee it goes away but I think it's a good idea to try.

```
2021-08-06T04:20:12.1491247Z     WARNING: Failed to export metrics. Error message: DEADLINE_EXCEEDED: deadline exceeded after 4.372567456s. [buffered_nanos=3462370469, remote_addr=localhost/127.0.0.1:46833]
2021-08-06T04:20:12.1492123Z 
2021-08-06T04:20:12.1492903Z OtlpConfigTest > configureMetricExporter() FAILED
2021-08-06T04:20:12.1494066Z     org.opentest4j.AssertionFailedError: 
2021-08-06T04:20:12.1495056Z     Expecting value to be true but was false
2021-08-06T04:20:12.1496710Z         at sun.reflect.NativeConstructorAccessorImp
```


sdk-extensions/autoconfigure/src/testOtlp/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigTest.java
```diff
@@ -133,14 +133,14 @@ void configureExportersGeneral() {
...
-    props.put("otel.exporter.otlp.timeout", "5s");
+    props.put("otel.exporter.otlp.timeout", "15s");
     ConfigProperties properties = ConfigProperties.createForTest(props);
     SpanExporter spanExporter = SpanExporterConfiguration.configureExporter("otlp", properties);
     MetricExporter metricExporter =
         MetricExporterConfiguration.configureOtlpMetrics(
             properties, SdkMeterProvider.builder().build());
 
-    assertThat(spanExporter).extracting("timeoutNanos").isEqualTo(TimeUnit.SECONDS.toNanos(5));
+    assertThat(spanExporter).extracting("timeoutNanos").isEqualTo(TimeUnit.SECONDS.toNanos(15));
```
jack-berg, line 143: You'll want to update `.join(10, TimeUnit.SECONDS)` to match the new timeout.

kubawach (approved):
LGTM

jsuereth (approved):

sdk-extensions/autoconfigure/src/testOtlp/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigTest.java
jkwatson: @anuraaga ?

jkwatson (approved):

sdk-extensions/autoconfigure/src/testOtlp/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigTest.java
anuraaga: My intention was to make sure the test still has a "real" timeout of 10s like most of our tests, the 15 is mostly to check the value is respected, while we still think the test should pass fast. If it's confusing I could raise the join timeout though.

PR #3454, opened 2021-08-07
Adds getParentSpanContext to ReadableSpan interface

HaloFour:
Adds the `getParentSpanContext` method to the `ReadableSpan` interface to get the `SpanContext` of the parent span, or `SpanContext#getInvalid` if the span is a root span.  I believe that this is already allowed by the opentelemetry-specification as it states:

> A function receiving this as argument MUST be able to access all information that was added to the span, as listed in the API spec.

and the API spec explicitly lists the parent span or span context as one of those pieces of information.

This information is already accessible via calling `toSpanData`, but that method is expensive as it synchronized and needs to make copies of the mutable state of the span.

### DIFF
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
@@ -202,6 +202,11 @@ final class RecordEventsReadableSpan implements ReadWriteSpan {
     return context;
   }
 
+  @Override
+  public SpanContext getParentSpanContext() {
+    return parentSpanContext;
+  }
+
   /**
    * Returns the name of the {@code Span}.
    *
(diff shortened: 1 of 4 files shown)


jkwatson:
This seems pretty reasonable to me. @anuraaga any issues with adding this to the SDK's public APIs?

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -451,10 +456,6 @@ private StatusData getSpanDataStatus() {
     }
   }
 
-  SpanContext getParentSpanContext() {
```
jkwatson, line 454: Is there already a unit test that covers this method? If not, can you add one?

jkwatson (approved):

anuraaga (approved):
Looks good, thanks!

PR #3455, opened 2021-08-07
Overhaul sdk.metrics package for new SDK:  Decouple Instrument + Storage per-metric.

jsuereth:
This PR does two major things:

1. Update Instrument names to match the API/SDK specification.
2. Decouple Instrument from Storage view a new `sdk.metrics.internal.state` package.
   - No `Aggregator` API should be used in `sdk.metrics` package.
   - The `sdk.metrics` packages should ONLY include exposed (stable) APIs for users.   This will (only) be:
      - `sdk.metrics.data` for exporting metrics
      - `sdk.metrics` for configuring the SdkMeterProvider.
      - Some TBD `sdk.metrics.view` and `sdk.metrics.aggregator` config.  For now we move the usage of these APIs behind some internal, with more refactorings to come in a follow-on PR.

This assumes that using `.internal.` packages is OK to denote "no exposed API here is intended to remain compatible, you should not depend on these".  This assumption is based on the sheer number of classes exposed in `internal` packages in other SDK components (e.g. threading libraries for trace).

A few callout on changes:

- Asynchronous / Observable instruments no longer have any type associated with them.  They didn't before but more officially now.
- Storage is based on *metric* identity not *instrument* identity.   Instruments register with a central hook (currently it's right on Meter* shared state, but can move with refactoring).
- "Accumulation" has been renamed to "MetricStorage".   The goal is to help associate storage with the resulting METRIC vs. instrument.  Ok renaming this.
- This does *not* rework the existing View/LabelsProcessor API.  That will be a follow-up PR once the groundwork has been laid here (to decouple instrument storage from metric).
- Error messages for instrument-name conflicts are more verbose, hopefully helping users identify conflicting instruments more easily.  I think this could use some more thought, but it's a step up.
- Passing attributes/context always may seem odd in this PR.   This is due to Exemplar sampling coming in a follow up PR.   Happy to tweak the design, just need to make sure attributes are available when sampling Exemplars and currently plan requires attributes to be saved from BEFORE labelsProcessor gets at them.
- There's still a few TODOs around multi-registration of asynchronous instruments.   Some of these need to be raised at the Metrics SiG.


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/DoubleUpDownCounterSdk.java
```diff
@@ -12,28 +12,26 @@
...
 import io.opentelemetry.sdk.metrics.common.InstrumentValueType;
+import io.opentelemetry.sdk.metrics.internal.state.BoundStorageHandle;
+import io.opentelemetry.sdk.metrics.internal.state.MeterProviderSharedState;
+import io.opentelemetry.sdk.metrics.internal.state.MeterSharedState;
+import io.opentelemetry.sdk.metrics.internal.state.WriteableMetricStorage;
 import java.util.function.Consumer;
 
-final class DoubleUpDownCounterSdk extends AbstractSynchronousInstrument
-    implements DoubleUpDownCounter {
+final class DoubleUpDownCounterSdk extends AbstractInstrument implements DoubleUpDownCounter {
```
jkwatson, line 24: For a future PR, I'd prefer that we rename all of these classes to have the `Sdk` as a prefix, rather than a suffix. This isn't an "Sdk" of type "DoubleUpDownCounter", it's the Sdk's implementation of the interface, so using "Sdk" as a prefix makes more sense to me and is consistent with how we ended up naming classes in the tracing SDK.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/common/InstrumentType.java
```diff
@@ -9,8 +9,8 @@
 public enum InstrumentType {
   COUNTER,
   UP_DOWN_COUNTER,
-  VALUE_RECORDER,
-  SUM_OBSERVER,
-  UP_DOWN_SUM_OBSERVER,
-  VALUE_OBSERVER,
+  HISTOGRAM,
+  OBSERVABLE_SUM,
```
jkwatson, line 13: 👍🏽

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/descriptor/MetricDescriptor.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.metrics.internal.descriptor;
```
jkwatson, line 6: this class is only used by the classes in the `internal.state` package. Does it need its own package?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/BoundStorageHandle.java
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.metrics.internal.state;
+
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.context.Context;
+
+/** A bound handle for recording measurements against a particular set of attributes. */
```
jkwatson, line 11: At some point we should add our boilerplate with warnings about using `internal` classes to all the public internal classes, but that can wait until things are a bit more settled.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MeterProviderSharedState.java
```diff
@@ -0,0 +1,60 @@
...
+  /** Returns the {@link Aggregator} to use for a given instrument. */
+  public <T> Aggregator<T> getAggregator(
+      MeterSharedState meterSharedState, InstrumentDescriptor descriptor) {
+    return getViewRegistry()
+        .findView(descriptor)
+        .getAggregatorFactory()
+        .create(getResource(), meterSharedState.getInstrumentationLibraryInfo(), descriptor);
+  }
+
+  // TODO: Move this.
```
jkwatson, line 51: where should it be moved to?

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/SdkMeterTest.java
```diff
@@ -31,20 +31,21 @@ void testLongCounter() {
...
+    sdkMeter
+        .counterBuilder("testLongCounter")
+        .setDescription("My very own counter")
+        .setUnit("metric tonnes")
+        .build();
 
     assertThatThrownBy(() -> sdkMeter.counterBuilder("testLongCounter").build())
         .isInstanceOf(IllegalArgumentException.class)
-        .hasMessage("Instrument with same name and different descriptor already created.");
+        .hasMessageContaining("Metric with same name and different descriptor already created.");
```
jkwatson, line 45: hmm. it's not a `Metric` that's being created here, but an `Instrument`. I think this message should be reverted to what it was previously.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/SynchronousMetricStorageTest.java
```diff
@@ -59,44 +64,56 @@ public Attributes onLabelsBound(Context ctx, Attributes lbls) {
...
       } finally {
         anotherDuplicateAggregatorHandle.release();
       }
     } finally {
       duplicateAggregatorHandle.release();
-      aggregatorHandle.release();
+      handle.release();
     }
 
+    // TODO: Is it necessary to test this?
```
jkwatson, line 114: is there a problem with testing it?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageRegistry.java
```diff
@@ -0,0 +1,76 @@
...
+   */
+  @SuppressWarnings("unchecked")
+  public <I extends MetricStorage> I register(I storage) {
+    MetricDescriptor descriptor = storage.getMetricDescriptor();
+    MetricStorage oldOrNewStorage =
+        registry.computeIfAbsent(descriptor.getName().toLowerCase(), key -> storage);
+    // Make sure the storage is compatible.
+    if (!oldOrNewStorage.getMetricDescriptor().isCompatibleWith(descriptor)) {
+      throw new IllegalArgumentException(
+          "Metric with same name and different descriptor already created.   Found: "
```
jkwatson, line 60: (Putting the comment in the source where it belongs). This message shouldn't say "Metric with same...", because there isn't a thing called a "Metric" here. It's the Instrument that is trying to be created with the same name. This message will be confusing to users to see it in their code, IMO.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/DoubleUpDownCounterSdk.java
jsuereth: I'm totally on-board.  If you look at my wip-metrics branch I had done that, and retrofitting commits against main has been hard due to it.  I also prefer `Sdk*`.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/descriptor/MetricDescriptor.java
jsuereth: Not when viewed from this PR no.   Eventually this will be used/returned by `View`s.   

Specifically:  Right now the pre-existing `View` code does not allow metric renaming.  The new SDK specification does.  

The code you see that creates this today is a stop-gap for that implementation.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/BoundStorageHandle.java
jsuereth: Ah, if you point me at the boilerplate, I can put it in here, or look at some kind of spotless hook to auto-add it to internal packages.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MeterProviderSharedState.java
jsuereth: My plan is to consolidate the code which determines default aggregations + view lookups in one class.  It's a bit fragmented here, but I did think the "MeasurementProcessor" was a nice component.  While I don't think it'll be exposed as an interface, I plan to add it back with a different component.

I can clean all these TODOs out though, somehow missed them in my review of the PR.  Leaving this open to check on plan of consolidating storage-wiring code into a single component.  E.g. see: https://github.com/jsuereth/opentelemetry-java/blob/wip-metrics-api/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/DefaultMeasurementProcessor.java#L51

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/BoundStorageHandle.java
jkwatson: Here's an example: https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/common/src/main/java/io/opentelemetry/sdk/internal/RateLimiter.java#L17-L18

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageRegistry.java
jsuereth: Well, except Views make this hella confusing.  Specifically this *COULD* be a name conflict from an instrument to a view that-has-a-rename.

I agree it's confusing to users.   I decided to consolidate on this notion of Metric.   When updating the View implementation I was going to add some provenance here so we can give good error messages.

Specifically, right now you see it lists the metric descriptors.   Post-View-Rework I expect this to have logic somewhat as follows:


1. If conflict involves two instruments with default aggregations, error as instrument conflict.
2. If conflict involves one instrument + one view re-name, error as instrument conflicting with "view" metric output.  Output the name of the view + its selector.
3. Conflicts involving multiple views *should* be caught when building the SDK and shouldn't make it to this level.


I can take a crack at showing that logic here or keeping it instrument based, since you can't have view renames just yet.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/SynchronousMetricStorageTest.java
jsuereth: It's an implementation detail.  Is it worth exposing this interface on the Storage class to test this?  

Additionally there are already tests in place that make sure if an Attribute-set is not reported within the interval that it stops showing up.  I don't think we need to also check the contract of the `tryUnmap` method here.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageRegistry.java
jkwatson: yeah...that will get even more confusing. I won't hold up this PR for it... and getting some actual user feedback will be valuable.

jsuereth: +100

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/BoundStorageHandle.java
jsuereth: Ok, added that to all internal packages.  Did a quick look at spotless plugin. Think we COULD add this as an automatic check, but it's more than a 5 minute task, so deferring.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/SynchronousMetricStorageTest.java
jkwatson: ah, I see. If removing this would let us narrow the interface, then 👍🏽

jkwatson (approved):
:shipit:

Issue #3443, opened 2021-08-05
Support comma-separated list of exporters for otel.traces.exporter


jkwatson:
This has been added to the spec and would be a nice-to-have addition to the auto-configuration module.

iNikem:
Hm... spec says nothing about if all those exporters should use the same span processor or not

jkwatson:
> Hm... spec says nothing about if all those exporters should use the same span processor or not

Yes, I know. I brought this up in the spec PR that added this, but it was decided it would be up to the individual SIGs to decide how to do that. General recommendation was to have each hooked up with their own `BatchSpanProcessor`, I believe.

anuraaga:
> General recommendation was to have each hooked up with their own BatchSpanProcessor, I believe.

I can't think of any benefit of this. I think we should hook up to the same BSP, just by using a composite span exporter.

jkwatson:
> > General recommendation was to have each hooked up with their own BatchSpanProcessor, I believe.
> 
> I can't think of any benefit of this. I think we should hook up to the same BSP, just by using a composite span exporter.

Also fine, and I agree.

jack-berg:
@jkwatson can you assign this issue to me?

jack-berg:
@jkwatson this issue can be closed. Looks like the comment I left on the PR #3529 didn't meet the criteria to close.

PR #3457, opened 2021-08-08
Add a new configurable metric exporter to autoconf

jsuereth:
- Creates ConfigurableMetricExporterProvider SPI
- Updates Metric setup to use SPI registry after built-in

### DIFF
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
@@ -9,6 +9,7 @@ import io.opentelemetry.exporter.logging.LoggingMetricExporter;
 import io.opentelemetry.exporter.otlp.metrics.OtlpGrpcMetricExporter;
 import io.opentelemetry.exporter.otlp.metrics.OtlpGrpcMetricExporterBuilder;
 import io.opentelemetry.exporter.prometheus.PrometheusCollector;
+import io.opentelemetry.sdk.autoconfigure.spi.ConfigurableMetricExporterProvider;
 import io.opentelemetry.sdk.metrics.SdkMeterProvider;
 import io.opentelemetry.sdk.metrics.export.IntervalMetricReader;
 import io.opentelemetry.sdk.metrics.export.IntervalMetricReaderBuilder;
@@ -21,10 +22,12 @@ import java.nio.file.Paths;
 import java.time.Duration;
 import java.util.Collections;
 import java.util.Map;
+import java.util.ServiceLoader;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 import javax.annotation.Nullable;
 
 final class MetricExporterConfiguration {
-
   static void configureExporter(
       String name, ConfigProperties config, SdkMeterProvider meterProvider) {
     switch (name) {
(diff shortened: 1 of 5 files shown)


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
```diff
@@ -41,11 +44,32 @@ static void configureExporter(
             "opentelemetry-exporter-logging");
         configureLoggingMetrics(config, meterProvider);
         return;
+      case "none":
+        return;
       default:
+        MetricExporter spiExporter = configureSpiExporter(name, config);
```
jkwatson, line 50: Are we ok that autoconfiguration only supports push based exporters using this method? Are there any other pull-based backends out there, aside from prometheus to worry about?

jkwatson: Maybe the bigger question is: is this going to be good enough, given the requirements that multiple metric exporters running at the same time need to be supported by the SDK?

jsuereth: So I looked at this.  MetricExporter *cannot* be a pull-based exporter.   Go take a gander at [prometheus exporter](https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/PrometheusCollector.java#L15), it's not a `MetricExporter`. 

Effectively we have the Hierarchy:

- Pull-Based Exporters
  - Prometheus
  - IntervalMetricReader
     - MetricExporter
       - OTLP
       - InMemory
       - ... extensions...
       
I was worried about putting this behind IMR until I looked at the details.

jkwatson: oh, yes, you are correct about that, technically. I. just to to make sure that putting this into the SPI is going to fulfill enough use-cases, and be future-proof enough to justify it, given the still ¯\_(ツ)_/¯ state of support for multiple exporters in the SDK.

jsuereth: This is fully inline w/ autoconf for trace (which only allows ONE autoconfigured exporter), even if trace can do mulitple in the SDK.

I'd suggest this can get merged and if we do support multiple here it's an autoconf change.

jkwatson: heh. https://github.com/open-telemetry/opentelemetry-java/issues/3443

Agree it's probably fine...I just want to make sure we've all thought it through. :)

jkwatson (approved):
looks good to me. 🤞🏽  it won't have to change much as the SDK evolves.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
jsuereth: Yeah. I think you should push on that part of this PR.   I'd *love* to be able to solve multi-exporter support (for trace + metrics), but I think we're ok for now.  Should I open any bugs around multi-exporter support?

jkwatson: > Yeah. I think you should push on that part of this PR. I'd _love_ to be able to solve multi-exporter support (for trace + metrics), but I think we're ok for now. Should I open any bugs around multi-exporter support?

do we know how this is going to work in the SDK yet? It's really easy for tracing/spans right now. Metrics is a different beast with the collection cycle(s), though.

anuraaga (approved):

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
jsuereth: Right now there's been two major directions discussed:

 One direction is very close to what we have now where multi-exporter semantics are handled by something like IMR and pull-based receivers become (somewhat) stateful.  I think this is what Reiley showed us yesterday in the SiG.

The other option is the multi-collection idea that's in the prototype on my wip-metrics branch.   

I think the first option will be technically easier too maintain and cleaner.  Both approaches, I suspect, will work well with this exposed exporter interface (we just may have another SPI for pull-based exporters like prometheus).

PR #3458, opened 2021-08-09
Add getAttribute to ReadableSpan

### LINKED ISSUE #3456: How to stamp a "tenant id" attribute on all spans?

I want to stamp a "tenant id" attribute on all spans.

Open to suggestions.

One option that I think would work well is to build a SpanProcessor, and onStart(), copy the "tenant id" attribute from the parent span down to the new span.

Then the application can stamp the "tenant id" on the SERVER span in a servlet filter, and it will get propagated down to all spans, including auto-instrumentation spans (if you register the SpanProcessor via new javaagent extension mechanism).

Currently though this would require calling toSpanData() on the active parent span which is expensive, so if going this route I'd like to see if it's ok to add a `getAttribute(AttributeKey)` method to ReadableSpan.

Another option is to put the "tenant id" into the `Context`, and have a SpanProcessor that copies the "tenant id" into the new span.

The main problem with this approach is that it doesn't work great with the javaagent because the `ContextKey` can't easily be shared between the application that is putting the "tenant id" into the `Context` and the SpanProcessor that lives in the javaagent (extension).

A possible mitigation to this problem is to put the "tenant id" context key into the bootstrap class loader, which is possible now via extensions, but this starts to feel like a pretty complex solution to a simple-ish use case.

Yet another option is to put the "tenant id" into baggage, which would not have the javaagent problem that custom context keys have. But there are no controls on baggage propagation over the wire (except to disable it completely), which makes me hesitant to recommend this as a general solution to this use case.

**Additional context**

I believe this is similar use case to https://github.com/open-telemetry/opentelemetry-specification/issues/1337

trask:
Resolves #3456


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -190,6 +190,14 @@ public SpanData toSpanData() {
     }
   }
 
+  @Override
+  @Nullable
+  public <T> T getAttribute(AttributeKey<T> key) {
```
Oberon00, line 195: Except for `hasEnded` (write-once) all other getters are for read-only fields. From a theoretical perspective, I would prefer *not* to have this getter, if we can find another solution to the original problem on the issue.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/ReadableSpan.java
```diff
@@ -78,4 +80,8 @@
    * @return the kind of the span.
    */
   SpanKind getKind();
+
+  /** Returns the value for the given {@link AttributeKey}, or {@code null} if not found. */
```
jkwatson, line 84: I think it's worth a note here about the fact that this call does require some locking, and that subsequent calls may result in different results, due to the inherent mutability of the underlying object.

jkwatson (approved):
I'm ok with this, as long as we add some documentation about the performance implications of using it.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/ReadableSpan.java
trask: do you want me to add that same note for `hasEnded()`, `getName()`, and `getLatencyNanos()` also?

jkwatson: that's actually probably a good idea, yes, if you don't mind.

PR #3460, opened 2021-08-10
Migrate B3 propagator integration test to Armeria.

anuraaga:


integration-tests/src/test/java/io/opentelemetry/B3PropagationIntegrationTest.java
```diff
@@ -20,67 +31,139 @@
...
-import org.testcontainers.shaded.okhttp3.Request;
-import org.testcontainers.shaded.okhttp3.Response;
-import org.testcontainers.shaded.okhttp3.ResponseBody;
-import spark.Service;
 
 /** Integration tests for the B3 propagators, in various configurations. */
-public class B3PropagationIntegrationTest {
+class B3PropagationIntegrationTest {
 
-  private static final int server1Port = new Random().nextInt(40000) + 1024;
```
anuraaga, line 43: While the probability is of course low, I did see a CI failure caused by both ports resolving to the same number which prompted this PR.

jkwatson (approved):
Thanks!

PR #3466, opened 2021-08-11
Fix OpenTelemetrySdk(.tracerProvider).tracerBuilder() being noop.

Oberon00:
The default method implementation of TracerProvider.tracerBuilder and OpenTelemetry.tracerBuilder is not very helpful, as demonstrated by this PR.

~~I will amend the PR later to provide a fix, just wanted to give you a quick heads-up  since the release of 1.5 seems imminent -- probably this should be fixed before. @jkwatson @anuraaga~~

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
+++ b/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
@@ -206,5 +206,10 @@ public final class GlobalOpenTelemetry {
     public ContextPropagators getPropagators() {
       return delegate.getPropagators();
     }
+
+    @Override
+    public TracerBuilder tracerBuilder(String instrumentationName) {
+      return delegate.tracerBuilder(instrumentationName);
+    }
   }
 }
--- a/sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
+++ b/sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
@@ -7,6 +7,7 @@ package io.opentelemetry.sdk;
 
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.api.trace.TracerBuilder;
 import io.opentelemetry.api.trace.TracerProvider;
 import io.opentelemetry.context.propagation.ContextPropagators;
 import io.opentelemetry.sdk.trace.SdkTracerProvider;
@@ -71,6 +72,11 @@ public final class OpenTelemetrySdk implements OpenTelemetry {
     @Override
     public Tracer get(String instrumentationName, String instrumentationVersion) {
       return delegate.get(instrumentationName, instrumentationVersion);
+    }
+
+    @Override
+    public TracerBuilder tracerBuilder(String instrumentationName) {
+      return delegate.tracerBuilder(instrumentationName);
     }
 
     public SdkTracerProvider unobfuscate() {
--- a/sdk/all/src/test/java/io/opentelemetry/sdk/OpenTelemetrySdkTest.java
+++ b/sdk/all/src/test/java/io/opentelemetry/sdk/OpenTelemetrySdkTest.java
@@ -73,9 +73,20 @@ class OpenTelemetrySdkTest {
   @Test
   void testShortcutVersions() {
     assertThat(GlobalOpenTelemetry.getTracer("testTracer1"))
-        .isEqualTo(GlobalOpenTelemetry.getTracerProvider().get("testTracer1"));
+        .isSameAs(GlobalOpenTelemetry.getTracerProvider().get("testTracer1"));
     assertThat(GlobalOpenTelemetry.getTracer("testTracer2", "testVersion"))
-        .isEqualTo(GlobalOpenTelemetry.getTracerProvider().get("testTracer2", "testVersion"));
+        .isSameAs(GlobalOpenTelemetry.getTracerProvider().get("testTracer2", "testVersion"));
+    assertThat(
+            GlobalOpenTelemetry.tracerBuilder("testTracer2")
+                .setInstrumentationVersion("testVersion")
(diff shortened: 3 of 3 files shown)


anuraaga (approved):
Thanks @Oberon00 - agree the default implementation was probably a miss, but glad we're able to work with it without breaking compatibility

Oberon00:
I think an alternative default implementation would be to return a working tracer provider that uses the other interface methods and just ignores setSchemaUrl.

Oberon00:
Build `(macos-latest, 17)` failed with a probably unrelated error:

```
> Task :sdk:trace:test

BatchSpanProcessorTest > exporterThrowsException() FAILED
    java.lang.NullPointerException: Cannot invoke "io.opentelemetry.sdk.common.CompletableResultCode.whenComplete(java.lang.Runnable)" because "code" is null
        at io.opentelemetry.sdk.common.CompletableResultCode.ofAll(CompletableResultCode.java:48)
        at io.opentelemetry.sdk.trace.export.MultiSpanExporter.shutdown(MultiSpanExporter.java:94)
        at io.opentelemetry.sdk.trace.export.BatchSpanProcessor$Worker.lambda$shutdown$2(BatchSpanProcessor.java:268)
        at io.opentelemetry.sdk.common.CompletableResultCode.whenComplete(CompletableResultCode.java:126)
        at io.opentelemetry.sdk.trace.export.BatchSpanProcessor$Worker.shutdown(BatchSpanProcessor.java:265)
        at io.opentelemetry.sdk.trace.export.BatchSpanProcessor$Worker.access$200(BatchSpanProcessor.java:125)
        at io.opentelemetry.sdk.trace.export.BatchSpanProcessor.shutdown(BatchSpanProcessor.java:110)
        at io.opentelemetry.sdk.trace.TracerSharedState.shutdown(TracerSharedState.java:96)
        at io.opentelemetry.sdk.trace.SdkTracerProvider.shutdown(SdkTracerProvider.java:115)
        at io.opentelemetry.sdk.trace.export.BatchSpanProcessorTest.cleanup(BatchSpanProcessorTest.java:69)
```

That looks more serious than a simple timeout. Should I create an issue?

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
```diff
@@ -73,6 +74,11 @@ public Tracer get(String instrumentationName, String instrumentationVersion) {
       return delegate.get(instrumentationName, instrumentationVersion);
     }
 
+    @Override
+    public TracerBuilder tracerBuilder(String instrumentationName) {
+      return delegate.tracerBuilder(instrumentationName);
+    }
```
jkwatson, line 80: why is this necessary? The default implementation should cover this, precisely, no?

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
```diff
@@ -206,5 +206,10 @@ public TracerProvider getTracerProvider() {
     public ContextPropagators getPropagators() {
       return delegate.getPropagators();
     }
+
+    @Override
+    public TracerBuilder tracerBuilder(String instrumentationName) {
+      return delegate.tracerBuilder(instrumentationName);
+    }
```
jkwatson, line 213: do I completely not understand how default methods and method dispatch works or something? I don't see why this is necessary given the default implementation.

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
Oberon00: `ObfuscatedTracerProvider` derives from the `TracerProvider` interface which has this default impl:

https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/trace/TracerProvider.java#L58-L60

`return DefaultTracerBuilder.getInstance();` is a no-op TracerBuilder.

jkwatson: ah. the issue here is in the ObfuscatedTracerProvider, not in this class at all. We should definitely fix the ObfuscatedTracerProvider. This particular class I don't care so much about since it works as its supposed to.

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
Oberon00: Here, the default implementation actually does something sensible:

https://github.com/open-telemetry/opentelemetry-java/blob/d9e1151336afc5778a110863bb9aa4178af567e3/api/all/src/main/java/io/opentelemetry/api/trace/TracerProvider.java#L75-L77

Still, we bypass any override that the delegate might have, calling `delegate.getTracerProvider().tracerBuilder(instrumentationName)` instead of `delegate.tracerBuilder(instrumentationName)`. This will hopefully do the same, but this bypass would still be very surprising in case one adds logging to `tracerBuilder` or an auto-instrumenting agent instruments it, etc.

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
Oberon00: This *is* the ObfuscatedTracerProvider which is a static inner class of OpenTelemetrySdk.

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
jkwatson: I assume this is also due to the missing implementation in the ObfuscatedTracerProvider. I'd rather fix it there, rather than patch the issue and not fix the underlying problem.

jkwatson: ugh. I also missed that this was the implementation in the obfuscated impl. Sorry for the noise.

jkwatson (approved):
thanks!

PR #3461, opened 2021-08-10
Clean up generating version.properties.

Oberon00:
Crucially, this adds `inputs.property("project.version", project.version)` to avoid stale builds.

PR #3467, opened 2021-08-11
Fix Resource version properties lookup and extract version file gener…

anuraaga:
…ation.

I think it's good practice to add the version to a file in any JARs we publish as it allows verification of version alignment at runtime as needed. So this extracts creation of the version specifier to the java conventions itself. Note that `MANIFEST` tends to not work well for this because fatjars often lose it.

### DIFF
--- a/buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
@@ -134,23 +134,28 @@ tasks {
       }
     }
   }
-
-  register("generateVersionResource") {
-    val moduleName = otelJava.moduleName
-    val propertiesDir = moduleName.map { file("build/generated/properties/${it.replace('.', '/')}") }
-
-    inputs.property("project.version", project.version.toString())
-    outputs.dir(propertiesDir)
-
-    doLast {
-      File(propertiesDir.get(), "version.properties").writeText("sdk.version=${project.version}")
-    }
-  }
 }
 
-sourceSets {
-  main {
-    output.dir("build/generated/properties", "builtBy" to "generateVersionResource")
+// Add version information to published artifacts.
+plugins.withId("maven-publish") {
+  tasks {
+    register("generateVersionResource") {
+      val moduleName = otelJava.moduleName
+      val propertiesDir = moduleName.map { file("build/generated/properties/${it.replace('.', '/')}") }
+
+      inputs.property("project.version", project.version.toString())
+      outputs.dir(propertiesDir)
+
+      doLast {
+        File(propertiesDir.get(), "version.properties").writeText("sdk.version=${project.version}")
+      }
+    }
+  }
+
+  sourceSets {
+    main {
+      output.dir("build/generated/properties", "builtBy" to "generateVersionResource")
+    }
   }
 }
 
--- a/buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
@@ -134,6 +134,24 @@ tasks {
       }
     }
   }
+
+  register("generateVersionResource") {
+    val moduleName = otelJava.moduleName
+    val propertiesDir = moduleName.map { file("build/generated/properties/${it.replace('.', '/')}") }
+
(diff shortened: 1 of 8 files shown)


sdk/common/src/test/java/io/opentelemetry/sdk/resources/ResourceTest.java
```diff
@@ -211,7 +211,8 @@ void testDefaultResources() {
     assertThat(attributes.get(ResourceAttributes.SERVICE_NAME)).isEqualTo("unknown_service:java");
     assertThat(attributes.get(ResourceAttributes.TELEMETRY_SDK_NAME)).isEqualTo("opentelemetry");
     assertThat(attributes.get(ResourceAttributes.TELEMETRY_SDK_LANGUAGE)).isEqualTo("java");
-    assertThat(attributes.get(ResourceAttributes.TELEMETRY_SDK_VERSION)).isNotNull();
```
anuraaga, line 214: This used to pass since it was `"unknown"` 😱

Oberon00:
I don't think we want/need this version.properties everywhere. Please take a look at #3461 where I only generate it into the sdk-common jar. Note that only one of #3461 or this PR can be merged.

anuraaga:
@Oberon00 the version files can be used to verify version alignment and have almost no impact on artifact size. Why not want them everywhere?

Oberon00:
So, I like the additional changes of this PR (looking in sdk/common/ instead of sdk/, additional tests) but I don't see why we need the version.properties in every jar. We have the Implementation-Version in the MANIFEST.MF in every jar already. As suggested in https://github.com/open-telemetry/opentelemetry-java/pull/3461#issuecomment-896070884, we could use the manifest version directly instead of generating *any* version.properties.

anuraaga:
Yeah but

> Note that MANIFEST tends to not work well for this because fatjars often lose it.

In my experience, MANIFEST tends to be mostly useless in practice since someone will file a bug when their fatjar causes problems with it.

Oberon00:
OK, that's a valid argument. It *is* potentially useful to have the version.properties in the API jar at least.

But I'd add the version.properties to the publishing conventions instead of the java conventions, or do we also want these for unpublished jars?

anuraaga:
@Oberon00 Agree that publish conventions would be enough, but the Gradle configuration was simpler in `java-conventions` with the availability of the `moduleName`. The particular issue is that we have `bom` artifacts that are published, but are not `java-library`. The downside for non-published is just a very small amount of build time I think so figured it's OK.

Oberon00:
How about mixing in the old approach and using `plugins.withId("maven-publish")` in the java conventions?

anuraaga:
Thanks, nice idea!

Oberon00:
Actually, instead of `maven-publish`, maybe using the ID of the otel publishing convention plugin would be better?

buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -136,6 +136,29 @@ tasks {
   }
 }
 
+// Add version information to published artifacts.
+plugins.withId("maven-publish") {
+  tasks {
+    register("generateVersionResource") {
+      val moduleName = otelJava.moduleName
+      val propertiesDir = moduleName.map { file("build/generated/properties/${it.replace('.', '/')}") }
```
Oberon00, line 144:
```suggestion
      val propertiesDir = moduleName.map { File(buildDir, "generated/properties/${it.replace('.', '/')}") }
```

buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -136,6 +136,29 @@ tasks {
...
+
+      doLast {
+        File(propertiesDir.get(), "version.properties").writeText("sdk.version=${project.version}")
+      }
+    }
+  }
+
+  sourceSets {
+    main {
+      output.dir("build/generated/properties", "builtBy" to "generateVersionResource")
```
Oberon00, line 157:
```suggestion
      output.dir("$buildDir/generated/properties", "builtBy" to "generateVersionResource")
```

Oberon00 (approved):

jkwatson (approved):
Thanks!

PR #3468, opened 2021-08-11
Add Resource.getAttribute() 

breedx-splk:
It's pretty common to see code that just wants to get the value of a single resource attribute...and there wasn't a method to do this, so there were law of demeter violations sprinkled about. This adds the `getResource(key)` method and cleans up some of the `.getAttributes().get(key)` calls. 

Vendors will be able to leverage this too.


jkwatson (approved):
This seems very reasonable to me. @anuraaga What do you say about this convenience method?

carlosalberto (approved):

anuraaga (approved):
Nice - this is similar to how we added `ResourceBuilder` too

PR #3472, opened 2021-08-12
Fix nullness in Resource (low priority).

anuraaga:


sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -120,7 +118,7 @@ private static String readVersion() {
       // we left the attribute empty
       return "unknown";
     }
-    return properties.getProperty("sdk.version");
+    return properties.getProperty("sdk.version", "unknown");
```
anuraaga, line 121: This seems to match the behavior of line 119. Otherwise I could of course have the caller check null

sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java
```diff
@@ -148,10 +146,6 @@ private static String readVersion() {
     return getAttributes().get(key);
   }
 
-  @Memoized
```
anuraaga, line 151: There seems to be a bug in auto-value where the use of the Memoized annotation causes Nullable annotations to not be present in the generated code correctly. Anyways, this stood out as unexpected to me as using Resource as a map key doesn't seem common enough to memoize this? Otherwise I'd rather remove auto-value and precompute this instead, this class only has two fields.

jkwatson: we use Resource as a map key in the OTLP exporter code, to group output by resource: https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/SpanAdapter.java#L70-L71

anuraaga: Thanks for noticing this! And I sent #3477 which seems like a good idea regardless of this. Does it mean we may as well stick with the AutoValue default?

jkwatson: I think the auto-value default is fine. If we find a use-case where we really need an efficient hashcode, we can go with some manual pre-computing of the value.

jkwatson (approved):
Do you want to create an issue to track optimizing the hashcode, or should we let that lie for now?

anuraaga:
> Do you want to create an issue to track optimizing the hashcode, or should we let that lie for now?

I think let lie since no pressing need.

PR #3474, opened 2021-08-12
Fix some nullability in SDK.

anuraaga:


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -370,7 +368,7 @@ public ReadWriteSpan setStatus(StatusCode statusCode, @Nullable String descripti
 
   @Override
   public ReadWriteSpan recordException(Throwable exception) {
-    recordException(exception, null);
+    recordException(exception, Attributes.empty());
```
anuraaga, line 371: Alternative is to extract a private method with a Nullable argument, but this seemed fine to me

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/data/StatusData.java
```diff
@@ -38,8 +39,8 @@ static StatusData error() {
    * Returns a {@link StatusData} with the given {@code code} and {@code description}. If {@code
    * description} is {@code null}, the returned {@link StatusData} does not have a description.
    */
-  static StatusData create(StatusCode code, String description) {
-    return ImmutableStatusData.create(code, description);
+  static StatusData create(StatusCode code, @Nullable String description) {
```
anuraaga, line 42: I just added Nullable to the signature since we document it in the javadoc. I'm not sure why jamp isn't showing a diff for this

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/data/StatusData.java
```diff
@@ -38,8 +39,8 @@ static StatusData error() {
...
+    return ImmutableStatusData.create(code, description != null ? description : "");
```
anuraaga, line 43: We use empty string for the description for our no-description singletons so figure this was intended. Currently the code seems buggy since it stores a null description which can be accessed by non-null getDescription

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/data/StatusData.java
anuraaga: Perhaps jcmp is smarter than expected? Probably not, but anyways just one thing to keep in mind is that adding `@Nullable` to a return type can cause kotlin code to stop compiling, while adding to a parameter cannot, so former is technically not semver compatible and later is (if we actually considered kotlin compatibility when defining our semver :P)

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java
```diff
@@ -389,10 +387,7 @@ public ReadWriteSpan recordException(Throwable exception, Attributes additionalA
     StringWriter writer = new StringWriter();
     exception.printStackTrace(new PrintWriter(writer));
     attributes.put(SemanticAttributes.EXCEPTION_STACKTRACE, writer.toString());
-
-    if (additionalAttributes != null) {
```
jkwatson, line 393: don't we still need to guard against this, since a user can ignore the non-nullability and pass in null?

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkSpanBuilder.java
```diff
@@ -234,6 +227,14 @@ public Span startSpan() {
         startEpochNanos);
   }
 
+  private AttributesMap attributes() {
+    AttributesMap attributes = this.attributes;
+    if (attributes == null) {
+      attributes = this.attributes = new AttributesMap(spanLimits.getMaxNumberOfAttributes());
```
jkwatson, line 233: hm. I'd definitely prefer to avoid the double assignment in a single line, just from a readability perspective.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkTracer.java
```diff
@@ -28,9 +28,13 @@ public SpanBuilder spanBuilder(String spanName) {
...
     if (sharedState.hasBeenShutdown()) {
-      return TracerProvider.noop()
-          .get(instrumentationLibraryInfo.getName(), instrumentationLibraryInfo.getVersion())
-          .spanBuilder(spanName);
+      Tracer tracer =
+          instrumentationLibraryInfo.getVersion() != null
+              ? TracerProvider.noop()
+                  .get(
+                      instrumentationLibraryInfo.getName(), instrumentationLibraryInfo.getVersion())
+              : TracerProvider.noop().get(instrumentationLibraryInfo.getName());
```
jkwatson, line 36: good catch.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/data/StatusData.java
jkwatson: I bet jamicmp just isn't smart enough to look at parameter annotations.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -250,8 +250,11 @@ private void flush() {
         }
       }
       exportCurrentBatch();
-      flushRequested.get().succeed();
-      flushRequested.set(null);
+      CompletableResultCode flushResult = flushRequested.get();
+      if (flushResult != null) {
+        flushResult.succeed();
+        flushRequested.set(null);
+      }
```
jkwatson, line 257: good catch here.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkTracer.java
Oberon00: But doesn't noop ignore all that anyway? Couldn't we just always pass the name?

Oberon00:
```suggestion
      Tracer tracer = TracerProvider.noop().get(instrumentationLibraryInfo.getName());
```

jkwatson: yes, this is correct. I remember looking at this and thinking the same, but somehow I forgot to write it down. thanks!

jkwatson (approved):
Thanks!

PR #3480, opened 2021-08-13
Clarify japicmp shouldn't be attempted until some time has passed aft…

anuraaga:
…er release workflow.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -31,7 +31,8 @@ for a list of major changes since last release.
 
 ## Update release versions in documentations and CHANGELOG files
 
-After releasing is done, you need to first update the docs.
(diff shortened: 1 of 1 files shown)


kubawach (approved):
LGTM (also an easy one).

Oberon00:
But this warning is only necessary because of #3481, right? Maybe we should retry with `--refresh-dependencies` if it fails without?

jkwatson:
> But this warning is only necessary because of #3481, right? Maybe we should retry with `--refresh-dependencies` if it fails without?

Well, you always will need to wait some amount of time, since maven central does take some time (sometimes a lot of time) to get fully synced up. But, we should definitely not do what is in #3481, since we look up `:latest` and gradle will definitely cache that for way too long.

jkwatson (approved):

PR #3481, opened 2021-08-13
Remove refresh-dependencies for japi

anuraaga:
@jkwatson Do we need to refresh dependencies for this command?


kubawach (approved):
LGTM (this was a quick one ;) )

jkwatson:
> @jkwatson Do we need to refresh dependencies for this command?

Yes. Since we use `:latest` as the version for looking up the latest-release, you need to `--refresh-dependencies` for gradle to pick up the actual latest after a release. (assuming you've run this command already before the release)

jkwatson (changes requested):
I don't think we should remove this, due to using the `:latest` version for doing the API diff generation.

anuraaga:
Got it - thanks for clarifying :)

PR #3490, opened 2021-08-14
Make proto dependency on gRPC compileOnly.

anuraaga:
otlp-http currently includes an unused gRPC dependency because of this which isn't intended. Our gRPC exporters actually already had implementation dependencies on these artifacts anyways.

### DIFF
--- a/exporters/otlp-http/trace/build.gradle.kts
+++ b/exporters/otlp-http/trace/build.gradle.kts
@@ -13,6 +13,7 @@ dependencies {
 
   implementation(project(":exporters:otlp:common"))
 
+  implementation("com.google.api.grpc:proto-google-common-protos")
   implementation("com.squareup.okhttp3:okhttp")
   implementation("com.squareup.okhttp3:okhttp-tls")
   implementation("com.squareup.okio:okio")
--- a/exporters/otlp-http/trace/build.gradle.kts
+++ b/exporters/otlp-http/trace/build.gradle.kts
@@ -13,12 +13,12 @@ dependencies {
 
   implementation(project(":exporters:otlp:common"))
 
-  implementation("com.google.api.grpc:proto-google-common-protos")
   implementation("com.squareup.okhttp3:okhttp")
   implementation("com.squareup.okhttp3:okhttp-tls")
   implementation("com.squareup.okio:okio")
 
   testImplementation(project(":sdk:testing"))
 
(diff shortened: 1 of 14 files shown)


exporters/otlp-http/trace/build.gradle.kts
```diff
@@ -13,6 +13,7 @@ dependencies {
 
   implementation(project(":exporters:otlp:common"))
 
+  implementation("com.google.api.grpc:proto-google-common-protos")
```
anuraaga, line 16: Needing a dependency on a gRPC proto in the simple HTTP exporter seems like a bug in the spec to me

exporters/otlp/metrics/build.gradle.kts
```diff
@@ -20,6 +20,7 @@ dependencies {
 
   implementation(project(":exporters:otlp:common"))
 
+  api("io.grpc:grpc-stub")
```
anuraaga, line 23: We have public API that uses ManagedChannel

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/GrpcStatusUtil.java
```diff
@@ -0,0 +1,36 @@
...
+package io.opentelemetry.exporter.otlp.internal;
+
+import com.google.protobuf.CodedInputStream;
+import java.io.IOException;
+
+/** Utilities for working with gRPC status without requiring dependencies on gRPC. */
+public final class GrpcStatusUtil {
+
+  /** Parses the message out of a serialized gRPC Status. */
+  public static String getStatusMessage(byte[] serializedStatus) throws IOException {
```
anuraaga, line 15: Realized it's easy to parse manually to reduce dependency. The main reason for finding the issue in the first place was to evaluate feasibility of proto-lite, and a dependency on the google proto would block that.

exporters/otlp-http/trace/build.gradle.kts
jkwatson: ooh. interesting. yeah, I had forgotten this was the case. And, I agree with you, although I bet the otlp proto folks will fight hard not to have to re-implement the basic stuff (happy to be wrong).

jkwatson (approved):
good cleanup!

PR #3491, opened 2021-08-15
View API additons

jsuereth:
Step 1 of migration towards new view API:

- Set the default aggregator for Histogram instrument to be a Histogram (matching pending spec)
- Add Meter-based selection criteria to views (based off [this specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#view))
  - Add MeterSelector type w/ name, version + schema_url patterns
  - Update view registry for new logic.  
  
Notes:

- Override rules for view specificity unclear.  Kept previous behavior of last-registered-wins in the case of two applicable views.
- The (major) refactoring of view to *be* the metric output definition is still pending some refactoring of Aggregator.  Specifically Aggregator + Instrument Descriptor are still tightly coupled, including aggregators leveraging the instrument type.   This will be done in a follow up PR and is why we don't expose the `view name` or `view description` configuration in this PR.
- The (minor) refactoring of `LabelsProcessor` is also left to a future PR.  That refactoring will add the `attribute_keys` and `extra_dimensions` parameters.


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/ViewRegistry.java
```diff
@@ -55,17 +62,42 @@ public static ViewRegistryBuilder builder() {
    * @param descriptor description of the instrument.
    * @return The {@link View} for this instrument, or a default aggregation view.
    */
-  public View findView(InstrumentDescriptor descriptor) {
-    LinkedHashMap<Pattern, View> configPerType = configuration.get(descriptor.getType());
-    for (Map.Entry<Pattern, View> entry : configPerType.entrySet()) {
-      if (entry.getKey().matcher(descriptor.getName()).matches()) {
+  public View findView(InstrumentDescriptor descriptor, InstrumentationLibraryInfo meter) {
+    for (Map.Entry<InstrumentSelector, View> entry : configuration.entrySet()) {
```
anuraaga, line 66: If we don't call `get` might be simpler to just use a `List` or array instead of `LinkedHashMap`.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/ViewRegistry.java
```diff
@@ -55,17 +62,42 @@ public static ViewRegistryBuilder builder() {
...
+
+  // Matches a meter selector against a meter.
+  private static boolean matchesMeter(MeterSelector selector, InstrumentationLibraryInfo meter) {
+    return matchesPattern(selector.getNamePattern(), meter.getName())
+        && matchesPattern(selector.getVersionPattern(), meter.getVersion())
+        && matchesPattern(selector.getSchemaUrlPattern(), meter.getSchemaUrl());
+  }
+
+  // Matches a pattern against  a value.  Null values are treated as empty strings.
+  private static boolean matchesPattern(Pattern pattern, String value) {
```
anuraaga, line 94:
```suggestion
  private static boolean matchesPattern(Pattern pattern, @Nullable String value) {
```

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/ViewRegistry.java
```diff
@@ -55,17 +62,42 @@ public static ViewRegistryBuilder builder() {
...
+    if (value == null) {
+      return pattern.matcher("").matches();
```
anuraaga, line 96:
```suggestion
      value = "";
```

To not duplicate the logic

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/MeterSelector.java
```diff
@@ -0,0 +1,86 @@
...
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * Provides means for selecting one or more Meters. Used for selecting instruments when constructing
+ * views.
+ */
+@AutoValue
+@Immutable
+public abstract class MeterSelector {
+  private static final Pattern MATCH_ALL = Pattern.compile(".*");
```
anuraaga, line 20: Ideally Java did it for us but it doesn't... Performance for simple selectors would go up a lot if we define a `Matcher` and don't use regex for matchAll or exact matches.

Maybe inspiration

https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplier.java#L329

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/DoubleHistogramSdkTest.java
```diff
@@ -269,9 +262,10 @@ void stressTest_WithDifferentLabelSet() {
                                 .hasEpochNanos(testClock.now())
                                 .hasCount(4_000)
                                 .hasSum(40_000)
-                                .hasPercentileValues(
-                                    valueAtPercentiles(9.0, 11.0)
-                                        .toArray(new ValueAtPercentile[0])))
+                        // .hasPercentileValues(
```
anuraaga, line 265: Add a TODO

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/MeterSelector.java
jsuereth: Agreed.  I might use that directly against Predicate<String> for now.  It's almost exactly what you have but leveraging a more powerful `Predicate` class.  LMKWYT.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/DoubleHistogramSdkTest.java
jsuereth: This was just an oversight, fixing.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/ViewRegistry.java
jsuereth: Done

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/RegisteredView.java
```diff
@@ -0,0 +1,25 @@
...
+import javax.annotation.concurrent.Immutable;
+
+/** Internal representation of a {@link View} and {@link InstrumentSelector}. */
+@AutoValue
+@Immutable
+abstract class RegisteredView {
+  /** Instrument fitler for applying this view. */
+  public abstract InstrumentSelector getInstrumentSelector();
+  /** The view to apply. */
+  public abstract View getView();
```
jkwatson, line 20: s/public //

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/ViewRegistry.java
```diff
@@ -29,19 +29,22 @@
...
           .build();
-  static final View SUMMARY =
-      View.builder().setAggregatorFactory(AggregatorFactory.minMaxSumCount()).build();
+  static final View DEFAULT_HISTOGRAM =
+      View.builder()
+          .setAggregatorFactory(
+              AggregatorFactory.histogram(
+                  Arrays.asList(
+                      5d, 10d, 25d, 50d, 75d, 100d, 250d, 500d, 750d, 1_000d, 2_500d, 5_000d,
+                      7_500d, 10_000d),
```
jkwatson, line 38: It feels like this should be a constant somewhere.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/ViewRegistry.java
```diff
@@ -55,17 +58,34 @@ public static ViewRegistryBuilder builder() {
    * @param descriptor description of the instrument.
    * @return The {@link View} for this instrument, or a default aggregation view.
    */
-  public View findView(InstrumentDescriptor descriptor) {
-    LinkedHashMap<Pattern, View> configPerType = configuration.get(descriptor.getType());
-    for (Map.Entry<Pattern, View> entry : configPerType.entrySet()) {
-      if (entry.getKey().matcher(descriptor.getName()).matches()) {
-        return entry.getValue();
+  public View findView(InstrumentDescriptor descriptor, InstrumentationLibraryInfo meter) {
+    for (RegisteredView entry : reverseRegistration) {
```
jkwatson, line 62: I thought that Reilly's intent was to have the views applied with the first registered having priority. Any particular reason to reverse it here?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/InstrumentSelector.java
```diff
@@ -36,18 +39,37 @@ public static Builder builder() {
...
+
+    /** Sets the {@link Pattern} for instrument names that will be selected. */
+    public final Builder setInstrumentNamePattern(Pattern instrumentNamePattern) {
+      return setInstrumentNameFilter(StringPredicates.regex(instrumentNamePattern));
+    }
+
+    /** Sets the exact instrument name that will be selected. */
+    public final Builder setInstrumentName(String instrumentName) {
+      return setInstrumentNameFilter(StringPredicates.exact(instrumentName));
+    }
```
jkwatson, line 66: should probably have doc that states that these two are "last set wins" options.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/MeterSelector.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /** Sets the {@link Pattern} for matching schema urls. */
+    public final Builder setSchemaUrlPattern(Pattern pattern) {
+      return setSchemaUrlFilter(StringPredicates.regex(pattern));
+    }
+
+    /** Sets a specifier for selecting Instruments by name. */
+    public final Builder setSchemaUrl(String url) {
+      return setSchemaUrlFilter(StringPredicates.exact(url));
+    }
```
jkwatson, line 93: do we need both of these?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/MeterSelector.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.regex.Pattern;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * Provides means for selecting one or more Meters. Used for selecting instruments when constructing
+ * views.
+ */
+@AutoValue
+@Immutable
+public abstract class MeterSelector {
```
jkwatson, line 20: I don't see test coverage for this class anywhere.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/ViewRegistry.java
jsuereth: See comment on the PR.  Mostly I kept the reversal behavior so I didn't break existing tests, otherwise need to shuffle them.  Combine that with the Specification not *clearly* denoting order.  I can flip it though if we think that's the way the spec is going.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/MeterSelector.java
jsuereth: Switched to this from one of @anuraaga's comments on performance.   You can do exact matching w/ regex, but it's clunky.  `Pattern.compile(Pattern.quote(input))`.  I can switch back, it's not a huge deal either way

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/ViewRegistry.java
jkwatson: cool. no worries; I think this is something we can evolve, for sure.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/MeterSelector.java
jkwatson: ah, yeah. hmm. I don't think view lookup should be a hot-path thing, though, so I'm not so worried about performance, personally. Unless I'm mistaken that it might be on the users hot-path somewhere?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/StringPredicates.java
```diff
@@ -0,0 +1,32 @@
...
+package io.opentelemetry.sdk.metrics.internal.view;
+
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+
+/** Re-usable string predicates. */
+public final class StringPredicates {
+  private StringPredicates() {}
+
+  /** A string predicaate that matches all strings. */
```
jack-berg, line 15:
```suggestion
  /** A string predicate that matches all strings. */
```

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/StringPredicates.java
```diff
@@ -0,0 +1,32 @@
...
+  public static final Predicate<String> ALL = value -> true;
+
+  /** A string predicate that does exact string matching. */
+  public static Predicate<String> exact(String match) {
+    return input -> (input != null) && match.equals(input);
```
jack-berg, line 20: Intellij tells me the `(input != null)` check is unnecessary before `match.equals(`.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/InstrumentSelector.java
```diff
@@ -36,18 +39,37 @@ public static Builder builder() {
...
+
+    /**
+     * Sets the {@link MeterSelector} for which {@link io.opentelemetry.api.metrics.View}s will be
```
jack-berg, line 69:
```suggestion
     * Sets the {@link MeterSelector} for which {@link io.opentelemetry.sdk.metrics.view.View}s will be
```

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/InstrumentSelector.java
```diff
@@ -36,18 +39,37 @@ public static Builder builder() {
...
+     * included.
+     */
+    abstract Builder setMeterSelector(MeterSelector meterSelector);
```
jack-berg, line 72: I believe this needs to be public.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/view/ViewRegistryTest.java
```diff
@@ -16,6 +17,10 @@
 import org.junit.jupiter.api.Test;
 
 class ViewRegistryTest {
+
+  private static final InstrumentationLibraryInfo METER =
```
jack-berg, line 21: #nit
```suggestion
  private static final InstrumentationLibraryInfo INSTRUMENTATION_LIBRARY_INFO =
```

jack-berg (commented):
Noticed a few small things

anuraaga (approved):
LGTM with the others' comments

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/ViewRegistry.java
jsuereth: Personally, I think the way java had it leads to a much more flexible configuration model where "last added wins", if we combine this with a clear mechanism where configuration is loaded and in what order.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/InstrumentSelector.java
jsuereth: added

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/StringPredicates.java
jsuereth: removed

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/InstrumentSelector.java
jsuereth: That was a huge typo, should have been "Meter", not "View". Fixed.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/MeterSelector.java
jsuereth: Added

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/view/ViewRegistryTest.java
jsuereth: fixed

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/ViewRegistry.java
jsuereth: Move it locally for now.  Will shift as Aggregator re-work occurs.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/MeterSelector.java
jsuereth: It *could* be on the hot-path.  Instruments can be pulled off meters at any point in the lifecycle of the meter.  Depending on instrumentation, this could be as frequent as once-per-request.  While we recommend storing these somewhere more static / longer-lived, I think the predicate-based code (once I saw it) will be easier to maintain/extend.

If it's all the same, I'd like to use the predicate solution here (and elsewhere). Happy to remove some public methods if that simplifies things.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/InstrumentSelector.java
jsuereth: Fixed!

PR #3493, opened 2021-08-16
Test TraceMarshaler with random SpanData.

anuraaga:
And fix TraceMarshaler marshaling of schemaUrls.

This generates 100 different SpanData every CI run. Though I've ran through 100000 locally with no error.

It did solve the most important issue of automatically detecting schemaUrl had drifted. I think this gives good confidence and we can go ahead and switch the marshaler used by the exporters.

### DIFF
--- a/exporters/otlp/trace/src/test/java/io/opentelemetry/exporter/otlp/trace/SpanDataRandomizerRegistry.java
+++ b/exporters/otlp/trace/src/test/java/io/opentelemetry/exporter/otlp/trace/SpanDataRandomizerRegistry.java
@@ -276,6 +276,10 @@ public final class SpanDataRandomizerRegistry implements RandomizerRegistry {
   private final class SpanContextRandomizer implements Randomizer<SpanContext> {
     @Override
     public SpanContext getRandomValue() {
+      boolean isInvalid = booleanRandomizer.getRandomValue();
+      if (isInvalid) {
+        return SpanContext.getInvalid();
+      }
       String traceId = randomHex(TraceId.getLength());
       String spanId = randomHex(SpanId.getLength());
       TraceFlags traceFlags = TraceFlags.fromByte(byteRandomizer.getRandomValue());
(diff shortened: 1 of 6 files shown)


exporters/otlp/trace/src/test/java/io/opentelemetry/exporter/otlp/trace/SpanDataRandomizerRegistry.java
```diff
@@ -0,0 +1,345 @@
...
+    statusDataRandomizer =
+        new AutoValueRandomizer<>("io.opentelemetry.sdk.trace.data.AutoValue_ImmutableStatusData");
+    traceStateRandomizer = new TraceStateRandomizer();
+  }
+
+  @Override
+  @Nullable
+  public Randomizer<?> getRandomizer(Field field) {
+    // TODO(anuraaga): Make work for autovalue, unfortunately it only adds Nullable to constructor
+    // parameters / getters but not to fields.
```
jkwatson, line 165: It would be good to create an issue for this, so we don't lose track of it.

exporters/otlp/trace/src/test/java/io/opentelemetry/exporter/otlp/trace/TraceMarshalerTest.java
```diff
@@ -125,6 +140,39 @@ void marshalAndSizeRequest_InstrumentationLibrary() throws IOException {
             testSpanDataWithInstrumentationLibrary(InstrumentationLibraryInfo.create("", ""))));
   }
 
+  @RepeatedTest(10000)
```
jkwatson, line 143: do we want 10,000 for CI? Any rough guess how long this takes to run?

jkwatson:
This is a great approach! Thanks for digging in to this. I fear metrics will be considerably trickier, but at least getting traces converted over is an excellent move!

jkwatson (approved):
Thanks!

exporters/otlp/trace/src/test/java/io/opentelemetry/exporter/otlp/trace/TraceMarshalerTest.java
anuraaga: Oops, was drift back to 100

PR #3499, opened 2021-08-16
Fix javadoc

jack-berg:
Went to add javadoc links to the new OTLP http modules, and noticed that a bunch of broken javadoc links. Went through and fixed all the ones I found.

Separately, there are a bunch of readmes that are missing altogether, or are missing javadoc links. Figure I'll add those in a separate PR.


jkwatson (approved):
Thanks!!!

PR #3502, opened 2021-08-17
Optimize 1 & 2 element attributes creation

### LINKED ISSUE #2765: Improve performance of Attributes.of(myKey, "value")

jkwatson:
resolves #2765 

This handles `of(k,v)`, `of(k1,v1,k2,v2)` and a single element AttributesBuilder.

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/common/ArrayBackedAttributes.java
+++ b/api/all/src/main/java/io/opentelemetry/api/common/ArrayBackedAttributes.java
@@ -26,6 +26,14 @@ final class ArrayBackedAttributes extends ImmutableKeyValuePairs<AttributeKey<?>
     super(data, keyComparator);
   }
 
+  /**
+   * Only use this constructor if you can guarantee that the data has been de-duped and sorted.
+   * @param data the raw data
+   */
+  ArrayBackedAttributes(Object[] data) {
+    super(data, KEY_COMPARATOR_FOR_CONSTRUCTION);
+  }
+
   @Override
   public AttributesBuilder toBuilder() {
     return new ArrayBackedAttributesBuilder(new ArrayList<>(data()));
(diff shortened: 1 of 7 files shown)


api/all/src/main/java/io/opentelemetry/api/common/ArrayBackedAttributes.java
```diff
@@ -26,6 +26,16 @@ private ArrayBackedAttributes(Object[] data, Comparator<AttributeKey<?>> keyComp
...
   }
 
+  /**
+   * Only use this constructor if you can guarantee that the data has been de-duped, sorted by key
+   * and contains no null values or null/empty keys.
+   *
+   * @param data the raw data
+   */
+  ArrayBackedAttributes(Object[] data) {
+    super(data, KEY_COMPARATOR_FOR_CONSTRUCTION);
```
anuraaga, line 36: Doesn't `super` still call `sortAndFilter` for our presorted ones? We could probably have the others explicitly pass in `Comparator.naturalOrder` and implement the one-arg super constructor as a simple field set

jkwatson: oh drat. You are totally correct. Yes, I will refactor this a bit.

jkwatson: fixed. please take a look.

anuraaga (approved):

PR #3503, opened 2021-08-18
Plugin the custom marshaler to otlp:trace span exporter.

anuraaga:
This is almost exactly like @bogdandrutu's #2890 though I made a few tweaks to avoid any dependency on the generated gRPC code to in the future allow us to remove dependency on the generated proto completely.

### DIFF
--- a/exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/MarshalerInputStream.java
+++ b/exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/MarshalerInputStream.java
@@ -15,7 +15,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import javax.annotation.Nullable;
 
-// Adapated from gRPC ProtoInputStream but using our Marshaller
+// Adapted from gRPC ProtoInputStream but using our Marshaller
 // https://github.com/grpc/grpc-java/blob/2c2ebaebd5a93acec92fbd2708faac582db99371/protobuf-lite/src/main/java/io/grpc/protobuf/lite/ProtoInputStream.java
 final class MarshalerInputStream extends InputStream implements Drainable, KnownLength {
   @Nullable private Marshaler message;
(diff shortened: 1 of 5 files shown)


exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/MarshalerInputStream.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  @Override
+  public int drainTo(OutputStream target) throws IOException {
+    int written;
+    if (message != null) {
+      written = message.getSerializedSize();
+      CodedOutputStream cos = CodedOutputStream.newInstance(target);
+      message.writeTo(cos);
+      cos.flush();
```
anuraaga, line 35: For better or worse we have more test coverage since the days of #2890 - the Armeria-based tests that actually exercise Netty showed this missing flush from the old PR. I point it out since it means even after the random SpanData generation for verifying marshaling behavior, there are still points that could cause issues. But I think with the tests in this repo and eventually the java instrumentation smoke tests, we'll be fine and have to make the final leap.

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/MarshalerInputStream.java
```diff
@@ -0,0 +1,103 @@
...
+import com.google.protobuf.CodedOutputStream;
+import io.grpc.Drainable;
+import io.grpc.KnownLength;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import javax.annotation.Nullable;
+
+// Adapated from gRPC ProtoInputStream but using our Marshaller
```
jkwatson, line 18:
```suggestion
// Adapted from gRPC ProtoInputStream but using our Marshaller
```

jkwatson (approved):
🤞🏽

exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/MarshalerInputStream.java
```diff
@@ -0,0 +1,103 @@
...
+  MarshalerInputStream(Marshaler message) {
+    this.message = message;
+  }
+
+  @Override
+  public int drainTo(OutputStream target) throws IOException {
+    int written;
+    if (message != null) {
+      written = message.getSerializedSize();
+      CodedOutputStream cos = CodedOutputStream.newInstance(target);
```
jsuereth, line 33: FYI - There can be a huge performance swing in creating a CodedOutputStream without passing that serialized size in as the second parameter.  If you haven't benchmarked the difference, highly recommend you try it out.  I saw a large swing in results otherwise due to buffer resizing.

PR #3504, opened 2021-08-18
Add a custom proto compiler handler that generates classes with only …

anuraaga:
…proto field numbers / enum values.

Depends on #3503 but figured I'd send it out for initial review, it's mostly just some Kotlin code to bridge proto files to javapoet invocations.


exporters/otlp/trace/src/jmh/java/io/opentelemetry/exporter/otlp/trace/RequestMarshalBenchmarks.java
```diff
@@ -70,9 +71,18 @@
...
     protoRequest.writeTo(CodedOutputStream.newInstance(protoOutput));
     return protoOutput;
   }
+
+  private static ExportTraceServiceRequest toRequest(TraceMarshaler.RequestMarshaler request)
+      throws IOException {
+    byte[] buf = new byte[request.getSerializedSize()];
+    request.writeTo(CodedOutputStream.newInstance(buf));
+    return ExportTraceServiceRequest.newBuilder()
+        .setUnknownFields(UnknownFieldSet.newBuilder().mergeFrom(buf).build())
```
jsuereth, line 85: Wow, what's the performance boost for this?

anuraaga: Bogdan originally wrote these benchmarks, but I think it's to compare serializing without having to vendor in the entire `MarshalerInputStream` / stub code you saw. But in the end going with the vendored in code had a significant improvement so we went with it (don't have the numbers on hand)

jkwatson:
So, with this change, our published (alpha) `opentelemetry-proto` will become effectively unusable as a standalone artifact. Is that correct, or am I misunderstanding what's going on here?

I don't think this project should be in the business of publishing proto bindings, so personally I think this is fine, but I believe there are folks at AWS who are currently using our proto bindings for something. Am I remembering that correctly, @anuraaga?

anuraaga:
Nah, opentelemetry-proto is untouched - just that we don't use it anymore in our OTLP exporters with regards to SpanData (once everything is switched over to marshaler). The plugin here is just customizing the output that is directly embedded into the exporter artifacts as an internal package, not the proto artifact itself.

I think there is a Java OTLP ingestion API using the proto bindings we publish - it's simple for us to keep it so not planning on removing them.

jkwatson:
> I think there is a Java OTLP ingestion API using the proto bindings we publish - it's simple for us to keep it so not planning on removing them.

It's also used in the agent testing code to verify the data being sent is correct, so glad it's untouched.

jkwatson (approved):

PR #3507, opened 2021-08-19
Compute buffer size for grpc marshaler

anuraaga:
This copies the logic from `AbstractMessageLite` in the proto library. We may look into potential optimization on top of it but this is good to baseline with protobuf.

Thanks for the callout @jsuereth


jkwatson (approved):
@jsuereth look good to you?

anuraaga:
Let me go ahead and merge this in since it's the same as proto library. I have a followup coming that should make this codepath a corner case.

PR #3510, opened 2021-08-19
Update View-wiring to match SDK specificaiton

jsuereth:
- Look at views in order of registration
- Allow views to change name/description of instruments in addition to aggregation

Future work for this to be good for users:

- Do not crash (by default) when a metric name conflict is detected, instead issue an error.
-  Allow multiple views per-instrument as long as there are no name conflicts
-  Improve error message + provenance of metric-name clashes.

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/descriptor/MetricDescriptor.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/descriptor/MetricDescriptor.java
@@ -7,6 +7,8 @@ package io.opentelemetry.sdk.metrics.internal.descriptor;
 
 import com.google.auto.value.AutoValue;
 import com.google.auto.value.extension.memoized.Memoized;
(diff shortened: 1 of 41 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/descriptor/MetricDescriptor.java
```diff
@@ -25,6 +27,14 @@ public static MetricDescriptor create(String name, String description, String un
     return new AutoValue_MetricDescriptor(name, description, unit);
   }
 
+  /** Constructs a metric descriptor for a given View + instrument. */
+  public static MetricDescriptor create(View view, InstrumentDescriptor instrument) {
+    final String name = (view.getName() == null) ? instrument.getName() : view.getName();
+    final String description =
```
jkwatson, line 33: Nit: we don't generally make our local variables `final` unless there's a real reason for it.

jkwatson (approved):
Looks like a good step. I like the separation of concerns between the InstrumentDescriptor and the MetricDescriptor.

PR #3519, opened 2021-08-20
Vendor in required bits of CodedOutputStream and remove protobuf-java…

anuraaga:
… dependency

Because we only use the version of CodedOutputStream that writes to an OutputStream in both gRPC and okhttp marshalers, this only vendors in that version. We end up with about 1300 loc for the coder - by saying goodbye to the 700KB protobuflite or 1.5MB protobuf libraries, I think this is worth it as binary size will not be a reason to avoid the proto serializer. Checking the class files locally, it's about 23KB of uncompressed code. This is a lot smaller than jackson-core too for reference which would likely be our dependency for a JSON encoding approach. The code is 99.9% copied as-is, just with much deletion, and one line with a boolean check on unsafe removed.

@jsuereth called it that removing the buffering slowed things down (the approach I took was to remove the `AbstractBuffered` and replacing writes to the buffer with `out.write`). So didn't write anything custom, the code is as-is. Will follow up with some ideas for actually making changes to improve the performance, for example threadlocal reused CodedOutputStream and checking the varint encoding.

I removed support for Unsafe to reduce the amount of code vendored and because it performs significantly better without it :O I guess Unsafe is really for the byte buffers, byte arrays are very fast on the JVM already.

After
```
Benchmark                                                                         (numSpans)  Mode  Cnt      Score      Error   Units
MarshalerInputStreamBenchmarks.marshalToByteArray                                         16  avgt   10     13.004 ±    0.346   us/op
MarshalerInputStreamBenchmarks.marshalToByteArray:·gc.alloc.rate                          16  avgt   10   1432.026 ±   36.460  MB/sec
MarshalerInputStreamBenchmarks.marshalToByteArray:·gc.alloc.rate.norm                     16  avgt   10  29304.005 ±    0.001    B/op
MarshalerInputStreamBenchmarks.marshalToByteArray:·gc.churn.G1_Eden_Space                 16  avgt   10   1434.033 ±   78.589  MB/sec
MarshalerInputStreamBenchmarks.marshalToByteArray:·gc.churn.G1_Eden_Space.norm            16  avgt   10  29341.737 ± 1246.071    B/op
MarshalerInputStreamBenchmarks.marshalToByteArray:·gc.churn.G1_Old_Gen                    16  avgt   10      0.008 ±    0.007  MB/sec
MarshalerInputStreamBenchmarks.marshalToByteArray:·gc.churn.G1_Old_Gen.norm               16  avgt   10      0.171 ±    0.147    B/op
[...]

PR #3520, opened 2021-08-20
Make proto dependency in otlp-common compileOnly.

anuraaga:
This is a bit of a mess but will get cleaned up after 1) migrating HTTP exporter to TraceMarshaler (coming soon after #3519) and 2) Adding a MetricMarshaler (coming eventually). On the bright side, forgetting to add `:proto` where it's needed results in a quick compile failure.

### DIFF
--- a/exporters/otlp/common/build.gradle.kts
+++ b/exporters/otlp/common/build.gradle.kts
@@ -11,17 +11,24 @@ otelJava.moduleName.set("io.opentelemetry.exporter.otlp.internal")
 
 dependencies {
   api(project(":api:all"))
-  api(project(":proto"))
   api(project(":sdk:all"))
   api(project(":sdk:metrics"))
 
+  // We only use the protos for HTTP, logging-otlp, and metric export for now. Let them expose the
+  // proto dependency in their POMs instead of here. This is fine because this artifact only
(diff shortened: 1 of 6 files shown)


exporters/otlp/common/build.gradle.kts
```diff
@@ -11,17 +11,24 @@ otelJava.moduleName.set("io.opentelemetry.exporter.otlp.internal")
...
   api(project(":sdk:metrics"))
 
+  // We only use the protos for HTTP, logging-otlp, and metric export for now. Let them expose the
+  // proto dependency in their POMs instead of here. This is fine because this artifact only
+  // contains internal code.
+  compileOnly(project(":proto"))
+
   implementation("com.google.protobuf:protobuf-java")
 
+  // Similar to above note about :proto, we include a helper shared by the gRPC exporters but do not
```
anuraaga, line 24: I suppose what we really need for these is `exporter-grpc-common`, which would also be used from the jaeger exporter. It is yet another artifact, but "makes sense" and would allow Jaeger TLS support to be functioning (presumably no one uses it with TLS right now since without customizing certs it's rare that it could function)

jkwatson (approved):

PR #3522, opened 2021-08-20
Add OTLP protocol configurability to autoconfigure

jack-berg:
Related to this [open spec issue](https://github.com/open-telemetry/opentelemetry-specification/issues/1879).

A note on the implementation: I was getting annoyed with how repetitive the configuration was for OTLP exporters. The extraction of config properties for OTLP grpc traces / metrics exporters is mostly the same. Adding support for protocol configurability of protocol means more copy / pasta or adding some abstraction. Went with adding an abstraction. 

I initially added a `OtlpExporterBuilder` interface located in `:exporters:otlp:common` with methods common across all the OTLP exporters, but ran into all sorts of `NoClassDefFoundError` errors I couldn't get past. The solution I landed is to have a method that accepts a bunch of consumers which are invoked after config properties are extracted. Its pretty clean IMO, with one downside being a method that accepts 6 arguments.

### DIFF
--- a/sdk-extensions/autoconfigure/README.md
+++ b/sdk-extensions/autoconfigure/README.md
@@ -62,6 +62,9 @@ The [OpenTelemetry Protocol (OTLP)](https://github.com/open-telemetry/openteleme
 | otel.exporter.otlp.timeout   | OTEL_EXPORTER_OTLP_TIMEOUT  | The maximum waiting time, in milliseconds, allowed to send each OTLP trace and metric batch. Default is `10000`.  |
 | otel.exporter.otlp.traces.timeout   | OTEL_EXPORTER_OTLP_TRACES_TIMEOUT  | The maximum waiting time, in milliseconds, allowed to send each OTLP trace batch. Default is `10000`.  |
 | otel.exporter.otlp.metrics.timeout   | OTEL_EXPORTER_OTLP_METRICS_TIMEOUT  | The maximum waiting time, in milliseconds, allowed to send each OTLP metric batch. Default is `10000`.  |
+| otel.experimental.exporter.otlp.protocol | OTEL_EXPERIMENTAL_EXPORTER_OTLP_PROTOCOL | The transport protocol to use on OTLP trace and metrics requests. Options include `grpc` and `http/protobuf`. Default is `grpc`. |
+| otel.experimental.exporter.otlp.traces.protocol | OTEL_EXPERIMENTAL_EXPORTER_OTLP_TRACES_PROTOCOL | The transport protocol to use on OTLP trace requests. Options include `grpc` and `http/protobuf`. Default is `grpc`. |
+| otel.experimental.exporter.otlp.metrics.protocol | OTEL_EXPERIMENTAL_EXPORTER_OTLP_METRICS_PROTOCOL | The transport protocol to use on OTLP metrics requests. Options include `grpc` and `http/protobuf`. Default is `grpc`. |
 
 To configure the service name for the OTLP exporter, add the `service.name` key
 to the OpenTelemetry Resource ([see below](#opentelemetry-resource)), e.g. `OTEL_RESOURCE_ATTRIBUTES=service.name=myservice`.
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
@@ -8,6 +8,8 @@ package io.opentelemetry.sdk.autoconfigure;
 import static io.opentelemetry.sdk.autoconfigure.OtlpConfigUtil.DATA_TYPE_METRICS;
 
 import io.opentelemetry.exporter.logging.LoggingMetricExporter;
+import io.opentelemetry.exporter.otlp.http.metric.OtlpHttpMetricExporter;
+import io.opentelemetry.exporter.otlp.http.metric.OtlpHttpMetricExporterBuilder;
 import io.opentelemetry.exporter.otlp.metrics.OtlpGrpcMetricExporter;
 import io.opentelemetry.exporter.otlp.metrics.OtlpGrpcMetricExporterBuilder;
 import io.opentelemetry.exporter.prometheus.PrometheusCollector;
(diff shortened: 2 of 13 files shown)


sdk-extensions/autoconfigure/README.md
```diff
@@ -62,6 +62,9 @@ The [OpenTelemetry Protocol (OTLP)](https://github.com/open-telemetry/openteleme
 | otel.exporter.otlp.timeout   | OTEL_EXPORTER_OTLP_TIMEOUT  | The maximum waiting time, in milliseconds, allowed to send each OTLP trace and metric batch. Default is `10000`.  |
 | otel.exporter.otlp.traces.timeout   | OTEL_EXPORTER_OTLP_TRACES_TIMEOUT  | The maximum waiting time, in milliseconds, allowed to send each OTLP trace batch. Default is `10000`.  |
 | otel.exporter.otlp.metrics.timeout   | OTEL_EXPORTER_OTLP_METRICS_TIMEOUT  | The maximum waiting time, in milliseconds, allowed to send each OTLP metric batch. Default is `10000`.  |
+| otel.exporter.otlp.protocol | OTEL_EXPORTER_OTLP_PROTOCOL | The transport protocol to use on OTLP trace and metrics requests. Options include `grpc` and `http/protobuf`. Default is `grpc`. |
+| otel.exporter.otlp.traces.protocol | OTEL_EXPORTER_OTLP_TRACES_PROTOCOL | The transport protocol to use on OTLP trace requests. Options include `grpc` and `http/protobuf`. Default is `grpc`. |
+| otel.exporter.otlp.metrics.protocol | OTEL_EXPORTER_OTLP_METRICS_PROTOCOL | The transport protocol to use on OTLP metrics requests. Options include `grpc` and `http/protobuf`. Default is `grpc`. |
```
jkwatson, line 67: I chatted with @anuraaga about this, and because the spec hasn't been approved yet for these values, that we should use the `OTEL_EXPERIMENTAL_EXPORTER_` name prefix for these until the spec gets approved (or rejected!).

jkwatson:
Thanks for this @jack-berg 

Can you separate this into 2 PRs: 1 for the new experimental env vars, and 1 for the refactoring?

jack-berg:
> Can you separate this into 2 PRs: 1 for the new experimental env vars, and 1 for the refactoring?

Sure. Just to clarify, the preference is to add the new experimental env vars, with the repetitious reading of config properties in one PR. Then in a followup PR, abstract away the repetitious reading of config properties.

jkwatson:
> > Can you separate this into 2 PRs: 1 for the new experimental env vars, and 1 for the refactoring?
> 
> Sure. Just to clarify, the preference is to add the new experimental env vars, with the repetitious reading of config properties in one PR. Then in a followup PR, abstract away the repetitious reading of config properties.

Or, in the other order...whichever works best for you. :)

jack-berg:
I've rebased onto main which has the refactor, and added the `experimental` prefix!

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtil.java
```diff
@@ -19,6 +19,18 @@
   static final String DATA_TYPE_TRACES = "traces";
   static final String DATA_TYPE_METRICS = "metrics";
 
+  static String getOtlpProtocol(String dataType, ConfigProperties config) {
+    String protocol =
+        config.getString(String.format("otel.experimental.exporter.otlp.%s.protocol", dataType));
```
anuraaga, line 24: Nit: let's just use string concatenation instead of `String.format` - here, indeed the performance implication doesn't really matter but suggest just always using string concatenation where possible to avoid the cognitive load of reasoning out the performance implication. The readability penalty is debatable anyways, i.e. concatenation looks closer to modern language's interpolation syntax

anuraaga (approved):

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtil.java
```diff
@@ -19,6 +19,18 @@
...
+    if (protocol == null) {
+      protocol = config.getString("otel.experimental.exporter.otlp.protocol");
+    }
+    if (protocol == null) {
+      protocol = "grpc";
+    }
+    return protocol;
```
kubawach, line 31: Totally NIT: `return (protocol == null ? grpc : protocol)` - as an idiom for "default value"

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/SpanExporterConfiguration.java
```diff
@@ -38,7 +40,7 @@ static SpanExporter configureExporter(String name, ConfigProperties config) {
 
     switch (name) {
       case "otlp":
-        return configureOtlpSpans(config);
+        return configureOtlp(config);
```
kubawach, line 43: 👍

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
```diff
@@ -77,29 +79,58 @@ private static void configureLoggingMetrics(
...
+      OtlpHttpMetricExporterBuilder builder = OtlpHttpMetricExporter.builder();
 
-    OtlpConfigUtil.configureOtlpExporterBuilder(
-        DATA_TYPE_METRICS,
-        config,
-        builder::setEndpoint,
-        builder::addHeader,
-        builder::setTimeout,
-        builder::setTrustedCertificates);
+      OtlpConfigUtil.configureOtlpExporterBuilder(
```
kubawach, line 99: The `OtlpConfigUtil` calls here and in `SpanExporterConfiguration` make me wish for a Builder interface that would contain required 4 methods and allow us to remove the duplication. But this is out of scope of this change, so entirely up to you.

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtilTest.java
```diff
@@ -0,0 +1,42 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.google.common.collect.ImmutableMap;
+import java.util.Collections;
+import org.junit.jupiter.api.Test;
+
+class OtlpConfigUtilTest {
+
+  @Test
+  void getOtlpProtocolDefault() {
```
kubawach, line 18: 👍

kubawach (approved):
LGTM :)

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
jack-berg: I actually went down that road first and added the builder interface to `:exporters:otlp:common`. The problem was that I was stuck getting `NoClassDefFoundError` at runtime when the `OtlpExporterBuilder` interface is referenced. The pattern in the autoconfigure module of using `ClasspathUtil.checkClassExists(...)` seems to be deterministic at runtime in that java doesn't try to load the class until after the `ClasspathUtil.checkClassExists` call. But for some reason I couldn't figure out, that behavior wasn't the same with `OtlpExporterBuilder`.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtil.java
jack-berg: I'm partial to the readability of `String.format`, but am more in favor of having / following a convention on these sorts of things.

jkwatson (approved):

PR #3523, opened 2021-08-20
Remove outdated docs in otlp package-info.java files

jack-berg:
Noticed that the `package-info.java` files in `:exporters:otlp:metrics` and `:exporters:otlp:trace` have some information that appears to be a remnant of a time when the exporters were configurable via `ConfigProperties`.


jkwatson (approved):
Thanks! I had forgotten this cruft was still hanging around.

PR #3524, opened 2021-08-20
Update the semantic conventions to 1.6.0

jkwatson:
I had to tweak the generator template a bit to account for some new constants that didn't map to java constant names.

### DIFF
--- a/buildscripts/semantic-convention/templates/SemanticAttributes.java.j2
+++ b/buildscripts/semantic-convention/templates/SemanticAttributes.java.j2
@@ -89,7 +89,7 @@ public final class {{class}} {
   public static final class {{class_name}} {
     {%- for member in attribute.attr_type.members %}
       /** {% filter escape %}{{member.brief | to_doc_brief}}.{% endfilter %} */
-      public static final {{ type }} {{ member.member_id | to_const_name }} = {{ print_value(type, member.value) }};
+      public static final {{ type }} {{ member.member_id | to_const_name | regex_replace(" ", "_") | regex_replace("^([0-9])","_$1") }} = {{ print_value(type, member.value) }};
     {%- endfor %}
       private {{ class_name }}() {}
   }
--- a/buildscripts/semantic-convention/generate.sh
+++ b/buildscripts/semantic-convention/generate.sh
@@ -4,7 +4,7 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 ROOT_DIR="${SCRIPT_DIR}/../../"
 
 # freeze the spec & generator tools versions to make SemanticAttributes generation reproducible
-SEMCONV_VERSION=1.5.0
+SEMCONV_VERSION=1.6.0
 SPEC_VERSION=v$SEMCONV_VERSION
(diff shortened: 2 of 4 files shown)


jack-berg (approved):

Oberon00:
> I had to tweak the generator template a bit to account for some new constants that didn't map to java constant names.

This was fixed in https://github.com/open-telemetry/opentelemetry-specification/pull/1863

jkwatson:
> > I had to tweak the generator template a bit to account for some new constants that didn't map to java constant names.
> 
> This was fixed in [open-telemetry/opentelemetry-specification#1863](https://github.com/open-telemetry/opentelemetry-specification/pull/1863)

Hmm. do you think we should pass on the 1.6.0 release, then, and wait for 1.7.0 to update the semconv?

Oberon00:
I have no opinion there. Maybe you could even cherry-pick that particular commit in the generator script on top of 1.6. Also, it seems there is a instrumentation SIG + semantic convention WG set up, so there might be more changes ahead in the future.

* https://github.com/open-telemetry/community/issues/802
* https://github.com/open-telemetry/community/issues/803
* https://github.com/open-telemetry/community/issues/805

buildscripts/semantic-convention/templates/SemanticAttributes.java.j2
```diff
@@ -89,7 +89,7 @@ public final class {{class}} {
   public static final class {{class_name}} {
     {%- for member in attribute.attr_type.members %}
       /** {% filter escape %}{{member.brief | to_doc_brief}}.{% endfilter %} */
-      public static final {{ type }} {{ member.member_id | to_const_name }} = {{ print_value(type, member.value) }};
+      public static final {{ type }} {{ member.member_id | to_const_name | regex_replace(" ", "_") | regex_replace("^([0-9])","_$1") }} = {{ print_value(type, member.value) }};
```
anuraaga, line 92: Let's just add file an issue to revert this in 1.7.0

anuraaga (approved):

buildscripts/semantic-convention/templates/SemanticAttributes.java.j2
jkwatson: 1.7.0 is going to break a bunch of stuff that went in in 1.6.0, due to this stuff. That's why I was wondering if we should skip 1.6.0. I'm fine either way.

anuraaga: Oh I read the linked issue and get it better now - so the `NetHostConnectionSubtypeValues` would change from what I understand. Hrm - I guess we can still go for it, it's what we have the schema URLs for for better or worse

jkwatson: I wonder if they'll create a schema for 1.7.0 that maps between the changes. And, how that will interact with the fact that we're having to actually change the keys here, since they aren't java compliant constant values.

Oberon00: I think a schema can only handle the values.

jkwatson: I guess that makes sense, yes. 👍🏽

PR #3525, opened 2021-08-20
Extract common OTLP config logic to utility function

jack-berg:
The refactor portion of Issue #3522.


jkwatson (approved):
Look great! Thanks!

PR #3526, opened 2021-08-20
Add missing API diff.

jack-berg:
A change was made to `:extensions:kotlin` without running the apidiff script.


jkwatson (approved):
Thanks! This particular diff is not amazingly useful, since japicmp doesn't grok kotlin, but we should keep it up to date anyway.

jkwatson:
> Thanks! This particular diff is not amazingly useful, since japicmp doesn't grok kotlin, but we should keep it up to date anyway.

Makes me wonder if we should disable jampcmp tasks for the kotlin module

anuraaga:
> Makes me wonder if we should disable jampcmp tasks for the kotlin module

Think we should I don't think we even changed anything

PR #3529, opened 2021-08-20
Support comma separated list for exporters for otel.traces.exporter.

jack-berg:
Resolves Issue #3443 .

### DIFF
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/SpanExporterConfiguration.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/SpanExporterConfiguration.java
@@ -6,6 +6,9 @@
 package io.opentelemetry.sdk.autoconfigure;
 
 import static io.opentelemetry.sdk.autoconfigure.OtlpConfigUtil.DATA_TYPE_TRACES;
+import static java.util.stream.Collectors.counting;
+import static java.util.stream.Collectors.groupingBy;
+import static java.util.stream.Collectors.joining;
 import static java.util.stream.Collectors.toMap;
 
 import io.opentelemetry.exporter.jaeger.JaegerGrpcSpanExporter;
@@ -36,19 +39,26 @@ final class SpanExporterConfiguration {
     List<String> exporterNamesList = config.getCommaSeparatedValues("otel.traces.exporter");
     Set<String> exporterNames = new HashSet<>(exporterNamesList);
     if (exporterNamesList.size() != exporterNames.size()) {
-      throw new ConfigurationException("otel.traces.exporter contains duplicates");
+      String duplicates =
+          exporterNamesList.stream()
+              .collect(groupingBy(Function.identity(), counting()))
+              .entrySet()
+              .stream()
+              .filter(entry -> entry.getValue() > 1)
+              .map(Map.Entry::getKey)
+              .collect(joining(",", "[", "]"));
+      throw new ConfigurationException("otel.traces.exporter contains duplicates: " + duplicates);
     }
-    if (exporterNames.contains(EXPORTER_NONE) && exporterNames.size() > 1) {
-      throw new ConfigurationException(
-          "otel.traces.exporter contains " + EXPORTER_NONE + " along with other exporters");
+    if (exporterNames.contains(EXPORTER_NONE)) {
+      if (exporterNames.size() > 1) {
+        throw new ConfigurationException(
+            "otel.traces.exporter contains " + EXPORTER_NONE + " along with other exporters");
+      }
+      return Collections.emptyMap();
     }
 
     if (exporterNames.isEmpty()) {
       exporterNames = Collections.singleton("otlp");
-    }
-
-    if (exporterNames.contains(EXPORTER_NONE)) {
-      return Collections.emptyMap();
     }
 
     Map<String, SpanExporter> spiExporters =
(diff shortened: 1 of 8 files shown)


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/SpanExporterConfiguration.java
```diff
@@ -17,25 +18,58 @@
...
+    if (exporterNames.contains(EXPORTER_NONE) && exporterNames.size() > 1) {
+      throw new ConfigurationException(
+          "otel.traces.exporter contains " + EXPORTER_NONE + " along with other exporters");
+    }
+
+    if (exporterNames.isEmpty()) {
+      exporterNames = Collections.singleton("otlp");
+    }
+
+    if (exporterNames.contains(EXPORTER_NONE)) {
```
anuraaga, line 50: Nit, you can restructure the check on line 41 instead of checking again

```java
if (exporterNames.contains(EXPORTER_NONE)) {
  if (exporterNames.size() > 1) {
    throw
  }
  return emptyMap
}
```

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java
```diff
@@ -44,32 +47,38 @@ static SdkTracerProvider configureTracerProvider(Resource resource, ConfigProper
...
-  static SpanProcessor configureSpanProcessor(
-      ConfigProperties config, SpanExporter exporter, String exporterName) {
-    if (exporterName.equals("logging")) {
-      return SimpleSpanProcessor.create(exporter);
+  static List<SpanProcessor> configureSpanProcessors(
+      ConfigProperties config, Map<String, SpanExporter> exportersByName) {
+    Map<String, SpanExporter> exportersByNameCopy = new HashMap<>(exportersByName);
+    List<SpanProcessor> spanProcessors = new ArrayList<>();
+
+    if (exportersByNameCopy.containsKey("logging")) {
```
anuraaga, line 66: Cool

anuraaga (approved):

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/SpanExporterConfiguration.java
```diff
@@ -17,25 +18,58 @@
...
-  @Nullable
-  static SpanExporter configureExporter(String name, ConfigProperties config) {
+  private static final String EXPORTER_NONE = "none";
+
+  // Visible for testing
+  static Map<String, SpanExporter> configureSpanExporters(ConfigProperties config) {
+    List<String> exporterNamesList = config.getCommaSeparatedValues("otel.traces.exporter");
+    Set<String> exporterNames = new HashSet<>(exporterNamesList);
+    if (exporterNamesList.size() != exporterNames.size()) {
+      throw new ConfigurationException("otel.traces.exporter contains duplicates");
```
kubawach, line 39: NIT: we could name the duplicates in the exception message (or log). In case of long exporter lists this may be helpful.

kubawach (approved):
Looks good

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/SpanExporterConfiguration.java
jack-berg: Nice that's better.

sdk-extensions/autoconfigure/src/testFullConfig/java/io/opentelemetry/sdk/autoconfigure/ConfigurableSpanExporterTest.java
```diff
@@ -10,59 +10,146 @@
...
+  @Test
+  void configureSpanExporters_multipleWithNone() {
+    ConfigProperties config =
+        DefaultConfigProperties.createForTest(ImmutableMap.of("otel.traces.exporter", "otlp,none"));
+
+    assertThatThrownBy(() -> SpanExporterConfiguration.configureSpanExporters(config))
+        .isInstanceOf(ConfigurationException.class)
+        .hasMessageContaining("otel.traces.exporter contains none along with other exporters");
+  }
+
```
jkwatson, line 62: good tests. Do we have a test for something silly like a single comma (or several commas)? I would hope that would fail somehow, but I didn't see in the code how it might fail.

jkwatson (approved):
Looks good. Maybe one pathological test case we could add, but otherwise 👍🏽

sdk-extensions/autoconfigure/src/testFullConfig/java/io/opentelemetry/sdk/autoconfigure/ConfigurableSpanExporterTest.java
jack-berg: That situation isn't possible, since `DefaultConfigProperties#getCommaSeparatedValues(..)` filters out blank and null entries. 

I guess the question is whether we trust the responses of `DefaultConfigProperties#getCommaSeparatedValues(..)` to do what the javadoc says, or need to add additional checks / tests. My inclination is to not add additional checks / tests, since there is a test in `ConfigPropertiesTest` that confirms this already. 
```
Returns a list-valued configuration property. The format of the original value must be comma-separated. Empty values will be removed.
```

jkwatson:
looks like this needs a rebase, then should be good to merge.

PR #3533, opened 2021-08-23
Move TraceMarshaler to otlp-common and use in otlp-http-trace

anuraaga:


exporters/otlp-http/trace/src/main/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporter.java
```diff
@@ -85,16 +84,14 @@
   @Override
   public CompletableResultCode export(Collection<SpanData> spans) {
     spansSeen.add(spans.size());
-    ExportTraceServiceRequest exportTraceServiceRequest =
-        ExportTraceServiceRequest.newBuilder()
-            .addAllResourceSpans(SpanAdapter.toProtoResourceSpans(spans))
-            .build();
+
+    TraceRequestMarshaler exportRequest = TraceRequestMarshaler.create(spans);
```
kubawach, line 88: NIT: rename variable to `spansMarshaller` ?

kubawach (approved):

exporters/otlp-http/trace/src/main/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporter.java
anuraaga: Since it corresponds to the export request, I like that name a bit better

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/CodedOutputStream.java
```diff
@@ -58,7 +58,7 @@
 //
 // Unneeded lines of code are deleted as is, without any modifications otherwise.
 @SuppressWarnings({"UngroupedOverloads", "InlineMeSuggester"})
-abstract class CodedOutputStream {
+public abstract class CodedOutputStream {
```
jkwatson, line 61: we should probably add our standard boilerplate for internal classes in javadoc here.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/Marshaler.java
```diff
@@ -3,11 +3,11 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package io.opentelemetry.exporter.otlp.trace;
+package io.opentelemetry.exporter.otlp.internal;
 
 import java.io.IOException;
 
-interface Marshaler {
+public interface Marshaler {
```
jkwatson, line 10: internal boilerplate here

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/TraceRequestMarshaler.java
```diff
@@ -27,58 +27,61 @@
 import java.util.List;
 import java.util.Map;
 
-final class TraceMarshaler {
+/** {@link Marshaler} to convert SDK {@link SpanData} to OTLP ExportTraceServiceRequest. */
+public final class TraceRequestMarshaler extends MarshalerWithSize implements Marshaler {
```
jkwatson, line 31: internal boilerplate here.

jkwatson (approved):
Seems fine to me, with some internal-scare javadoc added.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/TraceRequestMarshaler.java
anuraaga: I noticed we have a lot of public classes without the boilerplate, both in this package and others like context / api. Wonder if it's a sign of being a fruitless effort :P

jkwatson: I try to catch them when I remember, but yeah, we're definitely still missing some.

PR #3537, opened 2021-08-23
Revert "Populate labels in addition to attributes for now on metrics.…

anuraaga:
… (#3397)"

This reverts commit 5e6be2bb526d41b7f1a0e34442e07fbc3a683e7e.

From what I understand, the collector does use attributes and not labels now. So we should be able to revert this given we have some time until the next release. Please double-check.

The main motivation is to not port label marshaling to MetricMarshaler.


jkwatson:
have we had an official collector release with this support yet?

jkwatson:
Looks like yes!  https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.33.0 Sweet!

jkwatson (approved):
Thanks! We'll definitely want to make sure we document this in the CHANGELOG/release notes, since it will break folks who are on older collectors that are expecting labels.

PR #3538, opened 2021-08-23
Refactor package to io.opentelemetry.exporter.otlp.http.metrics for c…

jack-berg:
The other metrics packages in the project are plural. This makes `:exporters:otlp-http:metrics` consistent. 

Better to make the change now while the number of consumers is small. Plus, this package is still in alpha.


jack-berg:
Note that PR #3522 has a string reference to this package name, so whichever of this PR and #3522 is merged second will need to be updated accordingly.

jkwatson:
> Note that PR #3522 has a string reference to this package name, so whichever of this PR and #3522 is merged second will need to be updated accordingly.

Looks like this one needs the update.

anuraaga (approved):

PR #3539, opened 2021-08-23
Add gradle ktfmt spotless config to all projects

### LINKED ISSUE #3435: Autoformat/enforce formatting of build scripts (kts)
https://github.com/open-telemetry/opentelemetry-java/pull/3427 showed that the build scripts were not consistently formatted. To avoid such large (or any!) formatting-only PRs in the future, we should have something like spotless+googleJavaFormat for the kts files.

jack-berg:
Resolves #3435. 

Uses [ktfmt](https://github.com/facebookincubator/ktfmt) to apply `google-java-format` style to all `*.gradle.kts` build files.

Opening as a draft to gauge interest in the approach and the resultant style. 

See [kotlin spotless config](https://github.com/diffplug/spotless/tree/main/plugin-gradle#kotlin) docs for additional configuration options.

### DIFF
--- a/all/build.gradle.kts
+++ b/all/build.gradle.kts
@@ -1,8 +1,7 @@
-plugins {
-  id("otel.java-conventions")
-}
+plugins { id("otel.java-conventions") }
 
 description = "OpenTelemetry All"
+
 otelJava.moduleName.set("io.opentelemetry.all")
 
 tasks {
@@ -10,15 +9,11 @@ tasks {
   // aggregating jacoco reports and it doesn't work if this isn't at least as high as the
   // highest supported Java version in any of our projects. Most of our projects target
   // Java 8, except for jfr-events.
-  withType(JavaCompile::class) {
-    options.release.set(11)
-  }
+  withType(JavaCompile::class) { options.release.set(11) }
 
   val testJavaVersion: String? by project
   if (testJavaVersion == "8") {
-    test {
-      enabled = false
-    }
+    test { enabled = false }
   }
 }
 
@@ -27,9 +22,7 @@ dependencies {
     // Generate aggregate coverage report for published modules that enable jacoco.
     subproject.plugins.withId("jacoco") {
       subproject.plugins.withId("maven-publish") {
-        implementation(project(subproject.path)) {
-          isTransitive = false
-        }
+        implementation(project(subproject.path)) { isTransitive = false }
       }
     }
   }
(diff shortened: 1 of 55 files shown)


all/build.gradle.kts
```diff
@@ -1,24 +1,19 @@
-plugins {
-  id("otel.java-conventions")
-}
+plugins { id("otel.java-conventions") }
```
anuraaga, line 1: Don't think Gradle DSL looks good with all the one-liners (which would look good for normal lambda invocations in general, but not DSL)

all/build.gradle.kts
```diff
@@ -66,21 +59,25 @@ tasks.named<JacocoReport>("jacocoTestReport") {
...
+          !it.absolutePath.contains("AutoValue_") &&
+          // TODO(anuraaga): Remove exclusion after enabling coverage for jfr-events
+          !it.absolutePath.contains("io/opentelemetry/sdk/extension/jfr")
+      }
+    )
   }
   additionalSourceDirs(sourcesPath.incoming.artifactView { lenient(true) }.files)
-  executionData(coverageDataPath.incoming.artifactView { lenient(true) }.files.filter { it.exists() })
+  executionData(
+    coverageDataPath.incoming.artifactView { lenient(true) }.files.filter { it.exists() }
```
anuraaga, line 79: This change looks reasonable

api/all/build.gradle.kts
```diff
@@ -7,7 +7,9 @@ plugins {
 }
 
 description = "OpenTelemetry API"
+
```
anuraaga, line 10: Don't think these newlines improve (wonder why ktfmt would go this far)

bom-alpha/build.gradle.kts
```diff
@@ -16,15 +18,12 @@ rootProject.subprojects.forEach { subproject ->
 afterEvaluate {
   dependencies {
     constraints {
-      rootProject.subprojects
+      rootProject
+        .subprojects
```
anuraaga, line 22: This change looks like a good one

build.gradle.kts
```diff
@@ -9,28 +9,25 @@ plugins {
 }
 
 if (!JavaVersion.current().isJava11Compatible()) {
-  throw GradleException("JDK 11 or higher is required to build. " +
-    "One option is to download it from https://adoptopenjdk.net/. If you believe you already " +
-    "have it, please check that the JAVA_HOME environment variable is pointing at the " +
-    "JDK 11 installation.")
+  throw GradleException(
+    "JDK 11 or higher is required to build. " +
```
anuraaga, line 13: This change looks good

anuraaga (commented):
Thanks for taking a look at this @jack-berg - definitely interesting to try out formatters. I think there are significant gaps here so wouldn't want to go with this formatter unfortunately.

I notice that ktlint mentions `*.kts` files on its README and it's not mentioned on ktfmt so perhaps it has more idiomatic formatting of kts files - do you mind giving it a try?

anuraaga:
Realized ktlint is what I'm using in my distro :)

https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/build.gradle.kts#L71

Formats well, except for some corner cases like having to exclude settings.gradle

all/build.gradle.kts
Oberon00: I have no issue with that. I would much prefer a general kts/Kotlin formatter over no formatter, even if it sometimes leads to sub-optimal results.

jack-berg:
@anuraaga I opened an alternative [draft PR](https://github.com/open-telemetry/opentelemetry-java/pull/3543) using ktlint instead of ktfmt. It seems less opinionated, including allowing short closures to be on multiple lines. Personally I'm partial to closures that are not inlined, so I prefer ktlint as well.

anuraaga:
Closing since we used ktlint - thanks!

PR #3540, opened 2021-08-24
Add Marshaler for metrics

anuraaga:
I have not added random data generation checks for metrics mainly since they're still alpha. Instead, I copied `MetricAdapterTest` and use the exact test cases for the marshaler path. So we have the same code coverage for marshaler as we do for adapter, which I think is pretty good.

Unlike tracing, we didn't do any optimizations to MetricAdapter, so have a huge improvement in performance.

```
Benchmark                                                                Mode  Cnt      Score     Error   Units
MetricsRequestMarshalerBenchmark.adapter                                 avgt   10      6.752 ±   0.089   us/op
MetricsRequestMarshalerBenchmark.adapter:·gc.alloc.rate                  avgt   10   1184.995 ±  15.979  MB/sec
MetricsRequestMarshalerBenchmark.adapter:·gc.alloc.rate.norm             avgt   10  12592.003 ±   0.001    B/op
MetricsRequestMarshalerBenchmark.adapter:·gc.churn.G1_Eden_Space         avgt   10   1191.516 ±  76.866  MB/sec
MetricsRequestMarshalerBenchmark.adapter:·gc.churn.G1_Eden_Space.norm    avgt   10  12659.490 ± 728.041    B/op
MetricsRequestMarshalerBenchmark.adapter:·gc.churn.G1_Old_Gen            avgt   10      0.003 ±   0.002  MB/sec
MetricsRequestMarshalerBenchmark.adapter:·gc.churn.G1_Old_Gen.norm       avgt   10      0.033 ±   0.019    B/op
MetricsRequestMarshalerBenchmark.adapter:·gc.count                       avgt   10     98.000            counts
MetricsRequestMarshalerBenchmark.adapter:·gc.time                        avgt   10     42.000                ms
MetricsRequestMarshalerBenchmark.marshaler                               avgt   10      2.465 ±   0.088   us/op
MetricsRequestMarshalerBenchmark.marshaler:·gc.alloc.rate                avgt   10   1923.724 ±  67.211  MB/sec
MetricsRequestMarshalerBenchmark.marshaler:·gc.alloc.rate.norm           avgt   10   7464.001 ±   0.001    B/op
MetricsRequestMarshalerBenchmark.marshaler:·gc.churn.G1_Eden_Space       avgt   10   1923.354 ± 126.347  MB/sec
MetricsRequestMarshalerBenchmark.marshaler:·gc.churn.G1_Eden_Space.norm  avgt   10   7463.182 ± 441.363    B/op
MetricsRequestMarshalerBenchmark.marshaler:·gc.churn.G1_Old_Gen          avgt   10      0.005 ±   0.003  MB/sec
MetricsRequestMarshalerBenchmark.marshaler:·gc.churn.G1_Old_Gen.norm     avgt   10      0.019 ±   0.012    B/op
MetricsRequestMarshalerBenchmark.marshaler:·gc.count                     avgt   10    131.000            counts
MetricsRequestMarshalerBenchmark.marshaler:·gc.time                      avgt   10     57.000                ms
```


exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MarshalerUtil.java
```diff
@@ -59,6 +127,26 @@ static void marshalBytes(int fieldNumber, byte[] message, CodedOutputStream outp
...
+  static int sizeRepeatedFixed64(int fieldNumber, List<Long> values) {
+    return sizeRepeatedFixed64(fieldNumber, values.size());
+  }
+
+  private static int sizeRepeatedFixed64(int fieldNumber, int numValues) {
+    if (numValues == 0) {
+      return 0;
+    }
+    int dataSize = WireFormat.FIXED64_SIZE * numValues;
+    int size = 0;
```
jkwatson, line 139: any particular reason to initialize this to `0`, rather than just the first thing you add to it?

jkwatson: or just return the two things added together?

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MetricsRequestMarshaler.java
```diff
@@ -0,0 +1,892 @@
...
+        resourceMetricsMarshalers,
+        output);
+  }
+
+  private static int calculateSize(ResourceMetricsMarshaler[] resourceMetricsMarshalers) {
+    int size = 0;
+    size +=
+        MarshalerUtil.sizeRepeatedMessage(
+            ExportMetricsServiceRequest.RESOURCE_METRICS_FIELD_NUMBER, resourceMetricsMarshalers);
+    return size;
```
jkwatson, line 106: ? 
Some reason not to just return the value?

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MetricsRequestMarshaler.java
```diff
@@ -0,0 +1,892 @@
...
+      io.opentelemetry.sdk.metrics.data.AggregationTemporality temporality) {
+    switch (temporality) {
+      case CUMULATIVE:
+        return AggregationTemporality.AGGREGATION_TEMPORALITY_CUMULATIVE_VALUE;
+      case DELTA:
+        return AggregationTemporality.AGGREGATION_TEMPORALITY_DELTA_VALUE;
+    }
+    return AggregationTemporality.AGGREGATION_TEMPORALITY_UNSPECIFIED_VALUE;
+  }
+}
```
jkwatson, line 892: this class is ginormous. I wonder if we can break it up a bit.

jkwatson (approved):
Seems fairly reasonable to me, aside from a few odd bits that are probably left over from something else.

@jsuereth your eyes will be valuable on this as well.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MetricsRequestMarshaler.java
anuraaga: I guess we're not expecting it, but I think we're using this pattern (I'm just following our current pattern) to make it easier to add / remove fields.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MetricsRequestMarshaler.java
anuraaga: Yeah it's probably hairy - should I split in this PR?

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MarshalerUtil.java
anuraaga: See my other comment, it makes less sense for this class I guess but just to be consistent with the calculateSize methods

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MetricsRequestMarshaler.java
jkwatson: hm. interesting. i'm surprised IDEA doesn't yell about this. :)

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MetricsRequestMarshaler.java
jkwatson: This one, or a follow-up. I was definitely cross-eyed by the time I got about half way through.

anuraaga: Will do in a followup, I don't think it makes sense without a `metrics` subpackage since lots of related classes, but will require a lot of package private -> public transition of existing code.

anuraaga:
Going to go ahead and merge this in but as always @jsuereth happy to follow up on any suggestions.

PR #3541, opened 2021-08-24
Allow multiple views per instrument

jsuereth:
- Allow multiple views per instrument
- Log on view conflict instead of throwing an exception.
- Update error reporting (slightly)

This does *NOT*

- Attempt to only poll async metrics once-per-instrument (instead once-per-metric for now)
- Give the best error messages

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DuplicateMetricStorageException.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DuplicateMetricStorageException.java
@@ -1,0 +1,20 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.metrics.internal.state;
+
+import io.opentelemetry.sdk.metrics.internal.descriptor.MetricDescriptor;
+
+/** There are multiple metrics defined with the same name/identity. */
+class DuplicateMetricStorageException extends IllegalArgumentException {
+
+  DuplicateMetricStorageException(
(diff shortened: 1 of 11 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DuplicateMetricStorageException.java
```diff
@@ -0,0 +1,20 @@
...
+/** There are multiple metrics defined with the same name/identity. */
+class DuplicateMetricStorageException extends IllegalArgumentException {
+
+  DuplicateMetricStorageException(
+      MetricDescriptor existing, MetricDescriptor next, String message) {
+    // TODO: Better error messages including async vs. sync instruments.
+    super(message + " Found: " + existing + ", Want: " + next);
+  }
+
+  private static final long serialVersionUID = 1547329629200005982L;
```
jkwatson, line 19: this makes me extremely sad, that for some ungodly reason, we still need to put this cruft into our files.

jkwatson: also, it needs to move to the top of the class.  ;)

jkwatson (approved):
one little gripe & a corresponding nit, but this looks pretty solid to me.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DuplicateMetricStorageException.java
jsuereth: Agreed, fixed.

PR #3543, opened 2021-08-24
Add gradle ktlint spotless config to all projects

### LINKED ISSUE #3435: Autoformat/enforce formatting of build scripts (kts)

jack-berg:
Resolves #3435.

Alternative to PR #3539 using [ktlint](https://github.com/pinterest/ktlint) instead of ktfmt. Opened a second draft PR to compare the style.


trask:
this looks great to me, will plan to add same in otel-java-instrumentation if/when accepted here

jkwatson (approved):
Thanks!

anuraaga (approved):
Sweet good improvement!

PR #3545, opened 2021-08-25
Reuse a threadlocal CodedOutputStream for marshaling.

anuraaga:
Also raises the buffer size from 4K to 50K. We only expect a single thread exporting in 99% of apps, so the overall overhead is 50K. Some event-loop driven apps may avoid the BatchSpanProcessor, they will still generally only have 2 * num CPU threads, 64 threads would be about 3MB of overhead which seems OK. 1000 threads would be 50MB which is probably not OK, but not using the BatchSpanProcessor when processing requests with 1000 threads is not going to have good performance anyways and I think we can accept this. For reference, Jackson reuses a 8K buffer, while Jackson does not have any expectation on the number of threads 

There is an escape hatch flag (which I don't think we need to document until anyone happens to complain about it) in case this is prohibitive for some reason.

Other approaches could be a buffer pool, or not worrying about the async case and passing in a threadlocal from BSP.

### DIFF
--- a/exporters/otlp/trace/src/jmh/java/io/opentelemetry/exporter/otlp/trace/RequestMarshalState.java
+++ b/exporters/otlp/trace/src/jmh/java/io/opentelemetry/exporter/otlp/trace/RequestMarshalState.java
@@ -50,7 +50,7 @@ public class RequestMarshalState {
   private static final SpanContext SPAN_CONTEXT =
       SpanContext.create(TRACE_ID, SPAN_ID, TraceFlags.getSampled(), TraceState.getDefault());
 
-  @Param({"16"})
+  @Param({"16", "512"})
   int numSpans;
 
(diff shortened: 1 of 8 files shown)


exporters/otlp/trace/src/jmh/java/io/opentelemetry/exporter/otlp/trace/RequestMarshalState.java
```diff
@@ -50,7 +50,7 @@
   private static final SpanContext SPAN_CONTEXT =
       SpanContext.create(TRACE_ID, SPAN_ID, TraceFlags.getSampled(), TraceState.getDefault());
 
-  @Param({"16"})
+  @Param({"16", "512"})
```
anuraaga, line 53: 512 is the default BSP max export batch size

PR #3548, opened 2021-08-25
Wait for termination for gRPC shutdown.

anuraaga:
While it would be nice if this could be done asynchronously in the future, for now let's just create a thread it's not the end of the world.

The current code is not correct because `notifyStateChanged` notifies you when the state _isn't_ the value. As we've never called SHUTDOWN yet, `.notifyWhenStateChanged(ConnectivityState.SHUTDOWN, result::succeed);` always executes the callback inline. But because there's no way to get notified when shutdown is complete, there isn't any way of making that happen.

### DIFF
--- a/exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporter.java
+++ b/exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporter.java
@@ -8,7 +8,6 @@ package io.opentelemetry.exporter.otlp.trace;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.MoreExecutors;
-import io.grpc.ConnectivityState;
 import io.grpc.ManagedChannel;
 import io.grpc.Status;
 import io.opentelemetry.api.common.AttributeKey;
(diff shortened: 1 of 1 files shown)


exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporter.java
```diff
@@ -179,14 +178,29 @@ public static OtlpGrpcSpanExporter getDefault() {
...
     }
-    managedChannel.shutdown();
     this.spansSeen.unbind();
     this.spansExportedSuccess.unbind();
     this.spansExportedFailure.unbind();
+    managedChannel.shutdown();
+    // Remove thread creation if gRPC adds an asynchronous shutdown API.
+    // https://github.com/grpc/grpc-java/issues/8432
+    Thread thread =
+        new Thread(
```
Oberon00, line 191: Shouldn't this use some kind of (background/daemon) thread factory?

anuraaga: Didn't think it's needed since it's guaranteed to end in 10 (usually much less) seconds

jkwatson: I think at least setting the thread as a daemon would be a good idea, so it doesn't block JVM shutdown when people don't care about this case.

PR #3558, opened 2021-08-27
Move ConfigProperties / ConfigurationException to SPI package and con…

anuraaga:
…sistently pass ConfigProperties

Will move to separate artifact after this.


jsuereth:
IIUC - this breaks the SPI for autocnfiguration, so we should make sure to mark the breakage somewhere.

Specifically, this creates version breakages in usages like this: https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/tree/main/exporters/auto

anuraaga:
Yeah we'll call it out in release notes - on the bright side I'm thinking we'll be able to mark the separated SPI artifact stable by only putting trace classes in it.

jkwatson:
> Yeah we'll call it out in release notes - on the bright side I'm thinking we'll be able to mark the separated SPI artifact stable by only putting trace classes in it.

Is there some way we could keep the other one around and deprecate it and it's usages? @jsuereth would that be preferable for you?

anuraaga:
@jkwatson I can't come up with a strategy for `ConfigProperties` - if the SPI classes themselves like `ConfigurableSpanExporter` were also being moved, then I could keep the old ones around but those are in the right place referencing a subclass of `ConfigProperties` or similar. There might be a way with having two methods on each class with default implementations, but I'm worried it will have its own issues and think it's a bit much for an alpha artifact (we should still keep non-breaking best effort, similar to our approach for the metrics package).

sdk-extensions/autoconfigure/src/testFullConfig/java/io/opentelemetry/sdk/autoconfigure/FullConfigTest.java
```diff
@@ -197,7 +197,7 @@ void configures() throws Exception {
               .contains(
                   KeyValue.newBuilder()
                       .setKey("configured")
-                      .setValue(AnyValue.newBuilder().setStringValue("true").build())
+                      .setValue(AnyValue.newBuilder().setBoolValue(true).build())
```
jkwatson, line 200: interesting. is this a behavioral change somehow?

anuraaga: Nah I just made the test code consistent between trace and metrics for the test SPI implementations

jkwatson: cool. yeah, I think I figured that out eventually working through here. 👍🏽

jkwatson:
@jsuereth Last chance to "demand" a deprecation cycle before I merge. :)

jsuereth:
@jkwatson No, I won't demand a deprecation cycle (not while I'm also breaking the hell out of the metrics SDK).   Once it's clear what's stable or not, then I'll complain.  

However, i'd prefer if we did the MINIMUM cosmetic changes here, so to the extent we can make sure the refactoring we're doing is unlikely to happen again, that'd be great :)

anuraaga:
@jsuereth Yeah for most of the SPIs it should just be a package regex rename, the providers have a method added but now it's all consistent and should be good. I intend for these to be released as non-alpha now :)

PR #3562, opened 2021-08-28
Rework LabelsProcessor to have access to context - Update View SDK API

jsuereth:
- Rename `LabelsProcessor` -> `AttributesProcessor`
- Update interface to denote if `Context` needs to be used (disabling the `bind` performance improvement).
- Allow baggage to be appended to measurements (via View API)
- Create semi-end-to-end baggage-in-metric test case

Notes:
- Not super happy with `ViewBuilder` interface.   For suggestions, please read [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#view) and offer ideas.  Currently specification only calls for filtering by attribute keys and adding "extra attributes" from baggage.
- Likely there's some cleanup that can be done on AttributesProcessor itself, that's "in" for this PR.
- Aggregations (and metric storage) still need some significant rework.  This PR, once again, makes this look a bit gunkier to bring us to the public SDK specification quicker, and hiding the pieces we'll need to refactor over time.


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/AttributesProcessor.java
```diff
@@ -0,0 +1,198 @@
...
+   * @param context The context associated with the measurement.
+   */
+  public abstract Attributes process(Attributes incoming, Context context);
+
+  /**
+   * If true, this ensures the `Context` argument of the attributes processor is always accurate.
+   * This will prevents bound instruments from pre-locking their metric-attributes and defer until
+   * context is available.
+   */
+  public abstract boolean usesContext();
```
jkwatson, line 45: Is this safe enough? Couldn't implementations be adding conditionally adding attributes from other things (like a pre-configured Resource, for example)?

jkwatson: In fact, isn't it the case that *any* dynamic manipulation of attributes will break the binding? Maybe we need this to be "isDynamic()" or something?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/ViewBuilder.java
```diff
@@ -52,19 +56,102 @@ public ViewBuilder setAggregatorFactory(AggregatorFactory aggregatorFactory) {
...
   }
 
+  /**
+   * Appends all key-values from baggage to all measurements.
+   *
+   * <p>Note: This runs after all other attribute processing added so far.
+   *
+   * @return this Builder.
+   */
+  public ViewBuilder addAppendBaggage() {
```
jkwatson, line 149: maybe "appendAllBaggageAttributes"?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/ViewBuilder.java
```diff
@@ -52,19 +56,102 @@ public ViewBuilder setAggregatorFactory(AggregatorFactory aggregatorFactory) {
...
+  /**
+   * Appends key-values from baggage to all measurements.
+   *
+   * <p>Note: This runs after all other attribute processing added so far.
+   *
+   * @param keyFilter Only baggage key values pairs where the key matches this predicate will be
+   *     appended.
+   * @return this Builder.
+   */
+  public ViewBuilder addAppendBaggageFilter(Predicate<String> keyFilter) {
```
jkwatson, line 121: maybe "appendFilteredBaggageAttributes" ?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/ViewBuilder.java
```diff
@@ -52,19 +56,102 @@ public ViewBuilder setAggregatorFactory(AggregatorFactory aggregatorFactory) {
...
+
+  /**
+   * Appends a static set of attributes to all measurements.
+   *
+   * <p>Note: This runs after all other attribute processing added so far.
+   *
+   * @param extraAttributes The static attributes to append to measurements.
+   * @return this Builder.
+   */
+  public ViewBuilder addAppendedAttributes(Attributes extraAttributes) {
```
jkwatson, line 107: maybe just "appendAttributes" ?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/ViewBuilder.java
```diff
@@ -52,19 +56,102 @@ public ViewBuilder setAggregatorFactory(AggregatorFactory aggregatorFactory) {
...
+
+  /**
+   * Filters measurement attributes using a given regex.
+   *
+   * <p>Note: This runs after all other attribute processing added so far.
+   *
+   * @param keyPattern the regular expression for selecting attributes by key name.
+   * @return this Builder.
+   */
+  public ViewBuilder addAttributeKeyPattern(Pattern keyPattern) {
```
jkwatson, line 92: how about "filterAttributesByKeyPattern"?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/ViewBuilder.java
```diff
@@ -52,19 +56,102 @@ public ViewBuilder setAggregatorFactory(AggregatorFactory aggregatorFactory) {
...
+
+  /**
+   * Filters measurement attributes using a given filter.
+   *
+   * <p>Note: This runs after all other attribute processing added so far.
+   *
+   * @param keyFilter filter for key names to include.
+   * @return this Builder.
+   */
+  public ViewBuilder addAttributeKeyFilter(Predicate<String> keyFilter) {
```
jkwatson, line 79: "filterAttributesByKey" ?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/AttributesProcessor.java
jsuereth: Yeah, this could be better documented.  One reason I tried not to expose this interface any longer was so only the extensions we provide are allowed.

To the extent this was intended to be a "pure functional" API, the only input that matters is context.  Happy to figure out how to document and change this portion.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/ViewBuilder.java
jsuereth: these names kind of imply an "order" to how they're done.  I was on the fence behind forcing an order or just doing them in the order they're called on the builder.   Do you have a preference?

jkwatson: I think just doing them in the order called will probably be the easiest to explain and document.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/AttributesProcessor.java
```diff
@@ -0,0 +1,198 @@
...
+      if (other instanceof JoinedAttribtuesProcessor) {
+        newList.addAll(((JoinedAttribtuesProcessor) other).processors);
+      } else {
+        newList.add(other);
+      }
+      return new JoinedAttribtuesProcessor(newList);
+    }
+
+    AttributesProcessor prepend(AttributesProcessor other) {
+      ArrayList<AttributesProcessor> newList = new ArrayList<>(processors.size() + 1);
```
jack-berg, line 192: #nit: ArrayList is irrelevant. Same for line 182. 
```suggestion
      List<AttributesProcessor> newList = new ArrayList<>(processors.size() + 1);
```

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/ViewBuilder.java
```diff
@@ -52,19 +56,102 @@ public ViewBuilder setAggregatorFactory(AggregatorFactory aggregatorFactory) {
...
-   * sets {@link LabelsProcessorFactory}.
+   * Specify the attributes processor for this view.
    *
-   * @param labelsProcessorFactory labels processor factory.
+   * <p>Note: This resets all attribute filters, baggage appending and other processing.
+   *
+   * @param processor The pre-processor for measurement attributes.
+   * @return this Builder.
+   */
+  public ViewBuilder setAttributesProcessor(AttributesProcessor processor) {
```
jack-berg, line 66: My preference would be to keep the surface area of configuring the attributes processor smaller, with just two methods:
```
public ViewBuilder setAttributesProcessor(AttributesProcessor processor);

public ViewBuilder addAttributeProcessor(AttributesProcessor processor) 
```

jsuereth: AttributesProcessor isn't specified.   I'd agree that this is simpler, but my goal was NOT to expose attribute processor publicly for two reasons:

1. I expect refactoring in the aggregator API (and a specification at some point).   It's likely this API would break at that point.
2.  We've been avoiding placing anything in a stable location that is unspecified (or where we're not just deviating, but inventing new functionality that diverges from the spec).

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/AttributesProcessor.java
```diff
@@ -0,0 +1,198 @@
...
+/**
+ * An {@code AttributesProcessor} is used by {@code View}s to define the actual recorded set of
+ * attributes.
+ *
+ * <p>An AttributesProcessor is used to define the actual set of attributes that will be used in a
+ * Metric vs. the inbound set of attributes from a measurement.
+ */
+@Immutable
+public abstract class AttributesProcessor {
+  private AttributesProcessor() {}
```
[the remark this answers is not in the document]
jsuereth: That's the point.  `setAttributesProcessor` needs to be marked "visible for testing".

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/AttributesProcessor.java
```diff
@@ -0,0 +1,198 @@
...
+
+/**
+ * An {@code AttributesProcessor} is used by {@code View}s to define the actual recorded set of
+ * attributes.
+ *
+ * <p>An AttributesProcessor is used to define the actual set of attributes that will be used in a
+ * Metric vs. the inbound set of attributes from a measurement.
+ */
+@Immutable
+public abstract class AttributesProcessor {
```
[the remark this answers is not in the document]
jsuereth: Agreed, however see my comments on locking it down.  Originally it was a class, but I'm limiting its extensibility purposefully until there's more solidarity on what `Aggregator` would look like if we allow customization in the SDK specification.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/AttributesProcessor.java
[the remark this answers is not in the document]
jack-berg: 👌

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/AttributesProcessor.java
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  /**
+   * Creates a processor which appends values from {@link Baggage}.
+   *
+   * <p>These attributes will not override those attributes provided by instrumentation.
+   *
+   * @param nameFilter a filter for which baggage keys to select.
+   */
+  public static AttributesProcessor appendBaggageByKeyName(Predicate<String> nameFilter) {
```
anuraaga, line 94: Hmm - could we leave out baggage from this initial PR? I think it can use more thought.

My initial thought is this isn't where to map baggage to attributes, that should probably happen in the instrumentation, who's "job" is to create the attributes for traces and metrics. At least in the way we've defined it so far.

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java#L140

That way, both traces and metrics will get the baggage. And metrics an use the normal filtering mechanism for attributes without a separate baggage-specific one. I suspect any approach we take to baggage we'd want to map into attributes, independently of signal, and let the signals consume them.

jsuereth: I hear what you're saying, but that discussion (likely) belongs in the specification sig.  Baggage is part of the View SDK specification right now:

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#view

> The extra dimensions which come from Baggage/Context (optional). If not provided, no extra dimension will be used. Please note that this only applies to synchronous Instruments.


In the interest of not letting perfect be the enemy of good, I think the current implementation is ok for a few reasons:

1. We've hidden (on purpose) Attributes processor in `internal` so we can evolve it going forward.
2. Baggage appending is unit tested to only be additive.  If instrumentation pre-adds baggage we won't overwrite those labels.
3. Most of the instrumentation work you have regarding baggage is unspecified (AFAIK), meanwhile we have a clear use case and interface via Views.  I suspect we can (and should) find a way to complement the two going forward, but it doesn't have to be now.  This is similar to how each signal needs to independently provide useful feature and then *also* work well when combined in the instrumenter interface.
4. When it comes to leveraging baggage-in-traces (or logs), I actually feel there's a missing component in the SDK that should be doing this, especially for the non-instrumentation cases.

anuraaga: The spec repo is a wall of MD and either no comments, or too many comments from a variety of biased vendors - I find discussions there very unproductive. So I'm hoping to take an approach of leaving feedback here and perhaps you will take it to other discussions, or not. I don't intend to block anything, just providing information / suggestions which may be useful. I know it's not fair for me to expect big changes without actually going through the grunt work of the spec - so it's fine if changes aren't made.

I think my biggest reaction here was that we have a path from baggage to metrics already, but even though there's demand we don't have it for traces, which is launched. It's surprising - but it's OK :-)

sdk-extensions/autoconfigure/src/testFullConfig/java/io/opentelemetry/sdk/autoconfigure/TestMeterProviderConfigurer.java
```diff
@@ -5,29 +5,24 @@
...
-                labels.toBuilder()
-                    .put("configured", config.getBoolean("otel.test.configured"))
-                    .build();
-
     for (InstrumentType instrumentType : InstrumentType.values()) {
       meterProviderBuilder.registerView(
           InstrumentSelector.builder().setInstrumentType(instrumentType).build(),
-          View.builder().setLabelsProcessorFactory(labelsProcessorFactory).build());
+          View.builder()
+              .appendAttributes(Attributes.builder().put("configured", true).build())
```
jkwatson, line 24: tiny nit, but this would probably read a bit better with just `Attributes.of(booleanKey("configured"), true)`

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/AttributesProcessor.java
jkwatson: A great point: how is it that we are building this into metrics before we build it into spans?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/ViewBuilder.java
```diff
@@ -51,18 +55,104 @@ public ViewBuilder setAggregation(Aggregation aggregation) {
...
+
+  /**
+   * Appends a static set of attributes to all measurements.
+   *
+   * <p>Note: This runs after all other attribute processing added so far.
+   *
+   * @param extraAttributes The static attributes to append to measurements.
+   * @return this Builder.
+   */
+  public ViewBuilder appendAttributes(Attributes extraAttributes) {
```
jkwatson, line 108: I'm somewhat curious about the use-case for this particular feature. Do you have a specific use in mind?

jkwatson (approved):
I think this is fine to move forward with. This is essentially all still prototype-work to make sure that spec is functional and hence worth trying.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/ViewBuilder.java
jsuereth: Besides our integration test? :)

Marking a canary server with appended attributes.

anuraaga (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/ViewBuilder.java
jkwatson: Is that not something that would be on the resource and hence automatically associated with the metrics?

anuraaga:
I think #3574 will fix the test failure

PR #3563, opened 2021-08-29
Cache AttributeKey hashCode

trask:
This turns out to be very helpful when looking up attribute keys in an attributes map.

I thought about using AutoValue's `@Memoize`, but the implementation for that is lazy double-checked lock (i.e. a volatile read), which seems unnecessary overhead in this case.

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/common/AttributeKeyImpl.java
+++ b/api/all/src/main/java/io/opentelemetry/api/common/AttributeKeyImpl.java
@@ -5,12 +5,26 @@
 
 package io.opentelemetry.api.common;
 
-import com.google.auto.value.AutoValue;
 import javax.annotation.Nullable;
 
 @SuppressWarnings("rawtypes")
-@AutoValue
-abstract class AttributeKeyImpl<T> implements AttributeKey<T> {
+final class AttributeKeyImpl<T> implements AttributeKey<T> {
+
+  private final AttributeType type;
+  private final String key;
+  private final int hashCode;
+
+  private AttributeKeyImpl(AttributeType type, String key) {
+    if (type == null) {
+      throw new NullPointerException("Null type");
+    }
+    this.type = type;
+    if (key == null) {
+      throw new NullPointerException("Null key");
+    }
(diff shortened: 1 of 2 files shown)


api/all/src/main/java/io/opentelemetry/api/common/AttributeKeyImpl.java
```diff
@@ -5,12 +5,26 @@
...
+  private AttributeKeyImpl(AttributeType type, String key) {
+    if (type == null) {
+      throw new NullPointerException("Null type");
+    }
+    this.type = type;
+    if (key == null) {
+      throw new NullPointerException("Null key");
+    }
+    this.key = key;
+    this.hashCode = buildHashCode();
```
trask, line 26: I don't love calling instance method from constructor, but the `EqualsVerifier.withCachedHashCode()` check seems to force that.

jkwatson: You can't make it a static method and pass in the fields?

jkwatson: oh, well, I think I'd rather skip that equalsverifier check in the test. What is that test doing under the hood anyway?

anuraaga (approved):
EqualsVerifier automatically generates some instances with reflection and makes sure equals and hashcode are implemented correctly. As we've implemented it by hand here I think it's worth keeping. The instance method `buildHashCode` isn't ideal but seems ok too.

Though I wonder what the difference is between `withCachedHashCode` and `withIgnoredFields("hashCode")`.

trask:
> `withIgnoredFields("hashCode")`

```
EqualsVerifier found a problem in class io.opentelemetry.api.common.AttributeKeyImpl.
-> Significant fields: equals relies on type, but hashCode does not.
  one has hashCode 1
  one has hashCode 1
```

jkwatson (approved):

jkwatson:
needs a rebase to pick up the fixed collector image

PR #3564, opened 2021-08-29
Allow creating a SpanContext without validation for internal use

trask:
The calls to TraceId.isValid() and SpanId.isValid() seem to be unnecessary inside of `SdkSpanBuilder.startSpan()` since traceId is from an existing SpanContext (so already validated) and spanId is from `IdGenerator`.

(this is showing up under profiling: https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/benchmark-overhead-jmh)

If it's acceptable, instead of creating a new SpanContext implementation for this, I can move the existing ImmutableSpanContext to the `internal` package and make it public, and exposing the new `createBypassingValidation()` method on it (I tend to prefer fewer concrete subclasses because it's friendlier to JIT inlining).


anuraaga (commented):
Thanks, this is one of the points I found in #3556!

Moving to `internal` package is good I think.

anuraaga:
By the way it does mean an `IdGenerator` returning invalid IDs will not have spans suppressed. I think this is an extreme rare case, and everything is broken anyways so it's probably not that bad. But we should double-check propagation code to make sure it won't throw exceptions if the IDs are e.g., wrong length or contain non-hex.

anuraaga (approved):

api/all/src/main/java/io/opentelemetry/api/internal/ImmutableSpanContext.java
```diff
@@ -51,6 +56,18 @@ static SpanContext create(
         /* valid= */ false);
   }
 
+  /**
+   * This method is provided as an optimization when {@code traceIdHex} and {@code spanIdHex} have
+   * already been validated. Only use this method if you are sure these have both been validated,
+   * e.g. when using {@code traceIdHex} from a parent {@link SpanContext} and {@code spanIdHex} from
+   * an {@code IdGenerator}.
+   */
+  public static SpanContext createBypassingValidation(
```
anuraaga, line 65: Can you add a unit test, also demonstrating this passing with some input that fails with the other method? For some coverage + documentation scones

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkSpanBuilder.java
```diff
@@ -191,7 +192,7 @@ public Span startSpan() {
     TraceState samplingResultTraceState =
         samplingResult.getUpdatedTraceState(parentSpanContext.getTraceState());
     SpanContext spanContext =
-        SpanContext.create(
+        ImmutableSpanContext.createBypassingValidation(
```
anuraaga, line 195: By the way, any thoughts on storing a boolean based on `tracerSharedState.getIdGenerator().getClassName()` to only skip validation on our SDK ID generators?

jkwatson: I was going to suggest something like this as well. I think that we probably should have this, to account for the case that you pointed out with a bad idGenerator letting things slip through unchecked.

It does make me think that there might be some value in providing a TCK for IdGenerator implementations, as well.

api/all/src/main/java/io/opentelemetry/api/internal/ImmutableSpanContext.java
```diff
@@ -32,7 +37,7 @@ private static AutoValue_ImmutableSpanContext createInternal(
         traceId, spanId, traceFlags, traceState, remote, valid);
   }
 
-  static SpanContext create(
+  public static SpanContext create(
```
jkwatson, line 40: this is failing CI due to missing javadoc

api/all/src/main/java/io/opentelemetry/api/internal/ImmutableSpanContext.java
```diff
@@ -0,0 +1,81 @@
...
+   * @return a new {@code SpanContext} with the given identifiers and options.
+   */
+  public static SpanContext create(
+      String traceIdHex,
+      String spanIdHex,
+      TraceFlags traceFlags,
+      TraceState traceState,
+      boolean remote,
+      boolean skipIdValidation) {
+    if (!skipIdValidation && SpanId.isValid(spanIdHex) && TraceId.isValid(traceIdHex)) {
```
jkwatson, line 66: Is this logic correct? If `skipIdValidation` is `true`, then `!skipIdValidation` is `false, so the `if` fails, and we drop through to always returning an invalid span context. I thought that "skipIdValidation" would mean that we had pre-validated as `valid`, not as `invalid`. 

Shouldn't this be `if (skipIdValidation || (SpanId.isValid(spanIdHex) && TraceId.isValid(traceIdHex))` ?

trask: ugh, boolean logic, i'm sure you're right, i will write test...

jkwatson: this make me think we might be missing some unit tests as well...

sdk/trace/src/test/java/io/opentelemetry/sdk/trace/SdkSpanBuilderTest.java
```diff
@@ -944,8 +944,8 @@ void spanDataToString() {
     assertThat(span.toSpanData().toString())
         .matches(
             "SpanData\\{spanContext=ImmutableSpanContext\\{"
-                + "traceId=[0-9a-f]{32}, "
-                + "spanId=[0-9a-f]{16}, "
+                + "traceId=(?!00000000000000000000000000000000)[0-9a-f]{32}, " // check not invalid
```
anuraaga, line 947: Is toString affected by this change?

trask: the negative lookahead is just to make sure it's not an invalid trace id, which would have caught my very bad boolean logic (I should have expanded on that comment to the right, but ran out of room 😂)

jkwatson: so... you got the boolean logic wrong, but I should trust you to get the regex right?  🤣

jkwatson: I kid, I kid. :)

trask: I promise 😂, I updated the test, and the test failed, then I fixed the boolean logic, and the test passed

anuraaga: Ah ok - the invalid spancontext is usually replaced by a valid one here I'm presuming which stopped happening with the wrong logic.

It's probably worth having a unit test that calls the `create` method directly setting the bool true or false

anuraaga (approved):

jkwatson (approved):

PR #3565, opened 2021-08-30
Use collector-contrib for tests since many components have been moved…

anuraaga:
… to it.

@trask Is this the only change we need to fix integration tests?

### DIFF
--- a/.github/workflows/docker-test-containers.yml
+++ b/.github/workflows/docker-test-containers.yml
@@ -12,7 +12,7 @@ jobs:
         include:
           - source: jaegertracing/all-in-one:1.17
             target_tag: jaeger
-          - source: otel/opentelemetry-collector-dev:latest
+          - source: otel/opentelemetry-collector-contrib-dev:latest
             target_tag: otel-collector-dev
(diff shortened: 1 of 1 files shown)


trask:
> @trask Is this the only change we need to fix integration tests?

👍

jsuereth (approved):

carlosalberto (approved):

.github/workflows/docker-test-containers.yml
```diff
@@ -12,7 +12,7 @@ jobs:
         include:
           - source: jaegertracing/all-in-one:1.17
             target_tag: jaeger
-          - source: otel/opentelemetry-collector-dev:latest
+          - source: otel/opentelemetry-collector-contrib-dev:latest
```
jkwatson, line 15: is "latest" a good thing to be using here? We've seen quite a bit of collector drift over the past year, which could end up causing unexpected test failures. Just wondering if we should start targeting a fixed version, rather than always grabbing latest.

jkwatson (approved):

jkwatson:
merging to get builds fixed (🤞🏽 )

PR #3570, opened 2021-08-31
Separate out autoconfigure-spi artifact with tracing autoconfiguratio…

anuraaga:
…n SPI.

I have not marked this artifact as alpha since I think it's ready to roll.

It means the same package is in both autoconfigure-spi and autoconfigure - this may cause JMS issues, but I'm not sure it's a big deal while metrics is alpha. If someone knows it is, I'd probably need to temporarily name the metrics SPIs package `autoconfigure.spi.alpha`


docs/apidiffs/current_vs_latest/opentelemetry-extension-trace-propagators.txt
```diff
@@ -1,2 +1,17 @@
 Comparing source compatibility of  against 
-No changes.
\ No newline at end of file
+***! MODIFIED CLASS: PUBLIC FINAL io.opentelemetry.extension.trace.propagation.B3ConfigurablePropagator  (not serializable)
```
anuraaga, line 2: Oops - sorry I didn't run a full build on the previous PR

jkwatson:
I'm not an expert, but I thought the Java module system would blow up if two modules tried to export the same package and were being loaded by the same classloader, which these would, I think. :(

docs/apidiffs/current_vs_latest/opentelemetry-extension-trace-propagators.txt
jkwatson: shame!

jkwatson:
Since we're breaking all of this, we could have `autoconfigure.spi.traces` and `autoconfigure.spi.metrics` to separate the packages. Thoughts?

anuraaga:
Split up, it seems good and does solve the JMS problem.

jkwatson (approved):
Thanks!

jsuereth (approved):
In for a penny....

If we're breaking SPI let's make sure we do it right :)

PR #3572, opened 2021-09-01
Update to Gradle 7.2

anuraaga:
And try ubuntu-20.04.

@jkwatson I tried incorporating #3568 into this to see how it goes, mainly since I know it won't use the build cache at all.


jkwatson (approved):

anuraaga:
@jkwatson Hmm can confirm I don't see them either. Manually editing the DOM in Chrome inspector didn't work either 😂

jkwatson:
> @jkwatson Hmm can confirm I don't see them either. Manually editing the DOM in Chrome inspector didn't work either 😂

I wonder if merging this might force the issue. We can always manually merge after checking what's passed until that bug is fixed.

anuraaga:
> I wonder if merging this might force the issue. We can always manually merge after checking what's passed until that bug is fixed.

Yeah I had that thought to. Let's try it

anuraaga:
@jkwatson I went ahead and removed the required status check on ubuntu-18.04. And now that can't be restored, searching in the checks only shows Build (macos-*). The GitHub bug appears to be something like only items that are already in the list are presented in the autocompletion...

jkwatson:
Ugh. A little more work for us until that gets fixed. Has anyone logged it with github  yet?

PR #3576, opened 2021-09-03
Preencode attribute keys into utf8 bytes.

anuraaga:
Includes #3575 

Pretty good improvement for the performance sensitive case, where attribute keys are always singletons.

This is net neutral if attribute keys aren't singleton but the user still exports in OTLP, as it's the same number of encodes as we have right now. It does move it to the API hot path though, but it's a fairly tiny operation.

This is only a loss if using non-singleton keys and a non-OTLP exporter in which case the utf8 encodings would be computed unnecessarily. If we write a Jaeger `Marshaler` at some point, the only exporters we have that would fall in this case is Zipkin and Jaeger thrift.

After
```
Benchmark                                                                (numSpans)  Mode  Cnt       Score       Error   Units
RequestMarshalBenchmarks.marshalCustom                                           16  avgt   10      11.181 ±     0.102   us/op
RequestMarshalBenchmarks.marshalCustom:·gc.alloc.rate                            16  avgt   10    1261.679 ±    11.406  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.alloc.rate.norm                       16  avgt   10   22226.885 ±     0.229    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Eden_Space                   16  avgt   10    1258.153 ±    90.851  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Eden_Space.norm              16  avgt   10   22168.051 ±  1703.885    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Survivor_Space               16  avgt   10       0.027 ±     0.010  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Survivor_Space.norm          16  avgt   10       0.477 ±     0.177    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.count                                 16  avgt   10     103.000              counts
RequestMarshalBenchmarks.marshalCustom:·gc.time                                  16  avgt   10      54.000                  ms
RequestMarshalBenchmarks.marshalCustom                                          512  avgt   10     392.017 ±    12.298   us/op
RequestMarshalBenchmarks.marshalCustom:·gc.alloc.rate                           512  avgt   10    1153.881 ±    36.414  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.alloc.rate.norm                      512  avgt   10  711994.909 ±     9.727    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Eden_Space                  512  avgt   10    1154.914 ±    78.755  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Eden_Space.norm             512  avgt   10  712439.936 ± 34116.007    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Survivor_Space              512  avgt   10       0.784 ±     0.110  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Survivor_Space.norm         512  avgt   10     483.333 ±    59.988    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.count                                512  avgt   10     116.000              counts
RequestMarshalBenchmarks.marshalCustom:·gc.time                                 512  avgt   10      66.000                  ms
```

Before
```
Benchmark                                                                (numSpans)  Mode  Cnt       Score       Error   Units
RequestMarshalBenchmarks.marshalCustom                                           16  avgt   10      11.906 ±     0.104   us/op
RequestMarshalBenchmarks.marshalCustom:·gc.alloc.rate                            16  avgt   10    1403.097 ±    12.485  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.alloc.rate.norm                       16  avgt   10   26308.168 ±     0.289    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Eden_Space                   16  avgt   10    1405.722 ±    49.540  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Eden_Space.norm              16  avgt   10   26356.067 ±   785.895    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Survivor_Space               16  avgt   10       0.034 ±     0.005  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Survivor_Space.norm          16  avgt   10       0.637 ±     0.099    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.count                                 16  avgt   10     141.000              counts
RequestMarshalBenchmarks.marshalCustom:·gc.time                                  16  avgt   10      76.000                  ms
RequestMarshalBenchmarks.marshalCustom                                          512  avgt   10     412.197 ±    23.345   us/op
RequestMarshalBenchmarks.marshalCustom:·gc.alloc.rate                           512  avgt   10    1242.080 ±    64.158  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.alloc.rate.norm                      512  avgt   10  805683.147 ±    10.334    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Eden_Space                  512  avgt   10    1244.567 ±    77.255  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Eden_Space.norm             512  avgt   10  807950.201 ± 57913.258    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Survivor_Space              512  avgt   10       0.065 ±     0.066  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Survivor_Space.norm         512  avgt   10      41.924 ±    41.964    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.count                                512  avgt   10     102.000              counts
RequestMarshalBenchmarks.marshalCustom:·gc.time                                 512  avgt   10      77.000                  ms
```


api/all/src/main/java/io/opentelemetry/api/internal/InternalAttributeKeyImpl.java
```diff
@@ -52,13 +58,17 @@ public String getKey() {
     return key;
   }
 
+  public byte[] getKeyUtf8() {
```
jack-berg, line 61: Could you get the best of both worlds by having this extra method, but instead of preencoding it at construction time, lazily encoding it when `getKeyUtf8()` is called, and memoizing the result?

Oberon00:
With this change, the `(String, ValueType)` overloads of the attribute setters (i.e. those that take a String instead of an AttributeKey, such as setAttribute on Span and SpanBuilder) might be worth deprecating, since there is now a significant performance overhead with these.

jkwatson:
> With this change, the `(String, ValueType)` overloads of the attribute setters (i.e. those that take a String instead of an AttributeKey, such as setAttribute on Span and SpanBuilder) might be worth deprecating, since there is now a significant performance overhead with these.

We've already got some pretty strong warnings on these methods about not using them if you don't have to. I don't think we should deprecate them, because they definitely are convenient for people who don't care about hyper performance optimization.

api/all/src/main/java/io/opentelemetry/api/internal/InternalAttributeKeyImpl.java
anuraaga: If it's good enough for Java hashcode, it's good enough for me thanks for the suggestion

api/all/src/main/java/io/opentelemetry/api/internal/InternalAttributeKeyImpl.java
```diff
@@ -3,18 +3,23 @@
...
+package io.opentelemetry.api.internal;
 
+import io.opentelemetry.api.common.AttributeKey;
+import io.opentelemetry.api.common.AttributeType;
+import java.nio.charset.StandardCharsets;
 import javax.annotation.Nullable;
 
-@SuppressWarnings("rawtypes")
-final class AttributeKeyImpl<T> implements AttributeKey<T> {
+/** Default AttributeKey implementation which preencodes to UTF8 for OTLP export. */
```
jkwatson, line 13: Needs the internal class boilerplate here.

api/all/src/main/java/io/opentelemetry/api/internal/InternalAttributeKeyImpl.java
```diff
@@ -52,13 +57,23 @@ public String getKey() {
     return key;
   }
 
+  /** Returns the key, encoded as UTF-8 bytes. */
+  public byte[] getKeyUtf8() {
+    byte[] keyUtf8 = this.keyUtf8;
```
jkwatson, line 62: Are we ok with not having this field be volatile, and just having multiple threads recalculate if necessary? I think that should be fine, since it'll still be better than computing every time on every thread, and avoiding extra memory barriers is also good.

jkwatson (approved):
Generally looks good. Needs a little boilerplate doc, and just one question.

api/all/src/main/java/io/opentelemetry/api/internal/InternalAttributeKeyImpl.java
anuraaga: Yup, and in practice with a BSP that doesn't happen anyways. It's same as String.hashCode

jsuereth (approved):
Starting to look like the protobuf optimisations finding their way into our core domain classes :)

jkwatson:
Needs a rebase, then should be good to go.

PR #3578, opened 2021-09-05
MetricReader and Multi-export features

jsuereth:
Adds MetricReader interface, and expands Java SDK to allow multiple exporters to be registered.

- Implements multi-export for SYNCHRONOUS INSTRUMENTS.  ASYNC work as-is but are currently broken for DELTA aggregation (in a trivial to reproduce way).
- Adds bare minimum tests or evolution of tests.
- Implements the following exporters:
   - OTLP (as-is)
   - InMemory (shifted to use MetricReader)
   - Prometheus (shifted to use MetricReader)
   - PeriodMetricReader (shifted from IntervalMetricReader + adapted to MetricReader interface)
- Updates autoconf to use new style of registration (albeit this can be improved).

## Implementation notes

The key interfaces are as follows:

```
interface MetricProducer {
   Collection<MetricData> collectAllMetrics();
}
interface MetricReader {
    CompletableResultCode flush();
    CompletableResultCode shutdown();
}
interface MetricReaderFactory {
      MetricReaer apply(MetricProducer producer);
}
interface MetricExporter {
   CompletableResultCode export(Collection<MetricData> batch);
   CompletableResultCode flush();
   CompletableResultCode shutdown();
}
interface SdkMeterProviderBuilder {
   SdkMetricProviderBuilder registerMetricReader(MetricReaderFactory readerFactory);
}
```

Effectively this accomplishes the following:

- Shutdown behavior, similar to TracerProvider, is available directly on the MeterProvider.
- SdkMeterProvide can know (with 100% certainty) how many readers there are *and* can influence their lifecycle (via flush/shutdown).

### Tracking metric collection

We introduced `CollectionHandle` which provides bit-set functionality (and integer equality) for tracking metric reader and reseting mutable interval-based state based on metric reader.   Effectively every metric reader is given a `CollectionHandle` which is a bit-value within the integer range (limiting us to 32 metric readers per sdk).   

### SynchronousMetricStorage

- DeltaMeasurementStorage
  - We have a mechanism (on every collection call) to grab current delta measurements from synchronous instruments.
  - Every measurement remembers which collection handles have read the measurement
  - This storage is automatically cleaned once a measurement has been read by all handles.
- CumualtiveMetricStorage
  - We have another mechanism to remember the last reported metric point per-collector

The basic algorithm is:

- Go grab recent delta measurements.
- Merge all unread delta measurements for this collector
- Merge delta-measurement w/ previous cumulative, if reporting cumulative.
- Perform garbage collection

### AsynchronousMetricStorage

This will be updated to track previous values for "mergeDiff" of `Sum` aggregation (or log errors for DELTA temporality depending on specification).


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
```diff
@@ -141,16 +139,17 @@ static MetricExporter configureOtlpMetrics(
...
-    IntervalMetricReaderBuilder readerBuilder =
-        IntervalMetricReader.builder()
-            .setMetricProducers(Collections.singleton(meterProvider))
-            .setMetricExporter(exporter);
+
     Duration exportInterval = config.getDuration("otel.imr.export.interval");
-    if (exportInterval != null) {
-      readerBuilder.setExportIntervalMillis(exportInterval.toMillis());
+    if (exportInterval == null) {
+      // TODO: What default for reading?
```
reyang, line 145: In the current spec PR - "The default value is 5000 (milliseconds)."

jsuereth: Wow, that's quite.... fast.   The default in OpenCensus was 1 minute.   Should we slow it down a bit?

reyang: I don't have strong opinion, I got the value (5000ms) from the [tracing SDK spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#batching-processor) for consistency (and I think consistency here is not something super important).

Note the difference - for span processor, the default 5000ms is the **delay between two consecutive exports**, so the actual interval is something like 5000ms + _exporting time_, in metrics we are taking different approach by using a fixed schedule (so the delay is 5000ms - _exporting time_).

jsuereth: I think there's also a huge difference in batch size + aggregation between metrics + traces.

Traces are known to be high-frequency + highly sampled.   Metrics we're leveraging aggregation to limit  bandwidth, so the export interval *should* be different.    Given what I know of our customers + vendor requests around metrics, I think a default of 1 minute should accommodate most users (and be too fast for some).

reyang: I know that for Microsoft most of the systems use either 20 seconds or 1 minute so I'm totally fine with 1 min.

reyang: I've updated the PR to change it to 60 seconds https://github.com/open-telemetry/opentelemetry-specification/pull/1888/commits/68f4fdc1156ae3a8eed3332006040b8648db32e1.

jkwatson: Hm. Having to wait a full minute, on top of ingest time for the backend is a very long time to wait for metrics to appear on new deployments. What's the harm in having it be faster?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
```diff
@@ -56,14 +70,54 @@ public MeterBuilder meterBuilder(@Nullable String instrumentationName) {
...
-    for (SdkMeter meter : meters) {
-      result.addAll(meter.collectAll(sharedState.getClock().now()));
+  /**
+   * Registers a new metric reader on this meter provider.
+   *
+   * @param <R> the type of the reader.
+   * @param factory a constructor for the reader, given access to SDK internals.
+   * @return the registered reader.
+   */
+  public <R extends MetricReader> R register(MetricReader.Factory<R> factory) {
```
jkwatson, line 80: I would really prefer that we don't allow this mutation to happen after the MeterProvider has been implemented. Why can't this be done up front on the builder?

jsuereth: Need to rework how autoconf module works.  It's possible to do this ahead of time, just need to go deeper on refactoring *and* it'll make tests a bit funkier.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
jsuereth: In OpenCensus there was the agility to have a different delay between when the first metrics are exported and the average time between each export.

My thinking on this is as follows:

- For short-lived processes, the shutdown hook + export is crucial, and one of the benefits of push-metrics.
- For most other metrics, one of the reasons to use metrics OVER logs or traces is because you can greatly shrink the sheer size of the data.   If we report too frequently we put more pressure on backends to collapse the data or risk being very expensive for users.
- For scenarios where we're serving HTTP traffic, if we send too frequently, we're unlikely to have explored/opened all metric-streams within the reporting interval.  This actually has memory/gc implications for the implementation.
- For initial-startup "is it working" scenarios, it probably makes sense to default a quick fire of metrics

TL;DR:

- Our default should balance "Most metric streams have a point recorded during this time" and "Users don't run a huge bill on storage", "Users can see changes live" and "Nice for debugging"
- I think the "is it working on startup" use case is enough to shorten time-to-initial report, but not the on-going reporting interval.  There's some other nice benefits we may be able to get here like restart-loop detection.  However I think the cost of reporting too frequently is a real thing.  Users should opt-in if they want higher resolution.

jsuereth:
@jkwatson @anuraaga This is ready for an "API" evaluation.   Still some known issues in the implementation I'm working through, but would love thoughts on the API before I dive too much further.

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/PrometheusCollector.java
```diff
@@ -5,36 +5,60 @@
...
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 
-public final class PrometheusCollector extends Collector {
-  private final MetricProducer metricProducer;
+public final class PrometheusCollector extends Collector
+    implements MetricReader, MetricReader.Factory {
+  private MetricProducer metricProducer;
```
jkwatson, line 20: `volatile` just to be sure?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/PeriodicMetricReader.java
```diff
@@ -0,0 +1,191 @@
...
+        return;
+      }
+      scheduledFuture =
+          scheduler.scheduleAtFixedRate(
+              scheduled, duration.toMillis(), duration.toMillis(), TimeUnit.MILLISECONDS);
+    }
+  }
+
+  /** Factory for {@link PeriodicMetricReader}. */
+  public static class Factory implements MetricReader.Factory {
```
jkwatson, line 95: pull this out to a top level for consistency with the rest of the SDK.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
```diff
@@ -132,32 +130,34 @@ static MetricExporter configureOtlpMetrics(
...
     Duration exportInterval = config.getDuration("otel.imr.export.interval");
-    if (exportInterval != null) {
-      readerBuilder.setExportIntervalMillis(exportInterval.toMillis());
+    if (exportInterval == null) {
+      exportInterval = Duration.ofMinutes(1);
     }
-    IntervalMetricReader reader = readerBuilder.build().startAndRegisterGlobal();
-    Runtime.getRuntime().addShutdownHook(new Thread(reader::shutdown));
+    // Register the reader (which starts it), and use a global shutdown hook rather than
+    // interval reader specific.
```
jkwatson, line 148: Registering the reader with the builder doesn't start it. It doesn't get started until the SdkMeterProvider is actually built, correct?

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
```diff
@@ -132,32 +130,34 @@ static MetricExporter configureOtlpMetrics(
...
-      ConfigProperties config, SdkMeterProvider meterProvider) {
+      ConfigProperties config, SdkMeterProviderBuilder sdkMeterProviderBuilder) {
     ClasspathUtil.checkClassExists(
         "io.opentelemetry.exporter.prometheus.PrometheusCollector",
         "Prometheus Metrics Server",
         "opentelemetry-exporter-prometheus");
-    PrometheusCollector.builder().setMetricProducer(meterProvider).buildAndRegister();
+    sdkMeterProviderBuilder.register(PrometheusCollector.create());
+    // TODO: Can we move this portion into the PrometheusCollector so shutdown on SdkMeterProvider
+    // will collapse prometheus too?
```
jkwatson, line 160: Not the config reading part, but the server creation/startup makes sense to move there, yes.

sdk-extensions/tracing-incubator/src/jmh/java/io/opentelemetry/sdk/extension/incubator/trace/ExecutorServiceSpanProcessorCpuBenchmark.java
```diff
@@ -38,7 +38,7 @@
 
   @State(Scope.Benchmark)
   public static class BenchmarkState {
-    private MetricProducer collector;
+    private InMemoryMetricReader collector;
```
jkwatson, line 41: Nit: I don't think we should call these variables `collector`, since that's doubly overloaded with prometheus and the otel collector.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/DefaultSdkMeterProvider.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Supplier;
+import java.util.logging.Logger;
+import javax.annotation.Nullable;
+
+/**
+ * {@code SdkMeterProvider} implementation for {@link MeterProvider}.
```
jkwatson, line 31: nit: `MeterProvider` isn't imported here, so the javadoc doesn't resolve the link properly.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/DefaultSdkMeterProvider.java
```diff
@@ -0,0 +1,122 @@
...
+    // Here we construct our own unique handle ids for this SDK.
+    // These are guaranteed to be unique per-reader for this SDK, and only this SDK.
+    Supplier<CollectionHandle> handleSupplier = CollectionHandle.createSupplier();
+    for (MetricReader.Factory readerFactory : readerFactories) {
+      CollectionHandle handle = handleSupplier.get();
+      // TODO: handle failure in creation or just crash?
+      MetricReader reader = readerFactory.apply(new LeasedMetricProducer(handle));
+      collectors.add(handle);
+      readers.add(reader);
+    }
```
jkwatson, line 73: I don't much like doing all this work in the constructor. Can we move this stuff to an `init` method of some sort to be called after construction is complete?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/DefaultSdkMeterProvider.java
```diff
@@ -0,0 +1,122 @@
...
+    List<CompletableResultCode> results = new ArrayList<>();
+    for (MetricReader reader : readers) {
+      results.add(reader.flush());
+    }
+    return CompletableResultCode.ofAll(results);
+  }
+
+  @Override
+  public CompletableResultCode shutdown() {
+    // TODO - prevent multiple calls.
```
jkwatson, line 96: or just ignore subsequent ones, returning the appropriate result code.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
```diff
@@ -5,69 +5,22 @@
...
-
-  SdkMeterProvider(
-      Clock clock, Resource resource, ViewRegistry viewRegistry, ExemplarSampler exemplarSampler) {
-    this.sharedState =
-        MeterProviderSharedState.create(clock, resource, viewRegistry, exemplarSampler);
-    this.registry =
-        new ComponentRegistry<>(
-            instrumentationLibraryInfo -> new SdkMeter(sharedState, instrumentationLibraryInfo));
-  }
+public interface SdkMeterProvider extends MeterProvider {
```
jkwatson, line 17: nit: with the change to an interface, the `builder` method doesn't need `public` on it any more.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/MetricExporter.java
```diff
@@ -19,8 +19,8 @@
 
   /**
    * Exports the collection of given {@link MetricData}. Note that export operations can be
-   * performed simultaneously depending on the type of metric reader being used. However, the {@link
-   * IntervalMetricReader} will ensure that only one export can occur at a time.
+   * performed simultaneously depending on the type of metric reader being used. However, the caller
+   * will ensure that only one export can occur at a time.
```
jkwatson, line 23:
```suggestion
   * MUST ensure that only one export can occur at a time.
```

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/MetricExporter.java
```diff
@@ -30,7 +30,7 @@
   /**
    * Exports the collection of {@link MetricData} that have not yet been exported. Note that flush
    * operations can be performed simultaneously depending on the type of metric reader being used.
-   * However, the {@link IntervalMetricReader} will ensure that only one export can occur at a time.
+   * However, the caller will ensure that only one export can occur at a time.
```
jkwatson, line 33:
```suggestion
   * However, the caller MUST ensure that only one export can occur at a time.
```

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/MetricReader.java
```diff
@@ -0,0 +1,55 @@
...
+import io.opentelemetry.sdk.common.CompletableResultCode;
+
+/**
+ * A registered reader of metrics.
+ *
+ * <p>This interface provides the {@link io.opentelemetry.sdk.metrics.SdkMeterProvider} a mechanism
+ * of global control over metrics during shutdown or memory pressure scenarios.
+ */
+public interface MetricReader {
+  /**
```
jkwatson, line 17:
```suggestion

  /**
```

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/MetricReader.java
```diff
@@ -0,0 +1,55 @@
...
+   * <p>For push endpoints, this should collect and report metrics as normal.
+   *
+   * <p>For pull endpoints, this may be called due to memory pressure or during shutdown of the
+   * application. Delta metrics collected in this way may be dropped, and the {@link MetricReader}
+   * is expected to handle this scenario.
+   *
+   * @return the result of the shutdown.
+   */
+  CompletableResultCode flush();
+  /**
```
jkwatson, line 32:
```suggestion
  
  /**
```

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/MetricReader.java
```diff
@@ -0,0 +1,55 @@
...
+   *
+   * <p>For push endpoints, this should shut down any scheduler threads.
+   *
+   * @return the result of the shutdown.
+   */
+  CompletableResultCode shutdown();
+
+  /** A constructor of {@link MetricReader}s. */
+  @FunctionalInterface
+  interface Factory {
```
jkwatson, line 46: Pull this up to the top level for consistency with other SDK classes/interfaces.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/MetricReader.java
```diff
@@ -0,0 +1,55 @@
...
+ * A registered reader of metrics.
+ *
+ * <p>This interface provides the {@link io.opentelemetry.sdk.metrics.SdkMeterProvider} a mechanism
+ * of global control over metrics during shutdown or memory pressure scenarios.
+ */
+public interface MetricReader {
+  /**
+   * Flushes metrics read by this reader.
+   *
+   * <p>In all scenarios, the {@link MetricProducer} should have its {@link
```
jkwatson, line 20:
```suggestion
   * <p>In all scenarios, the associated {@link MetricProducer} should have its {@link
```

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/export/CollectionHandle.java
```diff
@@ -0,0 +1,160 @@
...
+    this.index = index;
+  }
+
+  /** Construct a new (efficient) mutable set for tracking collection handles. */
+  public static Set<CollectionHandle> mutableSet() {
+    return new CollectionHandleSet();
+  }
+
+  /** Construct a new (mutable) set consistenting of the passed in collection handles. */
+  public static Set<CollectionHandle> of(CollectionHandle... handles) {
```
jkwatson, line 40: looks to be only used for testing, so we can make it non-public and notate that it's for testing

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/export/CollectionHandle.java
```diff
@@ -0,0 +1,160 @@
...
+    return index == ((CollectionHandle) other).index;
+  }
+
+  @Override
+  public String toString() {
+    return "CollectionHandle(" + index + ")";
+  }
+
+  // TODO: Better mechanism of this?
+  private static final AtomicInteger nextIndex = new AtomicInteger(1);
```
jkwatson, line 95: move this up to the top of the class, please. :) Also, why not start at zero, and call `incrementAndGet()` when needed?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
```diff
@@ -124,7 +145,11 @@ private AsynchronousMetricStorage(
   }
 
   @Override
-  public MetricData collectAndReset(long startEpochNanos, long epochNanos) {
+  public MetricData collectAndReset(
+      CollectionHandle collector,
+      Set<CollectionHandle> allCollectors,
```
jkwatson, line 150: it's a little weird, from an API perspective that these need to be passed in, but aren't ever used.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorage.java
```diff
@@ -25,5 +27,9 @@
    * @param epochNanos The timestamp for this collection.
    * @return The {@link MetricData} from this collection period, or {@code null}.
    */
-  MetricData collectAndReset(long startEpochNanos, long epochNanos);
+  MetricData collectAndReset(
+      CollectionHandle collector,
+      Set<CollectionHandle> allCollectors,
```
jkwatson, line 32: Two new params could use a bit of javadoc (yeah, it's internal...but I'm trying to figure it all out, so it would probably be helpful)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/SynchronousMetricStorage.java
```diff
@@ -55,23 +68,20 @@
                 metricDescriptor,
                 () -> sampler.createReservoir(view.getAggregation()));
     return new SynchronousMetricStorage<>(
-        metricDescriptor,
-        aggregator,
-        new InstrumentProcessor<>(aggregator, startEpochNanos),
-        view.getAttributesProcessor());
+        metricDescriptor, aggregator, view.getAttributesProcessor());
```
jkwatson, line 71: `startEpochNanos` parameter is no longer used, and can be removed.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/SynchronousMetricStorage.java
```diff
@@ -55,23 +68,20 @@
...
   }
 
   SynchronousMetricStorage(
       MetricDescriptor metricDescriptor,
       Aggregator<T> aggregator,
-      InstrumentProcessor<T> instrumentProcessor,
       AttributesProcessor attributesProcessor) {
     this.metricDescriptor = metricDescriptor;
     aggregatorLabels = new ConcurrentHashMap<>();
     collectLock = new ReentrantLock();
```
jkwatson, line 80: this could be moved up to the declaration, yes?

jkwatson: (and the aggregatorLabels as well)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/SynchronousMetricStorage.java
```diff
@@ -154,30 +164,173 @@ public void recordDouble(double value, Attributes attributes, Context context) {
...
+  /**
+   * This method leverages the `savedDeltas` and `reportHistory` to construct a final metric point
+   * to report for this synchronous instrument.
+   */
+  @GuardedBy("collectLock")
+  @Nullable
+  private MetricData buildMetricFor(
+      CollectionHandle collector, long startEpochNanos, long epochNanos) {
+    Map<Attributes, T> result = new HashMap<>();
+    // Next merge the delta w/ the last set of points.
```
jkwatson, line 180: I think some comments got out of order, since "next" is now before "first". :)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/SynchronousMetricStorage.java
```diff
@@ -154,30 +164,173 @@ public void recordDouble(double value, Attributes attributes, Context context) {
...
+      }
+    }
+    // TODO: Do we allow different pipelines?
+    Iterator<CollectionHandle> c = reportHistory.keySet().iterator();
+    while (c.hasNext()) {
+      if (!collectors.contains(c.next())) {
+        c.remove();
+      }
+    }
+  }
```
jkwatson, line 238: IDEA has some nifty suggestions for simplifying this code with the `Iterator.removeIf()` method.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/SynchronousMetricStorage.java
```diff
@@ -154,30 +164,173 @@ public void recordDouble(double value, Attributes attributes, Context context) {
...
+    return metricDescriptor;
+  }
+
+  /**
+   * This method leverages the `savedDeltas` and `reportHistory` to construct a final metric point
+   * to report for this synchronous instrument.
+   */
+  @GuardedBy("collectLock")
+  @Nullable
+  private MetricData buildMetricFor(
```
jkwatson, line 177: let's move this method down below where it's called.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/SynchronousMetricStorage.java
```diff
@@ -154,30 +164,173 @@ public void recordDouble(double value, Attributes attributes, Context context) {
...
+    while (c.hasNext()) {
+      if (!collectors.contains(c.next())) {
+        c.remove();
+      }
+    }
+  }
+
+  /** Collects bucketed metrics and resets the underlying storage for the next collection period. */
+  @Override
+  public MetricData collectAndReset(
```
jkwatson, line 242: this is definitely a hairy method in a hairy class. We'll want to figure out how to refactor/cleanup to make it processable by normal humans at some point.

jkwatson: (I see you've made some attempts at this already, so no criticism intended)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/SynchronousMetricStorage.java
jkwatson: also, what would you think about renaming it to `buildMetricsFor` ?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/testing/InMemoryMetricReader.java
```diff
@@ -0,0 +1,73 @@
...
+ * }
+ * </code></pre>
+ */
+public class InMemoryMetricReader implements MetricReader, MetricReader.Factory {
+  private volatile MetricProducer sdkCollection;
+  private volatile Collection<MetricData> latest = Collections.emptyList();
+
+  public InMemoryMetricReader() {
+    // Need to wait for registration for initial value.
+  }
```
jkwatson, line 47: this constructor declaration isn't needed, right? Or will you be filling in more stuff later?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/testing/InMemoryMetricReader.java
```diff
@@ -0,0 +1,73 @@
...
+ *
+ *   public static void main(String[] args) {
+ *     InMemoryMetricReaderExample example = new InMemoryMetricReaderExample();
+ *     example.printMetrics();
+ *   }
+ * }
+ * </code></pre>
+ */
+public class InMemoryMetricReader implements MetricReader, MetricReader.Factory {
+  private volatile MetricProducer sdkCollection;
```
jkwatson, line 42: strange name for this field. Maybe just "metricProducer"?

jkwatson:
This generally seems reasonable. A few naming bits here & there, and I notice that the test coverage of SynchronousMetricReader is a little spotty right now.

The MeterReaderFactory `apply` pattern is definitely a bit of a deviation from other patterns in the code, so that's something to ponder if there's an alternate approach, but I think we can iterate on that, for sure.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
jsuereth: Yeah, I'll update the comments.

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/PrometheusCollector.java
jsuereth: Shouldn't be necessary as this will be filled out before the class is registered w/ prometheus.  

We can add it, or I can document the reason it's ok not to be volatile here.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/DefaultSdkMeterProvider.java
jsuereth: Would this be called immediately after construction always anyway?

If we want immutable / final references to this config, it needs to be in the constructor.  I moved it this way to avoid having a "register" method be publicly exposed the could dynamically adjust configuration (and require locks).   We can shift this around, but having another method that mutates state feels just as bad as exposing "register" to me.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/DefaultSdkMeterProvider.java
jsuereth: Right, do we have an example of that impl I can copy-paste?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
jsuereth: doh!  I'll learn java one of these days.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/MetricReader.java
jsuereth: Right, had forgotten to do that, will also remove `@FunctionalInterface`.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
jsuereth: aren't used.... yet.


I need to solve async instrument DELTA temporality.  Java (today) has some bugs around DELTA metrics, but it does *try* to do delta sums, where you need to remember last-reported value per-collector.  `allCollectors` was there for dynamic registration and cleanup, but can be removed now that we aren't doing that.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorage.java
jsuereth: Definitely! This is why I asked for early review, will update javadocs to make this more clear.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/SynchronousMetricStorage.java
jsuereth: Actually..... I had "one big method" i refactored based on my ordering comments.  Will fix :)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/SynchronousMetricStorage.java
jsuereth: I'm ok renaming....

Plurality here is hard.  It's one metric-data, but has multiple streams, meh.   The other question I have is should we move this to Collection<MetricData>?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/SynchronousMetricStorage.java
jsuereth: totally agree!   If you help me with naming, I'll work on chunking this out into better pieces.   

To some extent, I think cleaning up the Aggregator interface can help a bit here.  To another extent, I went crazy with private data classes that could likely be better named.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/testing/InMemoryMetricReader.java
jsuereth: You're right, it's not.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/testing/InMemoryMetricReader.java
jsuereth: Can refactor, also I learned I can drop the volatile here, but will leave pending your comments on the prometheus variant.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
jkwatson: It would be really cool to have a dynamic system that would send metrics more frequently during unusual circumstances (like startup), but I have no idea how one might do that. :)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProviderBuilder.java
```diff
@@ -107,6 +111,17 @@ public SdkMeterProvider buildAndRegisterGlobal() {
     return meterProvider;
   }
 
+  /**
+   * Registers a {@link MetricReader} for this SDK.
+   *
+   * @param reader The factory for a reader of metrics.
+   * @return this
+   */
+  public SdkMeterProviderBuilder register(MetricReader.Factory reader) {
```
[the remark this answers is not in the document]
jsuereth: Fixed.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/PeriodicMetricReader.java
jsuereth: fixed

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/PrometheusCollector.java
jsuereth: Added documentation on why, PTAL and let me know if you'd still like volatile here.

sdk-extensions/tracing-incubator/src/jmh/java/io/opentelemetry/sdk/extension/incubator/trace/ExecutorServiceSpanProcessorCpuBenchmark.java
jsuereth: Tried to fix this across all these benchmarks.  I'm thinking there's a huge opportunity for some code sharing in these Span-processor benchmarks, but perhaps in a future PR.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/MetricReader.java
jsuereth: Fixed

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/export/CollectionHandle.java
jsuereth: Done.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorage.java
jsuereth: Added javadooc

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/DefaultSdkMeterProvider.java
jkwatson: heh. I looked through the code and found a variety of approaches, none of which look great. It will probably involve saving off the CompletableResultCode somewhere, and just returning it if it's already there.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/SynchronousMetricStorage.java
jkwatson: I think we can work on it over time. I don't think we necessarily need to block on this internal cleanup.

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/PrometheusCollector.java
jkwatson: Hm. I read your description and I'm not convinced. :) The JMM is a tricky beast and visibility of non-final, non-volatiles is likely to trip one up at the worst possible time (in production). Is "shared with other threads" a concrete JMM thing for non-final, non-volatiles?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/SynchronousMetricStorage.java
jsuereth: I actually came up with something I kind of like. PTAL.

TL;DR:

- DeltaStorage keeps "live" synchronous aggregations and dumps delta-measurements on demand.
- TemporalStorage remembers "last" reported point for (used awkwardly for both delta + cumulative)
- MetricStorage classes now focus on pulling all components together, e.g. AttributesProcessor.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/SynchronousMetricStorage.java
jsuereth: Done.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/SynchronousMetricStorage.java
jsuereth: nice! done.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/export/CollectionHandle.java
jsuereth: Removed

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/SynchronousMetricStorage.java
jsuereth: Removed

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/PeriodicMetricReaderFactory.java
```diff
@@ -0,0 +1,68 @@
...
+
+package io.opentelemetry.sdk.metrics.export;
+
+import io.opentelemetry.sdk.internal.DaemonThreadFactory;
+import java.time.Duration;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+
+/** Factory for {@link PeriodicMetricReader}. */
+public class PeriodicMetricReaderFactory implements MetricReaderFactory {
```
anuraaga, line 14: For our general pattern we'd generally make this have a package private constructor and have `PeriodicMetricReader.createFactory()` return it. Assuming we prefer that over static methods on MetricReaderFactory not clear for this one

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/export/CollectionHandle.java
```diff
@@ -0,0 +1,156 @@
...
+ */
+public final class CollectionHandle {
+  /** The index of this handle. */
+  private final int index;
+
+  private CollectionHandle(int index) {
+    this.index = index;
+  }
+
+  /** Construct a new (efficient) mutable set for tracking collection handles. */
```
anuraaga, line 34: @jsuereth I thought you said something about complete and working and optimizing in followups :P Just messing

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/export/CollectionHandle.java
```diff
@@ -0,0 +1,156 @@
...
+      return false;
+    }
+
+    @Override
+    public boolean containsAll(Collection<?> other) {
+      if (other instanceof CollectionHandleSet) {
+        BitSet result = new BitSet();
+        BitSet otherStorage = ((CollectionHandleSet) other).storage;
+        result.or(storage);
+        result.and(otherStorage);
```
anuraaga, line 125:
```suggestion
        BitSet result = storage.clone().and(((CollectionHandleSet) other).storage);
```

`BitSet.clone()` even has javadoc!

anuraaga (commented):
Did first run through to the middle

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/NoopSdkMeterProvider.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package io.opentelemetry.sdk.metrics;
+
+import io.opentelemetry.api.metrics.MeterBuilder;
+import io.opentelemetry.api.metrics.MeterProvider;
+import io.opentelemetry.sdk.common.CompletableResultCode;
+
+/** Implementation of SdkMeterProvider which does not collect metrics. */
+final class NoopSdkMeterProvider implements SdkMeterProvider {
```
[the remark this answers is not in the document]
jsuereth: There's a bunch of places that expect and SdkMeterProvider, but we want to disable collection of metrics when there are no registered metric-readers.   For me, this was the cleanest way to update all the code and keep it consistent.


Specficially `forceFlush` and `close` are only exposed on the providers.   We also don't really expose a mechanism to interact with them via GlobalOpenTelemetry, so actually I'd suspect we should think about exposing `forceFlush` in some fashion like this going forward.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/DefaultSdkMeterProvider.java
```diff
@@ -0,0 +1,118 @@
...
+    LeasedMetricProducer(CollectionHandle handle) {
+      this.handle = handle;
+    }
+
+    @Override
+    public Collection<MetricData> collectAllMetrics() {
+      Collection<SdkMeter> meters = registry.getComponents();
+      List<MetricData> result = new ArrayList<>(meters.size());
+      for (SdkMeter meter : meters) {
+        result.addAll(meter.collectAll(handle, collectors, sharedState.getClock().now()));
```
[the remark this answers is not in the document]
jsuereth: This is what I was suggesting around optimising the collection path.

I agree this isn't efficient (In Scala, I'd also be testing this).  However I wasn't able to get a (good) benchmark together to test this aspect of metric collection yet.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/MetricReader.java
```diff
@@ -0,0 +1,45 @@
...
+
+  /**
+   * Flushes metrics read by this reader.
+   *
+   * <p>In all scenarios, the associated {@link MetricProducer} should have its {@link
+   * MetricProducer#collectAllMetrics()} method called.
+   *
+   * <p>For push endpoints, this should collect and report metrics as normal.
+   *
+   * <p>For pull endpoints, this may be called due to memory pressure or during shutdown of the
```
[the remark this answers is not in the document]
jsuereth: This is copied from the spec, and it's for us to decide.

My $.02 is that we should ensure (specifically on pull exporters) that `.forceFlush` has the ability to clean used memory without allocations going forward.   That'll be a 'fun' set of refactorings, and on my longer-term list.  I think solving high-cardinality memory pressure is way higher priority than this particular aspect of forceFlush.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/PeriodicMetricReader.java
```diff
@@ -0,0 +1,134 @@
...
+            } else {
+              result.succeed();
+            }
+          });
+    }
+    return result;
+  }
+
+  void start(Duration duration) {
+    // Autoconfigure sends us null or 0 durations and expects us to just not start.
```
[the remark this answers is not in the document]
jsuereth: Yes.  Do you want it part of this PR?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/PeriodicMetricReaderFactory.java
jsuereth: I'm fine with that pattern here.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/export/CollectionHandle.java
jsuereth: I wrote this >6 months ago.   Haven't actually tested it in the current incarnation we have in this SDK, so we'll need to rework the `*Storage` classes to make sure this is actually used well :)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/export/CollectionHandle.java
jsuereth: bad habits of never trusting clone...

Also, clone returns `Object` so you have to cast here.

jsuereth: Fixed.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/PeriodicMetricReader.java
```diff
@@ -0,0 +1,134 @@
...
+ */
+public class PeriodicMetricReader implements MetricReader {
+  private static final Logger logger = Logger.getLogger(PeriodicMetricReader.class.getName());
+
+  private final MetricProducer producer;
+  private final MetricExporter exporter;
+  private final ScheduledExecutorService scheduler;
+
+  private final Scheduled scheduled;
+  @Nullable private volatile ScheduledFuture<?> scheduledFuture;
```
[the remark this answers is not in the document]
jsuereth: Done.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/PeriodicMetricReader.java
```diff
@@ -0,0 +1,134 @@
...
+
+  @Override
+  public CompletableResultCode flush() {
+    return scheduled.doRun();
+  }
+
+  @Override
+  public CompletableResultCode shutdown() {
+    final CompletableResultCode result = new CompletableResultCode();
+    if (scheduledFuture != null) {
```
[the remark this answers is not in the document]
jsuereth: Changed.

jsuereth:
Not sure how to solve the codecov issues, it seems like it's not seeing much coverage on existing exporters + autoconfigure.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/testing/InMemoryMetricReader.java
```diff
@@ -0,0 +1,74 @@
...
+ *
+ *   public static void main(String[] args) {
+ *     InMemoryMetricReaderExample example = new InMemoryMetricReaderExample();
+ *     example.printMetrics();
+ *   }
+ * }
+ * </code></pre>
+ */
+public class InMemoryMetricReader implements MetricReader, MetricReaderFactory {
+  // Note: we expect the `apply` method of `MetricReaderFactory` to be called
```
anuraaga, line 43: Even for testing think it's worth sticking to the same pattern we did to Prometheus (here it really should just be returning `::new` :)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DeltaMetricStorage.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  /**
+   * Collects the currently accumulated measurements from the concurrent-friendly synchronous
+   * storage.
+   *
+   * <p>All synchronous handles will be collected + reset during this method. Additionally cleanup
+   * related stale concurrent-map handles will occur. Any {@code null} measurements are ignored.
+   */
+  @GuardedBy("this")
```
anuraaga, line 100: I think this is supposed to be on fields, not on methods (and errorprone will ensure the marked fields are only accessed under said lock)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DeltaAccumulation.java
```diff
@@ -0,0 +1,39 @@
...
+class DeltaAccumulation<T> {
+  private final Map<Attributes, T> recording;
+  private final Set<CollectionHandle> readers;
+
+  DeltaAccumulation(Map<Attributes, T> recording) {
+    this.recording = recording;
+    this.readers = CollectionHandle.mutableSet();
+  }
+
+  boolean wasReadBy(CollectionHandle handle) {
```
anuraaga, line 27: Would help to understand the lifecycle by adding some javadocs here

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DeltaAccumulation.java
```diff
@@ -0,0 +1,39 @@
...
+    return readers.contains(handle);
+  }
+
+  boolean wasReadyByAll(Set<CollectionHandle> handles) {
```
anuraaga, line 31: Oh - well one reason I got confused was this should be `wasReadByAll` I guess

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DeltaMetricStorage.java
```diff
@@ -0,0 +1,138 @@
...
+    collectSynchronousDeltaAccumulationAndReset();
+    // Now build a delta result.
+    Map<Attributes, T> result = new HashMap<>();
+    for (DeltaAccumulation<T> point : unreportedDeltas) {
+      if (!point.wasReadBy(collector)) {
+        mergeInPlace(result, point.read(collector), aggregator);
+      }
+    }
+    // Now run a quick cleanup of deltas before returning.
+    unreportedDeltas.removeIf(delta -> delta.wasReadyByAll(collectors));
```
anuraaga, line 89: IIUC, one of the most significant changes from our current model is that rather than clearing all data every collection, we mark each time data has been read, and only after it's been read by all collectors do we clean up. Just want to confirm my understanding

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/testing/InMemoryMetricReader.java
jsuereth: We can't, you loose access to the instantiated class if you do so, which means you never have access to collect metrics.

This is designed so you can directly call "collect" and extract metrics.  (also, should this class be in the metric-testing artifact?)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DeltaMetricStorage.java
jsuereth: This did work for me in VisualStudio, but I can remove if you'd like.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DeltaMetricStorage.java
jsuereth: Yes.   We store *semi-aggregated* data every time any reader wants to collect metrics, then we reconstitute the full-delta per-reader and join w/ any cumulative storage before reporting.    There's a design "markdown" from Reiley that kind of clarifies this, and I can take a crack at something.  Hoping the new naming helps make this more clear.

Effectively the goal of this class is to get the Synchronous collection path (that's highly performance sensitive) into the reader thread as fast as possible.  We don't care too much about reader contention because we don't expect significant numbers of reader threads.

A few improvements over time:

- I'd like to pre-allocate delta storage and limit its size.  That should be hidden in this class and would be per-metric.
- I'd like to allow multiple reader "within interval" to avoid re-collecting synchronous metrics.
- I'd like to join multiple `MetricExporter`s into a single `MetricReader` if they all use periodic metric reader.

I can open tickets around these improvements, but wanted to make sure we're ok with how this code works now.

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/PrometheusCollector.java
```diff
@@ -21,20 +34,45 @@
 
   @Override
   public List<MetricFamilySamples> collect() {
+    if (metricProducer == null) {
```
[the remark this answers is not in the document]
jsuereth: Ah, this shouldn't remain, after the refactoring anyway.  It "can't" be null if all our `@Nullable` annotations are accurage.

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/PrometheusCollector.java
```diff
@@ -21,20 +34,45 @@
...
+  }
+
+  @Override
+  public CompletableResultCode shutdown() {
+    // TODO - Can we unsubscribe ourselves from callbacks?
+    return CompletableResultCode.ofSuccess();
+  }
+
+  /** Our implementation of the metric reader factory. */
+  private static class Factory implements MetricReaderFactory {
```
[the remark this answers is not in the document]
jsuereth: There's a refactoring (I can do in this PR or later) where we instantiate the HTTP server as part of the factory (configurable).   Would you like that in this PR? I'll add a note in the meantime.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/DefaultSdkMeterProvider.java
```diff
@@ -0,0 +1,118 @@
...
+ * OpenTelemetry}.
+ */
+final class DefaultSdkMeterProvider implements SdkMeterProvider {
+
+  private static final Logger LOGGER = Logger.getLogger(DefaultSdkMeterProvider.class.getName());
+  static final String DEFAULT_METER_NAME = "unknown";
+  private final ComponentRegistry<SdkMeter> registry;
+  private final MeterProviderSharedState sharedState;
+
+  // These are *only* mutated in our constructor, and safe to use concurrently after construction.
```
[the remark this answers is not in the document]
jsuereth: Fixed

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DeltaAccumulation.java
jsuereth: Ah, fixed.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DeltaAccumulation.java
jsuereth: added

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/NoopSdkMeterProvider.java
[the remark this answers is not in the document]
anuraaga: There was a strong position in the specs (sorry a bit lazy to search for links right now but let me know if not discoverable) to keep these methods out of the API no matter what. I and others (at least I remember JS mentioning this) do think this should be in the API so we are on the same page on that front.

I think it means we agree that to merge this before spec (anti-)discussions we should use instanceof to keep with the status quo for now and fix forward.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DeltaMetricStorage.java
anuraaga: Working means if a not synchronized method accesses the ArrayList without taking the lock. Is vscode really able to understand that with this pattern? And is Gradle able to? The requirement is for Gradle to catch it, and if it can then I'm missing parts of the behavior of this annotation so happy to know that.

docs/apidiffs/current_vs_latest/opentelemetry-exporter-logging-otlp.txt
```diff
@@ -1,2 +1,5 @@
 Comparing source compatibility of  against 
-No changes.
\ No newline at end of file
+***! MODIFIED CLASS: PUBLIC FINAL io.opentelemetry.exporter.logging.otlp.OtlpJsonLoggingMetricExporter  (not serializable)
```
jkwatson, line 2: eek. I'm guessing we messed up here and our metrics exporter ended up in an otherwise stable module

docs/apidiffs/current_vs_latest/opentelemetry-exporter-logging-otlp.txt
```diff
@@ -1,2 +1,5 @@
...
+	===  CLASS FILE FORMAT VERSION: 52.0 <- 52.0
+	---! REMOVED METHOD: PUBLIC(-) io.opentelemetry.sdk.common.CompletableResultCode flush()
```
jkwatson, line 4: Hrm. Since this is in a stable module, should we keep this method and just deprecated it?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/MetricExporter.java
```diff
@@ -30,11 +30,11 @@
   /**
    * Exports the collection of {@link MetricData} that have not yet been exported. Note that flush
    * operations can be performed simultaneously depending on the type of metric reader being used.
-   * However, the {@link IntervalMetricReader} will ensure that only one export can occur at a time.
+   * However, the caller MUST ensure that only one export can occur at a time.
    *
    * @return the result of the flush, which is often an asynchronous operation.
    */
-  CompletableResultCode flush();
+  CompletableResultCode forceFlush();
```
jkwatson, line 37: I'm against renaming this method. It breaks our existing API compatibility guarantees and makes the method out of alignment with the `SpanExporter` interface.  Is there a strong reason why we're adding the "force" to it?

docs/apidiffs/current_vs_latest/opentelemetry-exporter-logging-otlp.txt
jsuereth: Ah, I was kinda curious this showed up, I figured since it was the logging exporter that wasn't stable yet.  I'll re-add and deprecate.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/MetricExporter.java
jsuereth: I think maybe I'm misunderstanding some comments/conventions here.

I migrated all uses of `flush` to `forceFlush` to conform with SdkTracerProvider.   It looks like exporters use `flush` and the `SDK*Provider` uses `forceFlush`.

I'll revert this specific change.  Do we want SdkMeterProivider to exposes `forceFlush` or just `flush`?  From my perspective each is equally good, so if `flush` was exposed in a stable API, happy to change this.

jsuereth: Original comment (I likely took way to far): https://github.com/open-telemetry/opentelemetry-java/pull/3578#discussion_r715309363


For clarity on what we have:

- Exporter has:
  - `flush()`
  - `shutdown()`
- Sdk*Provider has:
  - `forceFlush`
  - `close()`
  - `shutdown()`
  
  If we're on board, I'll make sure metrics matches traces with that.

jkwatson: yeah, for good or ill, let's at least have the parallel interfaces have the same methods. I honestly don't remember why exporters got `flush` when the providers got `forceFlush`.

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/PrometheusCollector.java
```diff
@@ -26,15 +36,39 @@
...
+  }
+
+  @Override
+  public CompletableResultCode shutdown() {
+    CollectorRegistry.defaultRegistry.unregister(this);
+    return CompletableResultCode.ofSuccess();
+  }
+
+  /** Our implementation of the metric reader factory. */
+  // NOTE: This should be updated to (optionally) start the simpel Http server exposing the metrics
```
jkwatson, line 63:
```suggestion
  // NOTE: This should be updated to (optionally) start the simple Http server exposing the metrics
```

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/PeriodicMetricReader.java
```diff
@@ -0,0 +1,177 @@
...
+  public static MetricReaderFactory create(
+      MetricExporter exporter, Duration duration, ScheduledExecutorService scheduler) {
+    return new PeriodicMetricReaderFactory(exporter, duration, scheduler);
+  }
+
+  private final MetricProducer producer;
+  private final MetricExporter exporter;
+  private final ScheduledExecutorService scheduler;
+  private final Scheduled scheduled;
+  private final Object lock = new Object();
```
jkwatson, line 72: field declarations should go above static methods. You've been poisoned by @anuraaga !

jkwatson (approved):
This seems good enough to merge and polish as we go. :shipit:

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/PrometheusCollector.java
jsuereth: fixed

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/PeriodicMetricReader.java
jsuereth: Ha! I miss Scala where field declarations are the constructor.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DeltaMetricStorage.java
anuraaga: By the way, think I'm just waiting on this one to merge

anuraaga: Or we can merge after filing an issue about it if it seems hairy, since it's about thread safety want to make sure it doesn't get lost.

anuraaga (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DeltaMetricStorage.java
jsuereth: Didn't realize this was still open (sorry soo many comments I was missing things).  Just went with synchronized for now and we can rework the class internals over time (there's a few TODOs for it anyway).

PR #3579, opened 2021-09-06
Catch exceptions when loading a configurable SPI.

anuraaga:
In https://github.com/aws-observability/aws-otel-java-instrumentation/issues/77 I noticed that it's possible for a bad SPI implementation to prevent others from being usable since we load them eagerly without catching exceptions. I think it's worth making sure that one bad SPI doesn't remove the flag completely so this catches and logs a failure. If the bad SPI was actually specified for the flag, it would result in a ConfigurationException still as we'd want.

I think for the non-configurable SPIs (*ProviderConfigurer) it seemed still fine to bubble the exception as they are not just an option out of many but code that needs to be executed.

### DIFF
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/SpiUtil.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/SpiUtil.java
@@ -21,7 +21,7 @@ final class SpiUtil {
 
   static <T, U> Map<String, T> loadConfigurable(
       Class<U> spiClass,
-      List<String> requested,
+      List<String> requestedNames,
       Function<U, String> getName,
       BiFunction<U, ConfigProperties, T> getConfigurable,
       ConfigProperties config) {
@@ -32,7 +32,7 @@ final class SpiUtil {
       try {
         configurable = getConfigurable.apply(provider, config);
       } catch (Throwable t) {
-        Level level = requested.contains(name) ? Level.WARNING : Level.FINE;
+        Level level = requestedNames.contains(name) ? Level.WARNING : Level.FINE;
         logger.log(
             level, "Error initializing " + spiClass.getSimpleName() + " with name " + name, t);
         continue;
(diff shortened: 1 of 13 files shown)


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/SpiUtil.java
```diff
@@ -0,0 +1,46 @@
...
+final class SpiUtil {
+
+  private static final Logger logger = Logger.getLogger(SpiUtil.class.getName());
+
+  static <T, U> Map<String, T> loadConfigurable(
+      Class<U> spiClass,
+      List<String> requested,
+      Function<U, String> getName,
+      BiFunction<U, ConfigProperties, T> getConfigurable,
+      ConfigProperties config) {
```
jkwatson, line 27: now *that's* what I call a method signature.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/SpiUtil.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+final class SpiUtil {
+
+  private static final Logger logger = Logger.getLogger(SpiUtil.class.getName());
+
+  static <T, U> Map<String, T> loadConfigurable(
+      Class<U> spiClass,
+      List<String> requested,
```
jkwatson, line 24: maybe "requestedNames"?

jkwatson:
Is this being tested by providing throwing-implementations and just making sure the existing tests pass? Don't we need to request the `throwing` instances somewhere in the tests?

jkwatson:
> > Is this being tested by providing throwing-implementations and just making sure the existing tests pass? Don't we need to request the throwing instances somewhere in the tests?
> 
> Yeah wanted to test mainly that these don't interferece with correct flags (the bad flags will fail both before and after this PR). Do you think we need explicit tests for them?

But without specifying that you want them to be loaded, don't we not test the new code path(s) at all?

anuraaga:
@jkwatson We do! Since we eagerly load into a map as long as anything is specified.

jkwatson:
> @jkwatson We do! Since we eagerly load into a map as long as anything is specified. I added the throwers before the logic to see that several tests failed before the logic.

Ah, I see that now. Should we be bothering to load things that haven't been requested? I guess we need to for Resources, but would it be more efficient to just skip things that haven't been requested, even if they're on the classpath? Not sure I feel strongly about it...just thinking out loud.

jkwatson (approved):

anuraaga:
@jkwatson Yeah considered that too, think we might follow up with that.

PR #3580, opened 2021-09-06
Refactor attribute marshaling to use standard marshalers

anuraaga:
I had originally thought that marshaling attribute values would need special handling for binary vs json, because proto3 handles it's standard `Value` type specially. But I realized we don't use it, we have our own `AnyValue` type. So the JSON will follow all the standard rules of proto and doesn't need special casing.

I've refactored the keyvalue marshaling to follow the same pattern as all the other marshalers instead of having type-specific implementations of yet another abstract class. This will make it simpler to encode in JSON. It means some more allocation - it's just one more wrapper for the normal primitives. It is a heavy-weight array for list values - I think list values in spans (not resource which is preserialized) is not a common enough use case to worry too much. In fact, the protocol prevents efficient handling of them anyways - it only supports `repeated AnyValue` but should actually have type-specific array types with `repeated long`, etc fields to allow for packed encoding which is waaay cheaper than a repeated message.

Benchmark does exercise the array values, which is where the majority of performance degredation comes in. I'd probably remove them from the bench in the future.

After
```
Benchmark                                                                (numSpans)  Mode  Cnt       Score      Error   Units
RequestMarshalBenchmarks.marshalCustom                                           16  avgt   10      13.613 ±    0.183   us/op
RequestMarshalBenchmarks.marshalCustom:·gc.alloc.rate                            16  avgt   10    1236.334 ±   18.023  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.alloc.rate.norm                       16  avgt   10   26500.216 ±    0.323    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Eden_Space                   16  avgt   10    1245.134 ±   80.040  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Eden_Space.norm              16  avgt   10   26686.319 ± 1564.524    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Survivor_Space               16  avgt   10       0.010 ±    0.004  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Survivor_Space.norm          16  avgt   10       0.210 ±    0.091    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.count                                 16  avgt   10     125.000             counts
RequestMarshalBenchmarks.marshalCustom:·gc.time                                  16  avgt   10      77.000                 ms
RequestMarshalBenchmarks.marshalCustom                                          512  avgt   10     470.428 ±    4.177   us/op
RequestMarshalBenchmarks.marshalCustom:·gc.alloc.rate                           512  avgt   10    1093.740 ±   10.096  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.alloc.rate.norm                      512  avgt   10  810329.575 ±    6.421    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Eden_Space                  512  avgt   10    1095.705 ±    2.374  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Eden_Space.norm             512  avgt   10  811814.175 ± 8164.026    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Survivor_Space              512  avgt   10       0.437 ±    0.143  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Survivor_Space.norm         512  avgt   10     324.000 ±  105.829    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.count                                512  avgt   10     110.000             counts
RequestMarshalBenchmarks.marshalCustom:·gc.time                                 512  avgt   10      78.000                 ms
```

Before
```
Benchmark                                                                (numSpans)  Mode  Cnt       Score       Error   Units
RequestMarshalBenchmarks.marshalCustom                                           16  avgt   10      12.804 ±     0.117   us/op
RequestMarshalBenchmarks.marshalCustom:·gc.alloc.rate                            16  avgt   10    1293.442 ±    12.145  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.alloc.rate.norm                       16  avgt   10   26083.452 ±     0.317    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Eden_Space                   16  avgt   10    1295.033 ±    95.042  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Eden_Space.norm              16  avgt   10   26115.485 ±  1900.802    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Survivor_Space               16  avgt   10       0.030 ±     0.008  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Survivor_Space.norm          16  avgt   10       0.605 ±     0.154    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.count                                 16  avgt   10     106.000              counts
RequestMarshalBenchmarks.marshalCustom:·gc.time                                  16  avgt   10      70.000                  ms
RequestMarshalBenchmarks.marshalCustom                                          512  avgt   10     468.186 ±     5.668   us/op
RequestMarshalBenchmarks.marshalCustom:·gc.alloc.rate                           512  avgt   10    1082.336 ±    14.241  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.alloc.rate.norm                      512  avgt   10  798033.831 ±     8.300    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Eden_Space                  512  avgt   10    1085.378 ±    47.978  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Eden_Space.norm             512  avgt   10  800330.356 ± 36889.698    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Survivor_Space              512  avgt   10       0.482 ±     0.070  MB/sec
RequestMarshalBenchmarks.marshalCustom:·gc.churn.G1_Survivor_Space.norm         512  avgt   10     355.099 ±    52.251    B/op
RequestMarshalBenchmarks.marshalCustom:·gc.count                                512  avgt   10     109.000              counts
RequestMarshalBenchmarks.marshalCustom:·gc.time                                 512  avgt   10      74.000                  ms
```


exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/KeyValueMarshaler.java
```diff
@@ -0,0 +1,253 @@
...
+        });
+    return attributeMarshalers;
+  }
+
+  private final byte[] keyUtf8;
+  private final Marshaler value;
+
+  @SuppressWarnings("unchecked")
+  static KeyValueMarshaler create(AttributeKey<?> attributeKey, Object value) {
+    byte[] keyUtf8 = MarshalerUtil.toBytes(attributeKey.getKey());
```
anuraaga, line 44: Removed value null check since values can't be null

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/KeyValueMarshaler.java
```diff
@@ -0,0 +1,253 @@
...
+    switch (attributeKey.getType()) {
+      case STRING:
+        return new KeyValueMarshaler(
+            keyUtf8, new AnyStringFieldMarshaler(MarshalerUtil.toBytes((String) value)));
+      case LONG:
+        return new KeyValueMarshaler(keyUtf8, new AnyInt64FieldMarshaler((long) value));
```
anuraaga, line 50: Note the extra `new` compared to previous, one extra wrapper because of the oneof. In the future this could be removed using another switch statement instead elsewhere if it's worth it

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/KeyValueMarshaler.java
```diff
@@ -0,0 +1,253 @@
...
+            keyUtf8,
+            new AnyArrayFieldMarshaler(ArrayValueMarshaler.createString((List<String>) value)));
+      case LONG_ARRAY:
+        return new KeyValueMarshaler(
+            keyUtf8,
+            new AnyArrayFieldMarshaler(ArrayValueMarshaler.createInt64((List<Long>) value)));
+      case BOOLEAN_ARRAY:
+        return new KeyValueMarshaler(
+            keyUtf8,
+            new AnyArrayFieldMarshaler(ArrayValueMarshaler.createBool((List<Boolean>) value)));
```
anuraaga, line 66: Not only a `new` but the arrayvaluemarshaler will create an array of marshalers. The originally code was crafted mostly to avoid this I think. But IMO it should be fine for now and we can come back later. Without type-specific array marshalers we end up with quite a bit less code at least.

jkwatson (approved):
Thanks!

PR #3581, opened 2021-09-06
Make Marshaler.writeTo private and expose method to write binary into…

anuraaga:
… an OutputStream.

Callers of the Marshaler don't actually need to worry about `CodedOutputstream`, `Serializer`, (or later `JsonGenerator`) if we switch to only exposing an interface that accepts an `OutputStream`. While it's internal code, we do get to see the simplification by the reduction of code in various places that are calling the marshaler and we'll be able to continue to iterate on serialization patterns without larger refactorings.

### DIFF
--- a/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/Marshaler.java
+++ b/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/Marshaler.java
@@ -6,6 +6,7 @@
 package io.opentelemetry.exporter.otlp.internal;
 
 import java.io.IOException;
+import java.io.OutputStream;
 
 /**
  * Marshaler from an SDK structure to protobuf wire format.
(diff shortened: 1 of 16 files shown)


exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/Marshaler.java
```diff
@@ -6,15 +6,25 @@
...
+  public final void writeBinaryTo(OutputStream output) throws IOException {
+    CodedOutputStream cos = CodedOutputStream.newInstance(output);
+    writeTo(new ProtoSerializer(cos));
+    cos.flush();
+  }
+
+  /** Returns the number of bytes this Marshaler will write in proto binary format. */
+  public abstract int getBinarySerializedSize();
+
+  abstract void writeTo(Serializer output) throws IOException;
```
anuraaga, line 29: We were just talking about interface vs abstract class. Switched to abstract class here to be able to make this package private.

jkwatson (approved):

PR #3582, opened 2021-09-06
Base implementation of BaggageSpanProcessor

jsuereth:
- Allows configuring spans that should have attributes modified.
- Allows configuration of which baggage keys to propagate to spans.
- Lines up with Metric View API for how to append baggage.

Posted for initial comments and thoughts before refining further.


sdk-extensions/baggage-processor/src/main/java/io/opentelemetry/sdk/extension/trace/baggage/BaggageSpanProcessor.java
```diff
@@ -0,0 +1,71 @@
...
+  public boolean isEndRequired() {
+    return next.isEndRequired();
+  }
+
+  /**
+   * Returns a builder for a baggage span processor.
+   *
+   * @param next The next processor in the chain.
+   */
+  public static final BaggageSpanProcessorBuilder builder(SpanProcessor next) {
```
anuraaga, line 68:
```suggestion
  public static BaggageSpanProcessorBuilder builder(SpanProcessor next) {
```

sdk-extensions/baggage-processor/src/main/java/io/opentelemetry/sdk/extension/trace/baggage/BaggageSpanProcessorBuilder.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+
+/**
+ * Builder for {@link BaggageSpanProcessor}.
+ *
+ * <p>Allows configurable rules for what baggage to include and when to include it.
+ */
+public class BaggageSpanProcessorBuilder {
+  private final SpanProcessor next;
```
anuraaga, line 20: This might be an issue with the span processor API in general but while the pattern of wrapping could probably have been a better one for ensuring ordering, I don't think it's idiomatic for our SDK's current design. We'd expect the user to make sure this processor is added first, and not have wrapping

sdk-extensions/baggage-processor/src/test/java/io/opentelemetry/sdk/extension/trace/baggage/TestBaggageSpanProcessor.java
```diff
@@ -0,0 +1,89 @@
...
+import io.opentelemetry.sdk.trace.ReadWriteSpan;
+import io.opentelemetry.sdk.trace.SpanProcessor;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+public class TestBaggageSpanProcessor {
```
anuraaga, line 20:
```suggestion
class TestBaggageSpanProcessor {
```

sdk-extensions/baggage-processor/src/test/java/io/opentelemetry/sdk/extension/trace/baggage/TestBaggageSpanProcessor.java
```diff
@@ -0,0 +1,89 @@
...
+
+  @Test
+  void ignoresEmptyBaggage() {
+    BaggageSpanProcessor processor = BaggageSpanProcessor.builder(nextProcessor).build();
+
+    Baggage empty = Baggage.empty();
+    Context context = Context.root().with(empty);
+
+    processor.onStart(context, readWriteSpan);
+    Mockito.verify(nextProcessor).onStart(context, readWriteSpan);
```
anuraaga, line 32: We generally static import mockito methods

sdk-extensions/baggage-processor/src/test/java/io/opentelemetry/sdk/extension/trace/baggage/TestBaggageSpanProcessor.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  @Test
+  void appendsAllBaggage() {
+    BaggageSpanProcessor processor = BaggageSpanProcessor.builder(nextProcessor).build();
+    Baggage empty = Baggage.builder().put("test1", "value").put("test2", "value").build();
+    Context context = Context.root().with(empty);
+    processor.onStart(context, readWriteSpan);
+    Mockito.verify(nextProcessor).onStart(context, readWriteSpan);
+    Mockito.verify(readWriteSpan).setAttribute("test1", "value");
```
anuraaga, line 44: Consider replacing `verify` with `when` and returning the `readWriteSpan`. Mockito now automatically verifies when using `when` so don't need two statements. And it ensures the business logic can be changed to chain invocations if needed in the future (I guess it might not happen, but it prevents the possibility of a really mysterious NPE in the future if it does).

I believe the `never` calls aren't really needed either (maybe they're good for documentation though) since mockito would fail if there are unexpected invocations too

sdk-extensions/baggage-processor/src/main/java/io/opentelemetry/sdk/extension/trace/baggage/BaggageSpanProcessorBuilder.java
jsuereth: Interesting. I followed the pattern of the AsyncProcessor here.  Do we have any instances of pipelining processors outside of these two?

anuraaga: AsyncProcessor, BatchSpanProcessor, SimpleSpanProcessor are just exporter wrappers so they're a bit different from something that works with the actual span. A notable issue with this approach for baggage is that while intuitively, you'd expect modifications from the processor to only be available in `next` - in actuality, the span is mutated so other span processors in the pipeline, not just the `next`, will see the mutated span.

I think the only other non-exporter span processor we have is zpages

https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/zpages/src/main/java/io/opentelemetry/sdk/extension/zpages/TracezSpanProcessor.java

jsuereth: AsyncProcessor does not wrap Exporter, but processor, see: https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/async-processor/src/main/java/io/opentelemetry/sdk/extension/trace/export/DisruptorAsyncSpanProcessor.java#L72

It just looks funky because it hides the processor behind disruptor event queue fluffier.

sdk-extensions/baggage-processor/src/test/java/io/opentelemetry/sdk/extension/trace/baggage/TestBaggageSpanProcessor.java
jsuereth: This depends on whether someone is using strict-mock or lenient-mock mode.  I've learned to be a bit more explicit in negative assumptions (never) in case someone flips to lenient mode.

IF you think the `when` calls read better, I can flip this, I just felt the post-call verification felt more natural here.

sdk-extensions/baggage-processor/src/main/java/io/opentelemetry/sdk/extension/trace/baggage/BaggageSpanProcessorBuilder.java
anuraaga: Oops didn't notice async doesn't explicitly take exporter, I think the only use case anyone expects with it is exporter so I guess it always requires a SimpleSpanProcessor in between...

Either way my comment still stands though, especially that a span mutating processor has no reason to wrap another

sdk-extensions/baggage-processor/src/test/java/io/opentelemetry/sdk/extension/trace/baggage/TestBaggageSpanProcessor.java
anuraaga: It's two types of fragility, the other being methods that never return null returning null in the mocked version. I think we can expect to always be in the default strict mock mode.

sdk-extensions/baggage-processor/src/main/java/io/opentelemetry/sdk/extension/trace/baggage/BaggageSpanProcessorBuilder.java
jsuereth: In practice, we always recommend using a batch span processor.  For this specific processor I don't think preventing the usage of BSP is a great idea.

If a span-mutating process doesnt allow further processing, doesn't that mean it needs to have all functionality user could want?  Shouldn't all non-final processors wrap another?

anuraaga: The SDK has `addSpanProcessor` and allows multiple processors that execute in order. So it doesn't prevent BatchSpanProcessor (phew since it's basically required to export). That's basically what makes zpages possible.

It means a batch span processor added to the SDK after baggage processor will be affected by the baggage processor regardless of if it's next or not. So I think the general idiom is for non-exporting to not wrap anything, and async's API is probably a bug (it's still alpha, it's probably around time to delete that module since I haven't heard any demand for it yet).

anuraaga: The expressiveness of the export pipeline not being great came up around release but we decided it's too much to change the spec at that point. I was interested in getting rid of BSP type of exporting processors entirely. But for now we just have a list of processors in order - it doesn't make sense for a mutating processor to wrap since there is no isolation and non-wrapped will be effected too, defeating the expected meaning of wrapping.

sdk-extensions/baggage-processor/README.md
```diff
@@ -0,0 +1,9 @@
+# OpenTelemetry SDK Extension Baggage SpanProcessor
+
+[![Javadocs][javadoc-image]][javadoc-url]
+
+An implementation of the trace `SpanProcessors` that uses
```
jkwatson, line 5:
```suggestion
An implementation of a trace `SpanProcessor` that uses
```

sdk-extensions/baggage-processor/src/main/java/io/opentelemetry/sdk/extension/trace/baggage/BaggageSpanProcessor.java
```diff
@@ -0,0 +1,71 @@
...
+    if (!baggage.isEmpty() && spanFilter.test(span)) {
+      // TODO: append-only rather than overwrite with baggage.
+      baggage.forEach(
+          (k, v) -> {
+            if (baggageKeyFilter.test(k)) {
+              span.setAttribute(k, v.getValue());
+            }
+          });
+    }
+    next.onStart(parentContext, span);
```
jkwatson, line 45: shouldn't we check to make sure that `next.isStartRequired()` here?

sdk-extensions/baggage-processor/src/test/java/io/opentelemetry/sdk/extension/trace/baggage/TestBaggageSpanProcessor.java
```diff
@@ -0,0 +1,92 @@
...
+import io.opentelemetry.context.Context;
+import io.opentelemetry.sdk.trace.ReadWriteSpan;
+import io.opentelemetry.sdk.trace.SpanProcessor;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class TestBaggageSpanProcessor {
```
jkwatson, line 24: We name these classes with the `Test` at the end, rather than at the start.

PR #3583, opened 2021-09-06
Wire Exemplar Reservoir into Metrics SDK

jsuereth:
This PR does NOT expose the full configuration + API for exemplars, it simple does the smallest possible change to include ExemplarReservoir interface and wire it into aggregators/test the usage of the reservoir.

- Wire ExemplarReservoir into Aggregator interrface
- Expose bare minimum SDK configuration (for now), prior to `ExemplarFilter` PR.
- Wire exemplar reservoir into Histogram aggregation, including testing the usage of it.
- Create LongAccumulation class to store long values + exemplars in relevant aggregations.
- Create DoubleAccumulation class to store double values + exemplars in relevant aggregations.


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarReservoir.java
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * An interface for an exemplar reservoir of samples.
+ *
+ * <p>This represents a reservoir for a specific "point" of metric data.
+ */
+public interface ExemplarReservoir {
+
+  /** Offers a measurement to be sampled. */
+  public void offerMeasurementLong(long value, Attributes attributes, Context context);
```
anuraaga, line 22: Remove `public` from methods since interface

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarReservoir.java
```diff
@@ -0,0 +1,59 @@
...
+   * @return An (immutable) list of sampled exemplars for this point.
+   */
+  public List<Exemplar> collectAndReset(Attributes pointAttributes);
+
+  /** An exemplar reservoir that stores no exemplars. */
+  public static ExemplarReservoir empty() {
+    return EMPTY;
+  }
+
+  static final ExemplarReservoir EMPTY =
```
anuraaga, line 42: Move to some package private class to make sure not public

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarReservoir.java
```diff
@@ -0,0 +1,59 @@
...
+   * <p>Additionally, clears the reservoir for the next sampling period.
+   *
+   * @param pointAttributes the {@link Attributes} associated with the metric point. {@link
+   *     Exemplar}s should filter these out of their final data state.
+   * @return An (immutable) list of sampled exemplars for this point.
+   */
+  public List<Exemplar> collectAndReset(Attributes pointAttributes);
+
+  /** An exemplar reservoir that stores no exemplars. */
+  public static ExemplarReservoir empty() {
```
anuraaga, line 38: Nit, generally put the static methods in an interface at the top since they're the entry point

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarReservoir.java
anuraaga: It's probably ok to call them both `offerMeasurement` for better readability

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarSampler.java
```diff
@@ -0,0 +1,40 @@
...
+ * An interface that provides the two configuration components to exemplar sampling.
+ *
+ * <ul>
+ *   <li>(coming soon) A filter for which measurements should be sampled.
+ *   <li>A reservoir to store exemplars (and can do additional filtering)
+ * </ul>
+ */
+@AutoValue
+@Immutable
+public abstract class ExemplarSampler {
```
anuraaga, line 21: Add package private constructor

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarStorageStrategy.java
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.metrics.exemplar;
+
+import io.opentelemetry.sdk.metrics.view.Aggregation;
+
+/** A strategy for storing exemplars. */
+@FunctionalInterface
```
anuraaga, line 11: Confirm it's worth adding this with the tradeoff of not being able to add any default methods in the future. This doesn't seem like such a "functional" class

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarStorageStrategy.java
```diff
@@ -0,0 +1,24 @@
...
+public interface ExemplarStorageStrategy {
+  /** Constructs a new sampling reservoir for a given aggregator. */
+  ExemplarReservoir createReservoir(Aggregation aggregator);
+
+  /** A sampler which will never store Exemplars. */
+  static final ExemplarStorageStrategy ALWAYS_OFF = (agg) -> ExemplarReservoir.empty();
```
anuraaga, line 17: Probably in the same class as the above comment, let's move these into a package private class and expose them through static methods for consistency with other interfaces

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarReservoir.java
```diff
@@ -0,0 +1,59 @@
...
+import io.opentelemetry.sdk.metrics.data.Exemplar;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * An interface for an exemplar reservoir of samples.
+ *
+ * <p>This represents a reservoir for a specific "point" of metric data.
+ */
+public interface ExemplarReservoir {
```
anuraaga, line 19: Reservoir usually implies a type of sampling, should this be the class that's called `ExemplarSampler`? It looks very similar to the tracing Sampler

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarReservoir.java
anuraaga: With the above suggestion, this method would become `neverSample`

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarSampler.java
anuraaga: With the above suggestion, I guess this is `ExemplarSamplerConfig` or `ExemplarSamplerProvider`. Wonder if we actually need this class though instead of having all knobs on `ExemplarStrategy`

anuraaga (commented):
Just some small points

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarReservoir.java
jsuereth: the reservoir is instantiated *per* metric-timeseries.  I.e. one ExemplarSampler will be instantiating many reservoirs.

I tried to keep the two as similar as possible, only deviating where it makes sense, but I could have gotten the split wrong.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarSampler.java
jsuereth: This eventually get a measurement pre-filter, see: https://github.com/jsuereth/opentelemetry-java/blob/wip-metrics-api/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarSampler.java

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarReservoir.java
jsuereth: Went with "noSamples" instead to denote this doesn't allocate memory for samples vs. not taking samples.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarReservoir.java
jsuereth: Moved.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarReservoir.java
jsuereth: Done.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarStorageStrategy.java
jsuereth: Not sure.  There's a future PR to redo the `Aggregation` API so this method can be fully implemented.   From my perspective this should MOSTLY be a pattern match against aggregation type, optionally pulling out some aggregation config.

jkwatson: As far as I know, you can add `default` methods to `@FunctionalInterface` SMIs without breaking anything.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProviderBuilder.java
```diff
@@ -50,6 +53,12 @@ public SdkMeterProviderBuilder setResource(Resource resource) {
     return this;
   }
 
+  /** TODO: document. */
```
jkwatson, line 56: yessir. let's not forget.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/NoExemplarReservoir.java
```diff
@@ -0,0 +1,35 @@
...
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.sdk.metrics.data.Exemplar;
+import java.util.Collections;
+import java.util.List;
+
+/** Implementation of a reservoir that keeps no exemplars. */
+class NoExemplarReservoir implements ExemplarReservoir {
+
+  static final ExemplarReservoir NO_SAMPLES = new NoExemplarReservoir();
```
jkwatson, line 17: how about just calling this "INSTANCE". This name clash threw me when I saw it being used. we could also hide it behind a static method as we've done in other places in the codebase.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarStorageStrategy.java
jkwatson: But, I agree that this probably isn't adding a lot of value at the moment, since the JVM treats all SMIs as `@FunctionalInterface`s by default.

jkwatson: if we wanted this to use built-in types, we could make it a `Function<Aggregation, ExemplarReservoir>`. That might end up being more confusing, though.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/AggregatorHandle.java
```diff
@@ -105,6 +112,7 @@ public final void recordLong(long value) {
 
   @Override
   public final void recordLong(long value, Attributes attributes, Context context) {
```
jkwatson, line 114: I think this method should move up one in the class so it calls "down" to the other `recordLong` which then calls `doRecordLong`.  Same for the double method ordering.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/CountAggregator.java
```diff
@@ -8,54 +8,62 @@
...
         descriptor,
         temporality == AggregationTemporality.CUMULATIVE);
     this.temporality = temporality;
+    this.reservoirBuilder = reservoirBuilder;
   }
 
   @Override
-  public AggregatorHandle<Long> createHandle() {
-    return new Handle();
+  public AggregatorHandle<LongAccumulation> createHandle() {
```
jkwatson, line 45: Are we keeping the door open to not require the `LongAccumulation` wrapper when exemplars are turned off for a given instrument?

jkwatson: (this question applies to all the aggregators of course..I just thought about it when looking at this one).

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarSampler.java
```diff
@@ -0,0 +1,40 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.metrics.exemplar;
+
+import com.google.auto.value.AutoValue;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * An interface that provides the two configuration components to exemplar sampling.
```
jkwatson, line 12: s/interface/something else here/

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarSampler.java
```diff
@@ -0,0 +1,40 @@
...
+@AutoValue
+@Immutable
+public abstract class ExemplarSampler {
+
+  /** Configuration for exemplar storage. */
+  public abstract ExemplarStorageStrategy getStorage();
+
+  /**
+   * We hide the ability to create custom exemplar samplers until further specification work is
+   * stable.
```
jkwatson, line 28: not sure I grok this comment and how it applies to this method.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/LongLastValueAggregator.java
```diff
@@ -26,33 +30,37 @@
...
-final class LongLastValueAggregator extends AbstractAggregator<Long> {
+final class LongLastValueAggregator extends AbstractAggregator<LongAccumulation> {
+  private final Supplier<ExemplarReservoir> reservoirBuilder;
+
   LongLastValueAggregator(
       Resource resource,
       InstrumentationLibraryInfo instrumentationLibraryInfo,
-      MetricDescriptor descriptor) {
+      MetricDescriptor descriptor,
+      Supplier<ExemplarReservoir> reservoirBuilder) {
```
jkwatson, line 40: I don't think I would call these parameters "...builder". Probably just "reservoirSupplier" is a bit better.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/CountAggregator.java
jsuereth: Good question.   I wasn't planning to. I need to re-run the benchmark, as we're effectively boxing over the Long here anyway so I didn't see a huge slowdown.   This accumulations is NOT on the critical path, only when collecting.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/AggregatorHandle.java
jsuereth: Right, eventually this gets refactored/shrunk.  I've been trying to limit the size of these CLs so far, so all that extra gunk is still there (but not needed).

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarStorageStrategy.java
jsuereth: I expect the interface to (eventually) have more arguments (or a diffferent argument than just the raw "Aggregation", that's in a follow-up CL where we wire in the HistogramExemplarReservoir + DefaultExemplarReservoir, so going to avoid `Function` as we might need more than one parameter.

Took a crack at fixing LMKWYT.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarSampler.java
jsuereth: fixed.

anuraaga (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarReservoir.java
```diff
@@ -0,0 +1,41 @@
...
+  /** Offers a {@code double} measurement to be sampled. */
+  void offerMeasurement(double value, Attributes attributes, Context context);
+
+  /**
+   * Builds (an immutable) list of Exemplars for exporting from the current reservoir.
+   *
+   * <p>Additionally, clears the reservoir for the next sampling period.
+   *
+   * @param pointAttributes the {@link Attributes} associated with the metric point. {@link
+   *     Exemplar}s should filter these out of their final data state.
```
jkwatson, line 37: Whose responsibility is the filtering? Is this comment intended to be simply informative, or is this directing the Reservoir implementor to do the filtering? We should definitely make it clear if this is something that implementor needs to worry about or not.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarReservoirFactory.java
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+package io.opentelemetry.sdk.metrics.exemplar;
+
+import io.opentelemetry.sdk.metrics.view.Aggregation;
+
+/** A factory for {@link ExemplarReservoir}s that can leverage Aggregation configuration. */
+public interface ExemplarReservoirFactory {
+  /**
+   * Constructs a new sampling reservoir for a given metric stream.
```
jkwatson, line 13: I think this comment needs to be fleshed out a bit. Something like :
```
Constructs a new sampling reservoir. This will be applied to a single metric stream and will be responsible
for capturing Exemplar data associated with the metric recordings.
```

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarSampler.java
jkwatson: definitely needs a package private constructor, name change or not.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/CountAggregator.java
```diff
@@ -8,54 +8,62 @@
...
-    return 1L;
+  public LongAccumulation accumulateLong(long value) {
+    return LongAccumulation.create(1L);
   }
 
   @Override
-  public Long merge(Long a1, Long a2) {
-    return a1 + a2;
+  public LongAccumulation merge(LongAccumulation a1, LongAccumulation a2) {
+    return LongAccumulation.create(a1.getValue() + a2.getValue(), a2.getExemplars());
```
jkwatson, line 61: why not the set of both a1 and a2's exemplars? I'm not sure I would have known this was the right way to implement `merge` with exemplars in play.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleHistogramAggregator.java
```diff
@@ -58,7 +65,8 @@ public final HistogramAccumulation merge(HistogramAccumulation x, HistogramAccum
     for (int i = 0; i < x.getCounts().length; ++i) {
       mergedCounts[i] = x.getCounts()[i] + y.getCounts()[i];
     }
-    return HistogramAccumulation.create(x.getSum() + y.getSum(), mergedCounts);
+    // Note: we always preserve the left-side exemplars as these are the "newer" ones.
```
jkwatson, line 68: This seems to be the opposite of the CountAggregator choice.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleLastValueAggregator.java
```diff
@@ -28,33 +32,37 @@
...
-  public Double accumulateDouble(double value) {
-    return value;
+  public DoubleAccumulation accumulateDouble(double value) {
+    return DoubleAccumulation.create(value);
   }
 
   @Override
-  public Double merge(Double a1, Double a2) {
+  public DoubleAccumulation merge(DoubleAccumulation a1, DoubleAccumulation a2) {
     // TODO: Define the order between accumulation.
```
jkwatson, line 59: lol.

jkwatson: sorry...just a little punchy today, and after seeing 2 different merge choice, I saw this. :)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/CountAggregator.java
jsuereth: Two main points:

1. I'm generally not happy with this particular implementation here.  Without doing a refactor of the Aggregator interface, there's a lot of subtle details in methods that are either not documented (or buggy, e.g. how merge is used between Sync + Async aggregation w/ Sum).   a1 = previous, a2 = current when exemplars exist.   For Async instruments (where you'd want to merge), they'd both be empty.
2. According to the spec, it's best-effort to preserve exemplars (on purpose) so we can optimise code paths.  Additionally, for cumulative instruments, it's expected you show recent exemplars.  There is no expectation of perserverance.

If you think it's worth it (in this CL) to do some light refactoring + documentation of the merge method, I'd be more than happy to kick that off.  I'd been avoiding anything major until all the pieces are in place.   However, it's really getting unwieldy.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleHistogramAggregator.java
jsuereth: Yeah, I wasn't paying close enough attention to my backport.  I had flipped the order when reworking the Aggregator interface, but haven't driven that change here yet.  Fixed.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarReservoirFactory.java
jsuereth: Added your description exactly for now.  If you have any better descriptive ideas, let me know.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarReservoir.java
jsuereth: The implementation needs to do it, added a more descriptive comment here.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarSampler.java
jsuereth: Made this package private.  PTAL at #3599 if you want to see what `ExemplarSampler` will look like once completed.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/ExemplarReservoir.java
jkwatson: This looks the same as it was before to me. Missed push, perhaps?

jkwatson: oh, I see. It's below in the `@return`. How important is this? It *sounds* like it's very important, but I think it would be easy to miss this requirement as currently written.

jkwatson (approved):
Let's :shipit:

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/CountAggregator.java
jkwatson: This is fine for now. As long as we're doing it consistently, that's a good start.

PR #3586, opened 2021-09-08
Extract trace / span ID serialization logic to Serializer to allow ch…

anuraaga:
…anging for JSON.

Since trace / span ID don't follow protobuf conventions in OTLP, we can't just have a single `serializeBytes` method handling them so added them to the `Serializer`. By moving the byte cache into that layer, it seems to have actually simplified the code quite a bit too, and given metrics the caching automatically.

No change in performance. I initially tried using invalid IDs instead of `null` as currently in this PR but that tanked performance.


jkwatson (approved):
👍🏽

PR #3588, opened 2021-09-09
Implement LogAdapter

ran8080:
This PR Implements the LogAdapter class as a part of the following issue:
[Logging POC for log4j, logback](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3055)

This PR is not yet implementing any source of LogExporter, to keep things short and neat.
 
- Implemented the LogAdapter class to Convert LogRecord (data class) to LogRecord Protobuff implementation
- Written needed unit tests
- Added Resource and InstrumentationLibraryInfo members to LogRecord to support the adaptation.


atoulme (approved):
LGTM

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/LogRecord.java
```diff
@@ -32,9 +36,23 @@ static LogRecord create(
       AnyValue body,
```
jkwatson, line 36: I would love a follow-on to convert this to a `String`.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/LogAdapter.java
```diff
@@ -0,0 +1,111 @@
...
+import io.opentelemetry.sdk.logging.data.LogRecord;
+import io.opentelemetry.sdk.resources.Resource;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/** Converter from SDK {@link LogRecord} to OTLP {@link ResourceLogs}. */
+public class LogAdapter {
```
jkwatson, line 24: does this need to be `public`?

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/LogRecord.java
ran8080: Sure, let me know if you want it in this PR, or is it fine to modify this on our next PR?

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/LogAdapter.java
jack-berg: Eventually it will since the various exporters consume these adapters from different packages.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/LogAdapter.java
```diff
@@ -0,0 +1,111 @@
...
+              result.computeIfAbsent(logRecord.getResource(), unused -> new HashMap<>());
+      List<io.opentelemetry.proto.logs.v1.LogRecord> logList =
+          libraryInfoListMap.computeIfAbsent(
+              logRecord.getInstrumentationLibraryInfo(), unused -> new ArrayList<>());
+      logList.add(toProtoLogRecord(logRecord));
+    }
+    return result;
+  }
+
+  /** returns a new {@link io.opentelemetry.proto.logs.v1.LogRecord} from a {@link LogRecord}. */
```
jack-berg, line 78: #nit: 
```suggestion
  /** Returns a new {@link io.opentelemetry.proto.logs.v1.LogRecord} from a {@link LogRecord}. */
```

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/LogAdapter.java
```diff
@@ -0,0 +1,111 @@
...
+    return logsBuilder.build();
+  }
+
+  private static Map<
+          Resource, Map<InstrumentationLibraryInfo, List<io.opentelemetry.proto.logs.v1.LogRecord>>>
+      groupByResourceAndLibrary(Collection<LogRecord> logRecordList) {
+    Map<Resource, Map<InstrumentationLibraryInfo, List<io.opentelemetry.proto.logs.v1.LogRecord>>>
+        result = new HashMap<>();
+    for (LogRecord logRecord : logRecordList) {
+
```
jack-berg, line 66: #nit: Unintentional blank line?
```suggestion
```

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/LogRecord.java
jkwatson: a follow-on PR is fine.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/LogAdapter.java
jkwatson: Please add our standard javadoc boilerplate for public internal classes:
```
 * at any time.
```

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/LogRecordBuilder.java
```diff
@@ -7,21 +7,36 @@
...
+  private InstrumentationLibraryInfo instrumentationLibraryInfo;
   private long timeUnixNano;
   private String traceId = "";
   private String spanId = "";
   private int flags;
   private LogRecord.Severity severity = LogRecord.Severity.UNDEFINED_SEVERITY_NUMBER;
   private String severityText;
   private String name;
-  private AnyValue body = AnyValue.stringAnyValue("");
+  private String body = "";
```
jkwatson, line 24: interesting default. Does the proto require the body to have a value, or can it be proto-null?

jkwatson (approved):
Seems like a reasonable start.

ran8080:
Thanks for the help John

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/LogAdapter.java
anuraaga:
```suggestion
public final class LogAdapter {
```

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/LogAdapter.java
```diff
@@ -0,0 +1,115 @@
...
+            .setName(logRecord.getName())
+            .setBody(getLogRecordBodyAnyValue(logRecord))
+            .setSeverityText(logRecord.getSeverityText())
+            .setSeverityNumber(
+                SeverityNumber.forNumber(logRecord.getSeverity().getSeverityNumber()))
+            .setTimeUnixNano(logRecord.getTimeUnixNano())
+            .setTraceId(
+                UnsafeByteOperations.unsafeWrap(
+                    OtelEncodingUtils.bytesFromBase16(
+                        logRecord.getTraceId(), logRecord.getTraceId().length())))
```
anuraaga, line 95:
```suggestion
                        logRecord.getTraceId(), TraceId.getLength())))
```

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/LogAdapter.java
```diff
@@ -0,0 +1,115 @@
...
+            .setSpanId(
+                UnsafeByteOperations.unsafeWrap(
+                    OtelEncodingUtils.bytesFromBase16(
+                        logRecord.getSpanId(), logRecord.getSpanId().length())));
```
anuraaga, line 99:
```suggestion
                        logRecord.getSpanId(), SpanId.getLength())));
```

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/LogAdapter.java
```diff
@@ -0,0 +1,115 @@
...
+
+    logRecord
+        .getAttributes()
+        .forEach((key, value) -> builder.addAttributes(CommonAdapter.toProtoAttribute(key, value)));
+
+    io.opentelemetry.proto.logs.v1.LogRecord protoLogRecord = builder.build();
+    builder.clear();
```
anuraaga, line 106:
```suggestion
```

We're not doing any caching here so don't need to clear builders

sdk/testing/build.gradle.kts
```diff
@@ -9,6 +11,7 @@ otelJava.moduleName.set("io.opentelemetry.sdk.testing")
 dependencies {
   api(project(":api:all"))
   api(project(":sdk:all"))
+  api(project(":sdk-extensions:logging"))
```
anuraaga, line 14: This package doesn't seem to be related to the PR, can we revert it?

anuraaga (approved):

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/LogRecord.java
```diff
@@ -22,19 +24,35 @@ public static LogRecordBuilder builder() {
...
+      InstrumentationLibraryInfo instrumentationLibraryInfo,
       long timeUnixNano,
       String traceId,
       String spanId,
       int flags,
       Severity severity,
       String severityText,
       String name,
-      AnyValue body,
+      String body,
```
jsuereth, line 36: Why this deviation?  https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-body

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/LogRecordBuilder.java
jsuereth: Body is optional: https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/logs/v1/logs.proto#L116


However, I'm very curious why we dropped `AnyValue` for `String` here.   I'd prefer we keep body as a structured value (with convenience methods to use a string-any value) vs. removing the flexibility.

jsuereth (changes requested):

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/LogRecord.java
jsuereth: Why are we deviating from the log model here?

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/LogRecordBuilder.java
jkwatson: I want to only add the complexity/flexibility if we have an actual use-case. Let's start with String only, please.

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/LogRecord.java
atoulme: My use case is not tied to traditional java logging. I would foresee that we would log data, most likely serialized to json. A string therefore is fine, maybe a byte array would do too.

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/LogRecordBuilder.java
jsuereth: At a bare minimum, I'd ask that you attend the logging SiG with this feedback on the beta model and suggestions around SDK integrations.

However, regarding use cases, what's the plan for logging exceptions in Java in OTLP?  Naively, I thought body would be an excellent opportunity to define something better than a big long string.

jsuereth (commented):
Removing request for changes, however I think deviating from the log model should be raised in the logging sig and called out in the code/tracked in some way.   I don't think this deviation is due to a limitation between the specification and Java.

jsuereth (approved):

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/LogRecord.java
jsuereth: Most java logging library have structured logging, it's just not used much.   Additionally, ALL of them have exceptions that should be dealt with.

- [Slf4j LoggingEvent](http://www.slf4j.org/api/org/slf4j/event/LoggingEvent.html) supports key-value pairs.  It's not clear to me these belong as attributes, and if you look at recommended usage, I believe these are likely JSON-like body attributes in practice.
- [Logstash](https://www.innoq.com/en/blog/structured-logging/) added hacky ways to encode key-value pairs that show up in the message and not in the message.
- [Java Logging has parameters](https://docs.oracle.com/javase/7/docs/api/java/util/logging/LogRecord.html#getParameters()), which it's unclear/unspecified if these belong in `attributes` or `body` of log
- [log4j has StringMapMessage](https://logging.apache.org/log4j/log4j-2.9.1/log4j-api/apidocs/index.html?org/apache/logging/log4j/message/StringMapMessage.html) which is used to encode key-value pairs

Those are just the ones I know of when I was reworking the build logging framework for sbt.

I agree with the notion of keeping simplicity and as long as we're ok adding AnyValue back in when we integrate these log messages, I'm on board.

jkwatson:
> Removing request for changes, however I think deviating from the log model should be raised in the logging sig and called out in the code/tracked in some way.   I don't think this deviation is due to a limitation between the specification and Java.

If there is no need for this complexity, we should be free to provide an API for appenders that is relevant to the logging frameworks we will need to support. I don't think the logging SIG or the otlp model should be required to leak into our API surface without a compelling reason.

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/LogRecordBuilder.java
jkwatson: I believe exceptions go into the attributes not the body in Java logging.

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/LogRecord.java
jkwatson: I *think* that all the java logging frameworks end up reducing down to String by the time you're implementing a log appender. Yes, the APIs let you send in all sorts of crazy stuff, but you have to provide formatters for that stuff, and the formatters (again, as far as I'm aware) all format to a String by the time we'd be picking up the data to convert to OTLP.

ran8080:
@jkwatson Hey, Closed branch means we will merge it to the next version? or something else?
Not sure I'm quite get this part yet.

ran8080:
@jkwatson Approved branches will be usually merged to the next version I guess?
When will the code be added to main branch?

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/LogRecord.java
jkwatson: One thing that is missing from this whole discussion is specification from the logging sig about how all of these various and sundry logging frameworks' data is supposed to be mapped into OTLP.  Yes, we have inter-format conversion recipes, but that won't help appender implementors for a specific logging framework.  In particular, it's telling that there is no mention of an Exception type in the current spec (although there is a mention of stack traces in the ECS transformation, and that just goes into an attribute).

I fear that if we put `AnyValue` as the body for our LogRecord, it will make it all the more confusing for an appender implementor to know what they're supposed to put where. At least if the body is a `String` it's pretty clear that you want the formatted log message to go there.

All that being said, if there is a compelling use-case for having an arbitrary `AnyValue` thing as the body, then we can add an API for that. To start with, though, I'd like to make the common thing (String) easy, and target making the less common thing possible when/if we find a use-case for it.

jkwatson:
I'd like to wait to merge this until after 1.6.0 is released, and to give further time for discussion of the String/AnyValue issue.

ran8080:
@jkwatson Sure, np. We will base our next pr for the OTLPLoggingExporter on top of the PR anyway.

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/LogRecord.java
anuraaga: FWIW, +1 to keeping body as string at least for now. I can't find a difference in meaning between a body with key/value pairs and also having attributes with key/value pairs. Indexing behavior seems the same, index both in general, for example, `StringMapMessage` seems to be the most explicitly a body of key/value pairs at least based on its name, but usages seem to look like attributes, or at least they do need to be indexed the same as attributes would be.

https://github.com/search?q=stringmapmessage&type=code

I have a feeling the idiomatic way of handling this would be to convert all fields of a map to attributes, except for a field named `message` (just by convention, there is no strict meaning to this in the frameworks) which would be copied into body. I've only used JSON logging with Stackdriver before, but I recall it doing something similar to this when scraping.

This is also only assuming it's even possible to do this, though @jkwatson points out appenders may not even get anything other than a string after formatting. I haven't dug into that but that would make it even more compelling to stick to string for now.

jkwatson:
@ran8080 it looks like there are a few unresolved comments from @anuraaga . Can you address them before we merge this? Thanks!

ran8080:
@jkwatson Hey, I just addressed all of @anuraaga comments and pushed the changes. Should be ready for merge soon.

sdk/testing/build.gradle.kts
jkwatson: @ran8080 this line should be deleted, yes?

docs/apidiffs/current_vs_latest/opentelemetry-exporter-otlp-trace.txt
```diff
@@ -1,2 +1,4 @@
 Comparing source compatibility of  against 
-No changes.
\ No newline at end of file
+***! MODIFIED CLASS: PUBLIC FINAL io.opentelemetry.exporter.otlp.trace.OtlpGrpcSpanExporterBuilder  (not serializable)
+	===  CLASS FILE FORMAT VERSION: 52.0 <- 52.0
+	---! REMOVED METHOD: PUBLIC(-) io.opentelemetry.exporter.otlp.trace.OtlpGrpcSpanExporterBuilder setCompression(java.lang.String)
```
jkwatson, line 4: what happened here? I think your rebase got messed up to include this.

sdk/testing/build.gradle.kts
```diff
@@ -1,3 +1,5 @@
+
+
```
jkwatson, line 2:
```suggestion
```

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/LogAdapter.java
```diff
@@ -0,0 +1,113 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.exporter.otlp.internal;
```
jkwatson, line 6: I just noticed (very sorry for seeing this so late) that this got put into the otlp-common module. I don't think this class belongs in common does it? Shouldn't this be in a new otlp-logs module, in parallel to otlp-traces and otlp-metrics?

jkwatson: @ran8080 I don't think we can merge this PR in its current state due to the introduction of additional transitive dependencies added to the otlp-common module. Will you have time to correct this issue and put this code into a new module, specific to logging?

exporters/otlp/common/build.gradle.kts
```diff
@@ -17,6 +17,7 @@ dependencies {
   api(project(":api:all"))
   api(project(":sdk:all"))
   api(project(":sdk:metrics"))
+  api(project(":sdk-extensions:logging"))
```
anuraaga, line 20: @jkwatson I think we can make this, and probably all of these SDK dependencies, `compileOnly` so that should avoid a downstream effect

anuraaga:
```suggestion
  compileOnly(project(":sdk-extensions:logging"))
```

PR #3592, opened 2021-09-10
Implementations of Exemplar Reservoirs

jsuereth:
Depends on #3583

- Adds fixed-size exemplar reservoir
- Adds histogram-bucket exemplar reservoir

Both of these are naive samplers.  The first uses a simplistic algorithm to ensure exemplars and refresh periodically until a collection interval.   The second always preserves latest measurement per-histogram bucket.

I have benchmarked these within the metric prototype, but the wiring is not complete to do so in the current codebase.   

- For worst-case histogram exemplar sampling I saw a 35ns / operation increase in operations (5ns for API-ONLY calls, 40ns for SDK calls, 75ns for SDK-WITH-EXEMPLAR calls).
- For worst-case fixed-size exemplar sampling (with bucket count = # of cpus), we saw (on Counter instruments) 20ns / operation increase (5ns API-ONLY, 20ns SDK, 40ns SDK-WITH_EXEMPLAR)
- In practice, we expect exemplar codepaths to only be enabled at the tracing sample rate.

The focus on performance was retaining 0-allocations in hotpath (if using bound instruments), which this does achieve.   I'm sure there's more optimization we can do going forward.


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/FixedSizeExemplarReservoir.java
```diff
@@ -0,0 +1,182 @@
...
+    return storage.length;
+  }
+  /**
+   * Determines the sample bucket for a given measurement.
+   *
+   * @return The bucket to sample into or -1 for no sampling.
+   */
+  protected int bucketFor(double value, Attributes attributes, Context context) {
+    // TODO: hijackable random for testing.
+    long index = ThreadLocalRandom.current().nextLong(numMeasurements.sum() + 1);
```
jkwatson, line 65: Just FYI, TLR is super flaky and unreliable on Android (depending on API version). It is always initialized with the same seed, making it terrible for anything you actually want to be random.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/FixedSizeExemplarReservoir.java
```diff
@@ -0,0 +1,182 @@
...
+
+    private boolean hasValue = false;
+    private double value;
+    private Attributes attributes;
+    private String spanId;
+    private String traceId;
+    private long recordTime;
+
+    void offerMeasurement(double value, Attributes attributes, Context context) {
+      lock.lock();
```
jkwatson, line 121: It's worth trying this with an old-school `synchronized(object)` approach. We found that approach to be significantly faster for some usages. Definitely worthy of a micro-benchmark.

jsuereth: Noticing ~5ms improvement on that, going to update this with that change.   Still not happy with overall overhead of exemplars, but still chasing possible optimisations.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/FixedSizeExemplarReservoir.java
jsuereth: Agreed.  WDYT of adding something like this - https://github.com/census-instrumentation/opencensus-java/blob/master/impl_core/src/main/java/io/opencensus/implcore/trace/internal/RandomHandler.java

Specifically, I think there's a lot of reasons NOT to use random (or TLR) directly

jkwatson: right now, I'm sure you've seen, we have a quick-and-dirty Android detector to decide which IdGenerator to use. I could imagine we abstract that to include all random's that are needed in the SDK.

jsuereth: separate PR, or include in this one?

jkwatson: If you want a separate PR, please create an issue to track it so we don't forget about it.

jsuereth: Added it in this PR, but definitely needs a look over from you and @anuraaga

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/TracerSharedState.java
```diff
@@ -37,9 +37,7 @@
       List<SpanProcessor> spanProcessors) {
     this.clock = clock;
     this.idGenerator = idGenerator;
-    this.idGeneratorSafeToSkipIdValidation =
-        idGenerator instanceof RandomIdGenerator
-            || idGenerator instanceof AndroidFriendlyRandomIdGenerator;
+    this.idGeneratorSafeToSkipIdValidation = idGenerator instanceof RandomIdGenerator;
```
jkwatson, line 40: 👍🏽

sdk/common/src/main/java/io/opentelemetry/sdk/common/RandomHolder.java
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.common;
```
anuraaga, line 6: This should probably go in internal

sdk/common/src/main/java/io/opentelemetry/sdk/common/RandomHolder.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.util.Random;
+import javax.annotation.concurrent.ThreadSafe;
+
+/** Interface that pulls the most efficient random per-platform. */
+@ThreadSafe
+public interface RandomHolder {
```
anuraaga, line 13: `RandomSupplier` seems a bit more idiomatic. Actually I guess this can be a util class which returns `Supplier<Random>` rather than defining an interface

sdk/common/src/main/java/io/opentelemetry/sdk/common/RandomHolder.java
```diff
@@ -0,0 +1,50 @@
...
+  static RandomHolder androidFriendly() {
+    return AndroidFriendlyRandomHolder.INSTANCE;
+  }
+
+  /**
+   * Returns a custom random number holder.
+   *
+   * @param random The implementation of random to use.
+   */
+  static RandomHolder create(Random random) {
```
anuraaga, line 47: Using `Supplier<Random>` allows removing this I think

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/AbstractFixedSizeExemplarReservoir.java
```diff
@@ -0,0 +1,152 @@
...
+    if (bucket != -1) {
+      this.storage[bucket].offerMeasurement(value, attributes, context);
+    }
+  }
+
+  @Override
+  public final List<Exemplar> collectAndReset(Attributes pointAttributes) {
+    // Note: we are collecting exemplars from buckets piecemeal, but we
+    // could still be sampling exemplars during this process.
+    List<Exemplar> results = new ArrayList<>();
```
anuraaga, line 74:
```suggestion
    List<Exemplar> results = new ArrayList<>(storage.length);
```

I guess it's common to have a full reservoir?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/FixedSizeExemplarReservoir.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * A Reservoir sampler with fixed size that stores the given number of exemplars.
+ *
+ * <p>This implementation uses a un-unweighted/naive algorithm for sampler where the probability of
+ * sampling decrease as the number of observations continue. The collectAndReset method resets the
+ * count of observations, making the probability of sampling effectively 1.0.
+ *
+ * <p>Additionally this implementation ONLY exports double valued exemplars.
+ */
+public class FixedSizeExemplarReservoir extends AbstractFixedSizeExemplarReservoir {
```
anuraaga, line 23:
```suggestion
public final class FixedSizeExemplarReservoir extends AbstractFixedSizeExemplarReservoir {
```

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/FixedSizeExemplarReservoir.java
```diff
@@ -0,0 +1,56 @@
...
+  private final LongAdder numMeasurements = new LongAdder();
+
+  /**
+   * Instantiates an exemplar reservoir of fixed size.
+   *
+   * @param clock The clock to use when annotating measurements with time.
+   * @param size The number of exemplars to preserve.
+   * @param randomHolder The random number generated to use for sampling.
+   */
+  public FixedSizeExemplarReservoir(Clock clock, int size, RandomHolder randomHolder) {
```
anuraaga, line 34: We'd generally expose these as factory methods on `ExemplarReservoir` instead of making the classes themselves public, is that possible?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/FixedSizeExemplarReservoir.java
```diff
@@ -0,0 +1,56 @@
...
+    super(clock, size);
+    this.randomHolder = randomHolder;
+  }
+
+  @Override
+  protected int bucketFor(double value, Attributes attributes, Context context) {
+    // Purposefuly truncate here.
+    int count = (int) numMeasurements.sum() + 1;
+    int index = this.randomHolder.getRandom().nextInt(count > 0 ? count : 1);
```
anuraaga, line 43: Is there a reason to use random instead of just rotating through the cells (we do so here if copy-pasta can help https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/zpages/src/main/java/io/opentelemetry/sdk/extension/zpages/SpanBucket.java#L20)? Especially given the "we are collecting exemplars from buckets piecemeal", there doesn't seem to be any probability guarantee and FIFO seems like it could be better, it seems like there would never be anything better about keeping an old exemplar around compared to a new one, while having the disadvantage of being less fresh.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/AbstractFixedSizeExemplarReservoir.java
```diff
@@ -0,0 +1,152 @@
...
+  public final void offerMeasurement(double value, Attributes attributes, Context context) {
+    int bucket = bucketFor(value, attributes, context);
+    if (bucket != -1) {
+      this.storage[bucket].offerMeasurement(value, attributes, context);
+    }
+  }
+
+  @Override
+  public final List<Exemplar> collectAndReset(Attributes pointAttributes) {
+    // Note: we are collecting exemplars from buckets piecemeal, but we
```
anuraaga, line 72: This probably should be on a javadoc, though I guess not here since it's a package private class

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/exemplar/FixedSizeExemplarReservoirTest.java
```diff
@@ -0,0 +1,129 @@
...
+import io.opentelemetry.api.trace.TraceFlags;
+import io.opentelemetry.api.trace.TraceState;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.sdk.common.RandomHolder;
+import io.opentelemetry.sdk.testing.time.TestClock;
+import java.time.Duration;
+import java.util.Random;
+import org.junit.jupiter.api.Test;
+
+public class FixedSizeExemplarReservoirTest {
```
anuraaga, line 26: Jupiter, so no public class / method per jupiter idiom

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/FixedSizeExemplarReservoir.java
jkwatson: IIRC, randomly picking a replacement is actually very important in the statistics of reservoir sampling. If you do it in order, you tend to favor some things over others, and I believe the goal is to be as unbiased as possible. I am not an expert in this, however.

jsuereth: This basically an implementation of [Algorithm R - the simple algorithm on wikipedia](https://en.wikipedia.org/wiki/Reservoir_sampling), which has good probabilistic sampling of points.

As @jkwatson already said, this gives us better randomness in sampling.   Always keeping latest != better, from an observabiltiy perspective.   Increasing randomness *hopefully* helps us attain an exemplar of usefulness.   This algorithm is nice in that it guarantees we get at least ONE exemplar for low-volume measurement streams, but also provides a nice randomness to avoid bias.

The other factor here, that may not be apparent and should be, is that exemplar reservoir is cleaned every collection cycle.   This means the count is reset during a reporting interval, so we're only sampling from the most recent time period.

sdk/common/src/main/java/io/opentelemetry/sdk/common/RandomHolder.java
jsuereth: Moved!

sdk/common/src/main/java/io/opentelemetry/sdk/common/RandomHolder.java
jsuereth: You're speaking my (functional) language now.  Great suggestion!

sdk/common/src/main/java/io/opentelemetry/sdk/common/RandomHolder.java
jsuereth: Yep, cleaned things up nicely.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/AbstractFixedSizeExemplarReservoir.java
jsuereth: Unclear, actually.  For histograms I'd argue it's NOT common given our fixed size bucketing strategy.

Can we defer this for some sort of profiling?

docs/apidiffs/current_vs_latest/opentelemetry-sdk-common.txt
```diff
@@ -1,2 +1,8 @@
 Comparing source compatibility of  against 
-No changes.
\ No newline at end of file
++++  NEW INTERFACE: PUBLIC(+) ABSTRACT(+) io.opentelemetry.sdk.common.RandomHolder  (not serializable)
+	+++  CLASS FILE FORMAT VERSION: 52.0 <- n.a.
+	+++  NEW SUPERCLASS: java.lang.Object
+	+++  NEW METHOD: PUBLIC(+) STATIC(+) io.opentelemetry.sdk.common.RandomHolder androidFriendly()
+	+++  NEW METHOD: PUBLIC(+) STATIC(+) io.opentelemetry.sdk.common.RandomHolder create(java.util.Random)
+	+++  NEW METHOD: PUBLIC(+) ABSTRACT(+) java.util.Random getRandom()
+	+++  NEW METHOD: PUBLIC(+) STATIC(+) io.opentelemetry.sdk.common.RandomHolder platformDefault()
```
jkwatson: do we really need these to be added to the public API? I think if they need to be accessed outside the package, we should definitely make them `internal`

jsuereth: Moved :)

jkwatson: I somehow caught github in some weird temporary state where this was showing. sorry for the noise.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/FixedSizeExemplarReservoir.java
jsuereth: Yep, fixing now.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/exemplar/FixedSizeExemplarReservoirTest.java
jsuereth: still not used to jupiter, fixing.

docs/apidiffs/current_vs_latest/opentelemetry-sdk-common.txt
jsuereth: You started reviewing as I was pushing changes :)  Not noise, you literally commented as I was fixing locally.  Concurrency at its finest.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/AbstractFixedSizeExemplarReservoir.java
jsuereth: It's a bit of a concurrency/implementation detail.   Nominally subclasses shouldn't care/notice.  If it shows up, it's a bit of an (acceptable) bug.  I think fixing this to be perfect would impact performance way too much to be worth it.  It's possible we can fix this with a timestamp solution too.  The odds this happens is so esoterically low, I don't think it's worth loosing performance BUT would be good if someone else agrees with that assessment of risk vs. reward.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/AbstractFixedSizeExemplarReservoir.java
```diff
@@ -0,0 +1,152 @@
...
+  protected final int maxSize() {
+    return storage.length;
+  }
+  /**
+   * Determines the sample bucket for a given measurement.
+   *
+   * @return The bucket to sample into or -1 for no sampling.
+   */
+  protected abstract int bucketFor(double value, Attributes attributes, Context context);
+  /** Callback to reset any local state after a {@link #collectAndReset} call. */
```
jkwatson, line 54:
```suggestion

    /** Callback to reset any local state after a {@link #collectAndReset} call. */
```

jsuereth: ugh... why won't spotlessApply fix this for me... (FYI - I looked into adding this to spotlessApply and got lost, so it's on me).

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/exemplar/FixedSizeExemplarReservoirTest.java
```diff
@@ -0,0 +1,135 @@
...
+                    .hasValue(1)
+                    .hasFilteredAttributes(all)
+                    .hasTraceId(TRACE_ID)
+                    .hasSpanId(SPAN_ID));
+  }
+
+  @Test
+  public void multiMeasurements_preservesLatestSamples() {
+    AttributeKey<Long> key = AttributeKey.longKey("K");
+    // We cannot mock random in latest jdk, so we create an override.
```
jkwatson, line 111: curious what changed with this. Any ideas?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/AbstractFixedSizeExemplarReservoir.java
```diff
@@ -0,0 +1,152 @@
...
+
+  protected final int maxSize() {
+    return storage.length;
+  }
+  /**
+   * Determines the sample bucket for a given measurement.
+   *
+   * @return The bucket to sample into or -1 for no sampling.
+   */
+  protected abstract int bucketFor(double value, Attributes attributes, Context context);
```
jkwatson, line 53: I wonder if this method shouldn't be "reservoirIndexFor", since we don't really call things "buckets" unless it's a histogram, I don't think.

jkwatson (approved):
Looks ok to me.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/exemplar/FixedSizeExemplarReservoirTest.java
jsuereth: The error showed bytebuddy looking up some name that was considered JDK internal in 17.  I'm not really sure what it was, just decided to immediately workaround.


Here's the error:

```
org.mockito.exceptions.base.MockitoException: 
    Mockito cannot mock this class: class java.util.Random.

    Mockito can only mock non-private & non-final classes.
    If you're not sure why you're getting this error, please report to the mailing list.


    Java               : 17
    JVM vendor name    : Azul Systems, Inc.
    JVM vendor version : 17-ea+35
    JVM name           : OpenJDK 64-Bit Server VM
    JVM version        : 17-ea+35
    JVM info           : mixed mode, sharing
    OS name            : Mac OS X
    OS version         : 10.15.7


    Underlying exception : java.lang.IllegalStateException: Cannot access annotation property public abstract java.lang.String jdk.internal.util.random.RandomSupport$RandomGeneratorProperties.name()
        at app//io.opentelemetry.sdk.metrics.exemplar.FixedSizeExemplarReservoirTest.multiMeasurements_preservesLatestSamples(FixedSizeExemplarReservoirTest.java:113)

        Caused by:
        java.lang.IllegalStateException: Cannot access annotation property public abstract java.lang.String jdk.internal.util.random.RandomSupport$RandomGeneratorProperties.name()
```

jkwatson: interesting. I wonder if it's a mockito bug that they know about.

sdk/common/src/main/java/io/opentelemetry/sdk/internal/RandomSupplier.java
```diff
@@ -0,0 +1,44 @@
...
+
+  /**
+   * Returns an android-friendly random holder.
+   *
+   * <p>On android, ThreadLocalRandom is instantiated with the same initial seed on all new threads
+   * leading to poor randomness.
+   */
+  public static Supplier<Random> androidFriendly() {
+    return AndroidFriendlyRandomHolder.INSTANCE;
+  }
```
anuraaga, line 43:
```suggestion
  public static Supplier<Random> platformDefault() {
    // note: check borrowed from OkHttp's check for Android.
    if ("Dalvik".equals(System.getProperty("java.vm.name"))) {
      return AndroidFriendlyRandomHolder.INSTANCE;
    }
    return ThreadLocalRandom::current;
  }
```

anuraaga (approved):

PR #3595, opened 2021-09-10
Remove functions from CodedOutputStream we don't need.

anuraaga:
Our current pattern of marshaling always writes strings as byte[], and writes tags out separately, so removed support for them. Didn't remove things like support for `fixed32`, a proto type that OTLP doesn't happen to use now but no reason it may not come in the future. So this PR doesn't remove any ability to marshal, just code we copied in that isn't necessary with our patterns for any sort of OTLP message.


exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/CodedOutputStream.java
```diff
@@ -900,72 +642,6 @@ void reset(OutputStream out) {
       totalBytesWritten = 0;
     }
 
-    @Override
-    void writeTag(final int fieldNumber, final int wireType) throws IOException {
```
jsuereth, line 904: We're always doing our own joins for field number + wire type?

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/WireFormat.java
```diff
@@ -48,13 +48,7 @@
   static final int MAX_VARINT64_SIZE = 10;
   static final int MAX_VARINT_SIZE = 10;
 
-  static final int WIRETYPE_VARINT = 0;
-  static final int WIRETYPE_FIXED64 = 1;
-  static final int WIRETYPE_LENGTH_DELIMITED = 2;
-  static final int WIRETYPE_FIXED32 = 5;
```
jsuereth, line 54: We do have fixed32 usage in the logs proto in OTLP and we may have some coming in metrics in OTLP 0.10

jsuereth (approved):
Definitely a fan of reducing this API.

Regarding type support, we do have fixed32 in the log proto right now.  Also, do you have the wiretype enum values duplicated elsewhere?

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/CodedOutputStream.java
```diff
@@ -61,11 +58,12 @@
 // - No support for ByteString or ByteBuffer
 // - No support for message set extensions
 // - No support for Unsafe
+// - No support for Java String, only UTF-8 bytes
+// - No support for writing fields with tag, we alway write tags separately
```
kubawach, line 62: 👌

kubawach (approved):

anuraaga:
Yeah we use the wire types at compile time here

https://github.com/open-telemetry/opentelemetry-java/blob/main/buildSrc/src/main/kotlin/io/opentelemetry/gradle/ProtoFieldsWireHandler.kt#L82

When extracting the Serializer interface I noticed a significant performance regression which I attribute to Java not being able to do constant folding on the varint tag computation. So we precompute the tags instead in that handler.

anuraaga:
@jkwatson Ok to merge this?

jkwatson:
> @jkwatson Ok to merge this? It should bring our codecov up

Do it

Issue #3594, opened 2021-09-10
[Metrics] An Aggregation Temporality of delta，isMonotonic always true


clarck:
**Describe the bug**
Use AggregationTemporality.DELTA mode create MetricData，isMonotonic always is true

**Steps to reproduce**
```java
SdkMeterProvider meterProvider = SdkMeterProvider.builder()
                .registerView(InstrumentSelector.builder()
                                .setInstrumentType(InstrumentType.COUNTER)
                                .build(),
                                View.builder()
                                .setLabelsProcessorFactory(LabelsProcessorFactory.noop())
                                .setAggregatorFactory(AggregatorFactory.count(AggregationTemporality.DELTA))
                                .build())
                .build();
        GlobalMeterProvider.set(meterProvider);
```

Export data results:
```
...
 "metrics":[
                        {
                            "description":"Processed jobs",
                            "sum":{
                                "aggregation_temporality":"AGGREGATION_TEMPORALITY_DELTA",
                                "data_points":[
                                    {
                                        "exemplars":[

                                        ],
                                        "labels":[
                                            {
                                                "key":"request.cmd",
                                                "value":"cmd"
                                            }
                                        ],
                                        "start_time_unix_nano":1631194617077000000,
                                        "time_unix_nano":1631194627089000000,
                                        "value":2
                                    },
                                    {
                                        "exemplars":[

                                        ],
                                        "labels":[
                                            {
                                                "key":"request.cmd.cost",
                                                "value":"cost_time"
                                            }
                                        ],
                                        "start_time_unix_nano":1631194617077000000,
                                        "time_unix_nano":1631194627089000000,
                                        "value":2
                                    }
                                ],
                                "is_monotonic":true
                            },
                            "name":"processed_jobs2",
                            "unit":"1"
                        }
                    ]
...
```

**What did you expect to see?**
Use AggregationTemporality.CUMULATIVE mode create MetricData，isMonotonic is true; AggregationTemporality.DELTA mode create MetricData，isMonotonic is false.

jack-berg:
I believe the behavior you've outlined is correct.

The [metrics data model](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#sums) says the following about the monotonic flag on sum data points:
> For delta monotonic sums, this means the reader SHOULD expect non-negative values.
For cumulative monotonic sums, this means the reader SHOULD expect values that are not less than the previous value.

[...]

PR #3596, opened 2021-09-10
Fix An Aggregation Temporality of delta，isMonotonic always true. (#3594)

clarck:
Fix An Aggregation Temporality of delta，isMonotonic always true.


jsuereth:
The count aggregation only counts the number of measurements (not the sum of their values).  It should always be monotonic And starts at 0.

clarck:
> The count aggregation only counts the number of measurements (not the sum of their values). It should always be monotonic And starts at 0.

Yes, you are right. I understand it as a description of increment and accumulation. After all, the cumulative reported value is more and more, and the incremental report is not necessarily

PR #3600, opened 2021-09-10
Upgrade Gradle of examples to 7.2., fix println using non-existent property

Oberon00:
By using the same (or newer) gradle version as in the main build we can in theory start looking into making the examples buildable with a composite build (cf. https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3626)


examples/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.8.1-bin.zip
-distributionSha256Sum=fd591a34af7385730970399f473afabdb8b28d57fd97d6625c388d090039d6fd
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.2-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
+distributionSha256Sum=f581709a9c35e9cb92e16f585d2c4bc99b2b1a5f85d2badbd3dc6bff59e1e6dd
```
Oberon00, line 6: Copied from main build.

examples/settings.gradle
```diff
@@ -2,7 +2,7 @@ pluginManagement {
     plugins {
         id "com.diffplug.spotless" version "5.6.1"
         id "com.github.johnrengelman.shadow" version "6.1.0"
-        id 'com.google.protobuf' version '0.8.8'
+        id 'com.google.protobuf' version '0.8.17'
```
Oberon00, line 5: Had to upgrade this since the old version was accessing the no longer existing `compile` configuration internally.

examples/build.gradle
```diff
@@ -3,7 +3,16 @@ plugins {
     id "com.github.johnrengelman.shadow" apply false
 }
 
-println("Building against OpenTelemetry version: ${project.properties["io.opentelemetry.version"]}")
```
Oberon00, line 6: This always printed null, and a property which this name is nowhere to be found and not used anywhere else.

examples/build.gradle
```diff
@@ -3,7 +3,16 @@ plugins {
...
+ext {
```
Oberon00, line 6: Moved the ext block outside the subprojects block since subprojects will fall back to their parent project's ext properties and this makes it possible to print the version here.

examples/build.gradle
jkwatson: yep, this was left over from when we could build against different versions with a provided property. Good catch!

examples/gradle/wrapper/gradle-wrapper.properties
jkwatson: I think gradle itself will generate these for you if you use the `wrapper` task, but this is also fine. :)

jkwatson (approved):
Thanks!

examples/gradle/wrapper/gradle-wrapper.properties
Oberon00: I did use the wrapper task, but it removed the SHA 😕  maybe there is some extra option.

jkwatson: weird. it shouldn't do that, I wouldn't think.

PR #3551, opened 2021-08-25
Add attribute length limits to SpanLimits.

jack-berg:
The spec [states](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attribute-limits) that the SDK should provide a way to truncate the length of attributes, and that these limits should be configurable via [environment variables](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#span-limits-).

This PR starts that work by expanding `SpanLimits` to include `getMaxAttributeLength()`. A followup PR would add environment variable configurable to the autoconfigure module.

PR #3602, opened 2021-09-10
Add autoconfigure for span attribute value length limits

jack-berg:
Followup to Issue #3551.

Also noticed an inconsistency between the name of the new span attribute value length setter and getter so I've fixed that as well. Will be good to get that aligned before the release 😬!


jkwatson (approved):
Looks good, and good catch on the mis-aligned method names.

PR #3603, opened 2021-09-10
Dumb down project naming.

Oberon00:
Instead of using tricks to be able to use shorter project names, just name the folders like we want the final published artifact to be named.
This should enable composite builds to use opentelemetry-java without hassle, but I'm currently still verifying this (there seems to be an issue with the substituted bom or similar)


Oberon00:
Wish there was a way to hide "Renamed without changes" 😕

buildSrc/src/main/kotlin/otel.publish-conventions.gradle.kts
```diff
@@ -14,10 +27,8 @@ publishing {
         versionParts[0] += "-${release}"
         version = versionParts.joinToString("-")
       }
-      groupId = "io.opentelemetry"
       afterEvaluate {
         // not available until evaluated.
-        artifactId = base.archivesName.get()
```
Oberon00, line 20: Getting rid of this and ensuring that project name == archive name == artifact ID was key here.

buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -175,7 +167,7 @@ dependencies {
   dependencyManagement(platform(project(":dependencyManagement")))
   afterEvaluate {
     configurations.configureEach {
-      if (isCanBeResolved && !isCanBeConsumed) {
+      if (!isCanBeConsumed && this != dependencyManagement) {
```
Oberon00, line 170: This change is required for the composite build to work:

So this is not enough to make composite builds work as there is at least one other problem, namely how dependency management is set up. The constraints seem to be only applied if the ultimately built project has the constraints applied. For example, without this change, if I try inside the examples subdirectory:

```
$ ./gradlew classes --include-build ..

...

FAILURE: Build failed with an exception.

* What went wrong:
Could not determine the dependencies of task ':opentelemetry-examples-zipkin:compileJava'.
> Could not resolve all task dependencies for configuration ':opentelemetry-examples-zipkin:compileClasspath'.
   > Could not find io.zipkin.reporter2:zipkin-reporter:.
     Required by:
         project :opentelemetry-examples-zipkin > project :odin-java:exporters:opentelemetry-exporter-zipkin
```

We can see that `opentelemetry-exporter-zipkin` has the dependency `api("io.zipkin.reporter2:zipkin-reporter")`. However, the api configuration extends from nothing, so it does not have dependency management applied, and opentelemetry-examples-zipkin:compileClasspath does not use the java-conventions plugin so it does not have dependency management applied either.

buildSrc/src/main/kotlin/otel.publish-conventions.gradle.kts
```diff
@@ -1,10 +1,23 @@
+import gradle.kotlin.dsl.accessors._33c50961cbf9e9e490400740be7b5d25.base
+
 plugins {
   `maven-publish`
   signing
 
   id("otel.japicmp-conventions")
 }
 
+base {
```
Oberon00, line 10: Moved from `buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts`

examples/sdk-usage/src/main/java/io/opentelemetry/sdk/example/ConfigureSpanProcessorExample.java
```diff
@@ -25,6 +27,9 @@
   private static final Tracer tracer = openTelemetry.getTracer("ConfigureSpanProcessorExample");
 
   public static void main(String[] args) {
+    System.out.println(
```
Oberon00, line 30: I added this to test the included build, I can remove or adapt it if desired.

settings.gradle.kts
```diff
@@ -25,54 +25,75 @@ dependencyResolutionManagement {
...
+include(":sdk-extensions:opentelemetry-sdk-extension-autoconfigure-spi")
+include(":sdk-extensions:opentelemetry-sdk-extension-aws")
+include(":sdk-extensions:opentelemetry-sdk-extension-logging")
+include(":sdk-extensions:opentelemetry-sdk-extension-resources")
+include(":sdk-extensions:opentelemetry-sdk-extension-tracing-incubator")
+include(":sdk-extensions:opentelemetry-sdk-extension-jaeger-remote-sampler")
+include(":sdk-extensions:opentelemetry-sdk-extension-jfr-events")
+include(":sdk-extensions:opentelemetry-sdk-extension-zpages")
+
+// Map project names to directory names.
```
Oberon00, line 77: I added this instead of just renaming the folders for now. I would prefer renaming the folders but that makes the PR *huge* (1070 changed files). Can do in a follow-up.

anuraaga:
As someone who writes a lot of code in this repo, the longer names will definitely hurt the contributor experience I think. Publishing snapshots usually works easily enough and I don't think we've had as much need to write -instrumentation code against HEAD -java code since releasing 1.0. The linked issue problem about gradle-plugins will probably be handled without using a composite build https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4112

Composite builds have been around for a while but the limitations seem to not be getting fixed - this makes me think the feature is not very popular.

So I'm not too enthusastic about this change TBH.

Oberon00:
I was thinking that it is odd that our artifact IDs start with `opentelemetry-` despite already having a group of `io.opentelemetry`. Renaming our artifacts would be one way to fix the names. We could publish POM-only artifacts with an api dependency on the new name under the old name.

For me, right now I always need to squint hard to find the artifact name for any given project.

Oberon00:
> As someone who writes a lot of code in this repo, the longer names will definitely hurt the contributor experience I think.

Would the typical contributor write much Gradle code?

iNikem:
> I was thinking that it is odd that the artifact IDs start with `opentelemetry-` despite already having a group of `io.opentelemetry`. Renaming our artifacts would be one way to fix the names. We could publish POM-only artifacts with an api dependency on the new name under the old name.
> 
> Also, for me, right now I always need to squint hard to find the artifact name for any given project.

`opentelemetry-` prefix is important when you look at the list of jar files your application has/distributes. If you see `sdk.jar` it will tell you nothing what SDK is it about. `opentelemetry-sdk.jar` is much more telling. So that prefix is nor for consuming artifacts from the Maven Central, but for later accountability in application distribution package.

Oberon00:
That makes sense. In that case, it might be possible to rename just the jar but not the artifact ID? But I don't see the problem with long project names anyway and would prefer to avoid any extra configuration if possible, since that makes everything more complex.

iNikem:
I agree that simplicity of having project names equal to the desired artifact names and GAV coordinates is important enough to accept longer folder names during development.

anuraaga:
> Would the typical contributor write much Gradle code?

The typical contributor will run many Gradle tasks which is what gets more tedious.

> I agree that simplicity of having project names equal to the desired artifact names and GAV coordinates is important enough to accept longer folder names during development.

We should use the same pattern in our three repos for sanity - so `./gradlew :instrumentation:opentelemetry-api:opentelemetry-api-1.0:opentelemetry-javaagent-instrumentation-api-1.0:check`? :P

Flattening into just `./gradlew :opentelemetry-javaagent-instrumentation-api-1.0:check` is sort of ok in isolation, but has its own problems (javaagent library testing artifacts get randomly placed).

Losing control of your folder structure seems like a reasonable downside that is being missed. I'm still a -1 vote on it.

Oberon00:
>The typical contributor will run many Gradle tasks which is what gets more tedious. 

That's a point. Counterpoint: I usually don't run per project or I use the command line history, so the longer names wouldn't bother me. Also, you can abbreviate kebap-case names: https://docs.gradle.org/current/userguide/command_line_interface.html#sec:name_abbreviation E.g. `./gradlew :exp:oEP:classes` actually executes `:exporters:opentelemetry-exporter-prometheus:classes`

Oberon00:
> `./gradlew :instrumentation:opentelemetry-api:opentelemetry-api-1.0:opentelemetry-javaagent-instrumentation-api-1.0:check`

Should be doable with `./gradlew :instrumentation:oA:oA-1.0:oJIA-1.0:check`. (using `oJavaagentIA` or variants should also work)

anuraaga:
`./gradlew :instrumentation:oA:oA-1.0:oJIA-1.0:check` - I guess this sort of works (personally it takes so long to generate kebab case shorthand in my hand but admittedly should be able to get used to it).

We're still losing control of our folder naming scheme, which I think is a big enough downside. Basically we have to decide whether we accept tradeoffs to support composite builds - given that these bugs have been long-standing within Gradle, my inclination is not to support them. Publishing to maven local tends to work well enough for those times when doing cross-repo development, which I don't think happens that much anymore.

While we theoretically can apply this pattern to the instrumentation repo, I'm pretty sure we wouldn't. I think compared to this repo, there it's just way too big of a regression compared to our current, at least IMO, easy to understand structure of three folders per instrumentation clearly being subcomponents of that project.

https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/grpc-1.6

Currently 99% of the build logic is shared and I'm interested in publishing a Gradle plugin soon so we can actually use just one source for them. This change I honestly expect to toss a big wrench into that.

I am -1 for this still, but wouldn't block if there are enough votes otherwise and someone is committed to applying the pattern across the three repos. @trask @jkwatson please share your thoughts :)

jkwatson:
I'm against this change unless there is a really compelling reason for it.

trask:
> @trask @jkwatson please share your thoughts :)

I added to next week Tue SIG meeting agenda so we can discuss pros/cons of applying to the instrumentation repo with @iNikem since he opened the original issue.

Oberon00:
As an alternative to this, there is also the possibility to declare substitutions in the consuming project: https://docs.gradle.org/current/userguide/composite_builds.html#included_build_declaring_substitutions
Maybe there would be some way to provide some `.gradle` or `.gradle.kts` file that could be included by the consumer to enable composite builds.

> @anuraaga 
> Composite builds have been around for a while but the limitations seem to not be getting fixed - this makes me think the feature is not very popular.


> Basically we have to decide whether we accept tradeoffs to support composite builds - given that these bugs have been long-standing within Gradle, my inclination is not to support them.

The alternative conclusion that nobody changes the artifact ID, so nobody is affected by this limitation would be just as valid. At Dynatrace we use composite builds extensively for our larger repositories.


Maybe changing just the JAR name but not including `opentelemetry-` in the artifact ID would keep the composite build working.

iNikem:
What about keeping the folder structure but changing project names? Not artifacts names/coordinates, but sub-projects names? Just an idea, I haven't tried it.

Oberon00:
> What about keeping the folder structure but changing project names?

That's already what is being done here. The issue @anuraaga has are the task names, not the folder names AFAIK.

anuraaga:
Keeping folder names the same as the task path is the #1 priority. It's the only thing I would try to block regardless of a vote - it's just too confusing to know how to run commands if you can't trust the folder structure. This repo started with that, it was so hard.

The scheme I am still fairly receptive to is a 100% flat structure - all project folders are at the top level. I think it's possible to apply to this repo, but I think it's impossible to apply to the instrumentation repo because our artifact names do not sort well alphabetically there. Nor should they have to - this is a Gradle bug a PR to fix Gradle would have no tradeoffs :-D

anuraaga:
BTW filed https://github.com/gradle/gradle/issues/18291

Oberon00:
> Keeping folder names the same as the task path is the #1 priority.

I agree that this is desirable. This PR only lets them diverge because you would otherwise have 1070 files to review and the GH UI becomes very slow (see earlier PR versions).

anuraaga:
Yup I got that, this repo by itself we could then converge in the next PR on just the flat structure. I don't think it scales to repos like the instrumentation repo though - because we have such repos (large monorepos) we should take it into account. Loosing consistency among the repos is of course an option - but as possibly the only one who is proactive in all three it would be quite painful for me.

jkwatson:
Whatever happens here, we must not change the artifact names or maven coordinates of anything that we have marked stable at this point.

Oberon00:
Doesn't look like a gradle-side solution is coming 😕 https://github.com/gradle/gradle/issues/18291 was closed as duplicate of the very old https://github.com/gradle/gradle/issues/847.

If the main concern is gradle task names, maybe a task rule could be used to dispatch tasks like `:instrumentation-apache-camel-2.20-javaagent to `:instrumentation:apache-camel-2.20:opentelemetry-instrumentation-apache-camel-2.20-javaagent` (not sure how the current project name actually is). Though personally I think I would be content just using command line history or an IDE or the kebap abbreviations

iNikem:
Actually, I now learned that there is a way to solve my original problem in open-telemetry/opentelemetry-java-instrumentation#3626. Yes, automatic mapping of included project's modules to maven coordinates does not work for our repos atm. But you can manually specify dependency substitution rules as in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4179/files#diff-0b68cb406e55a2431167c05be6ceb1c7e64c13ef9f460dc0b9c80cc860a2ea7d. It's not ideal, but it works.

Oberon00:
How does this `.using(project(":"))` work? Does `:` refer to the current project or the root project? Wouldn't you need to specify the exact subproject?

iNikem:
> How does this `.using(project(":"))` work? Does `:` refer to the current project or the root project? Wouldn't you need to specify the exact subproject?

`using` select a project from the _included_ build. In the referenced case we include `gradle-plugin` project which does not have any subprojects.

iNikem:
> OK, so you basically have to spell out for gradle what the publish configuration does in a different format.
> I wonder if it would be possible to generate a `dependencySubstitution` for use in other projects automatically from open-telemetry java.

If we have just to strip `opentelemetry-` prefix, then sure :)

PR #3609, opened 2021-09-13
Split trace and metrics marshalers to packages instead of huge classes.

anuraaga:
GitHub doesn't render this diff well, but I just made some classes / methods in the common package `public` or `protected` and used IntelliJ move inner class refactoring for everything else.

All the new lines of code are import statements or javadoc, there is 0 additional bytecode.


jkwatson (approved):
I think this is a good change...makes it easier to understand where to find what.

PR #3611, opened 2021-09-13
Add OTLP integration test for gRPC and HTTP Trace and Metric exporters

jack-berg:
Per a conversation in last week's SIG meeting, I've added an integration test confirming compatibility between the OTLP exporters and the collector. This isn't (currently) an exhaustive test of all the fields available on the protos, but it should act as a decent starting point.

### DIFF
--- a/integration-tests/src/test/java/io/opentelemetry/OtlpExporterIntegrationTest.java
+++ b/integration-tests/src/test/java/io/opentelemetry/OtlpExporterIntegrationTest.java
@@ -57,16 +57,14 @@ import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.awaitility.Awaitility;
+import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Order;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.extension.RegisterExtension;
-import org.testcontainers.DockerClientFactory;
 import org.testcontainers.containers.BindMode;
 import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.containers.wait.strategy.Wait;
-import org.testcontainers.junit.jupiter.Container;
 import org.testcontainers.junit.jupiter.Testcontainers;
 import org.testcontainers.utility.DockerImageName;
 
@@ -83,7 +81,6 @@ class OtlpExporterIntegrationTest {
   private static final Integer COLLECTOR_OTLP_GRPC_PORT = 4317;
   private static final Integer COLLECTOR_OTLP_HTTP_PORT = 4318;
   private static final Integer COLLECTOR_HEALTH_CHECK_PORT = 13133;
-  private static final Integer IN_PROCESS_OTLP_GRPC_PORT = 4319;
   private static final Resource RESOURCE =
       Resource.getDefault().toBuilder()
           .put(ResourceAttributes.SERVICE_NAME, "integration test")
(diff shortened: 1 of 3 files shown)


integration-tests/src/test/java/io/opentelemetry/OtlpExporterIntegrationTest.java
```diff
@@ -0,0 +1,355 @@
...
+          .put(ResourceAttributes.SERVICE_NAME, "integration test")
+          .build();
+  private static final Logger LOGGER =
+      Logger.getLogger(OtlpExporterIntegrationTest.class.getName());
+
+  static {
+    if (DockerClientFactory.instance().isDockerAvailable()) {
+      // If docker is available, expose the port the in-process OTLP gRPC server will run on before
+      // the collector is initialized so the collector can connect to it.
+      exposeHostPorts(IN_PROCESS_OTLP_GRPC_PORT);
```
jkwatson, line 98: Is this necessary? I don't see us having to do this in, for example, the `JaegerRemoteSamplerIntegrationTest` which is needing to do something similar.

jkwatson: oh I see. We're sending OTLP back to this process again from the collector!

integration-tests/src/test/java/io/opentelemetry/OtlpExporterIntegrationTest.java
```diff
@@ -0,0 +1,355 @@
...
+                KeyValue.newBuilder()
+                    .setKey("key")
+                    .setValue(AnyValue.newBuilder().setStringValue("value").build())
+                    .build()));
+    assertThat(protoSpan.getEventsCount()).isEqualTo(1);
+    assertThat(protoSpan.getEvents(0).getName()).isEqualTo("event");
+    assertThat(protoSpan.getLinksCount()).isEqualTo(1);
+    Link link = protoSpan.getLinks(0);
+    assertThat(link.getTraceId()).isNotEmpty();
+    assertThat(link.getSpanId()).isNotEmpty();
```
jkwatson, line 220: If we save off the generated ids, we could easily assert them here, right?

jkwatson (approved):
Thanks!

integration-tests/src/test/java/io/opentelemetry/OtlpExporterIntegrationTest.java
```diff
@@ -0,0 +1,355 @@
...
+  private static final Integer IN_PROCESS_OTLP_GRPC_PORT = 4319;
+  private static final Resource RESOURCE =
+      Resource.getDefault().toBuilder()
+          .put(ResourceAttributes.SERVICE_NAME, "integration test")
+          .build();
+  private static final Logger LOGGER =
+      Logger.getLogger(OtlpExporterIntegrationTest.class.getName());
+
+  static {
+    if (DockerClientFactory.instance().isDockerAvailable()) {
```
anuraaga, line 95: I think we can do in a BeforeAll to avoid this check (we have it in the `TestContainers` annotation above)

integration-tests/src/test/java/io/opentelemetry/OtlpExporterIntegrationTest.java
anuraaga: I guess this will cause the test to fail if 4319 is already used? If that's right, it'd be nice if we could avoid it. If we move to BeforeAll, I guess we'd also be able to reference the server's automatic port (without passing an explicit value to `httpPort`)

integration-tests/src/test/java/io/opentelemetry/OtlpExporterIntegrationTest.java
jack-berg: That's true. We'll also have to manually start / stop the `GRPC_SERVER` and the `COLLECTOR` container since we'll need to call `exposeHostPorts(..)` after the grpcServer (so that we can use the randomly assigned port) and before the collector is started.

integration-tests/src/test/java/io/opentelemetry/OtlpExporterIntegrationTest.java
jack-berg: Yup!

integration-tests/src/test/java/io/opentelemetry/OtlpExporterIntegrationTest.java
jack-berg: Yes. Was being a bit lazy I suppose, but best to do everything we can to avoid port conflicts.

anuraaga (approved):
Cool, thanks!

PR #3621, opened 2021-09-15
Write enum values by name in JSON.

### LINKED ISSUE #3620: JSON enums should be serialized by name, not number
#3616 temporarily makes JSON logs worse by serializing enums by number. These can still be parsed by protobuf-json libraries, but we should restore them to being serialized by name for readability as logs are meant to be.

anuraaga:
Fixes #3620


jsuereth (approved):

jkwatson (approved):
Thanks!

so...if we were to use this format for sending data to the collector via the http/json protocol, would it work? Would the enums be properly grokked?

anuraaga:
Yeah, JSON parsing of proto supports both number or name so either would be fine. But writers are supposed to use the name, the small sidestep to number was not "as correct".

PR #3622, opened 2021-09-15
Remove experimental prefix from otlp protocol config properties

jack-berg:
Updating the config properties for protocol now that they're official in the [spec](https://github.com/open-telemetry/opentelemetry-specification/pull/1880).

### DIFF
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtil.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtil.java
@@ -22,9 +22,9 @@ final class OtlpConfigUtil {
   static final String DATA_TYPE_METRICS = "metrics";
 
   static String getOtlpProtocol(String dataType, ConfigProperties config) {
-    String protocol = config.getString("otel.experimental.exporter.otlp." + dataType + ".protocol");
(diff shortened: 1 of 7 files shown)


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtil.java
```diff
@@ -22,9 +22,9 @@
   static final String DATA_TYPE_METRICS = "metrics";
 
   static String getOtlpProtocol(String dataType, ConfigProperties config) {
-    String protocol = config.getString("otel.experimental.exporter.otlp." + dataType + ".protocol");
```
jkwatson, line 25: should we continue to support the `experimental` versions for one release, for those who might be using them?

jack-berg: Sure that seems reasonable.

jkwatson: I would have them documented, but clearly marked as deprecated in the docs.

jkwatson (approved):

anuraaga (approved):

PR #3624, opened 2021-09-15
Rename the logging AnyValue to Body and allow only String values for now.

jkwatson:


sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/Body.java
```diff
@@ -0,0 +1,62 @@
...
+@Immutable
+public abstract class Body {
+
+  /** An enum that represents all the possible value types for an {@code AnyValue}. */
+  public enum Type {
+    STRING
+  }
+
+  /**
+   * Returns an {@code AnyValue} with a string value.
```
breedx-splk, line 27:
```suggestion
   * Returns an {@code Body} with a string value.
```

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/Body.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * A class that represents all the possible values for a log message body. An {@code AnyValue} can
+ * currently only have 1 types of values: {@code String}, represented through {@code AnyValue.Type}.
+ * This class will likely be extended in the future to include additional body types supported by
+ * the OpenTelemetry log data model.
+ */
+@Immutable
+public abstract class Body {
+
+  /** An enum that represents all the possible value types for an {@code AnyValue}. */
```
breedx-splk, line 21:
```suggestion
  /** An enum that represents all the possible value types for an {@code Body}. */
```

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/Body.java
```diff
@@ -0,0 +1,62 @@
...
+  public enum Type {
+    STRING
+  }
+
+  /**
+   * Returns an {@code AnyValue} with a string value.
+   *
+   * @param stringValue The new value.
+   * @return an {@code AnyValue} with a string value.
```
breedx-splk, line 30:
```suggestion
   * @return a {@code Body} with a string value.
```

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/Body.java
```diff
@@ -0,0 +1,62 @@
...
+   */
+  public static Body stringBody(String stringValue) {
+    return StringBody.create(stringValue);
+  }
+
+  /** Returns the string value of this {@code Body}. */
+  public String asString() {
+    throw new UnsupportedOperationException(
```
breedx-splk, line 38: Since we have a string body, why not just return that. I find this exception surprising for a `Body` that just contains a string. Or maybe I'm just getting hung up on AutoValue stuff and it's handled by the `StringBody` subtype?

sdk-extensions/logging/src/test/java/io/opentelemetry/sdk/logging/data/BodyTest.java
```diff
@@ -0,0 +1,20 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+
+class BodyTest {
+
+  @Test
+  void stringValue() {
+    Body value = Body.stringBody("foobar");
+    assertThat(value.asString()).isEqualTo("foobar");
```
breedx-splk, line 17: 👍🏻

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/Body.java
```diff
@@ -0,0 +1,62 @@
...
+
+package io.opentelemetry.sdk.logging.data;
+
+import com.google.auto.value.AutoValue;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * A class that represents all the possible values for a log message body. An {@code AnyValue} can
+ * currently only have 1 types of values: {@code String}, represented through {@code AnyValue.Type}.
```
breedx-splk, line 14:
```suggestion
 * A class that represents all the possible values for a log message body. A {@code Body} can
 * currently only have 1 types of values: {@code String}, represented through {@code Body.Type}.
```

breedx-splk (changes requested):
I support this change!

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/Body.java
jkwatson: there will be a bunch more types of bodies in the future. yeah, this is a little weird, but this method implementation should never actually be hit by anything, since we'll only create things as sub-classes. I'd like to see if we can convince AutoValue that this should be an interface, rather than an abstract class.

breedx-splk (approved):

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/Body.java
jkwatson: looks like it's easy to make it an interface. Will amend.

jsuereth (approved):
Agree!  This sets us up for the future well, with a nice current state :)

jkwatson:
@ran8080 This will impact your PR. Any concerns?

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/data/Body.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  /** Returns the string value of this {@code Body}. */
+  String asString();
+
+  Type getType();
+
+  @Immutable
+  @AutoValue
+  abstract class StringBody implements Body {
```
anuraaga, line 43: Move to separate file (currently it is public since it's a member of an interface

jkwatson: done

anuraaga (approved):

PR #3625, opened 2021-09-16
Migrate Jaeger exporter to marshaler

anuraaga:
While performance of this exporter isn't such a huge deal as it's not our main protocol, since we do include it in the agent we can knock off a couple of MB from it since it's the only one that depends on protobuf-java / grpc-protobuf right now.


docs/apidiffs/current_vs_latest/opentelemetry-exporter-jaeger.txt
```diff
@@ -1,2 +1,1372 @@
 Comparing source compatibility of  against 
-No changes.
\ No newline at end of file
+---! REMOVED CLASS: PUBLIC(-) FINAL(-) io.opentelemetry.exporter.jaeger.proto.api_v2.Collector  (not serializable)
```
anuraaga, line 2: @jkwatson We accidentally had these in the public API. We need to make a semconv decision on keeping these or not. I attempted to keep without any change to the API but had trouble doing it without also vendoring in `google.protobuf.Timestamp` / `google.protobuf.Duration`. I can try harder but wondering if this could be considered an OK breaking change semconv wise since it seems unlikely they're used (and we could patch them back in if it caused a problem).

exporters/jaeger-proto/build.gradle.kts
```diff
@@ -0,0 +1,17 @@
+plugins {
+  id("otel.protobuf-conventions")
+  id("otel.publish-conventions")
```
anuraaga, line 3: We ourselves were using protos in another module, the sdk-autoconfigure tests. So I realized that the best we can do is to separate out this artifact. We could still make a call to remove this line until we get a complaint, though keeping this and calling out in release notes that users should add this artifact, would allow still maintaining semconv without worries.

exporters/jaeger/src/test/java/io/opentelemetry/exporter/jaeger/PostSpansRequestMarshalerTest.java
```diff
@@ -25,22 +27,31 @@
...
-import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 import org.junit.jupiter.api.Test;
 
-/** Unit tests for {@link Adapter}. */
-class AdapterTest {
+class PostSpansRequestMarshalerTest {
```
anuraaga, line 47: The GitHub diff makes obvious this is a line-by-line port of the previous test coverage to the marshaler. I think this is enough, so we don't need to do anything like the randomizer tests.

kubawach (approved):
LGTM (I see the "repeated" naming convention here ;) )

docs/apidiffs/current_vs_latest/opentelemetry-exporter-jaeger.txt
jkwatson: I think we definitely don't want to consider some proto classes as a part of our public APIs to support. I'm ok with hiding this class, and, as you said, if we have complaints we can add them in a patch release. I'd rather do that than expose all of those protobuf classes.

docs/apidiffs/current_vs_latest/opentelemetry-exporter-jaeger-proto.txt
```diff
@@ -0,0 +1,1355 @@
+Comparing source compatibility of  against 
++++  NEW CLASS: PUBLIC(+) FINAL(+) io.opentelemetry.exporter.jaeger.proto.api_v2.Collector  (not serializable)
+	+++  CLASS FILE FORMAT VERSION: 52.0 <- n.a.
+	+++  NEW METHOD: PUBLIC(+) STATIC(+) com.google.protobuf.Descriptors$FileDescriptor getDescriptor()
+	+++  NEW METHOD: PUBLIC(+) STATIC(+) void registerAllExtensions(com.google.protobuf.ExtensionRegistryLite)
+	+++  NEW METHOD: PUBLIC(+) STATIC(+) void registerAllExtensions(com.google.protobuf.ExtensionRegistry)
++++  NEW CLASS: PUBLIC(+) STATIC(+) FINAL(+) io.opentelemetry.exporter.jaeger.proto.api_v2.Collector$PostSpansRequest  (not serializable)
```
jkwatson, line 7: I'd definitely prefer it if we don't have to expose any of this stuff as a public API for otel.

anuraaga: We currently already do - so you'd like to try removing them from the public API and see if anyone has issues?

jkwatson: how were they not in the apidiffs previously?

jkwatson: oh, is this because we only generate diffs, and they've been public from the beginning of time? 😢

anuraaga: Sort of - the huge removal in the other artifact correspondence to this addition. So the API has moved artifacts while otherwise unchanged in its publicness from before.

jkwatson: yeah. ugh. This sort of makes me want to not touch this module, since this is a pretty big breaking change, if someone happens to be relying on this stuff for some unknown reason. I need to think on this a bit.

anuraaga: Sure - FWIW, with an artifact move and switch to compileOnly, it's easy to add this artifact to a build if the classes are being used. We don't consider needing additional POM changes to be semconv breaking so came up with this approach to try to keep it non-breaking.

exporters/jaeger-proto/build.gradle.kts
jkwatson: Do you mind adding a bit to the pending CHANGELOG for this as a part of this PR, so we don't forget? Thanks!

exporters/jaeger/src/main/java/io/opentelemetry/exporter/jaeger/BatchMarshaler.java
```diff
@@ -0,0 +1,46 @@
...
+import java.io.IOException;
+import java.util.List;
+
+final class BatchMarshaler extends MarshalerWithSize {
+
+  static BatchMarshaler create(List<SpanData> spans, Resource resource) {
+    SpanMarshaler[] spanMarshalers = SpanMarshaler.createRepeated(spans);
+    ProcessMarshaler processMarshaler = ProcessMarshaler.create(resource);
+    return new BatchMarshaler(spanMarshalers, processMarshaler);
+  }
```
jkwatson, line 23: Looks like this snuck up above the fields/constructor. ;)

exporters/jaeger/src/main/java/io/opentelemetry/exporter/jaeger/KeyValueMarshaler.java
```diff
@@ -0,0 +1,132 @@
...
+      return new ArrayList<>();
+    }
+
+    List<KeyValueMarshaler> marshalers = new ArrayList<>(attributes.size());
+    attributes.forEach((attributeKey, o) -> marshalers.add(create(attributeKey, o)));
+    return marshalers;
+  }
+
+  static KeyValueMarshaler create(AttributeKey<?> key, Object value) {
+    byte[] keyUtf8 = MarshalerUtil.toBytes(key.getKey());
```
jkwatson, line 41: static methods go below fields/constructor, correct?

exporters/jaeger/src/main/java/io/opentelemetry/exporter/jaeger/LogMarshaler.java
```diff
@@ -0,0 +1,71 @@
...
+
+    int droppedAttributesCount = event.getDroppedAttributesCount();
+    if (droppedAttributesCount > 0) {
+      fields.add(
+          KeyValueMarshaler.create(
+              KEY_EVENT_DROPPED_ATTRIBUTES_COUNT, (long) droppedAttributesCount));
+    }
+
+    return new LogMarshaler(timestamp, fields);
+  }
```
jkwatson, line 48: some more here.

exporters/jaeger/build.gradle.kts
```diff
@@ -12,21 +14,32 @@ dependencies {
   api(project(":sdk:all"))
   api("io.grpc:grpc-api")
 
-  implementation(project(":sdk:all"))
+  protoSource(project(":exporters:jaeger-proto"))
+
+  // TODO(anuraaga): otlp-common has a lot of code not specific to OTLP now. As it's just internal
+  // code, this mysterious dependency is possibly still OK but we may need a rename or splitting.
+  implementation(project(":exporters:otlp:common"))
```
jkwatson, line 21: yeah, this is pretty weird to see in here, agreed.

jkwatson:
Lots of cases of putting static methods above the constructor/fields, but generally looks like aside from that.

exporters/jaeger-proto/build.gradle.kts
anuraaga: Sure added

PR #3626, opened 2021-09-16
Remove SpanAdapter.

anuraaga:
With a release behind us it's an appropriate time to remove this now unused class.


exporters/otlp/common/src/test/java/io/opentelemetry/exporter/otlp/internal/traces/TraceRequestMarshalerTest.java
```diff
@@ -5,193 +5,298 @@
...
-          "0123456789abcdef",
-          TraceFlags.getSampled(),
-          TraceState.getDefault());
-  private static final SpanContext PARENT_SPAN_CONTEXT =
       SpanContext.create(TRACE_ID, SPAN_ID, TraceFlags.getSampled(), TraceState.getDefault());
 
   @Test
-  void marshalAndSizeRequest() throws IOException {
-    assertMarshalAndSize(Arrays.asList(testSpanData(), testSpanData(), testSpanData()));
+  void toProtoResourceSpans() {
```
anuraaga, line 65: Brought in the test cases from `SpanAdapterTest`

exporters/otlp/trace/src/test/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterTest.java
```diff
@@ -135,13 +139,27 @@ void testExport_MultipleSpans() {
...
-      assertThat(fakeCollector.getReceivedSpans())
-          .isEqualTo(SpanAdapter.toProtoResourceSpans(spans));
+      assertThat(fakeCollector.getReceivedSpans()).isEqualTo(toResourceSpans(spans));
     } finally {
       exporter.shutdown();
     }
   }
 
+  private static List<ResourceSpans> toResourceSpans(List<SpanData> spans) {
+    return Arrays.stream(ResourceSpansMarshaler.create(spans))
```
kubawach, line 149: NIT: this can be extracted into a test util class as duplicated in two places.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/traces/SpanEventMarshaler.java
```diff
@@ -21,25 +21,29 @@
   private final KeyValueMarshaler[] attributeMarshalers;
   private final int droppedAttributesCount;
 
-  static SpanEventMarshaler[] create(List<EventData> events) {
+  static SpanEventMarshaler[] createRepeated(List<EventData> events) {
```
kubawach, line 24: Also a NIT: I'd leave the "create" name overloaded - for single and list arguments as IMO it's a cleaner API.

kubawach (approved):
Looks good :)

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/traces/SpanEventMarshaler.java
anuraaga: This was actually the only place that missed our current naming scheme of createRepeated.

I think I like the different names but we could consider changing all of them in a different PR.

exporters/otlp/trace/src/test/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterTest.java
anuraaga: They are two different Gradle projects so it's not so simple. If we had an unpublished project like testing-internal I could put it in there but for now think a copy is fine.

PR #3631, opened 2021-09-17
Update more dependencies.

anuraaga:
@jkwatson I found the junit issue you mentioned, will sort it out separately


exporters/otlp/common/src/test/java/io/opentelemetry/exporter/otlp/internal/traces/SpanDataRandomizerRegistry.java
```diff
@@ -1,361 +0,0 @@
-/*
```
anuraaga, line 1: Forgot to delete this class in the other PR

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
```diff
@@ -172,7 +172,7 @@ private static void configurePrometheusMetrics(
     } catch (IOException e) {
       throw new IllegalStateException("Failed to create Prometheus server", e);
     }
-    Runtime.getRuntime().addShutdownHook(new Thread(server::stop));
+    Runtime.getRuntime().addShutdownHook(new Thread(server::close));
```
jkwatson, line 175: Is this the same thing, just renamed, or a different function?

jkwatson (approved):

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
anuraaga: Yeah - previous one is deprecated and our build prevents using deprecated stuff.

anuraaga: It does make me wonder how to release autoconfigure as stable with the dependency on this pre-1.0 Prometheus artifact which is known to break some times. We may need to go ahead and implement our own endpoint...

Issue #3550, opened 2021-08-25
Exponential Histogram Prototype


jsuereth:
This bug tracks work/progress around generating exponential histograms in the Java SDK.

- [x] Add exponential histogram to `sdk.metrics.data` package
- [ ] Add exponential histogram to OTLP exporters
- [ ] Create an exponential histogram Aggregator

jamesmoessis:
@jsuereth I'm going to implement this now that the OTEP seems in it's final form. Would these three items be three separate PRs or one big one?

jkwatson:
> @jsuereth I'm going to implement this now that the OTEP seems in it's final form. Would these three items be three separate PRs or one big one?

Please favor smaller, focussed PRs if at all possible. Thanks!

jamesmoessis:
PR for OTLP marshalers: https://github.com/open-telemetry/opentelemetry-java/pull/3995

Once the above PR gets merged, I think it's safe to close this issue.

jamesmoessis:
@jsuereth We have the data model, aggregator, and OTLP exporters all merged now - so we can close this one 🎉

PR #3637, opened 2021-09-20
Data classes for exponential histogram prototype (#3550)

jamesmoessis:
- This addresses the first task from #3550.
- Adds exponential histogram to `sdk.metrics.data`.
- No tests becase there aren't any existing tests for classes in this package 
- This PR should allow work on an aggregator for the exponential histogram to start. 

New to OTel contribution so any constructive feedback is welcomed.


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleExponentialHistogramBuckets.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+@AutoValue
+@Immutable
+public abstract class DoubleExponentialHistogramBuckets {
+  DoubleExponentialHistogramBuckets() {}
+
+  /**
+   * Create DoubleExponentialHistogramBuckets.
+   *
+   * @param offset Signed integer representing the bucket index of the first entry in bucketCounts.
```
jkwatson, line 26: Can offset be negative? If so, what does that mean, exactly?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleExponentialHistogramData.java
```diff
@@ -0,0 +1,44 @@
...
+ * <p><i>Note: This is called "DoubleExponentialHistogramData" to reflect which primitives are used
+ * to record it, however "ExponentialHistogram" is the equivalent OTLP type.</i>
+ */
+@Immutable
+@AutoValue
+public abstract class DoubleExponentialHistogramData
+    implements Data<DoubleExponentialHistogramPointData> {
+  DoubleExponentialHistogramData() {}
+
+  public static DoubleExponentialHistogramData create(
```
jkwatson, line 27: this method needs some javadoc. Thanks!

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleExponentialHistogramBuckets.java
jkwatson: Is "offset" here the exponent of the first bucket? So, `offset=-3` means that the first bucket is at `0.001` ?

If so, maybe `offset` isn't the best name for this parameter. `firstBucketExponent` maybe?

jsuereth: There's also a "scale" that determines a portion of the exponent.  

This is really the "number of skipped buckets" because we expect the area around 0 to be uninhabited, so it's an optimization.

Specifically you have   `(base^(offset+i)` IIRC (and a negative number means you're in fractional starting buckets).

The exponent also has 2^scale = base.   Not sure this helps with naming (I'm bad at naming) but I'd be a fan of sticking to the otel proto naming unless we really really hate the name.

jamesmoessis: The bucket boundaries are calculated via the `scale` and `offset`.

Firstly, `base = (2^(2^-scale))`, and then the bucket lower bound it calculated by `lower_bound = base^(i+offset)` like @jsuereth mentioned. The simplest case is when `scale = 0` and `offset=0`. That would mean your buckets are lower bounded at 1.0, 2.0, 4.0, 8.0, 16.0, etc. With `offset = -3` your buckets are lower bounded at 0.125, 0.25, 0.5, 1.0, 2.0, 4.0, 16.0, etc.

I think it makes sense to stick to the `offset` name since that's what it's called in the protocol. In terms of the javadoc description, I based it off the [protocol's description](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L535), but it could probably be clearer.

How does this sound instead? `@param offset Signed integer which shifts the bucket boundaries according to `lower_bound = base^(offset+i)`.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleExponentialHistogramData.java
jamesmoessis: I missed that, thanks. Just pushed it up 👍

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleExponentialHistogramBuckets.java
jkwatson: If you could put all that text that you wrote into the class level javadoc, that would be best. I don't want us to have to somehow depend on the specs or the proto definitions to understand the code here.

jamesmoessis: Alright, I've: 
- Put more detail on the class level javadocs, with examples of how bucket boundaries are calculated.  
- Put more detail on the `@param scale` and altered the `@param offset` javadoc to reflect my last comment.  
- Added javadoc for `getScale()` and `getOffset()`

jkwatson: Thanks! Very helpful for those of us who aren't intimately familiar with the art.

jkwatson:
> One question I have for @anuraaga and @jkwatson (and @jamesmoessis).
> 
> Given the cost of exponential histogram generation, would it make more sense for the "data" package to be an interface-only experience and we defer actual storage for an implementation? Specifically, the state of the art seems to be dynamic linked-list or bucket-expansion style classes. I think copying the data out of these representations into the data model may be expensive, looking to see if we can throw ourselves a bone by sticking to interfaces specifically.

Seems totally reasonable to me.

jsuereth:
Also @jamesmoessis Take a look at metrics-testing, where we have "assertions" that help make it easy to write unit tests.  Not required in this PR, but definitely have come in handy.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/DoubleExponentialHistogramPointData.java
```diff
@@ -0,0 +1,102 @@
...
+        count,
+        base,
+        zeroCount,
+        positiveBuckets,
+        negativeBuckets,
+        exemplars);
+  }
+
+  @Override
+  public abstract long getStartEpochNanos();
```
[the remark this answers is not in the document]
jamesmoessis: Sounds good, I've written them and pushed them up. I've also realised I could remove the overrides for this method and a couple of others.

jamesmoessis:
> Given the cost of exponential histogram generation, would it make more sense for the "data" package to be an interface-only experience and we defer actual storage for an implementation?

I agree, and it's why I've backed off implementing any real logic in this PR. I could probably move it to a pure interface. Is something like this what you are imagining? 

```
public interface ExponentialHistogramData extends Data<ExponentialHistogramPointData> {
  ...
}

---

public interface ExponentialHistogramPointData extends PointData {
  ...
}
```

> Specifically, the state of the art seems to be dynamic linked-list or bucket-expansion style classes.

Which implementations are you referring to here? I would love to have a look and see the best ways of implementing.

jsuereth:
@jamesmoessis 
> Is something like this what you are imagining?

Exactly.  The requirements I imagine for usefulness are:

1. A set of interface/methods that make extracting OTLP simple
2. A way of constructing "dummy" data classes when testing export-only paths
3. A set of assertj classes to simplify asserting results of instruments (similar to [this](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics-testing/src/main/java/io/opentelemetry/sdk/testing/assertj/metrics/DoubleHistogramPointDataAssert.java)).

Not suggesting anything but (1) for this specific PR.

> Which implementations are you referring to here? I would love to have a look and see the best ways of implementing.

- [HdrHistogram](https://github.com/HdrHistogram/HdrHistogram/tree/master/src/main/java/org/HdrHistogram) uses base10, but has similar behavior.
- [NrSketch](https://github.com/newrelic-experimental/newrelic-sketch-java) New Relic's algorithm specfically against this proposal.
- [Prometheus SparseHistogram (go)](https://github.com/prometheus/client_golang/tree/sparsehistogram)

To a lesser extent, you can look at [DDSketch ](https://github.com/DataDog/sketches-java), which may share similarities in storage, but not usage.

jsuereth:
Also @jmacd added more links to prototypes, test code and codegen for constants here: https://github.com/open-telemetry/oteps/pull/179

jamesmoessis:
@jsuereth thanks for the explanations.

I've changed the abstract classes here to pure interfaces like you suggested. Essentially, a bunch of getters that map directly to the OTLP definitions. 

I assume implementations of this class would be created in the aggregators package? I've left bucket mapping/indexing out because I believe these details are slightly more contentious and could be figured out in the implementations.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/ExponentialHistogramPointData.java
```diff
@@ -0,0 +1,81 @@
...
+
+  /**
+   * Returns the base, which is calculated via the scale {@link #getScale()}. The larger the base,
+   * the further away bucket boundaries are from each other.
+   *
+   * <p><code>base = 2^(2^-scale)</code>
+   *
+   * @return the base.
+   */
+  double getBase();
```
jsuereth, line 57: Should this have a default implementation?  Is it only for convenience (as scale already exists)

jsuereth (approved):

jsuereth:
> Are you envisioning the implementations of these interfaces to be a mutable representation, to avoid copying?

I'd like to leave that possibility open to use, yeah.   For initial version feel free to copy-data and just get it working (assuming @jkwatson @anuraaga are on board with that).   Using interfaces, hopefully keeps your options open for how to optimise later.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/ExponentialHistogramPointData.java
jamesmoessis: Yeah, the base is only there for convenience, it isn't in the OTLP. I've just pushed a default method for it 👍

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/ExponentialHistogramBuckets.java
```diff
@@ -0,0 +1,49 @@
...
+   */
+  int getOffset();
+
+  /**
+   * The bucket counts is a of counts representing number of measurements that fall into each
+   * bucket.
+   *
+   * @return the bucket counts.
+   */
+  List<Long> getBucketCounts();
```
jkwatson, line 41: Do we think we need `Long` level numbers for the bucket counts? That's a *lot* of recordings for a single bucket.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/ExponentialHistogramPointData.java
jkwatson: Did you forget to push? I don't see it here yet.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/ExponentialHistogramBuckets.java
jmacd: In the NrSketch implementation there is a [backing array that uses variable-width counts](https://github.com/newrelic-experimental/newrelic-sketch-java/blob/main/src/main/java/com/newrelic/nrsketch/WindowedCounterArray.java).

See also OTel-Go: https://github.com/open-telemetry/opentelemetry-go/pull/2261

jsuereth: Looking at the details here, I think exposing `List<Long>` may be... overkill for Java.    (@anuraaga will prove me wrong with some fun optimisation around `List`), but from every example @jmacd lists, I think we want to have the flexibilty to use a primitive array and keep the primitives.    That implies exposing a `long getBucketCountAt(int idx)` vs. an actual list (which is what the Go SDK + NR impl are doing).

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/ExponentialHistogramPointData.java
jamesmoessis: No, it's pushed, see here: https://github.com/open-telemetry/opentelemetry-java/pull/3637/files#diff-c3c2d267864d3dea2c06ce35c0a4ce4ade2834d1e4c0898742ada1d08f6eb221R57

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/ExponentialHistogramBuckets.java
jamesmoessis: It would make sense to use something similar to the `WindowedCounterArray` backed by the `MultiTypeCounterArray` from the NrSketch that @jmacd mentioned.

I can replace `List<Long> getBucketCounts()` with `long getBucketCountAt(int idx)`. Though, it does raise the question in my mind of how the entire collection of counts would be gathered easily. I can't see that in the Go SDK. Am I incorrectly assuming that it's a necessary method to have for aggregation/exporting?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/ExponentialHistogramPointData.java
```diff
@@ -0,0 +1,83 @@
...
+
+  /**
+   * Returns the base, which is calculated via the scale {@link #getScale()}. The larger the base,
+   * the further away bucket boundaries are from each other.
+   *
+   * <p><code>base = 2^(2^-scale)</code>
+   *
+   * @return the base.
+   */
+  default double getBase() {
```
anuraaga, line 57: I don't see `base` in the OTLP proto, I suspect we don't need this as a public method on PointData (presumably we'll have a private method somewhere to compute this where needed)

anuraaga (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/ExponentialHistogramBuckets.java
anuraaga: By the way, I expect marshaling to look like

```java
int size = 0;
for (int i = 0; i < getNumBuckets(); i++) {
  size += CodedOutputStream.computeUint32SizeNoTag(getBucketCountAt(i));
}
writeTag(fieldNumber, size)
for (int i = 0; i < getNumBuckets(); i++) {
  writeUint32NoTag(getBucketCountAt(i);
}
```

anuraaga: By the way, just curious if any consideration was given to a sparse array given this comment in the proto

```
 This field is expected to have many buckets,
    // especially zeros, so uint64 has been selected to ensure
    // varint encoding.
```

```
repeated fixed32 bucket_indexes;
repeated fixed64 bucket_counts;
```

No need to write out zeros

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/ExponentialHistogramBuckets.java
```diff
@@ -0,0 +1,52 @@
...
+   * The bucket count is the number of measurements that fall into a bucket corresponding to the
+   * given index. Index can be negative, in which case the bucket would be lower bounded somewhere
+   * between 0 and 1 (or 0 and -1 if these buckets represent negative measurements).
+   *
+   * <p>Bucket boundaries are inclusive lower bound and exclusive upper bound.
+   *
+   * @param index signed int corresponding to the relevant bucket.
+   * @return the number of measurements in the bucket.
+   */
+  long getBucketCountAt(int index);
```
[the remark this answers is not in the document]
jamesmoessis: Probably needs to be the `upper_limit_index - lower_limit_index` since the index can be negative and doesn't always start from 0. It might be easier to implement Iterable, I would like to hear your thoughts on that.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/ExponentialHistogramPointData.java
jamesmoessis: It's just for convenience / visibility. It doesn't necessarily need to be there. I can remove it if you feel strongly.

jmacd: I'm not sure that `base` is practically useful. In the logarithm implementation, there is a multiplicative factor `log2(E) * 2**scale` we find especially useful in order to use the builtin natural logarithm function.

jamesmoessis: Alright, seems that folks would prefer to go without it, so I've removed it

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/ExponentialHistogramBuckets.java
jamesmoessis: @anuraaga @jsuereth @jkwatson I've made a few changes: 

I've removed the `List<Long>` and replaced it with these three methods with the aim of being flexible for implementations.
  - `long getBucketCountAt(int index)`
  - `int getNumBuckets()`
  - `int getStartIndex()`
 
@anuraaga so your for loop would look like 
```
for (int i = getStartIndex(); i < getNumBuckets(); i++)
```

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/ExponentialHistogramBuckets.java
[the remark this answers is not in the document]
jamesmoessis: As mentioned above, I've added `getNumBuckets()` and `getStartIndex()` to solve for this.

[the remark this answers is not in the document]
anuraaga: Thanks for iterating on this - I'm sorry for the back and forth, but this complexity of the indexes makes me prefer the iterable a lot then. Can we go back to `List<Long>`? Sorry about that.

We'll just cast to an internal hypothetical `LongList` that stores and allows iterating on `long[]` in the exporters.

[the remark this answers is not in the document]
jamesmoessis: `LongList` is neat, that sounds good to me. I'll make the changes tomorrow.

jamesmoessis:
@jkwatson @anuraaga @jsuereth I've pushed the change to move back to `List<Long> getBucketCounts()` as discussed. As mentioned, this can still be optimised in the implementation.

If there's no other points of discussion, I think this PR is in its final form.

anuraaga (approved):
Thanks @jamesmoessis - let me go ahead and merge this as if anything comes up we can easily follow up

PR #3639, opened 2021-09-21
Migrate Zipkin integration test from junit4 to junit5 + testcontainers

anuraaga:
This is the last junit4 test in the repo aside from our explicit junit4 rule test in sdk-testing

### DIFF
--- a/exporters/zipkin/src/test/java/io/opentelemetry/exporter/zipkin/ZipkinSpanExporterEndToEndHttpTest.java
+++ b/exporters/zipkin/src/test/java/io/opentelemetry/exporter/zipkin/ZipkinSpanExporterEndToEndHttpTest.java
@@ -7,6 +7,9 @@ package io.opentelemetry.exporter.zipkin;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import com.linecorp.armeria.client.WebClient;
+import com.linecorp.armeria.common.AggregatedHttpResponse;
+import com.linecorp.armeria.common.HttpStatus;
 import io.opentelemetry.api.common.Attributes;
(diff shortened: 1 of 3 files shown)


exporters/zipkin/src/test/java/io/opentelemetry/exporter/zipkin/ZipkinSpanExporterEndToEndHttpTest.java
```diff
@@ -170,4 +177,17 @@ private static Span buildZipkinSpan(InetAddress localAddress) {
         .putTag(ZipkinSpanExporter.OTEL_STATUS_CODE, "OK")
         .build();
   }
+
+  private static List<Span> getTrace(String traceId) {
+    AggregatedHttpResponse response =
+        client.get(zipkinUrl("/api/v2/trace/" + traceId)).aggregate().join();
+    if (response.status().equals(HttpStatus.NOT_FOUND)) {
```
Oberon00, line 184: Using *just* the HTTP status for this seems dangerous as you would also get 404 if you had a typo like `api/v2/trac/`.

More importantly, shouldn't an exception be thrown here for other non-2xx responses?

anuraaga: I think this is reasonable enough for the method since it's also used from all the other tests, which would fail for such a big issue.

Could add an exception for other error cases but I think it's unexpected enough to keep things simple (instead of explicit exception it'll be json parsing exception).

jkwatson (approved):

PR #3640, opened 2021-09-21
Switch mirroring job to repo-scoped GHCR images since they work now.

anuraaga:


.github/workflows/docker-test-containers.yml
```diff
@@ -11,13 +11,13 @@ jobs:
       matrix:
         include:
           - source: jaegertracing/all-in-one:1.17
-            target_tag: jaeger
-          - source: otel/opentelemetry-collector-contrib-dev:latest
-            target_tag: otel-collector-dev
+            target_image: jaeger
+          - source: otel/opentelemetry-collector-contrib:latest
```
anuraaga, line 15: @jkwatson I also switched from -dev to normal as we discussed before not targeting it for our build I think

PR #3645, opened 2021-09-22
Remove dependency on okhttp-tls

anuraaga:
okhttp-tls has a dependency on the fairly large bouncycastle for certificate generation. We don't actually need it's generation so the thought of excluding bouncycastle came to mind - but I realized we already had all the relevant code here anyways.


jkwatson (approved):
Good cleanup!

trask:
thanks 🎉

PR #3646, opened 2021-09-22
Use concatenation instead of stringbuilder in RERS.

anuraaga:
Randomly found this and let IntelliJ automatically convert this - we generally prefer concatenation since basically same if not better readability and if we ever get to compile against Java 9+ we'll get free performance.


jkwatson (approved):
Sure!

PR #3647, opened 2021-09-22
Fix some nullness.

anuraaga:


sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/WorkerExporter.java
```diff
@@ -111,7 +111,10 @@ public void flush(Collection<SpanData> batch, AbstractQueue<ReadableSpan> queue)
       }
     }
     exportCurrentBatch(batch);
-    flushSignal.get().succeed();
-    flushSignal.set(null);
+    CompletableResultCode flushResult = flushSignal.get();
```
Oberon00, line 114: Is this actually possible?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/exemplar/AbstractFixedSizeExemplarReservoir.java
```diff
@@ -93,11 +93,10 @@ public final void offerMeasurement(double value, Attributes attributes, Context
    * <p>Allocations are acceptable in the {@link #getAndReset(Attributes)} method.
    */
   private class ReservoirCell {
-    private boolean hasValue = false;
```
Oberon00, line 96: Nice!

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/context/SettableContextStorageProvider.java
```diff
@@ -41,6 +42,7 @@ public Scope attach(Context toAttach) {
     }
 
     @Override
+    @Nullable
```
Oberon00, line 45: Shouldn't this result in an API change being listed by the API compare task?

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/WorkerExporter.java
anuraaga: Doubt it - I could use a requireNonNull instead to signal that but I guess this is still a bit more robust to refactoring.

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/context/SettableContextStorageProvider.java
anuraaga: Believe I ran the build - we've seen the diffs not detect all changes in annotations.

sdk-extensions/tracing-incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/WorkerExporter.java
Oberon00: Please add a comment then that we do not expect it to be null. Otherwise readers of the code may be left wondering. (also a warning log might make sense, but I guess that's a difficult topic)

jkwatson (approved):
so many nulls

PR #3648, opened 2021-09-22
Make LogRecord required parameters final.

anuraaga:
I think this is our convention - it solves a nullness issue which is why I found it.


jkwatson (approved):
Seems good to me.

Issue #1768, opened 2020-10-07
Consider compile-time checking of @Nullable and @Nonnull


Oberon00:
We currently have some nullability annotations in the code, but don't really check them. Errorprone only provides some very basic checks (like ReturnMissingNullable) out of the box. We may consider using additional tooling to check this more strictly, e.g. https://github.com/uber/NullAway might be usable.

anuraaga:
Yup I definitely think we should add this. Recently added it to java-instrumentation build

https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1203

breedx-splk:
Since #3668 was merged I think we should close this issue. 
What do you think @Oberon00 and @anuraaga ?

anuraaga:
Thanks - will close with 3845

PR #3655, opened 2021-09-24
Fix more nullness

anuraaga:
For #1768


exporters/jaeger/src/main/java/io/opentelemetry/exporter/jaeger/MarshalerCollectorServiceGrpc.java
```diff
@@ -42,19 +42,17 @@ public PostSpansResponse parse(InputStream stream) {
         }
       };
 
-  private static volatile MethodDescriptor<PostSpansRequestMarshaler, PostSpansResponse>
-      getPostSpansMethod;
+  private static final MethodDescriptor<PostSpansRequestMarshaler, PostSpansResponse>
```
anuraaga, line 45: The lazy technique I had copied from gRPC generated code was presumably for reasons related to having lots of protobufs in a single app. We don't have such an issue to require laziness I guess

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/LogAdapter.java
```diff
@@ -96,6 +94,14 @@ private static InstrumentationLibraryLogs buildInstrumentationLibraryLogs(
             .setSpanId(
                 UnsafeByteOperations.unsafeWrap(
                     OtelEncodingUtils.bytesFromBase16(logRecord.getSpanId(), SpanId.getLength())));
+    String severityText = logRecord.getSeverityText();
```
anuraaga, line 97: Actual potential NPE

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/PrometheusCollectorBuilder.java
```diff
@@ -8,10 +8,11 @@
 import io.opentelemetry.sdk.metrics.export.MetricProducer;
 import io.prometheus.client.Collector;
 import java.util.Objects;
+import javax.annotation.Nullable;
 
 /** Builder for {@link PrometheusCollector}. */
 public class PrometheusCollectorBuilder {
-  private MetricProducer metricProducer;
+  @Nullable private MetricProducer metricProducer;
```
anuraaga, line 15: I would normally change to having the builder constructor accept this, but I know this whole class is under the scalpal in #3578

sdk-extensions/async-processor/src/main/java/io/opentelemetry/sdk/extension/trace/export/DisruptorEventQueue.java
```diff
@@ -205,11 +192,12 @@ public void onEvent(final DisruptorEvent event, long sequence, boolean endOfBatc
           case ON_START:
             @SuppressWarnings("unchecked")
             final SimpleImmutableEntry<ReadWriteSpan, Context> eventArgs =
-                (SimpleImmutableEntry<ReadWriteSpan, Context>) readableSpan;
+                (SimpleImmutableEntry<ReadWriteSpan, Context>)
```
anuraaga, line 195: These requireNonNull are safe based on the code usage. Presumably we're supposed to have actually different event types for the different arguments. My suggestion is to delete this artifact instead if it seems unused :P

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSampler.java
```diff
@@ -45,14 +46,16 @@
...
-      String serviceName,
+      @Nullable String serviceName,
       ManagedChannel channel,
       int pollingIntervalMs,
       Sampler initialSampler,
       boolean closeChannel) {
     this.channel = channel;
     this.closeChannel = closeChannel;
-    this.serviceName = serviceName;
+    // TODO(anuraaga): This probably needs to be replaced with using the Resource, but the spec
```
anuraaga, line 56: snark

sdk-extensions/logging/src/main/java/io/opentelemetry/sdk/logging/export/BatchLogProcessor.java
```diff
@@ -209,8 +209,16 @@ public void run() {
 
     private CompletableResultCode forceFlush() {
       CompletableResultCode flushResult = new CompletableResultCode();
-      this.flushRequested.compareAndSet(null, flushResult);
-      return this.flushRequested.get();
+      if (this.flushRequested.compareAndSet(null, flushResult)) {
+        return flushResult;
+      }
+      flushResult = this.flushRequested.get();
```
anuraaga, line 215: Appears to have been a real, though highly improbable, race, caught by the nullness checker

anuraaga: Actually realized it's reasonable probableness - a pending flush from before this was invoked could rather easily complete between the two lines of code I guess.

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/PrometheusCollectorBuilder.java
jsuereth: Speaking of which, would you be ok removing conflicting parts from this PR and moving into that PR?

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/LogAdapter.java
jkwatson: proto classes will NPE if passed a null? 😭

Oberon00:
I'm wondering: You make quite a few null-ness PR but they are not linked to any issue or otherwise grouped together. Is there some common theme behind them? Do we plan to enforce this null-ness with tooling? I think these PRs should all mention a common issue if they have common cause.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/LogAdapter.java
Oberon00: Proto3 never returns any nulls from getters for primitives (even for strings that are never set, you get empty strings), so it makes sense that it would also reject setting them. I have to admit that I was bitten by that too 😄

sdk-extensions/async-processor/src/main/java/io/opentelemetry/sdk/extension/trace/export/DisruptorEventQueue.java
```diff
@@ -205,11 +192,12 @@ public void onEvent(final DisruptorEvent event, long sequence, boolean endOfBatc
...
+                    requireNonNull(readableSpan, "readableSpan");
```
jkwatson, line 196: feels a bit scary to end up throwing exceptions here. What happens if this method throws?

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSampler.java
jkwatson: why not? I don't think anyone objects to samplers being constructed with the Resource.

anuraaga: I object to it for reasons I've mentioned several times I think :P I've given up on the train of thought since the UX issues don't seem to resonate with anyone it seems

sdk-extensions/async-processor/src/main/java/io/opentelemetry/sdk/extension/trace/export/DisruptorEventQueue.java
anuraaga: Went with null check pattern. Noticed that BSP will throw either way calling `span.getSpanContext`, though yeah probably something to fix there

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/PrometheusCollectorBuilder.java
anuraaga: I'll merge main into your PR if this gets merged first, it will be helpful to get the artifacts working with nullaway if possible

anuraaga (commented):
@Oberon00 It's #1768 which I probably should have been linking too. After the remaining nullness issues are worked out, I'll enable it in the build

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSampler.java
anuraaga: Anyways snarked the snark

jsuereth (approved):

kubawach (approved):
Looks good :)

jkwatson (approved):

PR #3656, opened 2021-09-24
Fix nullness in trace-propagators

anuraaga:


api/all/src/main/java/io/opentelemetry/api/internal/Contract.java
```diff
@@ -89,9 +89,9 @@
  * verifying nullness. We copy the annotation to avoid an external dependency.
  */
 @Documented
-@Retention(RetentionPolicy.SOURCE)
+@Retention(RetentionPolicy.CLASS)
```
anuraaga, line 92: @jkwatson It turns out this needs to be CLASS so that other artifacts that use an annotated method (so usages of `StringUtils.isNullOrEmpty` outside of the api/all project) can be verified. I know you had a concern of us accidentally adding this to a non-internal API, so for now I've just made it package private since we only have one usage (well technically I wanted to use it in one other place but the workaround is fine)

extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/Common.java
```diff
@@ -31,7 +31,11 @@
 
   private Common() {}
 
-  static SpanContext buildSpanContext(String traceId, String spanId, String sampled) {
+  static SpanContext buildSpanContext(
+      @Nullable String traceId, @Nullable String spanId, @Nullable String sampled) {
+    if (traceId == null || spanId == null) {
```
anuraaga, line 36: I wouldn't add these if I added `Contract` to `isTraceIdValid` and `isSpanIdValid`. As far as I know there isn't any performance penalty for this sort of null check and it seems OK to just do this instead.

api/all/src/main/java/io/opentelemetry/api/internal/Contract.java
jkwatson: ok

jkwatson (approved):

PR #3658, opened 2021-09-24
Pass a cause to ConfigurationException somewhere.

anuraaga:
It's no big deal but it's a bit annoying to see the opentelemetry-spi package in codecov have 50% coverage as it just looks like a project with low coverage even though it isn't really. The simple fix is to pass and then ignore this NumberFormatException as a cause (the other location we were calling the constructor that has a cause is an invalid cert exception that is tedious to verify)


jkwatson (approved):
🤷🏽

PR #3664, opened 2021-09-27
Don't crash on null spans in BSP

anuraaga:
Noticed this in #3655 - since this is on the hot path in the user's app it can't have a chance of crashing

### DIFF
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
@@ -91,7 +91,7 @@ public final class BatchSpanProcessor implements SpanProcessor {
 
   @Override
   public void onEnd(ReadableSpan span) {
-    if (!span.getSpanContext().isSampled()) {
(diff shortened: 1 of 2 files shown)


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -91,7 +91,7 @@ public boolean isStartRequired() {
 
   @Override
   public void onEnd(ReadableSpan span) {
```
Oberon00, line 93: Since this callback is only called by SDK code, it should not be possible for null to occur here.

anuraaga: It's a public API method, so we can easily conceive a wrapping `SpanProcessor` that delegates to it that may pass in nulls problematically. While we could say that such a processor has a problem, if the NPE is in our code than users will blame our code for crashing their app.

Oberon00 (approved):

jkwatson (approved):

PR #3665, opened 2021-09-27
Build with Java 17

anuraaga:
Doesn't change the target compatibility to 17.

Does still build fine if Gradle is run with Java 11. Now it can also be run with Java 17 (with the gradle.properties workaround)

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -18,7 +18,6 @@ jobs:
         test-java-version:
           - 8
           - 11
-          - 16
           - 17
         include:
           - os: ubuntu-20.04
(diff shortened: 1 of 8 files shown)


buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -25,7 +25,7 @@ base {
 
 java {
   toolchain {
-    languageVersion.set(JavaLanguageVersion.of(11))
+    languageVersion.set(JavaLanguageVersion.of(17))
```
Oberon00, line 28: This would require a README/CONTRIBUTING.md update since it adds a new requirement for the build-environment. Or would gradle auto-download Java 17?

anuraaga: Gradle auto downloads indeed

anuraaga: So i.e., it's still just min Java 11 to run the build

.github/workflows/pr-build.yml
```diff
@@ -18,7 +18,6 @@ jobs:
         test-java-version:
           - 8
           - 11
-          - 16
```
jkwatson, line 21: 👍🏽

jkwatson (approved):

PR #3666, opened 2021-09-27
Update autoconfigure to append signal path to otlp http endpoint if n…

### LINKED ISSUE #3650: Autoconfigure SHOULD append /v1/(traces|metrics) to OTEL_EXPORTER_OTLP_ENDPOINT
See the description of "Endpoint (OTLP/HTTP)" at https://github.com/open-telemetry/opentelemetry-specification/blob/v1.6.1/specification/protocol/exporter.md#configuration-options:

> When using `OTEL_EXPORTER_OTLP_ENDPOINT`, exporters SHOULD follow the collector convention of appending the version and signal to the path (e.g. `v1/traces` or `v1/metrics`), if not present already.

Note this does not apply to the signal-specific envvars.

This is actually an important SHOULD, because otherwise OTEL_EXPORTER_OTLP_ENDPOINT cannot sensibly be used because metrics and traces would be sent to the same URL.

Not sure about the "if not present already" part though, that seems like a spec defect, because configuring a URL ending with /v1/traces as endpoint for all signals would mean that metrics are also sent to /v1/traces, which makes no sense at all. Will file a spec issue.

EDIT: I should mention that I was made aware of this issue by @rypdal. Thank you!

jack-berg:
…ot present

Resolves #3650.


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtil.java
```diff
@@ -46,6 +48,16 @@ static void configureOtlpExporterBuilder(
...
     if (endpoint == null) {
       endpoint = config.getString("otel.exporter.otlp.endpoint");
+      String signalPath = signalPath(dataType);
+      String protocol = getOtlpProtocol(dataType, config);
+      if (endpoint != null
+          && protocol.equals(PROTOCOL_HTTP_PROTOBUF)
+          && !endpoint.endsWith(signalPath)) {
+        if (!endpoint.endsWith("/")) {
+          endpoint += "/";
+        }
```
jack-berg, line 58: I took a little liberty from the spec here by appending a `/` before the signal path if it doesn't exist. 

A strict interpretation would cause `http://localhost:4317/foo` to be transformed to `http://localhost:4317/foov1/traces`

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtil.java
```diff
@@ -46,6 +48,16 @@ static void configureOtlpExporterBuilder(
     String endpoint = config.getString("otel.exporter.otlp." + dataType + ".endpoint");
     if (endpoint == null) {
       endpoint = config.getString("otel.exporter.otlp.endpoint");
+      String signalPath = signalPath(dataType);
+      String protocol = getOtlpProtocol(dataType, config);
+      if (endpoint != null
+          && protocol.equals(PROTOCOL_HTTP_PROTOBUF)
+          && !endpoint.endsWith(signalPath)) {
```
jkwatson, line 55: This seems like it precludes someone pointing at a totally different endpoint of their own choosing. I'm not sure that's a good thing, since perhaps their pointing at a proxy or some such that is doing the mapping to the actual path?

If we want to force the signal path like this, I would recommend that we use URI.resolve(), rather than hand-crafting the URL to use. If we use that method, it doesn't matter what they've put on the end of their host/port...we'll use the one we know is right.

But, again, this seems dangerous and that it would break anyone who is doing something not 100% standard with their ingest.

Perhaps the best thing would be to construct a URI, then see if it has a path component, and leave it alone if it does, otherwise resolve it to the required signal-specific path.

jkwatson: Hm. ok, I see how that probably makes sense since it probably wasn't really usable as-is, anyway. I think the URI.resolve() implementation is a little cleaner, since we don't have to worry about the trailing slashes, etc. Unless, of course, we think that people may want to pre-pend some sort of pre-path before the actual signal path? That's probably a very narrow edge case, though, and I'm not sure those people would have anything usable right now anyway.

jack-berg: Pushed up a commit using `URI.resolve()` instead of string concatenation. Personally, I prefer the string concatenation approach since URI required handling a checked exception and the behavior of `.resolve(..)` [isn't super intuitive](https://stackoverflow.com/a/2535032/14621080). Don't feel strongly about this.

jkwatson: I don't feel super strongly about it, either. Checking for validity seems like a positive thing from URI, though, doesn't it? And, yeah, `resolve()` is a bit weird around slash-handling (which I honestly didn't remember to be the case when I suggested it).

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtilTest.java
```diff
@@ -58,4 +62,39 @@ void getOtlpProtocolDefault() {
                         "otel.exporter.otlp.traces.protocol", "qux"))))
         .isEqualTo("qux");
   }
+
+  @Test
+  void configureOtlpExporterBuilder_HttpProtobufEndpoint() {
+    assertThat(configureHttpProtobufEndpoint(DATA_TYPE_TRACES, "http://localhost:4317"))
+        .isEqualTo("http://localhost:4317/v1/traces");
+    assertThat(configureHttpProtobufEndpoint(DATA_TYPE_TRACES, "http://localhost:4317/"))
+        .isEqualTo("http://localhost:4317/v1/traces");
```
anuraaga, line 71: Can you add a test for `http://localhost:4317/collector/v1/traces` -> `http://localhost:4317/collector/v1/traces` and `http://localhost:4317/collector` -> `http://localhost:4317/collector/v1/traces` ?

Oberon00: This is asking for more than the spec specifies. Configuring a per-signal endpoint should not append anything, so `http://localhost:4317/collector` should stay as just that. As worded in the spec, the "/v1/traces" suffix is merely convention for the collector. If I have a backend that does not follow this convention, I should be able to override the per-signal endpoint with the per-signal configuration, without having anything changed in the URL.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtil.java
Oberon00: I think that's OK. If someone needs a path without slash (or something entirely different), they can configure it with the per-signal configuration.
Still, it should probably also be fixed in the spec.

Oberon00 (commented):
In the PR description and docs, you say you will append the per-signal path "if not present already". Do you mean "if the per-signal config is not already set"? If so, please clarify the wording.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtil.java
```diff
@@ -46,6 +50,17 @@ static void configureOtlpExporterBuilder(
...
+      String protocol = getOtlpProtocol(dataType, config);
+      if (endpoint != null && protocol.equals(PROTOCOL_HTTP_PROTOBUF)) {
+        URI endpointUri;
+        try {
+          endpointUri = new URI(endpoint);
+        } catch (URISyntaxException e) {
+          throw new ConfigurationException("Invalid endpoint, must be a URI: " + endpoint, e);
+        }
+        String signalPath = signalPath(dataType);
+        endpoint = endpointUri.resolve(signalPath).toString();
```
Oberon00, line 62: Won't `resolve` with a leading slash in the argument strip of the entire path part of the URL, leaving only the hostname? The tests don't cover that. I suspect `URI.create("http://localhost:4317/collector").resolve("/v1/traces")` will result in http://localhost:4317/v1/traces`, *not* the probably expected `http://localhost:4317/collector/v1/traces`.

anuraaga: Yeah why I asked for the test :-)

> I think that would violate the spec. Configuring a per-signal endpoint should not append anything, so http://localhost:4317/collector should stay as just that.

I agree with this and think the spec allows for it. I guess it means we get something like this

`http://localhost:4317` -> `http://localhost:4317/v1/traces`
`http://localhost:4317/` -> `http://localhost:4317/` (otherwise, there'd be no way to host at a root like `http://otlp-http.com/`, not appending if a root path is passed explicitly seems like the only way to support it)
`http://localhost:4317/collector` -> `http://localhost:4317/collector`

Oberon00: No, the difference is not the slash but whether you set the shared configuration for all signals (auto-appending) or the per-signal config (URL is used as-is).

Oberon00: Hmm, it *is* relatively clear on that part:

> When using OTEL_EXPORTER_OTLP_ENDPOINT, exporters SHOULD follow the collector convention of appending the version and signal to the path

Note the
> When using OTEL_EXPORTER_OTLP_ENDPOINT

This exludes the OTEL_EXPORTER_OTLP_signal_ENDPOINT vars

https://github.com/open-telemetry/opentelemetry-specification/blob/v1.6.1/specification/protocol/exporter.md#configuration-options

anuraaga: Ah ok that sounds pretty good then

jack-berg:
@Oberon00 
> In the PR description and docs, you say you will append the per-signal path "if not present already". Do you mean "if the per-signal config is not already set"? If so, please clarify the wording.

By "if not present already" I'm referring to the current wording of the spec which says that the signal specific path (i.e. `v1/traces`) is appended if `otel.exporter.otlp.endpoint` is specified and doesn't already end with `v1/traces`. The signal specific configuration (i.e. `otel.exporter.otlp.traces.endpoint`) still takes precedence over `otel.exporter.otlp.endpoint`.

I see you have a PR that changes the spec wording to say that the signal specific path MUST always be appended. That would change the behavior as follows:
```
current: otel.exporter.otlp.endpoint=http://localhost:4317/v1/traces => http://localhost:4317/v1/traces
as proposed: otel.exporter.otlp.endpoint=http://localhost:4317/v1/traces => http://localhost:4317/v1/traces/v1/traces
``` 
I think the java SDK follow the behavior as currently worded until your spec PR is merged. I'm happy to update the behavior when that occurs.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtil.java
jack-berg: I had this test in my initial commit, which I've reverted to (along with using string concatenation instead of URI.resolve().

The weakness of string concatenation is that the behavior will be wrong if there is a query string or a fragment included. The spec appears to disallow query strings / fragments with this line:

> The endpoint MUST be a valid URL with scheme (http or https) and host, and MAY contain a port and path.

I.e. by not explicitly allowing a query string / fragment the assumption is that they're not allowed. The best logic would:
1. parse the input as a `URL`
2. confirm there aren't any parts included that aren't explicitly allowed
3. get the path from the URL, and manipulate it to include include the signal type path (accounting for trailing slashes)
4. rebuild a new url from the parts, using the newly constructed path, and convert it to a string

I can do this. Let me know if ya'll are interested or if you think its overkill.

jkwatson: That seems like a good plan. And, having unit tests to cover all the possible strange cases will be very useful documentation, at least internally.

Oberon00:
What will the metric export URL be with the current behavior if I specify `otel.exporter.otlp.endpoint=http://localhost:4317/v1/traces` ? I would assume that metrics are then exported to `http://localhost:4317/v1/traces` which would make no sense, because if you only cared about traces you should just set the per-signal endpoint.

jack-berg:
> What will the metric export URL be with the current behavior if I specify otel.exporter.otlp.endpoint=http://localhost:4317/v1/traces ? I would assume that metrics are then exported to http://localhost:4317/v1/traces which would make no sense, because if you only cared about traces you should just set the per-signal endpoint.

This is the behavior of the `main` branch today. 

> If we interpret the spec very literally, it could also be http://localhost:4317/v1/traces/v1/metrics (which it would be with the new behavior, making it perhaps more clear that the initial part should be omitted).

This is the behavior of the PR as it currently stands.

jkwatson:
> > What will the metric export URL be with the current behavior if I specify otel.exporter.otlp.endpoint=http://localhost:4317/v1/traces ? I would assume that metrics are then exported to http://localhost:4317/v1/traces which would make no sense, because if you only cared about traces you should just set the per-signal endpoint.
> 
> This is the behavior of the `main` branch today.
> 
> > If we interpret the spec very literally, it could also be http://localhost:4317/v1/traces/v1/metrics (which it would be with the new behavior, making it perhaps more clear that the initial part should be omitted).
> 
> This is the behavior of the PR as it currently stands.

It sounds like we definitely need some spec clarification on this. :)

jack-berg:
The commit I just pushed up now adds URL validation for endpoints that is more strict, but also more in inline with the spec, as I suggested I could do in [this comment](https://github.com/open-telemetry/opentelemetry-java/pull/3666#discussion_r717933212). 

It also has test cases that show:
- A variety of valid endpoints
- A variety of invalid endpoints
- How the signal path is appended to the endpoint if the protocol is `http/protobuf` in a variety of scenarios

I suggest that we don't add any more scope to this PR.

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtilTest.java
```diff
@@ -58,4 +67,121 @@ void getOtlpProtocolDefault() {
...
+    assertThat(configureHttpProtobufEndpoint(DATA_TYPE_TRACES, "http://localhost:4317/foo"))
+        .isEqualTo("http://localhost:4317/foo/v1/traces");
+    assertThat(configureHttpProtobufEndpoint(DATA_TYPE_TRACES, "http://localhost:4317/foo/"))
+        .isEqualTo("http://localhost:4317/foo/v1/traces");
+    assertThat(configureHttpProtobufEndpoint(DATA_TYPE_TRACES, "http://localhost:4317/v1/traces"))
+        .isEqualTo("http://localhost:4317/v1/traces");
+    assertThat(configureHttpProtobufEndpoint(DATA_TYPE_TRACES, "http://localhost:4317/v1/metrics"))
```
anuraaga, line 153: Thanks - this is a descriptive and good case to have here

anuraaga (approved):

jack-berg:
Now that @Oberon00 's [spec PR](https://github.com/open-telemetry/opentelemetry-specification/pull/1975) has been merged, I've updated this PR to always append the signal path when `otel.exporter.otlp.endpoint` is specified. This is reflected in the test cases as well.

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtilTest.java
```diff
@@ -58,4 +67,121 @@ void getOtlpProtocolDefault() {
...
+    assertThat(configureHttpProtobufEndpoint(DATA_TYPE_TRACES, "http://localhost:4317/"))
+        .isEqualTo("http://localhost:4317/v1/traces");
+    assertThat(configureHttpProtobufEndpoint(DATA_TYPE_TRACES, "http://localhost:4317/foo"))
+        .isEqualTo("http://localhost:4317/foo/v1/traces");
+    assertThat(configureHttpProtobufEndpoint(DATA_TYPE_TRACES, "http://localhost:4317/foo/"))
+        .isEqualTo("http://localhost:4317/foo/v1/traces");
+    assertThat(configureHttpProtobufEndpoint(DATA_TYPE_TRACES, "http://localhost:4317/v1/traces"))
+        .isEqualTo("http://localhost:4317/v1/traces/v1/traces");
+    assertThat(configureHttpProtobufEndpoint(DATA_TYPE_TRACES, "http://localhost:4317/v1/metrics"))
+        .isEqualTo("http://localhost:4317/v1/metrics/v1/traces");
```
jkwatson, line 154: this one makes sense to me (I mean...it's almost certainly wrong, but it's what the user requested). However, the one where we add `/v1/traces` when it's already on the end seems surprising to me. Yes, it's precisely what the user is requesting, but I suspect this will happen and users will complain. Are we were we don't want to check to make sure that the generic endpoint hasn't already had the `/v1/<signal>` tacked onto it before duplicating it?

Oberon00: This is exactly what the spec is forbidding. Better the users notice the error earlier than much later when they have deployed that config everywhere and want to start using metrics and it ends up at `/v1/traces/v1/metrics`.
If you suggest that the configuration should be "smart" and detect any signal path (e.g., replacing `/v1/metrics` with `/v1/traces` for traces, leaving it in-place for metrics) I suggest a spec PR. I wouldn't be against it, if we can manage to log some warning at least...

Oberon00: Maybe I misunderstood and you meant the signal-specific thing? That seems like a bug indeed. I'll comment above.

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtilTest.java
```diff
@@ -58,4 +67,121 @@ void getOtlpProtocolDefault() {
...
+    assertThat(configureHttpProtobufEndpoint(DATA_TYPE_TRACES, "http://localhost:4317"))
+        .isEqualTo("http://localhost:4317/v1/traces");
+    assertThat(configureHttpProtobufEndpoint(DATA_TYPE_TRACES, "http://localhost:4317/"))
+        .isEqualTo("http://localhost:4317/v1/traces");
+    assertThat(configureHttpProtobufEndpoint(DATA_TYPE_TRACES, "http://localhost:4317/foo"))
+        .isEqualTo("http://localhost:4317/foo/v1/traces");
+    assertThat(configureHttpProtobufEndpoint(DATA_TYPE_TRACES, "http://localhost:4317/foo/"))
+        .isEqualTo("http://localhost:4317/foo/v1/traces");
+    assertThat(configureHttpProtobufEndpoint(DATA_TYPE_TRACES, "http://localhost:4317/v1/traces"))
+        .isEqualTo("http://localhost:4317/v1/traces/v1/traces");
```
Oberon00, line 152:
```suggestion
    assertThat(configureHttpProtobufEndpoint(DATA_TYPE_TRACES, "http://localhost:4317/v1/traces"))
        .isEqualTo("http://localhost:4317/v1/traces");
```

The spec mandates that signal-specific URLs are left alone and /v1/traces is *not* appended. What if you have an endpoint that needs an URL not ending in that? I don't think the OTLP spec mandates a particular URL path.

Oberon00 (changes requested):

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtilTest.java
Oberon00: > The per-signal endpoint configuration options take precedence and can be used to override this behavior (the URL is used as-is for them, without any modifications). 

and

> 1. For the per-signal variables (OTEL_EXPORTER_OTLP_<signal>_ENDPOINT), the URL MUST be used as-is without any modification. The only exception is that if an URL contains no path part, the root path / MUST be used (see Example 2).

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp (and table above)

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtilTest.java
jkwatson: No, I was suggesting that if the user provides a trace endpoint as the "generic" endpoint, that we don't mess with it for traces, since it's likely(?) that they are only using tracing, and don't care about metrics and what endpoint might have been populated for it.

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtilTest.java
Oberon00: OK, maybe I misunderstand what the test does because the implementation at L56 in `sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtil.java` looks like it's doing the right thing, not appending the signal path for the per-signal config.

jkwatson: yes, it's quite hard to figure out what cases are being tested in these tests, at least from the github diff.

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtilTest.java
Oberon00: That's currently against the spec and I think delaying the error until the point in time where the user uses any other signal is not good. It would be better to ensure that the (404?) error that occurs with the doubled signal path is easy to find (e.g. by logging).
I could be convinced of the "fully smart" method, stripping off any known trailing signal path from the generic URL, but making *only* the configured path work seems to be a bad balance.

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtilTest.java
Oberon00: Maybe a better function name could already help?

jkwatson: yes, maybe `configureHttpProtobufEndpoint` could be renamed to `configureNonSignalSpecificEndpoint` or something to make it super clear in the `OtlpConfigUtilTest`

jack-berg: > OK, maybe I misunderstand what the test does because the implementation at L56 in sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtil.java looks like it's doing the right thing, not appending the signal path for the per-signal config.

Yes these test cases are specifically testing the behavior of the signal agnostic `otel.exporter.otlp.endpoint` property. 

I can try to improve the names of the function(s) to improve clarity.

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtilTest.java
jkwatson: ok. I understand that side of the argument. This is potentially a breaking change for some users, so we should be very sure to call it out very clearly in our release notes/CHANGELOG.  @jack-berg Can you craft a CHANGELOG entry for this PR? Thanks!

jkwatson (approved):
With a CHANGELOG entry explaining the potential breakage, this is 👍🏽

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtilTest.java
```diff
@@ -58,4 +71,219 @@ void getOtlpProtocolDefault() {
...
+        .isInstanceOf(ConfigurationException.class)
+        .hasMessageContaining("OTLP endpoint must not have a path:");
+  }
+
+  private static ThrowingCallable configureEndpointCallable(Map<String, String> properties) {
+    return () -> configureEndpoint(DATA_TYPE_TRACES, properties);
+  }
+
+  @Test
+  void configureOtlpExporterBuilder_HttpGenericEndpointKey() {
```
jack-berg, line 141: I've tried to improve the naming conventions in the tests. There are now 3 tests:
- `configureOtlpExporterBuilder_HttpGenericEndpointKey` validates the behavior when the generic `otel.exporter.otlp.endpoint` is used to configure the http endpoint
- `configureOtlpExporterBuilder_HttpTracesEndpointKey` validates the behavior when the trace specific `otel.exporter.otlp.traces.endpoint` is used to configure the http endpoint (included validating that it overrides the generic endpoint value)
- `configureOtlpExporterBuilder_HttpMetricsEndpointKey` validates the behavior when the metrics specific `otel.exporter.otlp.metrics.endpoint` is used to configure the http endpoint (included validating that it overrides the generic endpoint value)

I think the method names are pretty good. In any case, they're better than repeating a ~7 line `ImmutableMap.of(..)` 20+ times. If its not immediately obvious what the `configureEndpointForHttp` does, take a quick look at it - IMO it greatly improves the readability of the test cases.

jkwatson:
Thanks!

PR #3668, opened 2021-09-28
Enable nullaway except for metrics.

anuraaga:
Will enable for metrics after #3578

### DIFF
--- a/buildSrc/src/main/kotlin/otel.errorprone-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.errorprone-conventions.gradle.kts
@@ -7,8 +7,6 @@ plugins {
   id("net.ltgt.nullaway")
 }
 
-val enableNullaway: String? by project
-
 val disableErrorProne = properties["disableErrorProne"]?.toString()?.toBoolean() ?: false
 
 tasks {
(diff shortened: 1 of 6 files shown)


buildSrc/src/main/kotlin/otel.errorprone-conventions.gradle.kts
```diff
@@ -68,15 +66,21 @@ tasks {
...
+          // Disable nullaway by default, we enable for main sources below.
           severity.set(CheckSeverity.OFF)
         }
-        annotatedPackages.add("io.opentelemetry")
+      }
+    }
+  }
+
+  // Enable nullaway on main sources.
+  named<JavaCompile>("compileJava") {
```
kubawach, line 80: Niiiice

kubawach (approved):
👍

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/SpanConverter.java
```diff
@@ -40,10 +41,9 @@ static SpanKind mapKind(@Nullable io.opencensus.trace.Span.Kind ocKind) {
     return SpanKind.INTERNAL;
   }
 
-  @Nullable
   static Span fromOtelSpan(@Nullable io.opentelemetry.api.trace.Span otSpan) {
     if (otSpan == null) {
-      return null;
+      return BlankSpan.INSTANCE;
```
jkwatson, line 46: this is a behavior change. Do we have a unit test for this?

anuraaga: The oc shim has a lot of complication that makes it difficult for me to figure out some of the test cases - that's why it has very low total coverage already. Hoping someone from oc can give this artifact some attention.

I found this behavior in the OC scope manager so copied it since it seemed like a good way to reduce nullness.

jkwatson (approved):

PR #3670, opened 2021-09-28
Add OTLP logs exporter

anuraaga:
With the migration to Marshaler figured it'd be good to wire this up to make sure logs doesn't get blocked by anything. I basically copied every class from the span exporter and did find replace of Span -> Log, Trace -> Logs, etc until it compiled


exporters/otlp/logs/src/main/java/io/opentelemetry/exporter/otlp/logs/OtlpGrpcLogExporter.java
```diff
@@ -0,0 +1,187 @@
...
+   * Creates a new OTLP gRPC Logs Reporter with the given name, using the given channel.
+   *
+   * @param channel the channel to use when communicating with the OpenTelemetry Collector.
+   * @param timeoutNanos max waiting time for the collector to process each log batch. When set to 0
+   *     or to a negative value, the exporter will wait indefinitely.
+   * @param compressionEnabled whether or not to enable gzip compression.
+   */
+  OtlpGrpcLogExporter(ManagedChannel channel, long timeoutNanos, boolean compressionEnabled) {
+    // TODO: telemetry schema version.
+    Meter meter = GlobalMeterProvider.get().meterBuilder("io.opentelemetry.exporters.otlp").build();
```
jsuereth, line 69: In general, should OTEL itself be using the global provider? I noticed this in the other exporters and was curious, since I thought we were recommending most instrumentation not use the global provider.

exporters/otlp/logs/src/main/java/io/opentelemetry/exporter/otlp/logs/OtlpGrpcLogExporter.java
```diff
@@ -0,0 +1,187 @@
...
+            result.succeed();
+          }
+
+          @Override
+          public void onFailure(Throwable t) {
+            logsExportedFailure.add(logs.size());
+            Status status = Status.fromThrowable(t);
+            switch (status.getCode()) {
+              case UNIMPLEMENTED:
+                logger.log(
```
jsuereth, line 117: IIUC - This won't lead to a log-loop-export because you're using a throttling logger, however this COULD potentially place a new log back into the queue to export.

exporters/otlp/logs/src/main/java/io/opentelemetry/exporter/otlp/logs/OtlpGrpcLogExporter.java
```diff
@@ -0,0 +1,187 @@
...
+
+    MarshalerLogsServiceGrpc.LogsServiceFutureStub exporter;
+    if (timeoutNanos > 0) {
+      exporter = logsService.withDeadlineAfter(timeoutNanos, TimeUnit.NANOSECONDS);
+    } else {
+      exporter = logsService;
+    }
+
+    Futures.addCallback(
+        exporter.export(request),
```
jsuereth, line 103: You should make sure to detach grpc Context here to make sure you aren't accidentally attaching to something else that was on the same thread.

grpc Context in recent version is much more sensitive to re-use.  Hoping to find time later this week to reproduce a test case for auto-instrumentation.

jsuereth (approved):
Looks good to me, just some mostly unrelated to the specifics questions.

kubawach (approved):
LGTM :)

exporters/otlp/logs/src/main/java/io/opentelemetry/exporter/otlp/logs/OtlpGrpcLogExporter.java
jkwatson: We'll use this until metrics is stable and we can allow people to provide their own instance of the `MeterProvider`. This way we can at least hide any API instability, as it keeps this as an `implementation` dependency, rather than `api`.

exporters/otlp/logs/src/main/java/io/opentelemetry/exporter/otlp/logs/OtlpGrpcLogExporter.java
jkwatson: This is a good callout. What should we be doing in here, to avoid log-looping on this path. Maybe System.out is a "good enough" fallback? I don't know of a way to automatically exclude our logger here from being wired up to the exporter.

jkwatson (approved):
Should be good enough for people to start kicking the tires, I think.

exporters/otlp/logs/src/main/java/io/opentelemetry/exporter/otlp/logs/OtlpGrpcLogExporter.java
anuraaga: I'm not sure if it applies here. The common use case is the BatchSpanProcessor so the thread is independent and shouldn't affect the app I think. If a user uses the SingleSpanProcessor for some reason (not possible with the javaagent, and generally not recommended by us), I feel like respecting an existing context is actually correct behavior - if we're exporting spans during a request, then that request's timeout actually should be applied to the export request since we're affecting it.

Let me know if this understanding makes sense.

PR #3677, opened 2021-09-29
Remove unused CodedOutputStream methods.

anuraaga:
The only unused methods remaining are for types that we don't use yet, such as `sfixed32`, though we would want to make it easy to add these fields to the protocol and leave them. There is no possible future use case for these other removed methods, without bigger changes, especially the out of space stuff that doesn't apply to output stream encoding.


kubawach (approved):
Looks good!

PR #3678, opened 2021-09-29
Exclude okhttp dependency from grpc-okhttp tests

anuraaga:
`grpc-okhttp` actually bundles in okhttp - they also have a few API knobs that use real okhttp classes, but we don't actually use these knobs. The tests passing with this change seem to confirm that we don't need to actually bring in the ancient okhttp version into downstream builds.

For reference:
https://github.com/grpc/grpc-java/issues/6119

### DIFF
--- a/exporters/otlp/logs/build.gradle.kts
+++ b/exporters/otlp/logs/build.gradle.kts
@@ -48,7 +48,9 @@ dependencies {
 
   add("testGrpcOkhttpImplementation", "com.linecorp.armeria:armeria-grpc")
   add("testGrpcOkhttpImplementation", "com.linecorp.armeria:armeria-junit5")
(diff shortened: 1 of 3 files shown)


jack-berg:
Are there any downstream builds which have a dependency on a test source set?

Also, is it worth excluding this dependency from the compile only dependencies? I.e. `compileOnly("io.grpc:grpc-okhttp")`

anuraaga:
@jack-berg Yeah I was mostly trying to confirm it could be safe downstream to do the same as what we do here. But I realized that because we have `setChannel`, it's relatively easy for a user to configure the client in a way that would still break. Ideally we didn't have `setChannel` / gRPC exposed into the public API but perhaps in 2.x.

PR #3683, opened 2021-09-30
Deprecate async-processor which was added to contrib.

anuraaga:
This may cause a couple of days of awkwardness where this is deprecated but the contrib artifact hasn't had it's first published version yet. I think it's ok since it should be a small number of days and we're not actually removing the code.


jkwatson (approved):
👍🏽 . If, in the future, it turns out there is a lot of demand, we can consider moving it back here for full support.

PR #3685, opened 2021-09-30
Implements json log exporter

ohadza:


jkwatson:
@ ohadza It looks like you need to sign the CLA before we can move forward with this PR. Thanks!

exporters/logging-otlp/src/main/java/io/opentelemetry/exporter/logging/otlp/OtlpJsonLoggingLogExporter.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * A {@link LogExporter} which writes {@linkplain LogRecord logs} to a {@link Logger} in OTLP JSON
+ * format. Each log line will include a single {@code ResourceLogs}.
+ */
+public final class OtlpJsonLoggingLogExporter implements LogExporter {
+
+  //  private static final MessageMarshaller marshaller =
```
anuraaga, line 27: Looks like there's some drift we can delete

exporters/logging-otlp/src/main/java/io/opentelemetry/exporter/logging/otlp/OtlpJsonLoggingLogExporter.java
```diff
@@ -0,0 +1,67 @@
...
+
+  /** Returns a new {@link OtlpJsonLoggingLogExporter}. */
+  public static LogExporter create() {
+    return new OtlpJsonLoggingLogExporter();
+  }
+
+  private OtlpJsonLoggingLogExporter() {}
+
+  @Override
+  public CompletableResultCode export(Collection<LogRecord> logs) {
```
anuraaga, line 43: We ended up not marking this artifact alpha so there is some danger here. Will we rename the class to LogData to match other SDKs? That would be a breaking so if we don't make a separate artifact we need to lock that down.

exporters/logging-otlp/src/main/java/io/opentelemetry/exporter/logging/otlp/OtlpJsonLoggingLogExporter.java
ohadza: Thanks!

exporters/logging-otlp/src/main/java/io/opentelemetry/exporter/logging/otlp/OtlpJsonLoggingLogExporter.java
jkwatson: Hmm. I wonder if we should hold off a bit on this exporter, then.

jack-berg:
@anuraaga does moving the log classes solve the problem blocking this PR? `:sdk:log` will still have to be alpha, and so this non-alpha logging module will still need to depend on an alpha module.

ohadza:
My branch got mess so I open a new one #3744 
I'm closing this PR

PR #3686, opened 2021-09-30
update the semconv to v1.7.0 and the build tools to 0.7.0

jkwatson:


jsuereth (approved):

semconv/src/main/java/io/opentelemetry/semconv/trace/attributes/SemanticAttributes.java
```diff
@@ -787,6 +811,30 @@
...
+   * one for received message.
+   *
+   * <p>Notes:
+   *
+   * <ul>
+   *   <li>This way we guarantee that the values will be consistent between different
+   *       implementations.
+   * </ul>
+   */
+  public static final AttributeKey<Long> MESSAGE_ID = longKey("message.id");
```
Oberon00, line 828: Interesting that this was not present previously.

jkwatson: Looks like it was added here:  https://github.com/open-telemetry/opentelemetry-specification/pull/1843. By you. :)

Oberon00: Oh, this is for RPC! Didn't realize... Yeah we should definitely change that name 😄 

https://github.com/open-telemetry/opentelemetry-specification/pull/1914#discussion_r708388214

Oberon00: Hmm, the id of the group is `rpc.message`, not just message. I thought `id` instead of `prefix` would be used for the name. But I guess that would lead to ugly identifiers in some cases where we split client/server into different groups. On the other hand, these identifiers may even be helpful.

Please consider that (well) before declaring the semconv artifact stable.

Oberon00: It may make sense to add the group ID to the generated javadoc, otherwise this documentation here is misleading, since the RPC context is completely missing.

jkwatson: Is the group information even available to the templates at the moment? I poked around and tried various things but couldn't sort it out. It's very likely I missed it somewhere, though.

PR #3687, opened 2021-09-30
Add support for none compression keyword

jack-berg:
[Spec](https://github.com/open-telemetry/opentelemetry-specification/pull/1955) now says that `none` must be a supported compression type.

### DIFF
--- a/exporters/otlp-http/metrics/src/test/java/io/opentelemetry/exporter/otlp/http/metrics/OtlpHttpMetricExporterTest.java
+++ b/exporters/otlp-http/metrics/src/test/java/io/opentelemetry/exporter/otlp/http/metrics/OtlpHttpMetricExporterTest.java
@@ -7,6 +7,7 @@ package io.opentelemetry.exporter.otlp.http.metrics;
 
 import static io.opentelemetry.api.common.AttributeKey.stringKey;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import com.google.protobuf.InvalidProtocolBufferException;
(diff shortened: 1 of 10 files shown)


exporters/otlp-http/metrics/src/test/java/io/opentelemetry/exporter/otlp/http/metrics/OtlpHttpMetricExporterTest.java
```diff
@@ -94,6 +96,43 @@ void setup() {
                 HELD_CERTIFICATE.certificatePem().getBytes(StandardCharsets.UTF_8));
   }
 
+  @Test
+  @SuppressWarnings("PreferJavaTimeOverload")
+  void validConfig() {
```
jack-berg, line 101: Did this in another place and I think its useful to show explicitly config we expect to support.

One thing that's strange is that we're too permissive with the values we accept for `setEndpoint`.  We only confirm that the values are valid URIs, and that the scheme is `http` or `https`. The test cases on 111-117 are a small subset of what is accepts. The autoconfigure module now does additional validation, but arguably that validation should be moved to the exporter builders instead.

exporters/otlp-http/trace/src/main/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporterBuilder.java
```diff
@@ -76,15 +76,17 @@ public OtlpHttpSpanExporterBuilder setEndpoint(String endpoint) {
...
     requireNonNull(compressionMethod, "compressionMethod");
     checkArgument(
-        compressionMethod.equals("gzip"),
-        "Unsupported compression method. Supported compression methods include: gzip.");
-    this.compressionEnabled = true;
+        compressionMethod.equals("gzip") || compressionMethod.equals("none"),
+        "Unsupported compression method. Supported compression methods include: gzip, none.");
+    if (compressionMethod.equals("gzip")) {
+      this.compressionEnabled = true;
+    }
```
jack-berg, line 89: This method and `setEndpoint` are good candidates for helpers in the `:exporters:otlp:common` module.

jkwatson (approved):
Thanks!

anuraaga (approved):

PR #3689, opened 2021-10-01
Add timeout between delta collection

jsuereth:
Avoids a scenario where rapid collection cycles across multiple readers can do funny things.

A few important points:

- This sacrifices some accuracy for better runtime behavior in the presence of a poor-acting / aggressive metric reader
- This helps alleviate *some* concerns around Exemplar sampling when merging deltas.  (We always drop LHS exemplars during merge, which means we only keep latest exemplars from a delta right now).
- I placed some code to deal with clock-resets to avoid issues we saw in OpenCensus.   It's unclear how the whole SDK works in that scenario, or what we should do if we DO detect it.
- The timeout mechanism should be globally configurable eventually.  Just wanted to do a short cleanup for the next Java SDK release.

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DeltaMetricStorage.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DeltaMetricStorage.java
@@ -15,6 +15,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
@@ -29,6 +30,12 @@ class DeltaMetricStorage<T> {
   private final ConcurrentHashMap<Attributes, AggregatorHandle<T>> activeCollectionStorage =
       new ConcurrentHashMap<>();
   private final List<DeltaAccumulation<T>> unreportedDeltas = new ArrayList<>();
+  private long lastCollectionTimeNanos = 0;
+
+  // Minimum amount of time we allow between synchronous collections.
+  // This meant to reduce overhead when multiple exporters attempt to read metrics quickly.
+  // TODO: This should be configurable at the SDK level.
+  private static final long MINIMUM_COLLECTION_INTERVAL_NANOS = TimeUnit.MILLISECONDS.toNanos(100);
 
   DeltaMetricStorage(Aggregator<T> aggregator) {
     this.aggregator = aggregator;
@@ -71,12 +78,13 @@ class DeltaMetricStorage<T> {
    *
    * @param collector The current reader of metrics.
    * @param collectors All possible readers of metrics.
+   * @param epochNanos the timestamp of the current collection in nanoseconds since Jan 1, 1970 GMT.
    * @return The delta accumulation of metrics since the last read of a the specified reader.
    */
   public synchronized Map<Attributes, T> collectFor(
-      CollectionHandle collector, Set<CollectionHandle> collectors) {
+      CollectionHandle collector, Set<CollectionHandle> collectors, long epochNanos) {
     // First we force a collection
-    collectSynchronousDeltaAccumulationAndReset();
+    collectSynchronousDeltaAccumulationAndReset(epochNanos);
     // Now build a delta result.
     Map<Attributes, T> result = new HashMap<>();
     for (DeltaAccumulation<T> point : unreportedDeltas) {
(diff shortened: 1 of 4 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DeltaMetricStorage.java
```diff
@@ -96,7 +104,19 @@ public BoundStorageHandle bind(Attributes attributes) {
    * <p>All synchronous handles will be collected + reset during this method. Additionally cleanup
    * related stale concurrent-map handles will occur. Any {@code null} measurements are ignored.
    */
-  private synchronized void collectSynchronousDeltaAccumulationAndReset() {
+  private synchronized void collectSynchronousDeltaAccumulationAndReset(long epochNanos) {
+    // First check for a clock reset, and make sure we can limp along.
```
jkwatson, line 108: I'm curious how this can end up happening. I thought that the nano clock was guaranteed to be monotonically increasing, and since our clocks are pinned to creation millis + monotonic nanos, I would think this couldn't happen. 

Do you have a scenario that was uncovered in OC where this was violated?

jsuereth: Here's one of the original issues: https://github.com/census-instrumentation/opencensus-java/issues/2068
 
 TL;DR; I was unable to reproduce, and while it "should not happen", I'd rather be safe.
 
 Specifically for the nano-clock used in OTel I'm not sure if this will apply, but I also think we don't need to take chances.   IIRC This is one of a very few set of locations where timestamps matter in our metric collection vs. OpenCensus where there was the disruptor between measruement collection and aggregation.

jkwatson: I suppose someone could implement a very badly behaved clock, since they can provide their own. Can you add a comment in here about the unlikely/impossibility of a clock reset when using the standard Clock implementation?

anuraaga: Oops seemed to have missed hitting publish so affects you comment @jkwatson. I don't think we should use Clock or anything tied to wall time here

jsuereth: @anuraaga one of my goals is to avoid hitting any kind of clock repeatedly once collection starts, i.e. we peg it once and then use a value throughout.

It it makes sense to not used anchored clock for this, I can swap to using some other time that is also passed down, however I'd like to avoid doing a clock-lookup per-metric per-instrumentation-library, as would be the case here.

Additionally, when it comes to Exemplar time measurement, it sounds like that should ALSO not use Anchored clock, so there's a bit of a time cleanup to occur here.

Going to open a thread on slack to ask some questions.

jsuereth: Added #3691 to handle exemplars going forward *and* updated exemplars to call `.now` in this PR.

Still trying to understand the asks here on the slack channel.   The timestamp I'm using here is not associated with an AnchroedClock and I'm not sure what to anchor against to resolve this.   The timestamp is a wall-clock timestamp relating to the metric collection interval (start/current timestamps for aggregation temporality + gauges).  I built this logic trying to avoid adding a new clock and just re-using the existing lookup per-collection.

anuraaga: Left some notes in Slack but adding here too.

I singled out Exemplar having forgotten that normal metrics report a wall time too. Anytime a wall time is needed, indeed we use wall time - `Clock.getDefault().now()` seems like a reasonable source for it. For Exemplars, using the span's anchored clock would prevent a time jump from causing the Exemplar to report a time outside a span's window - this may not be a big enough issue on the backend to care about it too much.

But anywhere that duration is needed, we should always use `System.nanoTime()` to avoid invalid durations due to time jump - this is a general best practice  that I'd like to not create exceptions for. This collection interval is a duration, so it should use monotonic time, i.e. `System.nanoTime()`, not wall time. I don't think using `epochNanos`, which is a wall time, is appropriate here based on the principal of "never use wall time for a duration", a valuable principal to hold.

jsuereth: Thanks for copying slack convo here.  My next steps:

- I'm going to create a solution in SdkMetricProvider that does duration tracking using system nano time.
- A flag will be passed into metric storage denoting whether or not synchronous collection is needed, or we can just use the current reporting values.
- I'm going to fix the Clock.nanos() vs. Clock.now() call for exemplars in this PR (if it's ok).

Effectively after this PR, any area where we need wall-time in metrics SHOULD be using Clock.now(), and any location that needs duration will use Clock.nanos().

jkwatson: Completely aside from this discussion, I think all of this is why it is very important that we figure out a first class Timer implementation for otel metrics, so our users don't have to worry about these details. :)

anuraaga: > Effectively after this PR, any area where we need wall-time in metrics SHOULD be using Clock.now(), and any location that needs duration will use Clock.nanos().

Just want to confirm whether you're planning on doing it in this PR #3689 or a new one to follow up?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/DefaultSdkMeterProvider.java
```diff
@@ -45,6 +47,12 @@
   private final Set<CollectionHandle> collectors;
   private final List<MetricReader> readers;
   private final AtomicBoolean isClosed = new AtomicBoolean(false);
+  private final AtomicLong lastCollectionTimestamp = new AtomicLong(0);
```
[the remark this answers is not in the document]
jsuereth: Good point, fixed.

PR #3694, opened 2021-10-05
Change otlp http/protobuf default port from 4317 to 4318

jack-berg:
To reflect [updates](https://github.com/open-telemetry/opentelemetry-specification/pull/1975) to the spec.

The flip flopping is unfortunate, but luckily the collectors behavior still defaults to 4318 for http so I think its rather unlikely anyone has dependencies on the previous 4317 default.


jkwatson:
I would have sworn I already did this PR, but it appears not. Perhaps I am going mad.

jkwatson (approved):
Thanks!

PR #3697, opened 2021-10-06
Replace :proto usage with published io.opentelemetry.proto

anuraaga:


exporters/otlp/common/build.gradle.kts
```diff
@@ -11,19 +11,15 @@ plugins {
...
 
   api(project(":api:all"))
   api(project(":sdk:all"))
   api(project(":sdk:metrics"))
   compileOnly(project(":sdk-extensions:logging"))
 
-  // We only use the protos for HTTP, logging-otlp, and metric export for now. Let them expose the
-  // proto dependency in their POMs instead of here. This is fine because this artifact only
-  // contains internal code.
-  compileOnly(project(":proto"))
```
anuraaga, line 25: I had forgotten to remove this before, and it was ending up putting protobuf-java on the compile classpath and we accidentally used it. Unit tests didn't catch it since our tests always rely on protobuf-java and include it on the runtime classpath. Luckily this shouldn't happen anymore with this compileOnly removed

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/CodedInputStream.java
```diff
@@ -0,0 +1,245 @@
...
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+
+/**
+ * Minimal copy of protobuf-java's CodedInputStream, currently only used in GrpcStatusUtil.
+ *
+ * <p>This class is internal and is hence not for public use. Its APIs are unstable and can change
+ * at any time.
+ */
+public final class CodedInputStream {
```
anuraaga, line 52: Vendored in due to above

PR #3702, opened 2021-10-07
Fix SdkMeterProvider shutdown

anuraaga:


anuraaga:
Not completely sure but suspect this is the cause of some flakiness in this test (old non-closed reader collects metrics before new one can and the test can't find the data).

https://github.com/open-telemetry/opentelemetry-java/blob/main/integration-tests/src/testOtlpCommon/java/io/opentelemetry/integrationtest/OtlpExporterIntegrationTest.java#L286

PR #3704, opened 2021-10-07
Add a self-managed Prometheus HTTP server which does not integrate wi…

anuraaga:
…th an apps Registry

Got worried by having the `extends Collector` in our public API with the prometheus client library being pre-1.0 so felt like adding this. It's not clear to me what the expectations are for whether OTel metrics should share a registry with the app or not - perhaps the Collector version could live in a separate alpha artifact. Yes I understand our own artifact is still alpha :P But I'm optimistic that could change before the prom client and it'd be good for our API to anticipate that.


jsuereth:
FYI - You should ping me on things like this.  I actually had a branch that did just this (was lacking unit tests).

anuraaga:
> How much of a maintenance burden are we taking on by maintaining our own server implementation? That's my main concern with this.

235 lines of code. It's something, but I think it's fairly reasonable. It does look like an improvement to the UX when seeing the autoconfigure changes, and lets us remove the Collector due to public API reasons when stabilizing this module so hopefully it makes sense.

PR #3713, opened 2021-10-08
Deprecate SpanDataBuilder for removal from incubator.

anuraaga:


anuraaga:
@jkwatson Let's get this in 1.7, and if there's an issue we can revert for 1.8

jkwatson:
> @jkwatson Let's get this in 1.7, and if there's an issue we can revert for 1.8

oh no! Saw this too late!

PR #3714, opened 2021-10-08
Add a AutoConfiguredComponentWrapper SPI

### LINKED ISSUE #3695: Make MetricExporterConfiguration and SpanExporterConfiguration public
I have a need to write a delegating exporter that filters out sensitive data before data are exported by real exporters. Currently I have copied code from `SpanExporterConfiguration` so that I can initialize the real exporter in the same way that OT would config it.

The problem is that `SpanExporterConfiguration` is package-private and many helper classes it uses are also package-private. Thus copying the code to use for a delegator is very annoying.

It would be nice that at least some of the methods can be public so that creating delegating exporter would be an easier task:

- Autoconfigure calling configure exporter - load SPI delegating exporter 
  - SPI delegating exporter calling configure exporter again - configure OTLP exporter

One thing to be careful is that when configureExporter function is recursively called, SPI should not be loaded twice, or otherwise it may go into recursive loop and stack overflow.

anuraaga:
Also graduates DelegatingSpanData out of incubator.

Fixes #3695


anuraaga:
@edwardxia Can you check if this pattern will work fine for you?

jkwatson:
Hmm. Although this does solve the very narrow use-case at hand, I wonder if we're thinking too small here. If we start providing SPIs like this, I feel like we're going to end up with a giant pile of SPIs that will be hard to figure out what to do with. 

Would it make sense to take a step back and figure out a more general solution to the problem of "take something autoconfigured and tweak it a bit", rather than proliferate the SPIs?

anuraaga:
@jkwatson How about a single SPI with multiple methods for each of the autoconfigured pieces, `AutoconfiguredComponentWrapper`? That could reduce SPI proliferation

PR #3724, opened 2021-10-11
Prototype for Exponential Histogram Aggregator 

jamesmoessis:
Thought I'd raise a draft PR since this piece of work is becoming quite large, and it would probably be better to have some iterative review on it. I apologise for the size, but it was necessary to get any sort of functionality for testing. 

This histogram autoscales according to the recordings it receives. Each histogram starts at scale 20 and downscales if a recording does not fit (if calculated index can't be represented by `int`, or value would cause there to be more buckets than allowed). The max number of buckets is 320. I have taken these values from the [NrSketch](https://github.com/newrelic-experimental/newrelic-sketch-java) and the [Go implementation](https://github.com/open-telemetry/opentelemetry-go/pull/2261/files) done by @jmacd. The unit tests demonstrate this working. 

The Indexer used is a simple Logarithm mapper as seen in `DoubleExponentialHistogramBuckets.LogarithmMapper`. I haven't done any fancy indexing techniques. 

The recordings are stored via the `DoubleExponentialHistogramBuckets` which uses ~~NrSketch's circular backing array, the `WindowedCounterArray`, which `MultiTypeCounterArray` for variable bit-length. I've taken them directly from NrSketch which is Apache-2.0 so I've retained their copyright notice next to the Opentelemetry one. However the spotlessJava doesn't like this currently so the build fails. These classes need some additional cleanup too.~~ EDIT: changed this to a reference implementation `MapCounter` as discussed. 

For this to come out of draft, there are some todos which I have commented throughout the code: 
- [x] Thread safety
- [x] Write and test `merge()` which actually aggregates the accumulations together. 
- [ ] Write more tests to push the boundaries of the histogram. - EDIT: will do more of this in a later PR once we have settled on multiple indexing strategies. See discussion below 
- [x] Create assert types in `metrics-testing` and use them for the tests
- [x] `LongList` optimisation in `getBucketCounts()` mentioned previously #3637


jamesmoessis:
Question for @jsuereth - does an Accumulation need to be immutable? I ask because it would likely be more efficient to merge `accumulation` directly into `previousAccumulation` rather than creating an entirely new accumulation. This seems to be what other implementations are doing. 

If it needs to be immutable that's also fine, was just wondering if I should create a new accumulation or merge into an existing one.

jamesmoessis:
@yzhuge Are these algorithms supposed to work with recording the full double range in a single histogram? I am getting some inaccuracies while testing with `Double.MAX_VALUE`. The bucket's lower bound calculated at that index is slightly larger than the recorded value of `Double.MAX_VALUE`. It's off by one bucket. 

```java
// scale = -3. 
long i = valueToIndex(Double.MAX_VALUE); //  i = 128
BigDecimal lowerBound = BigDecimal.valueOf(256).pow( (int) i); // 256^128
assertThat(BigDecimal.valueOf(Double.MAX_VALUE)).isGreaterThanOrEqualTo(lowerBound); // fails. lower bound > max double
// but passes if I set i = 129
```

The indexing strategy is as shown in `valueToIndex` is in the PR, the simple `scalb` method.

This seems inconsistent to me, would you have any idea what this is due to?

jamesmoessis:
@yzhuge thankyou for the explanation, that makes sense. 

For the purposes of keeping this PR simple, I want just the one indexing strategy for now. I will make a note that other indexing strategies are optimal at different scales, and the code here is set up easily so they can be switched out for each other when scale changes. I can address that in a separate PR.

sdk/metrics-testing/src/main/java/io/opentelemetry/sdk/testing/assertj/metrics/ExponentialHistogramPointDataAssert.java
```diff
@@ -0,0 +1,45 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.testing.assertj.metrics;
+
+import io.opentelemetry.sdk.metrics.data.ExponentialHistogramPointData;
+import org.assertj.core.api.Assertions;
+
+public class ExponentialHistogramPointDataAssert
```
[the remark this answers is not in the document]
jamesmoessis: Done

sdk/metrics-testing/src/main/java/io/opentelemetry/sdk/testing/assertj/metrics/MetricAssertions.java
```diff
@@ -51,6 +53,15 @@ public static DoubleSummaryPointDataAssert assertThat(DoubleSummaryPointData poi
     return new DoubleSummaryPointDataAssert(point);
   }
 
+  public static ExponentialHistogramPointDataAssert assertThat(
```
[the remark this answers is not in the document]
jamesmoessis: Done

jamesmoessis:
@jsuereth I've implemented the `diff()` as part of the necessary refactor. It works similar to `merge()` and it uses the same underlying methods, however it decrements the counts. It does perform scaling where necessary.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MapCounter.java
```diff
@@ -0,0 +1,117 @@
...
+    return result.longValue();
+  }
+
+  @Override
+  public boolean isEmpty() {
+    return backing.isEmpty();
+  }
+
+  private void doIncrement(int index, long delta) {
+    AtomicLong value = backing.get(index);
```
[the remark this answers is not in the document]
jamesmoessis: I've had to change this method a bit more due to implementing `diff()` - it can take negative delta values now. It means the method is a little more complicated, and I also made it `synchronized`. 

As it currently works, `doIncrement()` probably doesn't need to be synchronized because of the way the callers work. The `Handle` is synchronized and the accumulations are immutable. But this may change in the future. 

If there's any thoughts on that let me know. (cc @jsuereth) 

https://github.com/open-telemetry/opentelemetry-java/pull/3724/files#diff-d689ce363b439db97c5d29f73118c4ff822068aa8013e1847db14f9d0f2dfbedR108

sdk/metrics-testing/src/main/java/io/opentelemetry/sdk/testing/assertj/metrics/ExponentialHistogramAssert.java
```diff
@@ -0,0 +1,61 @@
...
+
+  /** Ensures that {@code aggregation_temporality} field is {@code CUMULATIVE}. */
+  public ExponentialHistogramAssert isCumulative() {
+    isNotNull();
+    if (actual.getAggregationTemporality() != AggregationTemporality.CUMULATIVE) {
+      failWithActualExpectedAndMessage(
+          actual,
+          "aggregationTemporality: CUMULATIVE",
+          "Exepcted Histogram to have cumulative aggregation but found <%s>",
+          AggregationTemporality.CUMULATIVE,
```
jsuereth, line 31: I believe this is copy-pasted from my own mistaken use of this method.   You should drop this parameter, as you only have one %s which is the actual temporality.

jsuereth (approved):
Minor comment, but this is looking great for an  initial implementation/hook into the SDK!  Great work.

jkwatson:
@jsuereth is this something we want in the SDK for the next release, or are we waiting for spec stability around it?

sdk/metrics-testing/src/main/java/io/opentelemetry/sdk/testing/assertj/metrics/ExponentialHistogramAssert.java
jamesmoessis: Done!

jsuereth:
@jkwatson Regarding stability, only the MetricData interface piece matters (which is effectively based on our stable protocol), The rest is hidden in internal and cannot be exposed in the current public API.  From that standpoint I think this is ok.

@jkwatson / @anuraaga If you have suggestions on how to "hide" or "denote experimental" anything in MetricData, let me know.   I'd prefer (if possible) to just put a disclaimer on the class itself (and the enum value).   While ExponentialHistogram *cannot* change in breaking ways in our protocol, we may want room to encode it in a different way.   For now it's a pure interface, so I think we may have enough flexibility here, but PTAL at that area.

anuraaga:
@jkwatson @jsuereth This is an alpha module, I think it's always ok to merge if the current state looks fine.

@jamesmoessis Can you merge main since it's been a little while? I tried but don't seem to have permission to push to this PR

jamesmoessis:
@anuraaga I've merged main

sdk/metrics-testing/src/main/java/io/opentelemetry/sdk/testing/assertj/metrics/ExponentialHistogramAssert.java
```diff
@@ -0,0 +1,60 @@
...
+  protected ExponentialHistogramAssert(ExponentialHistogramData actual) {
+    super(actual, ExponentialHistogramAssert.class);
+  }
+
+  /** Ensures that {@code aggregation_temporality} field is {@code CUMULATIVE}. */
+  public ExponentialHistogramAssert isCumulative() {
+    isNotNull();
+    if (actual.getAggregationTemporality() != AggregationTemporality.CUMULATIVE) {
+      failWithActualExpectedAndMessage(
+          actual,
```
anuraaga, line 28: We'll need to add test cases for this class - should we file an issue for it or is it simple enough now? Or can we delete the class for now if it's not useful?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/MetricData.java
```diff
@@ -285,4 +307,18 @@ default DoubleHistogramData getDoubleHistogramData() {
...
+   *
+   * @return the {@link ExponentialHistogramData} if type is {@link
+   *     MetricDataType#EXPONENTIAL_HISTOGRAM}, otherwise a default empty data.
+   */
+  default ExponentialHistogramData getExponentialHistogramData() {
+    if (getType() == MetricDataType.EXPONENTIAL_HISTOGRAM) {
+      return (ExponentialHistogramData) getData();
+    }
+    return DoubleExponentialHistogramData.EMPTY;
+  }
```
anuraaga, line 323: Probably easier to add a test case than a TODO

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleExponentialHistogramBuckets.java
```diff
@@ -0,0 +1,287 @@
...
+    }
+    return totalCount;
+  }
+
+  void downscale(int by) {
+    if (by == 0) {
+      return;
+    } else if (by < 0) {
+      logger.warning(
+          "downScale() expects non-negative integer but was given"
```
anuraaga, line 88: Is this a programming bug in our codebase? Then we can throw IllegalStateException instead. Otherwise probably can add a test case

anuraaga (approved):
Thanks @jamesmoessis - I understand there will be some more test coverage in the future but tried to identify stuff that might not be "pushing the boundaries of the histogram" :) But let me know if you'd rather work later, keeping an eye on https://app.codecov.io/gh/open-telemetry/opentelemetry-java

jsuereth:
> @jkwatson @jsuereth This is an alpha module, I think it's always ok to merge if the current state looks fine.

We're trying to get Metrics graduated and out of alpha quickly.  I think we should start treating new-feature code review different than bug-fix / friction-fix / tuning code reviews (e.g. Jack's cardinality limits) differently going forward for metrics to make the job of releasing easier.

I.e. rather than viewing Metrics as alpha (where we can continue to dig holes we can't get out of), view it as "Beta" or "approaching stable".

That said my comment around this PR stands.  I think it's fine to include, the publicly exposed parts are based on stable protocol and flexible enough we can tweak the implementation going forward.

jmacd:
I will review this work in depth some time today, if that will help. I have a branch with an OTel-Go implementation of this and I'd like to compare and make notes.

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/MetricAdapter.java
```diff
@@ -92,6 +92,8 @@ private static String cleanMetricName(String descriptorMetricName) {
         return Collector.Type.SUMMARY;
       case HISTOGRAM:
         return Collector.Type.HISTOGRAM;
+      case EXPONENTIAL_HISTOGRAM:
+        return Collector.Type.UNKNOWN; // todo exporter for exponential histogram
```
jmacd, line 96: There's an implied question here -- what should an exporter or processor do when it sees this data and only knows about the old style of histogram.  The same question comes up (and is even more pressing) for the OTel collector. I prototyped a converted to produce explicit-boundary histograms here: https://github.com/open-telemetry/opentelemetry-collector/pull/3841

For Prometheus, there will be special considerations -- an auto-scaling aggregator is going to create problems.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/BucketMapper.java
```diff
@@ -0,0 +1,21 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.metrics.internal.aggregator;
+
+/**
+ * Interface defining methods for mapping values to buckets for {@link
+ * io.opentelemetry.sdk.metrics.data.ExponentialHistogramBuckets}.
+ */
+interface BucketMapper {
```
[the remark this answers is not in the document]
jmacd: True, there are reasons to have up to two implementations that we know about. The logarithm implementation stands in as a single implementation that works for all scales, but there are faster implementations if you restrict scale to non-negative values and non-positive values. It might be prudent to leave this an implementation detail until it comes to adding a second -- when I measured the difference between the logarithm method and the lookup table it was very slight, around 10ns difference. Is it worth having several more classes loaded to save 10ns per histogram observation? (Not sure.)

To those wondering why negative scale values matter, since they are indeed "extreme": for an auto-scaling histogram like our several prototypes, I believe it is easier to allow scale to go negative as a natural consequence of extreme input-value range than to check for and defend against extreme values at various other locations in the code.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleExponentialHistogramBuckets.java
jmacd: I believe this should be IllegalStateException.

jmacd (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleExponentialHistogramBuckets.java
```diff
@@ -48,6 +45,9 @@
   }
 
   public boolean record(double value) {
+    if (value == 0.0) {
+      throw new IllegalStateException("Illegal attempted recording of zero at bucket level.");
```
jkwatson, line 49: this will potentially crash the user's app, correct? Are we really ok with that, rather than ignoring the recording, and using the ThrottlingLogger to make sure we don't spam the logs too hard?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/BucketMapper.java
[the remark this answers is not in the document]
jamesmoessis: @jmacd The logarithm issue for me is less about the speed and more about the edge case where it returns an index that's off by one. I saw this when testing `Double.MIN_VALUE` and `Double.MAX_VALUE` in the same histogram, but it can also happen in more cases where the recorded value is close to a power of two. 

Some more discussion on that elsewhere in this PR: https://github.com/open-telemetry/opentelemetry-java/pull/3724#issuecomment-947870925

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleExponentialHistogramBuckets.java
jamesmoessis: It's indicative of a bug if this happens. The `Handle` ensures that all zero values go towards a separate counter `zeroCount`. This avoids `Math.log(0)`. 

Up to you if we log or throw an exception, I was just aligning with what @anuraaga said here on a similar issue: https://github.com/open-telemetry/opentelemetry-java/pull/3724#discussion_r745332146

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleExponentialHistogramBuckets.java
jamesmoessis: Done!

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/MetricAdapter.java
jamesmoessis: Interesting point. Converting to explicit boundary histogram could work. I also wonder how various backends would handle that. I certainly don't have the answers for that.

jamesmoessis:
Thanks @jmacd feel free to message me on Slack if you ever want to chat about exponential histogram stuff.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleExponentialHistogramBuckets.java
anuraaga: I'm not too sure, but one difference between the two is that one is a private method but this one is public.

Perhaps my comment wasn't clear enough, by programming bug I meant a programming bug of the OTel SDK, not user app. If this is a public API that can be called by a user, and it's their problematic code that causes 0.0 to be passed, then it's a case we wouldn't throw an exception.

sdk/metrics-testing/src/main/java/io/opentelemetry/sdk/testing/assertj/metrics/ExponentialHistogramAssert.java
jamesmoessis: I've gone ahead and added tests for it. Codecov seems to be green now.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleExponentialHistogramBuckets.java
jamesmoessis: This brings another issue to mind, this method should be package private, so I've got ahead and made it so. This is not a user facing method. If 0 is passed to this, it would mean there's a bug in the SDK. 

Given this info, should it be a log or an exception?

anuraaga (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleExponentialHistogramBuckets.java
jkwatson: exception is fine. I see this case is guarded by the caller. Can. you add a comment to that effect right before the line where you throw here? Thanks!

jamesmoessis: Sounds good, I've added a comment 👍

PR #3732, opened 2021-10-12
Add LogData interface

jack-berg:
Adds a `LogData` interface, implemented by `LogRecord`. Also elevates `Severity` to a top level class. 

This PR is in response to this [convo](https://github.com/open-telemetry/opentelemetry-java/pull/3685#discussion_r725338939).

The goal of this is to continue to mature the logging SDK. It seems like the [json log exporter](https://github.com/open-telemetry/opentelemetry-java/pull/3685) PR is blocked because it a dependency on the alpha `:sdk-extensions:logging` module. The logging [data model](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md) is still experimental. The logging otlp protos are [beta](https://github.com/open-telemetry/opentelemetry-proto). Once the proto reaches stability, the separation of the interface / implementation would potentially allow the stable release of a limited set of log functionality while the data model / sdk spec mature.


anuraaga:
Let's continue to follow up, there are big fish to fry

PR #3736, opened 2021-10-13
Add test tasks as dependencies to test report.

anuraaga:
Without dependencies, we get deprecation warnings like this one

```
 - Gradle detected a problem with the following location: '/Users/aanuraag/git/opentelemetry-java/sdk-extensions/logging/build/jacoco/test.exec'. Reason: Task ':all:codeCoverageReport' uses this output of task ':sdk-extensions:logging:test' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.2/userguide/validation_problems.html#implicit_dependency for more details about this problem.
```

This pattern isn't mentioned in the docs https://docs.gradle.org/current/samples/sample_jvm_multi_project_with_code_coverage.html but seems to have solved the warning.

### DIFF
--- a/all/build.gradle.kts
+++ b/all/build.gradle.kts
@@ -22,6 +22,8 @@ tasks {
   }
 }
 
+val testTasks = mutableListOf<Task>()
+
 dependencies {
   rootProject.subprojects.forEach { subproject ->
     // Generate aggregate coverage report for published modules that enable jacoco.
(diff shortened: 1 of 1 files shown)


iNikem:
https://melix.github.io/blog/2021/10/gradle-quickie-dependson.html :)

iNikem:
Replacing `dependsOn(testTasks)` with `executionData(testTasks)`? It seems to be working...

iNikem:
But you do depend on task outputs here, not on tasks themselves. You need coverage data that test tasks provide.

PR #3743, opened 2021-10-14
Add a PropertySource SPI to allow autoconfigure to have defaults defi…

anuraaga:
…ned.

The desire for the ability to tweak autoconfiguration config has come up in several situations

- Defining different defaults for a vendor distro of an autoconfiguration-based wrapper such as the AWS Lambda wrapper
- Falling back to sane defaults when user doesn't provide e.g., for service.name in the maven extension. A maven extension isn't a "service" in the traditional sense and unknown_service is a worse default than it could be
- The javaagent's file-based properties, which currently require the javaagent to reimplement the entire DefaultConfigProperties

I think more too - rule of three says this can be a good addition.


anuraaga:
Now I'm wondering about an `AutoConfigurationBuilder` approach for both this and others. SPI could be passed the builder and then call whatever it needs

```
builder.addSpanExporterWrapper(exporter -> exporter)
  .addPropertySource(() -> props)
  .build
```

Any thoughts on if I should try that?

PR #3753, opened 2021-10-15
Add an auto-configuration builder.

anuraaga:
This is an alternative to #3743 and #3714. I like it better

I realized that many uses don't need to use SPI to configure autoconfiguration - they're calling `OpenTelemetrySdkAutoConfiguration.inintialize()` and could instead just configure a builder which is simpler, and should be more flexible than an interface that only presents configuration through default no-op methods. We can remove some of the initialize overloads too.

The SPI that accepts the builder can still be exposed for downstream of downstream - for example, a javaagent end user could still implement an extension module with an implementation of the SPI if they need to themselves configure auto-configuration.


jkwatson:
I think this is a good idea. It allows a single SPI to configure pretty much anything that would need to be configured, without exploding the amount of SPIs. Do you think we could get rid of some of the existing ones with this approach?

jkwatson:
> It'd be cool if: 1. We had this customizer SPI. 2. our exporters adopted a convention of having a `toBuilder()` method so that you could do something like this to easily enhance the configuration done by autoconfiguration:
> 
> ```
> autoConfiguration.addSpanExporterCustomizer(spanExporter -> {
>             if (spanExporter instanceof OtlpGrpcSpanExporter) {
>               OtlpGrpcSpanExporter otlpGrpcSpanExporter = (OtlpGrpcSpanExporter) spanExporter;
>               return otlpGrpcSpanExporter.toBuilder()
>                   .addHeader("foo", "bar")
>                   .build();
>             }
>             return spanExporter;
>           });
> ```

We'd definitely need to make sure that the creation of an exporter didn't start it, though, so this would work. I'm looking at you, Prometheus exporter!

Issue #3750, opened 2021-10-14
Add LogSink#getBuilder() method


jack-berg:
**Is your feature request related to a problem? Please describe.**
The `LogSinkSdkProvider` API is awkward:

- Its `LogSink get(String insturmentationName, String instrumentationVersion)` method drops the arguments on the floor.  
- Callers of `LogSink#offer(LogRecord)` have to know about the instrumentation library, and resource when constructing log record instances.

**Describe the solution you'd like**
I'd like to see `LogSink` have a method like `LogRecordBuilder logBuilder()` whose implementation returns a `LogRecordBuilder` initialized with the resource and instrumentation library. 

The `LogSink` would have a reference to the resource and instrumentation library from the `LogSinkSdkProvider` which instantiated it.

jack-berg:
Alternatively, the `LogSink` could expose a `#getResource()` method to assist with building.

jack-berg:
This was resolved in #3759.

PR #3758, opened 2021-10-15
Log sdk refactor

jack-berg:
This refactor of the log sdk aims to make it more in line with traces, and metrics. It also makes it more intuitive for `LogSink` implementations to create logs with the appropriate resource and instrumentation library (Issue #3750). 

Its just a draft so I haven't completed all the javadoc or complete tests yet, but I did write [this test](https://github.com/jack-berg/opentelemetry-java/blob/log-sdk-refactor/sdk/logs/src/test/java/io/opentelemetry/sdk/logs/DemoTest.java#L24-L68) to demonstrate usage.


jack-berg:
I'm bad at opening draft pull requests.

PR #3775, opened 2021-10-20
Allow attribute values to be null explicitly

anuraaga:
In instrumentation, we've found it's very common to need to populate attributes that may or may not be there, depending on the type of request or conditions such as network failure. The attributes are applied generically, often through some interceptor type of mechanism, so all requests flow through despite populating different set of attributes.

Currently, we expect users of the attributes API to always check for null before sending in but I think this use case is common enough that we can explicitly support it, even though we prefer to reduce Nullable in our API where possible.

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/common/ArrayBackedAttributesBuilder.java
+++ b/api/all/src/main/java/io/opentelemetry/api/common/ArrayBackedAttributesBuilder.java
@@ -8,6 +8,7 @@ package io.opentelemetry.api.common;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import javax.annotation.Nullable;
 
 class ArrayBackedAttributesBuilder implements AttributesBuilder {
   private final List<Object> data;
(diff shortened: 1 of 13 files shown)


iNikem:
Some documentation should probably say somewhere that keys with null values will be ignored.

jkwatson:
> Is there a thought that some day `setAttribute` with `null` value might do something other than no-op?

For me, it's not that it might change, and more for users coming in and trying to understand the API and being surprised by the behavior, not knowing that the spec forbids nulls.

trask:
> > Is there a thought that some day `setAttribute` with `null` value might do something other than no-op?
> 
> For me, it's not that it might change, and more for users coming in and trying to understand the API and being surprised by the behavior, not knowing that the spec forbids nulls.

what would users expect other than no-op? (maybe I'm too close, in instrumentation repo we pass in `null` a lot)

jkwatson:
> > For me, it's not that it might change, and more for users coming in and trying to understand the API and being surprised by the behavior, not knowing that the spec forbids nulls.
> 
> what would users expect other than no-op? (maybe I'm too close, in instrumentation repo we pass in `null` a lot)

To remove the value (replace it with a null)

anuraaga:
Yeah I'm going to make it a separate method so hopefully that's ok. It think that while writing instrumentation we've found that the pattern of conditionally populating attributes is too common to ignore from a usability perspective.

Or perhaps we can just ignore usability since in the end it wasn't applied to the project's engineering values for some reason ;)

PR #3777, opened 2021-10-20
Move SemanticAttributes key into versioned package.

### LINKED ISSUE #3764: Make semconv packages explicitly versioned
OpenTelemetry semantic convention are considered to be a living standard to which changes are allowed over time.
As the conventions change over time, every version of the Otel specification captures/freezes a version of semantic
conventions. The telemetry sources are expected to explicitly declare the version of the semantic conventions
they use by recording the corresponding Schema URL in the emitted telemetry.

This repo currently declares a single semconv package. When the semantic conventions change in the specification
we will be forced to change the semconv package, which may break users of semconv package (e.g. if a name of an
attribute is change it will break the importing code, it won't compile any more).

We iterated on this problem in Go SDK and the solution we came up with is to have one semconv package per version
of semantic conventions, see [the PR](https://github.com/open-telemetry/opentelemetry-go/pull/1987) and [current structure](https://github.com/open-telemetry/opentelemetry-go/tree/main/semconv).

I suggest that we implement the same approach here, in this repo.

One alternate is to keep current design, but it doesn't match the intent of telemetry schemas and how they allow
semantic conventions to evolve.

**Additional context**
See [OTEP0152]( https://github.com/open-telemetry/oteps/blob/main/text/0152-telemetry-schemas.md) for the motivation of schemas and how the changes are allowed.

anuraaga:
Fixes #3764

We don't have to accept this but want to move things forward. I think what is unavoidable is that we must have different package names for incompatible versions of attribute definitions, there is nothing else that can work with Java and avoid dependency hell. Whether different packages should live in different artifacts is more technical and not as complicated of an issue.

There doesn't seem to be anything other than two package versioning schemes
1) Package per spec version, many packages are duplicates. To use a new attribute, an instrument author will almost always have to update all their import statements, instead of only rarely (so far breaking key names has been rare). This will disincentivise actually implementing new attributes.
2) Package version only incremented when an incompatible change is actually made. Out of sync with spec version but prevents duplicate packages. Instrumentations will still export a schema URL but it can still make it harder to understand what version of the spec something is targeting. Though I'm not sure whether 1) actually helps in that regard, both approaches seem to allow instrumentation to target 1.8.0 of the OTel API while using an older version of semantic attributes like 1.6.0, in the case that 1.7.0 introduced a breaking change.

I'm strongly against the first option since it seems to provide too much harm for the benefit, so proposing the second option in this PR.


jkwatson:
oof. I sure do wish that the semantic attributes obeyed semver. :( 

I think this approach is probably the right one, but I really don't like tagging something as "v1" when that version won't have any meaning to anyone, and the semantic conventions haven't even been marked as stable. Maybe until the spec marks them as stable, we should use "v0"? 

Another question: when/if a breaking change is introduced, do we deprecate the previous version, but leave it around in-perpetuity?

trask:
i think i finally understand this proposal(?)

so if an attribute is renamed or removed, the old one can just be marked deprecated, and would not require bumping the package name(?)

what kind of change _would_ require bumping the package name?

jkwatson:
> i think i finally understand this proposal(?)
> 
> so if an attribute is renamed or removed, the old one can just be marked deprecated, and would not require bumping the package name(?)
> 
> what kind of change _would_ require bumping the package name?

A removal of an attribute would, since the newly generated file wouldn't have it at all.

trask:
> > what kind of change _would_ require bumping the package name?
> 
> A removal of an attribute would, since the newly generated file wouldn't have it at all.

how about factoring in old versions somehow when generating, so the generated file could still have the old `@Deprecated` constant and not require a package bump?

anuraaga:
I realized that this PR is completely about UX. Given OTel values don't include usability, this sort of change doesn't align with the project so doesn't really make sense. Less is more.

trask:
> I realized that this PR is completely about UX. Given OTel values don't include usability, this sort of change doesn't align with the project so doesn't really make sense. Less is more.

I don't understand this, I thought we were having a productive discussion...

PR #3779, opened 2021-10-20
Update to Gradle 7.3 and fail build if binary compatibility broken

anuraaga:
Found the magic to allow japicmp to enforce binary compatibility.

Compatible change that passed, and still prints diff as normal
https://github.com/open-telemetry/opentelemetry-java/commit/c51d55fb0eebdab197aeae0d339cb22706683895

Incompatible change that fails, while also outputing the diff for local debugging
https://github.com/open-telemetry/opentelemetry-java/commit/c23ac3131e2da48a0eaa131957e5ece1b78ef166


anuraaga:
Ran locally with java 11, but looks like something about this change triggers bad behavior with java 17 will see what's up

`General error during canonicalization: PermittedSubclasses requires ASM9`

PR #3791, opened 2021-10-26
Autoconfigure experimental OTLP retry

jack-berg:
I'm taking another try at making progress on OTLP retry! 😬 

The PR #3636 draft PR demonstrated some strategies for implementing the OTLP retry required as a "MUST" in the spec. There were some good conversations and the resolution was to go to try to improve the wording of the spec so that the java implementation could be confident in its alignment.

I've [attempted](https://github.com/open-telemetry/opentelemetry-specification/pull/1974) to improve the spec language, but that has stalled due to what I would describe as a catch 22. The spec says SDKs "MUST" implement retry, but lacks specificity needed by SDKs to implement with confidence that they'll be aligned with future language. At the same time the spec is reluctant to add more specific language without supporting data.

This PR proposes a strategy for adding _experimental_ support for OTLP retry configuration. I hope it can help progress the issue by providing additional data to influence spec language. Here's what I propose:
- Define a retry policy as a bundle of the parameters described in the gRPC retry specification.
- For gRPC exporters using a gRPC client (i.e. not okhttp), define a method for configuring the retry policy via `DefaultGrpcExporterBuilder#addRetryPolicy(RetryPolicy)`. This method is on a class in an internal package. Neither the gRPC over okhttp or http/protobuf exporters have support for retry in this PR. 
- Provide an internal accessor for accessing the `DefaultGrpcExporterBuilder` delegate of the exporter builders. I.e. you can access the `OtlpGrpcSpanExporterBuilder#delegate` field reflectively, and call `addRetryPolicy()` on it. 
- Provide an experimental autoconfigure parameter for enabling a reasonable default retry policy via `otel.experimental.exporter.otlp.retry.enabled`. When enabled, this calls sets up the retry policy via the internal accessors.

The net affect is the ability to add a retry policy without any additional public API surface area. The compromise needed to accomplish this was an accessor that uses reflection. Maybe there's another way, but I couldn't think of one. 

**Why should we care about retry?**
This isn't a theoretical problem - I run otel for applications and I regularly see failed exports. Some applications may be able to tolerate intermittent failures in exports, but some will not. Lack of tolerance for network faults is a gap for production readiness.


jack-berg:
@jkwatson / @anuraaga what are you thoughts on including the upcoming release?

anuraaga:
@jkwatson Want to review this? Otherwise we can probably go ahead and merge

jkwatson:
> @jkwatson Want to review this? Otherwise we can probably go ahead and merge

Should be fine from my perspective. I can give it a thorough review tomorrow if you think it needs it

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/grpc/DefaultGrpcExporterBuilder.java
```diff
@@ -147,4 +164,25 @@ public DefaultGrpcExporterBuilder(
...
   }
+
+  /**
+   * Reflectively access a {@link DefaultGrpcExporterBuilder} instance in field called "delegate" of
+   * the instance.
+   *
+   * @throws IllegalArgumentException if the instance does not contain a field called "delegate" of
+   *     type {@link DefaultGrpcExporterBuilder}
+   */
+  public static <T> DefaultGrpcExporterBuilder<?> getDelegateBuilder(Class<T> type, T instance) {
```
jkwatson, line 175: should we scope `T` to at least have some bounds? Do we have a common interface we could enforce here to make it a bit clearer what this is for/about? This is a weird API to have be public at the moment.

jack-berg: > should we scope T to at least have some bounds? Do we have a common interface we could enforce here to make it a bit clearer what this is for/about? 

Unfortunately not. 

> This is a weird API to have be public at the moment.

Yeah its not ideal, but at least its an internal package. There's tons of public stuff in the internal packages that is questionable if you view it as truly public. 

The alternative to this was to have this functionality package private in the autoconfigure module. I like this approach though because it allows someone to programmatically configure a retry policy that is different from the default one that is used when you enable it via env vars.

jkwatson: Somehow I missed this was `internal`. Carry on!

tylerbenson:
@jack-berg is this still "experimental"?  Should we consider making it official at this point?

PR #3806, opened 2021-10-29
Actually flush in ExecutorServiceSpanProcessor

anuraaga:
The test for forceFlush for this processor is flaky because the processor doesn't actually force a flush. Now it does.

It's pretty complex fix though, I couldn't come up with a better pattern. Let me know with some code if anyone has a better idea.

@Oberon00 Are you still interested in this processor? Wondering if anyone is actually using it.


jack-berg (approved):

Oberon00:
No, actually I never used it, I don't know if it makes sense to maintain it 😕
My thought initially was that it might replace some other processor, but if it adds more complex code that nobody uses, I won't complain if it is removed.

PR #3808, opened 2021-10-30
Add diagnostic messages for metric registration conflicts.

jsuereth:
## Goals

- Output conflicting description/view/etc. information so you now what we found wrong.
- Ensure the SOURCE of the information is clear (which instrument registration, which view configuration).
- Allow source to be a config file in the (hopefully near) future.

## Changes
- Move instrument-descriptor to `internal` as we're adding very non-stable public methods to it and it wasn't used in any to-be-stable public API.
- Add new `SourceInfo` class that remembers where something was defined (either a custom location, or via stack trace).
- Expand `InstrumentDescriptor` and `View` to include `SourceInfo`, defaulting to stack-trace-on-registration.
- Create new utility, with tests, for the (large) error message on conflict.
- Create utility to enable/disable debugging on metrics SDK

## Moved to future work
- Figure out what to do when a view renames a metric *and* the instrument descriptor of the conflict is compatible
  (Succeeds today) https://github.com/open-telemetry/opentelemetry-java/issues/3809
- Error if the same name but different instrument is registered on the same Meter, but a view is renaming one of them.

Looking for early feedback on the approach.


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/debug/DebugConfig.java
```diff
@@ -0,0 +1,43 @@
...
+ * at any time.
+ */
+public final class DebugConfig {
+  private static final String ENABLE_METRICS_DEBUG_PROPERTY = "io.opentelemetry.sdk.metrics.debug";
+  private static boolean enabled;
+
+  private DebugConfig() {}
+
+  static {
+    // Should we also look for an environment variable?
```
jkwatson, line 21: yes, at least for consistency with the rest of the configuration available

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/debug/SourceInfo.java
```diff
@@ -0,0 +1,64 @@
...
+   *
+   * <p>This will attempt to ignore SDK classes.
+   */
+  static SourceInfo fromCurrentStack() {
+    if (!DebugConfig.isMetricsDebugEnabled()) {
+      return noSourceInfo();
+    }
+    // Use exceptions to pull current stack trace.
+    Exception e = new Exception();
+    return new StackTraceSourceInfo(e.getStackTrace());
```
jkwatson, line 50: how about using `Thread.getStackTrace()` here rather than creating an exception for it??

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/debug/DebugConfig.java
jack-berg: It'd be nice if we could use `ConfigProperties` here. But the implementation (`DefaultConfigProperties`) is in the autoconfigure module and is package private.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/descriptor/MetricDescriptor.java
```diff
@@ -41,12 +55,26 @@ public static MetricDescriptor create(View view, InstrumentDescriptor instrument
...
+  /** Returns true if another metric descriptor is compatible with this one. */
   public boolean isCompatibleWith(MetricDescriptor other) {
-    // TODO: implement.
-    return equals(other);
+    return sameString(getName(), other.getName())
+        && sameString(getDescription(), other.getDescription())
+        && sameString(getUnit(), other.getUnit());
+  }
+
+  private static boolean sameString(String lhs, String rhs) {
```
jkwatson, line 74: `Objects.equals(lhs, rhs)` ?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/debug/DebugConfig.java
```diff
@@ -0,0 +1,43 @@
...
+package io.opentelemetry.sdk.metrics.internal.debug;
+
+/**
+ * Determines if the SDK is in debugging mode (captures stack traces) or not.
+ *
+ * <p>This class is internal and is hence not for public use. Its APIs are unstable and can change
+ * at any time.
+ */
+public final class DebugConfig {
+  private static final String ENABLE_METRICS_DEBUG_PROPERTY = "io.opentelemetry.sdk.metrics.debug";
```
jack-berg, line 15: Shouldn't be follow the spec's recommendation for [language specific environment variables](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md)?

I.e. something like `otel.java.sdk.metrics.debug`, or `otel.experimental.sdk.metrics.debug`.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/debug/DebugConfig.java
```diff
@@ -0,0 +1,43 @@
...
+   *
+   * <p>This will grab stack traces on instrument/view registration.
+   */
+  public static boolean isMetricsDebugEnabled() {
+    return enabled;
+  }
+
+  /** Returns the message we send for how to enable better metrics debugging. */
+  public static String getHowToEnableMessage() {
+    return "To enable better debugging, run your JVM with -Dio.opentelemetry.sdk.metrics.debug=true";
```
jack-berg, line 36:
```suggestion
    return "To enable better debugging, run your JVM with -D" + ENABLE_METRICS_DEBUG_PROPERTY + "=true";
```

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/debug/DebugConfig.java
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  /** A mechanism to enable debugging for testing without having to recompile. */
```
jack-berg, line 39:
```suggestion
  /** A mechanism to enable debugging for testing without having to recompile. */
  // Visible for testing
```

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MeterSharedState.java
```diff
@@ -91,7 +91,7 @@ public final WriteableMetricStorage registerSynchronousMetricStorage(
       try {
         storage.add(getMetricStorageRegistry().register(currentStorage));
       } catch (DuplicateMetricStorageException e) {
-        logger.log(Level.WARNING, e, () -> "Failed to register metric.");
+        logger.log(Level.WARNING, DebugUtils.duplicateMetricErrorMessage(e), e);
```
jkwatson, line 94: using the original form (with the message Supplier) here might be a little better, in case someone doesn't care and turns this up to ERROR only.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/debug/TestSourceInfo.java
```diff
@@ -0,0 +1,37 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.metrics.internal.debug;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+
+class TestSourceInfo {
```
jack-berg, line 12:
```suggestion
class SourceInfoTest {
```

jack-berg: This applies to `TestMetricDescriptor`, `TestDebugConfig`, and the other `TestSourceInfo` as well.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DuplicateMetricStorageException.java
```diff
@@ -7,13 +7,30 @@
...
+ */
 class DuplicateMetricStorageException extends IllegalArgumentException {
   private static final long serialVersionUID = 1547329629200005982L;
+  private final MetricDescriptor existing;
+  private final MetricDescriptor conflict;
 
   DuplicateMetricStorageException(
       MetricDescriptor existing, MetricDescriptor next, String message) {
     // TODO: Better error messages including async vs. sync instruments.
     super(message + " Found: " + existing + ", Want: " + next);
```
jkwatson, line 24: I agree with the TODO here. I'm pretty sure I wouldn't understand what "Want" vs. "Found" would mean. :)

sdk/metrics/src/debugEnabledTest/java/io/opentelemetry/testing/TestSourceInfo.java
```diff
@@ -0,0 +1,126 @@
...
+import io.opentelemetry.sdk.metrics.common.InstrumentType;
+import io.opentelemetry.sdk.metrics.common.InstrumentValueType;
+import io.opentelemetry.sdk.metrics.internal.debug.SourceInfo;
+import io.opentelemetry.sdk.metrics.internal.descriptor.InstrumentDescriptor;
+import io.opentelemetry.sdk.metrics.internal.descriptor.MetricDescriptor;
+import io.opentelemetry.sdk.metrics.internal.state.DebugUtils;
+import io.opentelemetry.sdk.metrics.view.View;
+import org.junit.jupiter.api.Test;
+
+// Note: This class MUST be outside the io.opentelemetry.metrics package to work correctly.
```
jack-berg, line 19: Why?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/debug/SourceInfo.java
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * Constructs a custom source information object that is meant to represent a non-code source of
+   * metric configuration.
+   *
+   * @param sourcePath A URI, filename or other user-identifying feature of where the metric
+   *     configuration came from.
+   * @param lineNumber A line number in the source, denoting where the configuration resides.
+   */
+  static SourceInfo fromConfigFile(String sourcePath, int lineNumber) {
```
jkwatson, line 61: It seems a little premature to have this here, before we have any config file format defined.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/descriptor/InstrumentDescriptor.java
```diff
@@ -18,7 +27,8 @@ public static InstrumentDescriptor create(
       String unit,
       InstrumentType type,
       InstrumentValueType valueType) {
-    return new AutoValue_InstrumentDescriptor(name, description, unit, type, valueType);
+    return new AutoValue_InstrumentDescriptor(
+        name, description, unit, type, valueType, SourceInfo.fromCurrentStack());
```
jack-berg, line 31: Thinking ahead to when there is file based configuration of metrics and `CustomSourceInfo` is needed, `SourceInfo` is probably going to need to be an argument here so that the code that is interpreting the config file and setting up the metrics can set the appropriate `SourceInfo`.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/debug/SourceInfo.java
jsuereth: Watch this space... (but can remove if you'd like)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/debug/DebugConfig.java
jsuereth: Yes.  For now I duplicated what was used for `Context` API (which predated the specification).  Will update.

sdk/metrics/src/debugEnabledTest/java/io/opentelemetry/testing/TestSourceInfo.java
jkwatson: Because we skip any thing in the stack in `io.opentelemetry.metrics` package

jsuereth: We're truncating the "io.opentelemetry.sdk.metrics" package stack traces when we dump Instrument registration to avoid useless context.  I.e. we try to grab the stack trace where an instrument was registered, but deep from the bowels of the SDK.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/descriptor/InstrumentDescriptor.java
jsuereth: That won't be the case for InstrumentDescriptor.  Config-file will only affect `View` and `MetricDescriptor`.

You're right, but not in this location.

sdk/metrics/src/debugEnabledTest/java/io/opentelemetry/testing/TestSourceInfo.java
jack-berg: Right 👌

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/debug/SourceInfo.java
jkwatson: yeah, let's drop this until we have something to hook it up to. Glad it will be possible, though!

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/debug/SourceInfo.java
jsuereth: Nice call out.   Didn't actually realize that existed!  I've mostly done this from the compiler-side before :)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/descriptor/MetricDescriptor.java
jsuereth: Same, my java is pretty rusty, thanks for the call out!

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/debug/StackTraceSourceInfo.java
```diff
@@ -41,8 +41,9 @@ public String multiLineDebugString() {
     return "\tat unknown source";
   }
 
-  private static boolean isOTelSdkStackTrace(StackTraceElement e) {
+  private static boolean isUninterestingStackTrace(StackTraceElement e) {
     return (e.getClassName() != null)
-        && e.getClassName().startsWith("io.opentelemetry.sdk.metrics");
+        && (e.getClassName().startsWith("io.opentelemetry.sdk.metrics")
+            || e.getClassName().startsWith("java.lang"));
```
jkwatson, line 47: what's the case for this one?

jsuereth: Thread.currentThread().getStackTrace() includes the Thread class's stack.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DuplicateMetricStorageException.java
jsuereth: Updated PTAL

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/debug/DebugConfig.java
jsuereth: Attempted to unify with the logic.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DuplicateMetricStorageException.java
```diff
@@ -20,8 +20,7 @@
 
   DuplicateMetricStorageException(
       MetricDescriptor existing, MetricDescriptor next, String message) {
-    // TODO: Better error messages including async vs. sync instruments.
-    super(message + " Found: " + existing + ", Want: " + next);
+    super(message + " Found previous metric: " + existing + ", cannot register metric: " + next);
```
jkwatson, line 23: 👍🏽

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DuplicateMetricStorageException.java
jkwatson: looks good!

jack-berg (approved):

jkwatson (approved):
👍🏽

Issue #3804, opened 2021-10-28
Refine log fields


jack-berg:
A handful of questions have come up regarding log fields. This issue tracks them so we don't forget.

1. All the fields in the [log proto](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/logs/v1/logs.proto) are optional except timestamp, but only traceId, spanId, severityText, and Name are `@Nullable`. The other fields have default values. We should have a uniform strategy on this, or good reasons for breaking uniformity.
2. There have been [discussions](https://github.com/open-telemetry/opentelemetry-java/pull/3759#discussion_r738021302) about removing the `setBody(Body body)` method and simply keeping `setBody(String message)`. 
3. There have been [discussions](https://github.com/open-telemetry/opentelemetry-java/pull/3759#discussion_r738020720) about dropping or renaming the `name` field. 
4. We should make `setFlags(int flags)` more usable. Perhaps `setFlags(LogFlags)` akin to `TraceFlags`?

jack-berg:
I've included the change of the pattern to `LogEmitter.logBuilder().emit()` in PR #3759.

jack-berg:
I believe 2 is resolved now that spec [PR #2096](https://github.com/open-telemetry/opentelemetry-specification/pull/2096) was merged explaining why log body should be structured. 

> 2. There have been discussions about removing the setBody(Body body) method and simply keeping setBody(String message).

anuraaga:
Would it make sense to leave the structured version out of the SDK until someone asks for it? It mentions compatibility with certain existing log formats, but perhaps we can wait until we know this actually exists in the Java ecosystem, unless someone already has an example.

jack-berg:
I'd be on board to make it an internal implementation detail until required. Can keep the `setBody(String)` method and add a `setBody(Body)` overload when requested?

jack-berg:
All but one of these items were addressed in #3837. 

The last unaddressed item is about the log `name` field. This is being actively discussed in [the specification](https://github.com/open-telemetry/opentelemetry-specification/issues/2074). If the field is removed in the spec, I'll remove it here. But for now, I'll close this ticket.

PR #3811, opened 2021-11-01
Change Log flags field to TraceFlags

jack-berg:
Part of Issue #3804.

Adjust the log `int getFlags()` field to `TraceFlags getTraceFlags()`.

### DIFF
--- a/exporters/logging-otlp/src/test/java/io/opentelemetry/exporter/logging/otlp/OtlpJsonLoggingLogExporterTest.java
+++ b/exporters/logging-otlp/src/test/java/io/opentelemetry/exporter/logging/otlp/OtlpJsonLoggingLogExporterTest.java
@@ -12,6 +12,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 import io.github.netmikey.logunit.api.LogCapturer;
 import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.trace.TraceFlags;
 import io.opentelemetry.sdk.common.InstrumentationLibraryInfo;
 import io.opentelemetry.sdk.logs.data.LogData;
 import io.opentelemetry.sdk.logs.data.Severity;
@@ -34,7 +35,7 @@ class OtlpJsonLoggingLogExporterTest {
       LogData.builder(RESOURCE, InstrumentationLibraryInfo.create("instrumentation", "1"))
           .setName("testLog1")
           .setBody("body1")
-          .setFlags(0)
+          .setTraceFlags(TraceFlags.getDefault())
           .setSeverity(Severity.INFO)
           .setSeverityText("INFO")
           .setSpanId("8765432112345876")
@@ -47,7 +48,7 @@ class OtlpJsonLoggingLogExporterTest {
       LogData.builder(RESOURCE, InstrumentationLibraryInfo.create("instrumentation2", "2"))
           .setName("testLog2")
           .setBody("body2")
-          .setFlags(0)
+          .setTraceFlags(TraceFlags.getDefault())
           .setSeverity(Severity.INFO)
           .setSeverityText("INFO")
           .setSpanId("8765432112345875")
--- a/exporters/otlp-http/logs/src/test/java/io/opentelemetry/exporter/otlp/http/logs/OtlpHttpLogExporterTest.java
+++ b/exporters/otlp-http/logs/src/test/java/io/opentelemetry/exporter/otlp/http/logs/OtlpHttpLogExporterTest.java
@@ -22,6 +22,7 @@ import com.linecorp.armeria.testing.junit5.server.mock.MockWebServerExtension;
 import com.linecorp.armeria.testing.junit5.server.mock.RecordedRequest;
 import io.github.netmikey.logunit.api.LogCapturer;
 import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.trace.TraceFlags;
 import io.opentelemetry.exporter.otlp.internal.logs.ResourceLogsMarshaler;
 import io.opentelemetry.exporter.otlp.internal.okhttp.OkHttpExporter;
 import io.opentelemetry.proto.collector.logs.v1.ExportLogsServiceRequest;
@@ -329,7 +330,7 @@ class OtlpHttpLogExporterTest {
         .setTraceId(IdGenerator.random().generateTraceId())
         .setSpanId(IdGenerator.random().generateSpanId())
         .setEpoch(Instant.now())
-        .setFlags(0)
+        .setTraceFlags(TraceFlags.getDefault())
         .build();
   }
 }
--- a/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/logs/LogMarshaler.java
+++ b/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/logs/LogMarshaler.java
@@ -5,6 +5,7 @@
 
 package io.opentelemetry.exporter.otlp.internal.logs;
 
+import io.opentelemetry.api.trace.TraceFlags;
 import io.opentelemetry.exporter.otlp.internal.KeyValueMarshaler;
 import io.opentelemetry.exporter.otlp.internal.MarshalerUtil;
 import io.opentelemetry.exporter.otlp.internal.MarshalerWithSize;
@@ -25,7 +26,7 @@ final class LogMarshaler extends MarshalerWithSize {
   private final MarshalerWithSize anyValueMarshaler;
   private final KeyValueMarshaler[] attributeMarshalers;
   private final int droppedAttributesCount;
-  private final int flags;
+  private final TraceFlags traceFlags;
   @Nullable private final String traceId;
   @Nullable private final String spanId;
 
@@ -46,7 +47,7 @@ final class LogMarshaler extends MarshalerWithSize {
         attributeMarshalers,
         // TODO (trask) implement droppedAttributesCount in LogRecord
         0,
-        logData.getFlags(),
+        logData.getTraceFlags(),
         logData.getTraceId(),
         logData.getSpanId());
   }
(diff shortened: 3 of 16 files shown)


exporters/logging-otlp/src/test/java/io/opentelemetry/exporter/logging/otlp/OtlpJsonLoggingLogExporterTest.java
```diff
@@ -47,7 +48,7 @@
       LogData.builder(RESOURCE, InstrumentationLibraryInfo.create("instrumentation2", "2"))
           .setName("testLog2")
           .setBody("body2")
-          .setFlags(0)
+          .setTraceFlags(TraceFlags.getDefault())
```
jkwatson, line 51: might be nice to have one example that's sampled, as well. That goes for all the tests in this PR.

jack-berg: There are a lot of tests that create `LogData` but most of the functionality under test doesn't concern the trace flags field directly. I'd prefer to limit testing of the different shapes of trace flags to the tests that are directly concerned with serializing / deserializing trace flag values. That would include `OtlpJsonLoggingLogExporter`, `LogsRequestMarshalerTest`, which tests serialization of `LogData` to proto, and `OtlpExporterIntegrationTest`, which tests exporting to the collector and back to a java process.

If we have tests that validate the correctness of the low level code, then we can trust that it will work for higher level code that consumes it.

jkwatson: sounds good to me.  TraceFlags aren't a mandatory field, correct? We could just leave them out of the high level tests altogether in most cases, I'd assume.

anuraaga: Instead of three methods, should it be `setSpanContext`?

anuraaga: Another conceptual question, I don't think we have trace flags on metrics exemplar. It feels like the spancontext information added to a log would match what's added to an exemplar.

jkwatson: > Another conceptual question, I don't think we have trace flags on metrics exemplar. It feels like the spancontext information added to a log would match what's added to an exemplar.

@jsuereth @tigrannajaryan thoughts?

jkwatson: > Instead of three methods, should it be `setSpanContext`?

Interesting. @tigrannajaryan any thoughts on this?

tigrannajaryan: > Instead of three methods, should it be `setSpanContext`?

From data model perspective one degenerate case is when a log record is recorded in the context of a trace but without a span context, e.g. there is an incoming request which does have a context from which we can extract the trace id, but we don't create a span when we are processing. Imagine that we do record a log in this situation. So, we do know a trace id that we can put in the log record, but there is no span id to put in the log record.

This is in theory. In practice I don't know if our Trace API allows this. Can I somehow extract the trace id from the current context without creating a span?

If it is possible that I would want to record the trace id and traceflags, but not span id.

This is probably a very weak argument. We can just have a single method and allow the span id argument to be omitted/empty if this use case really exists.

jkwatson: At least in Java, the only way to get the trace id from a Span in the Context, so you'll have the whole SpanContext there.

jack-berg: > sounds good to me. TraceFlags aren't a mandatory field, correct? We could just leave them out of the high level tests altogether in most cases, I'd assume.

@jkwatson technically none of the fields are mandatory. You can call `logEmitter.logBuilder().emit()` without setting any fields. All the fields are either nullable or have default values. See [LogDataBuilder](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/logs/src/main/java/io/opentelemetry/sdk/logs/data/LogDataBuilder.java) for the specifics. 

But according to the [proto](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/logs/v1/logs.proto), every field is optional. Though timestamp is a bit of special case since while its not marked as `[Optional]`, a value of 0 indicates that the timestamp is unknown or missing. And really what's the difference between optional and unknown or missing?

jkwatson: I was just suggesting that for tests, we could have a minimal (whatever we decide that means) log entry to test with, unless we're testing the actual mapping to OTLP, then provide lots more detail/combinations.

breedx-splk (approved):
Looks ok to me.

exporters/logging-otlp/src/test/java/io/opentelemetry/exporter/logging/otlp/OtlpJsonLoggingLogExporterTest.java
anuraaga: Hmm - took another step back I have a feeling the builder should only accept `Context`, and the data should have a `getSpanContext()` (it could also have the three methods separately, but don't see a reason to separate as john explains https://github.com/open-telemetry/opentelemetry-java/pull/3811/files#r741313336). 

It seems that appender writers shouldn't need to worry about the details of spans - they would also only be calling `setContext` if they have a way to reference Context efficiently for some reason (think usually they won't), otherwise our code would just call `Context.current()` - I think this is similar to how the metrics SDK implements exemplars. Indeed, I lean towards not even having `setContext` really - appenders aren't instrumentation, I can't imagine a situation they have a Context available to pass in.

We could merge this PR as is but wonder if it's worth it, there seems to be some obvious cleanups we can apply that will just churn the exact lines modified here. Up to you.

jkwatson: This makes me wonder if we shouldn't default to using Context.current() like we do for span parenting.

kubawach (approved):
LGTM :)

exporters/logging-otlp/src/test/java/io/opentelemetry/exporter/logging/otlp/OtlpJsonLoggingLogExporterTest.java
jack-berg: Yeah I don't think its worth merging the PR if its just going to change in the near future. 

Some additional food for though:
- The Log4j2 [OpenTelemetryContextDataProvider](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/OpenTelemetryContextDataProvider.java) currently associates span context with Log4j2 log events.
- In my Log4j2 appender [prototype](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4375/files#diff-cce9527647ecff346de08cfc039060d71435d1455a03bd59858d1c71d0ee4298R51-R64), I read those fields out of the context map and associate them with the `LogBuilder` (note I haven't extracted the `TRACE_FLAGS` field yet). 
- If we depend on the log appenders calling `Context.current()` to get context, rather than first injecting the context into the log as is done in Log4j2, then we are assuming that the appenders are called synchronously from the same thread when a `logger.log(...)` is invoked, right? Because if the appender is not called synchronously on the same thread, then `Context.current()` won't result the right context. I think.

anuraaga: Interesting - the context provider can't inject Context itself because it needs to set a string map. So if we rely on the injection than baggage would also all need to be injected as key value pairs. Any plans for that out of curiosity?

I have a feeling that expecting the OTel appender to be synchronous is going to be more practical but not sure.

jack-berg: > So if we rely on the injection than baggage would also all need to be injected as key value pairs. Any plans for that out of curiosity?

I haven't heard anything about logs and baggage. @tigrannajaryan do you have any insight into this?

> I have a feeling that expecting the OTel appender to be synchronous is going to be more practical but not sure.

Log4j2 has the [AsyncAppender](https://logging.apache.org/log4j/2.x/manual/appenders.html#AsyncAppender), which allows you to wrap other appenders and causes log events to be written to them on a separate thread. We'd have to tell users they can't wrap the otel appender in an `AsyncAppender`. It would be a bit strange, but not outright wrong, for a user to use the otel appender with the `AsyncAppender`, since `BatchLogProcessor` already exports on a separate thread. 

[Logback](http://logback.qos.ch/manual/appenders.html#AsyncAppender) has the same concept.

jkwatson: Yes, I've been worrying about async appenders since the very early days of the logging project. I don't know that we have any good solution for that.

jack-berg: We discussed in the 11/4 Java + Instrumentation Office hours and decided to have a `LogBuilder#setContext(Context)` method and `LogData#getSpanContext()`.

jack-berg:
Replacing this PR with #3837, which includes the changes we discussed in the 11/4 Java + Instrumentation Office Hours.

PR #3812, opened 2021-11-01
Deprecate MeterProvider#get(..) methods

jack-berg:
In PR #3759 there was a [comment](https://github.com/open-telemetry/opentelemetry-java/pull/3759#discussion_r738014734) about removing the `SdkLogEmitterProvider#get(...)` methods which are made obsolete by `SdkLogEmitterProvider#logEmitterBuilder(..)`. 

The same is true for the `MeterProvider#get(..)`. This PR marks them as deprecated and changes all usages to `MeterProvider#meterBuilder(..)`.

### DIFF
--- a/api/metrics/src/main/java/io/opentelemetry/api/metrics/MeterProvider.java
+++ b/api/metrics/src/main/java/io/opentelemetry/api/metrics/MeterProvider.java
@@ -27,7 +27,9 @@ public interface MeterProvider {
    * @param instrumentationName The name of the instrumentation library, not the name of the
    *     instrument*ed* library.
    * @return a meter instance.
+   * @deprecated Use {@link #meterBuilder(String)}
    */
+  @Deprecated
   default Meter get(String instrumentationName) {
     return meterBuilder(instrumentationName).build();
   }
@@ -40,7 +42,9 @@ public interface MeterProvider {
    * @param instrumentationVersion The version of the instrumentation library.
    * @param schemaUrl Specifies the Schema URL that should be recorded in the emitted metrics.
    * @return a meter instance.
+   * @deprecated Use {@link #meterBuilder(String)}
    */
+  @Deprecated
   default Meter get(String instrumentationName, String instrumentationVersion, String schemaUrl) {
     return meterBuilder(instrumentationName)
         .setInstrumentationVersion(instrumentationVersion)
--- a/api/metrics/src/main/java/io/opentelemetry/api/metrics/MeterProvider.java
+++ b/api/metrics/src/main/java/io/opentelemetry/api/metrics/MeterProvider.java
@@ -27,9 +27,7 @@ public interface MeterProvider {
    * @param instrumentationName The name of the instrumentation library, not the name of the
    *     instrument*ed* library.
    * @return a meter instance.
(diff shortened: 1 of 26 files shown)


api/metrics/src/main/java/io/opentelemetry/api/metrics/MeterProvider.java
```diff
@@ -27,7 +27,9 @@
    * @param instrumentationName The name of the instrumentation library, not the name of the
    *     instrument*ed* library.
    * @return a meter instance.
+   * @deprecated Use {@link #meterBuilder(String)}
    */
+  @Deprecated
   default Meter get(String instrumentationName) {
```
jkwatson, line 33: I think we might want to keep this one around, since it'll be easy mode for application developers who don't care about version/schemaUrl. Thoughts?

jack-berg: If we do I'd like to add that helper to the log sdk as well.

@anuraaga do you have thoughts on this? Was your comment in the log sdk PR related to removing `get(String instrumentationName)` and `get(String instrumentationName, String instrumentationVersion, String schemaUrl)`?

mateuszrzeszutek: > If we do I'd like to add that helper to the log sdk as well.

Logs SDK will only be used by the appender instrumentation, not directly by the user, won't it?
The `MeterProvider` needs to be called whenever you want to add a custom counter to your app - and usually the instrumentation name is already confusing enough even without the builder and the other params.

jkwatson: I wonder if we should call it "loggerName" or some such in the logging SDK, as that will be more familiar to users of logging APIs.

jack-berg: > Logs SDK will only be used by the appender instrumentation, not directly by the user, won't it?
Because Otel Logs are covering both conventional application logs [as well as events](https://github.com/open-telemetry/opentelemetry-specification/issues/2074#issuecomment-954781513), I can definitely see a strong argument that the SDK _would_ be used directly in applications. 

> I wonder if we should call it "loggerName" or some such in the logging SDK, as that will be more familiar to users of logging APIs.
The spec currently doesn't talk about instrumentation library, and how its involved in mapping logs from logging frameworks. Not sure if that's covered under one of the outstanding issues, eg [Issue #2069](https://github.com/open-telemetry/opentelemetry-specification/issues/2069).

anuraaga: Sorry I did mean to keep one get method with just the library name, I probably missed the removal from the logs PR. It's nice to have the quick, but not parameter-exploding, factory method along a builder.

kubawach (approved):
Agree with leaving default, simple Meter get(name).

mateuszrzeszutek (approved):

jkwatson (approved):

anuraaga (approved):

PR #3814, opened 2021-11-02
Close response on successful response in HTTP exporter.

### LINKED ISSUE #3813: okhttp3.OkHttpClient - A connection was leaked. Did you forget to close a response body?
We are using the OpenTelemetry Java Agent 1.7.1.  We are continually getting the below warning when our spans are exported to our collector.  Is there any way to determine what is causing this warning?  When we run the agent in debug mode, we do not get any additional details of root cause.  This issue consumes all available threads until we stop the process to reclaim them.

[otel.javaagent 2021-11-01 20:04:28:616 -0500] [OkHttp ConnectionPool] WARN okhttp3.OkHttpClient - A connection to https://somehost.com:9999/ was leaked. Did you forget to close a response body?
java.lang.Throwable: response.body().close()
        at okhttp3.internal.platform.Platform.getStackTraceForCloseable(Platform.java:149)
        at okhttp3.internal.connection.Transmitter.callStart(Transmitter.java:116)
        at okhttp3.RealCall.enqueue(RealCall.java:92)
        at io.opentelemetry.exporter.otlp.http.trace.OtlpHttpSpanExporter.export(OtlpHttpSpanExporter.java:107)
        at io.opentelemetry.sdk.trace.export.BatchSpanProcessor$Worker.exportCurrentBatch(BatchSpanProcessor.java:305)
        at io.opentelemetry.sdk.trace.export.BatchSpanProcessor$Worker.run(BatchSpanProcessor.java:222)
        at java.base/java.lang.Thread.run(Unknown Source)

anuraaga:
Fixes #3813

### DIFF
--- a/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/okhttp/OkHttpExporter.java
+++ b/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/okhttp/OkHttpExporter.java
@@ -94,6 +94,7 @@ public final class OkHttpExporter<T extends Marshaler> {
               @Override
               public void onResponse(Call call, Response response) {
                 if (response.isSuccessful()) {
+                  response.close();
                   exporterMetrics.addSuccess(numItems);
                   result.succeed();
                   return;
--- a/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/okhttp/OkHttpExporter.java
+++ b/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/okhttp/OkHttpExporter.java
@@ -93,27 +93,28 @@ public final class OkHttpExporter<T extends Marshaler> {
 
               @Override
               public void onResponse(Call call, Response response) {
-                if (response.isSuccessful()) {
-                  response.close();
-                  exporterMetrics.addSuccess(numItems);
-                  result.succeed();
-                  return;
+                try (ResponseBody body = response.body()) {
+                  if (response.isSuccessful()) {
+                    exporterMetrics.addSuccess(numItems);
+                    result.succeed();
+                    return;
+                  }
+
+                  exporterMetrics.addFailed(numItems);
+                  int code = response.code();
+
+                  String status = extractErrorStatus(response, body);
+
+                  logger.log(
+                      Level.WARNING,
+                      "Failed to export "
+                          + type
+                          + "s. Server responded with HTTP status code "
+                          + code
+                          + ". Error message: "
+                          + status);
+                  result.fail();
                 }
-
-                exporterMetrics.addFailed(numItems);
-                int code = response.code();
-
-                String status = extractErrorStatus(response);
-
-                logger.log(
-                    Level.WARNING,
-                    "Failed to export "
-                        + type
-                        + "s. Server responded with HTTP status code "
-                        + code
-                        + ". Error message: "
-                        + status);
-                result.fail();
               }
             });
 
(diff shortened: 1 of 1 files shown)


jkwatson (approved):
should we release a patch for this? Seems like a bad bug in the okhttp implementations

carlosalberto (approved):

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/okhttp/OkHttpExporter.java
```diff
@@ -94,6 +94,7 @@ public void onFailure(Call call, IOException e) {
               @Override
               public void onResponse(Call call, Response response) {
                 if (response.isSuccessful()) {
+                  response.close();
```
jack-berg, line 97: We only close the response when `response.isSuccessful()`?

How about we change this to
```
  @Override
  public void onResponse(Call call, Response response) {
    try (ResponseBody responseBody = response.body()) {
      if (response.isSuccessful()) {
        exporterMetrics.addSuccess(numItems);
        result.succeed();
        return;
      }

      exporterMetrics.addFailed(numItems);
      int code = response.code();

      String status = extractErrorStatus(responseBody, response.message());

      logger.log(
          Level.WARNING,
          "Failed to export "
              + type
              + "s. Server responded with HTTP status code "
              + code
              + ". Error message: "
              + status);
      result.fail();
    }
  }
});
```
And `extractErrorStatus` to:
```
private static String extractErrorStatus(
    @Nullable ResponseBody responseBody, String responseMessage) {
  if (responseBody == null) {
    return "Response body missing, HTTP status message: " + responseMessage;
  }
  try {
    return GrpcStatusUtil.getStatusMessage(responseBody.bytes());
  } catch (IOException e) {
    return "Unable to parse response body, HTTP status message: " + responseMessage;
  }
}
```

jkwatson: Yes, agree with @jack-berg . This is more idiomatic okhttp usage.

breedx-splk: Yeah, the autocloseable is nice here. I had to look up/refresh what the behavior is for nullable objects in autocloseable usage, but it's null safe.  Passing that nullable around tho is a bit gross, but the guard was already in `extractErrorStatus()` so not a biggie.

There are some comments in the docs for `close()` about it being an error to close() some types of response, maybe that was the reason? Anyway, those cases (cache response, prior redirect response) don't seem relevant here anyway....so I'd +1 the autocloseable.

breedx-splk (commented):
Bummer that this doesn't come with a test case to ensure `close()` is called. Factoring out the Callback to its own class (even static inner) could help with that I suppose.

anuraaga:
> Bummer that this doesn't come with a test case to ensure close() is called. 

I couldn't come up with a way to test the response is closed, could you describe your idea more?

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/okhttp/OkHttpExporter.java
anuraaga: For reference, the error case was closing implicitly through the call to `.bytes()`. But agree the try resources is clearer

breedx-splk:
> I couldn't come up with a way to test the response is closed, could you describe your idea more?

I was just hoping to fill the coverage gap, is all. Not a big deal...but one thought was that factoring out the anonymous class into a named class could allow it to be instantiated and tested (that's all).

swartzkevinPFG:
The patch provided in for this issue in [#4581](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4581) has resolved this issue.. Thanks for your help.

PR #3815, opened 2021-11-02
Deprecate ExecutorServiceSpanProcessor.

anuraaga:
Removed benchmarks already instead of adding suppresswarnings(deprecation)

@Oberon00 last chance to revive it :)


jkwatson (approved):
I think, if someone really needs this, that it should move over to contrib, since it seems pretty redundant, at least from a functionality perspective.

jack-berg (approved):

PR #3816, opened 2021-11-02
Remove final from methods on final class.

anuraaga:
No big deal but just happened to be looking at the file


jkwatson (approved):
If there's time to lean, there's time to clean! Thanks!

Issue #3807, opened 2021-10-29
Log instrumentation should not have to depend on the log SDK


jack-berg:
As discussed [here](https://github.com/open-telemetry/opentelemetry-java/pull/3759#discussion_r738890665), there is a desire to not require log instrumentation to depend on the log SDK. 

At the same time, there is not currently a plan for a log API, so we must come up with other strategy for organizing artifacts to achieve this.

jack-berg:
Resolved with the addition of [logging appender API](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4917).

PR #3823, opened 2021-11-03
Let clock be overridden for log creators

### LINKED ISSUE #3817: Add ability to configure clock for SdkLogEmitterProvider
`SdkMeterProviderBuilder` and `SdkTracerProviderBuilder` each have a `setClock(Clock clock)` method, while `SdkLogEmitterProviderBuilder` does not.

****Should `SdkLogEmitterProvider` need the ability to customize the clock?****

The logs that are emitted from `LogEmitter` instances can have their timestamps set by the caller. If not set, they'll default to `System.currentTimeMillis()`. Arguably if log timestamps are not set they should default to `clock.now()`. 

On the other hand, the [log proto](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/logs/v1/logs.proto#L117) says that a timestamp field of 0 indicates an unknown or missing timestamp. So arguably, if a caller fails to the timestamp it should default to 0, or `null`.

breedx-splk:
Resolves #3817

Allow the clock to be overridden so that we aren't directly coupled to `System.currentTimeMillis()` and is more consistent with how metrics and traces handle temporality.

It is slightly different due to the `ComponentRegistry` used in `SdkLogEmitterProvider` but I think it's covered.

### DIFF
--- a/sdk/logs/src/main/java/io/opentelemetry/sdk/logs/LogBuilder.java
+++ b/sdk/logs/src/main/java/io/opentelemetry/sdk/logs/LogBuilder.java
@@ -53,7 +53,7 @@ public interface LogBuilder {
   /** Set the attributes. */
   LogBuilder setAttributes(Attributes attributes);
 
-  /** Sets the clock */
+  /** Sets the clock. */
   LogBuilder setClock(Clock clock);
 
   /** Emit the log to downstream {@link LogProcessor}(s). */
--- a/sdk/logs/src/main/java/io/opentelemetry/sdk/logs/LogBuilder.java
+++ b/sdk/logs/src/main/java/io/opentelemetry/sdk/logs/LogBuilder.java
@@ -6,6 +6,7 @@
 package io.opentelemetry.sdk.logs;
 
 import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.sdk.common.Clock;
 import io.opentelemetry.sdk.logs.data.Body;
 import io.opentelemetry.sdk.logs.data.Severity;
 import java.time.Instant;
@@ -51,6 +52,9 @@ public interface LogBuilder {
 
   /** Set the attributes. */
   LogBuilder setAttributes(Attributes attributes);
+
+  /** Sets the clock */
+  LogBuilder setClock(Clock clock);
 
   /** Emit the log to downstream {@link LogProcessor}(s). */
   void emit();
--- a/sdk/logs/src/main/java/io/opentelemetry/sdk/logs/LogBuilder.java
+++ b/sdk/logs/src/main/java/io/opentelemetry/sdk/logs/LogBuilder.java
@@ -6,7 +6,6 @@
 package io.opentelemetry.sdk.logs;
 
 import io.opentelemetry.api.common.Attributes;
-import io.opentelemetry.sdk.common.Clock;
 import io.opentelemetry.sdk.logs.data.Body;
 import io.opentelemetry.sdk.logs.data.Severity;
 import java.time.Instant;
@@ -52,9 +51,6 @@ public interface LogBuilder {
 
   /** Set the attributes. */
   LogBuilder setAttributes(Attributes attributes);
-
-  /** Sets the clock. */
-  LogBuilder setClock(Clock clock);
 
   /** Emit the log to downstream {@link LogProcessor}(s). */
   void emit();
--- a/sdk/logs/src/main/java/io/opentelemetry/sdk/logs/LogEmitterSharedState.java
+++ b/sdk/logs/src/main/java/io/opentelemetry/sdk/logs/LogEmitterSharedState.java
@@ -36,7 +36,7 @@ final class LogEmitterSharedState {
     return logProcessor;
   }
 
-  public Clock getClock() {
+  Clock getClock() {
     return clock;
   }
 
--- a/sdk/logs/src/main/java/io/opentelemetry/sdk/logs/LogEmitterSharedState.java
+++ b/sdk/logs/src/main/java/io/opentelemetry/sdk/logs/LogEmitterSharedState.java
@@ -5,6 +5,7 @@
 
 package io.opentelemetry.sdk.logs;
 
+import io.opentelemetry.sdk.common.Clock;
 import io.opentelemetry.sdk.common.CompletableResultCode;
 import io.opentelemetry.sdk.resources.Resource;
 import java.util.List;
@@ -18,11 +19,13 @@ final class LogEmitterSharedState {
   private final Object lock = new Object();
   private final Resource resource;
   private final LogProcessor logProcessor;
+  private final Clock clock;
   @Nullable private volatile CompletableResultCode shutdownResult = null;
 
-  LogEmitterSharedState(Resource resource, List<LogProcessor> logProcessors) {
+  LogEmitterSharedState(Resource resource, List<LogProcessor> logProcessors, Clock clock) {
     this.resource = resource;
     this.logProcessor = LogProcessor.composite(logProcessors);
+    this.clock = clock;
   }
 
   Resource getResource() {
@@ -31,6 +34,10 @@ final class LogEmitterSharedState {
 
   LogProcessor getLogProcessor() {
     return logProcessor;
+  }
+
+  public Clock getClock() {
+    return clock;
   }
 
   boolean hasBeenShutdown() {
--- a/sdk/logs/src/main/java/io/opentelemetry/sdk/logs/SdkLogEmitter.java
+++ b/sdk/logs/src/main/java/io/opentelemetry/sdk/logs/SdkLogEmitter.java
@@ -6,10 +6,8 @@
 package io.opentelemetry.sdk.logs;
 
 import io.opentelemetry.api.common.Attributes;
-import io.opentelemetry.sdk.common.Clock;
 import io.opentelemetry.sdk.common.InstrumentationLibraryInfo;
 import io.opentelemetry.sdk.logs.data.Body;
-import io.opentelemetry.sdk.logs.data.LogData;
 import io.opentelemetry.sdk.logs.data.LogDataBuilder;
 import io.opentelemetry.sdk.logs.data.Severity;
 import java.time.Instant;
@@ -20,20 +18,17 @@ final class SdkLogEmitter implements LogEmitter {
 
   private final LogEmitterSharedState logEmitterSharedState;
   private final InstrumentationLibraryInfo instrumentationLibraryInfo;
-  private final Clock clock;
 
   SdkLogEmitter(
       LogEmitterSharedState logEmitterSharedState,
-      InstrumentationLibraryInfo instrumentationLibraryInfo,
-      Clock clock) {
+      InstrumentationLibraryInfo instrumentationLibraryInfo) {
     this.logEmitterSharedState = logEmitterSharedState;
     this.instrumentationLibraryInfo = instrumentationLibraryInfo;
-    this.clock = clock;
   }
 
   @Override
   public LogBuilder logBuilder() {
-    return new SdkLogBuilder().setClock(clock);
+    return new SdkLogBuilder();
   }
 
   // VisibleForTesting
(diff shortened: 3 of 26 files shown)


sdk/logs/src/test/java/io/opentelemetry/sdk/logs/SdkLogEmitterTest.java
```diff
@@ -34,15 +35,19 @@
   private static final String INSTRUMENTATION_LIBRARY_NAME = SdkLogEmitter.class.getName();
   private static final String INSTRUMENTATION_LIBRARY_VERSION = "0.0.1";
   private static final String SCHEMA_URL = "http://schemaurl";
+  private static final long NOW = TimeUnit.MILLISECONDS.toNanos(System.currentTimeMillis());
 
   @Mock private LogProcessor logProcessor;
+  @Mock private Clock clock;
```
jkwatson, line 41: do we need a mock clock for this test? It looks like we're always assigning the time explicitly to the logs.

sdk/logs/src/test/java/io/opentelemetry/sdk/logs/SdkLogEmitterProviderTest.java
```diff
@@ -28,15 +30,19 @@
 @MockitoSettings(strictness = Strictness.LENIENT)
 class SdkLogEmitterProviderTest {
 
+  private static final long NOW = TimeUnit.MILLISECONDS.toNanos(System.currentTimeMillis());
   @Mock private LogProcessor logProcessor;
+  @Mock private Clock clock;
```
jkwatson, line 35: It doesn't look like this mock clock is needed for this test.

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/data/LogDataBuilder.java
```diff
@@ -27,6 +28,7 @@
   @Nullable private String severityText;
   @Nullable private String name;
   private Body body = Body.stringBody("");
+  private Clock clock = Clock.getDefault();
```
jkwatson, line 31: Is this the right place to save a copy of the clock? Why not just have the clock in the emitter/provider and have it do the work of defaulting the epoch time on the log data?

sdk/logs/src/test/java/io/opentelemetry/sdk/logs/SdkLogEmitterTest.java
breedx-splk: Ah I didn't realize that. I'll pull it in to a new test method.

sdk/logs/src/test/java/io/opentelemetry/sdk/logs/SdkLogEmitterProviderTest.java
breedx-splk: I thought it would be easy coverage, but I will move it to a dedicated test method.

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/data/LogDataBuilder.java
breedx-splk: I assumed that the zero check and defaulting to "now" was encapsulated in the `LogDataBuilder` for good reason. If we move it up, then it requires all users of the `LogDataBuilder` to go thru the `SLEProviderBuilder` and `SLEProvider` tangle.

Right now, it's possible for users to just say `LogData.builder().set(...).set(...).build()` to get a `LogData` instance. If we move the logic up, then that defaulting gets missed.

I see what you're getting at, though, and it does feel weird to have the clock defaulted in two places (here and in `SdkLogEmitterProviderBuilder`. Not sure what's the right thing to do here.

jkwatson: oh, I see the issue. I forgot the static `LogData.builder()` method didn't interact with the provider/emitter. Hmm. Ugly.  I need to ponder this. The builder is a little weird here in general.

jack-berg: Here's the (ok, "a" solution 😄) solution:

- Remove the `setClock(..)` method from `Logbuilder`.
- When `SdkLogEmitter#logBuilder()` is called, call `logBuilder#setEpoch(..)` with the current epoch according to the clock configured with the `SdkLogEmitterProvider`.
- The user can still override the epoch, but if not, its set to the epoch according to the clock at the time `logBuilder()` was called. And no need for `LogBuilder` to have a clock.

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/LogBuilder.java
```diff
@@ -52,6 +53,9 @@
   /** Set the attributes. */
   LogBuilder setAttributes(Attributes attributes);
 
+  /** Sets the clock */
+  LogBuilder setClock(Clock clock);
```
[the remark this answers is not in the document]
breedx-splk: It's been pulled out of here and the defaulting happens in the `LogDataBuilder`, which has a clock.

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/data/LogData.java
```diff
@@ -19,12 +19,6 @@
 @Immutable
 public interface LogData {
 
-  /** Returns a new {@link LogDataBuilder}. */
-  static LogDataBuilder builder(
```
jack-berg, line 23: What's the motivation for this?

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/SdkLogEmitter.java
```diff
@@ -29,21 +28,28 @@
...
-    return new SdkLogBuilder();
+    return new SdkLogBuilder(logEmitterSharedState, instrumentationLibraryInfo);
   }
 
   // VisibleForTesting
   InstrumentationLibraryInfo getInstrumentationLibraryInfo() {
     return instrumentationLibraryInfo;
   }
 
-  private final class SdkLogBuilder implements LogBuilder {
```
jack-berg, line 40: Besides some extra boilerplate 😁, what do we gain from this?

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/data/LogData.java
breedx-splk: It makes it more consistent with `SpanData` and doesn't couple the `LogData` to a clock impl. We anticipate that the usages are all internal anyway.

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/SdkLogEmitter.java
breedx-splk: In my brain, the implicit sharing of state (instance fields) between an outer and nonstatic inner class is much harder to reason about than just a vanilla class. Alternatively, we could make it a pure anonymous class or pull it up to the top level I suppose....but this seemed like a good compromise.

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/data/LogData.java
jkwatson: Yeah, as Jason said, I don't know if we have any use-cases at this point for non-internal usage of this builder, so let's keep it protected in the implementation until such time as we do.

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/data/LogDataBuilder.java
jkwatson: @breedx-splk and I talked through this option, and he (rightfully) asserted that if we just put the epoch value in at builder creation, then the builder wouldn't be reusable (without manually setting the epoch value on every usage thereafter).

breedx-splk: Yeah, for max accuracy, I think the time should be obtained from the clock at the moment of pojo creation, not builder creation.

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/SdkLogEmitter.java
anuraaga: Slight preference for making a top level class, just to be consistent with span

https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkSpanBuilder.java

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/data/LogDataBuilder.java
```diff
@@ -27,11 +28,26 @@
...
-  LogDataBuilder(Resource resource, InstrumentationLibraryInfo instrumentationLibraryInfo) {
+  private LogDataBuilder(
+      Resource resource, InstrumentationLibraryInfo instrumentationLibraryInfo, Clock clock) {
     this.resource = resource;
     this.instrumentationLibraryInfo = instrumentationLibraryInfo;
+    this.clock = clock;
+  }
+
+  /** Returns a new {@link LogDataBuilder} with the default clock. */
+  public static LogDataBuilder create(
```
anuraaga, line 42: Not for this PR I guess but I think we want a `LogDataBuilder` interface so that we can keep the implementation (and notably this method) private. I believe it's part of #3807 though

anuraaga (approved):

jkwatson (approved):
Thanks!

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/LogEmitterSharedState.java
```diff
@@ -33,6 +36,10 @@ LogProcessor getLogProcessor() {
     return logProcessor;
   }
 
+  public Clock getClock() {
```
jack-berg, line 39: This can be package private.

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/data/LogDataBuilder.java
jack-berg: It still feels wrong to have a clock associated with a record.

Another alternative is to take advantage of the fact that `SdkLogEmitter#SdkLogBuilder` delegates to `LogDataBuilder`, which means that we can track whether `SdkLogBuilder#setEpoch()` has been called with a boolean. If it hasn't been called when `emit()` is invoked, then we can set a default value using the clock.

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/data/LogData.java
jack-berg: Agree with consistency with `SpanData`.

Not sure what you mean by protected though, since the `LogDataBuilder#create()` methods are public and not in an internal package.

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/SdkLogEmitter.java
jack-berg: I like consistency, and think its fine to make this a top level class since its package private so ultimately an implementation detail.

In general though private classes (both static and non-static) are beneficial in that they make it clear to maintainers that class is only intended to be used by the outer class. I don't think we should make it a rule in this project that all classes should be top level, or that inner non-static classes aren't allowed - these language features have their place.

breedx-splk: Ok, thanks for this feedback. I'll move it up for consistency. 👍🏻

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/data/LogData.java
jkwatson: ugh. perhaps it should be in an internal package? Not sure.

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/data/LogDataBuilder.java
breedx-splk: Cool, yeah, follow-up effort I think too.

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/data/LogData.java
jack-berg: The equivalent to this for tracing is `TestSpanData`. That has its own builder, and a lives in a testing package: `io.opentelemetry.sdk.testing.trace`. 

We could do the same for symmetry. 

Part of me also thinks that its unnecessarily hard to create of these `*Data` interfaces. If I want to implement a processor that modifies the `*Data` before delegating to another processor, I have to bring my own implementation or use a class with "Test" in the name (or in the case of tracing, I can subclass `DelegatingSpanData`).

PR #3824, opened 2021-11-03
update the docs for 1.7.1

jkwatson:


docs/apidiffs/1.7.1_vs_1.7.0/opentelemetry-sdk.txt
```diff
@@ -0,0 +1,2 @@
+Comparing source compatibility of  against
```
breedx-splk, line 1: It looks weird that the versions are empty here. Is that normal?

breedx-splk (commented):
Looks good, just had a question about the diff files.

docs/apidiffs/1.7.1_vs_1.7.0/opentelemetry-sdk.txt
jkwatson: yeah.. this is a "feature" of japicmp, unfortunately. I have tried digging to fix it, but it ended up being more trouble than it's worth.

anuraaga (approved):

PR #3827, opened 2021-11-04
Demonstrate async metrics not reseting properly

jack-berg:
I noticed that async metrics aren't properly reseting when reporting to a delta reader. The issue is that accumulations (i.e. unique attributes) that don't appear in the most recent collection still get sent to the reader.

This results in double counting of the last reported value for a unique set of attributes.

### DIFF
--- a/sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/AsyncResetTest.java
+++ b/sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/AsyncResetTest.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.metrics;
+
+import static io.opentelemetry.sdk.testing.assertj.metrics.MetricAssertions.assertThat;
+
+import io.opentelemetry.api.common.Attributes;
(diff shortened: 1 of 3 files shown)


sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/AsyncResetTest.java
```diff
@@ -0,0 +1,65 @@
...
+import io.opentelemetry.api.metrics.ObservableLongMeasurement;
+import io.opentelemetry.sdk.metrics.testing.InMemoryMetricReader;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Consumer;
+import org.junit.jupiter.api.Test;
+
+class AsyncResetTest {
+
+  @Test
+  void demonstrateAsyncResetBug() {
```
jack-berg, line 20: This test duplicates the `TemporalMetricStorage#asynchronous_DeltaResetsAccumulations` test but in a clearer end-to-end fashion. I don't think we should keep this around, but thought it was helpful in illustrating the issue. 

I have an async observer that reports a different value for the attribute key `key` on each collection. The first collection, one metric data point is reported for `key=value1`. But on the second, when only `key=value1` is recorded metric data points are collected for both `key=value1` and `key=value2`.

jack-berg:
PR #3831 includes this change and builds on it. Closing this for now. Can re-open if #3831 is rejected.

PR #3828, opened 2021-11-05
Separate out coverage targets for unmaintained modules.

anuraaga:


.codecov.yaml
```diff
@@ -11,15 +11,17 @@ coverage:
   status:
     project:
       default:
-        enabled: yes
```
anuraaga, line 14: No such field in codecov.yaml

.codecov.yaml
```diff
@@ -11,15 +11,17 @@ coverage:
...
         target: 90%
         paths:
-          - api
```
anuraaga, line 17: I guess the intent when this file was made was to only enforce for api/sdk. But I think this was getting ignored or not handled properly due to the lack of the trailing `/`. Anyways exclusion seems better unless we end up needing even more fine grained control.

jkwatson (approved):

.codecov.yaml
```diff
@@ -11,15 +11,17 @@ coverage:
...
-          - sdk
+          - "!opencensus-shim/"
+          - "!opentracing-shim/"
+          - "!sdk-extensions/zpages/"
+      unmaintained:
+        target: 80%
+        paths:
+          - "opencensus-shim/"
```
jsuereth, line 22: FYI - Myself (and my team) are maintaining this and coordinating releases with OpenCensus.   So far, we haven't done much because we've been focused on OTel MEtrics SDK stabilization before continuing here.

anuraaga: #3829 Actions >> words

PR #3830, opened 2021-11-05
Retain accumulation history when reporting cumulative aggregation for async instruments

jack-berg:
If I'm exporting cumulative metrics, and have an async instrument that records one set of attributes on the first collection:
```
...
observableLongMeasurement.observe(10, Attributes.builder().put("key", "value1").build())
```
.. then on the next collection records another set of attributes:
```
...
observableLongMeasurement.observe(10, Attributes.builder().put("key", "value2").build())
```
Then I expect that the second export contains data points for both sets of attributes. Accumulations should be retained even if they don't appear in the most recent collection.

Currently, they do not. For async instruments and cumulative temporality, only the accumulations in the current collection appear.

This PR fixes that.

### DIFF
--- a/sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/AsyncResetTest.java
+++ b/sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/AsyncResetTest.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.metrics;
+
+import static io.opentelemetry.sdk.testing.assertj.metrics.MetricAssertions.assertThat;
+
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.metrics.ObservableLongMeasurement;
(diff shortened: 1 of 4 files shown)


sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/AsyncResetTest.java
```diff
@@ -0,0 +1,69 @@
...
+import io.opentelemetry.api.metrics.ObservableLongMeasurement;
+import io.opentelemetry.sdk.metrics.testing.InMemoryMetricReader;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Consumer;
+import org.junit.jupiter.api.Test;
+
+class AsyncResetTest {
+
+  @Test
+  void demonstrateAsyncCumulativeRetentionBug() {
```
jack-berg, line 20: This test is redundant to `TemporalMetricStorageTest#asynchronous_CumulativeRetainsHistory`, but I think it helps illustrate the point better.

I think we should remove it before merging.

jack-berg:
Discussed with @jsuereth - this is desired behavior. When async instrument stops reporting a particular stream, it should stop being collected.

PR #3831, opened 2021-11-05
Metric cardinality limits

jack-berg:
This PR adds some naive protections for metrics cardinality:

- It aggressively forgets metric streams when temporality is cumulative. Currently, if temporality is cumulative and async instruments stop recording for a metric stream for an export, that stream is removed. The PR changes sync instruments to have the same behavior. This is the most conservative thing we can do from a memory management perspective. In the past we've talked about having TTL, where if a stream isn't recorded for the TTL, it's removed. This is a more aggressive, and simpler to implement removal policy that I think is a good start.
- It caps the number of metric streams for each `MetricStorage`. There is one metric storage per view, so each view has a cap number of metric streams. I set the number to 2000, which I believe is consistent with the behavior of dotnet. Arguably this should be much lower. It should also (eventually) be configurable. 
- It makes the `SdkMeterProvider` minimum collection interval configurable. This was an outstanding TODO in the code. Allowing this to be set to 0 made for simpler tests.

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProviderBuilder.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProviderBuilder.java
@@ -5,6 +5,8 @@
 
 package io.opentelemetry.sdk.metrics;
 
+import static io.opentelemetry.api.internal.Utils.checkArgument;
+
 import io.opentelemetry.api.metrics.GlobalMeterProvider;
 import io.opentelemetry.sdk.common.Clock;
 import io.opentelemetry.sdk.metrics.exemplar.ExemplarFilter;
@@ -15,9 +17,11 @@ import io.opentelemetry.sdk.metrics.internal.view.ViewRegistryBuilder;
 import io.opentelemetry.sdk.metrics.view.InstrumentSelector;
 import io.opentelemetry.sdk.metrics.view.View;
 import io.opentelemetry.sdk.resources.Resource;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
+import java.util.concurrent.TimeUnit;
 
 /**
  * Builder class for the {@link SdkMeterProvider}. Has fully functional default implementations of
@@ -31,6 +35,7 @@ public final class SdkMeterProviderBuilder {
   private final List<MetricReaderFactory> metricReaders = new ArrayList<>();
   // Default the sampling strategy.
   private ExemplarFilter exemplarFilter = ExemplarFilter.sampleWithTraces();
+  private long minimumCollectionIntervalNanos = TimeUnit.MILLISECONDS.toNanos(100);
 
   SdkMeterProviderBuilder() {}
 
@@ -124,6 +129,20 @@ public final class SdkMeterProviderBuilder {
   }
 
   /**
+   * Configure the minimum duration between synchronous collections. If collections occur more
+   * frequently than this, synchronous collection will be suppressed.
+   *
+   * @param duration The duration.
+   * @return this
+   */
+  public SdkMeterProviderBuilder setMinimumCollectionInterval(Duration duration) {
+    Objects.requireNonNull(duration, "duration");
+    checkArgument(!duration.isNegative(), "duration must not be negative");
+    minimumCollectionIntervalNanos = duration.toNanos();
+    return this;
+  }
+
+  /**
    * Returns a new {@link SdkMeterProvider} built with the configuration of this {@link
    * SdkMeterProviderBuilder}. This provider is not registered as the global {@link
    * io.opentelemetry.api.metrics.MeterProvider}. It is recommended that you register one provider
@@ -135,6 +154,11 @@ public final class SdkMeterProviderBuilder {
    */
   public SdkMeterProvider build() {
     return new SdkMeterProvider(
-        metricReaders, clock, resource, viewRegistryBuilder.build(), exemplarFilter);
+        metricReaders,
+        clock,
+        resource,
+        viewRegistryBuilder.build(),
+        exemplarFilter,
+        minimumCollectionIntervalNanos);
   }
 }
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
@@ -162,6 +162,10 @@ public final class AsynchronousMetricStorage<T> implements MetricStorage {
     private Map<Attributes, T> currentAccumulation = new HashMap<>();
 
     public void record(Attributes attributes, T accumulation) {
+      if (currentAccumulation.size() >= MAX_ACCUMULATIONS) {
+        // TODO: provide useful diagnostics when max accumulations has been reached
+        return;
+      }
       // TODO: error on metric overwrites
       currentAccumulation.put(attributes, accumulation);
     }
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
@@ -10,6 +10,7 @@ import io.opentelemetry.api.metrics.ObservableDoubleMeasurement;
 import io.opentelemetry.api.metrics.ObservableLongMeasurement;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.sdk.common.InstrumentationLibraryInfo;
+import io.opentelemetry.sdk.internal.ThrottlingLogger;
 import io.opentelemetry.sdk.metrics.data.AggregationTemporality;
 import io.opentelemetry.sdk.metrics.data.MetricData;
 import io.opentelemetry.sdk.metrics.exemplar.ExemplarFilter;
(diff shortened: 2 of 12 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtils.java
```diff
@@ -20,6 +20,7 @@ private MetricStorageUtils() {}
    */
   static <T> void mergeInPlace(
       Map<Attributes, T> result, Map<Attributes, T> toMerge, Aggregator<T> aggregator) {
+    result.entrySet().removeIf(entry -> !toMerge.containsKey(entry.getKey()));
```
breedx-splk, line 23: This feels off to me. 

Is there supposed to be some hard guarantee that what we're merging in here should always have the same keyset?  This change feels like a different operation now -- more like `combineCommon` or something.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProviderBuilder.java
```diff
@@ -123,6 +128,20 @@ public SdkMeterProviderBuilder registerMetricReader(MetricReaderFactory reader)
...
+  /**
+   * Configure the minimum duration between synchronous collections. If collections occur more
+   * frequently than this, synchronous collection will be suppressed.
+   *
+   * @param duration The duration.
+   * @return this
+   */
+  public SdkMeterProviderBuilder setMinimumCollectionInterval(Duration duration) {
+    Objects.requireNonNull(duration, "duration");
+    checkArgument(!duration.isNegative(), "duration must not be negative");
```
breedx-splk, line 140: 👍🏻 nice and thorough

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DeltaMetricStorage.java
```diff
@@ -47,9 +53,13 @@ public BoundStorageHandle bind(Attributes attributes) {
       return aggregatorHandle;
     }
 
-    // Missing entry or no longer mapped, try to add a new entry.
+    // Missing entry or no longer mapped. Try to add a new one if not exceeded cardinality limits.
     aggregatorHandle = aggregator.createHandle();
     while (true) {
+      if (activeCollectionStorage.size() >= MAX_ACCUMULATIONS) {
+        // TODO: provide useful diagnostics when max accumulations has been reached
+        return NOOP_STORAGE_HANDLE;
```
breedx-splk, line 61: I def understand not wanting to fill up the logs when this condition is hit, but I do feel like especially while we're in alpha (and probably even later) it's important to have SOME indication that data is being dropped. How about logging max once a minute or maybe just a one-shot?

breedx-splk (changes requested):
Looks good to me. Two small thoughts to maybe address. Thanks!

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorage.java
```diff
@@ -20,6 +20,9 @@
  */
 public interface MetricStorage {
 
+  /** The max number of metric accumulations for a particular {@link MetricStorage}. */
+  Integer MAX_ACCUMULATIONS = 2000;
```
anuraaga, line 24: This should be private somewhere not public for end users as when added to an interface.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DeltaMetricStorage.java
```diff
@@ -47,9 +53,13 @@ public BoundStorageHandle bind(Attributes attributes) {
       return aggregatorHandle;
     }
 
-    // Missing entry or no longer mapped, try to add a new entry.
+    // Missing entry or no longer mapped. Try to add a new one if not exceeded cardinality limits.
     aggregatorHandle = aggregator.createHandle();
     while (true) {
+      if (activeCollectionStorage.size() >= MAX_ACCUMULATIONS) {
```
anuraaga, line 59: To confirm, we've added a cap without adding any LRU behavior. Seems fine to me.

jack-berg: Yes. Additionally, for synchronous instruments exporting as cumulative, metric streams are "forgotten" when they don't record during a collection.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DeltaMetricStorage.java
jack-berg: Yeah I think it's important too. I like logging once when the limit is exceeded, or perhaps every 10th time? I.e. something like:
```
AtomicLong exceedsMaxAccumulationsCount = new AtomicLong();
...
if (activeCollectionStorage.size() >= MAX_ACCUMULATIONS) {
       if (exceedsMaxAccumulationsCount.getAndIncrement() % 10 == 0) {
          logger.log("Max accumulations exceeded for instrument ....");
       }
       return NOOP_STORAGE_HANDLE;
}
```

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtils.java
jack-berg: This is actually part of what makes this simple solution possible. 

Currently, `DeltaMetricStorage` stores accumulations for synchronous instruments. It stores them as deltas, has no cap on the number of accumulations, and resets accumulations after each collection. When a cumulative collection happens, the delta accumulations are merged in the previous `Map<Attribute, T> result`.

So enforcing a max number of accumulations in `DeltaMetricStorage` alone doesn't actually bound memory usage - it only adds a bound for the number of accumulations in a single collection. This code forgets accumulations that weren't recorded in the most recent collection, and therefore allows the max accumulation enforcement in `DeltaMetricStorage` actually be the overall bound, not just the bound for a single collection. 

Arguably though, this code should be lifted up to `MetricStorageUtils`, so that this function have a single responsibility.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
```diff
@@ -162,6 +162,10 @@ public MetricDescriptor getMetricDescriptor() {
     private Map<Attributes, T> currentAccumulation = new HashMap<>();
 
     public void record(Attributes attributes, T accumulation) {
+      if (currentAccumulation.size() >= MAX_ACCUMULATIONS) {
+        // TODO: provide useful diagnostics when max accumulations has been reached
```
jsuereth, line 166: Could you add a throttling logger here for that?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DeltaMetricStorage.java
```diff
@@ -47,9 +53,13 @@ public BoundStorageHandle bind(Attributes attributes) {
...
+        // TODO: provide useful diagnostics when max accumulations has been reached
```
jsuereth, line 60: Can you add a throttling logger here for that?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DeltaMetricStorage.java
jsuereth: @jkwatson has a ThrottlingLogger you can use here.  Please do!

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorage.java
jsuereth: Also why is this `Integer` ?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtils.java
jsuereth: I think this is a nice elegant solution.  Please update the javadoc to denote that you're limiting the output keyset to match the `toMerge` keyset.

I'd also be ok if you made a different function `mergeAndLimitKeysInPlace()` or some such.   Since these are just helpers for our own usage, I think it's find to have it here.


Same comment for `diffInPlace`

jsuereth (commented):
Looks great!   Few minor nit-picks and an ask to use the ThrottlingLogger (think we use it elsewhere in the metrics SDK already).

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorage.java
jack-berg: > Also why is this Integer

Ah yes, can be `int`.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtils.java
jack-berg: Updated the javadoc, but left the method names the same. 👍

anuraaga (approved):

jsuereth (approved):

jkwatson:
@jack-berg needs a quick rebase then ready to merge.

PR #3833, opened 2021-11-05
Handle multiple hyphens in container prefix when parsing cgroups

### LINKED ISSUE #3832: Container detection fails in some kubernetes environments
**Describe the bug**
On cgroups with entries with the container format:
```11:perf_event:/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod4415fd05_2c0f_4533_909b_f2180dca8d7c.slice/cri-containerd-713a77a26fe2a38ebebd5709604a048c3d380db1eb16aa43aca0b2499e54733c.scope``` containerId will return as null.

**Steps to reproduce**
Any cgroup where the prefix has 2 dashes.

ContainerId should be ```713a77a26fe2a38ebebd5709604a048c3d380db1eb16aa43aca0b2499e54733c```

**What did you see instead?**
Returns as ```null```

v1.6.0

**Environment**
JDK11 but can be any
k8s version unknown.

**Additional context**
The issue arises because ContainerResource does not expect a second '-' in the prefix.

breedx-splk:
resolves #3832

### DIFF
--- a/sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
+++ b/sdk-extensions/resources/src/main/java/io/opentelemetry/sdk/extension/resources/ContainerResource.java
@@ -67,7 +67,7 @@ public final class ContainerResource {
       Optional<String> value =
           lines
               .filter(line -> !line.isEmpty())
-              .map(line -> getIdFromLine(line))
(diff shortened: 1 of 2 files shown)


lo-jason (approved):
looks good to me

jkwatson (approved):
Thanks!

PR #3834, opened 2021-11-05
Backfill test coverage for MetricStorageUtils

breedx-splk:
While reviewing #3831 I was thinking about the `MetricStorageUtils` implementation and couldn't find any tests so I made some.

### DIFF
--- a/sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtilsTest.java
+++ b/sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtilsTest.java
@@ -30,7 +30,7 @@ class MetricStorageUtilsTest {
   private Map<Attributes, String> toMerge;
 
   @BeforeEach
-  private void setup() {
+  public void setup() {
     result = new HashMap<>();
     result.put(a, "A");
     result.put(b, "B");
--- a/sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtilsTest.java
+++ b/sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtilsTest.java
@@ -57,7 +57,7 @@ class MetricStorageUtilsTest {
   }
 
   @Test
-  void diff() {
+  void diffInPlace() {
     Aggregator<String> agg = new TestAggregator();
     MetricStorageUtils.diffInPlace(result, toMerge, agg);
 
--- a/sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtilsTest.java
+++ b/sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtilsTest.java
@@ -21,16 +21,16 @@ import org.junit.jupiter.api.Test;
 
 class MetricStorageUtilsTest {
 
-  static final Attributes a = Attributes.of(AttributeKey.stringKey("a"), "a");
-  static final Attributes b = Attributes.of(AttributeKey.stringKey("b"), "b");
-  static final Attributes c = Attributes.of(AttributeKey.stringKey("c"), "c");
-  static final Attributes d = Attributes.of(AttributeKey.stringKey("d"), "d");
-  static final Attributes e = Attributes.of(AttributeKey.stringKey("e"), "e");
-  Map<Attributes, String> result;
-  Map<Attributes, String> toMerge;
+  private static final Attributes a = Attributes.of(AttributeKey.stringKey("a"), "a");
+  private static final Attributes b = Attributes.of(AttributeKey.stringKey("b"), "b");
+  private static final Attributes c = Attributes.of(AttributeKey.stringKey("c"), "c");
+  private static final Attributes d = Attributes.of(AttributeKey.stringKey("d"), "d");
+  private static final Attributes e = Attributes.of(AttributeKey.stringKey("e"), "e");
+  private Map<Attributes, String> result;
+  private Map<Attributes, String> toMerge;
 
   @BeforeEach
-  void setup() {
+  private void setup() {
     result = new HashMap<>();
     result.put(a, "A");
     result.put(b, "B");
@@ -60,7 +60,7 @@ class MetricStorageUtilsTest {
   }
 
   @SuppressWarnings("unchecked")
-  private Aggregator<String> buildConcatAggregator() {
+  private static Aggregator<String> buildConcatAggregator() {
     Aggregator<String> agg = mock(Aggregator.class);
     doAnswer(
             invocation -> {
--- a/sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtilsTest.java
+++ b/sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtilsTest.java
@@ -46,8 +46,8 @@ class MetricStorageUtilsTest {
     Aggregator<String> agg = buildConcatAggregator();
     MetricStorageUtils.mergeInPlace(result, toMerge, agg);
 
-    assertThat(result).containsOnly(entry(a, "A"), entry(b, "BB'"), entry(c, "C"), entry(d, null));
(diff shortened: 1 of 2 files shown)


sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtilsTest.java
```diff
@@ -0,0 +1,100 @@
...
+    toMerge.put(c, "C");
+    toMerge.put(e, null);
+  }
+
+  @Test
+  void mergeInPlace() {
+    Aggregator<String> agg = new TestAggregator();
+    MetricStorageUtils.mergeInPlace(result, toMerge, agg);
+
+    assertThat(result.keySet()).containsExactlyInAnyOrder(a, b, c, d);
```
anuraaga, line 52: We can use a map assertion

```
assertThat(result).containsOnly(
  entry (a, "a"),
  entry (b, "bb")
...
```

anuraaga (approved):

jkwatson (approved):
Thanks!

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtilsTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+class MetricStorageUtilsTest {
+
+  static final Attributes a = Attributes.of(AttributeKey.stringKey("a"), "a");
+  static final Attributes b = Attributes.of(AttributeKey.stringKey("b"), "b");
+  static final Attributes c = Attributes.of(AttributeKey.stringKey("c"), "c");
+  static final Attributes d = Attributes.of(AttributeKey.stringKey("d"), "d");
+  static final Attributes e = Attributes.of(AttributeKey.stringKey("e"), "e");
+  Map<Attributes, String> result;
+  Map<Attributes, String> toMerge;
```
jack-berg, line 33: #nit: All these can be private.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtilsTest.java
```diff
@@ -0,0 +1,100 @@
...
+    Aggregator<String> agg = new TestAggregator();
+    MetricStorageUtils.diffInPlace(result, toMerge, agg);
+
+    assertThat(result.keySet()).containsExactlyInAnyOrder(a, c, d);
+    assertThat(result.get(a)).isEqualTo("A");
+    assertThat(result.get(c)).isEqualTo("C");
+    assertThat(result.get(d)).isNull();
+  }
+
+  private static class TestAggregator implements Aggregator<String> {
```
jack-berg, line 70: #nit: This could be accomplished in fewer lines using a mock. No strong preference though 🤷 .

jack-berg (approved):

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtilsTest.java
```diff
@@ -0,0 +1,100 @@
...
+    assertThat(result.get(a)).isEqualTo("A");
+    assertThat(result.get(b)).isEqualTo("BB'");
+    assertThat(result.get(c)).isEqualTo("C");
+    assertThat(result.get(d)).isNull();
+  }
+
+  @Test
+  void diffInPlace() {
+    Aggregator<String> agg = new TestAggregator();
+    MetricStorageUtils.diffInPlace(result, toMerge, agg);
```
jsuereth, line 62: I'm a bit confused by this test.  IIUC - you're using the presence of a `null` accumulation to denote that one of the values is gone?

That's not really what this is meant to do.  Accumulations (for well-behaved accumulators) should never be returning `null`.  `Diff` and `Merge` should just be `-` and `+`, in fact just using a SumAggregator directly might make this test easier to read/understand.

Given @anuraaga's push for using Nullaway, I'd prefer we don't encode tests that use `null` return values in Aggregator, as I think we need to go clean that up.

jsuereth (commented):
So I think I'd like to see two things in these tests:

1. Use SumAggregator *or* an aggregator that does not return `null`.
2. The test should not rely on the `null` behavior of `compute`, that's an implementation detail.   These maps should NOT contain `null` values in them, the only `null` is when `compute` is asking for a value where there wasn't a previous value.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtilsTest.java
breedx-splk: My preference for tests is to just omit access specifiers for brevity and clarity...but I can change it if folks generally think there's something wrong with that.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtilsTest.java
breedx-splk: Thanks! Every time I learn more about assertj the more I like it.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtilsTest.java
jkwatson: We tend to keep things private in this repo, so yes, private would be good here.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtilsTest.java
jkwatson: would testing for `result.containsKey(xxx)` being false be a better assertion?

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtilsTest.java
breedx-splk: I tried it out, agreed it's better. Thanks for the suggestion.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtilsTest.java
breedx-splk: Whoops, it was not my intention to have `diff()` return null in that mock -- I think I just overlooked it. Thanks for catching that. Let me fix that up.

breedx-splk: Sorry that it was probably confusing with that diff returning null....please have another look with the current form that uses a mock and is more explicit.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtilsTest.java
```diff
@@ -0,0 +1,83 @@
...
+  private static final Attributes e = Attributes.of(AttributeKey.stringKey("e"), "e");
+  private Map<Attributes, String> result;
+  private Map<Attributes, String> toMerge;
+
+  @BeforeEach
+  private void setup() {
+    result = new HashMap<>();
+    result.put(a, "A");
+    result.put(b, "B");
+    result.put(d, null);
```
jsuereth, line 37: We shouldn't allow `null` accumulations.

jsuereth (approved):
One minor nit around having `null` in here, otherwise LGTM

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtilsTest.java
breedx-splk: There's nothing to prevent a user from passing in a map with null values, though, and because there definitely is code that checks for null values, I think it makes sense to have coverage in that area.

PR #3835, opened 2021-11-06
Expand test coverage of OpenCensus-Shim by testing metrics integrations.

jsuereth:
- Refresh the metric adapting code to account for all OpenCensus points.
- Move code into an easier-to-unit-test location
- Deprecate previous Exporter -> Exporter adaption process, favoring hooking the `MetricReaderFactory` as a better (available) option.
- Add a lot more tests.
- Update documentation based on new tests.

### Things I did not do, but wish were easier

- I'd like to have a hook in the autoconfigure module that detects the availability of this class *and* setting an environment flag, which would automatically attach OpenCensus metrics to any registered metric reader (or some such to-be-specified behavior).
- I'd love if i could access the `Resource` off the SDK in some fashion.   We *could* pass it down the `MetricReaderFactory` interface, but looking for opinions there.
- I did *not* refresh / move the tracing testing nor did I get a trace + metrics test working.   That would be a pending PR.


opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/metrics/MetricAdapter.java
```diff
@@ -0,0 +1,363 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.regex.MatchResult;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import javax.annotation.Nullable;
+
+/** Adapts an OpenCensus metric into the OpenTelemetry metric data API. */
+public final class MetricAdapter {
```
anuraaga, line 48: Move to internal if can't make it package private

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/metrics/MetricAdapter.java
```diff
@@ -0,0 +1,363 @@
...
+      InstrumentationLibraryInfo.create("io.opentelemetry.opencensusshim", null);
+
+  /**
+   * Converts an open-census metric into the OTLP format.
+   *
+   * @param otelResource The resource associated with the opentelemetry SDK.
+   * @param censusMetric The OpenCensus metric to convert.
+   */
+  public static MetricData convert(Resource otelResource, Metric censusMetric) {
+    // Note: we can't just adapt interfaces, we need to do full copy because OTel data API uses
```
anuraaga, line 62: We changed it to an interface, is there something that still needs to be switched?

https://github.com/open-telemetry/opentelemetry-java/pull/3783/files

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/metrics/MetricAdapter.java
```diff
@@ -0,0 +1,363 @@
...
+                        DoubleHistogramPointData.create(
+                            // Report Gauge histograms as DELTA with "instantaneous" time window.
+                            isGauge ? endTimestamp : startTimestamp,
+                            endTimestamp,
+                            attributes,
+                            distribution.getSum(),
+                            mapBoundaries(distribution.getBucketOptions()),
+                            mapCounts(distribution.getBuckets()),
+                            mapExemplars(distribution.getBuckets())),
+                    sumamry -> null,
```
anuraaga, line 221:
```suggestion
                    summary -> null,
```

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/metrics/MetricAdapter.java
```diff
@@ -0,0 +1,363 @@
...
+    return result;
+  }
+
+  private static ExemplarData mapExemplar(Exemplar exemplar) {
+    // Look for trace/span id.
+    String spanId = null;
+    String traceId = null;
+    if (exemplar.getAttachments().containsKey("SpanContext")) {
+      // We need to use `io.opencensus.contrib.exemplar.util.AttachmentValueSpanContext`
+      // The `toString` will be the following:
```
anuraaga, line 326: @bogdandrutu ;)

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/metrics/MetricAdapter.java
```diff
@@ -0,0 +1,363 @@
...
+      // SpanContext(traceId={traceId}, spanId={spanId}, traceOptions={traceOptions})
+      // We *attempt* parse it rather than pull in yet another dependency.
+      String spanContextToString = exemplar.getAttachments().get("SpanContext").getValue();
+      Matcher m =
+          Pattern.compile("SpanContext\\(traceId=([0-9A-Ga-g]+), spanId=([0-9A-Ga-g]+),.*\\)")
```
anuraaga, line 331: Make private static final constant for regex

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/metrics/MultiMetricProducer.java
```diff
@@ -0,0 +1,30 @@
...
+
+package io.opentelemetry.opencensusshim.metrics;
+
+import io.opentelemetry.sdk.metrics.data.MetricData;
+import io.opentelemetry.sdk.metrics.export.MetricProducer;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+/** Class that wraps multiple metric producers into one. */
```
anuraaga, line 14: We have a similar class for exporter, should this be in the SDK as an official class? Or only has value for metrics shim since otherwise only the OTel SDK produces metrics?

opencensus-shim/src/test/java/io/opentelemetry/opencensusshim/OpenTelemetryMetricExporterTest.java
```diff
@@ -0,0 +1,227 @@
...
+                      .hasValue(50)
+                      .attributes()
+                      .hasSize(1)
+                      .containsEntry(tagKey.getName(), tagValue.asString()));
+    } finally {
+      otelExporter.stop();
+    }
+  }
+
+  // Straight copy-paste from PeriodicMetricReaderTest.  Should likely move into metrics-testing.
```
anuraaga, line 164: I think our standard pattern can be InMemory + awaitility - any reason this class would also be needed?

opencensus-shim/src/test/java/io/opentelemetry/opencensusshim/metrics/OpenCensusMetricProducerTest.java
```diff
@@ -0,0 +1,68 @@
...
+        View.create(
+            View.Name.create("task_latency_distribution"),
+            "The distribution of the task latencies.",
+            LATENCY_MS,
+            Aggregation.Distribution.create(LATENCY_BOUNDARIES),
+            Collections.emptyList());
+    Stats.getViewManager().registerView(view);
+    STATS_RECORDER.newMeasureMap().put(LATENCY_MS, 50).record();
+    // Wait for measurement to hit the aggregator.
+    Thread.sleep(1000);
```
anuraaga, line 47: Replace Thread.sleep with awaitility (two locations I think)

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/metrics/MetricAdapter.java
jsuereth: It's an unimplementable interface because all of the methods return AutoValues that you can't implement.

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/metrics/MultiMetricProducer.java
jsuereth: Which class are you talking about?   I think I use the composite pattern in a few places, but not this one just yet.

Regarding longer term plans, post SDK stable launch, I hope to expose (via specification) a hook on SdkMeterProvider to register something like this for adapting external metric systems.  For now, I just wanted better code coverage and there was a quick win for a nicer hook with the new metrics SDK.  (Note: Whatever we do here + in Go will drive the [OpenCensus compatiblity specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/opencensus.md#opencensus-compatibility)

opencensus-shim/src/test/java/io/opentelemetry/opencensusshim/OpenTelemetryMetricExporterTest.java
jsuereth: No, just need to re-work some numbers, but that's fine by me. I actually couldn't remember the name of the await utiltiy....

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/metrics/MetricAdapter.java
jsuereth: Done.

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/metrics/MetricAdapter.java
jsuereth: Fixed

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/metrics/MetricAdapter.java
jsuereth: done

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/metrics/MetricAdapter.java
jsuereth: As an FYI - I created a test directly against the exemplar-utils library, but still did not take a dependency on it in the shim.

anuraaga (approved):

opencensus-shim/build.gradle.kts
```diff
@@ -21,4 +21,12 @@ dependencies {
 
   testImplementation("org.slf4j:slf4j-simple")
   testImplementation("io.opencensus:opencensus-impl")
+  testImplementation("io.opencensus:opencensus-contrib-exemplar-util")
+}
+
+tasks.named<Test>("test") {
+  // We must force a fork per-test class because OpenCensus pollutes globals with no restorative
```
anuraaga, line 28: Is it possible to split test sets to work around this? forkEvery(1) goes sloooow

jsuereth: Only for some of the tests.  I've tried to limit what's there to just the ones we need, but for some of the coverage I do needs to actually set up OpenCensus.

OpenCensus itself tended to use heavy mocks/mocking for various metric things.  Really appreciate in OTel the global-cleaning utiities.

anuraaga: Yeah if we can identify those tests that need it and split them into test suites at least it doesn't have to fork for every single one.

jkwatson:
> I'd like to have a hook in the autoconfigure module that detects the availability of this class and setting an environment flag, which would automatically attach OpenCensus metrics to any registered metric reader (or some such to-be-specified behavior).

I think this is something that should go through the spec or something official, yes? Something like "how to enable the OpenCensus integration" so it's at least similar across languages?

jsuereth:
@jkwatson 

> I think this is something that should go through the spec or something official, yes? Something like "how to enable the OpenCensus integration" so it's at least similar across languages?

Yes, that's the plan.   However, I'd like to get an implementation *ahead* of the spec to justify what will work across features.    Also looking to make similar changes on the Go side over time (the current prototypes the OpenCensus spec is based on).

PR #3837, opened 2021-11-08
Rework log fields

jack-berg:
Solves several of the issues in #3804. Replaces #3811.

This PR includes a couple of changes to log fields, based on conversations in the 11/4 SIG:
- `setTraceId(..)`, `setSpanId(..)`, and `setFlags(..)` are replaced with `setContext(Context)`
- `getTraceId()`, `getSpanId()` and `getFlags()` are replaced with `SpanContext getSpanContext()`
- `setAttributes(..)` now replaces attributes instead of merging them. This is less surprising.
- `setBody(Body)` is removed. Only `setBody(String)` is currently supported. On the read side `Body getBody()` still returns a `Body`.
- The nullability of fields has been been addressed. Generally, null is avoid if the response has an alternativel
  - `SpanContext getSpanContext()` returns `SpanContext.getInvalid()` if unset.
  - `Severity getSeverity()` returns `Severity#UNDEFINED_SEVERITY_NUMBER` if unset.
  - `String getSeverityText()` is nullable and returns null if unset.
  - `String getName()` is nullable and returns null if unset.
  - `Body getBody()` returns `Body#emptyBody()` if unset.
  - `Attributes getAttributes()` returns `Attributes#empty()` if unset.


sdk/logs/src/main/java/io/opentelemetry/sdk/logs/data/Body.java
```diff
@@ -31,8 +32,18 @@ static Body stringBody(String stringValue) {
     return StringBody.create(stringValue);
   }
 
+  /**
+   * Return an empty {@code Body}.
+   *
+   * @return a {@code Body} without a value.
+   */
+  static Body emptyBody() {
```
anuraaga, line 40:
```suggestion
  static Body empty() {
```

anuraaga: Some APIs may repeat the type name in factories for static import but I guess we haven't been

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/data/EmptyBody.java
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.logs.data;
+
+class EmptyBody implements Body {
+
+  static final EmptyBody INSTANCE = new EmptyBody();
```
anuraaga, line 10: Can use enum singleton pattern too

anuraaga (approved):
Looks much nicer thanks!

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/data/Body.java
jack-berg: I like this better. I think we can change `stringBody(String)` to just `string(String)` to match this.

anuraaga (approved):

jkwatson (approved):

PR #3843, opened 2021-11-09
A few optimizations and documentation fixes

reugn:


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/export/CollectionHandle.java
```diff
@@ -37,15 +38,13 @@ private CollectionHandle(int index) {
...
+   * Construct a new (mutable) set consisting of the passed in collection handles.
    *
    * <p>Used by tests.
    */
   static Set<CollectionHandle> of(CollectionHandle... handles) {
     Set<CollectionHandle> result = mutableSet();
-    for (CollectionHandle handle : handles) {
-      result.add(handle);
-    }
+    result.addAll(Arrays.asList(handles));
```
anuraaga, line 47: Can you revert this one? Doesn't seem to be an obvious improvement

anuraaga (commented):
Thanks for the cleanups

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/DefaultAggregation.java
```diff
@@ -12,7 +12,7 @@
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-/** Aggregation that selects the specified default based on insturment. */
+/** Aggregation that selects the specified default based on instrument. */
```
anuraaga, line 15: Actually it would be good to separate PRs for typo fixes and actual logic changes if that's OK

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/export/CollectionHandle.java
reugn: @anuraaga reverted.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/DefaultAggregation.java
anuraaga: Thanks @reugn - as this is small let me go ahead and merge, in the future would be good to split cosmetic and logic changes.

anuraaga (approved):

PR #3845, opened 2021-11-09
Enable nullaway for metrics SDK

### LINKED ISSUE #1768: Consider compile-time checking of @Nullable and @Nonnull

anuraaga:
Had left this off during metrics SDK churn but given #3844 and that things should have settled down somewhat, I think it's a good time to enable this

Fixes #1768


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtils.java
```diff
@@ -19,7 +20,12 @@ private MetricStorageUtils() {}
    * <p>Note: This mutates the result map.
    */
   static <T> void mergeInPlace(
-      Map<Attributes, T> result, Map<Attributes, T> toMerge, Aggregator<T> aggregator) {
+      @Nullable Map<Attributes, T> result,
```
anuraaga, line 23: @jsuereth Is `Collections.emptyMap()` (so making `getLastAccumulation` not-nullable) an alternative? Or does null and empty mean something different?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/Aggregator.java
```diff
@@ -50,8 +50,10 @@
    * instruments to create {@code Accumulation} that are passed to the processor.
    *
    * @param value the given value to be used to create the {@code Accumulation}.
-   * @return a new {@code Accumulation} for the given value.
+   * @return a new {@code Accumulation} for the given value, or {@code null} if there are no
```
anuraaga, line 53: I'm assuming there's no "empty value" we can use instead?

jsuereth: This shouldn't ever be returning `null` AFAIK.  I believe the only use case for `null` here is the `NoAggregator` / default.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
```diff
@@ -66,9 +65,12 @@ public static MetricStorage empty() {
         new ObservableDoubleMeasurement() {
           @Override
           public void observe(double value, Attributes attributes) {
-            measurementAccumulator.record(
-                attributesProcessor.process(attributes, Context.current()),
-                aggregator.accumulateDoubleMeasurement(value, attributes, Context.current()));
+            T accumulation =
+                aggregator.accumulateDoubleMeasurement(value, attributes, Context.current());
+            if (accumulation != null) {
```
jsuereth, line 70: Accumulation should not be `null`, and we should enforce that.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtils.java
jsuereth: `result` MUST be a writable map.  I'm also curious where the `null` is coming from here, probably poor checking upstream.

jsuereth (commented):
Looks like there's some upstream null check missing.  Most of the methods you're adding nullable to should NOT allow nullable.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtils.java
anuraaga: Looks like #3831 fixed this, sweet

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
anuraaga: Would you be able to follow up after merging this? Given the return value is an arbitrary `T`, I don't see any way to easily fix this line to not return null, which is what's propagating up to here.

https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/AggregatorHandle.java#L94

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageUtils.java
```diff
@@ -24,6 +24,9 @@ private MetricStorageUtils() {}
    */
   static <T> void mergeInPlace(
       Map<Attributes, T> result, Map<Attributes, T> toMerge, Aggregator<T> aggregator) {
+    if (result == null || toMerge == null) {
```
jack-berg, line 27: I dont think these checks are required anymore after the updates to remove `@Nullable` from `TemporalMetricStorage#LastReportedAccumulation`.

jack-berg (approved):
Minor comments, but good to include these checks 👍

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/Aggregator.java
jsuereth: I'm a fan of either or both of those.  Right now EmptyAggregator shouldn't have any of its methods called in practice, it's just used as a placeholder to disable metrics via views, the storage should never actually be wired up with it, so all of this gymnastics is just to appease our tooling since we can't express that aspect simply.

TL;DR; I'm a fan of Jack's proposal

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
```diff
@@ -153,14 +153,15 @@ public void close() {
       if (!disableSynchronousCollection) {
         lastCollectionTimestamp.lazySet(currentNanoTime);
       }
+      CollectionInfo info = collectionInfoMap.get(handle);
+      if (info == null) {
+        return Collections.emptyList();
```
jsuereth, line 158: This should likely through some kind of logic error bug.  We own both the Handle-creation *and* the collectionInfoMap, so we should never have a value that's not in the map, if we do the SDK itself has a bug.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
jsuereth: We use `null` to denote no-recording and drop bad cells in the Map of Attribute -> value.  However that's only for Synchronous storage, I'm curious how it's bubbling up to here, how can I look into that?


The TL;DR: though is you should consider an accumulation of `null` here a bug, and we should throw/log so we can fix the logic error.

jsuereth (commented):
Two comments on making sure we get enough logs/errors to account for logic errors that may occur (leading to NPEs).

Thanks for tackling this work!  Will be super exciting to have good null checks going forward!

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
anuraaga: Made it an exception - looked through and it looks like the map is a side channel to workaround a circular dependency between reader and collectioninfo. We can probably improve in the future by refactoring to have the reader's info (temporality settings) passed separately from the collectioninfo

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
anuraaga: I think this is the same as @jack-berg's proposal above. I'd like to merge this before making such a large refactoring

jkwatson (approved):

PR #3847, opened 2021-11-09
OTLP metric temporality configuration

### LINKED ISSUE #3790: Allow OTLP exporters to select aggregation temporality based on environment configuration
Per-spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/otlp.md

jack-berg:
Resolves #3790.

Make the preferred temporality of the OTLP grpc / http metric exporter configurable. Can be configured programmatically or via env / system properties.

### DIFF
--- a/exporters/otlp-http/metrics/src/test/java/io/opentelemetry/exporter/otlp/http/metrics/OtlpHttpMetricExporterTest.java
+++ b/exporters/otlp-http/metrics/src/test/java/io/opentelemetry/exporter/otlp/http/metrics/OtlpHttpMetricExporterTest.java
@@ -133,6 +133,20 @@ class OtlpHttpMetricExporterTest {
                 OtlpHttpMetricExporter.builder()
                     .setTrustedCertificates("foobar".getBytes(StandardCharsets.UTF_8)))
         .doesNotThrowAnyException();
+
+    assertThatCode(
+            () ->
+                OtlpHttpMetricExporter.builder()
+                    .setPreferredTemporality(AggregationTemporality.DELTA))
+        .doesNotThrowAnyException();
+    assertThat(
+            OtlpHttpMetricExporter.builder()
+                .setPreferredTemporality(AggregationTemporality.DELTA)
+                .build()
+                .getPreferredTemporality())
+        .isEqualTo(AggregationTemporality.DELTA);
+    assertThat(OtlpHttpMetricExporter.builder().build().getPreferredTemporality())
+        .isEqualTo(AggregationTemporality.CUMULATIVE);
   }
 
   @Test
@@ -168,6 +182,10 @@ class OtlpHttpMetricExporterTest {
         .isInstanceOf(IllegalArgumentException.class)
         .hasMessage(
             "Unsupported compression method. Supported compression methods include: gzip, none.");
+
+    assertThatThrownBy(() -> OtlpHttpMetricExporter.builder().setPreferredTemporality(null))
+        .isInstanceOf(NullPointerException.class)
+        .hasMessage("preferredTemporality");
   }
 
   @Test
--- a/exporters/otlp/metrics/src/main/java/io/opentelemetry/exporter/otlp/metrics/OtlpGrpcMetricExporterBuilder.java
+++ b/exporters/otlp/metrics/src/main/java/io/opentelemetry/exporter/otlp/metrics/OtlpGrpcMetricExporterBuilder.java
@@ -12,6 +12,7 @@ import io.grpc.ManagedChannel;
 import io.opentelemetry.exporter.otlp.internal.grpc.GrpcExporter;
 import io.opentelemetry.exporter.otlp.internal.grpc.GrpcExporterBuilder;
 import io.opentelemetry.exporter.otlp.internal.metrics.MetricsRequestMarshaler;
+import io.opentelemetry.sdk.metrics.data.AggregationTemporality;
 import java.net.URI;
 import java.time.Duration;
 import java.util.concurrent.TimeUnit;
@@ -27,9 +28,13 @@ public final class OtlpGrpcMetricExporterBuilder {
   private static final String DEFAULT_ENDPOINT_URL = "http://localhost:4317";
   private static final URI DEFAULT_ENDPOINT = URI.create(DEFAULT_ENDPOINT_URL);
   private static final long DEFAULT_TIMEOUT_SECS = 10;
+  private static final AggregationTemporality DEFAULT_TEMPORALITY =
+      AggregationTemporality.CUMULATIVE;
 
   // Visible for testing
   final GrpcExporterBuilder<MetricsRequestMarshaler> delegate;
+
+  private AggregationTemporality preferredTemporality = DEFAULT_TEMPORALITY;
 
   OtlpGrpcMetricExporterBuilder() {
     delegate =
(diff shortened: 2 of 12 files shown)


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtil.java
```diff
@@ -114,6 +115,22 @@ static void configureOtlpExporterBuilder(
...
 
+  static void configureOtlpAggregationTemporality(
+      ConfigProperties config, Consumer<AggregationTemporality> setAggregationTemporality) {
+    String temporalityStr = config.getString("otel.exporter.otlp.metrics.temporality");
+    if (temporalityStr == null) {
+      return;
+    }
+    AggregationTemporality temporality;
+    try {
+      temporality = AggregationTemporality.valueOf(temporalityStr);
```
anuraaga, line 126: Should we do `toUpperCase`?

jack-berg: I couldn't find any precedent for parsing config to enum values, but for the propagators which have a known set of config values all  in lower case, we don't convert to lower when checking equality.

anuraaga: Yeah - so I expect users to generally want to pass these in as lower-case since enum values are generally lowercase. I guess `valueOf` will fail for that. How about just adding a switch instead then to be similar to the others?

jsuereth (approved):

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtil.java
anuraaga: Actually noticed spec says case-insensitive. So I guess we would want `toUpperCase` to accept the lowercase versions

jack-berg: Ah so it does. It appears to be the only env var that is to be interpreted with case insensitivity. Although, it is still experimental and subject to change. 

Pushed a commit with the update 👍.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
```diff
@@ -101,6 +101,7 @@ static MetricExporter configureOtlpMetrics(
           builder::setTimeout,
           builder::setTrustedCertificates,
           (unused) -> {});
+      OtlpConfigUtil.configureOtlpAggregationTemporality(config, builder::setPreferredTemporality);
```
breedx-splk, line 104: Why treat this differently than the other setters? Are you opposed to adding another chunk to the monster method? I'm just thinking about consistency.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtil.java
```diff
@@ -114,6 +115,22 @@ static void configureOtlpExporterBuilder(
...
+      return;
+    }
+    AggregationTemporality temporality;
+    try {
+      temporality = AggregationTemporality.valueOf(temporalityStr.toUpperCase());
+    } catch (IllegalArgumentException e) {
+      throw new ConfigurationException(
+          "Unrecognized aggregation temporality: " + temporalityStr, e);
+    }
+    setAggregationTemporality.accept(temporality);
```
breedx-splk, line 131: Move this and the declaration of `temporality` var into the `try` block.

breedx-splk (approved):
Looks good to me. I had a few questions/ideas, but nothing critical. Good stuff 👍🏻

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java
jack-berg: Because this only applies to the metric exporters. The other method sets the common properties for span and metric exporters.

exporters/otlp-http/metrics/src/test/java/io/opentelemetry/exporter/otlp/http/metrics/OtlpHttpMetricExporterTest.java
```diff
@@ -133,6 +133,20 @@ void validConfig() {
...
+                    .setPreferredTemporality(AggregationTemporality.DELTA))
+        .doesNotThrowAnyException();
+    assertThat(
+            OtlpHttpMetricExporter.builder()
+                .setPreferredTemporality(AggregationTemporality.DELTA)
+                .build()
+                .getPreferredTemporality())
+        .isEqualTo(AggregationTemporality.DELTA);
+    assertThat(OtlpHttpMetricExporter.builder().build().getPreferredTemporality())
+        .isEqualTo(AggregationTemporality.CUMULATIVE);
```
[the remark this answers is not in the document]
jack-berg: That's true. Though that makes it easier to refactor the default temporality to be delta, which is definitely not the right thing, without having a test fail.

exporters/otlp/metrics/src/main/java/io/opentelemetry/exporter/otlp/metrics/OtlpGrpcMetricExporterBuilder.java
```diff
@@ -27,10 +28,14 @@
   private static final String DEFAULT_ENDPOINT_URL = "http://localhost:4317";
   private static final URI DEFAULT_ENDPOINT = URI.create(DEFAULT_ENDPOINT_URL);
   private static final long DEFAULT_TIMEOUT_SECS = 10;
+  private static final AggregationTemporality DEFAULT_TEMPORALITY =
```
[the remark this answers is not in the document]
jack-berg: > Does it make sense for exporters to have their own defaults?

I think it does. I think temporality is going to be one of the main design dimensions when building an exporter, since there are implications on the backend design.

Also, `preferredTemporality()` is nullable. If an exporter returns null then the [TemporalityUtils#resolveTemporality](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/TemporalityUtils.java#L21-L38) is called to resolve it. So effectively `TemporalityUtils` owns the definition of default.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtil.java
jack-berg: I don't think that's right. The try block is trying to explicitly catch a distinct type of error that occurs when parsing a string to an enum. Including this in the try block would cause errors in `setAggregationTemporality#accept`  to manifest with misleading exceptions that read `Unrecognized aggregation temporality:`.

There's no expectation that an error will occur on this line, and unit tests exist which exercise it according to its usage so I think its appropriate to keep it outside a try / catch.

jkwatson (approved):
👍🏽

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtil.java
breedx-splk: I guess I just think it's kinda ugly to declare a scoped variable that is like half initialized and forces the reader to visually scan to grok that you won't hit the setter in the invalid case. I dunno. 

What I've done elsewhere is to make a similar `valueOf` but called `fromString` or something that encapsulates the try/catch behavior. Then the caller looks a lot cleaner...but the `ConfigurationException` is NOT specific to the enum and wouldn't make sense in this case, so I dunno.

jack-berg: Yeah enum parsing has always been frustratingly verbose.

anuraaga (approved):

PR #3849, opened 2021-11-10
Allow specifying SPI classloader during autoconfiguration.

### LINKED ISSUE #3634: autoconfigure indirectly uses Thread.currentThread().getContextClassLoader()
This has cost me some time debugging: I used OTel inside Google Cloud Functions https://cloud.google.com/functions/docs/first-java and added `opentelemetry-sdk-extension-resources` to the classpath. In unit tests, everything worked but not when actually executing the test with the `com.google.cloud.functions.invoker.runner.Invoker`. The reason turned out to be that that one uses a custom context class loader. I assume Java Web Servers might be affected by this too.

I was able to work around this using

```java
final ClassLoader contextClassLoader = Thread.currentThread().getContextClassLoader();
Thread.currentThread().setContextClassLoader(getClass().getClassLoader());
try {
  this.otelSdk = OpenTelemetrySdkAutoConfiguration.initialize();
} finally {
  Thread.currentThread().setContextClassLoader(contextClassLoader);
}
```

The fix / behaviour change in autoconfigure would be to use the [`ServiceLoader.load` overload taking a Loader](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html#load-java.lang.Class-java.lang.ClassLoader-) and passing the classloader of the autoconfigure class.

I'm not completely sure if this is a bug (at all or in otel-java), but I am sympathetic to the opinion that relying on the context class loader is an anti-pattern, without having enough background knowledge to be sure about it.

anuraaga:
Fixes #3634

### DIFF
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
@@ -43,14 +43,14 @@ public final class AutoConfiguredOpenTelemetrySdkBuilder implements AutoConfigur
 
   private Supplier<Map<String, String>> propertiesSupplier = Collections::emptyMap;
 
+  private ClassLoader serviceClassLoader =
+      AutoConfiguredOpenTelemetrySdkBuilder.class.getClassLoader();
+
   private boolean setResultAsGlobal = true;
 
-  AutoConfiguredOpenTelemetrySdkBuilder() {
-    for (AutoConfigurationCustomizerProvider customizer :
-        ServiceLoader.load(AutoConfigurationCustomizerProvider.class)) {
-      customizer.customize(this);
-    }
-  }
+  private boolean customized;
+
+  AutoConfiguredOpenTelemetrySdkBuilder() {}
 
   /**
    * Sets the {@link ConfigProperties} to use when resolving properties for auto-configuration.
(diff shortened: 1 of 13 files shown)


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
```diff
@@ -43,14 +43,14 @@
 
   private Supplier<Map<String, String>> propertiesSupplier = Collections::emptyMap;
 
+  private ClassLoader serviceClassLoader =
```
anuraaga, line 46: This isn't the context classloader anymore - most of what I know of SPI is from the Armeria codebase where this is always the default we use, I have a feeling it has less gotchas than context, though both will naturally have some gotchas so we allow customizing.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
```diff
@@ -148,19 +148,36 @@ public AutoConfiguredOpenTelemetrySdkBuilder setResultAsGlobal(boolean setResult
...
+    return this;
+  }
+
   /**
    * Returns a new {@link AutoConfiguredOpenTelemetrySdk} holding components auto-configured using
    * the settings of this {@link AutoConfiguredOpenTelemetrySdkBuilder}.
    */
   @SuppressWarnings("deprecation") // Using classes which will be made package-private later.
   public AutoConfiguredOpenTelemetrySdk build() {
+    if (!customized) {
```
jkwatson, line 165: this is just to make sure that if the builder is re-used, we don't SPI and customize more than once?

jack-berg (approved):

jkwatson (approved):

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
anuraaga: Yeah

cyrille-leclerc:
Could we have forgotten to introduce the same serviceClassloader mechanism in [OpenTelemetryResourceAutoConfiguration#configureResource()](https://github.com/open-telemetry/opentelemetry-java/blob/v1.10.0-rc.1/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetryResourceAutoConfiguration.java#L60) when loading the `ResourceProvider`?

See https://github.com/open-telemetry/opentelemetry-java/blob/v1.10.0-rc.1/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetryResourceAutoConfiguration.java#L60

I create the feature request:
- https://github.com/open-telemetry/opentelemetry-java/issues/3998

PR #3854, opened 2021-11-11
Use a wrapper of primitive long array for histogram counts.

anuraaga:
Since we've talked about needing this for the exponential histogram, which may have many buckets, figured may as well add for the current one first. While the impact shouldn't be as big for a normal histogram, I think, it's still a reasonable optimization I think.

### DIFF
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/internal/PrimitiveLongList.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/internal/PrimitiveLongList.java
@@ -29,7 +29,7 @@ public final class PrimitiveLongList {
   }
 
   /**
-   * Returns a primitive array with the values of the list. The list should generally have been *
+   * Returns a primitive array with the values of the list. The list should generally have been
    * created with {@link PrimitiveLongList#wrap(long[])}.
    */
   public static long[] toArray(List<Long> list) {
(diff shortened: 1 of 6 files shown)


sdk/common/src/main/java/io/opentelemetry/sdk/internal/PrimitiveLongList.java
```diff
@@ -0,0 +1,68 @@
...
+  /**
+   * Returns a list that wraps the primitive array. Modifications in the array will be visible in
+   * the list.
+   */
+  public static List<Long> wrap(long[] values) {
+    return new LongListImpl(values);
+  }
+
+  /**
+   * Returns a primitive array with the values of the list. The list should generally have been *
```
jack-berg, line 32:
```suggestion
   * Returns a primitive array with the values of the list. The list should generally have been
```

sdk/common/src/test/java/io/opentelemetry/sdk/internal/PrimitiveLongListTest.java
```diff
@@ -0,0 +1,42 @@
...
+package io.opentelemetry.sdk.internal;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import java.util.Arrays;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+class PrimitiveLongListTest {
```
jack-berg, line 15: I think it would be nice to add an assertion of sameness:
```
    long[] arr = new long[] {1, 2, 3};
    assertThat(PrimitiveLongList.toArray(PrimitiveLongList.wrap(arr))).isSameAs(arr);
```

jack-berg (approved):

jkwatson:
build failing on both java 8 instances?

anuraaga:
Java 8 has a terrible IndexOutOfBounds exception message :P

jsuereth (approved):
Nice!!!

jkwatson (approved):

PR #3860, opened 2021-11-11
Optimize `StringUtils.isPrintableString`

reugn:


api/all/src/main/java/io/opentelemetry/api/internal/StringUtils.java
```diff
@@ -85,6 +85,9 @@ private static String padStart(String string, int minLength, char padChar) {
    * @return whether the {@code String} contains only printable characters.
    */
   public static boolean isPrintableString(String str) {
+    if (isNullOrEmpty(str)) {
+      return false;
```
jack-berg, line 89: This changes the implied contract of `isPrintableString`. It suggests we should change `String str` to be `@Nullable`, and update the documentation to clarify that null and empty strings are _not_ considered printable. 

However, the motivation behind this change is not clear. Is it really better this way?

reugn: @jack-berg changed the contract to Nullable. It feels better for such a kind of utility method. But on the other hand, it ended up as an inessential rework. Please reject, if the previous looks better to you.

anuraaga: TBH I don't think we benefit much from this change.

Also for future reference, a more appropriate PR title could have been `Check for null in isPrintableString instead of caller` - seeing the word "optimize" I was expecting a performance optimization :P

reugn: It's neglectable, but I still think it's better to have the parameter Nullable. Agree with you as for the title.

kubawach: There is only a single usage of this method, so the impact of the change is quite limited. There is other question tho - whether this kind of methods (static, helper, boolean) are expected to carry null checks or not? My personal taste is that they are indeed. In other words, I'd approve :)

jkwatson: Since this change is on an `internal` class, I would not make it. It adds little value, IMO.

PR #3862, opened 2021-11-12
Add getAttribute() to SpanData and EventData

breedx-splk:
Quite a few places where we call `getAttributes().get(key)` on `SpanData` (and spans) and also `EventData`. This adds the singular `getAttribute()` to simplify it a bit and fix a small law of Demeter violation (technicality).

### DIFF
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/data/SpanData.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/data/SpanData.java
@@ -5,6 +5,7 @@
 
 package io.opentelemetry.sdk.trace.data;
 
+import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.trace.SpanContext;
 import io.opentelemetry.api.trace.SpanKind;
@@ -12,6 +13,7 @@ import io.opentelemetry.sdk.common.InstrumentationLibraryInfo;
 import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.sdk.trace.SpanLimits;
 import java.util.List;
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
 /**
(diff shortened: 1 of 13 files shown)


jkwatson:
Outside of testing use-cases, I'm not sure this is actually going to end up being a very useful addition to the *Data APIs. The main consumer of them is exporters, and exporters will (almost) always want to just grab the full collection of attributes and iterate over them.

I'm not saying this is bad, but it might be more of an internal use-case for tests, that we might be able to tackle via our testing module.

breedx-splk:
> Outside of testing use-cases, I'm not sure this is actually going to end up being a very useful addition to the *Data APIs. The main consumer of them is exporters, and exporters will (almost) always want to just grab the full collection of attributes and iterate over them.
> 
> I'm not saying this is bad, but it might be more of an internal use-case for tests, that we might be able to tackle via our testing module.

Yeah that's fair. 🤷🏻

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/data/SpanData.java
```diff
@@ -93,6 +95,16 @@ default String getParentSpanId() {
    */
   Attributes getAttributes();
 
+  /**
+   * Returns the value for a given attribute key.
```
anuraaga, line 99: or null if the attribute is not set

anuraaga (approved):

anuraaga:
I'm ok with this but it's true that for end users this should generally be covered by the assertj helper

jack-berg:
I think there's a design principle / [best practice](https://github.com/open-telemetry/opentelemetry-java/blob/main/CONTRIBUTING.md#best-practices-that-we-follow) being implicitly discussed with this PR: Do we favor a smaller API surface area or one that's more ergonomic?

In practice we're somewhere in the middle. For example we recently added `AttributesBuilder#removeIf(..)` method despite it being possible, though inconvenient, to take an `Attributes` and remove select entries from it.

I suggest we adopt the practice of limiting the API surface area, unless there stands to be a non-trivial (alas, a subjective decision) performance benefit from extending it.

breedx-splk:
`homer-simpson-quietly-backing-into-the-bushes.gif` 🤪

jkwatson:
> I think there's a design principle / [best practice](https://github.com/open-telemetry/opentelemetry-java/blob/main/CONTRIBUTING.md#best-practices-that-we-follow) being implicitly discussed with this PR: Do we favor a smaller API surface area or one that's more ergonomic?
> 
> In practice we're somewhere in the middle. For example we recently added `AttributesBuilder#removeIf(..)` method despite it being possible, though inconvenient, to take an `Attributes` and remove select entries from it.
> 
> I suggest we adopt the practice of limiting the API surface area, unless there stands to be a non-trivial (alas, a subjective decision) performance benefit from extending it.

I think there's another case that isn't performance-related, and that's developer UX (not us, but our users). In this particular case, I think it wasn't going to help people outside the project too much, even if it would make our lives a tad easier.

PR #3867, opened 2021-11-12
Refactor `MeterSharedState`

reugn:


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorage.java
```diff
@@ -45,4 +46,15 @@ MetricData collectAndReset(
...
       boolean suppressSynchronousCollection);
+
+  /**
+   * Determines whether this storage is an empty metric storage.
+   *
+   * <p>Uses the reference comparison since {@link EmptyMetricStorage} is singleton.
+   *
+   * @return true if is empty.
+   */
+  default boolean isEmpty() {
```
jsuereth, line 57: Nit: You could make this be a static method similar to the `register` method you added, rather than adding a new instance-level interface method that's really only used to turn off storage.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MeterSharedState.java
```diff
@@ -102,48 +99,43 @@ public final void registerLongAsynchronousInstrument(
...
+        .forEach(this::register);
+  }
+
+  private <S extends MetricStorage> S register(S storage) {
+    try {
+      return getMetricStorageRegistry().register(storage);
+    } catch (DuplicateMetricStorageException e) {
+      logger.log(Level.WARNING, e, () -> DebugUtils.duplicateMetricErrorMessage(e));
     }
+    return null;
```
jsuereth, line 139: Nit: As we're moving to non-null here, might make sense to use a "NoMetricStorage", or alter the return to be Optional<S>

jsuereth (approved):
Nice clean up!

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MeterSharedState.java
anuraaga: We avoid `Optional` in this codebase (it's equivalent to Nullable given our checker setup) but an empty type would be good if possible

reugn: It's a private stealth method. Having it void would be the best, but I need the returned value in the first use case (could iterate again to register though).

kubawach (approved):
Look good

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MeterSharedState.java
anuraaga: Think this method needs the Nullable annotation - I merged in main to the PR to have the nullness checking activated

reugn: Added.

anuraaga (approved):

PR #3870, opened 2021-11-12
Cleanup trace incubator deprecations

jack-berg:
- Remove SpanDataBuilder
- Remove DelegatingSpanData


jack-berg:
I should look closer at outstanding PRs. This has already been done!

PR #3871, opened 2021-11-13
Move key/value validation from ImmutableBaggage to the W3CBaggagePropagator

### LINKED ISSUE #2200: Consider moving the baggage entry validation code to the W3CBaggagePropagator
Currently, baggage entries (keys & values) are validated for W3C-correctness when they are added to the Baggage via the builder. This operation is not cheap, as it requires character-by-character inspection of both key & value. 

We should consider moving the validation to the propagator, so in-process Baggage usage would be cheaper and not need to conform to the W3C specification.

breedx-splk:
Resolves #2200

Not sure how much folks [still?] like this idea, but this is maybe what it might look like?

The validation now happens during `inject()` time, and invalid baggage entries are just dropped from propagation.

I wasn't sure about the `extract()` case, it looked like the parser was probably the place to do incoming validation?  [This note](https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/baggage/propagation/Parser.java#L19) suggests that it's doing some intentional validation.


jkwatson:
I'm not sure this is a good idea. Should our API allow *any* key to be used for baggage, and then throw that data away when going to propagate it downstream? I mean...this would be a sneaky way to implement in-process-only baggage items (just prefix your keys with some random emoji), but I wonder if that's the intent here.

I think we need some spec clarification on how baggage is supposed to work, and the interop with the W3C baggage propagator.

breedx-splk:
> I'm not sure this is a good idea. <snip>

Heh, well, yeah....you wrote #2200.  🙃 

If you're happy to close that issue and this PR, I'm happy....but maybe I just misunderstood the spirit of the original issue?

jkwatson:
> > I'm not sure this is a good idea.
> 
> Heh, well, yeah....you wrote #2200. 🙃
> 
> If you're happy to close that issue and this PR, I'm happy....but maybe I just misunderstood the spirit of the original issue?

We ended up encoding all the possible string *values* already in the propagator, and lifting any restrictions on what they can contain. I'm just not sure about the key, since that can't really be encoded (at least by the W3C spec), so it's a more complex case. I'd like some more opinions than just mine, though.

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/W3CBaggagePropagator.java
```diff
@@ -48,20 +50,29 @@ private W3CBaggagePropagator() {}
     if (baggage.isEmpty()) {
       return;
     }
+    StringBuilder headerContent = baggageToString(baggage);
+
+    if (headerContent.length() > 0) {
```
anuraaga, line 55: I think better to do this in `baggageToString` and return a `String` instead of builder

anuraaga (commented):
Clarifying in the spec repo could be helpful. In the meantime, I'm in support of this PR since validation isn't mentioned in the otel spec at all

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/baggage/api.md#set-value

It is mentioned in the w3c baggage spec, and that seems to be exactly what this PR reflects.

jkwatson (approved):
Let's do it. shouldn't negatively impact anyone at this point, since it will only allow more internal baggage entries and won't change the behavior of propagation

api/all/src/main/java/io/opentelemetry/api/baggage/propagation/W3CBaggagePropagator.java
```diff
@@ -48,20 +50,35 @@ private W3CBaggagePropagator() {}
     if (baggage.isEmpty()) {
       return;
     }
+    String headerContent = baggageToString(baggage);
+
+    if (!headerContent.isEmpty()) {
```
anuraaga, line 55: @breedx-splk Fixed some nits in https://github.com/open-telemetry/opentelemetry-java/pull/3871/commits/001f5aa241d1ab376b25cad46e359f02bb1d2d04

anuraaga (approved):

breedx-splk:
Are we waiting on this one? I forget.

anuraaga:
Thanks for the reminder, had meant to merge I think

PR #3872, opened 2021-11-13
Use LinkedList for better performance

reugn:


anuraaga:
We don't except non-append operations to be common in practice so I'm not sure this change would improve performance in practice. We'll need JMH numbers to consider this change.

anuraaga:
Oops, realized that even append is adding to the front. We could probably either reverse the list, provide a reversed list view, or some other approach not requiring the allocation overhead of a linkedlist

reugn:
@anuraaga it's an internal tweak, the returned object on build is an ArrayList.

anuraaga:
Ah in that case even more so, we need to compare the JMH numbers. For example compared to reversing an ArrayList

knutwannheden:
I think an ArrayDeque would be the appropriate data structure to use here. I could provide an alternate PR.

jkwatson:
> I think an ArrayDeque would be the appropriate data structure to use here. I could provide an alternate PR.

Any internal implementation changes like this will need to be accompanied by benchmarks that clearly show a performance improvement.

anuraaga:
@reugn This was superceded by #4325. A benchmark was added there too so if you see anything left to try it will be easier to accept. Thanks

PR #3880, opened 2021-11-15
Break out logs appender API

### LINKED ISSUE #3807: Log instrumentation should not have to depend on the log SDK

jack-berg:
Resolves #3807.

The most interesting decision is what to name the log appender API artifact / module.

I've put it in `:api:logs-appender`. I also considered `:extensions:logs-appender`, but it didn't feel right for `:sdk:logs` to have a dependency on an extension module. I also chose "logs-appender" instead of "log-appender" for consistency with the plural logs and metrics SDK module names. I'm open to other suggestions.


jkwatson:
I don't think we should add an API API for this at this point. There is no spec for it, and I'm not convinced there will be in the near future. Appenders can and should depend directly on the SDK if they need to be sending data to the SDK. Users who need to use an appender can also depend on the SDK, IMO.

jack-berg:
This appeared to be the preferred direction for a time.

I'm just as happy sticking with only an SDK until something in the spec changes that suggests an API. After all, despite the module name including the `-appender` suffix, it looks an awful lot like the trace and metric APIs, which is not the current intent on the spec.

 I'm fine dropping this, and instead having instrumentation modules place a `compileOnly` dependency on the log sdk.

anuraaga:
I think in Java we have a unique situation where the javaagent will add the appender to a user's app (I guess we're planning this). We usually add instrumentation into, such as this custom appender, into the bootstrap classloader affecting the entire app but we very much don't want to add SDKs there, we keep them in the agent classloader.

As long as there's an API split out somewhere, for example `opentelemetry-java/sdk/logs-sdk-api` (Javalicious!), that would be fine. That being said, if some instrumentation like RUM will want to send "events" via OTel logs, I'm curious how we can get by without an actual logs API in the end. Perhaps both of these points could be raised in the logs group as reasons to reconsider - but if there's no API in the spec, I think for our Java reasons we do need a split here of some sort.

jkwatson:
> I think in Java we have a unique situation where the javaagent will add the appender to a user's app (I guess we're planning this). We usually add instrumentation into, such as this custom appender, into the bootstrap classloader affecting the entire app but we very much don't want to add SDKs there, we keep them in the agent classloader.
> 
> As long as there's an API split out somewhere, for example `opentelemetry-java/sdk/logs-sdk-api` (Javalicious!), that would be fine. That being said, if some instrumentation like RUM will want to send "events" via OTel logs, I'm curious how we can get by without an actual logs API in the end. Perhaps both of these points could be raised in the logs group as reasons to reconsider - but if there's no API in the spec, I think for our Java reasons we do need a split here of some sort.

For RUM, we're planning on designing a RUM API for otel which will be implemented with the various SDKs. The hope is that it'll be RUM specific, though, and not a part of the general use APIs.

I'd be good with an sdk-scoped "appender api" for now until we have more clarity about what (if any) a final logs/appender API might look like.

jack-berg:
> We usually add instrumentation into, such as this custom appender, into the bootstrap classloader affecting the entire app but we very much don't want to add SDKs there, we keep them in the agent classloader.

Sorry if this is a naive question - I haven't studied the java agent's classloader setup: Is there a way to include the appenders in the agent but with a `compileOnly` dependency on the SDK, such that the app has to have a runtime dependency on the SDK for the appenders to "activate"?

anuraaga:
@jack-berg - the idea of using a user's library has come to mind, for example to redirect the agent's logs through a user's logback.xml. But it's tricky, there can be dependency conflicts between the agent and the user's app that cause LinkageError or similar problems with the approach. Avoiding dependency conflicts is the motivation for the classpath separation and why we like to keep more complicated components, like the SDK, in the agent classloader to not affect the app.

jack-berg:
@anuraaga in today's SIG meeting we discussed the possibility of including an appender API in `opentelemetry-java-instrumentation` instead. The idea being that its not needed outside of the appender use cases so we can avoid possible confusion arising related to what purpose an appender API serves.

jack-berg:
Closing this in favor of adding an appender API in `opentelemetry-java-instrumentation`. Created this [ticket](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4790) to track the work.

PR #3887, opened 2021-11-16
Change EmptyAggregator to Drop Aggregator, remove @Nullable

### LINKED ISSUE #3886: Rename EmptyAggregator to DropAggregator
To reflect recent update to the [spec](https://github.com/open-telemetry/opentelemetry-specification/pull/2101). 

Should also solve the problem with `Aggregator#toMetricData(..)` annotated `@Nullable`, which propagates into several other places in the SDK.

jack-berg:
Resolves #3886.

Also removes several `@Nullable` annotations as proposed [here](https://github.com/open-telemetry/opentelemetry-java/pull/3845#discussion_r747891816).


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/MetricData.java
```diff
@@ -170,6 +171,16 @@ static MetricData createExponentialHistogram(
         data);
   }
 
+  /**
+   * Get an empty instance of metric data. Empty metric data should not be used except as an
+   * indication that recordings were dropped via {@link Aggregation#drop()}.
+   *
+   * @return an empty MetricData.
+   */
+  static MetricData empty() {
```
anuraaga, line 180: Should we leave this out of public API until supporting custom aggregations?

jsuereth: The Drop aggregator is exposed in the SDK, so it's possible to see this given this PR.


Second question though - Can you add some tests where this shows up in Exporters?  I'm curious if they need to be updated for this data type, I suspect it will cause breakages.

jack-berg: > Second question though - Can you add some tests where this shows up in Exporters? I'm curious if they need to be updated for this data type, I suspect it will cause breakages.

The way I handled this was with a `MetricData.isEmpty()` check in `MeterSharedState#collectAll(...)`, and `EmptyMetricData.isEmpty()` always returns `true`:
```
  public List<MetricData> collectAll(
      CollectionInfo collectionInfo,
      MeterProviderSharedState meterProviderSharedState,
      long epochNanos,
      boolean suppressSynchronousCollection) {
    Collection<MetricStorage> metrics = getMetricStorageRegistry().getMetrics();
    List<MetricData> result = new ArrayList<>(metrics.size());
    for (MetricStorage metric : metrics) {
      MetricData current =
          metric.collectAndReset(
              collectionInfo,
              meterProviderSharedState.getResource(),
              getInstrumentationLibraryInfo(),
              meterProviderSharedState.getStartEpochNanos(),
              epochNanos,
              suppressSynchronousCollection);
      // Ignore if the metric data doesn't have any data points, for example when aggregation is
      // Aggregation#drop()
      if (!current.isEmpty()) {
        result.add(current);
      }
    }
    return result;
  }
```

jack-berg: Added a unit test in latest commit

anuraaga: > Should we leave this out of public API until supporting custom aggregations?

Not sure if this was answered - does it work to leave this out of the public API for now?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/EmptyMetricData.java
```diff
@@ -0,0 +1,57 @@
...
+final class EmptyMetricData implements MetricData {
+
+  static final EmptyMetricData INSTANCE = new EmptyMetricData();
+
+  private EmptyMetricData() {}
+
+  @Override
+  public Resource getResource() {
+    throw new UnsupportedOperationException("EmptyMetricData does not support getResource().");
+  }
```
anuraaga, line 20: Let's go ahead and add unit tests for these methods. Also in case a caller doesn't call `isEmpty` should we return default values instead of throwing?

jack-berg: I thought about returning default values but couldn't come up with reasonable "empty" values for all the fields. And I didn't want to make any of the fields `@Nullable`, or have inconsistent behavior where some accessors throw and others return "empty" values. 

If we do our job right and check for `isEmpty()` in the SDK, this _shouldn't_ ever escape to users.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/MetricData.java
jack-berg: Possibly, yes. 

The `MetricData` interface includes `Data<?> getData()`, and the `Data<T extends PointData>` interface is package private. This means that any implementation of `MetricData` has to live in `io.opentelemetry.sdk.metrics.data`. The fact that `MetricData` has a public API that returns a package private interface is probably a bug. 

But while its true that `MetricData` implementations needs to be in `io.opentelemetry.sdk.metrics.data`, there's no way (short of reflection) to have a `EmptyMetricData` implementation without some sort of public accessor, since a public accessor is needed for `DropAggregator` and other internal accesses. 

But if we make `Data` public, we can move `EmptyMetricData` to an internal package and remove `MetricData.empty()`.

anuraaga: > The fact that MetricData has a public API that returns a package private interface is probably a bug.

Nice catch! Sounds like a plan

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/EmptyMetricData.java
```diff
@@ -0,0 +1,75 @@
...
+   *
+   * @return an empty MetricData.
+   */
+  public static MetricData getInstance() {
+    return INSTANCE;
+  }
+
+  @Override
+  public Resource getResource() {
+    throw new UnsupportedOperationException("EmptyMetricData does not support getResource().");
```
anuraaga, line 37: Perhaps there isn't much point but let's go ahead and add unit tests for this class specifically that just check `.isInstanceOf` on the exception. It can be considered a double-check that the semantics are indeed what we intended

anuraaga (approved):

jsuereth (approved):

PR #3893, opened 2021-11-17
Make InstrumentSelect#getInstrumentType() nullable

### LINKED ISSUE #3892: InstrumentSelector.getInstrumentType should be nullable
Right now, if you try to register a view with an instrument selector using a instrument name regex, you get an error saying: `Missing required properties: instrumentType`.

```
.registerView(InstrumentSelector.builder()
  .setInstrumentNameRegex(".*")
  .build(),
  ...)
```

The `InstrumentSelector` javadoc suggests that instrumentation type should be nullable such that you can just specify an instrument name selector:

```
  /**
   * Returns {@link InstrumentType} that should be selected. If null, then this specifier will not
   * be used.
   */
  public abstract InstrumentType getInstrumentType();
```

Also, instrument type is optional according to the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#view):

> The Instrument selection criteria (required), which covers:
The type of the Instrument(s) (optional).

jack-berg:
Resolves #3892.


jsuereth (approved):
Yay!!!

jkwatson (approved):

PR #3897, opened 2021-11-17
OTLP exporter - oneof primtives must have their default values serialized.

### LINKED ISSUE #3889: Metric point marshaller not handling `oneof` in proto correctly
**Describe the bug**
When reporting Gauge or Sum values of 0, you expect to see one of the two [oneof values](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L380) filled out.

**Steps to reproduce**
Create a Counter, report repeated values of 0.  The resulting proto field will not have `as_double` or `as_int` filled out.

For `oneof` fields, even primitives, you need to serialize the field number + value.

**What did you see instead?**
The field is not serialized, assuming the default will get used.

OpenTelemetry 1.7.x

jsuereth:
Fix #3889 

There may be more instances of this, just fixing the obvious ones.

### DIFF
--- a/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MarshalerUtil.java
+++ b/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MarshalerUtil.java
@@ -177,25 +177,25 @@ public final class MarshalerUtil {
 
   /** Returns the size of a double field. */
   public static int sizeDouble(ProtoFieldInfo field, double value) {
-    return sizeDouble(field, value, /* force= */ false);
-  }
-  /** Returns the size of a double field. */
-  public static int sizeDouble(ProtoFieldInfo field, double value, boolean force) {
-    if (!force && value == 0D) {
+    if (value == 0D) {
       return 0;
     }
+    return sizeDoubleOptional(field, value);
+  }
+  /** Returns the size of a double field. */
+  public static int sizeDoubleOptional(ProtoFieldInfo field, double value) {
     return field.getTagSize() + CodedOutputStream.computeDoubleSizeNoTag(value);
   }
 
   /** Returns the size of a fixed64 field. */
   public static int sizeFixed64(ProtoFieldInfo field, long value) {
-    return sizeFixed64(field, value, /* force= */ false);
-  }
-  /** Returns the size of a fixed64 field. */
-  public static int sizeFixed64(ProtoFieldInfo field, long value, boolean force) {
-    if (!force && value == 0L) {
+    if (value == 0L) {
       return 0;
     }
+    return sizeFixed64Optional(field, value);
+  }
+  /** Returns the size of a fixed64 field. */
+  public static int sizeFixed64Optional(ProtoFieldInfo field, long value) {
     return field.getTagSize() + CodedOutputStream.computeFixed64SizeNoTag(value);
   }
 
--- a/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MarshalerUtil.java
+++ b/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MarshalerUtil.java
@@ -177,18 +177,26 @@ public final class MarshalerUtil {
 
   /** Returns the size of a double field. */
   public static int sizeDouble(ProtoFieldInfo field, double value) {
-    if (value == 0D) {
+    return sizeDouble(field, value, /* force= */ false);
+  }
+  /** Returns the size of a double field. */
+  public static int sizeDouble(ProtoFieldInfo field, double value, boolean force) {
+    if (!force && value == 0D) {
       return 0;
     }
     return field.getTagSize() + CodedOutputStream.computeDoubleSizeNoTag(value);
   }
 
   /** Returns the size of a fixed64 field. */
-  public static int sizeFixed64(ProtoFieldInfo field, long message) {
-    if (message == 0L) {
+  public static int sizeFixed64(ProtoFieldInfo field, long value) {
+    return sizeFixed64(field, value, /* force= */ false);
+  }
+  /** Returns the size of a fixed64 field. */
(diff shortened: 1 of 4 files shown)


jsuereth:
as an FYI - this is likely an issue with non-string Attribute serialzition, would you like me to fix that in the current PR?  cc @anuraaga @jkwatson

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/Serializer.java
```diff
@@ -95,7 +95,12 @@ public void serializeInt64(ProtoFieldInfo field, long value) throws IOException
 
   /** Serializes a protobuf {@code fixed64} field. */
   public void serializeFixed64(ProtoFieldInfo field, long value) throws IOException {
-    if (value == 0) {
+    serializeFixed64(field, value, /* force= */ false);
+  }
+
+  /** Serializes a protobuf {@code fixed64} field. */
+  public void serializeFixed64(ProtoFieldInfo field, long value, boolean force) throws IOException {
```
anuraaga, line 102: Better to just make `writeFixed64` public I think. If the name is confusing, it could be renamed to `serializeRequiredFixed64` or something

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MarshalerUtil.java
```diff
@@ -177,18 +177,26 @@ public static int sizeInt32(ProtoFieldInfo field, int message) {
...
   }
 
   /** Returns the size of a fixed64 field. */
-  public static int sizeFixed64(ProtoFieldInfo field, long message) {
-    if (message == 0L) {
+  public static int sizeFixed64(ProtoFieldInfo field, long value) {
+    return sizeFixed64(field, value, /* force= */ false);
+  }
+  /** Returns the size of a fixed64 field. */
+  public static int sizeFixed64(ProtoFieldInfo field, long value, boolean force) {
```
anuraaga, line 195: We don't need a force flag, the `== 0L` check can remain in the other method and this one can have a name like `sizedFixed64ForOneof` or `sizedRequiredFixed64` (interesting how the serialization is required for an optional field :P)

jsuereth: It's not an optional field, it's a oneof.  You actually handle this for [AnyValue](https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/KeyValueMarshaler.java#L143), just missed that here.  Didn't think it was worth making a new Marshaller instance for it.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/Serializer.java
jsuereth: Yeah, I'd like to make it public.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MarshalerUtil.java
anuraaga: The name can be whatever but I'd avoid this boolean pattern. If we add any optional field, it would call the same method (optional == oneof), hence referring to optional fields.

jsuereth: Ah, right.   Basically this is:   proto2 *or* proto3Optional *or* proto3 oneof-primtive.

primitive oneof = proto3optional = proto2 in the case of field-presence being mandatory.

In this case it's just the fact it's a oneof.   I can either call the boolean "fieldPresenceRequired" or somehow denote it that way?

anuraaga: I'd prefer separate methods with different names to a boolean, it keeps the logic simpler. An amazing name doesn't come to mind but some form of these suggestions should be ok.

anuraaga: I realized an alternative to figure out naming would be to just stick to our current pattern of not calling a helper for oneofs, for now.

https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/KeyValueMarshaler.java#L128

FYI, on the back burner has been changing the codegen to allow something like `Span.NAME.size(nameUtf8)` instead of using util methods, where the method would implement oneof semantics directly where appropriate.

anuraaga:
> as an FYI - this is likely also an issue with non-string Attribute serialzition

Probably better for another PR since this metric issue is more obvious.

Don't think I see an issue at a glance (and unlike the metric code, the span + attributes code were hardened more strongly using randomized tests vs real proto so I think have less chance of issues, though of course not zero chance as any code)

https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/KeyValueMarshaler.java#L122

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MarshalerUtil.java
jsuereth: That'd be really nice.  I just used "Optional" for now and we can understand that oneof == optional in proto3 encoding.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/Serializer.java
jsuereth: I added an optional method to mirror the "size" method calls, as I think that reduces confusion the most (and simplifies reading this later).

PR #3901, opened 2021-11-18
Don't use global meter provider for trace export

anuraaga:
To prepare for removing GlobalMeterProvider.

- I'll add to logs in a separate PR, currently it's not autoconfigured.
- Currently, the metric exporter doesn't generate metrics since GlobalMeterProvider isn't ready in time. Metric exporter is a unique one so I guess we'll add a mutation-based approach only for it later.

### DIFF
--- a/docs/apidiffs/current_vs_latest/opentelemetry-exporter-otlp-http-trace.txt
+++ b/docs/apidiffs/current_vs_latest/opentelemetry-exporter-otlp-http-trace.txt
@@ -1,2 +1,4 @@
 Comparing source compatibility of  against 
-No changes.
+***  MODIFIED CLASS: PUBLIC FINAL io.opentelemetry.exporter.otlp.http.trace.OtlpHttpSpanExporterBuilder  (not serializable)
+	===  CLASS FILE FORMAT VERSION: 52.0 <- 52.0
+	+++  NEW METHOD: PUBLIC(+) io.opentelemetry.exporter.otlp.http.trace.OtlpHttpSpanExporterBuilder setMeterProvider(io.opentelemetry.api.metrics.MeterProvider)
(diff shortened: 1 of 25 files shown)


jkwatson:
This is the right change, but this will force the issue of stabilizing the metrics SDK in the next release. Are we ready for that?

jack-berg:
> This is the right change, but this will force the issue of stabilizing the metrics SDK in the next release. Are we ready for that?

@jkwatson I think you mean the metric API, not the SDK, since the OTLP exporters have a dependency only on `MeterProvider`. 

And actually, `:exporters:otlp:trace` already has an API dependency on `:exporters:otlp:common`, which has a API dependency on `:api:metrics`. Is that allowed?

jkwatson:
> > This is the right change, but this will force the issue of stabilizing the metrics SDK in the next release. Are we ready for that?
> 
> @jkwatson I think you mean the metric API, not the SDK, since the OTLP exporters have a dependency only on `MeterProvider`.

Whoops! You are totally correct. belay my concern.

> 
> And actually, `:exporters:otlp:trace` already has an API dependency on `:exporters:otlp:common`, which has a API dependency on `:api:metrics`. Is that allowed?

I think we discussed this when that change was made; I don't remember why the dependency is an api dependency, and not an implementation dependency, but the idea is that if you're using the otel bom, then everything should be fine, and the transitive dependency shouldn't hurt anything. At least, that's what I remember.

jkwatson (approved):
Thanks! It's a big step!

anuraaga:
> And actually, :exporters:otlp:trace already has an API dependency on :exporters:otlp:common, which has a API dependency on :api:metrics. Is that allowed?

@jack-berg I think this was actually an accident. It didn't leak through to users though because we only ever have an `implementation` dependency on `:exporters:otlp:common`

PR #3902, opened 2021-11-18
Populate cloud.platform in aws resources

### LINKED ISSUE #3898: AWS Resources are not adding all necessary attributes
**Describe the bug**
AWS Resources in the SDK extension are not adding all the required attributes, so the collector is not extracting origin for instance.
The collector was changed to require the cloud platform attribute, but the Java counter part is not adding this info (https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/426244389f8f6326ba368946cda28eba85284adf/exporter/awsxrayexporter/internal/translator/segment.go#L224)

The Golang version of the ECS detector, and all others are adding more info (https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/426244389f8f6326ba368946cda28eba85284adf/processor/resourcedetectionprocessor/internal/aws/ecs/ecs.go#L57)

It would be awesome to have this fixed in the Java side so AWS X-Ray can decode the origin successfully, and also have more info in the AWS X-Ray Dashboard

**Steps to reproduce**
Create one trace and export to AWS Collector with X-Ray exporter, check the trace segment origin, and there will be nothing

AWS X-Ray with correct information of origin and ECS/EKS info.

**What did you see instead?**
Nothing =]

anuraaga:
Fixes #3898


sdk-extensions/aws/src/test/java/io/opentelemetry/sdk/extension/aws/resource/LambdaResourceTest.java
```diff
@@ -22,7 +23,7 @@
   @Test
   void shouldNotCreateResourceForNotLambda() {
     Attributes attributes = LambdaResource.buildResource(emptyMap()).getAttributes();
-    assertThat(attributes.isEmpty()).isTrue();
+    assertThat(attributes).isEmpty();
```
kubawach, line 26: :)

kubawach (approved):
LGTM :) I really like small changes to tests improving the readability.

jkwatson (approved):

jack-berg (approved):

PR #3903, opened 2021-11-18
Extract gRPC exporter test into shared base class.

anuraaga:
There's currently _tons_ of duplication within exporter tests. This is just the first big cleanup, will extract shared logic for all those testGrpc* test sets and HTTP exporters in the future too. This also makes the default tests use the okhttp implementation, which is our current default (guess need to rename `DefaultGrpcExporter`...)


dependencyManagement/build.gradle.kts
```diff
@@ -15,7 +15,7 @@ val DEPENDENCY_BOMS = listOf(
   "com.fasterxml.jackson:jackson-bom:2.13.0",
   "com.google.guava:guava-bom:31.0.1-jre",
   "com.google.protobuf:protobuf-bom:3.18.1",
-  "com.linecorp.armeria:armeria-bom:1.11.0",
+  "com.linecorp.armeria:armeria-bom:1.13.2",
```
anuraaga, line 18: Needed this bugfix https://github.com/line/armeria/commit/0018233e7a9f6eab6da51f046fc3203135aa1a55

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/grpc/OkHttpGrpcExporter.java
```diff
@@ -146,14 +149,36 @@ public void onResponse(Call call, Response response) {
...
-                logger.log(
-                    Level.WARNING,
-                    "Failed to export "
-                        + type
-                        + "s. Server responded with "
-                        + codeMessage
-                        + ". Error message: "
-                        + errorMessage);
+
+                if (GRPC_STATUS_UNIMPLEMENTED.equals(status)) {
```
anuraaga, line 153: Noticed this user friendliness that is present in DefaultGrpcExporter was missing

exporters/otlp/logs/src/test/java/io/opentelemetry/exporter/otlp/logs/OtlpGrpcLogsExporterTest.java
```diff
@@ -154,257 +147,8 @@ void invalidConfig() {
...
-      assertThat(exporter.export(Collections.singletonList(generateFakeLog())).isSuccess())
-          .isFalse();
-    } finally {
-      exporter.shutdown();
-    }
-  }
-
-  @Test
-  void testExport_PermissionDenied() {
-    fakeCollector.setReturnedStatus(Status.PERMISSION_DENIED);
```
anuraaga, line 351: The old tests had a bunch of status codes but I didn't copy them all since they didn't seem to test anything special

exporters/otlp/testing-internal/src/main/java/io/opentelemetry/exporter/otlp/testing/internal/AbstractGrpcTelemetryExporterTest.java
```diff
@@ -0,0 +1,332 @@
...
+    LoggingEvent log =
+        logs.assertContains(
+            "Failed to export "
+                + type
+                + "s. Server responded with gRPC status code 13. Error message:");
+    assertThat(log.getLevel()).isEqualTo(Level.WARN);
+  }
+
+  @Test
+  void errorWithMessage() {
```
anuraaga, line 232: Added a couple of new tests checking error message handling

exporters/otlp/testing-internal/src/main/java/io/opentelemetry/exporter/otlp/testing/internal/AbstractGrpcTelemetryExporterTest.java
```diff
@@ -0,0 +1,332 @@
...
+                + "receiver in the \"pipelines\" section of the configuration. "
+                + "Full error message: UNIMPLEMENTED");
+    assertThat(log.getLevel()).isEqualTo(Level.ERROR);
+  }
+
+  protected abstract V createExporter(String endpoint);
+
+  protected abstract V createExporterWithTimeout(String endpoint, Duration timeout);
+
+  protected abstract CompletableResultCode shutdownExporter(V exporter);
```
anuraaga, line 303: Perhaps we can have an `.internal.TelemetryExporter` interface to model these

kubawach (approved):
Looks good :)

exporters/otlp/logs/src/test/java/io/opentelemetry/exporter/otlp/logs/OtlpGrpcLogsExporterTest.java
```diff
@@ -8,77 +8,70 @@
...
-import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.extension.RegisterExtension;
-import org.slf4j.event.Level;
-import org.slf4j.event.LoggingEvent;
 
-class OtlpGrpcLogsExporterTest {
+class OtlpGrpcLogsExporterTest
+    extends AbstractGrpcTelemetryExporterTest<LogData, ResourceLogs, OtlpGrpcLogExporter> {
```
jkwatson, line 32: I'm very very leery about creating a test class hierarchy. This inevitably leads to tangled, unmaintainable, incomprehensible test code. Can we use some static helper methods that can be shared, rather than creating a class hierarchy?

exporters/otlp/testing-internal/src/main/java/io/opentelemetry/exporter/otlp/testing/internal/AbstractGrpcTelemetryExporterTest.java
```diff
@@ -0,0 +1,320 @@
...
+public abstract class AbstractGrpcTelemetryExporterTest<T, U extends Message, V> {
+
+  static final ConcurrentLinkedQueue<Object> exportedResourceTelemetry =
+      new ConcurrentLinkedQueue<>();
+
+  @Nullable static volatile ArmeriaStatusException grpcError;
+
+  @RegisterExtension
+  static final ServerExtension server =
+      new ServerExtension() {
```
jack-berg, line 63: What are your thoughts on extracting this out into its own `OtlpGrpcServerExtension` class? The autoconfigure module exercises the otlp exports as well and the same type of setup is repeated over there as well.

exporters/otlp/logs/src/test/java/io/opentelemetry/exporter/otlp/logs/OtlpGrpcLogsExporterTest.java
jack-berg: I think a big part of keeping the test code comprehensible if we go down the class hierarchy route is being meticulous about limiting test scope. 

For example, this class asserts common export patterns shared between the exporters.  Subclasses should only implement the required abstract methods and test nothing else. If other verifications are needed, create a new test class. So if following this advice, the subclass `OtlpGrpcLogsExporterTest` would  move the config validation tests to a different class.

anuraaga: I think moving the config tests out makes sense. I agree @jkwatson with the sentiment of not creating spaghetti, but in this case the tests are 100% the same, so I think it's appropriate here

exporters/otlp/testing-internal/src/main/java/io/opentelemetry/exporter/otlp/testing/internal/AbstractGrpcTelemetryExporterTest.java
anuraaga: I think that's a good idea, would like to do it in a followup since the scope of this PR was this one test, but will go onto the others too later

jkwatson (approved):
provisional approval...I'll revoke it if I see another layer of test hierarchy pop up!

PR #3917, opened 2021-11-19
Lower codecov patch target.

anuraaga:
I guess it's pretty common to have patch diffs with less than 90% coverage that don't strongly affect the coverage of the build.


kubawach (approved):
Makes sense :)

jack-berg (approved):

jkwatson (approved):

PR #3919, opened 2021-11-19
Promote metrics API to stable

anuraaga:
I was hoping to only add to OpenTelemetry this PR and merge the files in in another, but a circular dependency prevented that.

I will deprecate GlobalOpenTelemetry and maybe add the OpenTelemetry.getMeter type of helper methods in future PRs.


api/all/src/main/java/io/opentelemetry/api/OpenTelemetry.java
```diff
@@ -39,6 +40,9 @@ static OpenTelemetry propagating(ContextPropagators propagators) {
   /** Returns the {@link TracerProvider} for this {@link OpenTelemetry}. */
   TracerProvider getTracerProvider();
 
+  /** Returns the {@link MeterProvider} for this {@link OpenTelemetry}. */
+  MeterProvider getMeterProvider();
```
jack-berg, line 44: Isn't it a breaking change to add this without a `default` method?

api/all/src/main/java/io/opentelemetry/api/metrics/NoopMeter.java
```diff
@@ -56,7 +33,7 @@
  * </ul>
  */
 @ThreadSafe
-public class NoopMeter implements Meter {
+class NoopMeter implements Meter {
```
jack-berg, line 36: What do you think about renaming these to `DefaultMeter` and `DefaultTracerProvider` to mirror tracing? Not super important because they're package private, but lowers the cognitive load if someone goes and comparing tracing to metrics.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
```diff
@@ -109,11 +110,14 @@ static OpenTelemetrySdk newOpenTelemetrySdk(
...
-        OpenTelemetrySdk.builder()
-            .setTracerProvider(tracerProvider)
-            .setPropagators(propagators)
-            .build();
+    OpenTelemetrySdkBuilder sdkBuilder =
+        OpenTelemetrySdk.builder().setTracerProvider(tracerProvider).setPropagators(propagators);
+
+    if (meterProvider instanceof SdkMeterProvider) {
+      sdkBuilder.setMeterProvider((SdkMeterProvider) meterProvider);
+    }
```
jack-berg, line 118: We're stuck with this pattern of possibly not having an `SdkMeterProvider` until the metric SDK stabilizes and we remove the "metrics are off by default" feature, right?

sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java
```diff
@@ -83,4 +101,31 @@ public SdkTracerProvider unobfuscate() {
...
+    ObfuscatedMeterProvider(SdkMeterProvider delegate) {
+      this.delegate = delegate;
+    }
+
+    @Override
+    public MeterBuilder meterBuilder(String instrumentationName) {
+      return delegate.meterBuilder(instrumentationName);
+    }
+
+    public SdkMeterProvider unobfuscate() {
```
jack-berg, line 127: Does this need to be public?

jack-berg (approved):

api/all/src/main/java/io/opentelemetry/api/OpenTelemetry.java
anuraaga: Oops thanks! Hmm thought the japi check would have caught this too bad

api/all/src/main/java/io/opentelemetry/api/OpenTelemetry.java
```diff
@@ -39,6 +40,11 @@ static OpenTelemetry propagating(ContextPropagators propagators) {
   /** Returns the {@link TracerProvider} for this {@link OpenTelemetry}. */
   TracerProvider getTracerProvider();
 
+  /** Returns the {@link MeterProvider} for this {@link OpenTelemetry}. */
```
anuraaga, line 43: I think this is a good use of default methods for API evolution. But I'm wondering separate from that whether I should add default to the other methods for consistency and because a single-signal OpenTelemetry seems to be a generally useful concept. Not for this PR I think but something we could do.

extensions/noop-api/src/main/java/io/opentelemetry/extension/noopapi/NoopOpenTelemetry.java
```diff
@@ -6,18 +6,20 @@
...
 /**
  * An implementation of {@link OpenTelemetry} that is completely no-op. Unlike {@link
- * OpenTelemetry#noop()}, this implementation does not support in-process context propagation at
- * all. This means that no objects are allocated nor {@link ThreadLocal}s used in an application
- * using this implementation. This can be a good option for use in frameworks shared across a large
- * number of servers to introduce instrumentation without forcing overhead on any users of the
- * framework. If such overhead is not a concern, always use either {@link OpenTelemetry#noop()},
- * {@link OpenTelemetry#propagating(ContextPropagators)}, or the OpenTelemetry SDK.
+ * OpenTelemetry#noop()}, this implementation does not support in-process context propagation
+ * atjavajava all. This means that no objects are allocated nor {@link ThreadLocal}s used in an
```
jamesmoessis, line 17: atjavajava all?

api/all/src/main/java/io/opentelemetry/api/OpenTelemetry.java
jack-berg: That would be:
```
ContextPropagators getPropagators();
TracerProvider getTracerProvider();
```
I don't have any aversion to adding default methods for all, but a user could effectively do this already by just providing noop implementations for the signals they don't need. Would improve the ergonomics and consistency though.

extensions/noop-api/src/main/java/io/opentelemetry/extension/noopapi/NoopOpenTelemetry.java
jkwatson, line 17:
```suggestion
 * at all. This means that no objects are allocated nor {@link ThreadLocal}s used in an
```

jkwatson (approved):

PR #3922, opened 2021-11-19
Catch throwable in async callbacks, and propagate if fatal

jack-berg:
Followup to this [conversation](https://github.com/open-telemetry/opentelemetry-java/pull/3895#discussion_r751829424).


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
```diff
@@ -168,6 +169,17 @@ public MetricData collectAndReset(
     }
   }
 
+  // Taken from RxJava throwIfFatal, which was taken from scala
+  private static void propagateIfFatal(Throwable t) {
```
reugn, line 173: Yes, Scala `NonFatal` would come in handy here.
Nit: worth making it a public utility method for very possible further reuse.

jkwatson: I don't think OpenTelemetry should be in the business of providing general utilities. :)

anuraaga: I can definitely see this being used elsewhere so we can have an .internal helper. But can be later too

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
```diff
@@ -168,6 +169,17 @@ public MetricData collectAndReset(
...
+    if (t instanceof VirtualMachineError) {
+      throw (VirtualMachineError) t;
```
anuraaga, line 175: Can you add unit tests? I guess the constructors are public :)

anuraaga (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
```diff
@@ -168,6 +169,17 @@ public MetricData collectAndReset(
...
+    } else if (t instanceof ThreadDeath) {
+      throw (ThreadDeath) t;
+    } else if (t instanceof LinkageError) {
+      throw (LinkageError) t;
+    }
+  }
+
```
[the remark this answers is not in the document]
anuraaga: OOME is a subclass

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
```diff
@@ -145,13 +145,14 @@ public MetricData collectAndReset(
...
         metricUpdater.run();
-      } catch (RuntimeException e) {
+      } catch (Throwable e) {
         logger.log(
             Level.WARNING,
             "An exception occurred invoking callback for instrument "
                 + getMetricDescriptor().getName()
                 + ".",
             e);
+        propagateIfFatal(e);
```
anuraaga, line 155: Let's put this above the log statement. We're already hosed so the log statement may just throw again in a weird way.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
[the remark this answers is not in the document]
jkwatson: So it is. TIL

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
jsuereth: +1, need to avoid allocations in the event you catch OOM.

jsuereth (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
anuraaga: By the way here's just one of many locations we would want to apply the new pattern

https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/MultiSpanExporter.java#L45

sdk/common/src/main/java/io/opentelemetry/sdk/internal/ThrowableUtil.java
```diff
@@ -0,0 +1,30 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.internal;
+
+/**
+ * This class is internal and is hence not for public use. Its APIs are unstable and can change at
+ * any time.
+ */
+public class ThrowableUtil {
```
anuraaga, line 12:
```suggestion
public final class ThrowableUtil {
```

sdk/common/src/main/java/io/opentelemetry/sdk/internal/ThrowableUtil.java
```diff
@@ -0,0 +1,30 @@
...
+
+  private ThrowableUtil() {}
```
anuraaga, line 14: We usually put at the end for these static utils

sdk/common/src/test/java/io/opentelemetry/sdk/internal/ThrowableUtilTest.java
```diff
@@ -0,0 +1,25 @@
...
+import org.junit.jupiter.api.Test;
+
+class ThrowableUtilTest {
+
+  @Test
+  void propagateIfFatal_Throws() {
+    assertThatCode(() -> propagateIfFatal(new RuntimeException("Error!")))
+        .doesNotThrowAnyException();
+    assertThatCode(() -> propagateIfFatal(new Exception("Error!"))).doesNotThrowAnyException();
+    assertThatCode(() -> propagateIfFatal(new VirtualMachineError("Error!") {}))
```
anuraaga, line 20: Let's use `assertByThrownBy` for readability for the throwing ones. It's interesting how assertj seems to have just aliased these two

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorageTest.java
```diff
@@ -120,7 +124,26 @@ void longAsynchronousAccumulator_AttributesProcessor_used() {
...
+    // Should catch runtime exceptions
+    assertThat(
+            collectAndReset(
+                unused -> {
+                  throw new RuntimeException("Error!");
+                }))
+        .isEqualTo(null);
+
+    // Should propagate fatal exceptions
+    assertThatCode(
```
anuraaga, line 137: Ditto (or maybe just delete, not sure if it's worth having this test case in every single location we use that helper)

jack-berg: Yeah wasn't sure what your thoughts would be on this. Ideally we'd be able to verify that the method was called, but because its static, the only way to really verify it is by duplicating the coverage from `ThrowableUtilTest`.

anuraaga (approved):

jkwatson (approved):

PR #3923, opened 2021-11-19
Remove deprecations from autoconfigure module

jack-berg:
The autoconfiguration module is looked pretty solid now with only `AutoConfiguredOpenTelemetrySdk` and its builder left as publicly accessible.


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java
```diff
@@ -1,168 +0,0 @@
...
-  @Deprecated
-  public static OpenTelemetrySdk initialize(boolean setResultAsGlobal, ConfigProperties config) {
-    return AutoConfiguredOpenTelemetrySdk.builder()
-        .setResultAsGlobal(setResultAsGlobal)
-        .setConfig(config)
-        .build()
-        .getOpenTelemetrySdk();
-  }
-
-  static OpenTelemetrySdk newOpenTelemetrySdk(
```
jack-berg, line 87: This logic was moved to `AutoConfiguredOpenTelemetrySdk#build()`, and `#configureMeterProvider` was moved to `MeterProviderConfiguration`.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ResourceConfiguration.java
```diff
@@ -5,48 +5,24 @@
...
 
-/**
- * Auto-configuration for the OpenTelemetry {@link Resource}.
- *
- * @deprecated Use {@link AutoConfiguredOpenTelemetrySdk#getResource()}.
- */
-@Deprecated
-public final class OpenTelemetryResourceAutoConfiguration {
+/** Auto-configuration for the OpenTelemetry {@link Resource}. */
+final class ResourceConfiguration {
```
jack-berg, line 20: Renamed and merged with `EnvironmentResource`.

jkwatson:
I want to hold off on this for a moment. @breedx-splk just found a use-case that we may have broken with these removals, namely being able to just get the auto-configured Resource, without building a whole SDK.

breedx-splk (commented):
Hey @jack-berg this is good stuff. Coincidentally, I just filed #3924 around the same time due to there not being a clear approach to handling the Resource from agent code. 

I would prefer that we get #3924 figured out prior to removing all these deprecations (or if we come up with a simple idea, as part of this PR maybe).

jkwatson (approved):

jack-berg:
Resolved conflicts. Should be good to go now 👍

anuraaga (approved):

PR #3927, opened 2021-11-23
make W3CTraceContextPropagator.extractContextFromTraceParent public

leviwilson:
In order to propagate a distributed trace in a lambda invocation, we needed to pass along the `traceparent` value to the lambda and create the invocation `Span` from that `traceparent` (if provided).

We didn't want to write our own parsing logic, so looking at how this repository does it we ended up copying / pasting the code.

This PR exposes `W3CTraceContextPropagator.extractContextFromTraceParent` as `public static` rather than `private static` so we can just use that.

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/trace/propagation/W3CTraceContextPropagator.java
+++ b/api/all/src/main/java/io/opentelemetry/api/trace/propagation/W3CTraceContextPropagator.java
@@ -190,7 +190,7 @@ public final class W3CTraceContextPropagator implements TextMapPropagator {
     }
   }
 
-  private static SpanContext extractContextFromTraceParent(String traceparent) {
+  public static SpanContext extractContextFromTraceParent(String traceparent) {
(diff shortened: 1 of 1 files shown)


jkwatson:
If we are going to expose this particular functionality, I don't think this is the right way to do it. I could possibly be in favor of creating a W3CUtils (or something like that) class that had the parsing functionality exposed, but even that is a bit of a stretch, I think, for what we would want to support publically as a part of our public APIs.

trask:
@leviwilson here's an alternative to copy-pasting the code:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/discussions/4546#discussioncomment-1572327

leviwilson:
@jkwatson thanks for the feedback. I wasn't sure on the "right" approach either...I was looking for some sort of `Context` builder that could achieve the trace propagation and it _looks_ like what @trask has provided is a good alternative for what I was trying to do (thanks).

I will go ahead and close this PR for now, thanks!

PR #3928, opened 2021-11-23
Remove bound instrument from metric API

### LINKED ISSUE #3913: Remove bound metric instruments
Last call for this - we have discussed removing instrument binding in the past, but currently are on track for releasing stable with them. There seems to be no mention of binding in the spec

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md

Therefore I strongly advocate for removing them from our initial stable metrics release and adding them after spec work - they don't appear to be a Java-specific concern to me where we'd have API that's not in the spec.

jack-berg:
Resolves #3913.

The bound instrument interfaces have been moved to an internal package of the SDK, since @jsuereth anticipates their use in the near future in the not-yet-specified batch API.


anuraaga:
@jack-berg I think I would move the internal Bound* interfaces into the sdk-metrics artifact for now if keeping them

jack-berg:
> @jack-berg I think I would move the internal Bound* interfaces into the sdk-metrics artifact for now if keeping them

Yup, that's what I did. Moved them to the `io.opentelemetry.sdk.metrics.internal.instrument` package in the `:sdk:metrics` module.

anuraaga:
Oops - the package names were more obvious than the file paths I was trying to read!

anuraaga (approved):

jkwatson (approved):

PR #3935, opened 2021-11-25
Move gRPC TLS exporter tests into base class and use in grpc-classpat…

anuraaga:
…h test sets.

Also makes small tweaks to logs for consistency that were uncovered by the tests


anuraaga:
The deleted tests included compression tests I didn't port over for same reason as here

https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/otlp/trace/src/testOkhttpOnly/java/io/opentelemetry/exporter/otlp/trace/OkHttpOnlyExportTest.java#L85

They're covered in the integration tests so no need to worry about them here given they're tricky to test for the okhttp-only case

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/grpc/DefaultGrpcExporter.java
```diff
@@ -89,7 +89,7 @@ public void onFailure(Throwable t) {
                         + "This usually means that your collector is not configured with an otlp "
                         + "receiver in the \"pipelines\" section of the configuration. "
                         + "Full error message: "
-                        + t.getMessage());
```
jkwatson, line 92: why get rid of the throwable message here?

jkwatson (approved):

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/grpc/DefaultGrpcExporter.java
anuraaga: I don't think it was intended - the throwable is only wrapping Status, so it's message has some extra boilerplate that causes it not to match the other implementation with no extra information.

PR #3936, opened 2021-11-25
Implement retry for OkHttpGrpcExporter

anuraaga:


jack-berg:
FYI, I've been working on something similar for the okhttp http/protobuf otlp exporters.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/okhttp/OkHttpRetrier.java
```diff
@@ -0,0 +1,86 @@
...
+  private final Function<Response, Boolean> isRetriableError;
+  private final Sleeper sleeper;
+  private final BoundedLongGenerator randomLong;
+
+  /** Constructs a new retrier. */
+  public OkHttpRetrier(RetryPolicy retryPolicy, Function<Response, Boolean> isRetriableError) {
+    this(
+        retryPolicy,
+        isRetriableError,
+        TimeUnit.NANOSECONDS::sleep,
```
jack-berg, line 29: Nice way to improve testability!

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/okhttp/OkHttpRetrier.java
```diff
@@ -0,0 +1,86 @@
...
+      return response;
+    }
+    long nextBackoffNanos = retryPolicy.getInitialBackoff().toNanos();
+    // Made an attempt already, starting with the 2nd.
+    for (int attempt = 2; attempt <= retryPolicy.getMaxAttempts(); attempt++) {
+      // https://github.com/grpc/proposal/blob/master/A6-client-retries.md#exponential-backoff
+      // Careful - the document says "retry attempt" but our maxAttempts includes the original
+      // request. The numbers are tricky but the unit test should be more clear.
+      long upperBoundNanos = Math.min(nextBackoffNanos, retryPolicy.getMaxBackoff().toNanos());
+      long backoffNanos = randomLong.get(upperBoundNanos);
```
jack-berg, line 58: I think this logic is correct, but IMO it reads cleaner if its extracted into a helper function. Here's what I have locally for a `http/protobuf` retry interceptor:
```
private static int backoffMillisWithJitter(int attempt, RetryPolicy retryPolicy) {
    return RANDOM.nextInt(backoffMillisTarget(attempt, retryPolicy));
  }

  // Visible for testing
  static int backoffMillisTarget(int attempt, RetryPolicy retryPolicy) {
    long initialBackoffMillis = retryPolicy.getInitialBackoff().toMillis();
    if (attempt == 1) {
      return (int) initialBackoffMillis;
    }
    long target =
        (long) (initialBackoffMillis * Math.pow(retryPolicy.getBackoffMultiplier(), attempt));
    return (int) Math.min(target, retryPolicy.getMaxBackoff().toMillis());
  }
```
Would likely need to be extracted to a helper function once the `http/protobuf` retry stuff lands anyway.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/okhttp/OkHttpRetrier.java
```diff
@@ -0,0 +1,86 @@
...
+
+      try {
+        sleeper.sleep(backoffNanos);
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        return response;
```
jack-berg, line 64: Do we want to return here or raise a runtime exception?

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/okhttp/OkHttpRetrier.java
```diff
@@ -0,0 +1,86 @@
...
+import io.opentelemetry.exporter.otlp.internal.RetryPolicy;
+import java.io.IOException;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
+import okhttp3.Interceptor;
+import okhttp3.Response;
+
+/** Retrier of OkHttp requests. */
+public final class OkHttpRetrier implements Interceptor {
```
jack-berg, line 17: `RetryInterceptor`? `OkHttp` is implied because it implements the okhttp `Interceptor` interface.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/okhttp/OkHttpRetrier.java
anuraaga: Sorry for writing this while you also had code in the work! I was thinking that the only difference between the two would be the definition if `isRetryable` but otherwise the same intercceptor could be reused (I don't think it hurts to have support for retry-after even in the one used by gRPC). See anything that would prevent that?

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/okhttp/OkHttpRetrier.java
anuraaga: From what I understand calling interrupt and returning is the correct way of handling an interrupt exception, though I've never really run into it differing much from throwing in practice

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/okhttp/OkHttpRetrier.java
anuraaga: Also I had a similar pattern but changed to this one to avoid redundant math in Math.pow, similar to what gRPC code was doing. I don't think it's a huge deal but would keep it for that reason if it is possible to reuse the interceptor instead of just a helper method.

jack-berg: No problem! I do think this interceptor can be reused. 

One note: I believe that if `chain.proceed(Request)` is called and throws an `IOException` that we should retry, rather than letting it propagate to `Callback#onFailure(Call, IOException)`.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/okhttp/OkHttpRetrier.java
jack-berg: Yeah I think in practice the thread will only be interrupted at shutdown so it shouldn't matter much that we return.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/okhttp/OkHttpRetrier.java
anuraaga: Unfortunately it will throw IOException for things like a bad TLS certificate which we shouldn't retry on. Some smart checking of `instanceof` on the exception might work but I'm not sure the best way to implement that right now.

anuraaga: Oh yeah, OkHttp has automatic retry on connection failure so I guess we can generally live without retrying on IOException? https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/-builder/retry-on-connection-failure/

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/grpc/OkHttpGrpcExporter.java
```diff
@@ -229,6 +226,18 @@ public CompletableResultCode shutdown() {
     return CompletableResultCode.ofSuccess();
   }
 
+  static boolean isRetryable(Response response) {
+    // Only retry on gRPC codes which will always come with an HTTP success
```
anuraaga, line 230: @jack-berg BTW, this is unlikely to be the best behavior (i.e., 503 seems like an obvious one to retry on) - gRPC retry lives in a bubble that misses standard HTTP load balancers. But since spec doesn't mention HTTP codes for gRPC retry, went with this for now.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/okhttp/OkHttpRetrier.java
jack-berg: Yeah, probably fine ignoring IOException.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/grpc/OkHttpGrpcExporter.java
jack-berg: Yeah that's a good point. There could be a load balancer in front of the gRPC server that gets in the way of the gRPC bubble.

jack-berg (approved):

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/okhttp/RetryInterceptor.java
```diff
@@ -0,0 +1,86 @@
...
+      return response;
+    }
+    long nextBackoffNanos = retryPolicy.getInitialBackoff().toNanos();
+    // Made an attempt already, starting with the 2nd.
+    for (int attempt = 2; attempt <= retryPolicy.getMaxAttempts(); attempt++) {
+      // https://github.com/grpc/proposal/blob/master/A6-client-retries.md#exponential-backoff
+      // Careful - the document says "retry attempt" but our maxAttempts includes the original
+      // request. The numbers are tricky but the unit test should be more clear.
+      long upperBoundNanos = Math.min(nextBackoffNanos, retryPolicy.getMaxBackoff().toNanos());
+      long backoffNanos = randomLong.get(upperBoundNanos);
```
ayushtkn, line 58: This can return any number like from 0 to upperBoundNanos, and in worst case all the retries can go off in like 1 or two nanos.

As per the doc attached as comment, the jitter should be 0.2, this doesn't operate that way

trask: @ayushtkn can you open an issue for this, so your suggestion isn't lost? thanks!

ayushtkn: Sure I created one: https://github.com/open-telemetry/opentelemetry-java/issues/7004

PR #3939, opened 2021-11-29
Remove supported temporality

jack-berg:
Per [recent change](https://github.com/open-telemetry/opentelemetry-specification/pull/2154) to the spec.

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/TemporalityUtils.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/TemporalityUtils.java
@@ -6,7 +6,6 @@
 package io.opentelemetry.sdk.metrics.internal.state;
 
 import io.opentelemetry.sdk.metrics.data.AggregationTemporality;
-import java.util.EnumSet;
 import javax.annotation.Nullable;
 
 final class TemporalityUtils {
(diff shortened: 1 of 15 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/TemporalityUtils.java
```diff
@@ -15,25 +14,14 @@ private TemporalityUtils() {}
...
-   * @param supported All aggregation temporalities supported by the exporter.
    * @param preferred The preferred temporality of the exporter.
    */
-  static AggregationTemporality resolveTemporality(
-      EnumSet<AggregationTemporality> supported, @Nullable AggregationTemporality preferred) {
-    // Next assume preferred should always win.
+  static AggregationTemporality resolveTemporality(@Nullable AggregationTemporality preferred) {
+    // If preferred temporality is provided respect it
     if (preferred != null) {
       return preferred;
```
jkwatson, line 22: we could just do a simple one-liner for this whole method now, right?
`return preferred == null ? AggregationTemporality.CUMULATIVE : preferred;`

jkwatson (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/TemporalityUtils.java
jack-berg: Yes. Will update.

anuraaga (approved):

PR #3945, opened 2021-12-01
Enable animal sniffer on metrics

jack-berg:
Came across a TODO in the code to enable animal sniffer once we're closer to releasing a stable version and figured now is the time.

Is there any better alternative to replace `LongAdder/DoubleAdder` than `AtomicLong` ?


anuraaga:
We can't replace the *Adder for performance reasons. The options are

1) Have counter interfaces that fallback to Atomic* if the Adder class can't be found
2) Vendor in Adder classes
3) Remove TODO and do nothing, if metrics on Android has no use case

@jkwatson Any thoughts in particular about 3)?

jkwatson:
I think there will probably be use cases in Android, although I don't know what they might be. We should have a really good story on why we don't support them if we aren't going to. 

I would prefer an Android friendly fallback if at all possible. High performance will be much less of an issue on mobile, I believe.

anuraaga:
Cool - let's go with 1)

jack-berg:
Sounds good. Will get this sorted out today 👍

sdk/metrics/build.gradle.kts
```diff
@@ -48,3 +46,7 @@ tasks {
     dependsOn(testing.suites)
   }
 }
+
+animalsniffer {
+  ignore("java.util.concurrent.atomic.DoubleAdder", "java.util.concurrent.atomic.LongAdder")
```
jack-berg, line 51: I tried using the `@IgnoreJRERequirement` annotation on the classes that use `j.u.c.a.DoubleAdder` and `j.u.c.a.LongAdder` but it didn't work. It got rid of of some of the animal sniffer errors, but there was always one pesky error remaining.

Added the annotation at the class level like so:
```
@IgnoreJRERequirement
final class JreLongAdder implements LongAdder {

  private final java.util.concurrent.atomic.LongAdder delegate =
      new java.util.concurrent.atomic.LongAdder();
...
```

And still see errors like:
```
[Undefined reference] io.opentelemetry.sdk.metrics.internal.concurrent.(JreDoubleAdder.java::1)
  >> java.util.concurrent.atomic.DoubleAdder
```

Not sure what I'm doing wrong, but this allowed me to get around the errors with what I believe is a safe implementation.

anuraaga: I will play around a bit too. A Gradle based solution is OK too but I hope we can specify our classes to ignore from checking, not the JDK classes themselves. Otherwise the latter can easily get used outside those classes accidentally.

jack-berg: > Otherwise the latter can easily get used outside those classes accidentally.

Yeah exactly. Definitely prefer to do this via annotation if possible.

anuraaga (approved):
@jack-berg Found a way to exclude our classes, rather than the used classes, from animalsniffer. Also tweaked the keySet fix

jkwatson (approved):

PR #3949, opened 2021-12-02
Rename .observe to .record

### LINKED ISSUE #3914: Rename ObservableMeasurement.observe -> ObservableMeasurement.record?
I feel as if `Consumer<Measurement>` is the observer and doing the observing - the method on the callback seems like it's actually recording, and is meant to be "semantically" the same as the record methods on the synchronous instruments.

anuraaga:
Also removed MUST be non-negative from docs, this is true only of Counter instruments I think.

Fixes #3914

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/metrics/ObservableDoubleMeasurement.java
+++ b/api/all/src/main/java/io/opentelemetry/api/metrics/ObservableDoubleMeasurement.java
@@ -13,7 +13,9 @@ public interface ObservableDoubleMeasurement extends ObservableMeasurement {
    * Records a measurement.
    *
    * @param value The measurement amount.
+   * @deprecated Use {@link #record(double)}.
    */
+  @Deprecated
   default void observe(double value) {
     record(value);
   }
@@ -23,7 +25,9 @@ public interface ObservableDoubleMeasurement extends ObservableMeasurement {
    *
    * @param value The measurement amount.
    * @param attributes A set of attributes to associate with the count.
+   * @deprecated Use {@link #record(double, Attributes)}.
    */
+  @Deprecated
   default void observe(double value, Attributes attributes) {
     record(value, attributes);
   }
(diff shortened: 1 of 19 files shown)


jack-berg (approved):
The plan is to leave these deprecated for the release candidate, and remove them for the next published version (which will be the first version which includes metrics as part of the stable API)?

jack-berg:
This should resolve #3914.

jkwatson (approved):

api/all/src/main/java/io/opentelemetry/api/metrics/ObservableDoubleMeasurement.java
```diff
@@ -12,15 +12,38 @@
...
   /**
    * Records a measurement with a set of attributes.
    *
-   * @param value The measurement amount. MUST be non-negative.
+   * @param value The measurement amount.
    * @param attributes A set of attributes to associate with the count.
+   * @deprecated Use {@link #record(double, Attributes)}.
    */
-  void observe(double value, Attributes attributes);
+  @Deprecated
```
breedx-splk, line 30: should we include a message about what to use instead?

api/all/src/main/java/io/opentelemetry/api/metrics/ObservableDoubleMeasurement.java
```diff
@@ -12,15 +12,38 @@
   /**
    * Records a measurement.
    *
-   * @param value The measurement amount. MUST be non-negative.
+   * @param value The measurement amount.
+   * @deprecated Use {@link #record(double)}.
    */
-  void observe(double value);
+  @Deprecated
```
breedx-splk, line 18: Include a message about using `record(value)` instead?

breedx-splk (approved):
Looks ok to me, but I didn't verify the MUST/non-negative thing. Would be nice to have a note in the deprecation about using the new name. 🥂

api/all/src/main/java/io/opentelemetry/api/metrics/ObservableDoubleMeasurement.java
anuraaga: Yeah it's there in the javadoc, this is the annotation 😛

PR #3951, opened 2021-12-02
Fix logs exporter module name

elevenzhan:
Fix logs exporter module name and description

### DIFF
--- a/exporters/otlp/logs/src/main/java/io/opentelemetry/exporter/otlp/logs/MarshalerLogsServiceGrpc.java
+++ b/exporters/otlp/logs/src/main/java/io/opentelemetry/exporter/otlp/logs/MarshalerLogsServiceGrpc.java
@@ -17,7 +17,9 @@ import io.opentelemetry.exporter.otlp.internal.grpc.MarshalerServiceStub;
 import io.opentelemetry.exporter.otlp.internal.logs.LogsRequestMarshaler;
 import java.io.InputStream;
 
-// Adapted from the protoc generated code for LogsServiceGrpc.
+/**
+ * Adapted from the protoc generated code for LogsServiceGrpc.
+ */
 final class MarshalerLogsServiceGrpc {
 
   private static final String SERVICE_NAME = "opentelemetry.proto.collector.logs.v1.LogsService";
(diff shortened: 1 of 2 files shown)


jack-berg (approved):
Thanks!

breedx-splk (approved):
Yikes. 😁 Thanks thanks.

exporters/otlp/logs/src/main/java/io/opentelemetry/exporter/otlp/logs/MarshalerLogsServiceGrpc.java
```diff
@@ -17,7 +17,7 @@
 import io.opentelemetry.exporter.otlp.internal.logs.LogsRequestMarshaler;
 import java.io.InputStream;
 
-// Adapted from the protoc generated code for LogsServiceGrpc.
+/** Adapted from the protoc generated code for LogsServiceGrpc. */
```
anuraaga, line 20: This isn't really javadoc so a comment it more appropriate. Let's revert this change

jkwatson: I realized my comment was in "pending status". 🤦🏽

elevenzhan: Thanks, but I think this style is aligned with other source codes in project; 
In addition, Javadoc single-line comment can not pass spotlessJavaCheck, just like this:
```
/**
 * Adapted from the protoc generated code for LogsServiceGrpc.
 */
```
Do I need revert this change or keep it, @anuraaga @jkwatson

anuraaga: > In addition, Javadoc single-line comment can not pass spotlessJavaCheck, just like this:

I don't follow this - spotless is currently passing right? Let's go ahead and rever tthis - this is an internal class so it doesn't need real javadoc, this is an implementation detail comment targeted at code authors, not docs for end users. We like to make that distinction where possible.

elevenzhan: I see, already revert

jkwatson (approved):

PR #3953, opened 2021-12-03
Migrate most testsets to test suites.

anuraaga:


context/build.gradle.kts
```diff
@@ -70,7 +70,7 @@ testing {
 }
 
 tasks {
-  named("check") {
```
Oberon00, line 73: I think this should be kept as-is to profit from conviguration-avoidance: https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:old_vs_new_configuration_api_overview

anuraaga: It's not clear from the docs but it seems the kotlin dsl for these seems to use the lazy configuration APIs based on code navigating. Switched for the compile time safety

exporters/otlp/trace/build.gradle.kts
```diff
@@ -53,8 +27,50 @@ dependencies {
...
+    }
+    val testGrpcOkhttp by registering(JvmTestSuite::class) {
+      dependencies {
+        implementation(project(":exporters:otlp:testing-internal"))
+
+        implementation("io.grpc:grpc-okhttp")
+        implementation("io.grpc:grpc-stub")
+      }
+    }
+    val testSpanPipeline by registering(JvmTestSuite::class) {
```
Oberon00, line 56: I think if we do not use the result of registering, then the `by registering` syntax looks odd. Isn't there some `register(JvmTestSuide::class)` API that could be used instead?

anuraaga: `register` with a string is the alternative. I find it more consistent to just always use this form even when not using the variable, here we happen to not have any need for variable form but if we happened to mix the two patterns it is confusing I think.

jack-berg (approved):

PR #3960, opened 2021-12-06
Configurable temporality

jack-berg:
According to the spec, the [in-memory](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/in-memory.md) and [standard output](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/stdout.md) exporters must allow aggregation temporality to be specified. This makes preferred temporality configurable at creation time.

### DIFF
--- a/exporters/logging/src/main/java/io/opentelemetry/exporter/logging/LoggingSpanExporter.java
+++ b/exporters/logging/src/main/java/io/opentelemetry/exporter/logging/LoggingSpanExporter.java
@@ -18,6 +18,9 @@ import java.util.logging.Logger;
 public final class LoggingSpanExporter implements SpanExporter {
   private static final Logger logger = Logger.getLogger(LoggingSpanExporter.class.getName());
 
+  /** Class constructor. */
+  public LoggingSpanExporter() {}
+
   @Override
   public CompletableResultCode export(Collection<SpanData> spans) {
     // We always have 32 + 16 + name + several whitespace, 60 seems like an OK initial guess.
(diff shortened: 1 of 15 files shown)


exporters/logging/src/main/java/io/opentelemetry/exporter/logging/LoggingSpanExporter.java
```diff
@@ -18,6 +18,9 @@
 public final class LoggingSpanExporter implements SpanExporter {
   private static final Logger logger = Logger.getLogger(LoggingSpanExporter.class.getName());
 
+  /** Class constructor. */
+  public LoggingSpanExporter() {}
```
jack-berg, line 22: Noticed that there are public constructors in `:exporters:logging`. Thought about deprecating them, and adding `public static LoggingSpanExporter create()` for symmetry, but figured that would just be clutter since we can never actually get rid of the public constructor. 

Instead, I opted to at least make the public constructors explicit.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/testing/InMemoryMetricExporterTest.java
```diff
@@ -99,7 +108,7 @@ void testShutdown_export() {
   }
 
   @Test
-  void test_flush() {
+  void flush() {
```
jack-berg, line 111: I normally try to stick with the test naming convention of the file I'm modifying, but in this case the `test_` prefix is purely redundant.

exporters/logging/src/main/java/io/opentelemetry/exporter/logging/LoggingSpanExporter.java
anuraaga: Doh - I think it's fine to add the create and deprecate to at least show the consistent pattern within our codebase.

jsuereth (approved):

anuraaga (approved):

PR #3966, opened 2021-12-07
Drop unused CountAggregator

jack-berg:
The `CountAggregator` was unused and not mentioned in the [sdk spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#aggregation).


breedx-splk (commented):
(Disclaimer: I'm a metrics n00b and could definitely be wrong here)

My read is that Views can be customized with an Aggregation, and that this is an aggregator that just produces a count. The [example given in the View spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#view) talks about HTTP response codes being built as a histogram but some users might want a view that aggregates as a count (eg. to show a count of outgoing responses). Seems like a good thing to have...

Is it possible to do the same thing maybe with one of the other aggregations? Or maybe this is a hard line stance of "If it isn't spec'd get rid of it"?

jkwatson:
> (Disclaimer: I'm a metrics n00b and could definitely be wrong here)
> 
> My read is that Views can be customized with an Aggregation, and that this is an aggregator that just produces a count. The [example given in the View spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#view) talks about HTTP response codes being built as a histogram but some users might want a view that aggregates as a count (eg. to show a count of outgoing responses). Seems like a good thing to have...
> 
> Is it possible to do the same thing maybe with one of the other aggregations? Or maybe this is a hard line stance of "If it isn't spec'd get rid of it"?

This class wasn't public or instantiatable by the end-user. When we change to support custom aggregations, then someone will be able to write this trivially, as needed.

breedx-splk:
> This class wasn't public or instantiatable by the end-user. When we change to support custom aggregations, then someone will be able to write this trivially, as needed.

Cool, thanks for the clarification. I missed the non-public thing (I should have assumed, LOL)

breedx-splk (approved):

anuraaga (approved):

PR #3967, opened 2021-12-08
Migrate to official Gradle GitHub action.

anuraaga:


iNikem:
Out of curiosity: why this change?

anuraaga:
@iNikem Noticed the official action, since we disable the remote cache support of the old one anyways, this seems to have similar features but more active maintenance.

PR #3973, opened 2021-12-08
Add @IgnoreJRERequirement to JreDoubleAdder

jack-berg:
We omitted `@IgnoreJRERequirement` on one class in #3945.

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/concurrent/JreDoubleAdder.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/concurrent/JreDoubleAdder.java
@@ -5,6 +5,9 @@
 
 package io.opentelemetry.sdk.metrics.internal.concurrent;
 
+import org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement;
+
+@IgnoreJRERequirement
 final class JreDoubleAdder implements DoubleAdder {
 
   private final java.util.concurrent.atomic.DoubleAdder delegate;


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/concurrent/JreDoubleAdder.java
```diff
@@ -5,6 +5,9 @@
 
 package io.opentelemetry.sdk.metrics.internal.concurrent;
 
+import org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement;
+
+@IgnoreJRERequirement
```
anuraaga, line 10: This annotation unfortunately isn't doing anything (we had to exclude the class entirely). We could still add it but I guess better to remove from JreLongAdder which I left in accidentally I think.

jack-berg: Ah I see. I knew it didn't do anything but thought you had added it to `JreLongAdder` just as a reminder to future maintainers. 

Happy to remove it from both places.

anuraaga: Keeping in both as a reminder makes sense to me, I'm OK with either. To clarify, I had no lofty ideas in mind when keeping in JreLongAdder :P

anuraaga (approved):

PR #3977, opened 2021-12-09
Consistent behavior when metric instruments violate rules

### LINKED ISSUE #3861: Consistent behavior when metric instruments violate rules
Discussed in Issue #3518 [here](https://github.com/open-telemetry/opentelemetry-java/issues/3518#issuecomment-965576185), but creating own issue to track.

Currently, async instruments can record values that produce `MetricData` which violate their monotonic properties. For example:
```
AtomicLong count = new AtomicLong(100);
    meter
        .counterBuilder("async-counter")
        .buildWithCallback(
            measurement ->
                measurement.observe(count.decrementAndGet()));
```
When export is cumulative, this will produce sequential metric data that are monotonic but which have decreasing values. When export is delta, this will produce monotonic metric data whose values are negative.

This behavior which trusts that the callback is well behaved is inconsistent with synchronous counts (`SdkLongCounter`, `SdkDoubleCounter`), which throw if you try to record negative values.

Minimally, the behavior should be consistent. As suggested by @jsuereth, we also might allow the SDK to run in a verify mode, which throws when rules are violated. And when verify is disabled, produce diagnostic log messages instead.

jack-berg:
Addresses several possible violations:
- Counters can't add negative values. Current behavior is to throw. New behavior is to log with a throttling logger.
- Histograms can't record negative values. Current behavior is to allow it. New behavior is to log with a throttling logger.
- Async instruments can't record duplicate values for the same key. Current behavior is to allow it. New behavior is to log with a throttling logger. 
- Observable sums must be monotonically increasing. Current behavior is to allow it. New behavior is to log with a throttling logger, if and only if the default sum aggregation is used.

I'm very much unsure about my implementation of the check for the fourth violation. Synchronous instruments don't need access to previous recordings in order to ensure the rules are followed, but observable sums do since you need to compare the current recorded value to the previous. My implementation adds this check to the `AsynchronousMetricStorage`, which retains a copy of previous accumulations, and asks the configured `Aggregator` if the current accumulation is valid given the previous. The problem with this is that the check is ignored if you switch the aggregator to something else, such as last value. But there isn't any other natural place to store the previous value such that a check can be consistently implemented.

I included it in this PR to prompt discussion, but am not committed to it. 

Resolves #3861.


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkDoubleCounter.java
```diff
@@ -30,16 +35,15 @@ private SdkDoubleCounter(InstrumentDescriptor descriptor, WriteableMetricStorage
...
   public void add(double increment, Attributes attributes, Context context) {
-    BoundStorageHandle aggregatorHandle = storage.bind(attributes);
-    try {
-      if (increment < 0) {
-        throw new IllegalArgumentException("Counters can only increase");
-      }
-
-      aggregatorHandle.recordDouble(increment, attributes, context);
-    } finally {
-      aggregatorHandle.release();
```
jack-berg, line 41: This is already taking place in `storage.recordDouble`.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/Aggregator.java
```diff
@@ -74,6 +74,15 @@ default T accumulateDoubleMeasurement(double value, Attributes attributes, Conte
     return handle.accumulateThenReset(attributes);
   }
 
+  /**
+   * Validate that the asynchronous {@code currentAccumulation} is compatible with the {@code
+   * previousAccumulation}.
+   *
+   * @throws Exception if the {@code currentAccumulation} is invalid
+   */
+  default void validateAsyncAccumulation(T previousAccumulation, T currentAccumulation)
```
jack-berg, line 83: The only aggregator that implements this are the sum aggregators `DoubleSumAggregator` and `LongSumAggregator`, which is a smell. 

This might be more generally useful as a reset detection mechanism. I.e. during collection we ask the aggregator if the current accumulation is compatible with the previous accumulation, and if not, we indicate that a reset a reset in the `MetricData`.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
```diff
@@ -92,13 +92,16 @@ public void record(double value) {
...
-    final MetricDescriptor metricDescriptor = MetricDescriptor.create(view, instrument);
+    MetricDescriptor metricDescriptor = MetricDescriptor.create(view, instrument);
     Aggregator<T> aggregator =
         view.getAggregation().createAggregator(instrument, ExemplarFilter.neverSample());
-    final AsyncAccumulator<T> measurementAccumulator = new AsyncAccumulator<>(instrument);
-    final AttributesProcessor attributesProcessor = view.getAttributesProcessor();
+    AsyncAccumulator<T> measurementAccumulator = new AsyncAccumulator<>(instrument, aggregator);
+    if (Aggregator.drop() == aggregator) {
+      return empty();
+    }
```
jack-berg, line 101: Not sure why this optimization made it to `doubleAsynchronousAccumulator` but not `longAsynchronousAccumulator`.

Got rid of all these the `final` because they reduce readability and don't do anything. If compile time checks are smart enough to know if a `final` variable has been mutated, then the compiler can also be smart enough to apply final optimizations to any variable that is effectively final, but doesn't have the `final` keyword.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
```diff
@@ -194,14 +202,41 @@ public void record(Attributes attributes, T accumulation) {
                 + ").");
         return;
       }
-      // TODO: error on metric overwrites
+
+      // Check there is not already a recording for the attributes
```
anuraaga, line 206: I can't find making this an error in the spec

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#asynchronous-updowncounter-creation

`User code is recommended not to provide more than one Measurement with the same attributes in a single callback. If it happens, the SDK MAY decide how to handle it. For example, during the callback invocation if two measurements value=1, attributes={pid:4, bitness:64} and value=2, attributes={pid:4, bitness:64} are reported, the SDK can decide to simply let them pass through (so the downstream consumer can handle duplication), drop the entire data, pick the last one, or something else.`

I personally am not a fan of dropping (strongly) or logging (less strongly) for this, it's similar to how `Span.setAttribute` can be called with the same key multiple times, rarely needed but seems overstepping by treating as an error.

anuraaga (approved):
Do you feel like asserting the log messages?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkDoubleCounter.java
```diff
@@ -19,8 +20,12 @@
 import io.opentelemetry.sdk.metrics.internal.state.MeterSharedState;
 import io.opentelemetry.sdk.metrics.internal.state.WriteableMetricStorage;
 import java.util.function.Consumer;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 final class SdkDoubleCounter extends AbstractInstrument implements DoubleCounter {
+  private static final ThrottlingLogger logger =
```
anuraaga, line 27: I guess it's better to have a static Logger, non-static ThrottlingLogger to prevent logs from instrument blocking another one forever based on the timing.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
jack-berg: That's the line that I was going off of. The current behavior is "last recording wins", and this PR changes it to "first recording wins, and log". 

I don't have a strong preference for whether first or last wins, but I think that logging (at some level) is important because without logging it could be difficult to diagnose.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
```diff
@@ -194,14 +202,41 @@ public void record(Attributes attributes, T accumulation) {
...
+          // TODO: consider resetting the metric stream instead of logging here
+          logger.log(
+              Level.WARNING,
+              "Instrument "
+                  + instrument.getName()
+                  + " has recorded an invalid value: "
+                  + e.getMessage());
+          return;
+        }
+      }
```
jack-berg, line 232: @jsuereth curious what you think about this. Repeating a blurb from the PR description for context:

Observable sums must be monotonically increasing. Current behavior is to allow it. New behavior is to log with a throttling logger, if and only if the default sum aggregation is used.

Synchronous instruments don't need access to previous recordings in order to ensure the rules are followed, but observable sums do since you need to compare the current recorded value to the previous. My implementation adds this check to the AsynchronousMetricStorage, which retains a copy of previous accumulations, and asks the configured Aggregator if the current accumulation is valid given the previous. The problem with this is that the check is ignored if you switch the aggregator to something else, such as last value. But there isn't any other natural place to store the previous value such that a check can be consistently implemented.

jsuereth: Do you still report the value?

It is allowed, and acceptable, for cumulative sum to "roll" i.e.. go past int max and return int min.  These are commonly used to compute rates at query time, so they're always diffed and backends generally have wrap-around protection logic.

While I understand what you want to do here, I think we could allow users to provide any value and document expectations.  We shouldn't limit async in this way unless we're positive we're not preventing legitimate uses (which is almost impossible)

jsuereth:
Sorry I missed this earlier.  Been dealing with log4j fun.  I'll try to make sure I give a full review early tomorrow

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
jack-berg: Yeah I mentioned in the PR description but I'll say it again, I'm very unsure about the value of this check. I found a situation where you can use the API / SDK to produce OTLP data that violates the monotonicity, which seems wrong. Made the change to prompt conversation so we don't accidentally release something we regret (FYI, this only affects the SDK, not the API). 

My main problem is that the check only actually works if you're using an async counter and accumulating with the default sum aggregator. So if you chance to a different aggregator, you can suddenly record non-monotonic measurements. 

I don't see any obvious resolution to this. One thought I had is that maybe in certain edge cases its ok that SDKs produce OTLP data that violate rules, and its just up to backends to detect this and react. (They could either reject the data or reset the stream, etc)

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkDoubleCounter.java
jsuereth: Good catch!

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkLongHistogram.java
```diff
@@ -28,6 +34,14 @@ private SdkLongHistogram(InstrumentDescriptor descriptor, WriteableMetricStorage
 
   @Override
   public void record(long value, Attributes attributes, Context context) {
+    if (value < 0) {
+      throttlingLogger.log(
```
jsuereth, line 38: Regarding dropping the point completely, I'm still on the fence here.

1. we *could* update this message to explain how this can break prometheus exporters which it would)
2. OTLP (while violated today) actually handles the situation just fine.  We just can't determine if prometheus-export is ok.

I think this represents the best "first release" option, as at least users understand this could be an issue w/ Prometheus compatibility.   It seems like a restriction we can ease once we figure out a better mechanism for prometheus compatibility here.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/Aggregator.java
jsuereth: You'd also want to do this for Histogram if we ever support gauge/async histograms.  The "sum" and "count" in histogram need to match w/ Prometheus sum/count sub-metrics for compatibility.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleSumAggregator.java
```diff
@@ -55,6 +55,15 @@ public DoubleAccumulation accumulateDoubleMeasurement(
     return DoubleAccumulation.create(value);
   }
 
+  @Override
+  public void validateAsyncAccumulation(
+      DoubleAccumulation previousAccumulation, DoubleAccumulation currentAccumulation)
+      throws Exception {
+    if (isMonotonic() && currentAccumulation.getValue() < previousAccumulation.getValue()) {
+      throw new Exception("Values must be monotonically increasing.");
```
jsuereth, line 63: [blocking] Very much against throwing here for various reasons.


1. It is *expected* that counters are allowed to roll-over (i.e. go from Int.Max to Int.Min).  Backends detect this reset.   At a minimum we can detect this issue and reset a start timestamp when it happens.   
2. We *should not prevent this use case*.  For high-rate counters (e.g. request count during high load) this will lead to exceptions in async insturments.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
jsuereth: The primary issue I have with the check is it makes an important metrics use-case broken.  Specifically counter-roll-over.   Unless you can verify the counter drop is not the result of int overflow, we should not be doing this check.

jsuereth (changes requested):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkLongHistogram.java
jack-berg: > I think this represents the best "first release" option

Right. Better to have strict requirements that we ease, since it will be hard / impossible to add restrictions later.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleSumAggregator.java
jack-berg: > At a minimum we can detect this issue and reset a start timestamp when it happens.

I don't believe we have any precedent for reseting the start timestamp yet - I'm interested in ideas on how this would work. 

> this will lead to exceptions in async insturments.

It's only throwing because throwing and catching is the idiomatic way to indicate an error has occurred. In reality, the only caller of this (AsynchronousMetricStorage) catches and logs, and skips recording the value when this occurs. 

We talked in this in the 12/16 6pm PST SIG. The SDK is in a tough spot with async counters because there's no natural way to validate that values are monotonically increasing. This approach checks in the aggregator, which means that the validation rules change based on how you're aggregating, which isn't right - an async counter doesn't all the sudden get to break monotonicity because the aggregation is last value. And to your point, roll-overs are valid - how do we differentiate between valid and invalid?

But doing nothing means that the SDK can emit OTLP data that violates the monotonicity description. An approach I think is reasonable (and others may have agreed - can't quite remember) is to do nothing here. By doing nothing we take the stance that the SDK is going to do its best to validate recordings and produce valid data. But occasional edge cases like this one may prevent that. It's up to backends to handle this in an appropriate way.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
jack-berg: Replied [here](https://github.com/open-telemetry/opentelemetry-java/pull/3977#discussion_r771641874).

TL;DR is that I think we should remove this check. By doing so we are accepting that in certain edge cases the SDK will produce data that violates rules.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleSumAggregator.java
jsuereth: > I don't believe we have any precedent for reseting the start timestamp yet - I'm interested in ideas on how this would work.

We have some preliminary discussion from > 1 year ago, when the data-model SiG was meeting regularly.  Specifically, we do discuss [reset here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#resets-and-gaps).   We decided, in the end, to attempt to model resets using purely timestamps.  This is purely our datamodel.  There are practical implications to this.


If you're interested in the design space and want to look into solutions, I suggest diving into [this issue](https://github.com/prometheus/prometheus/issues/6669) which dives deep into practical concerns with async counters and expectations of reset in SUMs *plus* the major issue happens from OpenMetrics/Prometheus forcing positive-only histograms.

Additionally, take look at why cumulative values dropping is ok in practice:  the [resets](https://prometheus.io/docs/prometheus/latest/querying/functions/#resets) function in PromQL.

There's a great discussion to be had over whether the SDK should do reset-like funcitonality to ensure the PromQL `reset` method and our start/stop timestamps (specified in our data model) align.   AFAIK, not many backends use (only) timestamps for reset detection.  But most will use value drops (particularly anything touting promql compatibility).

> We talked in this in the 12/16 6pm PST SIG. The SDK is in a tough spot with async counters because there's no natural way to validate that values are monotonically increasing. This approach checks in the aggregator, which means that the validation rules change based on how you're aggregating, which isn't right - an async counter doesn't all the sudden get to break monotonicity because the aggregation is last value. And to your point, roll-overs are valid - how do we differentiate between valid and invalid?

My point is I don't think we can.   We should assume good intent from the user (and document expectations) in these situations.  If you come up with something that won't break prometheus expectations, then I'd be all for something better.

TL;DR; I still think doing nothing here is better than preventing recordings.  If you'd like to log (and still record) that's fine as well.

jack-berg:
I've rolled back the check trying to confirm that async counters only record montonically increasing values. 

PR should be good to go now 👍

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleSumAggregator.java
jack-berg: > TL;DR; I still think doing nothing here is better than preventing recordings. If you'd like to log (and still record) that's fine as well.

I've rolled it back. The log message isn't worth the complexity to implement.

jsuereth (approved):

PR #3982, opened 2021-12-10
Update CHANGELOG for 1.10.0 RC1

anuraaga:
I have intentionally not updated docs / japicmp since it's an interim release. I'm wondering if we should also skip the CHANGELOG and reflect the final state in 1.10.0 without the RC steps in it, open to either approach. We do need the notes in the GitHub releases at least though to provide documentation on the deprecations / removals

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,39 @@
 
 ## Version 1.10.0 (unreleased):
 
+## Version 1.10.0 RC1 2021-12-10:
+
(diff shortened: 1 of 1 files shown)


jkwatson:
I think it makes sense to include this in the changelog. We can always roll up all the rc changes to the main changelog entry for 1.10 when it's made official.

jkwatson (approved):
Thanks!!!

PR #3983, opened 2021-12-11
Add http/protobuf retry

jack-berg:
- Add experimental retry support for http/protobuf exporters
- Create new `io.opentelemetry.exporter.otlp.internal.retry` package relocate retry related classes
- Rename addRetryPolicy to setRetryPolicy for consistency with other setters
- Wire up retry to autoconfigure for http/protobuf

It wasn't entirely clear which http status codes were retryable, so I opened an issue with the [spec](https://github.com/open-telemetry/opentelemetry-specification/issues/2217). Assumed the following are retryable: 429, 502, 503, 504.


exporters/otlp-http/logs/build.gradle.kts
```diff
@@ -10,8 +10,7 @@ otelJava.moduleName.set("io.opentelemetry.exporter.otlp.http.logs")
 
 dependencies {
   api(project(":sdk:logs"))
-
-  implementation(project(":exporters:otlp:common"))
+  api(project(":exporters:otlp:common"))
```
anuraaga, line 13: I think sticking with implementation is better, even if it means duplicating into some build files

sdk-extensions/autoconfigure/src/testOtlpHttp/java/io/opentelemetry/sdk/autoconfigure/OtlpHttpServerExtension.java
```diff
@@ -0,0 +1,117 @@
...
+
+  static {
+    try {
+      canonicalHostName = InetAddress.getByName("localhost").getCanonicalHostName();
+    } catch (UnknownHostException e) {
+      throw new IllegalStateException("Error resolving canonical host name.", e);
+    }
+  }
+
+  private final HeldCertificate heldCertificate;
```
anuraaga, line 43: While we're extracting this, you could consider switching back to `SelfSignedCertificate` from armeria as the incompatibility with okhttp was fixed

exporters/otlp-http/logs/build.gradle.kts
jack-berg: We talked about this back [here](https://github.com/open-telemetry/opentelemetry-java/pull/3791#discussion_r737575914), but I've tested it and confirmed that we can get away with implementation. 

Best to keep it as implementation until we actually promote something in common out of internal. Will update the grpc modules as well.

anuraaga (approved):

PR #3984, opened 2021-12-12
Set OT Shim's SpanBuilder attributes before the Span is created.

### LINKED ISSUE #3962: SpanBuilderShim: set span attribute before starting a span
In OpenTracing, we can elevate a trace's importance by adding a `sampling.priority` tag to a span. This doesn't seem to be supported in the opentracing-shim library. `SpanBuilderShim` starts a Span before its attributes are copied. This would cause the sampling decision to be made without considering the value of `sampling.priority`. 

Is it possible to copy the `attributes/tags` before starting a Span so we can have our own Sampler implementation which can respect the value of `sampling.priority`?

```
...

for (int i = 0; i < this.spanBuilderAttributeKeys.size(); i++) {
  AttributeKey key = this.spanBuilderAttributeKeys.get(i);
  Object value = this.spanBuilderAttributeValues.get(i);
  builder.setAttribute(key, value);
}

io.opentelemetry.api.trace.Span span = builder.startSpan();

if (error) {
  span.setStatus(StatusCode.ERROR);
}
...
```

carlosalberto:
Fixes #3962

### DIFF
--- a/opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanBuilderShim.java
+++ b/opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanBuilderShim.java
@@ -215,7 +215,7 @@ final class SpanBuilderShim extends BaseShimObject implements SpanBuilder {
     }
 
     // Attributes passed to the OT SpanBuilder MUST
-    // me set before the OTel Span is created,
+    // be set before the OTel Span is created,
     // so those attributes are available to the Sampling API.
     for (int i = 0; i < this.spanBuilderAttributeKeys.size(); i++) {
       AttributeKey key = this.spanBuilderAttributeKeys.get(i);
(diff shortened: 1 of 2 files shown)


opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/SpanBuilderShim.java
```diff
@@ -214,13 +214,16 @@ public Span start() {
       builder.setStartTimestamp(startTimestampMicros, TimeUnit.MICROSECONDS);
     }
 
-    io.opentelemetry.api.trace.Span span = builder.startSpan();
-
+    // Attributes passed to the OT SpanBuilder MUST
+    // me set before the OTel Span is created,
```
shawn-canva, line 218:
```suggestion
    // be set before the OTel Span is created,
```

shawnsong, line 218: Typo?
```suggestion
    // be set before the OTel Span is created,
```

jkwatson (approved):

anuraaga (approved):

Issue #3952, opened 2021-12-03
ScopeManagerShim.activate(span) throws IllegalArgumentException when null is passed in


shawnsong:
`ScopeManagerShim.activate()` throws `IllegalArgumentException` if a `null` span is passed in. I'm not sure whether this is a bug or implemented on purpose but it is inconvenient to use as we need to do a `null` check before calling `activate`.

Currently, I need to do this to avoid the exception.
```
try (Scope scope = (span != null ? tracer.scopeManager().activate(span) : null)) {
  ...
}
```
It would be good if `null` is returned by the method as try-with-resources doesn't complain a `null` resource. 

**What version and what artifacts are you using?**
Artifacts: (`opentelemetry-opentracing-shim`)
Version: (`v1,9,1-alpha`)

carlosalberto:
Fixed via #3985

anpatellll:
@carlosalberto Whats the ETA of this being released?

PR #3985, opened 2021-12-12
Stop throwing an exception when ScopeManagerShim activate a null Span

shawnsong:
Details are described in: https://github.com/open-telemetry/opentelemetry-java/issues/3952

A scope is returned to make sure the context is restored.

### DIFF
--- a/opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/ScopeManagerShim.java
+++ b/opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/ScopeManagerShim.java
@@ -44,7 +44,12 @@ final class ScopeManagerShim extends BaseShimObject implements ScopeManager {
 
   @Override
   @SuppressWarnings("MustBeClosedChecker")
-  public Scope activate(Span span) {
+  public Scope activate(@Nullable Span span) {
+    if (span == null) {
+      SpanShim spanShim = new SpanShim(telemetryInfo(),
+          io.opentelemetry.api.trace.Span.getInvalid());
+      return new ScopeShim(Context.current().with(spanShim).makeCurrent());
+    }
     if (!(span instanceof SpanShim)) {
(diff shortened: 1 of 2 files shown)


shawnsong:
Hi @carlosalberto, do you mind taking a quick look at this PR? This is my first attempt :) Thanks.

opentracing-shim/src/test/java/io/opentelemetry/opentracingshim/TracerShimTest.java
```diff
@@ -80,6 +80,25 @@ void activateSpan() {
     assertThat(io.opentelemetry.api.baggage.Baggage.current()).isSameAs(EMPTY_BAGGAGE);
   }
 
+  @Test
+  void activateNullSpan() {
```
carlosalberto, line 84: What about starting the test with an actual active `Span`, _and then_ you activate/test the null value?

For testing purposes it's the same, but I get the feeling that changing it that way could make the test more readable.

opentracing-shim/src/main/java/io/opentelemetry/opentracingshim/ScopeManagerShim.java
```diff
@@ -44,7 +44,12 @@ public Span activeSpan() {
 
   @Override
   @SuppressWarnings("MustBeClosedChecker")
-  public Scope activate(Span span) {
+  public Scope activate(@Nullable Span span) {
+    if (span == null) {
+      SpanShim spanShim =
```
carlosalberto, line 49: I think we could cache a final private `SpanShim` instance that simply wraps OTel `Span.getInvalid()`, and we can re-use that for all invocations:

```java
if (span == null) {
  return new ScopeShim(Context.current().with(NOP_SPANSHIM).makeCurrent());
}
```

jkwatson (approved):

anuraaga (approved):

anuraaga:
Thanks!

Issue #3848, opened 2021-11-09
Optimisation and benchmarking of backing data structure for exponential histogram


jamesmoessis:
For the exponential histogram, we decided to start with the simplest possible implementation for the backing data structure, and then beat it from there. Some context here: https://github.com/open-telemetry/opentelemetry-java/pull/3724#discussion_r726722725. This is the ticket to beat it. 

This ticket is to track the work for the optimisation, testing, and benchmarking of these data structures. There are notable reference implementations such as NrSketch and DynaHist that the author may draw from. 

To implement the backing data structure, the author should implement `ExponentialCounter`.

jamesmoessis:
benchmarking PR: https://github.com/open-telemetry/opentelemetry-java/pull/3986

jsuereth:
A few things I found when benchmarking and investigating the initial prototype:

- `MapCounter` uses both `ConcurrentMap` and `AtomicLong`.   However, the current `Handle` puts both of these behind a strict lock/synchronize (as is done for explicit bucket histogram).  You can probably use a more efficient IntMap implementation with no-concurrency in the current impl.
- `MapCounter` uses a LOT of memory as you take samples.   The original NrSketch algorithm that's outlined as the baseline for Exponential histogram makes a good balance of "usually byte is enough to count measurements" because you're spread over a lot of buckets.   I've been running some tests myself on memory usage and performance and the current `MapCounter` seems to really miss the expected use case.   I'll need some time to get these into PR form, but I recommend running some here.
- Copying of data is ONLY done on the collection path or when re-scaling.   
  - The first, we can optimise later, but note it's usually done at a 1-minute interval, so the performance impact is a lot lower and usually out-of-band.   
  - For the second (re-scaling), we should probably look into one of two things:  (1) preserve the last scale value per-stream or (2) Verify whether our assumptions on bucket-size + scale match common use cases (e.g. do we "fast fall" into a  scale that accounts for most data, or do we see uniform scale-increase throughout a collection cycle).

jsuereth:
Here's a quick comparison of using a Circular-Buffer (with preallocated array of integers) vs. MapCounter:  https://github.com/open-telemetry/opentelemetry-java/compare/main...jsuereth:wip-exponential-counter-perf

[...]

Issue #3766, opened 2021-10-18
Allow mutable accumulations


jamesmoessis:
**Is your feature request related to a problem? Please describe.**
As discussed with @jsuereth in #3724, it would be beneficial to allow accumulations to be mutable. Specifically for the exponential histogram, immutability of accumulations requires a decent amount of CPU + memory. 

**Describe the solution you'd like**
`DeltaMetricStorage` relies on accumulations being immutable, however switching it to a clone + mutate approach would allow accumulations to be mutated while merging. i.e. merge into existing accumulation rather than creating a new one. 

Additionally, we should be careful about the lifecycle and ownership of accumulations. Anything to aid this is encouraged.

jack-berg:
Accumulations have been optimized out of the SDK implementation in #5154. This was in addition to several other recent metrics SDK optimizations, some of which were specific to exponential histograms.

Can further optimize allocations by going forward with something like the #5178 prototype, but that's a distinctly different problem than what is requested here.

PR #3986, opened 2021-12-12
Rework histogram benchmarks to include exponential + no-bucket.

jsuereth:
- Create unified histogram aggregation benchmark
- Add exponential histograms to benchmark
- Update benchmark to use same-value + random values

Current results:

- Exponential histogram is ~2x slower than ExplicitBucket (60 ns/op vs. 30 ns/op), but I think this can be recovered.
- number of  threads appears to impact Exponential histogram more than ExplicitBucket, need to investigate if this is an artifact or real.
- 1-bucket Explicit histogram vs. N-bucket Explicit histogram shows no significant performance difference (which makes sense, same # of locks / memory barriers, just simpler index lookup).


sdk/metrics/src/jmh/java/io/opentelemetry/sdk/metrics/internal/aggregator/HistogramBenchmark.java
```diff
@@ -0,0 +1,63 @@
...
+    public final void setup() {
+      aggregatorHandle = aggregation.getAggregator().createHandle();
+    }
+
+    public void record() {
+      this.aggregatorHandle.recordDouble(valueGen.generateValue());
+    }
+  }
+
+  @Benchmark
```
jamesmoessis, line 46: > number of threads appears to impact Exponential histogram more than ExplicitBucket, need to investigate if this is an artifact or real. 

I think this makes sense, since the exponential histogram has to spend more time in write lock, especially during rescaling. I am however kind of surprised that there isn't more variation/outliers across different recordings in the exponential histogram. 

I'd expect recordings that require rescaling to be much slower than recordings that fit into the existing range (I'm assuming the black horizontal lines indicate min/max time, correct me if I'm wrong). It appears that rescaling would be performed during warmup iterations which aren't measured. 

If we don't want to measure that outlier case of rescaling then that's reasonable, but it may also be interesting to see by how much a rescale slows down a recording.

sdk/metrics/src/jmh/java/io/opentelemetry/sdk/metrics/internal/aggregator/HistogramAggregationParam.java
```diff
@@ -0,0 +1,34 @@
...
+  EXPLICIT_DEFAULT_BUCKET(
+      new DoubleHistogramAggregator(
+          ExplicitBucketHistogramUtils.createBoundaryArray(
+              ExplicitBucketHistogramUtils.DEFAULT_HISTOGRAM_BUCKET_BOUNDARIES),
+          ExemplarReservoir::noSamples)),
+  EXPLICIT_SINGLE_BUCKET(
+      new DoubleHistogramAggregator(
+          ExplicitBucketHistogramUtils.createBoundaryArray(Collections.emptyList()),
+          ExemplarReservoir::noSamples)),
+  EXPONENTIAL(new DoubleExponentialHistogramAggregator(ExemplarReservoir::noSamples));
```
jamesmoessis, line 23: Eventually I expect there will be some configuration parameter for max buckets / expected range (like in the recent go implementation), which will to some extent affect the benchmark. Once that exists, it will be interesting to see how those parameters affect performance.

jamesmoessis (approved):
Nice, all looks good to me. 

In terms of what we can do to improve performance, there are a couple of tickets surrounding this: 
* Optimisation of the backing data structure. (`ExponentialCounter` implementation comparison) https://github.com/open-telemetry/opentelemetry-java/issues/3848 
* Allow the accumulation to be mutable so less copying is needed: https://github.com/open-telemetry/opentelemetry-java/issues/3766 (though it doesn't look like this benchmark covers collection) 

Also if you have other ideas around optimising it I would be interested to hear.

sdk/metrics/src/jmh/java/io/opentelemetry/sdk/metrics/internal/aggregator/HistogramBenchmark.java
jsuereth: I think we need to measure both, and benchmarking here is hard.

1. I actually *want* us to try to optimise for stable scale per-metric-stream.   Ideally we find a way for the scale to stabilize for a particular metric so re-scaling *should* be rare / mostly seen at start-up.
2. We should understand the cost of rescaling and be able to identify whether we're doing it too often.  I'll look into adding a benchmark specific to that.   Are there other scenarios you can call out that should be measured?

sdk/metrics/src/jmh/java/io/opentelemetry/sdk/metrics/internal/aggregator/HistogramAggregationParam.java
jsuereth: Yep!  I'd love to have those as well.

jsuereth:
@jamesmoessis Made some comments on the benchmark/optimisation PR.  I fully suspect the NrSketch algorithm will need to be pulled back in over time, but we can walk the benchmark path to prove it.

sdk/metrics/src/jmh/java/io/opentelemetry/sdk/metrics/internal/aggregator/HistogramBenchmark.java
jamesmoessis: Great that sounds good. There aren't other scenarios I can think of currently

sdk/metrics/src/jmh/java/io/opentelemetry/sdk/metrics/internal/aggregator/HistogramValueGenerator.java
```diff
@@ -0,0 +1,26 @@
...
+
+package io.opentelemetry.sdk.metrics.internal.aggregator;
+
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.function.DoubleSupplier;
+
+/** Methods of generating values for histogram benchmarks. */
+@SuppressWarnings("ImmutableEnumChecker")
+public enum HistogramValueGenerator {
+  SAME_VALUE(() -> 100.0056),
```
anuraaga, line 14: Curious if this value has any meaning, especially in terms of what the bucket is. Is it ok not to test various values?

sdk/metrics/src/jmh/java/io/opentelemetry/sdk/metrics/internal/aggregator/HistogramValueGenerator.java
```diff
@@ -0,0 +1,26 @@
...
+  RANDOM_WITHIN_2K(() -> ThreadLocalRandom.current().nextDouble(2000));
```
anuraaga, line 15: Random JMH tends to be not that useful, especially here it seems like numbers could be vastly different across runs. How about various golden lists of measurements that are all recorded in a single iteration rather than one random value?

anuraaga:
> I fully suspect the NrSketch algorithm will need to be pulled back in over time, but we can walk the benchmark path to prove it.

I agree and its good we have benchmarks now for it. I'd recommend going piece by piece, for example the MultiTypedArray may or may not be worth it in practice, even if other parts of the algorithm are, want to verify that. Also want to compare each part with anything in ddsketch that could be used instead

https://github.com/DataDog/sketches-java/tree/master/src/main/java/com/datadoghq/sketch/ddsketch/mapping

sdk/metrics/src/jmh/java/io/opentelemetry/sdk/metrics/internal/aggregator/HistogramValueGenerator.java
jsuereth: This was just to understand the cost of distributing over buckets vs. always placing in the same.

It's only useful if you're looking for where the slowdown is, not really as its own test.  I can remove, as this is NOT a use case to optimsie for.  The random value one is.

sdk/metrics/src/jmh/java/io/opentelemetry/sdk/metrics/internal/aggregator/HistogramValueGenerator.java
jsuereth: Two things:

1. The resizing/reallocating aspect of memory is important here.  `MapCounter` performs abysmally in other benchmarks I've targeted, but I can duplicate that in a no-warmup benchmark to show the cost over time.
2. While numbers CAN be vastly different over runs, I've seen pretty stable / repeated performance output here.  This is because the algorithms *should* be resilient to randomness in practice.

I can update this to randomnly generate the values and use the same ones each test, but I'll need to also go update to make sure warmup isn't preserving allocated memory in ways that won't reflect production.

jsuereth:
> I agree and its good we have benchmarks now for it. I'd recommend going piece by piece, for example the MultiTypedArray may or may not be worth it in practice, even if other parts of the algorithm are, want to verify that. 


I'm seeing ~ 4x slowdown of MapCounter vs. the circular-buffer approach, and a 2x slowdown of MapCounter vs. MultiTypedArray.  For performance tests, we need to track memory usage for MultiTypedArray.  It's a 2x slowdown over a simple ring-buffer of integers, but it's a significant memory reduction of metric streams.   Given expected use case scenarios (e.g. tracking latency with existing instrumentation + attributes), MapCounter is very obviously a problem.  

> Also want to compare each part with anything in ddsketch that could be used instead

> https://github.com/DataDog/sketches-java/tree/master/src/main/java/com/datadoghq/sketch/ddsketch/mapping

These are just the mapping functions.  We can use these to optimise the "which bucket to write to", but that's not "how to record buckets".    Datadog sketch storage stores doubles, not integers.  They support both Spare (MapCounter) + Dense (circular-buffer-like) storage.  They can also record partial counts and rebalance their storage based on that.   TL;DR; their algorithm for collapsing is different than our ExponentialHistograms, and we need to optimise this aspect of storage differently.

1. It's likely our bucket counts for delta-recordings remain *below* the byte/short threshold.
2. We expect to have many streams per-histogram.  E.g. for latency, at a minimum, we expect ONE of these things to be instantiated per-route.  Cost of memory allocation + stabilization is critical as this is in the hot-path, and highly allocated.
3. We expect to have many different counts in buckets vs. everything landing in the same (similar) buckets (especially starting at a scale of 20).

I'll work on providing benchmarks to back this, but I'd recommend reading relevant sketch algorithms/papers for performance tradeoffs to understand how we landed on the new-relic algorithm as our preferred approach in Otel.   DataDog sketch is nice, but we are aligned closer to Prometheus pending high resolution histograms.

sdk/metrics/src/jmh/java/io/opentelemetry/sdk/metrics/internal/aggregator/HistogramValueGenerator.java
```diff
@@ -0,0 +1,88 @@
...
+    // Add Double max value as our other extreme.
+    fixedBoundaries.add(Double.MAX_VALUE);
+    return ExplicitBucketHistogramUtils.createBoundaryArray(fixedBoundaries);
+  }
+
+  /** Create a pool of random numbers within bound, and of size. */
+  private static double[] randomPool(int size, double bound) {
+    double[] pool = new double[size];
+    for (int i = 0; i < size; i++) {
+      pool[i] = ThreadLocalRandom.current().nextDouble(bound);
```
anuraaga, line 75: Let's use a normal `Random` with a fixed seed for these two

jsuereth: done

anuraaga (approved):

PR #3989, opened 2021-12-14
Deprecate InMemoryMetricReader

### LINKED ISSUE #3943: InMemoryMetricReader vs InMemoryMetricExporter
The [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/in-memory.md) defines the `InMemoryMetricExporter`, but does [not mention](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#metricreader-operations) anything like the `InMemoryMetricReader`.

Practically speaking, `InMemoryMetricReader` is a bit more useful for testing, since it allows you to collect all metrics ad hoc. However, you can accomplish the same thing with `InMemoryMetricExporter` (albeit with worse ergonomics) by calling flush on the: `SdkMeterProvider`, `PeriodicMetricReader` the `InMemoryMetricExporter` is associated with, or on `InMemoryMetricExporter` itself. 

`InMemoryMetricReader` is really closer to a [pull metric exporter](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#pull-metric-exporter) like prometheus. Without a mention in the spec, should we remove `InMemoryMetricReader`?

On a related note, its strange and potentially incorrect that `InMemoryMetricExporter#flush()` forces the collection of metrics. The [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#forceflush-1) says that flush should hint that the export of any metrics the exporter has received prior to the call should be completed as soon as possible, but in our case flush initiates a metric collect.

jack-berg:
Resolves #3943.


sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/CardinalityTest.java
```diff
@@ -221,7 +237,9 @@ void cardinalityLimits_asynchronousInstrument() {
     meter.counterBuilder("async-counter1").buildWithCallback(callback);
     meter.counterBuilder("async-counter2").buildWithCallback(callback);
 
-    assertThat(deltaReader.collectAllMetrics())
+    meterProvider.forceFlush().join(10, TimeUnit.SECONDS);
```
anuraaga, line 240: I agree with the sentiment but find this to be a significantly worse UX when writing metrics tests. Can you confirm that users won't have to worry about flushing after we integrate metrics into our JUnit extension? If so then it seems fine to me

jack-berg: Its going to depend on the type of test they're writing, but generally, yes integration of metrics into our JUnit extension will help. We could extend `OpenTelemetryExtension` with something like the snippet below, to flush metrics before returning them for callers to assert against. 
```
  public List<MetricData> getMetrics() {
    return getMetrics(true);
  }

  public List<MetricData> getMetrics(boolean forceFlush) {
    if (forceFlush) {
      openTelemetry.getSdkMeterProvider().forceFlush().join(10, TimeUnit.SECONDS);
    }
    return metricExporter.getFinishedMetricItems();
  }
```

If they're writing tests that depend on being able to configure non standard things like the clock, or multiple readers, then they won't be able to leverage `OpenTelemetryExtension` (as it stands today) since it doesn't provide any config options.

Question though: the sdk spec defines requirements for our sdk, but we include testing utilities like `InMemorySpanExproter` and `InMemoryMetricExporter` in the :sdk:testing` module, not the sdk. And we include a fair number of other convenience utilities which are not specified. By the same logic, I would think we're entitled to include `InMemoryMetricReader` as a testing utility in `:sdk:testing`. It certainly improve testing patterns so I would be in favor of this if allowed by the spec.

anuraaga: Yeah thinking more, I don't think testing utils are an aspect of the spec. We need to provide testing utilities for helping with writing tests - including tests that don't have background threads polling for metrics :) So keeping InMemoryMetricReader in sdk-testing sounds better than removing it

jsuereth: FYI - regarding #3943 if the way the spec is worded leads to us removing this class then we need to go reword the specification.

The Speicfication was written to allow a design decision of EITHER "Exporter" does everything *or* "Reader is Pull", "Exporter is Push".   

.NET AFAIK is the only impl of the first, most of the rest of us do the second.   I'm against this change because I think we should go clarify the specification, not remove the best utiltiy we have for testing the metrics SDK.

jkwatson: How about just moving it to the test artifact, rather than deleting it?

jack-berg: > How about just moving it to the test artifact, rather than deleting it?

Yeah sounds like we're all in favor of that.

jack-berg:
I've reverted the original change. Instead, I've deprecated InMemoryMetricReader and moved a copy to `:sdk:metrics-testing`.

sdk/metrics-testing/src/main/java/io/opentelemetry/sdk/testing/exporter/InMemoryMetricReader.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  /** Returns all metrics accumulated since the last call. */
+  public Collection<MetricData> collectAllMetrics() {
+    MetricProducer metricProducer = this.metricProducer;
+    if (metricProducer != null) {
+      return metricProducer.collectAllMetrics();
+    }
+    return Collections.emptyList();
+  }
```
anuraaga, line 69: Not a big deal but we may as well add the test case for this case (collect before apply)

anuraaga (approved):

sdk/metrics-testing/src/main/java/io/opentelemetry/sdk/testing/exporter/InMemoryMetricReader.java
jack-berg: Done. 

While I was at it I noticed that there were two tests classes for this that were largely duplicative, so I merged them. Also noticed that the `shutdown()` method didn't actually do anything to prevent future calls to `collectAllMetrics()`, so I adjusted its behavior to set `metricProducer = null`, returning it to its state pre `apply(MetricProducer)`.

PR #3990, opened 2021-12-14
Remove deprecated metrics APIs

anuraaga:


docs/apidiffs/current_vs_latest/opentelemetry-api.txt
```diff
@@ -1,188 +1,2 @@
 Comparing source compatibility of  against 
-+++  NEW CLASS: PUBLIC(+) STATIC(+) io.opentelemetry.api.metrics.DefaultMeter$NoopDoubleHistogram  (not serializable)
```
anuraaga, line 2: Ah while I didn't think we should add a `1.9 vs 1.10-RC1` japi during the release, it looks like I should have at least refreshed current_vs_latest during the release, this is the diff vs RC1 (no diff)

bogdandrutu (approved):

jack-berg (approved):

jkwatson (approved):

docs/apidiffs/current_vs_latest/opentelemetry-api.txt
jkwatson: I wonder if we should also just generate the api diffs between 0.9.0 and the RC, for completeness. We can end up also having 0.9.0->0.10.0 when we release for real.

PR #3991, opened 2021-12-14
Use OkHttp for AWS resource detectors

anuraaga:
With OTLP exporters having a default dependency on OkHttp now, the vast majority of users will have OkHttp available on the classpath. There isn't a great reason to stick to HttpUrlConnection anymore - notably this will prevent resource detection requests from being traced during agent startup due to https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4642


sdk-extensions/aws/src/test/java/io/opentelemetry/sdk/extension/aws/resource/SimpleHttpClientTest.java
```diff
@@ -49,28 +49,11 @@ void testFetchString() {
...
     String urlStr = String.format("http://localhost:%s%s", server.httpPort(), "/path");
-    JdkHttpClient jdkHttpClient = new JdkHttpClient();
-    String result = jdkHttpClient.fetchString("GET", urlStr, requestPropertyMap, null);
+    SimpleHttpClient httpClient = new SimpleHttpClient();
+    String result = httpClient.fetchString("GET", urlStr, requestPropertyMap, null);
     assertThat(result).isEmpty();
   }
 
-  @Test
-  void requestBody() {
```
anuraaga, line 58: I deleted the code path taking a request body as that was only used by the moved-out xray sampler code

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource/SimpleHttpClient.java
```diff
@@ -0,0 +1,145 @@
...
+import javax.net.ssl.TrustManagerFactory;
+import javax.net.ssl.X509TrustManager;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.RequestBody;
+import okhttp3.Response;
+import okhttp3.ResponseBody;
+
+/** A simple HTTP client based on OkHttp. Not meant for high throughput. */
+final class SimpleHttpClient {
```
kubawach, line 31: Total NIT: as it's bound with OkHttp for good and bad I'd name it: "SimpleOkHttpClient" to note that.

kubawach (approved):
LGTM :)

jkwatson (approved):

sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource/SimpleHttpClient.java
anuraaga: Might end up swapping OkHttp back to HTTPUrlConnection or something else later, next time won't have to change class name references everywhere by using a more generic name here :)

PR #3994, opened 2021-12-15
Add support for log emitter provider

elevenzhan:
In this CL:
* add default noop log emitter, guaranteed to be consistent with tracer and metrics
* add logEmitterProvider api
* add unit test cases


anuraaga (commented):
@jack-berg Can you check out this PR?

anuraaga:
@trask Do I understand correctly that your PR is the same as this API?

https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4917

trask:
> @trask Do I understand correctly that your PR is the same as this API?
> 
> [open-telemetry/opentelemetry-java-instrumentation#4917](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4917)

I don't think so. open-telemetry/opentelemetry-java-instrumentation#4917 introduces a way for instrumentation to avoid direct dependency on SDK, which we need for javaagent bridging.

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/GlobalLogEmitterProvider.java
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.logs;
+
+/** This class is a temporary solution until log SDK is marked stable. */
```
jack-berg, line 8: Two thoughts:

1. Do we want a global accessor for the log sdk? There's not a precedent for providing global accessor for SDK components. `GlobalMeterProvider` provides access to `MeterProvider`, not `SdkMeterProvider`. `GlobalOpenTelemetry` provides access to `OpenTelemetry`, not `OpenTelemetrySdk`. Who would use this?

2.  Related to the javadoc: because there are no plans for a log api, this likely isn't a temporary solution.

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/DefaultLogEmitter.java
```diff
@@ -0,0 +1,80 @@
...
+    return INSTANCE;
+  }
+
+  @Override
+  public LogBuilder logBuilder() {
+    return new NoopSpanBuilder();
+  }
+
+  // Noop implementation of LogEmitter.Builder.
+  private static final class NoopSpanBuilder implements LogBuilder {
```
jack-berg, line 35: `NoopSpanBuilder` => `NoopLogBuilder`

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/DefaultLogEmitter.java
```diff
@@ -0,0 +1,80 @@
...
+
+  private static final DefaultLogEmitter INSTANCE = new DefaultLogEmitter();
+
+  static LogEmitter getInstance() {
+    return INSTANCE;
+  }
+
+  @Override
+  public LogBuilder logBuilder() {
+    return new NoopSpanBuilder();
```
jack-berg, line 31: Can use a singleton here instead of instantiating a new instance each time.

anuraaga:
Hi @elevenzhan - we are currently working on a logs appender API in the instrumentation API here

https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4917

This is because we only expect it to be used by log appenders, not by end users. Will that satisfy your use case?

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/GlobalLogEmitterProvider.java
elevenzhan: Ok, I see and It has been deleted; this can be defined in my business code in a private way

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/DefaultLogEmitter.java
elevenzhan: done

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/DefaultLogEmitter.java
elevenzhan: done

elevenzhan:
> Hi @elevenzhan - we are currently working on a logs appender API in the instrumentation API here
> 
> [open-telemetry/opentelemetry-java-instrumentation#4917](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4917)
> 
> This is because we only expect it to be used by log appenders, not by end users. Will that satisfy your use case?

Hi @anuraaga - I simplified the implementation and removed global provider; javaagent is a greatest way to resolve logs export, but in some unusable scenarios, It cannot overwrite these scenes; 
this CR provides a LogEmitterProvider api and noop implementation, that can be used anywhere who want to use log exporter, not just in javaagent

anuraaga:
> that can be used anywhere who want to use log exporter, not just in javaagent

The logs appender API in the java-instrumentation repo is just a library, it will be used for instrumentation both within the javaagent and without. So I think this one is mostly a dupe of that API

elevenzhan:
> > that can be used anywhere who want to use log exporter, not just in javaagent
> 
> The logs appender API in the java-instrumentation repo is just a library, it will be used for instrumentation both within the javaagent and without. So I think this one is mostly a dupe of that API

thanks @anuraaga, I see instrumentation-api-appender is consistent with current MR, so I will close this merge

PR #3995, opened 2021-12-15
OTLP Marshalers for Exponential Histogram

jamesmoessis:
## What has been added

This PR contains the marshalers and additional utilities required to export the exponential histogram: 

* Update OTLP 0.9.0 -> 0.11.0 (exponential histogram was added in 0.11.0) 
* Marshalers for the `MetricData`, `PointData`, and `ExponentialHistogramBuckets` all corresponding to their protobuf message type.
* Added size calculations and serializers for `sint32`, `uint64` and `repeated varint` which were required to write the buckets.
* Implemented the additional `Serializer` methods in `ProtoSerializer` and `JsonSerializer`.
* Tests, one for `ExponentialHistogram` and one for `ExponentialHistogramPointData` serialization. 

Linked issue: https://github.com/open-telemetry/opentelemetry-java/issues/3550


exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/ProtoSerializer.java
```diff
@@ -140,6 +151,18 @@ protected void writeEndRepeatedPrimitive() {
     // Do nothing
   }
 
+  @Override
+  protected void writeStartRepeatedVarint(ProtoFieldInfo field, int payloadSize)
```
jamesmoessis, line 155: The `ProtoSerializer` implementation needs the payload size to write repeated varints ([docs on that](https://developers.google.com/protocol-buffers/docs/encoding#packed)). It means the the size gets calculated twice for the payload, in two different places. Once when calculating overall size in the marshaler, and again when serializing. I don't think it's a huge deal but thought I'd call it out.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/Serializer.java
```diff
@@ -200,6 +217,24 @@ public void serializeRepeatedFixed64(ProtoFieldInfo field, long[] values) throws
...
   }
 
+  /** Serializes a {@code repeated uint64} field. */
+  public void serializeRepeatedUInt64(ProtoFieldInfo field, List<Long> values) throws IOException {
+    if (values.isEmpty()) {
+      return;
+    }
+
+    int payloadSize = 0;
+    for (long v : values) {
```
anuraaga, line 227: We know it's a `PrimitiveLongList` in practice so let's use `PrimitiveLongList.toArray` first. I guess I would make this method accept `long[]` and the caller do that conversion.

Same for sizing

anuraaga (commented):
Thanks!

jkwatson:
Does the collector support protocol v1.11? If not, we shouldn't release this.

jamesmoessis:
@jkwatson The collector looks like it supports 0.11.0 since October: https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.38.0

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/Serializer.java
jamesmoessis: Ah, nice I wasn't aware you changed the `MetricData` to use the `PrimitiveLongList`. I've changed the bucket marshaller to use `PrimitiveLongList.toArray(counts)` when calling the serializer 👍

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MarshalerUtil.java
```diff
@@ -107,6 +107,26 @@ private static int sizeRepeatedFixed64(ProtoFieldInfo field, int numValues) {
     return size;
   }
 
+  /**
+   * Returns the size of a repeated uint64 field.
+   *
+   * <p>Packed repeated fields contain the tag, an integer representing the incoming payload size,
+   * and an actual payload of repeated varints.
+   */
+  public static int sizeRepeatedUInt64(ProtoFieldInfo field, List<Long> values) {
```
anuraaga, line 116: Think we need to update this one too for the PrimitiveLongList pattern

anuraaga (approved):

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/MarshalerUtil.java
jamesmoessis: Yeah, I wasn't sure about that since I was keeping it consistent with `sizeRepeatedFixed64` which takes in a `Long<List>`. Anyway, I've just pushed the change to make this one receive a `long[]`.

jsuereth (approved):

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/ProtoSerializer.java
jamesmoessis: Sorry for the late reply, I've been off for the holiday period! 

Basically with the current flow, it might be possible but from what I can see it would be awkward and would have to somewhat deviate from the current pattern. It's mainly because the `calculateSize` is `private static` and the result is passed into `super`: https://github.com/open-telemetry/opentelemetry-java/pull/3995/files#diff-23872c8146e426d9040de8d826fd7fa28f7473da4384de0ab5c6b002701186d9R26

So to me there isn't somewhere where I can stop and set an instance field, since the size is passed to `super` which has to be done first. @anuraaga might have an idea of a clean way to memoize it?

anuraaga: Let's stick with this for now, thanks for looking into it

PR #4002, opened 2021-12-16
Make AutoConfiguredOpenTelemetrySdkBuilder.setConfig public

### LINKED ISSUE #4001: AutoConfiguredOpenTelemetrySdkBuilder#setConfig is not publicly visible
**Describe the bug**
Deprecated method `io.opentelemetry.sdk.autoconfigure.OpenTelemetrySdkAutoConfiguration#initialize` point to wrong code alternative (`AutoConfiguredSdk` should be `AutoConfiguredOpenTelemetrySdk`) and part of the code replacement doesn't work because `io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdkBuilder#setConfig` is not visible (default visibility).

Artifacts: io.opentelemetry.sdk.autoconfigure
Version: 1.9.1

radcortez:
- Fixes #4001

### DIFF
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
@@ -63,7 +63,7 @@ public final class AutoConfiguredOpenTelemetrySdkBuilder implements AutoConfigur
    * Sets the {@link ConfigProperties} to use when resolving properties for auto-configuration.
    * {@link #addPropertiesSupplier(Supplier)} will have no effect if this method is used.
    */
-  AutoConfiguredOpenTelemetrySdkBuilder setConfig(ConfigProperties config) {
+  public AutoConfiguredOpenTelemetrySdkBuilder setConfig(ConfigProperties config) {
     requireNonNull(config, "config");
     this.config = config;
     return this;


anuraaga:
@radcortez Sorry that the deprecation notice was confusing - the intent is `addPropertiesSupplier` is used by users so we don't need `setConfig` to be public

radcortez:
Yes, I had a look into `addPropertiesSupplier`, but it doesn't offer the same functionality as `OpenTelemetrySdkAutoConfiguration#initialize(boolean, ConfigProperties)`. This allows you to plug in your own configuration mechanism, while the replacement method only allows you to set defaults.

To be more specific on the use case, for Quarkus, our intent is to transparently support OTel configuration using Quarkus configuration mechanism, so users can use different configuration sources to configure their environment. The ability to provide an implementation of `ConfigProperties` allows us to do that. Of course, we also support system properties and environment variables, so the behavior is exactly the same.

anuraaga:
So to confirm you want to replace config with quarkus config and not support OTel environment variables? If so, it, at least so far, has not been a use case for autoconfigure. autoconfigure exists to implement the SDK environment variables, and provide some customization of defaults on top. This isn't set in stone but I think we lean towards wanting to have the OTel env var experience consistent among OTel users.

If not, then I feel like it's still possible to have a props supplier that uses the quarkus config. OTel env vars are also supported, quarkus config is also supported.

radcortez:
> So to confirm you want to replace config with quarkus config and not support OTel environment variables?

Yes and No :) The Quarkus config supports the exact same notation as env variables like OTel, so they are still supported like the OTel implementation, but just provided by Quarkus.

> If so, it, at least so far, has not been a use case for autoconfigure. autoconfigure exists to implement the SDK environment variables, and provide some customization of defaults on top. This isn't set in stone but I think we lean towards wanting to have the OTel env var experience consistent among OTel users.

I agree. Our goal is to offer the exact same experience as OTel, but at the same time offer the Quarkus experience.

> If not, then I feel like it's still possible to have a props supplier that uses the quarkus config. OTel env vars are also supported, quarkus config is also supported. Just OTel env variables take precedence, which I think aligns with OTel

In Quarkus, we have other config sources (properties files, custom user sources, etc). They have a defined order, so my preference would be to keep the order we look for stuff. We use the exact same order as OTel, regarding environment and system properties. Is it just that we may have things in between (or with higher precedence).

jack-berg:
The only hole in the property supplier solution is control over ordering. Specifically, the ability to have alternative config sources like files to have higher priority than system properties and environment variables.

@radcortez is being able to override system props and environment variables with alternative sources critical to quarkus?

radcortez:
> @radcortez is being able to override system props and environment variables with alternative sources critical to quarkus?

It is part of the MicroProfile Config API (being moved to Jakarta), which we use for all our configurations. It is not critical for Quarkus, but we cannot really control how users set up their sources.

We don't really need to set alternate sources when calling the OTel API. Implementing our own `ConfigProperties` allowed us to bridge with our own configuration and pick everything up. That is why I was proposing to keep `setConfig` public.

Alternatively, I could supply the properties myself, but I would need a way to specify that they should have priority over everything (because when we lookup a Quarkus config, we already query all available sources and return the highest priority source configuration, including system and environment).

anuraaga:
@radcortez Sorry for the late response - I'm still not too enthusiastic about this change, the autoconfigure module is an implementation of the spec's SDK environment variables spec

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md

Allowing custom `Config` allows OTel distributions that deviate from this spec, which is not what this autoconfigure is intended to allow. For example, the Config may ignore all env variables, or map keys differently. Our intent is for all Java apps using autoconfigure to have similar user experience and this applies to Quarkus too - it seems like our standard env vars / properties should work as is, with also being able to source properties from Quarkus config afterwards to provide an experience consistent for OTel users as well as Quarkus users. Are there any concrete UX gaps you see from OTel spec being prioritized?

I'm not completely opposed to the change but providing a hook that can completely change the autoconfiguration experience draws pause. Hoping for more thoughts. @kenfinnigan do you happen to have any ideas on this?

radcortez:
> @radcortez Sorry for the late response

No worries :)

>  Are there any concrete UX gaps you see from OTel spec being prioritized?

The only issue here is when users set up their own config sources that may have a higher priority than system properties or environment variables. I guess we could document an exception for the OTel configuration and refer that we follow the OTel way of doing configuration.

 > I'm not completely opposed to the change but providing a hook that can completely change the autoconfiguration experience draws pause.

There is an API to pass a Map with defaults if the configuration cannot be found in system / env. Maybe if there is also a Map that you can use to pass concrete values and skip the query to system / env is a more acceptable change?

anuraaga:
A map to skip the query system could have the same issue of possibly not being compliant with the env var spec.

I'd recommend for now sticking with and documenting the caveat that env vars / props will take priority over user config to follow OTel spec and if a user has an issue with that later we can revisit with that concrete case as it would still be a compatible change in the future I believe. Would that be ok?

radcortez:
Ok, let's forget it and revisit this if required. Thanks!

PR #4004, opened 2021-12-16
Customizer for SdkTracerProviderBuilder

### LINKED ISSUE #3964: AutoConfiguration of the IdGenerator
Provide a way to set the`IdGenerator` with auto-configuration. Right now, this seems to be the only option from the `SdkTracerProviderBuilder` not exposed with auto-configuration.

There are multiple ways to do this:
- Add the desired instance of the `IdGenerator` as a parameter in the `io.opentelemetry.sdk.autoconfigure.TracerProviderConfiguration#configureTracerProvider` and work the way up from there so it can be exposed.
- A new configuration property that takes the FQN of the `IdGenerator` to use
- ServiceLoader

**Additional context**
If you want to use the AWS `IdGenerator`, you need to manually set it up with the `TracerProvider`, but if you are using auto-configuration, you can only use the random default one.

radcortez:
- Fixes #3964

### DIFF
--- a/sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/AutoConfigurationCustomizer.java
+++ b/sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/AutoConfigurationCustomizer.java
@@ -7,6 +7,7 @@ package io.opentelemetry.sdk.autoconfigure.spi;
 
 import io.opentelemetry.context.propagation.TextMapPropagator;
 import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.sdk.trace.SdkTracerProviderBuilder;
 import io.opentelemetry.sdk.trace.export.SpanExporter;
 import io.opentelemetry.sdk.trace.samplers.Sampler;
 import java.util.Map;
@@ -66,4 +67,16 @@ public interface AutoConfigurationCustomizer {
    */
   AutoConfigurationCustomizer addPropertiesSupplier(
       Supplier<Map<String, String>> propertiesSupplier);
+
+  /**
+   * Adds a {@link BiFunction} to invoke the with the {@link SdkTracerProviderBuilder} to allow
+   * customization. The return value of the {@link BiFunction} will replace the passed-in argument.
+   *
+   * <p>Multiple calls will execute the customizers in order.
+   */
+  default AutoConfigurationCustomizer addTracerProviderCustomizer(
+      BiFunction<SdkTracerProviderBuilder, ConfigProperties, SdkTracerProviderBuilder>
+          tracerProviderCustomizer) {
+    return this;
+  }
 }
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
@@ -20,6 +20,7 @@ import io.opentelemetry.sdk.logs.SdkLogEmitterProvider;
 import io.opentelemetry.sdk.metrics.SdkMeterProvider;
 import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.sdk.trace.SdkTracerProvider;
+import io.opentelemetry.sdk.trace.SdkTracerProviderBuilder;
 import io.opentelemetry.sdk.trace.export.SpanExporter;
 import io.opentelemetry.sdk.trace.samplers.Sampler;
 import java.util.Collections;
@@ -44,6 +45,8 @@ public final class AutoConfiguredOpenTelemetrySdkBuilder implements AutoConfigur
 
   @Nullable private ConfigProperties config;
 
+  private BiFunction<SdkTracerProviderBuilder, ConfigProperties, SdkTracerProviderBuilder>
+      tracerProviderCustomizer = (a, unused) -> a;
   private BiFunction<? super TextMapPropagator, ConfigProperties, ? extends TextMapPropagator>
       propagatorCustomizer = (a, unused) -> a;
   private BiFunction<? super SpanExporter, ConfigProperties, ? extends SpanExporter>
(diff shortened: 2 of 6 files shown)


anuraaga:
Thanks for helping @radcortez - sorry for not being clear. Instead of a knob only for id generator, which would require lots of duplicated knobs, I was thinking just `addTracerProviderCustomizer(BiFunction<SdkTracerProviderBuilder, ConfigProperties, SdkTracerProviderBuilder)` which you would be able to use to set ID generator I think. Would that be OK?

radcortez:
> Thanks for helping @radcortez - sorry for not being clear. Instead of a knob only for id generator, which would require lots of duplicated knobs, I was thinking just `addTracerProviderCustomizer(BiFunction<SdkTracerProviderBuilder, ConfigProperties, SdkTracerProviderBuilder)` which you would be able to use to set ID generator I think. Would that be OK?

Sure, I'll update this.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
```diff
@@ -175,15 +193,20 @@ public AutoConfiguredOpenTelemetrySdk build() {
       }
     }
 
+    SdkTracerProviderBuilder tracerProviderBuilder = SdkTracerProvider.builder();
     ConfigProperties config = getConfig();
+    tracerProviderBuilder = tracerProviderCustomizer.apply(tracerProviderBuilder, config);
+
+    // TODO - with the builder customizer, the user may set stuff that is later overridden by us
     Resource resource = ResourceConfiguration.configureResource(config, resourceCustomizer);
+    tracerProviderBuilder.setResource(resource);
```
radcortez, line 202: The issue here is that exposing the builder, allows the consumer to set the `Resource`, that we override with our own lookup. This also happens with other stuff.

Should we also add a check if things we set into the builder directly, don't override them? I think it makes sense.

anuraaga: @radcortez Sorry for taking so long to get back to you. We already have an SPI for tracerprovider customization that has the same issue. Let's not worry about it for now as it's a small corner case, and we can address if everyone finds it as a bug. We just want this builder to have the same functionality as our current SPI

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
```diff
@@ -69,6 +72,21 @@ AutoConfiguredOpenTelemetrySdkBuilder setConfig(ConfigProperties config) {
     return this;
   }
 
+  /**
+   * Adds a {@link BiFunction} to invoke the with the {@link SdkTracerProviderBuilder} to allow
+   * customization. The return value of the {@link BiFunction} will replace the passed-in argument.
+   *
+   * <p>Multiple calls will execute the customizers in order.
+   */
+  public AutoConfiguredOpenTelemetrySdkBuilder addTracerProviderCustomizer(
```
anuraaga, line 81: Can you also add to this interface?

https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/AutoConfigurationCustomizer.java

radcortez: Sure.

anuraaga (approved):

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkTracerProviderBuilder.java
```diff
@@ -119,6 +120,22 @@ public SdkTracerProviderBuilder setSampler(Sampler sampler) {
...
+   *
+   * <p>Skipping the registration of the shutdown hook may cause unexpected behavior. This
+   * configuration is for SDK consumers that require control over the SDK lifecycle. In this case,
+   * alternatives must be provided by the SDK consumer to shut down the SDK.
+   *
+   * @param registerShutdownHook a boolean <code>true</code> will register the hook, otherwise
+   *     <code>false</code> will skip registration.
+   * @return this
+   */
+  public SdkTracerProviderBuilder setRegisterShutdownHook(boolean registerShutdownHook) {
```
jkwatson, line 134: I'd prefer this were just `registerShutdownHook` without the `set`. We have no reason we need to adhere to the very antiquated javabean spec here.

anuraaga (changes requested):
Sorry I didn't notice https://github.com/open-telemetry/opentelemetry-java/pull/4004/commits/12c48250cbcab5c9c20661d02ad7a7062eb293df had been added here when approving - can we go ahead and keep them separate PRs?

Sorry if I was too unclear, we want to add the knob on https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java, not this one. The SDK should not itself register any shutdown hooks, that's only autoconfiguration.

radcortez:
Sure, sorry. I'll do that tomorrow.

sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/AutoConfigurationCustomizer.java
```diff
@@ -66,4 +67,8 @@ AutoConfigurationCustomizer addSpanExporterCustomizer(
    */
   AutoConfigurationCustomizer addPropertiesSupplier(
       Supplier<Map<String, String>> propertiesSupplier);
+
+  AutoConfigurationCustomizer addTracerProviderCustomizer(
+      BiFunction<SdkTracerProviderBuilder, ConfigProperties, SdkTracerProviderBuilder>
+          tracerProviderCustomizer);
```
jack-berg, line 73: Need a default implementation for this since the `:sdk-extensions:autoconfigure-spi` module is stable.

jack-berg (approved):

anuraaga (approved):

PR #4009, opened 2021-12-17
Reduce volatile reads in GlobalOpenTelemetry.get

anuraaga:


trask (approved):

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
```diff
@@ -51,9 +51,11 @@ private GlobalOpenTelemetry() {}
...
   public static OpenTelemetry get() {
-    if (globalOpenTelemetry == null) {
+    OpenTelemetry openTelemetry = globalOpenTelemetry;
+    if (openTelemetry == null) {
       synchronized (mutex) {
-        if (globalOpenTelemetry == null) {
+        openTelemetry = globalOpenTelemetry;
+        if (openTelemetry == null) {
 
           OpenTelemetry autoConfigured = maybeAutoConfigure();
```
bogdandrutu, line 60: not related to your change (probably in a separate PR), if `maybeAutoConfigure` can be renamed would be good to have it as `makeAutoConfigureAndSet` because it is confusing that you don't set on that path.

bogdandrutu (approved):

jkwatson (approved):

jsuereth (approved):

jack-berg (approved):

PR #4015, opened 2021-12-20
Add log autoconfigure support

jack-berg:
Initial autoconfigure support for logs. Caveats:
- No SPI hooks
- Only supported exporter is OTLP
- No configuration for `BatchLogProcessor`


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/LogEmitterProviderConfiguration.java
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  // Visible for testing
+  @Nullable
+  static LogExporter configureExporter(
+      String name, ConfigProperties config, MeterProvider meterProvider) {
+    switch (name) {
+      case "otlp":
+        return configureOtlpLogs(config, meterProvider);
+      default:
```
trask, line 89: support for "logging" (SystemOutLogExporter)?

trask (approved):

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/LogEmitterProviderConfiguration.java
jack-berg: Wasn't sure. Its not mentioned as an exporter option in the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#exporter-selection), but none of other the logging exporters we allow are either. 

Can certainly do that if there's no reason not to.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/DefaultConfigProperties.java
```diff
@@ -156,6 +162,29 @@ public Duration getDuration(String name) {
     return filterBlanksAndNulls(value.split(","));
   }
 
+  /**
+   * Returns {@link ConfigProperties#getList(String)} as a {@link Set} after validating there are no
+   * duplicate entries.
+   *
+   * @throws ConfigurationException if {@code name} contains duplicate entries
```
jkwatson, line 169: is it really a problem if there's a duplicate?

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/LogEmitterProviderConfiguration.java
jkwatson: I have no problem with it.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/DefaultConfigProperties.java
jkwatson: I see this was existing behavior from the span exporter config... I wonder why it matters?

jack-berg: I can't think of a valid configuration where you would want to use the same exporter multiple times, especially since there's no way to have different config for each of the exporters. Until a use case for that comes up, I think its better to fail to help users avoid an accidental invalid config.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/LogEmitterProviderConfiguration.java
jack-berg: Included a commit to add support for this.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/DefaultConfigProperties.java
jkwatson: It doesn't seem harmful to me if we were to ignore duplicates, but I'm also ok with keeping the existing behavior.

trask:
putting in a request for this to make it into the release planned for end of this week (this will enable the recently added javaagent logging/appender instrumentation)

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/LogEmitterProviderConfigurationTest.java
```diff
@@ -0,0 +1,45 @@
...
+            resource, DefaultConfigProperties.createForTest(properties), MeterProvider.noop());
+    try {
+      assertThat(logEmitterProvider)
+          .extracting("sharedState")
+          .satisfies(
+              sharedState -> {
+                assertThat(sharedState).extracting("resource").isEqualTo(resource);
+                assertThat(sharedState)
+                    .extracting("logProcessor")
+                    .isEqualTo(LogProcessor.composite());
```
jkwatson, line 39: I'm not sure this is really verifying that we have a no-op. Just that our processor is a composite. Can we make a stronger assertion here?

sdk-extensions/autoconfigure/README.md
```diff
@@ -39,39 +39,47 @@ for more details.
...
-| otel.experimental.exporter.otlp.retry.enabled | OTEL_EXPERIMENTAL_EXPORTER_OTLP_RETRY_ENABLED | If `true`, enable [experimental retry support](#otlp-exporter-retry). Default is `false`. |
+| System property       | Environment variable  | Purpose                                                                                                                    |
+|-----------------------|-----------------------|----------------------------------------------------------------------------------------------------------------------------|
+| otel.traces.exporter  | OTEL_TRACES_EXPORTER  | List of exporters to be used for tracing, separated by commas. Default is `otlp`. `none` means no autoconfigured exporter. |
+| otel.metrics.exporter | OTEL_METRICS_EXPORTER | The exporter to be used for metrics. Default is `none`.                                                                    |
+| otel.logs.exporter    | OTEL_LOGS_EXPORTER    | The exporter to be used for logs. Default is `none`.                                                                       |
```
jkwatson, line 46: since this hasn't been specced yet (I assume!), shouldn't this be in the `experimental` namespace until it is?

jack-berg: Oddly enough, it has been [specced](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#exporter-selection).

There is no spec for configuring the log batch exporter settings. I'll probably add those in a future commit under an experimental namespace.

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/LogEmitterProviderConfigurationTest.java
jack-berg: It is verifying it, but in a cheating way:`LogProcessor.composite()` returns a `NoopLogProcessor`. Its doing it this way because `NoopLogProcessor` is package private and inaccessible to both `LogEmitterProviderConfiguration` and the test code. Would have to do something like this to validate its a noop:
```
      .isInstanceOfSatisfying(
          LogProcessor.class,
          logProcessor ->
              assertThat(logProcessor.getClass().getName())
                  .isEqualTo("io.opentelemetry.sdk.logs.NoopLogProcessor"));
```

Arguably, verifying that it's a `NoopLogProcessor` is beyond the scope of the test. `LogEmitterProviderConfiguration` doesn't do anything except add 0 or more `LogProcessor`s to the `SdkLogEmitterProviderBuilder`. Its not till `LogEmitterSharedState` is constructed in the `SdkLogEmitterProvider` constructor that the list of configured log processors is translated to a single log processor via `LogProcessor.composite(LogProcessor...)`. 

By having the test verify that the log processor is equal `LogProcessor.composite()` (with zero log processors), we verify that `LogEmitterProviderConfiguration` did not add any.

jkwatson: I totally missed that `composite` took varargs processors. 👍🏽

anuraaga (approved):

Issue #3549, opened 2021-08-25
Registering multiple asynchronous instruments


fabianishere:
**Describe the bug**
I have been experimenting with the latest OpenTelemetry Metrics release (version 1.5.0-alpha), but it has occurred to me that registering multiple asynchronous instruments with the same name will result in only one of the callbacks being invoked during a collection cycle.

I want to implement a pattern like this:
```kotlin
public class Host(public val uid: UUID, meter: Meter, ....) {
    init {
        meter.gaugeBuilder("cpu.usage")
            .setDescription("The amount of CPU resources used by the host")
            .setUnit("MHz")
            .buildWithCallback { result ->
                val cpuUsage = ...
                result.observe(cpuUsage, Attributes.of(ResourceAttributes.HOST_ID, uid.toString()))
            }
    }
}
```

**Steps to reproduce**
1. Create multiple asynchronous instruments with the same name
2. Observe that only one callback is invoked during the collection cycle

**What did you expect to see?**
Every asynchronous instrument with the same should have its callback invoked every collection cycle.

**What did you see instead?**
Only one of the asynchronous instrument callbacks is invoked every collection cycle.

**What version and what artifacts are you using?**
This is the Gradle version catalog I use for the OpenTelemetry artifacts:
```toml
[versions]
opentelemetry-main = "1.5.0"
opentelemetry-metrics = "1.5.0-alpha"
opentelemetry-semconv = "1.5.0-alpha"

[libraries]
opentelemetry-api-main = { module = "io.opentelemetry:opentelemetry-api", version.ref = "opentelemetry-main" }
opentelemetry-sdk-main = { module = "io.opentelemetry:opentelemetry-sdk", version.ref = "opentelemetry-main" }
opentelemetry-api-metrics = { module = "io.opentelemetry:opentelemetry-api-metrics", version.ref = "opentelemetry-metrics" }
opentelemetry-sdk-metrics = { module = "io.opentelemetry:opentelemetry-sdk-metrics", version.ref = "opentelemetry-metrics" }
opentelemetry-semconv = { module = "io.opentelemetry:opentelemetry-semconv", version.ref = "opentelemetry-semconv" }
```

**Environment**
Compiler: Oracle JDK 16.0.2
OS: macOS Big Sur

jsuereth:
This is a known limitation right now in the API.  I've been holding off on fixing this prior to aggregator refactor.  I'm happy to prioritize it next, before finishing up exemplar or multi-export.

[...]

PR #4020, opened 2021-12-21
Log warning when duplicate async instruments are registered

### LINKED ISSUE #4005: Log when duplicate async instruments are registered
If I use the API to register two async instruments with the same name, but different callback behavior, the first one registered wins. The callback of the second is never invoked, and there are no logs to tell me that.

For example:
```
    AtomicLong countUp = new AtomicLong(0);
    AtomicLong countDown = new AtomicLong(100);
    meter.gaugeBuilder("foo").buildWithCallback(observe -> observe.record(countUp.incrementAndGet()));
    meter.gaugeBuilder("foo").buildWithCallback(observe -> observe.record(countDown.decrementAndGet()));
```

We should provide useful log messages when conflicts like this happen since unlike synchronous instruments, we can't just return the same instrument that was previously registered.

jack-berg:
Resolves #4005.

The example in the issue now yields the following results:
```
    AtomicLong countUp = new AtomicLong(0);
    AtomicLong countDown = new AtomicLong(100);
    meter.gaugeBuilder("foo").buildWithCallback(observe -> observe.record(countUp.incrementAndGet()));
    meter.gaugeBuilder("foo").buildWithCallback(observe -> observe.record(countDown.decrementAndGet()));

....
WARNING: Found duplicate metric definition: foo
	at unknown source
		To enable better debugging, run your JVM with -Dotel.experimental.sdk.metrics.debug=true
Causes
- InstrumentType [OBSERVABLE_GAUGE] is async and already registered
```


kubawach (approved):
Looks good, although codecov complains about new cov < 80% ;)

jkwatson (approved):

jsuereth:
There's one valid use case of multiple registrations that I would like an answer to (via batch API) before we lock this down.  Specifically, registering callbacks that report different attributes.


```
    AtomicLong countUp = new AtomicLong(0);
    Attributes countUpAttr = ...;
    meter.gaugeBuilder("foo").buildWithCallback(observe -> observe.record(countUp.incrementAndGet(), countUpAtttr));
    
    AtomicLong countDown = new AtomicLong(100);
    Attributes countDownAttr = ...;
    meter.gaugeBuilder("foo").buildWithCallback(observe -> observe.record(countDown.decrementAndGet(), countDownAttr));
}
```

Nominally, users could just put each observation in a single callback and maybe that's ok.   Just want to make sure we document HOW we want users to do that, especially if we lock down use cases.

jack-berg:
> Nominally, users could just put each observation in a single callback and maybe that's ok. Just want to make sure we document HOW we want users to do that, especially if we lock down use cases.

Small clarification - this PR doesn't change the behavior. If you register multiple callbacks today the first will be called and subsequent ones are silently ignored. This just tries to help users debug why their callback isn't being called. 

I think we have to recommend that for now. An option for users is to use a helper function to transform several callbacks into a composite callback:

```
    meter.gaugeBuilder("foo").buildWithCallback(compositeCallback(
        observer -> observer.record(countUp.incrementAndGet(), upAttrs),
        observer -> observer.record(countDown.incrementAndGet(), downAttrs)
    ));

  ...

  static Consumer<ObservableDoubleMeasurement> compositeCallback(
      Consumer<ObservableDoubleMeasurement>... callbacks) {
    return observer -> {
      for (Consumer<ObservableDoubleMeasurement> callback : callbacks) {
        callback.accept(observer);
      }
    };
  }
```

jack-berg:
There's an outstanding issue related to registering multiple async callbacks [here](https://github.com/open-telemetry/opentelemetry-java/issues/3549).

I believe its possible to support this type of use case by. We'd simply have to make `AsynchronousMetricStorage` instances mutable such that when another callback is registered for the same instrument, that the storage is updated to have a callback which is a composite of the original and the new. 

However, the issue references that the spec is the limiting factor right now, which is likely due to the [following section](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument):

> Meter implementations MUST return an error when multiple Instruments are registered under the same Meter instance using the same name.

Thinking about it more and my other [comment](https://github.com/open-telemetry/opentelemetry-java/pull/4020#issuecomment-1001713645), a user could implement their own mutable callback to solve that problem. For example:

```
    MutableCallback mutableCallback = new MutableCallback();
    meter.gaugeBuilder("foo").buildWithCallback(mutableCallback);
    mutableCallback.add("up", observer -> observer.record(countUp.incrementAndGet(), upAttrs));
    mutableCallback.add("down", observer -> observer.record(countDown.decrementAndGet(), downAttrs));

  ...

  static class MutableCallback implements Consumer<ObservableDoubleMeasurement> {
    private final Map<String, Consumer<ObservableDoubleMeasurement>> callbacks = new HashMap<>();

    void add(String id, Consumer<ObservableDoubleMeasurement> callback) {
      callbacks.put(id, callback);
    }

    void remove(String id) {
      callbacks.remove(id);
    }

    @Override
    public void accept(ObservableDoubleMeasurement observable) {
      List<Consumer<ObservableDoubleMeasurement>> callbacksCopy = new ArrayList<>(callbacks.values());
      for (Consumer<ObservableDoubleMeasurement> callback : callbacksCopy) {
        callback.accept(observable);
      }
    }
  }
```

mateuszrzeszutek:
> There's one valid use case of multiple registrations that I would like an answer to (via batch API) before we lock this down. Specifically, registering callbacks that report different attributes.

Just for the record, I've encountered exactly this case in my Micrometer instrumentation -- ended up implementing a mutable composite callback, works fine; especially when you have to stop/remove certain measurements.

anuraaga (approved):
Let's get this logging in, and if we do in the future allow multiple callbacks even with the same API, it would be accepting more input than before by removing the log message and a semver-comaptible change

Issue #4013, opened 2021-12-20
Allow to unregister/stop/destroy metric instruments


mateuszrzeszutek:
**Is your feature request related to a problem? Please describe.**
Currently there is no way to stop/remove an instrument after it is created (see https://github.com/open-telemetry/opentelemetry-specification/issues/2232 for more details). This is a major difference when you compare OTel metrics API to Dropwizard or Micrometer, both of which allow removing instruments.

**Describe the solution you'd like**
For synchronous instruments we could just add a `stop()` (the name being whatever the spec decides) method to the interfaces representing each synchronous instrument.
Things get a bit more complicated with asynchronous instruments - right now `buildWithCallback()` methods return `void`, there's no object/concept in the API that would represent an already started async instrument. In case the spec issue is not resolved before metrics API going stable it will be pretty difficult to offer forward compatibility. How about we add initially empty interfaces for all async instruments and return them in `buildWithCallback` methods:
```java
interface ObservableLongCounter {
}

// ...
interface LongCounterBuilder {
    ObservableLongCounter buildWithCallback(Consumer<ObservableLongMeasurement> callback);
}
```
Then we could just add methods to `ObservableLongCounter` in a forward-compatible manner once the spec issue is fixed.

**Additional context**
https://github.com/open-telemetry/opentelemetry-specification/issues/2232
https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4919

anuraaga:
@mateuszrzeszutek Thanks for bringing this up - it does seem to make sense to return an empty interface and we need to do it before 1.10.0

anuraaga:
For naming, somehow e.g., `LongObservingCounter` feels most natural to me but `Observable` would be fine too - it seems like a counter that is observing a long value.

mateuszrzeszutek:
I just used the spec recommendation for the name

> It is highly recommended that implementations use the name `ObservableCounter` (or any language idiomatic variation

`LongObservingCounter` is perfectly fine too; anything that mentions "observe" + instrument name will do.

jkwatson:
We could add this later with an additional `build` method on the builders.

anuraaga:
@jkwatson Could you describe your idea? Is it `SyncThing build()` and `AsyncThing build(callback)`? Since adding a return type for callback is compile compatible, but not ABI compatible, I think it's best to do it now with an empty interface, though it's still possible in the future for a differently named method.

jkwatson:
[...]

PR #4022, opened 2021-12-22
Return an interface when registering metric callbacks.

anuraaga:
This gives us flexibility to provide functionality on a registered observable metric in the future, notably the ability to unregister them. I think it's likely we will add it after some spec discussion since it is good practice to allow any registered stateful thing to be unregistered - it seems idiomatic compared to `ScheduledFuture` returned from `ScheduledExecutorService` for example.

Helps with #4013 

@jkwatson I didn't quite follow the idea in https://github.com/open-telemetry/opentelemetry-java/issues/4013#issuecomment-998955236 but let me know if you want to try a different pattern. Return value, similar to `ScheduledFuture` seems best to me right now.


anuraaga:
Also one point to note is that this is compile-compatible, any existing apps will still compile without change when updating from 1.10.0-RC1 to 1.10.0-RC2 / stable. It's not binary-compatible so to avoid adding another method in the future, we would need to do this now. I expect there is enough of a chance of wanting to add APIs here to do this proactively, especially given it's not invasive for existing users in any way.

jsuereth:
My $.02 is that you should return a unique type per-instrument.   If we want this to future proof method on async instruments, this doesn't provide enough flexibility as-is.

Note: Originally the metrics API returned instruments on async registration.  I removed it, to make room for structural changes internally.  While I still think we need to make those changes, I don't think we have a hard dependency on returned Instrument types in Async like we did previously, so this should be fine.

anuraaga:
> If we want this to future proof method on async instruments, this doesn't provide enough flexibility as-is.

We'd be able to add a type parameter in the future as in my suggestion, but it doesn't have to be Observable*Measurement and could be instrument-specific. So I think it does have the flexibility, but such an approach may make using that future functionality a bit more awkward. I'd probably lean towards it if we think instrument-specific logic is not likely, or specific return types if we think it's likely.

anuraaga:
Well - I guess the specific return types don't have a significant issue either let me go with that, will update

anuraaga:
Updated

jsuereth:
> We'd be able to add a type parameter in the future as in my suggestion, but it doesn't have to be Observable*Measurement and could be instrument-specific. So I think it does have the flexibility, but such an approach may make using that future functionality a bit more awkward. I'd probably lean towards it if we think instrument-specific logic is not likely, or specific return types if we think it's likely.

I'm not really sure this leads to the best API.  We don't really know if we have a uniform API per-async instrument but we KNOW they differ in important ways (at least Gauge vs. Counter), and we know MOST APIs we're adapting have diverging methods per-instrument, so it's more likely we diverge than stay the same.

Again, we're talking unknown unknowns, but I think it's far better w/ per-instrument async return value (AsynchronousGauge) vs. a single "Observable".

jsuereth (approved):

kubawach (approved):
Looks fine :)

jkwatson (approved):
Do we need another RC release with this change, do you think?

anuraaga:
Yeah planning on doing one today

api/all/src/main/java/io/opentelemetry/api/metrics/ObservableLongUpDownCounter.java
```diff
@@ -0,0 +1,17 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.api.metrics;
+
+import java.util.function.Consumer;
+
+/**
+ * A reference to an observable metric registered with {@link
+ * DoubleGaugeBuilder#buildWithCallback(Consumer)}.
```
jack-berg, line 12: Typo here.

api/all/src/main/java/io/opentelemetry/api/metrics/ObservableDoubleUpDownCounter.java
```diff
@@ -0,0 +1,17 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.api.metrics;
+
+import java.util.function.Consumer;
+
+/**
+ * A reference to an observable metric registered with {@link
+ * DoubleGaugeBuilder#buildWithCallback(Consumer)}.
```
jack-berg, line 12: Typo here.

jack-berg (approved):

mateuszrzeszutek:
Nice 👍 
Thanks!

PR #4025, opened 2021-12-23
Reduce metrics stress test operations

anuraaga:
Currently the stress tests take several seconds, which I think is too much for unit tests that don't do I/O (even I/O tests are usually 300ms for first, and 10-20ms for remaining reusing the same server). It seemed OK to reduce the number of operations a lot, let me know if this seems reasonable and I will apply to the rest of the instruments.

### DIFF
--- a/sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/SdkLongUpDownCounterTest.java
+++ b/sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/SdkLongUpDownCounterTest.java
@@ -186,10 +186,10 @@ class SdkLongUpDownCounterTest {
     for (int i = 0; i < 4; i++) {
       stressTestBuilder.addOperation(
           StressTestRunner.Operation.create(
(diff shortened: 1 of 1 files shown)


anuraaga:
BTW, with this change the stress test is ~400ms on my fast VM. It's much better but still pretty long, let me know if further reduction of the operations is feasible (or deleting them)

jsuereth:
These tests were pretty critical for catching some hairy thread-related issues during the rewrite.   I'm not certain what number we can reduce to, it's a tradeoff of execution time vs. coverage at some point.  It's reasonable to try to fit these into a decent window, but I know from experience that these are valuable when touching any of the core aspects of synchronous collection.

anuraaga:
@jsuereth Any suggestion on how to proceed? I think the current tests are too slow and we do need to tweak them in some way.

jsuereth:
@anuraaga I was just trying to call out that I think deleting them (without suitable replacement) would be a mistake.   If you want to reduce them (or find a convenient way to skip over them) that's fine.   If I could remember what I had done that broke the hell out of them (something w/ the AggregatorHandle), I'd rebreak it and see how far we can reduce :)

PR #4030, opened 2021-12-27
Metrics API noop implementation

jack-berg:
- Optimize noop meter implementation by using singletons instead of instantiating new instances
- Reduce visibility of `DefaultMeter` internal classes from public to private
- Reduce visibility of `DefaultMeter#getInstance()` and `DefaultMeterProvider#getInstance()` from public to package private


jkwatson (approved):
👍🏽

anuraaga (approved):
Thanks! Had missed these

PR #4039, opened 2021-12-31
Fix the problem that README.md failed to display [javadoc-image]

JooKS-me:
There is a problem with the url in `sdk-extensions/autoconfigure/README.md`.


jkwatson (approved):
Thanks!

PR #4041, opened 2022-01-03
Remove grpc dependency from examples

### LINKED ISSUE #4040: Ran JaegerExample but got exception - UnsupportedOperationException: Only available on DefaultGrpcExporter
**Describe the bug**
I tried to run opentelemetry-java/examples/jaeger. I followed the steps described here, 
https://github.com/open-telemetry/opentelemetry-java/tree/main/examples/jaeger
In step3, when run this command, "java -cp build/libs/opentelemetry-examples-jaeger-0.1.0-SNAPSHOT-all.jar io.opentelemetry.example.jaeger.JaegerExample localhost 14250",
I got the following error
Exception in thread "main" java.lang.UnsupportedOperationException: Only available on DefaultGrpcExporter
 at io.opentelemetry.exporter.otlp.internal.grpc.OkHttpGrpcExporterBuilder.setChannel(OkHttpGrpcExporterBuilder.java:55)
 at io.opentelemetry.exporter.otlp.internal.grpc.OkHttpGrpcExporterBuilder.setChannel(OkHttpGrpcExporterBuilder.java:31)
 at io.opentelemetry.exporter.jaeger.JaegerGrpcSpanExporterBuilder.setChannel(JaegerGrpcSpanExporterBuilder.java:51)
 at io.opentelemetry.example.jaeger.ExampleConfiguration.initOpenTelemetry(ExampleConfiguration:40)
 at io.opentelemetry.example.jaeger.JaegerExample.main(JaegerExample:44)

openTelemetryVersion = "1.9.1"
https://github.com/open-telemetry/opentelemetry-java/blob/main/examples/build.gradle

**Environment**
Compiler: (e.g., "AdoptOpenJDK 16.0.1+9")
OS: (e.g., "macOS Big Sur  version 11.6.1")

**Additional context**

jack-berg:
Resolves #4040.

Another option would be to use `setEndpoint(String)` instead of `setChannel(ManagedChannel)` for [configuration](https://github.com/open-telemetry/opentelemetry-java/blob/main/examples/jaeger/src/main/java/io/opentelemetry/example/jaeger/ExampleConfiguration.java#L40).

### DIFF
--- a/examples/jaeger/README.md
+++ b/examples/jaeger/README.md
@@ -27,7 +27,7 @@ docker run --rm -it --name jaeger\
 
 ## 3 - Start the Application
 ```shell script
-java -cp build/libs/opentelemetry-examples-jaeger-0.1.0-SNAPSHOT-all.jar io.opentelemetry.example.jaeger.JaegerExample localhost 14250
+java -cp build/libs/opentelemetry-examples-jaeger-0.1.0-SNAPSHOT-all.jar io.opentelemetry.example.jaeger.JaegerExample http://localhost:14250
 ```
 ## 4 - Open the Jaeger UI
 
--- a/examples/jaeger/src/main/java/io/opentelemetry/example/jaeger/ExampleConfiguration.java
+++ b/examples/jaeger/src/main/java/io/opentelemetry/example/jaeger/ExampleConfiguration.java
@@ -5,8 +5,6 @@
 
 package io.opentelemetry.example.jaeger;
 
-import io.grpc.ManagedChannel;
-import io.grpc.ManagedChannelBuilder;
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.exporter.jaeger.JaegerGrpcSpanExporter;
@@ -26,18 +24,14 @@ class ExampleConfiguration {
   /**
    * Initialize an OpenTelemetry SDK with a Jaeger exporter and a SimpleSpanProcessor.
    *
-   * @param jaegerHost The host of your Jaeger instance.
-   * @param jaegerPort the port of your Jaeger instance.
+   * @param jaegerEndpoint The endpoint of your Jaeger instance.
    * @return A ready-to-use {@link OpenTelemetry} instance.
    */
-  static OpenTelemetry initOpenTelemetry(String jaegerHost, int jaegerPort) {
-    // Create a channel towards Jaeger end point
-    ManagedChannel jaegerChannel =
-        ManagedChannelBuilder.forAddress(jaegerHost, jaegerPort).usePlaintext().build();
+  static OpenTelemetry initOpenTelemetry(String jaegerEndpoint) {
     // Export traces to Jaeger
     JaegerGrpcSpanExporter jaegerExporter =
         JaegerGrpcSpanExporter.builder()
-            .setChannel(jaegerChannel)
+            .setEndpoint(jaegerEndpoint)
             .setTimeout(30, TimeUnit.SECONDS)
             .build();
 
--- a/examples/otlp/docker/docker-compose.yaml
+++ b/examples/otlp/docker/docker-compose.yaml
@@ -26,8 +26,7 @@ services:
       - "8888:8888"   # Prometheus metrics exposed by the collector
       - "8889:8889"   # Prometheus exporter metrics
       - "13133:13133" # health_check extension
-      - "55678"       # OpenCensus receiver
-      - "55681:55679" # zpages extension
+      - "55679:55679" # zpages extension
       - "4317:4317"   # otlp receiver
     depends_on:
       - jaeger-all-in-one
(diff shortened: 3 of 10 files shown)


exporters/jaeger/build.gradle.kts
```diff
@@ -42,8 +42,8 @@ wire {
   }
 }
 
-tasks {
-  compileJava {
-    source("$buildDir/generated/source/wire")
+sourceSets {
+  main {
+    java.srcDir("$buildDir/generated/source/wire")
```
jack-berg, line 47: Without this my IDE doesn't pick up the generated proto classes.

jkwatson (approved):

anuraaga:
Yeah let's change to `setEndpoint` that's what we want users to use so should be our example

exporters/jaeger/build.gradle.kts
anuraaga: Hmm - somehow it seems OK for me but maybe I'm sporadically in a good state. sourceSets does seem better

Can you make the same change to https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/otlp/common/build.gradle.kts#L72 ?

anuraaga:
So I guess let's go ahead and remove grpc-protobuf and grpc-netty-shaded from all the example projects

examples/jaeger/README.md
```diff
@@ -27,7 +27,7 @@ docker run --rm -it --name jaeger\
 
 ## 3 - Start the Application
 ```shell script
-java -cp build/libs/opentelemetry-examples-jaeger-0.1.0-SNAPSHOT-all.jar io.opentelemetry.example.jaeger.JaegerExample localhost 14250
+java -cp build/libs/opentelemetry-examples-jaeger-0.1.0-SNAPSHOT-all.jar io.opentelemetry.example.jaeger.JaegerExample http://localhost:14250
```
jack-berg, line 30: If we're going to use endpoint instead of a custom channel, then we ought to have the program accept a single endpoint argument in the format expected by `setEndpoint(String)`.

examples/otlp/docker/docker-compose.yaml
```diff
@@ -26,8 +26,7 @@ services:
       - "8888:8888"   # Prometheus metrics exposed by the collector
       - "8889:8889"   # Prometheus exporter metrics
       - "13133:13133" # health_check extension
-      - "55678"       # OpenCensus receiver
-      - "55681:55679" # zpages extension
```
jack-berg, line 30: There is no open census receiver configured, so no need for `55678`. 

`55681` is [used](https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/otlpreceiver/factory.go#L35) as the legacy oltp http port. Better to just use `55679` for zpages.

examples/otlp/src/main/java/io/opentelemetry/example/otlp/ExampleConfiguration.java
```diff
@@ -41,7 +41,11 @@ static OpenTelemetry initOpenTelemetry() {
     SdkTracerProvider tracerProvider =
         SdkTracerProvider.builder()
             .addSpanProcessor(spanProcessor)
-            .setResource(AutoConfiguredOpenTelemetrySdk.initialize().getResource())
```
jack-berg, line 44: This was causing an error because it sets `GlobalOpenTelemetry` by default, which is set again on line 47 (51 in the revised code) via `OpenTelemetrySdk.builder().setTracerProvider(tracerProvider).buildAndRegisterGlobal()`.

examples/otlp/src/main/java/io/opentelemetry/example/otlp/OtlpExporterExample.java
```diff
@@ -56,5 +56,7 @@ public static void main(String[] args) throws InterruptedException {
 
     // sleep for a bit to let everything settle
     Thread.sleep(2000);
+
+    System.out.println("Bye");
```
jack-berg, line 60: Currently with 1.9.1 the program hangs (applies to the jaeger example) instead of shutting down correctly. It shuts down properly with `1.10.0-SNAPSHOT`, which I believe is due to [the change to use daemon threads in the dispatcher](https://github.com/open-telemetry/opentelemetry-java/commit/74fb36b6cb79dd80dd0d57c330f85a110599254c). 

Anyway, having a log line here like the jaeger example was helpful to know that the example had actually done some work.

exporters/jaeger/build.gradle.kts
jack-berg: Done.

examples/jaeger/src/main/java/io/opentelemetry/example/jaeger/ExampleConfiguration.java
```diff
@@ -26,18 +24,14 @@
...
    * Initialize an OpenTelemetry SDK with a Jaeger exporter and a SimpleSpanProcessor.
    *
-   * @param jaegerHost The host of your Jaeger instance.
-   * @param jaegerPort the port of your Jaeger instance.
+   * @param jaegerEndpoint The endpoint of your Jaeger instance.
    * @return A ready-to-use {@link OpenTelemetry} instance.
    */
-  static OpenTelemetry initOpenTelemetry(String jaegerHost, int jaegerPort) {
-    // Create a channel towards Jaeger end point
-    ManagedChannel jaegerChannel =
```
anuraaga, line 35: Random aside - I was hoping most people weren't using `setChannel` in real apps but with it in our examples, now I'm less sure 😅

anuraaga (approved):

PR #4047, opened 2022-01-05
Add toString to SDK trace components and print autoconfigured SDK

anuraaga:
When debugging issues, it can be invaluable to know the resolved SDK configuration, for example a typo like `OTEL_TRACE_SAMPLER=traceidratio(0.1)` could result in all spans being traced despite seeming to have set the sampler. I think it's worth it for us to add toString to SDK components with their configuration, especially because we have autoconfigure, not programming-only configuration.

Note this was triggered by debugging a production issue related to sampling and wanting to know what the actual sampler being used is in a reliable way.

The toString isn't beautiful at all without newlines, I just stuck with what IntelliJ generates. But it doesn't necessarily need to be beautiful, just give the information which is currently not easy to get to (mostly in javaagent, where SDK is hidden into another classloader)

### DIFF
--- a/docs/apidiffs/current_vs_latest/opentelemetry-sdk.txt
+++ b/docs/apidiffs/current_vs_latest/opentelemetry-sdk.txt
@@ -4,6 +4,7 @@ Comparing source compatibility of  against
 	+++  NEW METHOD: PUBLIC(+) io.opentelemetry.api.metrics.MeterProvider getMeterProvider()
 	+++  NEW METHOD: PUBLIC(+) io.opentelemetry.sdk.logs.SdkLogEmitterProvider getSdkLogEmitterProvider()
 	+++  NEW METHOD: PUBLIC(+) io.opentelemetry.sdk.metrics.SdkMeterProvider getSdkMeterProvider()
+	+++  NEW METHOD: PUBLIC(+) java.lang.String toString()
 ***  MODIFIED CLASS: PUBLIC FINAL io.opentelemetry.sdk.OpenTelemetrySdkBuilder  (not serializable)
 	===  CLASS FILE FORMAT VERSION: 52.0 <- 52.0
 	+++  NEW METHOD: PUBLIC(+) io.opentelemetry.sdk.OpenTelemetrySdkBuilder setLogEmitterProvider(io.opentelemetry.sdk.logs.SdkLogEmitterProvider)
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
@@ -210,7 +210,8 @@ public final class AutoConfiguredOpenTelemetrySdkBuilder implements AutoConfigur
 
     if (setResultAsGlobal) {
       GlobalOpenTelemetry.set(openTelemetrySdk);
-      logger.log(Level.FINE, "Global OpenTelemetrySdk set to {0}", openTelemetrySdk);
+      logger.log(
+          Level.FINE, "Global OpenTelemetrySdk set to {0} by autoconfiguration", openTelemetrySdk);
     }
 
     return AutoConfiguredOpenTelemetrySdk.create(openTelemetrySdk, resource, config);
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
@@ -27,6 +27,8 @@ import java.util.Map;
 import java.util.ServiceLoader;
 import java.util.function.BiFunction;
 import java.util.function.Supplier;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import javax.annotation.Nullable;
 
 /**
@@ -35,6 +37,9 @@ import javax.annotation.Nullable;
  * behavior such as filtering out telemetry attributes.
  */
 public final class AutoConfiguredOpenTelemetrySdkBuilder implements AutoConfigurationCustomizer {
+
+  private static final Logger logger =
+      Logger.getLogger(AutoConfiguredOpenTelemetrySdkBuilder.class.getName());
 
   @Nullable private ConfigProperties config;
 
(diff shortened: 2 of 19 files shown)


anuraaga:
I will add metrics, etc in followups if this seems sound.

willarmiros (approved):
Thanks!

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
```diff
@@ -205,6 +210,7 @@ public AutoConfiguredOpenTelemetrySdk build() {
 
     if (setResultAsGlobal) {
       GlobalOpenTelemetry.set(openTelemetrySdk);
+      logger.log(Level.FINE, "Global OpenTelemetrySdk set to {0}", openTelemetrySdk);
```
jack-berg, line 213: Should this line go in `GlobalOpenTelemetry` instead?

docs/apidiffs/current_vs_latest/opentelemetry-sdk.txt
```diff
@@ -4,6 +4,7 @@ Comparing source compatibility of  against
 	+++  NEW METHOD: PUBLIC(+) io.opentelemetry.api.metrics.MeterProvider getMeterProvider()
 	+++  NEW METHOD: PUBLIC(+) io.opentelemetry.sdk.logs.SdkLogEmitterProvider getSdkLogEmitterProvider()
 	+++  NEW METHOD: PUBLIC(+) io.opentelemetry.sdk.metrics.SdkMeterProvider getSdkMeterProvider()
+	+++  NEW METHOD: PUBLIC(+) java.lang.String toString()
```
jkwatson, line 7: this is weird that jamicmp flags this, since `toString()` is on `Object`. 🤔

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -121,6 +121,11 @@ public CompletableResultCode forceFlush() {
     return worker.batch;
   }
 
+  @Override
+  public String toString() {
+    return "BatchSpanProcessor{" + "worker=" + worker + '}';
```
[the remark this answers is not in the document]
anuraaga: Yeah I was thinking of simplifying some of these but kept to letting IntelliJ take care of things. But agree, I should just go ahead and do that now.

docs/apidiffs/current_vs_latest/opentelemetry-sdk.txt
anuraaga: I believe it's because now it is possible for code to compile against `SdkMeterProvider.toString()` directly, so after adding these we can't remove them.

trask: oh interesting i hadn't thought about that before 👍

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
anuraaga: Hmm - not sure. I guess the logging should generally not be that helpful if a user is constructing their own SDK, it's really for resolving AutoConfiguration past all SPIs etc. Lean towards keeping it here.

docs/apidiffs/current_vs_latest/opentelemetry-sdk.txt
jkwatson: I'm missing something. Can you give me an example of what you mean?

anuraaga: If my app wrote this,
```
SdkMeterProvider provider = SdkMeterProvider.builder().build();
System.out.println(provider.toString());
```

Then the bytecode would link to `SdkMeterProvider.toString` since the concrete type of `provider` is `SdkMeterProvider`. This wouldn't be true for `System.out.println(provider)`, where within println the concrete type is `Object` and it links to `Object.toString()`, which virtually gets resolved to the subclass. A Java strangeness indeed

jkwatson: won't the runtime walk up the hierarchy either way, though? If this is true, I'm extremely surprised.

anuraaga: If you mean it would walk up to `Object.toString` if the bytecode said `invokevirtual SdkMeterProvider.toString` but we removed that method, no I don't think so, it's looking for that method only and will throw a link error.

jkwatson: I just tested this locally and it works as I expected. The runtime will walk up the hierarchy and call the parent implementation.

anuraaga: @jkwatson Showed me the bytecode too - sorry for red herring, was completely wrong and surprised by this!

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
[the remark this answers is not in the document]
anuraaga: Updated to remove the internal details from toString

jkwatson (approved):
thanks! This is something that has been needed for a long time, TBH

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
jack-berg: Yeah I buy that. Can we add some sort of indication that autoconfigure is the thing that is setting it? Something like "AutoConfigure setting Global OpenTelemetrySdk to {0}".

Also might be helpful to do this at a higher logging level. Pretty useful to know how your environment variables / system properties translated to the sdk config.

jack-berg (approved):

docs/apidiffs/current_vs_latest/opentelemetry-sdk.txt
trask: thx @jkwatson 🧐

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
anuraaga: I added autoconfiguration to the log message. I've left the level at least for now, if nothing else the output looks like sort of goop without any newlines / indenting so it might be a bit scary to log by default but we can revisit later I think.

PR #4048, opened 2022-01-05
Update errorprone

anuraaga:


api/all/src/main/java/io/opentelemetry/api/common/Attributes.java
```diff
@@ -140,6 +140,7 @@ static Attributes empty() {
    * Returns a {@link Attributes} instance with the given key-value pairs. Order of the keys is not
    * preserved. Duplicate keys will be removed.
    */
+  @SuppressWarnings("TooManyParameters")
```
jkwatson, line 143: lol.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/CodedOutputStream.java
```diff
@@ -62,7 +62,7 @@
 // - No support for writing fields with tag, we alway write tags separately
 // - Allow resetting and use a ThreadLocal instance
 //
-@SuppressWarnings({"UngroupedOverloads", "InlineMeSuggester"})
+@SuppressWarnings({"UnnecessaryFinal", "UngroupedOverloads", "InlineMeSuggester"})
```
jkwatson, line 65: why is this needed? Oh.... is this for final variables in the code?

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/CodedInputStream.java
```diff
@@ -49,6 +49,7 @@
  * <p>This class is internal and is hence not for public use. Its APIs are unstable and can change
  * at any time.
  */
+@SuppressWarnings("UnnecessaryFinal")
```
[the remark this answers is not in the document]
jack-berg: Is the `final` unnecessary because it has a private constructor?

jack-berg (approved):
Big fan of getting rid of the `final` keyword in methods where the var is already effectively final 👍

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/SynchronousMetricStorageTest.java
```diff
@@ -6,7 +6,6 @@
 package io.opentelemetry.sdk.metrics.internal.state;
 
 import static io.opentelemetry.sdk.testing.assertj.MetricAssertions.assertThat;
-import static io.opentelemetry.sdk.testing.assertj.OpenTelemetryAssertions.assertThat;
```
jkwatson, line 9: how was this not just a straight-up compiler error?

jkwatson (approved):
❤️  getting rid of all the un-needed `final` noise!

Oberon00:
Coming from C++, I view getting rid of final as bad. Final (like const in C/C++) means that you get a compiler error if you assign to the variable by accident.

jkwatson:
> Coming from C++, I view getting rid of final as bad. Final (like const in C/C++) means that you get a compiler error if you assign to the variable by accident. (Just adding this as a tidbit, don't view this as a vote for or against anything)

Sure...but if your methods are nice and tight and small, this shouldn't be an issue! It should be easy to see by inspection that you might be re-assigning a parameter. Also, much less of an issue when there are no pointers passed around to have caller side-effects in Java. :)

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/CodedInputStream.java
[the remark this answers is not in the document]
anuraaga: These two classes are mostly copied as is don't think we should drift them too much from upstream just for style checks

Issue #4056, opened 2022-01-05
Make SPI loader load implementations lazily to avoid instantiating unused implementations


ryanrupp:
**Is your feature request related to a problem? Please describe.**
I'm running into an issue that the SPI loader eagerly loading `Sampler` implementations can make it so that resources get acquired that I don't actually use, details:

Minor thing (mainly just some debug log noise + spawning another thread) but in my custom agent distro I'm using `AwsXrayIdGenerator` for Xray friendly IDs however the agent I have doesn't always run in the context of AWS (i.e. on-prem). The problem I'm running into is it looks the SPI loader for samplers seems to load all the samplers eagerly [here](https://github.com/open-telemetry/opentelemetry-java/blob/10f5efa14da30b4b6eb124a5c938f50904406186/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java#L150) which because I've included the AWS jar will then try to load `AwsXrayRemoteSampler` even though I don't have it enabled as the sampler. The `AwsXrayRemoteSampler` constructor is scheduling a thread [here](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSampler.java#L80) which will just fail to retrieve sampling info.

Technically AwsXrayRemoteSampler could initialize this thread lazily but it seems like it would be less error prone to push this responsibility onto the SPI loader which could lazily load custom samplers or only load the requested samplers. There's an argument [here](https://github.com/open-telemetry/opentelemetry-java/blob/10f5efa14da30b4b6eb124a5c938f50904406186/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/SpiUtil.java#L24) about which samplers were actually requested (so the default is a single value parentbased_always_on) but it's currently just using this to determine if a failed initialization should be logged as a warning or not.

[This](https://github.com/open-telemetry/opentelemetry-java/pull/3579) issue actually mentions the eager loading where the catch + log was added. Not sure if lazy loading these would potentially break something (even if unused).

Slack thread for reference [here](https://cloud-native.slack.com/archives/C014L2KCTE3/p1641412011206900).

**Describe the solution you'd like**
[...]

PR #4059, opened 2022-01-06
Make SPI factories instantiate implementations lazily

ryanrupp:
Previously, this would load all SPI factories and then eagerly instantiate each implementation. This would potentially result in implementations that would get instantiated even if they didn't end up getting used. An example was if the AWS contrib jar was included, this would result in an AwsXrayRemoteSampler being instantiated eagerly despite the configuration not using it. This could then result in errors or unused resources depending on the implementation i.e. in the AwsXrayRemoteSampler example it spawns a thread and makes a network request.

This change transitions the factories so that they only construct the implementation objects if the configuration specifically asks for it. Note the SPI factories themselves must still must be eagerly loaded to determine which factory names are available. This change applies to all auto-configuration areas following this pattern: Samplers, Metric Exporters, Text Map Propagators and Span Exporters.


ryanrupp:
Resolves https://github.com/open-telemetry/opentelemetry-java/issues/4056

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/SpiUtil.java
```diff
@@ -9,39 +9,100 @@
...
+          () -> {
+            try {
+              return getConfigurable.apply(provider, config);
+            } catch (Throwable t) {
+              Level level = requestedNames.contains(name) ? Level.WARNING : Level.FINE;
+              logger.log(
+                  level,
+                  "Error initializing " + spiClass.getSimpleName() + " with name " + name,
+                  t);
+              return null;
```
ryanrupp, line 72: If this fails to initialize you probably get a sort of confusing error in some areas i.e. you'll get the above log statement saying it couldn't initialize but then you'll get a subsequent error message saying something along the lines of the provider with <name> wasn't found, since this doesn't differentiate didn't exist and failed to load. Technically because this is now lazily initialized this could probably be changed to not swallow the exception although it's a behavior change.

ryanrupp: Looking at the change where this was added https://github.com/open-telemetry/opentelemetry-java/pull/3579 it looks like mainly this was due to the eagerness so I think I will move forward with removing the exception catching here and just allow it to propagate.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/NamedSpiManager.java
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.autoconfigure;
+
+import javax.annotation.Nullable;
+
+interface NamedSpiManager<T> {
```
anuraaga, line 10: Because this is a private implementation detail I don't think we get much from having this interface. Let's just have NamedSpiManager be the class you defined below, and if needed for convenience it can have an `.empty()` static factory that initializes it with an empty map, we don't need a dedicated implementation / interface to support the empty case.

anuraaga (commented):
Thanks a lot! Small point, but otherwise great change

jack-berg (approved):

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/NamedSpiManager.java
jkwatson: Agreed. This interface isn't adding much value to the implementation, IMO.

kubawach (approved):
Approve with assumption that @anuraaga comment regarding the new interface will be resolved :)

PR #4063, opened 2022-01-06
Add configuration to skip the shutdown hook registration

### LINKED ISSUE #4003: ShutdownHook in TracerProviderConfiguration
The `TracerProviderConfiguration` registers a ShutdownHook to close the `TracerProvider` (and shutdown the active span processor). In some environments, this may not be the most correct way to do it. Think for instance in a multi-app environment, each with its own `TracerProvider` and with different life cycles. The environments need to call `close` when the app terminates (or undeploys) and not on the JVM shutdown.

Provide a way to skip the registration of the shutdown hook. This also means that we need to provide a wat to access the `SdkTracerProvider` to call `.close` (or the `Closeable` instance).  Right now this does not seem to be possible if you use auto-configuration.

radcortez:
- Fixes #4003

### DIFF
--- a/sdk-extensions/autoconfigure/README.md
+++ b/sdk-extensions/autoconfigure/README.md
@@ -35,6 +35,15 @@ and not something we can control. If you require logging during shutdown hooks, 
 that might shut itself down in a shutdown hook, thus suppressing your log messages. See this [JDK bug](https://bugs.openjdk.java.net/browse/JDK-8161253)
 for more details.
 
+It is possible to skip the automatic registration of the shutdown hook by using the following configuration item:
+
+| System property                  | Environment variable             | Description                                                          |
+|----------------------------------|----------------------------------|----------------------------------------------------------------------|
+| otel.java.register.shutdown.hook | OTEL_JAVA_REGISTER_SHUTDOWN_HOOK | If `false` skips the shutdown hook registration. Default is `true`. |
+
+Skipping the registration of the shutdown hook may cause unexpected behaviour. This configuration is intended for SDK consumers that
+require control over the SDK lifecycle. In this case, alternatives must be provided by the SDK consumer to properly shut down the SDK.
+
 ## Exporters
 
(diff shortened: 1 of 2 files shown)


sdk-extensions/autoconfigure/README.md
```diff
@@ -35,6 +35,15 @@ and not something we can control. If you require logging during shutdown hooks,
 that might shut itself down in a shutdown hook, thus suppressing your log messages. See this [JDK bug](https://bugs.openjdk.java.net/browse/JDK-8161253)
 for more details.
 
+It is possible to skip the automatic registration of the shutdown hook by using the following configuration item:
+
+| System property                  | Environment variable             | Description                                                          |
+|----------------------------------|----------------------------------|----------------------------------------------------------------------|
+| otel.java.register.shutdown.hook | OTEL_JAVA_REGISTER_SHUTDOWN_HOOK | If `false` skips the shutdown hook registration. Default is `true`. |
```
jkwatson, line 42: If we're going to introduce non-officially-specified env vars, I think we should keep them in the `experimental` namespace, so if the spec does end up supporting this idea that we won't have to support two non-experimental options for the same thing.

anuraaga (commented):
@radcortez Did you need an env var for this? I think usually whether the shutdown hook is needed is known at coding time for each app. So was thinking we would add the option to the autoconfiguration builder but no env var.

radcortez:
> @radcortez Did you need an env var for this? I think usually whether the shutdown hook is needed is known at coding time for each app. So was thinking we would add the option to the autoconfiguration builder but no env var.

Sure. That was my thinking as well, but for some reason, I misunderstood that it was preferred as a configuration. My bad, I'll change it.

radcortez:
Moving this to https://github.com/open-telemetry/opentelemetry-java/pull/4004, since it requires the Customizer work for the SdkTracerProviderBuilder.

PR #4070, opened 2022-01-07
Collect coverage on Java 17 PR build.

anuraaga:
Already the case for main build probably accidentally drifted


jkwatson (approved):
coverage shouldn't vary by VM version, could it?

anuraaga:
Nah I figured we should generally just enable it on the latest LTS as a arbitrary principle. Having PR and main match though is beneficial for allowing the build cache to be shared.

PR #4072, opened 2022-01-07
Provide a way to register / skip the shutdown hook in AutoConfiguredOpenTelemetrySdkBuilder

### LINKED ISSUE #4003: ShutdownHook in TracerProviderConfiguration

radcortez:
- Fixes #4003

### DIFF
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
@@ -16,16 +16,20 @@ import io.opentelemetry.sdk.OpenTelemetrySdkBuilder;
 import io.opentelemetry.sdk.autoconfigure.spi.AutoConfigurationCustomizer;
 import io.opentelemetry.sdk.autoconfigure.spi.AutoConfigurationCustomizerProvider;
 import io.opentelemetry.sdk.autoconfigure.spi.ConfigProperties;
+import io.opentelemetry.sdk.common.CompletableResultCode;
 import io.opentelemetry.sdk.logs.SdkLogEmitterProvider;
 import io.opentelemetry.sdk.metrics.SdkMeterProvider;
 import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.sdk.trace.SdkTracerProvider;
 import io.opentelemetry.sdk.trace.export.SpanExporter;
 import io.opentelemetry.sdk.trace.samplers.Sampler;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.ServiceLoader;
+import java.util.concurrent.TimeUnit;
 import java.util.function.BiFunction;
 import java.util.function.Supplier;
 import java.util.logging.Level;
@@ -57,6 +61,8 @@ public final class AutoConfiguredOpenTelemetrySdkBuilder implements AutoConfigur
 
   private ClassLoader serviceClassLoader =
       AutoConfiguredOpenTelemetrySdkBuilder.class.getClassLoader();
+
+  private boolean registerShutdownHook = true;
 
   private boolean setResultAsGlobal = true;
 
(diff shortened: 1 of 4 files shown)


sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/AutoConfigurationCustomizer.java
```diff
@@ -66,4 +66,18 @@ AutoConfigurationCustomizer addSpanExporterCustomizer(
...
+   * Control the registration of a shutdown hook to shut down the SDK when appropriate.
+   *
+   * <p>Skipping the registration of the shutdown hook may cause unexpected behavior. This
+   * configuration is for SDK consumers that require control over the SDK lifecycle. In this case,
+   * alternatives must be provided by the SDK consumer to shut down the SDK.
+   *
+   * @param registerShutdownHook a boolean <code>true</code> will register the hook, otherwise
+   *     <code>false</code> will skip registration.
+   */
+  default AutoConfigurationCustomizer registerShutdownHook(boolean registerShutdownHook) {
```
jack-berg, line 80: Does this need to be defined in the interface? Is it insufficient as a method in `AutoConfiguredOpenTelemetrySdkBuilder` similar to `setResultAsGlobal(boolean setResultAsGlobal)`?

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
```diff
@@ -197,6 +216,10 @@ public AutoConfiguredOpenTelemetrySdk build() {
             spanExporterCustomizer,
             samplerCustomizer);
 
+    if (registerShutdownHook) {
+      Runtime.getRuntime().addShutdownHook(new Thread(tracerProvider::close));
```
jack-berg, line 220: `MeterProviderConfiguration` and `LogEmitterProviderConfiguration` also register shutdown hooks that should be conditional on `registerShutdownHook`.

sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/AutoConfigurationCustomizer.java
jack-berg, line 80: Is there anyway to unit test this?

radcortez: No, we can remove it from the interface. I've only added it here for consistency.

radcortez: There is no public API to query the registered shutdown hooks. One way would be to keep a reference to the `Thread` and then call `removeShutdownHook`.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
```diff
@@ -201,6 +219,14 @@ public AutoConfiguredOpenTelemetrySdk build() {
         LogEmitterProviderConfiguration.configureLogEmitterProvider(
             resource, config, meterProvider);
 
+    if (registerShutdownHook) {
+      Runtime.getRuntime().addShutdownHook(new Thread(tracerProvider::close));
```
anuraaga, line 223: Let's just use one thread, and call shutdown with `CompletableResultCode.ofAll().join` to reduce thread explosion on shutdown

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
```diff
@@ -201,6 +223,27 @@ public AutoConfiguredOpenTelemetrySdk build() {
...
+                    List<CompletableResultCode> shutdown = new ArrayList<>();
+                    shutdown.add(tracerProvider.shutdown());
+                    if (meterProvider instanceof SdkMeterProvider) {
+                      shutdown.add(((SdkMeterProvider) meterProvider).shutdown());
+                    }
+                    shutdown.add(logEmitterProvider.shutdown());
+                    CompletableResultCode.ofAll(shutdown).join(10, TimeUnit.SECONDS);
+                  }));
+
+      Runtime.getRuntime().addShutdownHook(new Thread(tracerProvider::close));
```
anuraaga, line 240: Looks like forgot to delete the old code with three threads

radcortez: Ops. Sorry. Fixed now.

anuraaga (approved):

jkwatson (approved):

jack-berg (approved):

PR #4073, opened 2022-01-09
Single logger instance for invalid API call

gcatanese:
Created a name logger (which is not enabled by default). I have added where I thought it is useful (possibly missing other spots - please suggest in case)

### DIFF
--- a/api/all/README.md
+++ b/api/all/README.md
@@ -19,5 +19,4 @@ to run a specific benchmark class.
 
 ---
 #### Logging API mis-use
-Define `logging.level.io.opentelemetry.ApiUsageLogging=TRACE` to enable TRACE logging for possible invalid or
-no-op API calls
+Define `logging.level.io.opentelemetry.ApiUsageLogging=TRACE` to enable TRACE logging for invalid API calls.
--- a/api/all/README.md
+++ b/api/all/README.md
@@ -19,5 +19,5 @@ to run a specific benchmark class.
 
 ---
 #### Logging API mis-use
-Define `logging.level.ApiUsageLogging=TRACE` to enable TRACE logging for possible invalid or
+Define `logging.level.io.opentelemetry.ApiUsageLogging=TRACE` to enable TRACE logging for possible invalid or
 no-op API calls
--- a/api/all/README.md
+++ b/api/all/README.md
@@ -16,3 +16,8 @@
 From the root of the repo run `./gradlew clean :api:jmh` to run all the benchmarks
 or run `./gradlew clean :api:jmh -PjmhIncludeSingleClass=<ClassNameHere>`
 to run a specific benchmark class.
+
+---
+#### Logging API mis-use
+Define `logging.level.ApiUsageLogging=TRACE` to enable TRACE logging for possible invalid or
+no-op API calls
--- a/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
+++ b/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
@@ -34,7 +34,6 @@ import javax.annotation.concurrent.ThreadSafe;
  */
 public final class GlobalOpenTelemetry {
 
-  public static final Logger API_USAGE_LOGGER = Logger.getLogger("ApiUsageLogging");
   private static final Logger logger = Logger.getLogger(GlobalOpenTelemetry.class.getName());
 
   private static final Object mutex = new Object();
--- a/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
+++ b/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
@@ -34,6 +34,7 @@ import javax.annotation.concurrent.ThreadSafe;
  */
 public final class GlobalOpenTelemetry {
 
+  public static final Logger API_USAGE_LOGGER = Logger.getLogger("ApiUsageLogging");
   private static final Logger logger = Logger.getLogger(GlobalOpenTelemetry.class.getName());
 
   private static final Object mutex = new Object();
--- a/api/all/src/main/java/io/opentelemetry/api/baggage/ImmutableEntryMetadata.java
+++ b/api/all/src/main/java/io/opentelemetry/api/baggage/ImmutableEntryMetadata.java
@@ -5,9 +5,8 @@
 
 package io.opentelemetry.api.baggage;
 
-import static io.opentelemetry.api.GlobalOpenTelemetry.API_USAGE_LOGGER;
-
 import com.google.auto.value.AutoValue;
+import io.opentelemetry.api.internal.ValidationUtil;
 import javax.annotation.concurrent.Immutable;
 
 @Immutable
@@ -26,7 +25,7 @@ abstract class ImmutableEntryMetadata implements BaggageEntryMetadata {
    */
   static ImmutableEntryMetadata create(String metadata) {
     if (metadata == null) {
-      API_USAGE_LOGGER.finest("ImmutableEntryMetadata.create: metadata is null");
+      ValidationUtil.log("metadata is null", new NullPointerException());
       return EMPTY;
     }
     return new AutoValue_ImmutableEntryMetadata(metadata);
--- a/api/all/src/main/java/io/opentelemetry/api/baggage/ImmutableEntryMetadata.java
+++ b/api/all/src/main/java/io/opentelemetry/api/baggage/ImmutableEntryMetadata.java
@@ -25,7 +25,7 @@ abstract class ImmutableEntryMetadata implements BaggageEntryMetadata {
    */
   static ImmutableEntryMetadata create(String metadata) {
     if (metadata == null) {
-      ValidationUtil.log("metadata is null", new NullPointerException());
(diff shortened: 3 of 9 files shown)


api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
```diff
@@ -34,6 +34,7 @@
  */
 public final class GlobalOpenTelemetry {
 
+  public static final Logger API_USAGE_LOGGER = Logger.getLogger("ApiUsageLogging");
```
anuraaga, line 37: Can you move this to a class inside the `internal` package, maybe called `ValidationUtil` as it's logging related to validation? And we'd like to not expose it in the user API

anuraaga: Let's also prefix the logger name with `io.opentelemetry`

api/all/src/main/java/io/opentelemetry/api/baggage/ImmutableEntryMetadata.java
```diff
@@ -24,6 +26,7 @@
    */
   static ImmutableEntryMetadata create(String metadata) {
     if (metadata == null) {
+      API_USAGE_LOGGER.finest("ImmutableEntryMetadata.create: metadata is null");
```
anuraaga, line 29: I think instead of exposing the `Logger` we can expose a `log` method that calls finest which we are calling everywhere.

anuraaga: Also instead of logging just a string with some location information, let's log a throwable to provide a complete stack trace 

`log(Level.FINEST, "metadata is null", new NullPointerException())`

api/all/src/main/java/io/opentelemetry/api/trace/Span.java
```diff
@@ -75,6 +78,7 @@ static Span getInvalid() {
    */
   static Span wrap(SpanContext spanContext) {
     if (spanContext == null || !spanContext.isValid()) {
+      API_USAGE_LOGGER.finest("Span.wrap: context is null or invalid");
```
anuraaga, line 81: Passing an invalid context is correct usage here, we need to split this if to only log for null

api/all/src/main/java/io/opentelemetry/api/trace/SpanId.java
```diff
@@ -95,6 +98,7 @@ public static String fromBytes(byte[] spanIdBytes) {
    */
   public static String fromLong(long id) {
     if (id == 0) {
+      API_USAGE_LOGGER.finest("SpanId.fromLong: id == 0");
```
anuraaga, line 101: This is ok usage so no log

api/all/src/main/java/io/opentelemetry/api/trace/TraceId.java
```diff
@@ -102,6 +105,8 @@ public static String fromBytes(byte[] traceIdBytes) {
    */
   public static String fromLongs(long traceIdLongHighPart, long traceIdLongLowPart) {
     if (traceIdLongHighPart == 0 && traceIdLongLowPart == 0) {
+      API_USAGE_LOGGER.finest(
+          "Trace.fromLongs: traceIdLongHighPart == 0 or traceIdLongLowPart == 0");
```
anuraaga, line 109: This is ok usage so no log

anuraaga (commented):
Thanks! There are probably other locations to use this but let's do it step by step

api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
gcatanese: Done. Only a small remark: enabling tracing on `io.opentelemetry` would include these logs too (maybe we could have considered just use TRACE on a standard per-class logger

api/all/README.md
```diff
@@ -16,3 +16,8 @@
 From the root of the repo run `./gradlew clean :api:jmh` to run all the benchmarks
 or run `./gradlew clean :api:jmh -PjmhIncludeSingleClass=<ClassNameHere>`
 to run a specific benchmark class.
+
+---
+#### Logging API mis-use
+Define `logging.level.ApiUsageLogging=TRACE` to enable TRACE logging for possible invalid or
```
jack-berg, line 22: Does this need to change now that the logger is named `io.opentelemetry.ApiUsageLogging` instead of `ApiUsageLogging`?

api/all/src/main/java/io/opentelemetry/api/baggage/ImmutableEntryMetadata.java
```diff
@@ -24,6 +25,7 @@
    */
   static ImmutableEntryMetadata create(String metadata) {
     if (metadata == null) {
+      ValidationUtil.log("metadata is null", new NullPointerException());
```
jack-berg, line 28: Can the throwable be generated inside the `log` method to DRY up the code?

api/all/README.md
gcatanese: Yes, sorry, I missed that

api/all/src/main/java/io/opentelemetry/api/baggage/ImmutableEntryMetadata.java
gcatanese: Yes, I see. Agree (I thought we would want more flexibility in terms of throwing different exceptions but it is probably not needed)

api/all/src/main/java/io/opentelemetry/api/trace/Span.java
```diff
@@ -74,7 +77,11 @@ static Span getInvalid() {
    * to propagate a valid {@link SpanContext} downstream.
    */
   static Span wrap(SpanContext spanContext) {
-    if (spanContext == null || !spanContext.isValid()) {
+    if (spanContext == null) {
+      ValidationUtil.log("context is null or invalid");
```
anuraaga, line 81:
```suggestion
      ValidationUtil.log("context is null");
```

api/all/src/main/java/io/opentelemetry/api/internal/ValidationUtil.java
```diff
@@ -0,0 +1,24 @@
...
+@Immutable
+public final class ValidationUtil {
+
+  private ValidationUtil() {}
+
+  private static final Logger API_USAGE_LOGGER =
+      Logger.getLogger("io.opentelemetry.ApiUsageLogging");
+
+  public static void log(String msg) {
+    API_USAGE_LOGGER.log(Level.FINEST, msg, new NullPointerException());
```
anuraaga, line 22:
```suggestion
    API_USAGE_LOGGER.log(Level.FINEST, msg, new AssertionError());
```

Since we use in some cases for non-null condition, let's use `AssertionError` instead

api/all/README.md
```diff
@@ -16,3 +16,8 @@
 From the root of the repo run `./gradlew clean :api:jmh` to run all the benchmarks
 or run `./gradlew clean :api:jmh -PjmhIncludeSingleClass=<ClassNameHere>`
 to run a specific benchmark class.
+
+---
+#### Logging API mis-use
+Define `logging.level.io.opentelemetry.ApiUsageLogging=TRACE` to enable TRACE logging for possible invalid or
+no-op API calls
```
jkwatson, line 23:
```suggestion
Define `logging.level.io.opentelemetry.ApiUsageLogging=TRACE` to enable TRACE logging for invalid API calls.
```

api/all/src/main/java/io/opentelemetry/api/baggage/ImmutableEntryMetadata.java
```diff
@@ -24,6 +25,7 @@
    */
   static ImmutableEntryMetadata create(String metadata) {
     if (metadata == null) {
+      ValidationUtil.log("metadata is null");
```
jkwatson, line 28: I wonder if we should provide some sort of contextual information with these log entries. If I saw this in my logs I might not know what would be the source of the logging, since all of them will be under the same logger name. I guess the AssertionError that gets logged would have the stack trace in it?

anuraaga: Yeah think the stack trace should be enough context unless there's anything else that comes to mind to add. But don't think we need to repeat method information.

gcatanese: I have just verified to confirm the stacktrace is logged

api/all/src/main/java/io/opentelemetry/api/trace/Span.java
```diff
@@ -55,6 +57,7 @@ static Span fromContext(Context context) {
   @Nullable
   static Span fromContextOrNull(Context context) {
     if (context == null) {
+      ValidationUtil.log(" context is null");
```
anuraaga, line 60:
```suggestion
      ValidationUtil.log("context is null");
```

anuraaga (approved):

jkwatson (approved):

PR #4085, opened 2022-01-12
Add log attribute limits

jack-berg:
Making log attributes compliant with the [common attribute limits](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attribute-limits).

I plan on integrating these into autoconfigure using the [generic attribute limit](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#attribute-limits) env vars.


sdk/logs/src/main/java/io/opentelemetry/sdk/logs/SdkLogBuilder.java
```diff
@@ -67,7 +70,11 @@ public LogBuilder setBody(String body) {
 
   @Override
   public LogBuilder setAttributes(Attributes attributes) {
```
jack-berg, line 72: Arguably we should expand the `LogBuilder` interface to include setters for single typed attributes for symmetry with `SpanBuilder`. E.g. `setAttribute(String key, String value)`, `setAttribute(String key, long value)`, etc.

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/SdkLogEmitterProviderBuilder.java
```diff
@@ -35,6 +37,23 @@ public SdkLogEmitterProviderBuilder setResource(Resource resource) {
...
+   * {@link LogEmitter#logBuilder()} is called.
+   *
+   * <p>The {@code logLimitsSupplier} must be thread-safe and return immediately (no remote calls,
+   * as contention free as possible).
+   *
+   * @param logLimitsSupplier the supplier that will be used to retrieve the {@link LogLimits} for
+   *     every {@link LogBuilder}.
+   * @return this
+   */
+  public SdkLogEmitterProviderBuilder setLogLimits(Supplier<LogLimits> logLimitsSupplier) {
```
jack-berg, line 51: I opted to only include the `Supplier<LogLimits>` version of the setter, and omitting the static `LogLimits` version. 

I can add the static version, but figured its pretty small cognitive load to use the supplier version.

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/SdkLogBuilder.java
```diff
@@ -17,10 +18,12 @@
 
   private final LogDataBuilder logDataBuilder;
   private final LogEmitterSharedState logEmitterSharedState;
+  private final LogLimits logLimits;
 
   SdkLogBuilder(LogEmitterSharedState logEmitterSharedState, LogDataBuilder logDataBuilder) {
     this.logEmitterSharedState = logEmitterSharedState;
     this.logDataBuilder = logDataBuilder;
+    this.logLimits = logEmitterSharedState.getLogLimits();
```
breedx-splk, line 26: Is it important to hold this reference? Does it limit the utility of the `Supplier` in the shared state then, or is the really the intention? (I probably don't understand the builder's lifecycle well enough).

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/SdkLogEmitterProvider.java
```diff
@@ -33,8 +34,12 @@ public static SdkLogEmitterProviderBuilder builder() {
...
   }
 
-  SdkLogEmitterProvider(Resource resource, List<LogProcessor> processors, Clock clock) {
-    this.sharedState = new LogEmitterSharedState(resource, processors, clock);
+  SdkLogEmitterProvider(
+      Resource resource,
+      Supplier<LogLimits> logLimitsSupplier,
+      List<LogProcessor> processors,
+      Clock clock) {
+    this.sharedState = new LogEmitterSharedState(resource, logLimitsSupplier, processors, clock);
```
breedx-splk, line 42: I think creating new instances in the constructor is a bit of a code smell (DI violation). If it's unreasonable to make the callers do this creation to pass in the real dependency (`LogEmitterSharedState`) then maybe a factory method is in order?

sdk/logs/src/test/java/io/opentelemetry/sdk/logs/SdkLogEmitterTest.java
```diff
@@ -38,4 +51,69 @@ void logBuilder() {
...
+  void logBuilder_maxAttributeLength() {
+    int maxLength = 25;
+    AtomicReference<LogData> seenLog = new AtomicReference<>();
+    SdkLogEmitterProvider logEmitterProvider =
+        SdkLogEmitterProvider.builder()
+            .addLogProcessor(seenLog::set)
+            .setLogLimits(() -> LogLimits.builder().setMaxAttributeValueLength(maxLength).build())
+            .build();
+    LogBuilder logBuilder = logEmitterProvider.get("test").logBuilder();
+    String strVal = IntStream.range(0, maxLength).mapToObj(i -> "a").collect(joining());
```
breedx-splk, line 65: FWIW, there's also `StringUtils.padLeft()` in the api.

sdk/logs/src/test/java/io/opentelemetry/sdk/logs/SdkLogEmitterTest.java
```diff
@@ -38,4 +51,69 @@ void logBuilder() {
...
+        .emit();
+
+    Attributes attributes = seenLog.get().getAttributes();
+
+    assertThat(attributes.get(stringKey("string"))).isEqualTo(strVal);
+    assertThat(attributes.get(booleanKey("boolean"))).isEqualTo(true);
+    assertThat(attributes.get(longKey("long"))).isEqualTo(1L);
+    assertThat(attributes.get(doubleKey("double"))).isEqualTo(1.0);
+    assertThat(attributes.get(stringArrayKey("stringArray")))
+        .isEqualTo(Arrays.asList(strVal, strVal));
```
breedx-splk, line 89: Great coverage!

breedx-splk (approved):
Looks good to me. Had a couple small questions/thoughts.

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/SdkLogBuilder.java
jack-berg: We need to have predictable semantics about when the log limits are resolved. Spans have the same concept, and `Supplier<SpanLimits>` is invoked when `Tracer.spanBuilder(String spanName)` is called. I think it makes sense for the log limits to resolved when the log builder is initialized, versus when attributes are first added.

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/SdkLogEmitterProvider.java
jack-berg: The only caller of this constructor is `SdkLogEmitterProviderBuilder#build()`. In this case requiring the caller to construct the `LogEmitterSharedState` shifts the builder from being a simple data carrier for configuration options to it knowing about the internals of the `SdkLogEmitterProvider`.

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/LogLimits.java
```diff
@@ -0,0 +1,70 @@
...
+   * instance.
+   *
+   * @return a {@link LogLimitsBuilder} initialized to the same property values as the current
+   *     instance.
+   */
+  public LogLimitsBuilder toBuilder() {
+    return new LogLimitsBuilder()
+        .setMaxNumberOfAttributes(getMaxNumberOfAttributes())
+        .setMaxAttributeValueLength(getMaxAttributeValueLength());
+  }
```
anuraaga, line 69: Looks like missing a test for this method

sdk/logs/src/test/java/io/opentelemetry/sdk/logs/SdkLogEmitterTest.java
```diff
@@ -38,4 +50,69 @@ void logBuilder() {
...
+                .put(stringArrayKey("stringArray"), Arrays.asList(strVal, tooLongStrVal))
+                .put(booleanArrayKey("booleanArray"), Arrays.asList(true, false))
+                .put(longArrayKey("longArray"), Arrays.asList(1L, 2L))
+                .put(doubleArrayKey("doubleArray"), Arrays.asList(1.0, 2.0))
+                .build())
+        .emit();
+
+    Attributes attributes = seenLog.get().getAttributes();
+
+    assertThat(attributes.get(stringKey("string"))).isEqualTo(strVal);
```
anuraaga, line 83: You might want to use `OpenTelemetryAssertions` for the attributes assert

sdk/logs/src/test/java/io/opentelemetry/sdk/logs/SdkLogEmitterTest.java
```diff
@@ -38,4 +50,69 @@ void logBuilder() {
...
+    AttributesBuilder attributesBuilder = Attributes.builder();
+    for (int i = 0; i < 2 * maxNumberOfAttrs; i++) {
+      attributesBuilder.put("key" + i, i);
+    }
+
+    logEmitterProvider.get("test").logBuilder().setAttributes(attributesBuilder.build()).emit();
+
+    Attributes attributes = seenLog.get().getAttributes();
+    assertThat(attributes.size()).isEqualTo(maxNumberOfAttrs);
+    // NOTE: cannot guarantee which attributes are retained, only that there are no more that the
```
anuraaga, line 115: Would put the comment above the code it's referring to, so just one line up

anuraaga (approved):

sdk/logs/src/main/java/io/opentelemetry/sdk/logs/LogLimits.java
jack-berg: Realized we are missing the whole test class for `LogLimits`! Added.

sdk/logs/src/test/java/io/opentelemetry/sdk/logs/SdkLogEmitterProviderBuilderTest.java
```diff
@@ -1,42 +0,0 @@
...
-
-import io.opentelemetry.sdk.common.Clock;
-import io.opentelemetry.sdk.logs.data.LogData;
-import io.opentelemetry.sdk.resources.Resource;
-import java.util.LinkedList;
-import java.util.List;
-import org.assertj.core.api.AssertionsForClassTypes;
-import org.junit.jupiter.api.Test;
-
-class SdkLogEmitterProviderBuilderTest {
```
jack-berg, line 20: Moved coverage to `SdkLogEmitterProviderTest` and `SdkLogEmitterTest` for more consistency:
- `SdkLogEmitterProviderTest` contains tests asserting the `SdkLogEmitterBuilder` behavior.
- `SdkLogEmitterTest` contains tests asserting that the configured clock is used to generate the epoch of the log (if the time is not explicitly set).

PR #4093, opened 2022-01-17
Rename RecordEventsReadableSpan to SdkSpan

anuraaga:
Before: Highly obtuse name
After: Consistent with every other (*) SDKFoo implementation of an API Foo.

Also note that the builder for this class is `SdkSpanBuilder`

(*) I didn't actually confirm but highly suspect this is true


jkwatson (approved):
Yes! This name has always been awkward. Thanks!

jack-berg:
👍 👍

PR #4097, opened 2022-01-19
Enable reflective setting of retry policy for OkHttpGrpcExporter

jack-berg:
Was updating some code to use the latest version of the agent, and to use the `OkHttpGrpcExporter` instead of the `DefaultGrpcExporter`, and realized that while `OkHttpGrpcExporter` supports setting a retry policy, its not actually possible to enable it programmatically or via environment variables. 😞 

This fixes that by adjusting the reflective access method to work with both `DefaultGrpcExporterBuilder` and `OkHttpGrpcExporterBuilder`.

### DIFF
--- a/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/grpc/DefaultGrpcExporterBuilder.java
+++ b/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/grpc/DefaultGrpcExporterBuilder.java
@@ -43,7 +43,7 @@ public final class DefaultGrpcExporterBuilder<T extends Marshaler>
   private boolean compressionEnabled = false;
   @Nullable private Metadata metadata;
   @Nullable private byte[] trustedCertificatesPem;
-  @Nullable private RetryPolicy retryPolicy;
+  @Nullable RetryPolicy retryPolicy;
   private MeterProvider meterProvider = MeterProvider.noop();
 
   /** Creates a new {@link DefaultGrpcExporterBuilder}. */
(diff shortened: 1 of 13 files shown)


exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/retry/RetryUtil.java
```diff
@@ -44,4 +47,29 @@ private RetryUtil() {}
...
   }
+
+  /**
+   * Reflectively access a {@link DefaultGrpcExporterBuilder} or {@link OkHttpGrpcExporterBuilder}
+   * instance in field called "delegate" of the instance, and set the {@link RetryPolicy}.
+   *
+   * @throws IllegalArgumentException if the instance does not contain a field called "delegate" of
+   *     type {@link DefaultGrpcExporterBuilder}
+   */
+  public static void setRetryPolicyOnDelegate(Object instance, RetryPolicy retryPolicy) {
```
jack-berg, line 58: This replaces `DefaultGrpcExporterBuilder#getDelegateBuilder`. Whereas the `#getDelegateBuilder` returned the delegate builder which allowed you to call `setRetryPolicy(...)`, this replacement is more to the point and just sets the retry policy on the delegate if it can.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/grpc/DefaultGrpcExporterBuilder.java
```diff
@@ -166,7 +166,11 @@ public DefaultGrpcExporterBuilder(
    *
    * @throws IllegalArgumentException if the instance does not contain a field called "delegate" of
    *     type {@link DefaultGrpcExporterBuilder}
+   * @deprecated Use {@link
+   *     io.opentelemetry.exporter.otlp.internal.retry.RetryUtil#setRetryPolicyOnDelegate(Object,
+   *     RetryPolicy)}
    */
+  @Deprecated
```
jkwatson, line 173: since this is `internal` can we just delete this?

anuraaga (approved):

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/otlp/internal/grpc/DefaultGrpcExporterBuilder.java
jack-berg: I would like to, yes 👍

PR #4112, opened 2022-01-24
BatchSpanProcessor metrics

### LINKED ISSUE #4109: BatchSpanProcessor created by TracerProviderConfiguration does not report metrics
**Describe the bug**
When using the `BatchSpanProcessor` created by `TracerProviderConfiguration.configureBatchSpanProcessor`, the metrics named `processedSpans` and `queueSize` are not reported  to a collector. This is because the BatchSpanProcessor is using the noop `MeterProvider` rather than the `MeterProvider` provided to `TracerProviderConfiguration.configureTracerProvider`

**Steps to reproduce**
I discovered this issue while observing the behavior of the OpenTelemetry Java Instrumentation project. But, minimal steps to reproduce would be to initialize the OpenTelemetry SDK using the autoconfigure API and create and export a span. This will create an SDK instance with a BatchSpanProcessor. The `SpanExporter` used by the SDK will have an `SdkMeterProvider` that exports metrics via OTLP, however the `BatchSpanProcessor` will have a `DefaultMeterProvider` instance that always creates `NoopMeterBuilder`, which do not actually result in metrics being created and reported 

I expected to see the  `processedSpans` and `queueSize` metrics in the debug logs of the OpenTelemetry collector that my SDK is exporting to

**What did you see instead?**
I only see other metrics created and exported by the OpenTelemetry SDK like `otlp.exporter.seen` and `otlp.exporter.exported`

Artifacts: io.opentelemetry:opentelemetry-sdk-extension-autoconfigure` with the `io.opentelemetry.exporter.otlp.trace.OtlpGrpcSpanExporter`
Version: 1.10.0, 1.10.1
How did you reference these artifacts? 
```kotlin
  implementation("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure")
```

**Environment**
Compiler: (e.g., "AdoptOpenJDK 11.0.10+9")
OS: (e.g., "OS X 12.1")

**Additional context**

These metrics were still being reported in OpenTelemetry SDK 1.9.1. I believe the change that resulted in this new behavior was [this line](https://github.com/open-telemetry/opentelemetry-java/pull/3901/files#diff-5cd39c6b59b999b012da4539f8e1fd4e1d404dbd813f7f982ae5f8423de35350R166) replacing the `GlobalMeterProvider.get()`, which I think was unintentional. The [BatchSpanProcessorBuilder](https://github.com/open-telemetry/opentelemetry-java/pull/3901/files#diff-86f374cd67a7b106ccb71712de8b5ea8f06199798c6272090e9191da15194c80R32) defaults to the noop `MeterProvider`, which makes sense. However, I believe we should add a call to `BatchSpanProcessorBuilder.setMeterProvider` in the `TracerProviderConfiguration.configureBatchSpanProcessor` method using the `MeterProvider` that is currently passed to `TracerProviderConfiguration.configureTracerProvider`. If this sounds like a change the team would like, I would be happy to make the pull request myself

ryandens:
closes #4109 

Configures the `BatchSpanProcessor` with the `MeterProvider` provided to the `TracerProviderConfiguration` in the autoconfigure extension.

I'm open to ideas for any additional test cases as well to help protect against this again in future refactors

### DIFF
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java
@@ -90,6 +90,7 @@ final class TracerProviderConfiguration {
   static BatchSpanProcessor configureBatchSpanProcessor(
       ConfigProperties config, SpanExporter exporter, MeterProvider meterProvider) {
     BatchSpanProcessorBuilder builder = BatchSpanProcessor.builder(exporter);
+    builder.setMeterProvider(meterProvider);
 
     Duration scheduleDelay = config.getDuration("otel.bsp.schedule.delay");
     if (scheduleDelay != null) {
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java
@@ -89,7 +89,8 @@ final class TracerProviderConfiguration {
   // VisibleForTesting
   static BatchSpanProcessor configureBatchSpanProcessor(
       ConfigProperties config, SpanExporter exporter, MeterProvider meterProvider) {
-    BatchSpanProcessorBuilder builder = BatchSpanProcessor.builder(exporter).setMeterProvider(meterProvider);
+    BatchSpanProcessorBuilder builder =
+        BatchSpanProcessor.builder(exporter).setMeterProvider(meterProvider);
 
     Duration scheduleDelay = config.getDuration("otel.bsp.schedule.delay");
     if (scheduleDelay != null) {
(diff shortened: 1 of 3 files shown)


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java
```diff
@@ -78,16 +80,17 @@ static SdkTracerProvider configureTracerProvider(
...
 
     return spanProcessors;
   }
 
   // VisibleForTesting
   static BatchSpanProcessor configureBatchSpanProcessor(
-      ConfigProperties config, SpanExporter exporter) {
+      ConfigProperties config, SpanExporter exporter, MeterProvider meterProvider) {
     BatchSpanProcessorBuilder builder = BatchSpanProcessor.builder(exporter);
+    builder.setMeterProvider(meterProvider);
```
jack-berg, line 93: #nit: slight preference for the following:
```suggestion
    BatchSpanProcessorBuilder builder = BatchSpanProcessor.builder(exporter).setMeterProvider(meterProvider);
```

jack-berg (approved):
Thanks!

anuraaga (approved):
Thanks!

PR #4116, opened 2022-01-25
Deprecate gRPC setChannel

anuraaga:
We'd generally like users to be using the okhttp transport and wouldn't even have this method in the API ideally to allow switching transports when we need. For now let's deprecate, and maybe hear back on some use cases which would probably be forwarded to the spec.

### DIFF
--- a/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerBuilder.java
+++ b/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerBuilder.java
@@ -98,7 +98,11 @@ public final class JaegerRemoteSamplerBuilder {
   /**
    * Sets the managed channel to use when communicating with the backend. Takes precedence over
    * {@link #setEndpoint(String)} if both are called.
+   *
+   * @deprecated Use {@link #setEndpoint(String)}. If you have a use case not satisfied by the
+   *     methods on this builder, please file an issue to let us know what it is.
    */
+  @Deprecated
   public JaegerRemoteSamplerBuilder setChannel(ManagedChannel channel) {
     requireNonNull(channel, "channel");
     delegate.setChannel(channel);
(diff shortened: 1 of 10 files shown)


sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerBuilder.java
```diff
@@ -98,7 +98,11 @@ public JaegerRemoteSamplerBuilder setInitialSampler(Sampler initialSampler) {
   /**
    * Sets the managed channel to use when communicating with the backend. Takes precedence over
    * {@link #setEndpoint(String)} if both are called.
+   *
+   * @deprecated Use {@link #setEndpoint(String)}. If you have a use case not satisfied by the
+   *     methods on this builder, please file an issue to let us know what it is.
    */
+  @Deprecated
   public JaegerRemoteSamplerBuilder setChannel(ManagedChannel channel) {
```
jack-berg, line 106: Should be able to delete this method since this code is more recent than the latest release.

jack-berg (approved):

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerBuilder.java
anuraaga: This method has been around for a while right? e.g.

https://github.com/open-telemetry/opentelemetry-java/blob/v1.9.0/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerBuilder.java#L58

jack-berg: Oops so it has. Was confusing it with something else.

PR #4120, opened 2022-01-25
Use noop meter provider if no readers registered

jack-berg:
I was working on adding support for metrics customizers in `AutoConfigurationCustomizer` and ran into an issue: We currently don't register the `SdkMeterProvider` with `OpenTelemetrySdkBuilder` unless an exporter is explicitly set. This presents issues when adding a meter provider customizer method because we no we won't be able to know whether or not an `SdkMeterProviderBuilder` was extended with a reader based on whether `otel.metrics.exporter` has been set. And actually, the current metrics SPI `ConfigurableMetricExporterProvider` suffers from this same issue: if I configure a reader via `ConfigurableMetricExporterProvider`, but don't change `otel.metrics.exporter`, then my SPI customization will be ignored. 

I don't actually thing the existing protective measure is doing anything at all. As it stands today, if `otel.metrics.exporter` is unset, `GlobalOpenTelemetry` will be set with a `MeterProvider` equal to `SdkMeterProvider.builder().build()`. 

I decided to fix the issue with autoconfigure and also optimize `SdkMeterProvider` to use the noop implementation if no readers are registers.

### DIFF
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
@@ -8,7 +8,6 @@ package io.opentelemetry.sdk.autoconfigure;
 import static java.util.Objects.requireNonNull;
 
 import io.opentelemetry.api.GlobalOpenTelemetry;
-import io.opentelemetry.api.metrics.MeterProvider;
 import io.opentelemetry.context.propagation.ContextPropagators;
 import io.opentelemetry.context.propagation.TextMapPropagator;
 import io.opentelemetry.sdk.OpenTelemetrySdk;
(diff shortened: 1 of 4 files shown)


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
```diff
@@ -269,12 +266,9 @@ public AutoConfiguredOpenTelemetrySdk build() {
         OpenTelemetrySdk.builder()
             .setTracerProvider(tracerProvider)
             .setLogEmitterProvider(logEmitterProvider)
+            .setMeterProvider(meterProvider)
             .setPropagators(propagators);
 
-    if (meterProvider instanceof SdkMeterProvider) {
```
anuraaga, line 274: 🙀

anuraaga (approved):

jsuereth (approved):

PR #4123, opened 2022-01-26
Make metric / log exporter closeable and improve behavior of some tests

anuraaga:
Enabling output streams on tests shows some places that are not closing gRPC channels, returning `null` from FooExporter (which fails when putting into the SimpleSpanProcessor's pending task map), and incorrect kotlin coroutines usage.

### DIFF
--- a/extensions/kotlin/src/test/kotlin/io/opentelemetry/extension/kotlin/KotlinCoroutinesTest.kt
+++ b/extensions/kotlin/src/test/kotlin/io/opentelemetry/extension/kotlin/KotlinCoroutinesTest.kt
@@ -79,8 +79,10 @@ class KotlinCoroutinesTest {
     val context1 = Context.root().with(ANIMAL, "cat")
     runBlocking(context1.asContextElement()) {
       assertThat(Context.current().get(ANIMAL)).isEqualTo("cat")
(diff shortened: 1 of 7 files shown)


extensions/kotlin/src/test/kotlin/io/opentelemetry/extension/kotlin/KotlinCoroutinesTest.kt
```diff
@@ -78,17 +78,19 @@ class KotlinCoroutinesTest {
   fun stressTest() {
     val context1 = Context.root().with(ANIMAL, "cat")
     runBlocking(context1.asContextElement()) {
+      assertThat(Context.current().get(ANIMAL)).isEqualTo("cat")
+      delay(10)
       assertThat(Context.current().get(ANIMAL)).isEqualTo("cat")
       for (i in 0 until 100) {
-        GlobalScope.launch {
```
anuraaga, line 83: Not sure why I used GlobalScope here but it was wrong, it starts a job in a new context meaning

1) The ANIMAL is null, not cat
2) the result is not collected into `runBlocking` (so assertion failure was getting ignored)

jsuereth (approved):

PR #4126, opened 2022-01-27
Enable loggers in all tests via slf4j.

anuraaga:
After enabling logging output on tests, we are able to catch unintended issues like leaked resources, but also do spam with expected log messages, especially when they have a huge junit stack trace. We can improve some.

At the same time our logging story is also confusing because we have logs on many tests and don't have them on many tests, all based on whether `slf4j-simple` happened to have been added, often to debug a failing test and left there so it organically absorbs more projects. Things can be more consistent if we enable logging everywhere.

This changes so that all test loggers are enabled via slf4j. The main reason to use slf4j is its ability to suppress based on system property, which makes it very easy to suppress log statements per test where they are known to be expected. 

Sending as draft, let me know if this pattern seems reasonable and then I will go through a full build and add more suppressions in this PR


sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/SdkMeterTest.java
```diff
@@ -20,7 +20,11 @@
 import io.opentelemetry.sdk.testing.exporter.InMemoryMetricReader;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
+import org.junitpioneer.jupiter.SetSystemProperty;
 
+@SetSystemProperty(
+    key = "org.slf4j.simpleLogger.log.io.opentelemetry.sdk.metrics.internal.state.MeterSharedState",
+    value = "off")
```
jkwatson, line 27: would it be better to change this to ERROR, rather than disabling altogether?

jack-berg:
Seems like a reasonable approach to me. 👍

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/SdkMeterTest.java
anuraaga: If we were to log an error there would be cases where we have a unit test for it, and we'd probably want off then in that case. Just always using off seems simpler and work well in enough cases to just stick with it

jkwatson: I think that makes sense. I'm slightly worried about something creeping in and spamming logs inadvertently, but lets go with it for now.

testing-internal/src/main/resources/simplelogger.properties
```diff
@@ -0,0 +1,4 @@
+org.slf4j.simpleLogger.defaultLogLevel = warn
```
jack-berg, line 1: Did you intend on switching the default log level from `info` => `warn`?

jack-berg (approved):

testing-internal/src/main/resources/simplelogger.properties
anuraaga: Yeah - the type of spam or resource issues we've seen would always log at warn, keeping the default at warn instead of info seems like a reasonable comproise vs having to add more `SuppressLogger`

PR #4128, opened 2022-01-27
Add DelegatingMetricData

jack-berg:
The metrics corollary to `DelegatingSpanData`. 

Opening as a draft because unlike `SpanData`, `MetricData` has publicly accessible static methods for creating instances, meaning that having a delegating metric data isn't as vital. Let me know if you think this is useful enough to add?

### DIFF
--- a/sdk/metrics/build.gradle.kts
+++ b/sdk/metrics/build.gradle.kts
@@ -25,6 +25,7 @@ dependencies {
   testImplementation(project(":sdk:metrics-testing"))
   testImplementation(project(":sdk:testing"))
   testImplementation("com.google.guava:guava")
+  testImplementation("com.google.guava:guava-testlib")
 
   jmh(project(":sdk:trace"))
   jmh(project(":sdk:metrics-testing"))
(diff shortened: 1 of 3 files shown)


jkwatson:
Personally, I'd leave it out until we need it. But, it also took a lot to convince me we needed the span version in the first place (I'm still not 100% convinced).

jack-berg:
Yeah after further consideration, I'm of the opinion that we don't need this. Reasons:
- `SpanData` doesn't have a concrete implementation that is accessible. `MetricData` does. 
- `SpanData` has many top level fields, which makes it inconvenient make a copy that changes just one or two fields. `MetricData` has 5 top level fields: `getResource`, `getInstrumentationLibraryInfo`, `getName`, `getDescription`, and `getUnit`. Its easy enough to copy these fields.
- A lot of the extra fields on metrics are nested under `Data<?> getData()`, which `DelegatingMetricData` does not address.

So overall `DelegatingMetricData` is not very high value IMO.

PR #4129, opened 2022-01-28
refactor: adjust tracer and meter provider customizer execution order

### LINKED ISSUE #4121: Adjust tracer provider customizer execution order and add metric/log provider customizer
I think the tracer provider customizer should be executed at the last step, so users can modify the auto configured fields. For example, if I want to set different resource to different provider. And I suggest expose metric provider customizer and log emitter provider customizer.

wallezhang:
Adjust tracer and meter provider customizer execution order to the last step. 
close #4121

### DIFF
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
@@ -15,6 +15,7 @@ import io.opentelemetry.sdk.OpenTelemetrySdkBuilder;
 import io.opentelemetry.sdk.autoconfigure.spi.AutoConfigurationCustomizer;
 import io.opentelemetry.sdk.autoconfigure.spi.AutoConfigurationCustomizerProvider;
 import io.opentelemetry.sdk.autoconfigure.spi.ConfigProperties;
+import io.opentelemetry.sdk.autoconfigure.spi.traces.SdkTracerProviderConfigurer;
 import io.opentelemetry.sdk.common.CompletableResultCode;
 import io.opentelemetry.sdk.logs.SdkLogEmitterProvider;
 import io.opentelemetry.sdk.metrics.SdkMeterProvider;
@@ -215,6 +216,7 @@ public final class AutoConfiguredOpenTelemetrySdkBuilder implements AutoConfigur
   public AutoConfiguredOpenTelemetrySdk build() {
     if (!customized) {
       customized = true;
+      mergeSdkTracerProviderConfigurer();
       for (AutoConfigurationCustomizerProvider customizer :
           ServiceLoader.load(AutoConfigurationCustomizerProvider.class, serviceClassLoader)) {
         customizer.customize(this);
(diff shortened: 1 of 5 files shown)


wallezhang:
@anuraaga Please take a review.

anuraaga (commented):
Thanks @wallezhang - can you add a unit test that would fail with the old behavior?

wallezhang:
> Thanks @wallezhang - can you add a unit test that would fail with the old behavior?

Now it seems that only the order of the tracer processor will be different, which actually won't affect the trace provider behavior.

anuraaga:
> Now it seems that only the order of the tracer processor will be different, which actually won't affect the trace provider behavior.

Can't you write a unit test where the customizer sets resource and validate that had an effect?

wallezhang:
> > Now it seems that only the order of the tracer processor will be different, which actually won't affect the trace provider behavior.
> 
> Can't you write a unit test where the customizer sets resource and validate that had an effect?

Yes, I can write a unit test to validate it has effect, but can't make the old behavior fail. For now only the order of processor is different.

anuraaga:
@wallezhang I think you can do something like this pseudocode which should fail with the old behavior

```
.addResourceCustomizer(resource -> resource.put("animal", "cat"))
// TracerProviderCustomizer gets priority since it's the most general
.addTracerProviderCustomizer(tracerProvider -> tracerProvider.setResource(Resource.builder().put("animal", "bear")))
```

And verify the value in the resource

wallezhang:
@anuraaga oh,I see what you mean. I think `tracerProviderCustomizer` in `AutoConfiguredOpenTelemetrySdkBuilder` class is useless since there is a `SdkTracerProviderConfigurer` interface in `:autoconfigure-spi` module. User can implement the interface to customize sdk trace provider builder.

wallezhang:
So I modified the `SdkTracerProviderConfigurer` SPI interface execution order.

jack-berg:
`AutoConfigurationCustomizer#addTracerProviderCustomizer(..)` and `SdkTracerProviderConfigurer ` perform the same function; however, `AutoConfigurationCustomizer#addTracerProviderCustomizer(..)` is meant to is meant to supersede  `SdkTracerProviderConfigurer`. 

As it stands in the PR, customizations added via `AutoConfigurationCustomizer#addTracerProviderCustomizer(..)` happen before autoconfiguration, while customizations via `SdkTracerProviderConfigurer` happen after autoconfiguration. This inconsistency is strange.

I suggest we change the flow so that `SdkTracerProviderConfigurer` implementations manifest as calls to `AutoConfigurationCustomizer#addTracerProviderCustomizer(..)`. I've demonstrated what this might look [here](https://github.com/jack-berg/opentelemetry-java/commit/d57a29359e8cc60eaba2836bec6fb1c5dfcbff31).

wallezhang:
> 

Yes, this flow looks like more reasonable. Thanks for your advice. I'll modify it.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java
```diff
@@ -280,6 +282,17 @@ public AutoConfiguredOpenTelemetrySdk build() {
     return AutoConfiguredOpenTelemetrySdk.create(openTelemetrySdk, resource, config);
   }
 
+  private void mergeSdkTracerProviderConfigurer() {
```
anuraaga, line 285: Thanks for this cleanup!

anuraaga (approved):

jack-berg (approved):

PR #4130, opened 2022-01-28
Add a testing-internal module and functionality to SuppressLogger for…

anuraaga:
… a test.

In #4126 I realized that SimpleLogger doesn't support mutating logLevel - I had just had good luck with SetSystemProperty initially that the logger got initialized after the property mutation. I'll still rework that PR to focus on enabling slf4j logging throughout for consistency but for suppression sticking to JUL suppression because JUL does support level mutation.

I also went ahead and moved mockito-extensions into this module as it was a reason I've wanted a testing-internal module for a while now.


exporters/logging/src/test/java/io/opentelemetry/exporter/logging/LoggingSpanExporterTest.java
```diff
@@ -38,7 +39,7 @@
 import org.junit.jupiter.api.extension.RegisterExtension;
 import org.slf4j.event.Level;
 
-/** Tests for the {@link LoggingSpanExporter}. */
+@SuppressLogger(LoggingSpanExporter.class)
```
jack-berg, line 42: I'm a bit confused on what is going on here. Whether this annotation is present or not I don't see log output for this when I run `./gradlew :exporters:logging:test`.

I thought it might be because the default log level is set to `warn`, but even after reducing it and removing the `@SurpressLogger` annotation, I don't see the logs produced by `LoggingSpanExporter`.

anuraaga: Thanks for looking into this so deeply! I've never used a logging.properties file before and can't believe that I have to copy in the default values (e.g. ConsoleHandler) when using one... 🙀

buildscripts/test-logging.properties
```diff
@@ -0,0 +1,7 @@
+handlers= java.util.logging.ConsoleHandler
+.level= INFO
```
jkwatson, line 2: formatting here is slightly different than the rest of the file. missing a space before the `=` sign.

dependencyManagement/build.gradle.kts
```diff
@@ -65,7 +65,7 @@ val DEPENDENCY_SETS = listOf(
     "org.mockito",
     "4.2.0",
     listOf("mockito-core", "mockito-junit-jupiter")
-  )
+  ),
```
jkwatson, line 68: ?

jkwatson (approved):
Just one small ask for some additional internal documentation

jack-berg (approved):

PR #4132, opened 2022-01-28
Retry on connection timeouts

jack-berg:
Some New Relic folks using the retry feature let me know that `RetryInterceptor` doesn't retry on transient network errors like connection timeout. I took a look and confirmed that this is the case. 

Anytime we call `chain.proceed(chain.request())` without wrapping in a try / catch, connection errors can occur which bubble up without retry. This PR wraps treats all `IOExceptions` that occur during calls to `chain.proceed(chain.request())` as retryable. 

Note that there a variety of settings on `OkHttpClient.Builder` that can impact the conditions that cause `IOException` to be throw:
- [callTimeout](https://square.github.io/okhttp/3.x/okhttp/okhttp3/OkHttpClient.Builder.html#callTimeout-java.time.Duration-) - we set this today to 10s, and allow users to configure it. The total amount of time allotted to the entire call, including resolving DNS, connecting, writing, reading, and any redirects and retries.
- [connectTimeout](https://square.github.io/okhttp/3.x/okhttp/okhttp3/OkHttpClient.Builder.html#connectTimeout-java.time.Duration-) - this defaults to 10 seconds. This controls the time allotted to connect to a TCP socket to the host. With this PR, a violation of this would trigger a retry.
- [readTimeout](https://square.github.io/okhttp/3.x/okhttp/okhttp3/OkHttpClient.Builder.html#readTimeout-java.time.Duration-) - this defaults to 10 seconds. This controls the time allocated to the TCP socket connection and individual read IO operations. With this PR, a violation of this would trigger a retry.
- [writeTimeout](https://square.github.io/okhttp/3.x/okhttp/okhttp3/OkHttpClient.Builder.html#writeTimeout-java.time.Duration-) - this defaults to 10 seconds. This controls the time allocated to individual write IO operations. With this PR, a violation of this would trigger a retry.

**Update** - I've adjusted this PR to limit the scope to retrying when `IOException` is instance of `SocketTimeoutException`. Its hard to know the conditions which produce `IOException`, so I think better to narrow retrying to only the things we know should definitely be retried.

### DIFF
--- a/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/retry/RetryInterceptor.java
+++ b/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/retry/RetryInterceptor.java
@@ -5,10 +5,13 @@
 
 package io.opentelemetry.exporter.internal.retry;
 
+import io.opentelemetry.sdk.internal.ThrottlingLogger;
 import java.io.IOException;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import okhttp3.Interceptor;
 import okhttp3.Response;
 
@@ -20,6 +23,9 @@ import okhttp3.Response;
  */
 public final class RetryInterceptor implements Interceptor {
 
+  private static final Logger logger = Logger.getLogger(RetryInterceptor.class.getName());
+
+  private final ThrottlingLogger throttlingLogger = new ThrottlingLogger(logger);
   private final RetryPolicy retryPolicy;
   private final Function<Response, Boolean> isRetryable;
   private final Sleeper sleeper;
@@ -48,34 +54,52 @@ public final class RetryInterceptor implements Interceptor {
 
   @Override
   public Response intercept(Chain chain) throws IOException {
-    Response response = chain.proceed(chain.request());
-    if (!Boolean.TRUE.equals(isRetryable.apply(response))) {
+    Response response = null;
+    IOException exception = null;
+    int attempt = 0;
+    long nextBackoffNanos = retryPolicy.getInitialBackoff().toNanos();
+    do {
+      if (attempt > 0) {
+        // Compute and sleep for backoff
+        // https://github.com/grpc/proposal/blob/master/A6-client-retries.md#exponential-backoff
+        long upperBoundNanos = Math.min(nextBackoffNanos, retryPolicy.getMaxBackoff().toNanos());
+        long backoffNanos = randomLong.get(upperBoundNanos);
+        nextBackoffNanos = (long) (nextBackoffNanos * retryPolicy.getBackoffMultiplier());
+        try {
+          sleeper.sleep(backoffNanos);
+        } catch (InterruptedException e) {
+          Thread.currentThread().interrupt();
+          break; // Break out and return response or throw
+        }
+        // Close response from previous attempt
+        if (response != null) {
+          response.close();
+        }
+      }
+
+      attempt++;
+      try {
+        response = chain.proceed(chain.request());
+      } catch (IOException e) {
+        throttlingLogger.log(
+            Level.FINE,
+            "Error executing "
+                + chain.request().method()
+                + " "
+                + chain.request().url()
+                + ": "
+                + e.getMessage());
+        exception = e;
+      }
+      if (response != null && !Boolean.TRUE.equals(isRetryable.apply(response))) {
+        return response;
+      }
+    } while (attempt < retryPolicy.getMaxAttempts());
+
+    if (response != null) {
       return response;
     }
-    long nextBackoffNanos = retryPolicy.getInitialBackoff().toNanos();
-    // Made an attempt already, starting with the 2nd.
-    for (int attempt = 2; attempt <= retryPolicy.getMaxAttempts(); attempt++) {
-      // https://github.com/grpc/proposal/blob/master/A6-client-retries.md#exponential-backoff
-      // Careful - the document says "retry attempt" but our maxAttempts includes the original
-      // request. The numbers are tricky but the unit test should be more clear.
-      long upperBoundNanos = Math.min(nextBackoffNanos, retryPolicy.getMaxBackoff().toNanos());
-      long backoffNanos = randomLong.get(upperBoundNanos);
-
-      try {
-        sleeper.sleep(backoffNanos);
-      } catch (InterruptedException e) {
-        Thread.currentThread().interrupt();
-        return response;
-      }
-      // Close response from previous attempt and try again.
-      response.close();
-      response = chain.proceed(chain.request());
-      if (!Boolean.TRUE.equals(isRetryable.apply(response))) {
-        return response;
-      }
-      nextBackoffNanos = (long) (nextBackoffNanos * retryPolicy.getBackoffMultiplier());
-    }
-    return response;
+    throw exception;
   }
 
   // Visible for testing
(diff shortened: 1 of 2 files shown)


exporters/otlp/common/src/test/java/io/opentelemetry/exporter/internal/retry/RetryInterceptorTest.java
```diff
@@ -118,6 +127,22 @@ void interrupted() throws Exception {
...
+  void connectTimeout() throws Exception {
+    when(random.get(anyLong())).thenReturn(1L);
+    doNothing().when(sleeper).sleep(anyLong());
+
+    // Connecting to a non-routable IP address to trigger connection timeout
+    assertThatThrownBy(
+            () ->
+                client.newCall(new Request.Builder().url("http://10.255.255.1").build()).execute())
+        .isInstanceOf(SocketTimeoutException.class)
+        .hasMessage("Connect timed out");
```
jkwatson, line 140: test failure seems to suggest this should be a lower-cased "connect" in this message.

jack-berg: Weird.. the message case seems to vary by JVM version 🤔

jkwatson: oh, that's extra fun!  :)

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/retry/RetryInterceptor.java
```diff
@@ -38,44 +47,72 @@ public RetryInterceptor(RetryPolicy retryPolicy, Function<Response, Boolean> isR
...
 
+      attempt++;
       try {
-        sleeper.sleep(backoffNanos);
-      } catch (InterruptedException e) {
-        Thread.currentThread().interrupt();
-        return response;
+        response = chain.proceed(chain.request());
+      } catch (IOException e) {
+        throttlingLogger.log(
```
anuraaga, line 89: I'm pretty sure callers will log this as well so we shouldn't log here

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/retry/RetryInterceptor.java
```diff
@@ -38,44 +47,72 @@ public RetryInterceptor(RetryPolicy retryPolicy, Function<Response, Boolean> isR
...
-      // Close response from previous attempt and try again.
-      response.close();
-      response = chain.proceed(chain.request());
-      if (!Boolean.TRUE.equals(isRetryable.apply(response))) {
+      if (response != null && !Boolean.TRUE.equals(isRetryable.apply(response))) {
         return response;
       }
-      nextBackoffNanos = (long) (nextBackoffNanos * retryPolicy.getBackoffMultiplier());
+      if (exception != null && !Boolean.TRUE.equals(isRetryableException.apply(exception))) {
+        throw exception;
```
anuraaga, line 103: Are we missing a test case?

exporters/otlp/common/src/test/java/io/opentelemetry/exporter/internal/retry/RetryInterceptorTest.java
```diff
@@ -50,9 +68,14 @@ void setUp() {
                 .setMaxAttempts(5)
                 .build(),
             r -> !r.isSuccessful(),
+            isRetryableException,
             sleeper,
             random);
-    client = new OkHttpClient.Builder().addInterceptor(retrier).build();
+    client =
+        new OkHttpClient.Builder()
+            .connectTimeout(Duration.ofSeconds(1))
```
anuraaga, line 76: Let's set this much lower so the test doesn't take 1s

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/retry/RetryInterceptor.java
```diff
@@ -38,44 +47,72 @@ public RetryInterceptor(RetryPolicy retryPolicy, Function<Response, Boolean> isR
...
+    if (response != null) {
+      return response;
     }
-    return response;
+    throw exception;
+  }
+
+  // Visible for testing
+  static boolean isRetryableException(IOException e) {
+    return e instanceof SocketTimeoutException;
```
anuraaga, line 115: I think we're not supposed to retry on readTimeout because all the data has already been sent and may have been processed by the server. Well it's my interpretation of the spirit of the spec. Do you know if we can determine what type of timeout it is from some detail on the exception? `instanceof` seems like it will hold for connect, read, and write, though not call

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/retry/RetryInterceptor.java
jack-berg: Thought was that users may be interested in seeing log information describing why retrys occur. But if we wanted to log that information, we'd want to do it for more than just the case where `IOException` is thrown. Will leave it for possible future enhancement.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/retry/RetryInterceptor.java
jack-berg: My read of the spec is that the read / write timeout behavior is unspecified, but that retrying is expected and acceptable. The relevant bit is [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#duplicate-data). When we write a request and a read timeout occurs waiting for a response, we don't know whether the server successfully processed the bytes or not. Better to guarantee delivery as the risk of duplicate data. 

If we wanted to have behavior that differs for read / write / connect timeout, its possible to do so by inspecting the `SocketTimeoutException` error message:
- For connect timeouts, error message is "Connect timed out"
- For read timeouts, error message is "Read timed out"
- For write timeouts, error message is "timeout"

Not super confident that the messages are reliable though because I've seen at least the case of the message vary by JVM. I.e. in one it was "Connect timed out", in another it was "connect timed out".

anuraaga: I guess it's OK - the reason I still have some concern as readTimeout specifically may not be a transient error. If the server always takes too long to return its response for the telemetry we send, we will always retry until max attempts all the time. This generally indicates either the timeout is too low or a bigger problem in general, so this is probably still OK but something to keep in mind. It wouldn't be OK if it caused a queue to grow unbounded somewhere, but I don't think we have that behavior with BSP, would be good if you can doublecheck.

anuraaga (approved):

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/retry/RetryInterceptor.java
jack-berg: > It wouldn't be OK if it caused a queue to grow unbounded somewhere, but I don't think we have that behavior with BSP, would be good if you can doublecheck that retries will not amplify a bad effect on the app itself.

Yeah the BSP will block for a configurable amount of time waiting for the export to complete, and the export itself has a configurable timeout that it can spend waiting for the export, which consists at least one attempt and possibly more if retry is enabled and the timeout allows it. If the exporter timeout is longer than the BSP timeout, then an export may take longer than the BSP is willing to wait, and the BSP may start another export before the first is finished. In this scenario, its not clear to me what happens when one of the grpc transports (i.e. netty, netty-shadd, or okhttp) is used. But when the okhttp client is used, we use `OkHttpUtil.newDispatcher`, which uses a thread pool with a max size of `Integer.MAX_VALUE`. 

So if many exports end up in flight simultaneously, there could be congestion that will slow things down but its not like the exports are backing up in a queue. However, this does underscore that its is important to pay attention to the settings for `OTEL_BSP_EXPORT_TIMEOUT` and `OTEL_EXPORTER_OTLP_TIMEOUT` together. I wouldn't advise having a `OTEL_EXPORTER_OTLP_TIMEOUT` that is greater than `OTEL_BSP_EXPORT_TIMEOUT`. 

But I don't feel strongly about the read timeout stuff. I would be ok checking that the exception has a message contains "Connect timed out" as criteria to retry. Can always make the check more broad in the future.

PR #4140, opened 2022-02-01
Javadoc code snippets use methods that don't exist

laurit:
For example `OpenTelemetry.get()`


anuraaga (commented):
Thanks @laurit - can you change to `openTelemetry`, a variable of type `OpenTelemetry`? That is the standard pattern we encourage when using the SDK (and in practice agent too since it's recommended to be registered as a bean type thing in one place, not accessed everywhere).

anuraaga (approved):

jack-berg (approved):

jsuereth (approved):

kubawach (approved):
Nice catch.

jkwatson (approved):

PR #4143, opened 2022-02-03
Allow multiple async callbacks, allow callbacks to be removed

jack-berg:
There has been discussion in the java SIG and at the spec level about allowing asynchronous instruments to have multiple callbacks. This PR demonstrates how that might function. One of the issues that comes up with supporting multiple callbacks is the need to remove or unregister them at a later time. This PR adds a `remove()` method to the `Observable*` instrument interfaces. 

Related java issues: #4013, #3549

Related spec issues: [#2249](https://github.com/open-telemetry/opentelemetry-specification/issues/2249), [#2232](https://github.com/open-telemetry/opentelemetry-specification/issues/2232)


api/all/src/main/java/io/opentelemetry/api/metrics/ObservableDoubleCounter.java
```diff
@@ -14,4 +14,4 @@
  * <p>This interface currently has no methods but may be extended in the future with functionality
  * such as canceling the observable.
  */
-public interface ObservableDoubleCounter {}
+public interface ObservableDoubleCounter extends ObservableInstrument {}
```
bogdandrutu, line 17: Why not using something more java like "AutoClosable"?

anuraaga: Yup let's use AutoCloseable

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
```diff
@@ -165,17 +147,59 @@ public MetricData collectAndReset(
...
+        throttlingLogger.log(
+            Level.WARNING,
+            "Attempting to remove callback for "
+                + getMetricDescriptor().getName()
+                + " after storage has been unregistered.");
+        return;
+      }
+      this.callbacks.remove(callback);
+      if (callbacks.size() == 0) {
+        metricStorageRegistry.unregister(this);
```
anuraaga, line 197: It doesn't seem like removing a callback should cause total unregistration. From my reading of the code, it looks like this should but wouldn't work

```
addCallback
removeCallback
addCallback
```

Second `addCallback` would be `unregistered`

If the storage gets recreated at a higher level then maybe it's ok but I'd expect symmetry either at this level or a higher level, currently add and remove aren't symmetric and makes me wonder if there is possible thread safety issues at that higher level if this is true.

anuraaga: Also I wonder if unregistration causes the metric to stop being reported (not sure so just confirming) - if that is true, then we should avoid it for another reason, we'd want async and sync behavior to be consistent. There is currently no way for a sync metric to stop being reported, so we would want to solve both at the same time rather than just one.

jack-berg: > If the storage gets recreated at a higher level then maybe it's ok 

That's what happening - the second time you call `addCallback` a new `AynchronousMetricStorage` is registered with the `MetricStorageRegistry` because none currently exists at that point. Its likely an unnecessary optimization, since it only frees up resources of asynchronous metrics that have no more callbacks registered at all, which is probably unusual. 

> Also I wonder if unregistration causes the metric to stop being reported (not sure so just confirming) 

Asynchronous metrics are different from synchronous metrics in that we delegate the management of the metrics stream to the callback. So today (i.e. without this code) if a callback stops reporting a metric stream, it will stop being exported to cumulative exporters. In contrast, synchronous instruments continue to report values to exporters even when there is no data reported to the metric stream during the collection interval. This PR doesn't address that behavior. 

Here's a little demo of what I mean:
```
    InMemoryMetricReader cumulativeReader = InMemoryMetricReader.create();
    Meter meter = SdkMeterProvider.builder()
        .registerMetricReader(cumulativeReader)
        .setMinimumCollectionInterval(Duration.ofSeconds(0))
        .build()
        .get("test-meter");

    AtomicLong counter = new AtomicLong(2);
    meter.counterBuilder("foo").buildWithCallback(measurement -> {
      long count = counter.decrementAndGet();
      if (count >= 0) {
        measurement.record(10);
      }
    });

    for (int i = 0; i < 5; i++) {
      System.out.println(cumulativeReader.collectAllMetrics());
    }
```
That code yields:
```
[...data...]
[...data...]
[]
[]
[]
```

api/all/src/main/java/io/opentelemetry/api/metrics/ObservableDoubleCounter.java
jack-berg: Agree with using `AutoCloseable`, but just want to mention that its a bit difficult to imagine a scenario where you'd be able to take advantage of try-with-resources with one of these async instruments. 

Still good to use `AutoCloseable` though because it's the idiomatic way to indicate that the class should be cleaned up after use.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
```diff
@@ -37,102 +40,62 @@
...
+  private final MetricDescriptor metricDescriptor;
   private final TemporalMetricStorage<T> storage;
-  private final Runnable metricUpdater;
+  private final AsyncAccumulator<T> accumulator;
+  private final O measurement;
 
-  /** Constructs asynchronous metric storage which stores nothing. */
-  public static MetricStorage empty() {
-    return EmptyMetricStorage.INSTANCE;
+  @GuardedBy("callbackLock")
```
anuraaga, line 54: I think `CopyOnWriteArrayList` is commonly used to lock / copy on write instead of our current lock / copy on read. Though unfortunately to atomically check `isEmpty` and `forEach`, we may need to avoid that class and implement it ourselves, with a `volatile Consumer[] callbacks` and locks in add/remove callback (almost entirely to avoid `toArray`'s copy)

jack-berg: Good point about `CopyOnWriteArrayList`.

I think we can get around having to do an atomic check of `isEmpty` and `forEach` with a bit of a trick. We can iterate through the callbacks and determine if callbacks was empty as a side effect of iteration:
```
  boolean hasCallbacks = false;
  for (Consumer<O> callback : callbacks) {
    hasCallbacks = true;
    callback.accept(measurement);
  }
  if (!hasCallbacks) {
    return EmptyMetricData.getInstance();
  }
```

jack-berg:
Now that [#2317](https://github.com/open-telemetry/opentelemetry-specification/pull/2317) has been merged in the spec, we should be unblocked to merge this. Before we do, a question: Is it clear / intuitive that when you call `Observable{Type}{Instrument}#close()` that the callback associated with the `#buildWithCallback()` is the one removed? Put another way, is it obvious that `#close()` removes one callback and not all the callbacks associated with the instrument?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkObservableInstrument.java
```diff
@@ -0,0 +1,54 @@
...
+  SdkObservableInstrument(
+      String instrumentName, List<AsynchronousMetricStorage<?, O>> storages, Consumer<O> callback) {
+    this.instrumentName = instrumentName;
+    this.storages = storages;
+    this.callback = callback;
+  }
+
+  @Override
+  public void close() {
+    if (removed) {
```
anuraaga, line 46: We should probably use an `AtomicBoolean` for this

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MeterSharedState.java
```diff
@@ -98,39 +99,49 @@ public final WriteableMetricStorage registerSynchronousMetricStorage(
...
+          Consumer<ObservableLongMeasurement> callback) {
+
+    List<AsynchronousMetricStorage<?, ObservableLongMeasurement>> storages =
+        meterProviderSharedState
+            .getViewRegistry()
+            .findViews(instrument, getInstrumentationLibraryInfo())
+            .stream()
+            .map(view -> AsynchronousMetricStorage.createLongAsyncStorage(view, instrument))
+            .filter(m -> !m.isEmpty())
+            .map(this::register)
```
anuraaga, line 115: Not a huge deal but `map` having side effects can lead to some potential confusion

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorageTest.java
```diff
@@ -54,7 +58,7 @@
 
   @BeforeEach
   void setup() {
-    spyAttributesProcessor = Mockito.spy(AttributesProcessor.noop());
+    spyAttributesProcessor = spy(AttributesProcessor.noop());
```
anuraaga, line 61: Since you happened to update this line (and maybe wish now you hadn't :-) ), consider migrating this to `@Spy`, generally prefer not to mix annotations and non-annotations in the same class without strong reason

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MeterSharedState.java
jack-berg: I agree. I think the alternative is worth it despite the additional code.

jack-berg:
@jkwatson are you comfortable with merging this?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkObservableInstrument.java
```diff
@@ -0,0 +1,54 @@
...
+  SdkObservableInstrument(
+      String instrumentName, List<AsynchronousMetricStorage<?, O>> storages, Consumer<O> callback) {
+    this.instrumentName = instrumentName;
+    this.storages = storages;
+    this.callback = callback;
+  }
+
+  @Override
+  public void close() {
+    if (removed.compareAndSet(false, true)) {
```
anuraaga, line 47: Nit, I think

```
if (!compareAndSet) {
 LOG
 return
}
```

is a bit of a better pattern so that logic can be increased at the unindented level, the false case is constant 2 LoC no matter what happens.

jkwatson (approved):

PR #4150, opened 2022-02-03
Rename instrument type enumeration values

jack-berg:
Rename InstrumentType enum values for consistency with API:
- `OBSERVABLE_SUM` => `OBSERVABLE_COUNTER`
- `OBSERVABLE_UP_DOWN_SUM` => `OBSERVABLE_UP_DOWN_COUNTER` 

InstrumentType is public in the metrics SDK so consistency with the names of the classes in the API matters:
- `ObservableDoubleCounter`
- `ObservableLongCounter`
- `ObservableDoubleUpDownCounter`
- `ObservableLongUpDownCounter`


anuraaga (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/common/InstrumentType.java
```diff
@@ -10,7 +10,7 @@
   COUNTER,
   UP_DOWN_COUNTER,
   HISTOGRAM,
-  OBSERVABLE_SUM,
```
jkwatson, line 13: as this is a public API, used for views, I think we should probably go through a deprecation cycle for these values.

jkwatson: interesting that this doesn't show up in the api diffs. 🤔

jsuereth: Does the metrics SDK participate in API diffs? I'm not sure if it does.

+1 to deprecation cycle.

jsuereth (approved):
Agree with John's comment on deprecating.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/common/InstrumentType.java
jkwatson: oh right..it doesn't because it's not stable yet. 🤦🏽

PR #4154, opened 2022-02-07
Add context wrapping methods for arguments found in CompletableFuture…

anuraaga:
… APIs.

Lately I get questions about propagating context when using `CompletableFuture` and have to give the unsatisfying answer that we don't currently have helpers for most of its methods. Given it's a common Java 8 API, we probably should.

Too bad we can't use `wrap` for all of them since it causes lambdas to resolve ambiguously.

### DIFF
--- a/context/src/main/java/io/opentelemetry/context/Context.java
+++ b/context/src/main/java/io/opentelemetry/context/Context.java
@@ -27,6 +27,11 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Supplier;
 import javax.annotation.Nullable;
 
 /**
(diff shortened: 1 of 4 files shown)


jkwatson:
In the CompletableFuture, are these functions executed on the submitting thread? If not, isn't there an issue with propagation into the thread which makes the calls? I'm a little confused about what the specific use-case these new methods are fulfilling. Can you provide an example or two?

anuraaga:
@jkwatson Yeah they execute on the submitting thread, e.g.

```java
Context ctx = Context.current();
String result = httpClient.get("/foo")
  .thenCompose(ctx.wrapFunction(response -> httpClient.get(response.getRedirect()))
  // Not necessarily needs to be wrapped though
  .thenApply(ctx.wrapFunction(response -> response.getResult()));
```

Using `CompletableFuture` for anything just a bit more complex than that will be pretty intensely unwieldy vs bringing in an async execution library but I think these sort of simple invocations are reasonably common.

Note an alternative is to tell users to always use `Async` versions of the methods

```java
Context ctx = Context.current();
String result = httpClient.get("/foo")
  .thenComposeAsync(response -> httpClient.get(response.getRedirect()), ctx.wrap(Runnable::run))
  // Not necessarily needs to be wrapped though
  .thenApplyAsync(response -> response.getResult(), ctx.wrap(Runnable::run));
```

where `Runnable::run` is an `Executor`. This seems more awkward to reason about (why use an Executor just to propagate context) but is still OK. If we wanted to avoid the new methods, I think we would want to add `Context.executor()` to at least avoid the `Runnable::run`

```java
interface Context
  // Returns an {@link Executor} which will invoke all tasks synchronously, making this context current around each
  // task invocation. If you need to execute tasks asynchronously, use {@link Context#wrap(Executor executor)).
  default Executor executor() {
    return wrap(Runnable::run);
  }
}

Context ctx = Context.current();
String result = httpClient.get("/foo")
  .thenComposeAsync(response -> httpClient.get(response.getRedirect()), ctx.executor())
  // Not necessarily needs to be wrapped though
  .thenApplyAsync(response -> response.getResult(), ctx.executor());
```

I can drop this PR for that approach if it sounds better, is the javadoc for `executor()` grokkable?

jkwatson:
I'm totally ok with the new methods. I just wasn't sure about the usage and where it would help. Thanks!

jkwatson (approved):

jack-berg (approved):

PR #4157, opened 2022-02-07
Add interface for interpreting exceptions recorded to spans

HaloFour:
Adds interface to allow an extension to provide a different implementation for recording exceptions to spans.


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SpanExceptionRecorder.java
```diff
@@ -0,0 +1,13 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.trace;
+
+import io.opentelemetry.api.common.AttributesBuilder;
+
+@FunctionalInterface
+public interface SpanExceptionRecorder {
+  AttributesBuilder recordException(Throwable exception, AttributesBuilder attributesBuilder);
```
HaloFour, line 12: I toyed with having the interface be responsible for adding the event to the Span but I thought that I'd start here with attributes first.

anuraaga:
Hi @HaloFour - being able to configure the exception recording in principle sounds good but as it's not a Java-specific concept and a fairly major feature, we'd want to go through the opentelemetry-specification to add it. Wondering if this was a PoC to start that.

Also I notice it's somewhat awkward that exception semantics are implemented in the SDK, all other semantic conventions are implemented in instrumentation. It means that instrumentation libraries present many knobs such as customizing span name, status extraction, etc, but if we allow exception extraction customization that has to be configured separately in the SDK. I don't know if there is a good way to avoid this but wanted to record

anuraaga:
A random idea that came to mind, which could also be in an extension is

```
public final class FormattedThrowable {
  private final Throwable t;

  public Attributes format()
}

recordException(Throwable t) {
  if (t instanceof FormattedThrowable) {
    setAttributes(((FormattedThrowable) t).format());
    return this;
  }
  ...
```

HaloFour:
@anuraaga 

> Also I notice it's somewhat awkward that exception semantics are implemented in the SDK, all other semantic conventions are implemented in instrumentation. It means that instrumentation libraries present many knobs such as customizing span name, status extraction, etc, but if we allow exception extraction customization that has to be configured separately in the SDK. I don't know if there is a good way to avoid this but wanted to record

I understand the concern, especially a desire to keep too many knobs outside of the SDK.  My use case is covered if some kind of knob was exposed in the Spring Boot aspect, but to me it feels like doing this at the instrumentation level is also not appropriate unless there is a common way to do it that all instrumentation that might record exceptions would understand.  And even then, any code calling `Span#recordException` would bypass that knob.

IMO this feels like a common concern better addressed in the SDK, as it is in any given logging framework.  I can understand the hesitance of wanting to expose that knob as a public API until it is well vetted, though.

> `public class FormattedThrowable extends Throwable {`

Does this suggest that anywhere that might throw would have to catch/rethrow in order to wrap the exception?  If so I really don't like that approach.

jack-berg:
@anuraaga in the example you describe [here](https://github.com/open-telemetry/opentelemetry-java/pull/4157#issuecomment-1032240772) would the trace SDK place an implementation dependency on the experimental module in order to access `FormattedThrowable`?

anuraaga:
@jack-berg Yup

HaloFour:
I think I'd rather explore PR #4162 which instead propagates the `Throwable` through the event so that an exporter may decide to interpret it.  It keeps the instrumentation from having to be aware that exceptions might be formatted in any special way, or to have to be aware of what that format should be.

HaloFour:
Closing in favor of #4162

PR #4160, opened 2022-02-08
Deprecate log name field

jack-berg:
The log name has been removed from the [spec](https://github.com/open-telemetry/opentelemetry-specification/pull/2271) and is [being deprecated](https://github.com/open-telemetry/opentelemetry-proto/pull/357) from the proto.

### DIFF
--- a/sdk/logs/src/main/java/io/opentelemetry/sdk/logs/data/LogDataBuilder.java
+++ b/sdk/logs/src/main/java/io/opentelemetry/sdk/logs/data/LogDataBuilder.java
@@ -86,6 +86,7 @@ public final class LogDataBuilder {
   }
 
   /** Set the name. */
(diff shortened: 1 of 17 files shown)


sdk/logs/src/main/java/io/opentelemetry/sdk/logs/data/LogDataBuilder.java
```diff
@@ -86,6 +86,7 @@ public LogDataBuilder setSeverityText(String severityText) {
   }
 
   /** Set the name. */
```
jkwatson, line 88: might be nice to document the deprecation in javadoc and explain that it will be removed with no replacement.

jkwatson (approved):

anuraaga (approved):

PR #4162, opened 2022-02-08
Add ExceptionEventData which wraps events that contain exceptions

HaloFour:
Adds `ExceptionEventData` interface extending `EventData` to represent an exception recorded to a span.  Processors/exporters can access the exception directly to enable alternate formatting.


anuraaga:
This seems reasonable to me - we'd probably want to check on the spec repo if anyone else has any thoughts on how to format exceptions across languages

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/data/ExceptionEventData.java
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.trace.data;
+
+import io.opentelemetry.api.common.Attributes;
+
+public interface ExceptionEventData extends EventData {
```
jack-berg, line 10: I believe this can be package private.

jack-berg: Oh nvm. It has to be public so that processors / exporters can check `instanceof ExceptionEventData`. This seems reasonable to me too. 

If we wanted to be cautious we could put this interface in `:sdk-extensions:tracing-incubator` so users have to add a dependency on an unstable artifact in order to use it.

HaloFour: The SDK would need to know of the interface in order for `Span#recordException` to add an event that implements it, wouldn't it?

jkwatson: What about putting this in an `internal` package for now, as we explicitly document anything in `internal` as being unstable. You would be free to use it, of course, accepting the risk of it changing in the future.

jack-berg: > The SDK would need to know of the interface in order for Span#recordException to add an event that implements it, wouldn't it?

Yes, it could place an implementation dependency on `:sdk-extensions:tracing-incubator`.

> What about putting this in an internal package for now, as we explicitly document anything in internal as being unstable. 

That would work too, but probably better to not promote using internal classes.

jkwatson: I don't want to promote it...but @HaloFour could try it out while we figure out spec issues.

jkwatson: Also, it could be seen as an implementation detail for now, with future promotion out of the `internal` packaging.

jack-berg: That's definitely reasonable - I don't have a strong preference.

HaloFour: If the `internal` package is sufficient I can move to that.

Otherwise how would you feel about having the `ExceptionEventData` interface in `tracing-incubator` and SDK picks it up if it detects it on the class path?

jkwatson: I'd be fine with that as well.

HaloFour: Finished pass at moving the `ExceptionEventData` interface and implementation to `tracing-incubator`.  Added a helper class that looks for it on the classpath and will use the `create` static method to create instances, otherwise it defaults to the existing behavior.  Not terribly thrilled with the manual use of reflection but it seems that `MethodHandles` and `LambdaMetaFactory` aren't available due to the target Android API version and desugaring seems to only apply to them when used with `invokedynamic`, but hopefully given this code is related to exceptions the overhead of invoking the method is relatively a drop in the bucket.

Also made the implementation of `ExceptionEventData` defer creating of the common attributes until the `getAttributes` method is called and the result is memoized, so existing exporters should work as expected if using this implementation.  That behavior is certainly up for discussion.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/ExceptionEventUtil.java
```diff
@@ -0,0 +1,74 @@
...
+final class ExceptionEventUtil {
+  private static final ExceptionEventFactory factory = initializeEventFactory();
+
+  static EventData create(Throwable exception, Attributes additionalAttributes, long epochNanos) {
+    return factory.create(exception, additionalAttributes, epochNanos);
+  }
+
+  private static ExceptionEventFactory initializeEventFactory() {
+    try {
+      Class<?> clazz =
```
jkwatson: I worry a bit about having this reflection here, which will get logged as a warning on every SDK startup on a JDK 11+. I'm not sure I have a better solution, though.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/data/ExceptionEventData.java
HaloFour: > Do users need to opt out?

Not necessarily, just thought it was safer to be able to test it in isolation.

I can get the PR together to move these classes back to `:sdk:tracing` under an `internal` package.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/internal/data/ImmutableExceptionEventData.java
```diff
@@ -0,0 +1,72 @@
...
+  ImmutableExceptionEventData() {}
+
+  @Override
+  public String getName() {
+    return SemanticAttributes.EXCEPTION_EVENT_NAME;
+  }
+
+  @Override
+  @Memoized
+  public Attributes getAttributes() {
```
anuraaga, line 45:
```suggestion
  public final Attributes getAttributes() {
```

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/internal/data/ImmutableExceptionEventData.java
```diff
@@ -0,0 +1,72 @@
...
+
+  ImmutableExceptionEventData() {}
+
+  @Override
+  public String getName() {
+    return SemanticAttributes.EXCEPTION_EVENT_NAME;
+  }
+
+  @Override
+  @Memoized
```
anuraaga, line 44: Just to point out this changes the `getAttributes` for exceptions from a normal field to volatile field in the new codepath. I think it's OK just want to make sure everyone notices it

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/internal/data/ExceptionEventData.java
```diff
@@ -0,0 +1,39 @@
...
+   * @return the {@link Throwable exception} of the {@link ExceptionEventData}
+   */
+  Throwable getException();
+
+  /**
+   * Return the additional {@link Attributes attributes} of the {@link ExceptionEventData}.
+   *
+   * @return the additional {@link Attributes attributes} of the {@link ExceptionEventData}
+   */
+  Attributes getAdditionalAttributes();
```
anuraaga, line 38: I think this can be a package private method on `ImmutableExceptionEventData` instead of on this interface

sdk/trace/src/test/java/io/opentelemetry/sdk/trace/SdkSpanTest.java
```diff
@@ -894,6 +895,11 @@ void recordException() {
                 .put(SemanticAttributes.EXCEPTION_MESSAGE, "there was an exception")
                 .put(SemanticAttributes.EXCEPTION_STACKTRACE, stacktrace)
                 .build());
+
+    assertThat(event).isInstanceOf(ExceptionEventData.class);
```
anuraaga, line 899: Consider `isInstanceOfSatisfying`

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/internal/data/ImmutableExceptionEventData.java
HaloFour: Going for the stretch goal of having `ExceptionEventData` not generate those attributes unless the exporter tries to use them.  Assuming the exporter only calls `getAttributes` once it's probably not actually necessary to memoize them.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/internal/data/ExceptionEventData.java
HaloFour: As with that stretch goal I wanted to enable an exporter to access the additional attributes without forcing the semantic attributes for the exception to be created.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/internal/data/ImmutableExceptionEventData.java
HaloFour: If there is hesitation in deferring the generation of those attributes I can definitely back that off.  When the interface was in `tracing-incubation` I felt it was worth exploring.  The only attribute I'd prefer to avoid generating is `exception.stacktrace` as in my Spring codebases they can be quite large, but I don't think it'd a big deal to consider that later.

anuraaga (approved):
Sorry for delay thought had already hit approve

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/internal/data/ExceptionEventData.java
```diff
@@ -0,0 +1,39 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.trace.internal.data;
+
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.sdk.trace.data.EventData;
+
+public interface ExceptionEventData extends EventData {
```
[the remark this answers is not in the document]
HaloFour: I added the missing `package-info`, is there other boilerplate that I'm missing?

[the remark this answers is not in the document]
HaloFour: oops, I see that verbiage is repeated on the [classes](https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/internal/ReadOnlyArrayMap.java#L37) as well, will add that.

HaloFour:
Just checking in to see what the thoughts are about this PR.

jkwatson (approved):

kubawach (approved):
Looks good, thanks!

jack-berg (approved):
Any reason not to merge this?

Issue #3108, opened 2021-04-05
Metric view registry system property autoconfiguration options


jack-berg:
**Is your feature request related to a problem? Please describe.**
Its hard to configure the views for `SdkMetricProvider` when using `opentelemetry-java-instrumentation`. The addition of `SdkMeterProviderConfigurer` makes it possible, but users are still required to package up their implementation in a special initializer jar. It should be easy to do basic metric view configuration via system properties. 

**Describe the solution you'd like**
I'd like to see a set of system properties / environment variables like:

| System property          | Environment variable     | Description                                                                       |
|--------------------------|--------------------------|-----------------------------------------------------------------------------------|
| otel.mvr.counter.aggregator | OTEL_MVR_COUNTER_AGGREGATOR | The aggregator factory used for COUNTER instruments. Default is `cumulative_sum`.|
| otel.mvr.up_down_counter.aggregator | OTEL_MVR_UP_DOWN_COUNTER_AGGREGATOR | The aggregator factory used for UP_DOWN_COUNTER instruments. Default is `cumulative_sum`.|
| otel.mvr.sum_observer.aggregator | OTEL_MVR_SUM_OBSERVER_AGGREGATOR | The aggregator factory used for SUM_OBSERVER instruments. Default is `cumulative_sum`.|
| otel.mvr.up_down_sum_observer.aggregator | OTEL_MVR_UP_DOWN_SUM_OBSERVER_AGGREGATOR | The aggregator factory used for UP_DOWN_SUM_OBSERVER instruments. Default is `cumulative_sum`.|
| otel.mvr.value_recorder.aggregator | OTEL_MVR_VALUE_RECORDER_AGGREGATOR | The aggregator factory used for VALUE_RECORDER instruments. Default is `summary`.|
| otel.mvr.value_observer.aggregator | OTEL_MVR_VALUE_OBSERVER_AGGREGATOR | The aggregator factory used for VALUE_OBSERVER instruments. Default is `last_value`.|

Supported values for `otel.mvr.{instrument}.aggregator` are

- "cumulative_sum": AggregatorFactory.sum(alwaysCumulative=true)
- "delta_sum": AggregatorFactory.sum(alwaysCumulative=false)
- "cumulative_count": AggregatorFactory.count(temporality=CUMULATIVE)
- "delta_count": AggregatorFactory.count(temporality=DELTA)
- "summary": AggregatorFactory.minMaxSumCount()
[...]

PR #4163, opened 2022-02-08
File based view configuration

jack-berg:
A while ago I opened an #3108 requesting system props / env var configuration for views. I no longer think we should expose view configuration via system props / env vars because trying to come up with a scheme to represent the config sounds like a mess.

However, I do still think its important to expose some mechanism for view configuration besides programatic configuration. In this PR I propose a file based YAML scheme which can be translated into a corresponding view configuration. This concept is not currently specified, but it has been discussed, and I think having a prototype would be valuable for collecting feedback.

If folks are directionally aligned with this, I would like to extend this in a followup PR to include a hook into autoconfiguration. Perhaps an implementation of `AutoConfigurationCustomizer#addMeterProviderCustomizer(..)` would be best.

I would also like to extend it to include support for regular expressions in the configuration. This should be straight forward, but I omitted it for now because it's not crucial to illustrate the concept. 

**Edit 2/10/2022**
I've gone ahead and extended this PR a bit:
- Added support for attribute filters based on keys
- Implemented `AutoConfigurationCustomizerProvider` so that it can be used with autoconfigure to load view config files from env vars / system props
- Renamed module to `:sdk-extensions:metric-incubator` so it can be a home for other incubating metric features


sdk-extensions/view-config/README.md
```diff
@@ -0,0 +1,63 @@
...
+Suppose `/Users/user123/view.yaml` had content as follows:
+
+```yaml
+- selector:
+    instrumentName: my-instrument
+    instrumentType: COUNTER
+    meterName: my-meter
+    meterVersion: 1.0.0
+    meterSchemaUrl: http://example.com
+  view:
```
jsuereth, line 23: I love this!

1. we may want to think about providing "default" views for our instrumentation APIs as well, regarding passing ALL attributes to a meter and reducing down to the semantic conventions).
2. We'd like to (eventually) have SDK configuration components be accessible via config, similar to this.   Are you planning to use this implementation to propose addition to the Specification?  I think you should as I'd like for some consistency in SDK config across languages.
3. Regarding "precedence" order, I assume it's just the order these files are passed into `registerViews`? 
4. Should we offer any "logging-like" configuration-lookups via things like ServiceLoader?

jack-berg: > 1. we may want to think about providing "default" views for our instrumentation APIs as well, regarding passing ALL attributes to a meter and reducing down to the semantic conventions).

Yes, this could be extended to include additional parameters which translate to the attribute / baggage components of the view. For example, could have `attributeFilterAllowableKeys: ["foo", "bar", "baz"]` to remove attribute not in the allow list.

>2. We'd like to (eventually) have SDK configuration components be accessible via config, similar to this. Are you planning to use this implementation to propose addition to the Specification? I think you should as I'd like for some consistency in SDK config across languages.

Yeah I like the idea for a file based configuration specification for the SDK. Configuration of the view API is a pretty urgent gap IMO, so I figured its a nice place to build prototypes that can be used to explore ideas and collect feedback. I definitely think that having a consistent spec for file based config that crosses language is important. 

>3. Regarding "precedence" order, I assume it's just the order these files are passed into registerViews?

Yes, if you were to call this with multiple files, the views would be registered in order in which the files were parsed, and for each file, in the order that the views were configured.

>4. Should we offer any "logging-like" configuration-lookups via things like ServiceLoader?

Do you mean like searching for configuration files in standard locations? If so, I'm not opposed, but personally would prefer to have an environment variable point to the absolute path of a config file.

jkwatson:
Putting this in an alpha sdk-extension is a great idea. Perfect place to vet the ideas and make sure our SDK has the appropriate programmatic configuration in place to make this work. 👍🏽

sdk-extensions/metric-incubator/src/main/java/io/opentelemetry/sdk/viewconfig/ViewConfig.java
```diff
@@ -0,0 +1,220 @@
...
+      assert viewConfigSpec.getViewSpecification() != null;
+      meterProviderBuilder.registerView(
+          toInstrumentSelector(viewConfigSpec.getSelectorSpecification()),
+          toView(viewConfigSpec.getViewSpecification()));
+    }
+  }
+
+  // Visible for testing
+  @SuppressWarnings("unchecked")
+  static List<ViewConfigSpecification> loadViewConfig(File viewConfigYamlFile) {
```
jack-berg, line 102: Updated from jackson to snakeyaml. Just manually extracting fields from generic list / and map types here. Snakeyaml works with the javabean spec, but not clear if its possible to make it play nice with autovalue builders. Its not pretty, but it works, and is only 17 lines longer than the jackson solution.

docs/apidiffs/current_vs_latest/opentelemetry-sdk-extension-view-config.txt
```diff
@@ -0,0 +1,4 @@
+Comparing source compatibility of  against
```
anuraaga, line 1: Looks like need to run `./gradlew check` to refresh this

sdk-extensions/metric-incubator/README.md
```diff
@@ -0,0 +1,77 @@
...
+           .setAggregation(Aggregation.histogram())
+           .filterAttributes(key -> new HashSet<>(Arrays.asList("foo", "bar")).contains(key))
+           .build());
+```
+
+If using [autoconfigure](../autoconfigure) with this artifact on your classpath, it will automatically load a list of view config files specified via environment variable or system property:
+
+| System property                      | Environment variable                 | Purpose                                                    |
+|--------------------------------------|--------------------------------------|------------------------------------------------------------|
+| otel.experimental.metric.view.config | OTEL_EXPERIMENTAL_METRIC_VIEW_CONFIG | List of absolute files containing view configuration YAML. |
```
anuraaga, line 56:
```suggestion
| otel.experimental.metrics.view_config | OTEL_EXPERIMENTAL_METRIC_VIEW_CONFIG | List of absolute paths to files containing view configuration YAML. |
```

sdk-extensions/metric-incubator/src/main/java/io/opentelemetry/sdk/viewconfig/ViewConfig.java
```diff
@@ -0,0 +1,220 @@
...
+ *             .build(),
+ *         View.builder()
+ *             .setName("new-instrument")
+ *             .setDescription("new-description")
+ *             .setAggregation(Aggregation.histogram())
+ *             .filterAttributes(key -> new HashSet<>(Arrays.asList("foo", "bar")).contains(key))
+ *             .build());
+ * }</pre>
+ */
+public class ViewConfig {
```
anuraaga, line 76:
```suggestion
public final class ViewConfig {
```

sdk-extensions/metric-incubator/src/main/java/io/opentelemetry/sdk/viewconfig/ViewConfig.java
```diff
@@ -0,0 +1,220 @@
...
+  @SuppressWarnings("unchecked")
+  static List<ViewConfigSpecification> loadViewConfig(File viewConfigYamlFile) {
+    Yaml yaml = new Yaml();
+    try {
+      List<ViewConfigSpecification> result = new ArrayList<>();
+      List<Map<String, Object>> viewConfigs =
+          yaml.load(Files.newBufferedReader(viewConfigYamlFile.toPath(), StandardCharsets.UTF_8));
+      for (Map<String, Object> viewConfigSpecMap : viewConfigs) {
+        Map<String, Object> selectorSpecMap =
+            Objects.requireNonNull(
```
anuraaga, line 110: We'd usually static import requireNonNull

sdk-extensions/metric-incubator/src/main/java/io/opentelemetry/sdk/viewconfig/ViewConfig.java
```diff
@@ -0,0 +1,220 @@
...
+              + " but was "
+              + value.getClass().getName());
+    }
+    return (T) value;
+  }
+
+  // Visible for testing
+  static View toView(ViewSpecification viewSpec) {
+    ViewBuilder builder = View.builder();
+    Optional.ofNullable(viewSpec.getName()).ifPresent(builder::setName);
```
anuraaga, line 176: Above `ofNullable` is providing some value through `map` I think and is ok but I would say `ifPresent` at least without some other mapping, is almost always an anti-pattern, null checks in if statements are simpler and keep the code easier to read

sdk-extensions/metric-incubator/src/main/java/io/opentelemetry/sdk/viewconfig/ViewConfig.java
```diff
@@ -0,0 +1,220 @@
...
+
+  private ViewConfig() {}
+
+  /**
+   * Load the view configuration YAML from the file and apply it to the {@link
+   * SdkMeterProviderBuilder}.
+   *
+   * @throws ConfigurationException if unable to interpret file contents
+   */
+  @SuppressWarnings("NullAway")
```
anuraaga, line 86: Perhaps something changed in an iteration of the PR but the methods being `assert`ed against null don't seem to be marked `Nullable` so I guess we don't need this or those assertions

anuraaga: For reference, nullaway will recognize `requireNonNull` but not `assert` since it is trying to ensure runtime safety (I wish it did check asserts personally though). So using it would be better than suppressing where it would be needed

sdk-extensions/metric-incubator/src/test/java/io/opentelemetry/sdk/viewconfig/ViewConfigCustomizerTest.java
```diff
@@ -0,0 +1,108 @@
...
+import org.junit.jupiter.api.Test;
+
+class ViewConfigCustomizerTest {
+
+  @Test
+  void customizeMeterProvider_Spi() {
+    System.setProperty("otel.traces.exporter", "none");
+    System.setProperty(
+        "otel.experimental.metric.view.config",
+        ViewConfigTest.class.getResource("/view-config-customizer-test.yaml").getFile());
```
anuraaga, line 30: I think there is huge demand to be able to support bundled resources with this sort of configuration property. It seems we can copy spring's practice, how about we add support for `classpath:`?

https://www.baeldung.com/spring-classpath-file-access#2-using-value

sdk-extensions/metric-incubator/src/test/java/io/opentelemetry/sdk/viewconfig/ViewConfigTest.java
```diff
@@ -0,0 +1,185 @@
...
+              assertThat(selectorSpec.getInstrumentName()).isEqualTo("name1");
+              assertThat(selectorSpec.getInstrumentType()).isEqualTo(InstrumentType.COUNTER);
+              assertThat(selectorSpec.getMeterName()).isEqualTo("meterName1");
+              assertThat(selectorSpec.getMeterVersion()).isEqualTo("1.0.0");
+              assertThat(selectorSpec.getMeterSchemaUrl()).isEqualTo("http://example1.com");
+              ViewSpecification viewSpec = viewConfigSpec.getViewSpecification();
+              assertThat(viewSpec.getName()).isEqualTo("name1");
+              assertThat(viewSpec.getDescription()).isEqualTo("description1");
+              assertThat(viewSpec.getAggregation()).isEqualTo("sum");
+              assertThat(viewSpec.getAttributeKeys()).isEqualTo(Arrays.asList("foo", "bar"));
```
anuraaga, line 67:
```suggestion
              assertThat(viewSpec.getAttributeKeys()).containsExactly("foo", "bar");
```

sdk-extensions/metric-incubator/src/test/java/io/opentelemetry/sdk/viewconfig/ViewConfigTest.java
```diff
@@ -0,0 +1,185 @@
...
+              assertThat(selectorSpec.getInstrumentName()).isEqualTo("name2");
+              assertThat(selectorSpec.getInstrumentType()).isEqualTo(InstrumentType.COUNTER);
+              assertThat(selectorSpec.getMeterName()).isEqualTo("meterName2");
+              assertThat(selectorSpec.getMeterVersion()).isEqualTo("2.0.0");
+              assertThat(selectorSpec.getMeterSchemaUrl()).isEqualTo("http://example2.com");
+              ViewSpecification viewSpec = viewConfigSpec.getViewSpecification();
+              assertThat(viewSpec.getName()).isEqualTo("name2");
+              assertThat(viewSpec.getDescription()).isEqualTo("description2");
+              assertThat(viewSpec.getAggregation()).isEqualTo("last_value");
+              assertThat(viewSpec.getAttributeKeys()).isEqualTo(Arrays.asList("baz", "qux"));
```
anuraaga, line 80:
```suggestion
              assertThat(viewSpec.getAttributeKeys()).containsExactly("baz", "qux");
```

sdk-extensions/metric-incubator/README.md
jack-berg: How about `otel.experimental.metrics.view-config` for consistency with [otel.java.disabled.resource-providers](https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk-extensions/autoconfigure#disabling-automatic-resourceproviders)?

sdk-extensions/metric-incubator/src/main/java/io/opentelemetry/sdk/viewconfig/ViewConfig.java
jack-berg: Yes, with the change to snakeyaml we're get stronger non-null guarantees than with jackson (at the expense of manual parsing code).

sdk-extensions/metric-incubator/src/main/java/io/opentelemetry/sdk/viewconfig/ViewConfig.java
jack-berg: Oh but its so many extra lines! I've gotten use to reading `Optional.ofNullable().ifPresent(..)` and don't mind it if lambdas can be used in `ifPresent()` (preferably method references), and when throughput is low so the additional allocation for optional is negligible. 

I get it though - not the style I prefer but good to have a convention and stick with it.

sdk-extensions/metric-incubator/src/test/java/io/opentelemetry/sdk/viewconfig/ViewConfigCustomizerTest.java
jack-berg: Good idea. Do you foresee any issuing accessing resources when this is packaged with the agent?

sdk-extensions/metric-incubator/src/test/java/io/opentelemetry/sdk/viewconfig/ViewConfigCustomizerTest.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.UUID;
+import org.junit.jupiter.api.Test;
+
+class ViewConfigCustomizerTest {
+
+  @Test
+  void customizeMeterProvider_Spi() {
+    System.setProperty("otel.traces.exporter", "none");
+    System.setProperty(
+        "otel.experimental.metrics.view.config",
```
anuraaga, line 31: Now that we have `classpath:` we can easily use junit-pioneer's `@SetSystemProperty` for these to make sure they're cleaned up

Well I was going to say that but realized we should probably just `addPropertiesSupplier`

sdk-extensions/metric-incubator/src/test/java/io/opentelemetry/sdk/viewconfig/ViewConfigCustomizerTest.java
```diff
@@ -0,0 +1,130 @@
...
+                    withConfigFileLocations(
+                        Collections.singletonList(
+                            ViewConfigTest.class
+                                .getResource("/empty-selector-config.yaml")
+                                .getFile()))))
+        .hasMessageContaining("Failed to parse view config")
+        .hasRootCauseMessage("selector is required");
+  }
+
+  private static ConfigProperties withConfigFileLocations(List<String> fileLocations) {
```
anuraaga, line 125: Tiny nit you could make this varargs and have the `Arrays.asList` in here instead

sdk-extensions/metric-incubator/src/test/java/io/opentelemetry/sdk/viewconfig/ViewConfigCustomizerTest.java
```diff
@@ -0,0 +1,130 @@
...
+                  Attributes.builder()
+                      .put("foo", "val")
+                      .put("bar", "val")
+                      .put("baz", "val")
+                      .put("qux", "val")
+                      .build());
+            });
+
+    assertThat(reader.collectAllMetrics())
+        .hasSize(1)
```
anuraaga, line 59:
```suggestion
```

`.Exactly` checks the size too may as well let it handle

sdk-extensions/metric-incubator/src/test/java/io/opentelemetry/sdk/viewconfig/ViewConfigTest.java
```diff
@@ -0,0 +1,214 @@
...
+        .isInstanceOf(ConfigurationException.class)
+        .hasMessageContaining("An error occurred reading view config file:");
+  }
+
+  @Test
+  void registerViews_FileFullConfig() {
+    SdkMeterProviderBuilder builder = SdkMeterProvider.builder();
+
+    ViewConfig.registerViews(
+        builder, new File(ViewConfigTest.class.getResource("/full-config.yaml").getFile()));
```
anuraaga, line 53: While in practice we don't expect `getFile` to fail when running tests it always makes me nervous calling it on a resource, how about moving the buffered reader stuff to `registerViews` so we can pass an `InputStream` in both from the SPI and tests?

anuraaga: Alternatively looks like you have a `bufferedReader` method in this class, but could still consider moving that into ViewConfig

jack-berg: If `registerViews` only accepts an `InputStream`, the API surface area becomes really simple and folks can use it however they want - whether that be from a string, file, etc. I like it.

anuraaga (approved):

PR #4168, opened 2022-02-11
Exponential Histogram Bucket Indexing Improvements

jamesmoessis:
## Changes 

- Delete BucketMapper interface and move logic to private method. 
- Write test for histogram at Double limits (min and max double values) 
- Fix corner case bug caused by integer overflow 

We need at least two approaches for calculating bucket indexes for it to be fully accurate. This is from the spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#producer-expectations

For positive and negative scales, the method for calculating the index will be different. 

Furthermore, the new indexing method allowed me to write a test for recording the histogram at its limits - `Double.MIN_VALUE` - `Double.MAX_VALUE`. Looks like it works fine with subnormals.


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AdaptingCircularBufferCounter.java
```diff
@@ -73,13 +73,13 @@ public boolean increment(int index, long delta) {
 
     if (index > endIndex) {
       // Move end, check max size
-      if (index - startIndex + 1 > backing.length()) {
+      if ((long) index - (long) startIndex + 1L > backing.length()) {
```
jamesmoessis, line 76: When recording `Double.MIN_VALUE` and `Double.MAX_VALUE` at scale 20 in the same histogram, the two indexes here differ so much that it causes an integer overflow. As a result, the if statement evaluates to false when it should be true. This causes an error later on because the histogram didn't downscale, and it tries to create an array of negative length. 

Casting to `long` fixes this, as it can't possibly overflow at the max scale of 20.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleExponentialHistogramAggregatorTest.java
```diff
@@ -122,6 +135,45 @@ void testInvalidRecording() {
     assertThat(acc.getZeroCount()).isEqualTo(0);
   }
 
+  @ParameterizedTest
```
jamesmoessis, line 138: Tested this for both `mapCounter` and `circularBufferCounter` impls because they both have index logic that can possibly overflow.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleExponentialHistogramBuckets.java
```diff
@@ -249,6 +245,34 @@ int getScaleReduction(long newStart, long newEnd) {
...
+  }
+
+  /**
+   * Maps a recorded double value to a bucket index.
+   *
+   * <p>The strategy to retrieve the index is specified in the OpenTelemtry specification:
+   * https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#exponential-buckets
+   *
+   * <p>If the index cannot be represented by a signed 32 bit integer, it is expected the histogram
+   * will be downscaled.
```
anuraaga, line 264:
```suggestion
   * <p>Indexes outside the range of a signed 32-bit integer indicate an incorrect scale.
```

Is this what this line means? Not quite sure what the expectation actually refers to

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleExponentialHistogramBuckets.java
```diff
@@ -249,6 +245,34 @@ int getScaleReduction(long newStart, long newEnd) {
     return scaleReduction;
   }
 
+  private static int getIndexByLogarithm(double value, int scale) {
+    double scaleFactor = Math.scalb(1D / Math.log(2), scale);
```
anuraaga, line 249: Let's store `logScaleFactor` in a field like it was before to avoid recalculating every time

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AdaptingCircularBufferCounter.java
anuraaga: Think it makes sense to reduce some noise with just one cast

`(long) index - startIndex + 1`

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleExponentialHistogramAggregatorTest.java
```diff
@@ -122,6 +135,45 @@ void testInvalidRecording() {
...
+
+    // if scale is -3, base is 256.
+    int base = 256;
+
+    // Verify the rule holds:
+    // base ^ (offset+i) <= (values recorded to bucket i) < base ^ (offset+i+1)
+
+    // lowest bucket:
+    assertThat(Math.pow(base, acc.getPositiveBuckets().getOffset()))
+        .isLessThanOrEqualTo(Double.MIN_VALUE);
```
anuraaga, line 166: Should this be isEqualTo? It shouldn't be possible for the double result of pow to be less than the MIN See below for related

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleExponentialHistogramAggregatorTest.java
```diff
@@ -122,6 +135,45 @@ void testInvalidRecording() {
...
+    assertThat(Math.pow(base, acc.getPositiveBuckets().getOffset() + 1))
+        .isGreaterThan(Double.MIN_VALUE);
+
+    // highest bucket
+    assertThat(Math.pow(base, acc.getPositiveBuckets().getOffset() + bucketCounts.size() - 1))
+        .isLessThanOrEqualTo(Double.MAX_VALUE);
+    assertThat(Math.pow(base, acc.getPositiveBuckets().getOffset() + bucketCounts.size()))
```
anuraaga, line 173: The result of `pow` is `double` so I don't see how it is possible for it to be `greaterThan(Double.MAX_VALUE)`. This assertion seems to be passing though maybe due to some non-obvious overflow happening, but there is probably a better assertion than this non-intuitive one.

jamesmoessis: This test case relies on the result of `Math.pow()` here returning `Infinity`. It's not entirely accurate, but it works and indicates that the upper bound of the bucket is placed correctly above the recorded value. I suppose I could use `BigDecimal` here instead? 

See above comment for problem with `BigDecimal`.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleExponentialHistogramAggregatorTest.java
jamesmoessis: This test case relies on `Math.pow()` rounding down to zero. Again, it's not entirely accurate, but it works and indicates the lower bound of the bucket is placed correctly. The bucket bounds are never explicitly calculated in the main code so it never becomes a problem there. 

I tried using `BigDecimal` here: 
```java
BigDecimal lowerBound = new BigDecimal("256").pow(acc.getPositiveBuckets().getOffset());
assertThat(lowerBound)
    .isLessThanOrEqualTo(new BigDecimal("4.9E-324")); // double min value as BigDecimal
```

but the `BigDecimal.pow()` throws and says invalid operation. I'm open to suggestions on better approaches for this. Let me know if you have any ideas.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleExponentialHistogramBuckets.java
jamesmoessis: I've gone ahead and done this.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleExponentialHistogramBuckets.java
jamesmoessis: I've removed this line, it isn't really necessary and only tangentially related to this method, so it's causing more confusion than it's worth.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AdaptingCircularBufferCounter.java
jamesmoessis: Good suggestion, all done.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleExponentialHistogramAggregatorTest.java
anuraaga: How about `.isEqualTo(0)` and `isEqualTo(Infinity)` here and below? I think it's a bit clearer than less than or equal to min, greater than max. We could add single line comments like Math.pow rounds down to zero or up to Infinity

jamesmoessis: Yeah that makes sense, I agree that's clearer. I've added comments and used `.isEqualTo(0)` and `isEqualTo(Infinity)`.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleExponentialHistogramBuckets.java
```diff
@@ -249,6 +251,35 @@ int getScaleReduction(long newStart, long newEnd) {
     return scaleReduction;
   }
 
+  private int getIndexByLogarithm(double value) {
+    return (int) Math.floor(Math.log(value) * scaleFactor);
```
anuraaga, line 255: I guess `(int)` cast without the floor would behave the same

anuraaga (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleExponentialHistogramBuckets.java
jamesmoessis: Not for negative values: 
```
(int) -2.2  == -2
(int) Math.floor(-2.2) == -3
```

jamesmoessis: But then again, this function should never receive negative values 🤔

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AdaptingCircularBufferCounter.java
jack-berg: Might be worth an explanatory comment.

jamesmoessis:
@jsuereth sorry I didn't see your comment earlier. I just ran the `HistogramBenchmark` locally, and I'm getting results a lot slower (100x) than what I can see in `histograms.png`. This is for both explicit bucket and exponential histograms. What kind of machine did you run the benchmark on initially?

jack-berg:
@jamesmoessis can you run the benchmark with and without this commit to produce a like for like comparison on the same machine?

jamesmoessis:
@jack-berg I've run with both - no significant differences between them. 

I've uploaded the results pages over here if you are interested at looking in the graphs. They are too big to include in a screenshot. https://github.com/atlassian-forks/opentelemetry-java/tree/perf-test/sdk/metrics/src

PR #4169, opened 2022-02-11
Experimental support for disabling resource keys

jack-berg:
Allow you to exclude resource attribute keys with `export OTEL_EXPERIMENTAL_DISABLED_RESOURCE_KEYS=foo,bar`.

This stems from conversations about filtering out `process.command_line` resource attributes, which are lengthy, and may contain sensitive information.

Its also possible to do accomplish something similar by using `otel.java.disabled.resource-providers` to disable the entire `ProcessResourceProvider`, but that results in other useful process resource attributes being filtered out.

Its also possible to split out a separate `ProcessCommandLineResourceProvider`, so the command line can be disabled on its own.

Its also possible to disable `process.command_line` in `ProcessResourceProvider` by default, and make it opt in through some env var / system property.

However, I think having a general mechanism for this is a nice tool because it allows users to be able to make fine-grained decisions about each resource attribute, weighing the cost from the additional bytes on the wire against the value it provides. 

I didn't find any conversation about this in the spec, but may have missed something. Its possible that this issue doesn't affect other language SDKs if they lack features to automatically detect resources (i.e. if the resource attributes are already opt in), but my hunch is that this would be a valuable feature to have general support.

### DIFF
--- a/sdk-extensions/autoconfigure/README.md
+++ b/sdk-extensions/autoconfigure/README.md
@@ -161,10 +161,11 @@ Supported values are
 The [OpenTelemetry Resource](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/sdk.md)
 is a representation of the entity producing telemetry.
 
-| System property          | Environment variable     | Description                                                                                                |
-|--------------------------|--------------------------|------------------------------------------------------------------------------------------------------------|
-| otel.resource.attributes | OTEL_RESOURCE_ATTRIBUTES | Specify resource attributes in the following format: key1=val1,key2=val2,key3=val3                         |
-| otel.service.name        | OTEL_SERVICE_NAME        | Specify logical service name. Takes precedence over `service.name` defined with `otel.resource.attributes` |
+| System property                          | Environment variable                     | Description                                                                                                |
+|------------------------------------------|------------------------------------------|------------------------------------------------------------------------------------------------------------|
+| otel.resource.attributes                 | OTEL_RESOURCE_ATTRIBUTES                 | Specify resource attributes in the following format: key1=val1,key2=val2,key3=val3                         |
+| otel.service.name                        | OTEL_SERVICE_NAME                        | Specify logical service name. Takes precedence over `service.name` defined with `otel.resource.attributes` |
+| otel.experimental.disabled.resource.keys | OTEL_EXPERIMENTAL_DISABLED_RESOURCE_KEYS | Specify resource attribute keys that are filtered.                                                         |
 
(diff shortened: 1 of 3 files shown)


jkwatson (approved):
👍🏽  given the experimental namespace

trask:
I like this, it's a lot easier for users to know attribute names they want to disable, compared to knowing the class name of the resource provider that they want to disable

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ResourceConfiguration.java
```diff
@@ -62,5 +66,23 @@ static Attributes getAttributes(ConfigProperties configProperties) {
     return resourceAttributes.build();
   }
 
+  // visible for testing
+  static Resource filterAttributes(Resource resource, ConfigProperties configProperties) {
+    Set<String> disabledKeys = new HashSet<>(configProperties.getList(DISABLED_ATTRIBUTE_KEYS));
+
+    Attributes filteredAttributes =
+        resource.getAttributes().toBuilder()
+            .removeIf(attributeKey -> disabledKeys.contains(attributeKey.getKey()))
```
anuraaga, line 75: FYI filed a couple of issues which would improve this code

#4175 #4176

anuraaga (approved):

sdk-extensions/autoconfigure/README.md
```diff
@@ -161,10 +161,11 @@ Supported values are
...
-| otel.service.name        | OTEL_SERVICE_NAME        | Specify logical service name. Takes precedence over `service.name` defined with `otel.resource.attributes` |
+| System property                          | Environment variable                     | Description                                                                                                |
+|------------------------------------------|------------------------------------------|------------------------------------------------------------------------------------------------------------|
+| otel.resource.attributes                 | OTEL_RESOURCE_ATTRIBUTES                 | Specify resource attributes in the following format: key1=val1,key2=val2,key3=val3                         |
+| otel.service.name                        | OTEL_SERVICE_NAME                        | Specify logical service name. Takes precedence over `service.name` defined with `otel.resource.attributes` |
+| otel.experimental.disabled.resource.keys | OTEL_EXPERIMENTAL_DISABLED_RESOURCE_KEYS | Specify resource attribute keys that are filtered.                                                         |
```
anuraaga, line 168: `otel.experimental.resource.disabled_keys`

jack-berg: How about `otel.experimental.resource.disabled-keys` for consistency with `otel.java.disabled.resource-providers`? 

Reminder that the `-` and `_` only matter for the readme because they're replaced by `.` when reading in from env / system props.

anuraaga (approved):

jsuereth (approved):

PR #4173, opened 2022-02-14
Add ability to do loose attribute matching while checking all keys.

### LINKED ISSUE #3993: More flexible AttributesAssert containsOnly feature
In the instrumentation repo, the groovy-based attribute assertions keep track of the attributes that were asserted against, and fail if there are any additional attributes outside of those expected ones, which is a nice validation against unexpected attributes.

It would be really helpful if the java-based `AttributesAssert` that we are starting to adopt from this repo had some way to do the same.

I'm not sure the best way to expose/implement this functionality. One option is potentially for `containsOnly` to accept flexible value matching (e.g. `hasEntrySatisfying`).

Some background at https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4872#issue-1077157302

anuraaga:
Fixes #3993 

Played around a bit and came up with this which seems to have good UX. The goals are

1) Can define all assertions with lambdas
2) Can use same method name for all types
3) All entrypoints in one location (OpenTelemetryAssertions)
4) Can mix exact and loose assertions (lower priority)

Took a lot of API to achieve these, especially 2) but I guess we want to prioritize ease of writing assertions and this seems OK but open to any feedback to make it better.

### DIFF
--- a/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/AttributeAssertion.java
+++ b/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/AttributeAssertion.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.testing.assertj;
+
+import com.google.auto.value.AutoValue;
+import io.opentelemetry.api.common.AttributeKey;
+import java.util.function.Consumer;
+import org.assertj.core.api.AbstractAssert;
+
+/** An assertion on an attribute key. */
+@AutoValue
+public abstract class AttributeAssertion {
+
+  // This method is not type-safe! But because the constructor is private, we know it will only be
+  // created through
+  // our factories, which are type-safe.
+  @SuppressWarnings("unchecked")
+  static AttributeAssertion create(
+      AttributeKey<?> key, Consumer<? extends AbstractAssert<?, ?>> assertion) {
+    return new AutoValue_AttributeAssertion(key, (Consumer<AbstractAssert<?, ?>>) assertion);
+  }
+
+  abstract AttributeKey<?> getKey();
+
+  abstract Consumer<AbstractAssert<?, ?>> getAssertion();
+
+  AttributeAssertion() {}
+}
--- a/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/OpenTelemetryAssertions.java
+++ b/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/OpenTelemetryAssertions.java
@@ -195,7 +195,7 @@ public final class OpenTelemetryAssertions extends Assertions {
    * Returns an {@link AttributeAssertion} that asserts the given {@code key} is present with the
    * given {@code value}.
    */
-  public static <T> AttributeAssertion attribute(AttributeKey<T> key, T value) {
+  public static <T> AttributeAssertion equalTo(AttributeKey<T> key, T value) {
     return AttributeAssertion.create(key, val -> val.isEqualTo(value));
   }
 
--- a/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/OpenTelemetryAssertions.java
+++ b/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/OpenTelemetryAssertions.java
@@ -13,8 +13,14 @@ import java.util.AbstractMap;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
+import org.assertj.core.api.AbstractBooleanAssert;
+import org.assertj.core.api.AbstractDoubleAssert;
+import org.assertj.core.api.AbstractLongAssert;
+import org.assertj.core.api.AbstractStringAssert;
 import org.assertj.core.api.Assertions;
+import org.assertj.core.api.ListAssert;
 
 /**
  * Entry point for assertion methods for OpenTelemetry types. To use type-specific assertions,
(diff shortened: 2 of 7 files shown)


sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/OpenTelemetryAssertions.java
```diff
@@ -111,6 +117,107 @@ public static EventDataAssert assertThat(EventData eventData) {
...
+   * given {@code value}.
+   */
+  public static <T> AttributeAssertion attribute(AttributeKey<T> key, T value) {
+    return AttributeAssertion.create(key, val -> val.isEqualTo(value));
+  }
+
+  // Unique interfaces to prevent generic functional interface clash. These are not interesting at
+  // all but are required to be able to use the same method name in methods like satisfies above.
+
+  public interface StringAssertConsumer extends Consumer<AbstractStringAssert<?>> {}
```
anuraaga, line 205: Left these nested without javadoc since these are just types corresponding to lambdas and should never be directly referenced by users.

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/SpanDataAssert.java
```diff
@@ -268,6 +272,57 @@ public SpanDataAssert hasAttributesSatisfying(Consumer<Attributes> attributes) {
...
+      Object value = actual.getAttributes().get(key);
+      // Already checked all keys are present.
+      requireNonNull(value);
+      AbstractAssert<?, ?> assertion = attributeValueAssertion(key, value);
+      attributeAssertion.getAssertion().accept(assertion);
+    }
+    return this;
+  }
+
+  // The return type of these assertions must match the parameters in methods like
```
anuraaga, line 305: This is the main trick. I wasn't sure I'd be able to have a single typesafe method that can accept arbitrary assertions but by having these match factories, and factories the only way to create the assertions, it seems to work. Notably, this is to avoid `Consumer<? extends AbstractAssert>`, which would not work. We're taking advantage of type erasure while ensuring safety through exposing a fixed set of factories.

Well at least I think so :) Would be good if others could reason this out too.

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/AttributeAssertion.java
```diff
@@ -0,0 +1,31 @@
...
+import com.google.auto.value.AutoValue;
+import io.opentelemetry.api.common.AttributeKey;
+import java.util.function.Consumer;
+import org.assertj.core.api.AbstractAssert;
+
+/** An assertion on an attribute key. */
+@AutoValue
+public abstract class AttributeAssertion {
+
+  // This method is not type-safe! But because the constructor is private, we know it will only be
```
anuraaga, line 17: The main purpose of having this class instead of using Map.Entry is to be able to make the constructor private

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/OpenTelemetryAssertions.java
```diff
@@ -111,6 +117,107 @@ public static EventDataAssert assertThat(EventData eventData) {
...
+  public static AttributeAssertion satisfies(
+      AttributeKey<List<Double>> key, DoubleListAssertConsumer assertion) {
+    return AttributeAssertion.create(key, assertion);
+  }
+
+  /**
+   * Returns an {@link AttributeAssertion} that asserts the given {@code key} is present with the
+   * given {@code value}.
+   */
+  public static <T> AttributeAssertion attribute(AttributeKey<T> key, T value) {
```
anuraaga, line 198: I had initially named `equals` but conflicts with `Test.equals()`, this or `equalTo` came to mind and this seemed better

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/SpanDataAssert.java
jack-berg: I think this tracks

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/OpenTelemetryAssertions.java
jack-berg: I slightly prefer equalTo, but don't feel too strong

jack-berg (approved):

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/OpenTelemetryAssertions.java
anuraaga: Yeah it does seem to correspond better with satisfies, switched to equalTo

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/SpanDataAssert.java
```diff
@@ -268,6 +272,57 @@ public SpanDataAssert hasAttributesSatisfying(Consumer<Attributes> attributes) {
...
+  // OpenTelemetryAssertions.sastisfies.
```
trask, line 306:
```suggestion
  // OpenTelemetryAssertions.satisfies.
```

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/SpanDataAssert.java
```diff
@@ -268,6 +272,57 @@ public SpanDataAssert hasAttributesSatisfying(Consumer<Attributes> attributes) {
...
+   */
+  public SpanDataAssert hasAttributesSatisfyingExactly(Iterable<AttributeAssertion> assertions) {
+    Set<AttributeKey<?>> actualKeys = actual.getAttributes().asMap().keySet();
+    Set<AttributeKey<?>> expectedKeys =
+        StreamSupport.stream(assertions.spliterator(), false)
+            .map(AttributeAssertion::getKey)
+            .collect(Collectors.toSet());
+    assertThat(actualKeys)
+        .as("span [%s] attribute keys", actual.getName())
+        .containsExactlyInAnyOrderElementsOf(expectedKeys);
```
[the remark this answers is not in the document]
anuraaga: Yeah I think that should be supported, should be fixed. Probably not in this PR but a followup we could consider adding `equalToIfPresent`, `satisfiesIfPresent`. The 100 in the instrumentation repo seem suspicious though, while a base class like HttpServerTest I can see it, but library-specific tests having so many seems like a possible copy-paste of incorrect practice.

PR #4178, opened 2022-02-16
Add Prometheus text format serializers.

anuraaga:
With metrics nearing stability, I think coupling the exporter to the prometheus client library has some risk as it may not make it in time. It's nice to also get some performance by directly serializing given how huge they can be (in particular, lots of arrays allocated for histograms in the adapter code that are eliminated). Many Java apps have a strong correlation between full GC and Prometheus scrapes :) I will move the `Collector` integration into an alpha extension instead so it can be decoupled from a stable release.

This only adds serialization logic, it doesn't use it yet or remove a few smaller usages of the client library that are left which will come in followups. 

Maintenance-wise, given the above benefits, I think 414 LoC of the serializer vs 340 LoC for the MetricAdapter is not that much more. Code is not quite as simple but prometheus format is relatively straight forward so doesn't seem too bad.

The output makes sure to match the client library 100% for openmetrics. It has some small deviations for legacy prometheus format which are compatible with prometheus, presumably the client library didn't mess with that format to avoid breaking unit tests relying on the strings.

### DIFF
--- a/exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/PrometheusType.java
+++ b/exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/PrometheusType.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.exporter.prometheus;
+
+import io.opentelemetry.sdk.metrics.data.AggregationTemporality;
+import io.opentelemetry.sdk.metrics.data.DoubleSumData;
+import io.opentelemetry.sdk.metrics.data.LongSumData;
+import io.opentelemetry.sdk.metrics.data.MetricData;
+
+// Four types we use are same in prometheus and openmetrics format
+enum PrometheusType {
+  GAUGE("gauge"),
+  COUNTER("counter"),
+  SUMMARY("summary"),
+  HISTOGRAM("histogram");
+
+  private final String typeString;
+
+  PrometheusType(String typeString) {
+    this.typeString = typeString;
+  }
+
+  static PrometheusType forMetric(MetricData metric) {
+    switch (metric.getType()) {
+      case LONG_GAUGE:
+      case DOUBLE_GAUGE:
+        return GAUGE;
+      case LONG_SUM:
+        LongSumData longSumData = metric.getLongSumData();
+        if (longSumData.isMonotonic()
+            && longSumData.getAggregationTemporality() == AggregationTemporality.CUMULATIVE) {
+          return COUNTER;
+        }
+        return GAUGE;
+      case DOUBLE_SUM:
+        DoubleSumData doubleSumData = metric.getDoubleSumData();
+        if (doubleSumData.isMonotonic()
+            && doubleSumData.getAggregationTemporality() == AggregationTemporality.CUMULATIVE) {
+          return COUNTER;
+        }
+        return GAUGE;
+      case SUMMARY:
+        return SUMMARY;
+      case HISTOGRAM:
+      case EXPONENTIAL_HISTOGRAM:
+        return HISTOGRAM;
+    }
+    throw new IllegalArgumentException(
+        "Unsupported metric type, this generally indicates version misalignment "
+            + "among opentelemetry dependencies. Please make sure to use opentelemetry-bom.");
+  }
+
+  String getTypeString() {
+    return typeString;
+  }
+}
--- a/exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/Serializer.java
+++ b/exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/Serializer.java
@@ -77,7 +77,7 @@ abstract class Serializer {
   }
 
   private void write(MetricData metric, Writer writer) throws IOException {
-    // TODO: Implement
+    // Not supported in specification yet.
     if (metric.getType() == MetricDataType.EXPONENTIAL_HISTOGRAM) {
       return;
     }
--- a/exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/Serializer.java
+++ b/exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/Serializer.java
@@ -1,0 +1,416 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// Includes work from:
+
+/*
+ * Prometheus instrumentation library for JVM applications
+ * Copyright 2012-2015 The Prometheus Authors
+ *
+ * This product includes software developed at
+ * Boxever Ltd. (http://www.boxever.com/).
+ *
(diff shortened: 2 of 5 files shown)


exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/Serializer.java
```diff
@@ -0,0 +1,416 @@
...
+        writer.write('0');
+      }
+      writer.write(Long.toString(millis));
+    }
+
+    @Override
+    void writeExemplar(
+        Writer writer, Collection<ExemplarData> exemplars, double minExemplar, double maxExemplar)
+        throws IOException {
+      for (ExemplarData exemplar : exemplars) {
```
jsuereth, line 385: I believe this code should also be writing dropped attributes as exemplar labels (after your trace_id/span_id code, with a limit on total size of labels written to prometheus).

anuraaga: Don't think we do that right now

https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/MetricAdapter.java#L327

Would like to keep this pr to just reproducing our current behavior and we can add features in the future

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/Serializer.java
```diff
@@ -0,0 +1,416 @@
...
+        new BufferedWriter(new OutputStreamWriter(output, StandardCharsets.UTF_8))) {
+      for (MetricData metric : metrics) {
+        write(metric, writer);
+      }
+      writeEof(writer);
+    }
+  }
+
+  private void write(MetricData metric, Writer writer) throws IOException {
+    // TODO: Implement
```
jsuereth, line 80: Can you open a bug for this one?

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/Serializer.java
```diff
@@ -0,0 +1,416 @@
...
+
+    List<ValueAtPercentile> valueAtPercentiles = point.getPercentileValues();
+    for (ValueAtPercentile valueAtPercentile : valueAtPercentiles) {
+      writePoint(
+          writer,
+          name,
+          valueAtPercentile.getValue(),
+          point.getAttributes(),
+          point.getEpochNanos(),
+          "quantile",
```
jsuereth, line 178: This a note for me, doubting my own code here.

Percentile != Quantile.  We had this issue in OpenCensus where we needed to switch from [0.0,1.0] to [0.0, 100.0].  


However, OTLP specifies summaries in `Quantile`.   Is this an instance where Java SDK is uses the wrong name?   I never noticed before, as I didn't pay close attention until working on some OpenCensus-Go bridges.

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/Serializer.java
```diff
@@ -0,0 +1,416 @@
...
+    try {
+      attributes.forEach(
+          new BiConsumer<AttributeKey<?>, Object>() {
+            private boolean wroteOne;
+
+            @Override
+            public void accept(AttributeKey<?> key, Object value) {
+              try {
+                if (wroteOne) {
+                  writer.write(',');
```
jsuereth, line 252: Would be nice to get test coverage for multi-attributes.

jsuereth (approved):

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/Serializer.java
jack-berg: [Wikipedia](https://en.wikipedia.org/wiki/Quantile) tells me that a percentile is a particular type of quantile, so is this a problem?

jack-berg (approved):

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/PrometheusType.java
```diff
@@ -0,0 +1,59 @@
...
+          return COUNTER;
+        }
+        return GAUGE;
+      case SUMMARY:
+        return SUMMARY;
+      case HISTOGRAM:
+      case EXPONENTIAL_HISTOGRAM:
+        return HISTOGRAM;
+    }
+    throw new IllegalArgumentException(
```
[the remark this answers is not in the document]
anuraaga: This can never be thrown when the BOM is used, if it did happen to be thrown then the export would fail. I've seen various linkage exceptions related to unaligned dependencies and think it's mostly not possible to use OTel without alignment so leant towards just making this an exception instead of log / fallback, but either approach would be fine here.

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/Serializer.java
anuraaga: I suspect this would be an issue in the spec repo. I changed this comment a bit to reflect this proposed spec wording

https://github.com/open-telemetry/opentelemetry-specification/pull/2266/files#diff-0efae13f08f98e62a81767d5daeff37ebb7ef8c50537c7b9013e72506e9b055aR1152

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/PrometheusType.java
```diff
@@ -0,0 +1,59 @@
...
+      case LONG_GAUGE:
+      case DOUBLE_GAUGE:
+        return GAUGE;
+      case LONG_SUM:
+        LongSumData longSumData = metric.getLongSumData();
+        if (longSumData.isMonotonic()
+            && longSumData.getAggregationTemporality() == AggregationTemporality.CUMULATIVE) {
+          return COUNTER;
+        }
+        return GAUGE;
```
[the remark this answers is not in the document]
anuraaga: I believe after we added preferred aggregation support it isn't possible when using our SDK. It could be possible if someone used the exporter without our SDK though (creating arbitrary MetricData).

exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/Serializer.java
anuraaga: I think @jsuereth is referring to the fact that the OTel quantile is not actually supposed to be that particular type of quantile. Spec says

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#summary-legacy

`within the interval [0.0, 1.0]`

We would at least need to change the name of the class to `ValueAtQuantile` and this obviously wrong javadoc

https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/ValueAtPercentile.java#L21

Would also need to make sure we don't have any logic such as validation that is using the percentile expectation

PR #4188, opened 2022-02-18
(4172) Vararg versions of AttributeBuilder.put

gcatanese:
@anuraaga Given the first shot at it, please let me if I understood correctly what you had in mind. Thanks


api/all/src/main/java/io/opentelemetry/api/common/AttributesBuilder.java
```diff
@@ -100,6 +101,18 @@ default AttributesBuilder put(String key, String... value) {
     return put(stringArrayKey(key), Arrays.asList(value));
   }
 
+  /**
+   * Puts a AttributeKey with String array into this.
+   *
+   * @return this Builder
+   */
+  default AttributesBuilder put(AttributeKey<List<String>> key, String... value) {
```
anuraaga, line 109: This looks mostly good - while I didn't verify with a compiler I think we need just one generic method for this issue

`<T> AttributesBuilder put(AttributeKey<List<T>> key, T... value)

If it doesn't compile we can just add the different versions

gcatanese: @anuraaga Yup, works with a generic type

gcatanese: I was too fast :-(

`<T> AttributesBuilder put(AttributeKey<T> key, T value)` and the new ` default <T> AttributesBuilder put(AttributeKey<List<T>> key, T... value)` collide when T is a List and value is null.

For example in ResourceTest the following doesn't compile
```
// ambiguous call  
attributes.put(stringArrayKey("NullArrayStringKey"), null);
```
I fiddled around but I cannot see how to prevent that. Any tip? Shall I opt for your other suggestion (method per type)?

anuraaga: I think this behavior is fine - `null` literal isn't an appropriate value for the value of either method anyways, so the fact that it requires a type cast, only in our unit test in practice, is fine

api/all/src/main/java/io/opentelemetry/api/common/AttributesBuilder.java
```diff
@@ -100,6 +101,19 @@ default AttributesBuilder put(String key, String... value) {
     return put(stringArrayKey(key), Arrays.asList(value));
   }
 
+  /**
+   * Puts a {@link AttributeKey} of generic List into this.
```
anuraaga, line 105:
```suggestion
   * Puts a List attribute into this.
```

anuraaga (approved):

jkwatson (approved):

kubawach (approved):
LGTM.

jack-berg (approved):

jkwatson:
all maintainers approve! Let's :shipit:

Issue #4189, opened 2022-02-18
OTLP: Implement marshaling of trace state in spans


knutwannheden:
I noticed that the OTLP marshaling of trace states in trace spans didn't work. Looking at the code I see that this is still a TODO, but I couldn't find any issue for it:

https://github.com/open-telemetry/opentelemetry-java/blob/044cadb47ec62a437ccb0ef4f8a4f99e3d1d7378/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/otlp/traces/SpanMarshaler.java#L116

and:

https://github.com/open-telemetry/opentelemetry-java/blob/044cadb47ec62a437ccb0ef4f8a4f99e3d1d7378/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/otlp/traces/SpanMarshaler.java#L155

With some guidance I could probably provide a PR.

PR #4191, opened 2022-02-18
Marshal trace state in OTLP exporter

knutwannheden:
Includes the trace state in the marshaled spans and span links, resolving the corresponding TODOs in the code.

Adds a static utility method to W3CTraceContextPropagator to serialize a `TraceState` to string (according to the W3C trace context recommendation.

Fixes: #4189

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/trace/propagation/W3CTraceContextPropagator.java
+++ b/api/all/src/main/java/io/opentelemetry/api/trace/propagation/W3CTraceContextPropagator.java
@@ -5,7 +5,8 @@
 
 package io.opentelemetry.api.trace.propagation;
 
-import static io.opentelemetry.api.internal.Utils.checkArgument;
+import static io.opentelemetry.api.trace.propagation.internal.W3CTraceContextEncoding.decodeTraceState;
+import static io.opentelemetry.api.trace.propagation.internal.W3CTraceContextEncoding.encodeTraceState;
 
 import io.opentelemetry.api.internal.OtelEncodingUtils;
 import io.opentelemetry.api.internal.TemporaryBuffers;
@@ -15,7 +16,6 @@ import io.opentelemetry.api.trace.SpanId;
 import io.opentelemetry.api.trace.TraceFlags;
 import io.opentelemetry.api.trace.TraceId;
 import io.opentelemetry.api.trace.TraceState;
-import io.opentelemetry.api.trace.TraceStateBuilder;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.propagation.TextMapGetter;
 import io.opentelemetry.context.propagation.TextMapPropagator;
@@ -27,7 +27,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.logging.Logger;
-import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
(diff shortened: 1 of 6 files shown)


api/all/src/main/java/io/opentelemetry/api/trace/propagation/W3CTraceContextPropagator.java
```diff
@@ -133,6 +133,14 @@ public static W3CTraceContextPropagator getInstance() {
       // No need to add an empty "tracestate" header.
       return;
     }
+    setter.set(carrier, TRACE_STATE, encodeTraceState(traceState));
+  }
+
+  /** Return the trace state encoded as a string according to the W3C specification. */
+  public static String encodeTraceState(TraceState traceState) {
```
anuraaga, line 140: Let's move this to a new class, `io.opentelemetry.api.trace.propagation.internal.W3CTraceContextEncoding`. We don't want to expose this as a public API for users, at least not yet.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/otlp/traces/SpanLinkMarshaler.java
```diff
@@ -37,21 +41,29 @@
 
   // Visible for testing
   static SpanLinkMarshaler create(LinkData link) {
+    TraceState traceState = link.getSpanContext().getTraceState();
+    byte[] traceStateUtf8 =
```
anuraaga, line 45: Let's shortcut a constant empty byte array when `traceState.isEmpty()`

exporters/otlp/common/src/test/java/io/opentelemetry/exporter/internal/otlp/traces/TraceRequestMarshalerTest.java
```diff
@@ -93,15 +103,20 @@ void toProtoResourceSpans() {
             InstrumentationLibrary.newBuilder().setName("testLib").setVersion("1.0").build());
   }
 
-  @Test
-  void toProtoSpan() {
+  static Stream<SpanContext> spanContexts() {
```
anuraaga, line 106: I don't think we need two tests let's just stick with adding trace state to the existing

exporters/otlp/common/src/test/java/io/opentelemetry/exporter/internal/otlp/traces/TraceRequestMarshalerTest.java
```diff
@@ -123,13 +138,14 @@ void toProtoSpan() {
...
                     .setTotalRecordedEvents(3)
-                    .setLinks(Collections.singletonList(LinkData.create(SPAN_CONTEXT)))
+                    .setLinks(Collections.singletonList(LinkData.create(spanContext)))
                     .setTotalRecordedLinks(2)
                     .setStatus(StatusData.ok())
                     .build()));
 
     assertThat(span.getTraceId().toByteArray()).isEqualTo(TRACE_ID_BYTES);
     assertThat(span.getSpanId().toByteArray()).isEqualTo(SPAN_ID_BYTES);
+    assertThat(span.getTraceState()).isEqualTo(encodeTraceState(spanContext.getTraceState()));
```
anuraaga, line 148: Let's put the actual trace state string inline, e.g. `"foo=bar,baz=qux".getBytes(UTF_8)"` so it's  more obvious the test is working

anuraaga (commented):
Thanks for the help @knutwannheden

api/all/src/main/java/io/opentelemetry/api/trace/propagation/W3CTraceContextPropagator.java
knutwannheden: Makes sense.

anuraaga (approved):
Thanks! Please sign the CLA so we can merge

knutwannheden:
I had some trouble with my local Docker setup, so I didn't notice the test failures. Fixing them now.

PR #4196, opened 2022-02-21
Add module-info to prometheus exporter

### LINKED ISSUE #4192: JPMS and prometheus exporter
From @laurit's https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5408#issue-1144660181

> Note that when using jpms user will need to use --add-modules jdk.httpserver so it would be best if prometheus exporter did not use these classes.

Short of documenting it (e.g. [here](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/README.md#prometheus-exporter)), wondering (as @laurit mentioned) if we can get rid of the dependency on `com.sun.net.httpserver`.

org.apache.hc.core5.http.impl.bootstrap.HttpServer in `org.apache.httpcomponents.core5:httpcore5` looks reasonable (actively developed, no transitive dependencies, 831 kb, may be worth shading into the prometheus exporter to avoid conflicts, especially since this is in an `impl` package doesn't give much confidence in API stability).

anuraaga:
Fixes #4192

With this adding module-info somewhere, it's probably best to go ahead and add it everywhere, will followup with that afterward. Currently it's only particularly important for prometheus and zpages which are using httpserver.

Tried using javamodularity plugin but don't think it works in toolchains mostly because of an old [TODO](https://github.com/java9-modularity/gradle-modules-plugin/blob/master/src/main/java/org/javamodularity/moduleplugin/extensions/DefaultModularityExtension.java#L74). It's approach of having two java compilation tasks output to the same directory also will probably break build caching. The alternative is a `HackForJpms.java` unfortunately.

When applying to all the projects, I will try out a DSL to define the module in the java file to let it autogenerate the hack to avoid exposing it.

### DIFF
--- a/exporters/prometheus/src/module/java/module-info.java
+++ b/exporters/prometheus/src/module/java/module-info.java
@@ -1,0 +1,7 @@
+@SuppressWarnings({"requires-automatic", "requires-transitive-automatic"})
+module io.opentelemetry.exporter.prometheus {
+  exports io.opentelemetry.exporter.prometheus;
+
+  requires transitive io.opentelemetry.sdk.metrics;
+  requires jdk.httpserver;
+}
--- a/exporters/prometheus/src/testJpms/java/io/opentelemetry/exporters/prometheus/test/JpmsTest.java
+++ b/exporters/prometheus/src/testJpms/java/io/opentelemetry/exporters/prometheus/test/JpmsTest.java
@@ -1,0 +1,22 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.exporters.prometheus.test;
+
+import io.opentelemetry.exporter.prometheus.PrometheusHttpServer;
+import io.opentelemetry.sdk.metrics.SdkMeterProvider;
+import org.junit.jupiter.api.Test;
+
+class JpmsTest {
+
+  @Test
+  void noLinkageErrors() {
+    SdkMeterProvider meterProvider =
+        SdkMeterProvider.builder()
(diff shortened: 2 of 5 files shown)


exporters/prometheus/src/testJpms/java/io/opentelemetry/exporters/prometheus/test/JpmsTest.java
```diff
@@ -0,0 +1,22 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.exporters.prometheus.test;
+
+import io.opentelemetry.exporter.prometheus.PrometheusHttpServer;
+import io.opentelemetry.sdk.metrics.SdkMeterProvider;
+import org.junit.jupiter.api.Test;
+
+class JpmsTest {
```
anuraaga, line 12: I think the long term approach will be to add an extra CI configuration running Java 17 with modules for all our normal tests instead of defining jpms-specific test sets. Went with this as an interim and to verify the fix.

exporters/prometheus/src/module/java/module-info.java
```diff
@@ -0,0 +1,7 @@
+@SuppressWarnings({"requires-automatic", "requires-transitive-automatic"})
+module io.opentelemetry.exporter.prometheus {
+  exports io.opentelemetry.exporter.prometheus;
+
+  requires transitive io.opentelemetry.sdk.metrics;
```
anuraaga, line 5: prometheus simpleclient doesn't define an automatic module name so doesn't work here, luckily that codepath is deprecated now

jkwatson (approved):

PR #4198, opened 2022-02-22
Remove unused Instrument interface.

anuraaga:
Not too sure what this was for but the javadoc seems to indicate a relation to the recently change to returning from async instruments.


jkwatson (approved):
thanks!

Issue #4201, opened 2022-02-22
DoubleExponentialHistogramBucketsTest extremely slow


anuraaga:
I haven't looked into detail but this test feels like it takes 1min+ to finish, which seems slower than expected.

jamesmoessis:
@anuraaga I think this can be closed

PR #4204, opened 2022-02-22
Fix slow ExponentialHistogram tests 

jamesmoessis:
Issue: https://github.com/open-telemetry/opentelemetry-java/issues/4201

The slow test is `diffAccumulation()` which was taking around 10 seconds on my machine. The problem is when the tests are comparing bucket counts between one empty and one non-empty bucket counts. The bucket set that is empty has `indexStart = Integer.MIN_VALUE` so comparing the buckets takes several seconds because it's iterating from such a low value. 

~~I've added the `isEmpty()` check to the beginning of the `sameBucketCounts()` method to fix this.~~

~~However, this brings a problem with `AdaptingCircularBufferArray`. In the case where an `AdaptingCircularBufferArray` has been `diff()`ed back down so it's empty, `isEmpty()` returns false because the index is no longer the null index. As a result I have added the O(maxSize) iteration into `isEmpty()` but I acknowledge it's less than ideal. I tried a few ways but I could never avoid adding an O(maxSize) check somewhere in `AdaptingCircularBufferArray`. I am kind of reluctant to slow down main code so test can be faster.~~

~~The loop I've added to `isEmpty()` returns within the first iteration in the vast majority of cases, because the lowest bucket will have non zero count almost always. So in reality it's probably negligible performance difference anyway.~~ 

@jsuereth I would appreciate your thoughts on this one :)

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleExponentialHistogramBuckets.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleExponentialHistogramBuckets.java
@@ -29,6 +29,7 @@ final class DoubleExponentialHistogramBuckets implements ExponentialHistogramBuc
   private ExponentialCounter counts;
   private int scale;
   private double scaleFactor;
+  private long totalCount;
 
   DoubleExponentialHistogramBuckets(
       int scale, int maxBuckets, ExponentialCounterFactory counterFactory) {
@@ -36,6 +37,7 @@ final class DoubleExponentialHistogramBuckets implements ExponentialHistogramBuc
     this.counts = counterFactory.newCounter(maxBuckets);
     this.scale = scale;
     this.scaleFactor = computeScaleFactor(scale);
+    this.totalCount = 0;
   }
 
   // For copying
@@ -44,6 +46,7 @@ final class DoubleExponentialHistogramBuckets implements ExponentialHistogramBuc
     this.counts = counterFactory.copy(buckets.counts);
     this.scale = buckets.scale;
     this.scaleFactor = buckets.scaleFactor;
+    this.totalCount = buckets.totalCount;
   }
 
   /** Returns a copy of this bucket. */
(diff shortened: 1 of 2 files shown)


anuraaga:
From what I can tell, we don't currently populate count when serializing

https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L478

I wonder if buckets should each keep track of their count (and sum)? Then isEmpty becomes just checking that

jamesmoessis:
@anuraaga that's a great point, we could keep track of count in the buckets. And if it isn't serialized into OTLP, that must be my mistake and I can fix that up. I will have a look into this today or tomorrow and push it to this PR

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/DoubleExponentialHistogramBuckets.java
```diff
@@ -306,7 +308,20 @@ public boolean equals(@Nullable Object obj) {
...
     int min = Math.min(this.counts.getIndexStart(), other.counts.getIndexStart());
+
+    // This check is so we avoid iterating from Integer.MIN_VALUE.
+    // In this case, we can assume that those buckets are empty.
+    // We start iterating from the other bucket index instead.
+    // They still may be equal as it is possible for another set of buckets
+    // to be empty but have a higher start index.
+    if (min == Integer.MIN_VALUE) {
+      min = Math.max(this.counts.getIndexStart(), other.counts.getIndexStart());
+    }
```
jamesmoessis, line 323: This problem kind of stems from `AdaptingCircularBufferCounter.isEmpty()` not returning correct result after a `diff()` that makes it empty. However I'm not sure if there's a way to fix it without putting in an _O(n)_ check somewhere in that class.  

This works well as a workaround and brings the `diffAccumulation()` test down from 16s to 0.002s.

jamesmoessis:
Alright @anuraaga I've written the code to keep track of the totalCount in the buckets. The OTLP serialisation already sends count so that's all good, but now the buckets memoize the count so it doesn't have to be calculated during the accumulation process.

jack-berg (approved):

anuraaga (approved):
Thanks!

PR #4205, opened 2022-02-22
Cleanup ExemplarData

anuraaga:
I deprecated in this PR but maybe we'll just delete anyways as there will probably be some harder changes coming up. 

- Record SpanContext instead of span / trace ID. Throughout the SDK we conventionally use invalid SpanContext to indicate not traced, and valid one to indicate traced. It seems to work well here, eliminating Nullable, and notably the cross-nullable relationship where either both are null or both are filled.

- Remove getValueAsDouble. We don't have it on e.g. SumData so this doesn't feel symmetric. getValueAsDouble seems to be primarily useful in prometheus conversion but such a utility can just live there. (Another location was the assertion but that should possibly be double / long specific like our other assertions). An alternative is to remove LongExemplarData, we don't wire it up to anything currently and I wonder if there is any plan for wiring it. If not, then it'd be much simpler to have only one for double.

### DIFF
--- a/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/otlp/metrics/ExemplarMarshaler.java
+++ b/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/otlp/metrics/ExemplarMarshaler.java
@@ -5,6 +5,7 @@
 
 package io.opentelemetry.exporter.internal.otlp.metrics;
 
+import io.opentelemetry.api.trace.SpanContext;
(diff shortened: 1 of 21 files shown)


jkwatson (approved):
Seems solid to me. Would love to hear from people more connected with exemplar usage, like @jsuereth

jack-berg (approved):

anuraaga:
@jsuereth Let me go ahead and merge this to keep on moving, if you get a chance to take a look later, happy to do any followups or even revert as needed, no worries.

PR #4207, opened 2022-02-22
Remove type-specific GaugeData

anuraaga:
This applies to the other instruments too which I will refactor after this one to confirm it makes sense.

I noticed there doesn't seem to be a need for type-specific versions of instrument data - while different instruments have different data / APIs, I don't think we can conceive of a need for this for the distinction between double / long. So sticking to the generic interface we already have works fine.

Also hides `ImmutableGaugeData.create()` for now - we will likely need to allow creation of metric data (unlike span data) when supporting custom aggregations but that is something we should reason out when the time comes.


jkwatson (approved):
Seems reasonable. @jack-berg @jsuereth ?

kubawach (approved):
I miss some context (why was this introduced in the first place?) but move to less complex data model is always welcome.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/MetricData.java
```diff
@@ -27,7 +28,7 @@ static MetricData createDoubleGauge(
       String name,
       String description,
       String unit,
-      DoubleGaugeData data) {
+      GaugeData<DoublePointData> data) {
```
jack-berg, line 31: This is a public API but the only implementation of `GaugeData<DoublePointData>` is `ImmutableGaugeData` which lives in an internal package. Can we move `ImmutableGaugeData` to `io.opentelemetry.sdk.metrics.data` and omit the "Immutable" prefix?

jack-berg (commented):
I agree with the idea in principal but have a comment.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/MetricData.java
anuraaga: Our general naming scheme has been `Immutable*` for the implementation of an interface as AutoValue. I think the class is used from the census bridge as well so I moved it to an internal package but think I can make it package private in the future.

anuraaga: Ah nevermind I got the brunt of the question after a reread. I think we'll be able to make all these factories internal until supporting custom aggregations so let me follow up with that. And renaming `Impl` to `Immutable*` per our other classes

jack-berg: > I think we'll be able to make all these factories internal until supporting custom aggregations so let me follow up with that.

This makes it much harder to make changes to metric data before exporting in a delegating metric exporter. A couple weeks back I proposed a [DelegatingMetricData](https://github.com/open-telemetry/opentelemetry-java/pull/4128#issuecomment-1026036078), but left it out before `MetricData` factories were public. 

I don't have a use case for this today myself, but I imagine someone will want to be able to change metrics before export. Suppose its probably best to publish stable with a minimal SDK API surface area, and let feature requests decide where we make additions.

PR #4213, opened 2022-02-24
Fixed minor typo in README regarding the alpha bom reference

mariusoe:
Fixing a typo in the readme where it states that the `opentelemetry-alpha-bom` dependency should be used for alpha dependencies. Correct would be `opentelemetry-bom-alpha`


anuraaga (approved):
Thanks!

jkwatson (approved):

Issue #4190, opened 2022-02-18
Add support for instrumentation library scope


jack-berg:
To reflect spec changes from [#2276](https://github.com/open-telemetry/opentelemetry-specification/pull/2276)

- [ ] Add `InstrumentationScopeInfo`
- [ ] Adjust `TracerProvider` docs to reflect that `name` is instrumentation scope name
- [ ] Adjust `MeterProvider` docs to reflect that `name` is instrumentation scope name
- [ ] Adjust `LogEmitterProvider` to reflect that `name` is instrumentation scope name
- [ ] Adjust `ReadableSpan`, `SpanData` to include accessor to `InstrumentationScopeInfo`, deprecate `#getInstrumentationLibraryInfo()`
- [ ] Adjust `MetricData` to include accessor to `InstrumentationScopeInfo`, deprecate `#getInstrumentationLibraryInfo()`
- [ ] Adjust `LogData` to include accessor to `InstrumentationScopeInfo`, deprecate `#getInstrumentationLibraryInfo()`
- [ ] Adjust zipkin exporter to export `otel.scope.name`, `otel.scope.version`
- [ ] Adjust jaeger exporter to export `otel.scope.name`, `otel.scope.version`

jkwatson:
for zipkin/jaeger, do we still export the old field name, too? Will that possibly break people?

jack-berg:
> for zipkin/jaeger, do we still export the old field name, too? Will that possibly break people?

Yes the spec says to export the old field name and the new name for compatibility

PR #4215, opened 2022-02-24
Instrumentation scope

jack-berg:
Related to #4190.

Solves the following items:
- Add InstrumentationScopeInfo
- Adjust TracerProvider docs to reflect that name is instrumentation scope name
- Adjust MeterProvider docs to reflect that name is instrumentation scope name
- Adjust LogEmitterProvider to reflect that name is instrumentation scope name

Updates `ComponentRegistry` to use new `InstrumentationScopeInfo` class internally, but `SpanData`, `MetricData`, and `LogData` still return `InstrumentationLibraryInfo`. Saving that for a subsequent PR.

### DIFF
--- a/sdk/common/src/main/java/io/opentelemetry/sdk/common/InstrumentationScopeInfo.java
+++ b/sdk/common/src/main/java/io/opentelemetry/sdk/common/InstrumentationScopeInfo.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.common;
+
+import static java.util.Objects.requireNonNull;
+
+import com.google.auto.value.AutoValue;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * Holds information about instrumentation scope.
+ *
+ * <p>Instrumentation scope is a logical unit of the application code with which emitted telemetry
+ * is associated. The most common approach is to use the instrumentation library as the scope,
+ * however other scopes are also common, e.g. a module, a package, or a class can be chosen as the
+ * instrumentation scope.
+ */
+@AutoValue
+@Immutable
+public abstract class InstrumentationScopeInfo {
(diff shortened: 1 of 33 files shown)


sdk/common/src/main/java/io/opentelemetry/sdk/common/InstrumentationScopeInfo.java
```diff
@@ -0,0 +1,60 @@
...
+import com.google.auto.value.AutoValue;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * Holds information about instrumentation scope.
+ *
+ * <p>Instrumentation scope is a logical unit of the application code with which emitted telemetry
+ * is associated. The most common approach is to use the instrumentation library as the scope,
+ * however other scopes are also common, e.g. a module, a package, or a class can be chosen as the
```
anuraaga, line 19:
```suggestion
 * however other scopes are also common, e.g. a module, a package, or a class may be chosen as the
```

sdk/common/src/main/java/io/opentelemetry/sdk/internal/InstrumentationScopeUtil.java
```diff
@@ -0,0 +1,27 @@
...
+package io.opentelemetry.sdk.internal;
+
+import io.opentelemetry.sdk.common.InstrumentationLibraryInfo;
+import io.opentelemetry.sdk.common.InstrumentationScopeInfo;
+
+/**
+ * This class is internal and is hence not for public use. Its APIs are unstable and can change at
+ * any time.
+ */
+public class InstrumentationScopeUtil {
```
anuraaga, line 15:
```suggestion
public final class InstrumentationScopeUtil {
```

anuraaga (approved):

sdk/common/src/main/java/io/opentelemetry/sdk/common/InstrumentationScopeInfo.java
anuraaga: Just to reinforce these aren't the only options

jkwatson (approved):

PR #4217, opened 2022-02-25
Move ExponentialHistogram data to internal package.

anuraaga:
We almost assuredly will mark metrics SDK stable without exponential histograms in the first version. So for now, we should move the code to an internal package.

I'm hoping not, but based on https://github.com/open-telemetry/opentelemetry-specification/pull/2304#discussion_r796054799 we may do similar for Exemplar if it's not finished for metrics SDK stable release. Optimistically leaving it in place for now.


jamesmoessis (approved):
LGTM, just not sure what the [.attach_pid10327](https://github.com/open-telemetry/opentelemetry-java/pull/4217/files#diff-64825ccbf3fd1468d30d4c2a7943ea8867b9aa5b510f3b98ff30c019eb5be847) file is. Is that supposed to be there?

jkwatson:
I approve..just need to put the "internal" scare text boilerplate on the public stuff moved in there.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/data/exponentialhistogram/ImmutableExponentialHistogramData.java
```diff
@@ -15,25 +16,32 @@
...
+  private static final ExponentialHistogramData EMPTY =
       ExponentialHistogramData.create(AggregationTemporality.CUMULATIVE, Collections.emptyList());
 
-  DoubleExponentialHistogramData() {}
+  public static ExponentialHistogramData empty() {
+    return EMPTY;
+  }
 
   /**
    * Create a DoubleExponentialHistogramData.
```
jkwatson, line 34: javadoc drift here

jkwatson (approved):
looks good. a bit of javadoc drift with the renames, but other than that, good to go.

PR #4218, opened 2022-02-28
Deprecate automatic histogram method for removal.

anuraaga:
As discussed at the SIG, we don't see a way to actually change the histogram that is returned by this method in the future, in which case it currently overlaps with `explicitBucketsHistogram`, and constrains us if we do end up wanting to encourage a different one, such as exponential, as the de-facto.

https://github.com/open-telemetry/opentelemetry-specification/issues/2382


jkwatson:
approved, but should we wait on some commentary from spec folks before merging?

anuraaga:
I suspect it's worth removing now and adding back later if for some reason it survives the spec but either works.

jkwatson (approved):

PR #4219, opened 2022-02-28
Extract HistogramData interface for DoubleHistogramData.

anuraaga:
Our data APIs are generally interfaces with autovalue private classes. This is already true of exponential histogram actually, this changes normal histogram to match the pattern. Also removes the word `Double` since histograms / summary don't have a double / long distinction in the OTLP data model. We could add Long versions later if we add it to OTLP later, it doesn't seem very likely.

https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/HistogramPointData.java
```diff
@@ -0,0 +1,44 @@
...
+  List<Double> getBoundaries();
+
+  /**
+   * The counts in each bucket.
+   *
+   * @return the read-only counts in each bucket. <b>do not mutate</b> the returned object.
+   */
+  List<Long> getCounts();
+
+  double getBucketLowerBound(int bucketIndex);
```
jkwatson, line 41: javadoc needed here and on the upper bound method

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/data/ImmutableHistogramData.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.Collections;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * A histogram metric point.
+ *
+ * <p>See:
+ * https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#histogram
+ *
+ * <p><i>Note: This is called "DoubleHistogram" to reflect which primitives are used to record it,
```
jkwatson, line 22: javadoc drift here

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/data/ImmutableHistogramPointData.java
```diff
@@ -19,7 +21,7 @@
  */
 @Immutable
 @AutoValue
-public abstract class DoubleHistogramPointData implements PointData {
+public abstract class ImmutableHistogramPointData implements HistogramPointData {
```
jkwatson, line 24: javadoc drift in this class as well.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/HistogramData.java
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.metrics.data;
+
+import java.util.Collection;
+
+public interface HistogramData extends Data<HistogramPointData> {
```
jkwatson, line 10: class level javadoc would be good here, since it's a public class

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/HistogramPointData.java
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.metrics.data;
+
+import java.util.List;
+
+public interface HistogramPointData extends PointData {
```
jkwatson, line 10: class-level javadoc needed here, as well

jkwatson (approved):

jsuereth:
For context, we used to have Long/Double histograms but decided to remove them for Histograms, so it's very unlikely we add it back.  The Java DataModel classes predate that decision IIRC.

Issue #4161, opened 2022-02-08
AbstractMethodError crash on Android 


ber4444:
**Describe the bug**
 ```
java.lang.AbstractMethodError: abstract method "void io.opentelemetry.api.common.Attributes.forEach(j$.util.function.BiConsumer)"
at io.opentelemetry.exporter.zipkin.ZipkinSpanExporter.generateSpan(SourceFile:39)
02-07 16:14:48.698 13877 13916 E AndroidRuntime:        at io.opentelemetry.exporter.zipkin.ZipkinSpanExporter.export(SourceFile:3)
02-07 16:14:48.698 13877 13916 E AndroidRuntime:        at na.g0$c.export(SourceFile:1)
02-07 16:14:48.698 13877 13916 E AndroidRuntime:        at na.j.export(SourceFile:10)
02-07 16:14:48.698 13877 13916 E AndroidRuntime:        at na.s0.export(SourceFile:18)
02-07 16:14:48.698 13877 13916 E AndroidRuntime:        at na.l0.export(SourceFile:15)
02-07 16:14:48.698 13877 13916 E AndroidRuntime:        at io.opentelemetry.sdk.trace.export.BatchSpanProcessor$Worker.exportCurrentBatch(SourceFile:3)
02-07 16:14:48.698 13877 13916 E AndroidRuntime:        at io.opentelemetry.sdk.trace.export.BatchSpanProcessor$Worker.run(SourceFile:8)
02-07 16:14:48.698 13877 13916 E AndroidRuntime:        at java.lang.Thread.run(Thread.java:919)
```

**Steps to reproduce**
Run sample Android app with the Splunk SDK that uses OpenTelemetry. 

**What did you expect to see?**
No crash

**What did you see instead?**
Crashes as the Splunk SDK initializes.

**What version and what artifacts are you using?**
Artifacts: https://github.com/signalfx/splunk-otel-android
Version: com.splunk:splunk-otel-android:0.12.0, seems to be using opentelemetry 1.6.0
How did you reference these artifacts? [(excerpt from your `build.gradle`, `pom.xml`, etc)](https://github.com/signalfx/splunk-otel-android/blob/97bdf58966bf878f555823ad0bb5ad4b0f3caa44/splunk-otel-android/build.gradle#L46)

**Environment**
OS: Android 10
[...]

PR #4221, opened 2022-02-28
Fix android desugaring for HashMap.forEach

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java/issues/4161
Android desugaring is not smart enough to handle calling `HashMap.forEach` through `Attributes.forEach`. This pr helps out desugaring by adding an explicit call to `HashMap.forEach` that desugaring can rewrite it.


jkwatson (approved):
Should we log a bug against the Android Gradle Plugin for this case explicitly? Or, can you add your findings to https://issuetracker.google.com/issues/202188674

anuraaga (approved):
Thanks!

PR #4222, opened 2022-02-28
Implement metric identity specification

jack-berg:
Implements changes merged in spec PR [#2317](https://github.com/open-telemetry/opentelemetry-specification/pull/2317).

In summary:
- Each instrument is identified by its name, description, unit, type, and value type. 
- If you ask a meter for an instrument with the same identity as one previously registered, you'll record to the same instrument.
- If you ask a meter for an instrument the same name as one previously registered but some difference in description, unit, type, or value type, you've created an identity conflict. Multiple metrics with the same name will be exported, and we'll log a diagnostic error message.
- Views also play a role in identity. Since you can use views to change the name, description, and aggregation, its fairly easy to create an identity conflict when you register multiple views that select the same instrument(s).

### DIFF
--- a/sdk/metrics/src/debugEnabledTest/java/io/opentelemetry/sdk/metrics/internal/descriptor/InstrumentDescriptorTest.java
+++ b/sdk/metrics/src/debugEnabledTest/java/io/opentelemetry/sdk/metrics/internal/descriptor/InstrumentDescriptorTest.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.metrics.internal.descriptor;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.opentelemetry.sdk.metrics.common.InstrumentType;
+import io.opentelemetry.sdk.metrics.common.InstrumentValueType;
+import org.junit.jupiter.api.Test;
+
+/**
+ * {@link InstrumentDescriptor#equals(Object)} must ignore {@link
+ * InstrumentDescriptor#getSourceInfo()}, which only returns a meaningful value when {@code
+ * otel.experimental.sdk.metrics.debug=true}.
+ */
+class InstrumentDescriptorTest {
+
+  @Test
+  void equals() {
+    InstrumentDescriptor descriptor =
+        InstrumentDescriptor.create(
+            "name", "description", "unit", InstrumentType.COUNTER, InstrumentValueType.LONG);
+
+    assertThat(descriptor)
+        .isEqualTo(
+            InstrumentDescriptor.create(
+                "name", "description", "unit", InstrumentType.COUNTER, InstrumentValueType.LONG));
+
+    // Validate getSourceInfo() is not equal for otherwise equal descriptors
+    assertThat(descriptor.getSourceInfo())
+        .isNotEqualTo(
+            InstrumentDescriptor.create(
+                    "name", "description", "unit", InstrumentType.COUNTER, InstrumentValueType.LONG)
+                .getSourceInfo());
+
+    // Validate that name, description, unit, type, and value type are considered in equals
+    assertThat(descriptor)
+        .isNotEqualTo(
+            InstrumentDescriptor.create(
+                "foo", "description", "unit", InstrumentType.COUNTER, InstrumentValueType.LONG));
+    assertThat(descriptor)
+        .isNotEqualTo(
+            InstrumentDescriptor.create(
+                "name", "foo", "unit", InstrumentType.COUNTER, InstrumentValueType.LONG));
+    assertThat(descriptor)
+        .isNotEqualTo(
+            InstrumentDescriptor.create(
+                "name", "description", "foo", InstrumentType.COUNTER, InstrumentValueType.LONG));
+    assertThat(descriptor)
+        .isNotEqualTo(
+            InstrumentDescriptor.create(
+                "name",
+                "description",
+                "unit",
+                InstrumentType.UP_DOWN_COUNTER,
+                InstrumentValueType.LONG));
+    assertThat(descriptor)
+        .isNotEqualTo(
+            InstrumentDescriptor.create(
+                "name", "description", "unit", InstrumentType.COUNTER, InstrumentValueType.DOUBLE));
+  }
+}
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/descriptor/MetricDescriptor.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/descriptor/MetricDescriptor.java
@@ -11,7 +11,6 @@ import io.opentelemetry.sdk.metrics.common.InstrumentType;
 import io.opentelemetry.sdk.metrics.common.InstrumentValueType;
 import io.opentelemetry.sdk.metrics.view.View;
 import java.util.Objects;
-import java.util.Optional;
 import javax.annotation.concurrent.Immutable;
 
 /**
@@ -27,7 +26,7 @@ import javax.annotation.concurrent.Immutable;
 public abstract class MetricDescriptor {
 
   /**
-   * Constructs a metric descriptor with no source instrument/view.
+   * Constructs a metric descriptor with no instrument and default view.
    *
    * <p>Used for testing + empty-storage only.
    */
@@ -35,8 +34,7 @@ public abstract class MetricDescriptor {
     return new AutoValue_MetricDescriptor(
         name,
         description,
-        unit,
-        Optional.empty(),
+        View.builder().build(),
         InstrumentDescriptor.create(
             name, description, unit, InstrumentType.OBSERVABLE_GAUGE, InstrumentValueType.DOUBLE));
   }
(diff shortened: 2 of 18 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/descriptor/MetricDescriptor.java
```diff
@@ -46,21 +44,32 @@ public static MetricDescriptor create(View view, InstrumentDescriptor instrument
...
+   */
   public abstract String getName();
 
+  /**
+   * The description of the descriptor, equal to {@link View#getDescription()} if not null, else
+   * {@link InstrumentDescriptor#getDescription()}.
+   */
   public abstract String getDescription();
 
-  public abstract String getUnit();
```
jack-berg, line 57: No reason for `getUnit()` to be a top level field since views can't modify it.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/descriptor/MetricDescriptor.java
```diff
@@ -46,21 +44,32 @@ public static MetricDescriptor create(View view, InstrumentDescriptor instrument
...
+  /** The view that lead to the creation of this metric. */
+  public abstract View getSourceView();
 
-  /** The view that lead to the creation of this metric, if applicable. */
-  public abstract Optional<View> getSourceView();
```
jack-berg, line 60: No reason for `getSourceView()` to be optional. Getting rid of optional simplifies code.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/SdkMeterTest.java
```diff
@@ -51,12 +51,7 @@ void testLongCounter() {
     assertThat(logs.getEvents()).isEmpty();
 
     sdkMeter.counterBuilder("testLongCounter").build();
-    assertThat(
-            logs.assertContains(
-                    loggingEvent -> loggingEvent.getLevel().equals(WARN),
-                    "Failed to register metric.")
-                .getThrowable())
-        .hasMessageContaining("Metric with same name and different descriptor already created.");
+    logs.assertContains("Found duplicate metric definition");
```
jack-berg, line 54: A lot of the tests in this code duplicate the test cases in `IdentityTest`, but with less complete validation. I think I should remove the duplicate scenarios to reduce maintenance burden.

sdk/metrics/src/debugEnabledTest/java/io/opentelemetry/sdk/metrics/internal/descriptor/InstrumentDescriptorTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * {@link InstrumentDescriptor#equals(Object)} must ignore {@link
+ * InstrumentDescriptor#getSourceInfo()}, which only returns a meaningful value when {@code
+ * otel.experimental.sdk.metrics.debug=true}.
+ */
+class InstrumentDescriptorTest {
+
+  @Test
+  void equals() {
```
anuraaga, line 22: Is it possible to use equalsverifier?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/descriptor/MetricDescriptor.java
```diff
@@ -46,21 +44,32 @@ public static MetricDescriptor create(View view, InstrumentDescriptor instrument
...
   /** The instrument which lead to the creation of this metric. */
   public abstract InstrumentDescriptor getSourceInstrument();
 
+  /** The FQCN of the view aggregation. */
```
anuraaga, line 68: How about using Class directly instead of String?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/descriptor/MetricDescriptor.java
```diff
@@ -73,15 +82,22 @@ public static MetricDescriptor create(View view, InstrumentDescriptor instrument
...
+   *   <li>{@link #getAggregationName()} is equal
+   *   <li>{@link InstrumentDescriptor#getName()} is equal
+   *   <li>{@link InstrumentDescriptor#getDescription()} is equal
+   *   <li>{@link InstrumentDescriptor#getUnit()} is equal
    *   <li>{@link InstrumentDescriptor#getType()} is equal
    *   <li>{@link InstrumentDescriptor#getValueType()} is equal
    * </ul>
    */
   public boolean isCompatibleWith(MetricDescriptor other) {
     return Objects.equals(getName(), other.getName())
```
anuraaga, line 94: Would be nice to not use `Objects.equals` for not null methods

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/descriptor/MetricDescriptor.java
anuraaga: Though otherwise if we already have the concept of a human-readable aggregation name for viewconfig, we should try to use that too

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MeterSharedState.java
```diff
@@ -146,24 +141,10 @@ public final WriteableMetricStorage registerSynchronousMetricStorage(
         new ArrayList<>();
     for (AsynchronousMetricStorage<?, ObservableDoubleMeasurement> storage : storages) {
       AsynchronousMetricStorage<?, ObservableDoubleMeasurement> registeredStorage =
-          register(storage);
-      if (registeredStorage != null) {
```
anuraaga, line 150: Nice simplification!

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageRegistry.java
```diff
@@ -5,71 +5,88 @@
...
  */
 public class MetricStorageRegistry {
-  // TODO: Maybe we store metrics *and* instrument interfaces separately here...
-  private final ConcurrentMap<String, MetricStorage> registry = new ConcurrentHashMap<>();
+  private static final Logger logger = Logger.getLogger(MetricStorageRegistry.class.getName());
+
+  private final Object lock = new Object();
+
+  @GuardedBy("lock")
+  private final ConcurrentMap<MetricDescriptor, MetricStorage> registry = new ConcurrentHashMap<>();
```
anuraaga, line 38: Can be normal `HashMap` since there's a lock now

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageRegistry.java
```diff
@@ -5,71 +5,88 @@
...
 
   /**
    * Returns a {@code Collection} view of the registered {@link MetricStorage}.
    *
    * @return a {@code Collection} view of the registered {@link MetricStorage}.
    */
   public Collection<MetricStorage> getMetrics() {
-    return Collections.unmodifiableCollection(new ArrayList<>(registry.values()));
+    synchronized (lock) {
+      return Collections.unmodifiableCollection(new ArrayList<>(registry.values()));
```
anuraaga, line 47: Old code so no worries, but noticed that the javadoc says "view" even though we copy

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageRegistry.java
```diff
@@ -5,71 +5,88 @@
...
+    for (MetricStorage storage : storages) {
+      // Skip the newly registered storage
+      if (storage == newStorage) {
+        continue;
+      }
+      MetricDescriptor existing = storage.getMetricDescriptor();
+      // Check compatibility of metrics which share the same case-insensitive name
+      if (existing.getName().equalsIgnoreCase(descriptor.getName())
+          && !existing.isCompatibleWith(descriptor)) {
+        logger.log(Level.WARNING, DebugUtils.duplicateMetricErrorMessage(existing, descriptor));
```
anuraaga, line 85: Let's use the `Supplier` version of `logging.log`

sdk/metrics/src/debugEnabledTest/java/io/opentelemetry/sdk/metrics/internal/descriptor/InstrumentDescriptorTest.java
jack-berg: I can't figure out how to make it work with an abstract class implemented by autovalue.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageRegistry.java
jack-berg: Android doesn't support the supplier version. We do exit early if `WARNING` isn't loggable.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/AggregationUtil.java
```diff
@@ -0,0 +1,77 @@
...
+   * @throws IllegalArgumentException if the aggregation is not recognized
+   */
+  public static String aggregationName(Aggregation aggregation) {
+    String name = nameByAggregation.get(aggregation.getClass());
+    if (name == null) {
+      throw new IllegalStateException(
+          "Unrecognized aggregation " + aggregation.getClass().getName());
+    }
+    return name;
+  }
```
jack-berg, line 76: @anuraaga this is my solution for being able to translate aggregations to / from human readable names. Its in an internal class so we're not obligated to support it, but it provides a central place for the conversion that both the view file config and `MetricDescriptor#getAggregationName()` can depend on.

jack-berg, line 76: @anuraaga this is my solution for being able to translate aggregations to / from human readable names. Its in an internal class so we're not obligated to support it, but it provides a central place for the conversion that both the view file config and `MetricDescriptor#getAggregationName()` can depend on.

anuraaga (approved):

PR #4223, opened 2022-02-28
Add scare text to the javadoc for GlobalOpenTelemetry

### LINKED ISSUE #4177: Missing documentation about GlobalOpenTelemetry usage
**Describe the bug**
As mentioned in https://github.com/open-telemetry/opentelemetry-java/issues/4110#issuecomment-1020287705 and https://cloud-native.slack.com/archives/C014L2KCTE3/p1644503120757649?thread_ts=1644485274.293099&cid=C014L2KCTE3, `GlobalOpenTelemetry` _should not be used in applications unless you have a use case that makes impossible to avoid_.

For clarity that behavior and such extreme cases should be documented 

**What did you see instead?**
There is no such documentation about it's preferred usage

Exists in `opentelemetry-api` `1.10.1`

jkwatson:
Resolves #4177

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
+++ b/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
@@ -27,7 +27,12 @@ import javax.annotation.concurrent.ThreadSafe;
  * <p>If using the OpenTelemetry SDK, you may want to instantiate the {@link OpenTelemetry} to
  * provide configuration, for example of {@code Resource} or {@code Sampler}. See {@code
  * OpenTelemetrySdk} and {@code OpenTelemetrySdk.builder} for information on how to construct the
- * SDK {@link OpenTelemetry}.
+ * SDK's {@link OpenTelemetry} implementation.
+ *
+ * <p>WARNING: Due to the inherent complications around initialization order involving this class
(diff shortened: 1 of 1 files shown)


api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java
```diff
@@ -27,7 +27,12 @@
  * <p>If using the OpenTelemetry SDK, you may want to instantiate the {@link OpenTelemetry} to
  * provide configuration, for example of {@code Resource} or {@code Sampler}. See {@code
  * OpenTelemetrySdk} and {@code OpenTelemetrySdk.builder} for information on how to construct the
- * SDK {@link OpenTelemetry}.
+ * SDK's {@link OpenTelemetry} implementation.
+ *
+ * <p>WARNING: Due to the inherent complications around initialization order involving this class
+ * and its single global instance, we strongly recommend *not* using GlobalOpenTelemetry unless you
+ * have a use-case that absolutely requires it. Please favor using instances of OpenTelemetry
```
anuraaga, line 34: Let's add something like "if you are using the javaagent, it is generally best to only call GlobalOpenTelemetry.get() once and pass that reference to locations that use it, often by providing it through dependency injection, e.g. defining a Spring bean."

anuraaga (approved):

trask (approved):

PR #4231, opened 2022-03-03
feat: Add enabled resource providers configuration

### LINKED ISSUE #4220: Add otel.java.enabled.resource.providers config key
In my case, I only need a few resource fields while so many resource fields are auto configured. More resource fields mean more storage space, and these fields don't care for my scenario.
I can only disable these auto configured resource fields by setting `otel.java.disabled.resource.providers` config. However there are too many `ResourceProvider` implementations I want to disable.

I want to add a config key `otel.java.enabled.resource.providers` that just configure `ResourceProvider` what I want to enable. Enabled resource providers configuration takes precedence over disabled configuration.
If the enabled config key is not configured, all resource providers will be enabled.

wallezhang:
Add a new configuration: `otel.java.enabled.resource.providers`. 

All resource providers will be enabled if this configuration key is not configured, otherwise only the specific resource providers will be enabled.

If a resource provider is configured in both `otel.java.enabled.resource.providers` and `otel.java.disabled.resource.providers`, then it will be disabled.

close #4220

### DIFF
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ResourceConfiguration.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ResourceConfiguration.java
@@ -31,6 +31,8 @@ final class ResourceConfiguration {
       BiFunction<? super Resource, ConfigProperties, ? extends Resource> resourceCustomizer) {
     Resource result = Resource.getDefault();
 
+    Set<String> enabledProviders = new HashSet<>(
+        config.getList("otel.java.enabled.resource.providers"));
     // TODO(anuraaga): We use a hyphen only once in this artifact, for
     // otel.java.disabled.resource-providers. But fetching by the dot version is the simplest way
     // to implement it for now.
@@ -38,6 +40,10 @@ final class ResourceConfiguration {
         new HashSet<>(config.getList("otel.java.disabled.resource.providers"));
     for (ResourceProvider resourceProvider :
         ServiceLoader.load(ResourceProvider.class, serviceClassLoader)) {
+      if (!enabledProviders.isEmpty() && !enabledProviders.contains(
+          resourceProvider.getClass().getName())) {
+        continue;
+      }
       if (disabledProviders.contains(resourceProvider.getClass().getName())) {
         continue;
       }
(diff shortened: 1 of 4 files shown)


wallezhang:
TBR

sdk-extensions/autoconfigure/src/testFullConfig/java/io/opentelemetry/sdk/autoconfigure/ResourceConfigurationTest.java
```diff
@@ -46,6 +49,27 @@ void emptyClassLoader() {
                 (r, c) -> r)
             .getAttributes();
 
+    assertProcessAttributeIsNull(attributes);
+  }
+
+  @Test
+  void onlyEnabledCustomResourceProvider() {
```
jkwatson, line 56: I think it would be good to have a unit test to verify the behavior when things are specified in both lists, especially when the same provider is in both lists.

wallezhang: Sounds reasonable, I will add the missing unit tests.

jkwatson (approved):
Thanks!

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ResourceConfiguration.java
```diff
@@ -31,13 +31,19 @@ static Resource configureResource(
       BiFunction<? super Resource, ConfigProperties, ? extends Resource> resourceCustomizer) {
     Resource result = Resource.getDefault();
 
+    Set<String> enabledProviders =
+        new HashSet<>(config.getList("otel.java.enabled.resource.providers"));
```
anuraaga, line 35: Should we validate that only one of these is set as it would probably result in confusing behavior?

anuraaga (approved):

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/ResourceConfiguration.java
wallezhang: I think the only missing case should be only setting the `otel.java.enabled.resource.providers` configuration. Only setting `` configuration is validated in other test case.

PR #4239, opened 2022-03-07
Extract View configuration interface so it only exposes public API.

anuraaga:
The primary objective of this PR is to make sure getSourceInfo and getAttributeProcessor are not public (requires `View` to be an interface). A secondary objective is to allow adding experimental options such as baggage in an internal-only way (requires `ViewBuilder` to be an interface).


jkwatson (approved):
Seems reasonable to me. 👍🏽

jack-berg (approved):

PR #4241, opened 2022-03-07
Inline MeterSelector into InstrumentSelector

anuraaga:
A strawmans proposal, don't have to merge it but I feel like the extra class makes the API more complicated without much user benefit.


jkwatson:
This seems pretty reasonable to me. Would love thoughts from @jack-berg and @jsuereth if possible.

jack-berg (approved):

PR #4242, opened 2022-03-07
Migrate PointData to interface

anuraaga:


jkwatson (approved):
Thanks

exporters/logging-otlp/src/test/java/io/opentelemetry/exporter/logging/otlp/OtlpJsonLoggingMetricExporterTest.java
```diff
@@ -13,9 +13,9 @@
 import io.opentelemetry.internal.testing.slf4j.SuppressLogger;
 import io.opentelemetry.sdk.common.InstrumentationLibraryInfo;
 import io.opentelemetry.sdk.metrics.data.AggregationTemporality;
-import io.opentelemetry.sdk.metrics.data.DoublePointData;
 import io.opentelemetry.sdk.metrics.data.MetricData;
 import io.opentelemetry.sdk.metrics.export.MetricExporter;
+import io.opentelemetry.sdk.metrics.internal.data.ImmutableDoublePointData;
```
jsuereth, line 18: Just catching up on things here.  I'm not sure this move was best for implementors of SDK exporters.  Right now, we are hiding all the concrete class (but also the constructors) for SDK data, which means either:

1. Downstream users need to create their own data hierarchy for testing.
2. Downstream users need to depend on internal construction mechanisms.

We should definitely provide constructors of these data points at a MINIMUM for users to test their own SDK extensions.

anuraaga: Yeah we still need to introduce a testing version of data construction similar to what we have for SpanData, probably with builders to avoid the long factory methods. For most cases though do the assertj helpers work?

PR #4245, opened 2022-03-08
Option for increasing BatchSpanProcessor throughput under slow network / ingestion

trask:
I'm guessing you would want this option to go through spec(?), but wanted to get some initial feedback here ~, including the alternate option (see #4246).~

The problem: under high telemetry load and slower network / ingestion, the BatchSpanProcessor single worker thread can spend a lot of time waiting for responses.

This PR: this option tries to send as many complete batches as possible (up to a new config option `maxConcurrentExports`), and then waits `exporterTimeoutNanos` for all of those batches to complete.

### DIFF
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
@@ -332,13 +332,14 @@ public final class BatchSpanProcessor implements SpanProcessor {
 
       try {
         CompletableResultCode result = spanExporter.export(Collections.unmodifiableList(batch));
-        result.whenComplete(() -> {
-          if (result.isSuccess()) {
-            processedSpansCounter.add(batch.size(), exportedAttrs);
-          } else {
-            logger.log(Level.FINE, "Exporter failed");
-          }
-        });
+        result.whenComplete(
+            () -> {
+              if (result.isSuccess()) {
+                processedSpansCounter.add(batch.size(), exportedAttrs);
+              } else {
+                logger.log(Level.FINE, "Exporter failed");
+              }
+            });
         return result;
       } catch (RuntimeException e) {
         logger.log(Level.WARNING, "Exporter threw an Exception", e);
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
@@ -20,6 +20,7 @@ import io.opentelemetry.sdk.trace.data.SpanData;
 import io.opentelemetry.sdk.trace.internal.JcTools;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.List;
 import java.util.Queue;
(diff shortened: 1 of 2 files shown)


anuraaga:
LGTM having this option for batch processor seems like it would be nice

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -231,11 +237,19 @@ public void run() {
...
         }
-        while (!queue.isEmpty() && batch.size() < maxExportBatchSize) {
+        List<CompletableResultCode> concurrentExports = new ArrayList<>();
+        while (!queue.isEmpty() && concurrentExports.size() < maxConcurrentExports) {
           batch.add(queue.poll().toSpanData());
+          if (batch.size() >= maxExportBatchSize) {
+            concurrentExports.add(exportCurrentBatch());
+          }
+        }
+        if (concurrentExports.isEmpty() && System.nanoTime() >= nextExportTime) {
```
jkwatson, line 247: what case is this trying to catch? It would only trigger if somehow the queue were empty, but the time had expired...but the queue was just empty! Why would you expect there to be something to export at this point?

jkwatson: yeah.. we talked about this last week. I asked @trask to add a comment about it, since it wasn't obvious to me what this was for.

trask:
Closing.

@jkwatson had the keen recollection that the spec explicitly [disallows calling the exporter concurrently](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#exportbatch):

> Export() will never be called concurrently for the same exporter instance. Export() can be called again only after the current call returns.

@jkwatson also had the great workaround to have the exporter always return `CompletableResultCode.ofSuccess()`, and then flush the exporter after flushing the `BatchSpanProcessor`, which could even be written as a wrapper around an existing exporter, e.g.

```
public class ConcurrentSpanExporter implements SpanExporter {

  private final SpanExporter delegate;
  private final int maxConcurrentExports;

  private final Set<CompletableResultCode> activeExportResults =
      Collections.newSetFromMap(new ConcurrentHashMap<>());

  public ConcurrentSpanExporter(SpanExporter delegate, int maxConcurrentExports) {
    this.delegate = delegate;
    this.maxConcurrentExports = maxConcurrentExports;
  }

  @Override
    return maybeAddToActiveExportResults(delegate.export(spans));
  }

  private CompletableResultCode maybeAddToActiveExportResults(CompletableResultCode result) {
    if (activeExportResults.size() >= maxConcurrentExports) {
      return result;
    }
    activeExportResults.add(result);
    result.whenComplete(() -> activeExportResults.remove(result));
  }

  @Override
  public CompletableResultCode flush() {
    return CompletableResultCode.ofAll(activeExportResults);
  }

  @Override
    return delegate.shutdown();
  }
}
```

Btw, not sure if it makes sense for the `BatchSpanProcessor` to officially support span exporters that don't return a real `CompletableResultCode`, e.g. by calling `flush` on the SpanExporter after it flushes the worker. I'll open an issue to discuss/track.

PR #4246, opened 2022-03-08
(Alternate) Option for increasing BatchSpanProcessor throughput under slow network / ingestion

trask:
I'm guessing you would want this option to go through spec(?), but wanted to get some initial feedback here, including the alternate option (see #4245).

I prefer this option, I call this option "use the single worker thread as much as possible".

This option conflicts with `exporterTimeoutNanos`, though I think that it makes more sense for downstream exporters to be responsible for their own timeout behavior anyways (e.g. `OTEL_EXPORTER_OTLP_TIMEOUT`).

With more complexity, it could probably support `exporterTimeoutNanos` as well, but I would probably prefer the other PR in that case from a simplicity perspective.

### DIFF
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
@@ -20,6 +20,7 @@ import io.opentelemetry.sdk.trace.data.SpanData;
 import io.opentelemetry.sdk.trace.internal.JcTools;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.List;
 import java.util.Queue;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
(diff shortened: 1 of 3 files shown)


anuraaga:
@trask For reference, why do you prefer this PR? It seems quite similar to the other one but has an additional Semaphore to reason about

trask:
> @trask For reference, why do you prefer this PR? It seems quite similar to the other one but has an additional Semaphore to reason about

This PR maximizes throughput of the single worker thread (with adequate `maxConcurrentExports`), where the single worker thread will be continuously busy serializing/writing requests over the wire, and never blocking on responses.

With this PR, I can say that the only way to get higher throughput from a single worker thread is to make the exporter's serialization faster.

The other PR still pauses the single worker thread after each set of batches to wait for responses, so it has less throughput (and if it encounters a sporadically slow response, could be significantly less).

trask:
discussed with @anuraaga, closing this in favor of #4245

even though having the timeout in two places (BSP and Exporter) is confusing, that 🚢 has sailed

> With more complexity, it could probably support `exporterTimeoutNanos` as well, but I would probably prefer the other PR in that case from a simplicity perspective.

this is always a possibility in the future, but doesn't seem worth the complexity for now

PR #4254, opened 2022-03-09
Convert ValueAtPercentile to ValueAtQuantile

### LINKED ISSUE #4206: Change ValueAtPercentile to ValueAtQuantile
I think @jsuereth is referring to the fact that the OTel quantile is not actually supposed to be that particular type of quantile. Spec says

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#summary-legacy

`within the interval [0.0, 1.0]`

We would at least need to change the name of the class to `ValueAtQuantile` and this obviously wrong javadoc

https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/ValueAtPercentile.java#L21

Would also need to make sure we don't have any logic such as validation that is using the percentile expectation

_Originally posted by @anuraaga in https://github.com/open-telemetry/opentelemetry-java/pull/4178#discussion_r809653928_

anuraaga:
Fixes #4206 

Note that the prometheus spec says exporting quantiles between [0.0,1.0] so no need to change logic in the prometheus exporter that was previously filling [0.0,100.0] (for users of OC shim, the only usage of Summary)

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#summaries-1


opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/internal/metrics/MetricAdapter.java
```diff
@@ -370,12 +370,13 @@ static long mapTimestamp(@Nullable Timestamp time) {
...
-  private static List<ValueAtPercentile> mapValueAtPercentiles(
+  private static List<ValueAtQuantile> mapValueAtPercentiles(
       List<Summary.Snapshot.ValueAtPercentile> valueAtPercentiles) {
-    List<ValueAtPercentile> result = new ArrayList<>(valueAtPercentiles.size());
+    List<ValueAtQuantile> result = new ArrayList<>(valueAtPercentiles.size());
     for (Summary.Snapshot.ValueAtPercentile censusValue : valueAtPercentiles) {
       result.add(
-          ImmutableValueAtPercentile.create(censusValue.getPercentile(), censusValue.getValue()));
+          ImmutableValueAtQuantile.create(
+              censusValue.getPercentile() / 100.0, censusValue.getValue()));
```
anuraaga, line 379: This is the only logic change

jack-berg (approved):

opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/internal/metrics/MetricAdapter.java
jkwatson: this does end up being a breaking change for any OC shim users, I guess. Do we have any?

jkwatson (approved):

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/otlp/metrics/ValueAtQuantileMarshaler.java
```diff
@@ -26,8 +26,8 @@
     return marshalers;
   }
 
-  private static ValueAtQuantileMarshaler create(ValueAtPercentile value) {
-    return new ValueAtQuantileMarshaler(value.getPercentile() / 100.0, value.getValue());
```
anuraaga, line 30: Sorry I lied here is the other logic change. So no change for opencensus + OTLP. Some change for opencensus + prometheus but we probably have to do it.

PR #4255, opened 2022-03-09
Migrate Aggregation to interface

anuraaga:
We don't currently support custom aggregations and enforce this at compile time with a private constructor on a class. However, we do anticipate supporting them, and any time we support a customization point like that, we generally expose it as an interface, not abstract class. Unfortunately switching from abstract class to interface is not semver compatible.

The workaround which seems reasonable is to move from the compile time prevention to a runtime check.


jack-berg (approved):

anuraaga:
On second thought I realized that class to interface may technically be compatible in this case. Lacking any public methods, there isn't any way to link a `invokevirtual` to the type. But it's probably too much of a nuance to rely on, presumably japicmp will agree.

jkwatson (approved):

PR #4256, opened 2022-03-09
Deprecate InstrumentationLibraryInfo

jack-berg:
Related to #4190.

- Adjust ReadableSpan, SpanData to include accessor to InstrumentationScopeInfo, deprecate #getInstrumentationLibraryInfo()
- Adjust MetricData to include accessor to InstrumentationScopeInfo, deprecate #getInstrumentationLibraryInfo()
- Adjust LogData to include accessor to InstrumentationScopeInfo, deprecate #getInstrumentationLibraryInfo()


exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/otlp/logs/InstrumentationScopeLogsMarshaler.java
```diff
@@ -9,42 +9,42 @@
...
   private static int calculateSize(
-      InstrumentationLibraryMarshaler instrumentationLibrary,
+      InstrumentationScopeMarshaller instrumentationScope,
       byte[] schemaUrlUtf8,
       List<Marshaler> logMarshalers) {
     int size = 0;
     size +=
         MarshalerUtil.sizeMessage(
-            InstrumentationLibraryLogs.INSTRUMENTATION_LIBRARY, instrumentationLibrary);
+            InstrumentationLibraryLogs.INSTRUMENTATION_LIBRARY, instrumentationScope);
```
anuraaga, line 47: Just merged the proto update, so think you need to merge main and update the keys now when marshaling

sdk/common/src/test/java/io/opentelemetry/sdk/common/InstrumentationLibraryInfoTest.java
```diff
@@ -10,18 +10,21 @@
 
 import org.junit.jupiter.api.Test;
 
-/** Tests for {@link InstrumentationLibraryInfo}. */
+/** Tests for {@link io.opentelemetry.sdk.common.InstrumentationLibraryInfo}. */
```
anuraaga, line 13: Looks like some qualified imports got automatically added by intellij or something

jack-berg: You have to qualify the imports, even in comments, or else you get an error for using deprecated classes.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/otlp/logs/InstrumentationScopeLogsMarshaler.java
jack-berg: Actually, it looks like the proto PR to change the field names hasn't merged yet: https://github.com/open-telemetry/opentelemetry-proto/pull/362

sdk/common/src/test/java/io/opentelemetry/sdk/common/InstrumentationLibraryInfoTest.java
anuraaga: Ah thought if it's in the same package that wouldn't be an issue but maybe not

jkwatson (approved):

jkwatson-verta:
We all missed that this was a breaking change, since we added methods to public interfaces without a default implementation. :(

PR #4257, opened 2022-03-10
Make MetricData factories internal

anuraaga:
As a reminder, we don't expect to need APIs to create MetricData until supporting custom aggregations. Either way, when we do present them we'll want to come up with a better API, probably using builders, as the current ones don't really follow our general goals for APIs by having a large number of parameters, many with the same `String` type allowing easy bugs.


jkwatson (approved):
thanks! good cleanup

PR #4258, opened 2022-03-10
Bring in micrometer shim

### LINKED ISSUE #4248: Move micrometer shim to the core repo
We currently have a shim from Micrometer to OTel in the instrumentation repo

https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/micrometer/micrometer-1.5/library

I would like to propose moving it to this repo

- Micrometer is an important part of the Java ecosystem and I think it is fair to expect core maintainers to ensure the integration with OTel works well for users
- The shim doesn't actually use instrumentation concepts such as semantic conventions or the instrumentation-api
- Having it here would allow creating internal utilities in the SDK to improve the integration experience where needed

/cc @mateuszrzeszutek

anuraaga:
Fixes #4248


micrometer-shim/build.gradle.kts
```diff
@@ -0,0 +1,54 @@
+plugins {
+  id("otel.java-conventions")
+  id("otel.publish-conventions")
+}
+
+description = "OpenTelemetry Micrometer Bridge"
+otelJava.moduleName.set("io.opentelemetry.micrometershim")
+
+dependencies {
+  api("io.micrometer:micrometer-core")
```
anuraaga, line 10: In instrumentation repo this is `compileOnly` but here went with `api` for now to match pattern in opentracing shim but may change it.

micrometer-shim/src/main/java/io/opentelemetry/micrometershim/Bridging.java
```diff
@@ -0,0 +1,55 @@
...
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.Statistic;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.config.NamingConvention;
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.common.AttributesBuilder;
+
+final class Bridging {
+
+  static Attributes tagsAsAttributes(Meter.Id id, NamingConvention namingConvention) {
```
anuraaga, line 17: Removed caches from this class since we don't (yet) have a concurrent cache in this repo

- I've always wondered whether we should cache `AttributesBuilder.put(String key` invocations in the API itself. I don't think there's anything unique to micrometer to do it specifically here
- #4222 will land soon to remove need for description cache

micrometer-shim/src/main/java/io/opentelemetry/micrometershim/OpenTelemetryMeterRegistryBuilder.java
```diff
@@ -0,0 +1,58 @@
...
+/** A builder of {@link OpenTelemetryMeterRegistry}. */
+public final class OpenTelemetryMeterRegistryBuilder {
+
+  // Visible for testing
+  static final String INSTRUMENTATION_NAME = "io.opentelemetry.micrometershim";
+
+  private final OpenTelemetry openTelemetry;
+  private Clock clock = Clock.SYSTEM;
+  private TimeUnit baseTimeUnit = TimeUnit.MILLISECONDS;
+  private NamingConvention namingConvention = NamingConvention.identity;
```
anuraaga, line 23: No config property read here like in instrumentation repo

anuraaga (commented):
Comments below for the only two locations business logic changed. Tests were modified fairly significantly though should still be testing the same cases

micrometer-shim/src/main/java/io/opentelemetry/micrometershim/Bridging.java
mateuszrzeszutek: > * I've always wondered whether we should cache `AttributesBuilder.put(String key` invocations in the API itself. I don't think there's anything unique to micrometer to do it specifically here

That would be awesome, I think - we've been recommending using `AttributeKey` everywhere to avoid additional allocations, but if we could fix that at the source that would be great.

mateuszrzeszutek (approved):

micrometer-shim/src/main/java/io/opentelemetry/micrometershim/OpenTelemetryMeterRegistryBuilder.java
```diff
@@ -0,0 +1,58 @@
...
+  public OpenTelemetryMeterRegistryBuilder setBaseTimeUnit(TimeUnit baseTimeUnit) {
+    this.baseTimeUnit = baseTimeUnit;
+    return this;
+  }
+
+  /** Sets the {@link NamingConvention} for this registry. */
+  public OpenTelemetryMeterRegistryBuilder setNamingConvention(NamingConvention namingConvention) {
+    this.namingConvention = namingConvention;
+    return this;
+  }
```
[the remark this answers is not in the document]
anuraaga: I think having common options here is good since it keeps all the configuration in one place and discoverable. Naming convention is quite key

mateuszrzeszutek:
Now that I think of it, there's going to be Micrometer 2.0 release soon - we'll most likely need another shim for that, I think we can safely assume that there're going to be incompatible changes in there. Should we include the version in the artifact/module name? E.g. `micrometer-1-shim`

micrometer-shim/src/main/java/io/opentelemetry/micrometershim/OpenTelemetryMeterRegistryBuilder.java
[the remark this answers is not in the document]
anuraaga: That being said better to start with less than more API, so went back

micrometer-shim/src/main/java/io/opentelemetry/micrometershim/UnsupportedReadLogger.java
```diff
@@ -0,0 +1,23 @@
...
+
+package io.opentelemetry.micrometershim;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+final class UnsupportedReadLogger {
+
+  static {
+    Logger logger = Logger.getLogger(OpenTelemetryMeterRegistry.class.getName());
```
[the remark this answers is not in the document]
anuraaga: Just like we don't want people to use our internal classes I don't want to use others' as well :P This logger class seems fine for OTel users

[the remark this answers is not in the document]
mateuszrzeszutek: Hmm, I had the impression that it could be used inside `MeterRegistry` implementations - but I've double checked it just now, it seems it's only used in `micrometer-core` itself.
Yeah, let's keep using JUL then.

mateuszrzeszutek (approved):

anuraaga:
@jack-berg @jkwatson Would be great if you could give this an ack, thanks!

jkwatson:
Will look tomorrow first thing

micrometer1-shim/src/main/java/io/opentelemetry/micrometershim/OpenTelemetryCounter.java
```diff
@@ -0,0 +1,86 @@
...
+import io.micrometer.core.instrument.Measurement;
+import io.micrometer.core.instrument.config.NamingConvention;
+import io.micrometer.core.instrument.util.MeterEquivalence;
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.metrics.DoubleCounter;
+import io.opentelemetry.api.metrics.Meter;
+import java.util.Collections;
+import javax.annotation.Nullable;
+
+@SuppressWarnings("HashCodeToString")
```
jkwatson, line 22: what is this warning about?

jkwatson (approved):
Didn't give it anything close to a thorough review, but it's marked as `alpha`, so I'm not overly concerned.

micrometer-shim/src/main/java/io/opentelemetry/micrometershim/Bridging.java
jkwatson: I'm a bit hesitant, only because we'd need to be careful about how much we cache... we wouldn't want a pathological API usage to OOM us.

jack-berg (approved):

micrometer1-shim/src/main/java/io/opentelemetry/micrometershim/OpenTelemetryCounter.java
anuraaga: I think since hashcode is defined but toString isn't. It seems like it'd be worth defining toString but will follow up

Issue #4251, opened 2022-03-08
Metrics Annotations


fstab:
Frameworks like Spring Boot, MicroProfile, and Dropwizard offer annotations like `@Timed` or `@Counted` for creating metrics from method calls.

The benefit is that developers can provide metrics for their business logic while still separating business code from metrics tracking.

```
@Timed
@Counted
public void processPayment() {
    // business code here
}
```

The OpenTelemetry Java auto instrumentation supports the `@WithSpan` annotation for tracing, but has nothing equivalent for metrics yet.

Brainstorming of a few ideas:

* For `@Timed` the type of histogram should be configurable, like `@Timed(histogramType = EXPONENTIAL)`. Other histogram types would be explicit buckets and default buckets.
* Explicit buckets should be configurable: `@Timed(buckets = { 0.001, 0.002, 0.003, 0.004 })`
* Static attributes should be configurable: `@Counted(attributes = { "key", "value" })`
* Dynamic attributes should be configurable similar to OpenTelemetry's `@SpanAttribute`.
* Dynamic attributes based on the return value and on an exception thrown would be great.

Moreover, it would be awesome to integrate this with `@WithSpan` so that develpers can say "I want to time this method AND have a Span". However, it should also be possible to use tracing without metrics, because histograms may introduce a lot of cardinality, and users might just want to create a Span without creating a histogram.

What do you think, is it worthwhile to put some more thoughts into this?

jkwatson:
I think things that are a strict SDK responsibility (bucket definitions for example) should not end up being referenced in an annotation, since the annotations should be a pure API artifact.

Aside from that detail, I think having something like this would be a great idea.

jkwatson:
At the moment, all configuration of how histograms work (and metrics aggregations in general) is the responsibility of the application _operator_ via SDK configuration, and is not a concern assigned to instrumentation authors. In the future, we may have what we've been calling a "hint" API that would allow instrumentation call sites to inform the SDK of what they would prefer, but that is not currently specified.

jkwatson:
> Thanks a lot for the explanation, this makes sense. In that case the annotations would be quite simple, as they only provide configuration options for the metric name and attributes.
> 
> I could open a PR to add them to `extensions/annotations/` so that we can continue the discussion based on concrete examples. What do you think?

That would be fine. Since these annotations are generally implemented by the instrumentation folks, I'd love some feedback on this proposal from @trask @mateuszrzeszutek @anuraaga and others.

[...]

PR #4260, opened 2022-03-12
Add Metrics Annotations

fstab:
Add annotations `@Timed` and `@Counted` for automatically instrumenting method invocations.

The annotations are similar to the annotations offered by Micrometer, MicroProfile metrics, and Dropwizard metrics.

The default attributes **class**, **method**, and **exception** are similar to the default attributes provided by Micrometer's annotations.

Here's an example of what a `@Counted` metric would look like in Prometheus format:

```
method_invocations_total{class="com.example.demo.Example", method="example", exception="None"} 1.0
```

This PR is a follow-up to the discussion in #4251.


extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/TimerAttribute.java
```diff
@@ -0,0 +1,36 @@
...
+ * <p>Application developers can use this annotation to signal OpenTelemetry auto-instrumentation
+ * that the attribute should be created.
+ *
+ * <p>If you are a library developer, then probably you should NOT use this annotation, because it
+ * is non-functional without the OpenTelemetry auto-instrumentation agent, or some other annotation
+ * processor.
+ */
+@Target(ElementType.PARAMETER)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface TimerAttribute {
```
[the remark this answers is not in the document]
mateuszrzeszutek: Could we perhaps reuse `SpanAttribute` somehow? Or deprecate it, and introduce a new annotation that would apply to both signals?

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Timed.java
```diff
@@ -0,0 +1,101 @@
...
+
+  /**
+   * Attribute name for the return value.
+   *
+   * <p>The name of the attribute for the return value of the method call. {@link Object#toString()}
+   * will be called on the return value to convert it to a String.
+   *
+   * <p>By default, the instrument will not have an attribute with the return value.
+   */
+  String returnValueAttribute() default "";
```
mateuszrzeszutek, line 87: A couple days ago somebody asked for the same thing for spans: https://github.com/open-telemetry/opentelemetry-java-instrumentation/discussions/5527
Instead of an attribute, perhaps this could be a separate annotation? That's used by both signals (traces and metrics)

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Timed.java
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * This annotation creates a {@link io.opentelemetry.api.metrics.LongHistogram Histogram} instrument
+ * observing the duration of invocations of the annotated method or constructor.
+ *
+ * <p>By default, the Histogram instrument will have the following attributes:
+ *
+ * <ul>
+ *   <li><b>class:</b> The fully qualified name of the class whose method is invoked.
+ *   <li><b>method:</b> The name of the annotated method, or "new" of the annotation is on a
+ *       constructor.
```
mateuszrzeszutek, line 23: Instead of `class` and `method`, WDYT about using `code.namespace` and `code.function`? These attributes are already in the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#source-code-attributes) (for spans, at least), and already are (or should be) captured by the agent.

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Timed.java
```diff
@@ -0,0 +1,101 @@
...
+ *   <li><b>exception:</b> The {@link Class#getSimpleName() simple name} of the Exception if an
+ *       Exception is thrown, or "None" if the method did not throw an Exception.
```
mateuszrzeszutek, line 25: Hmm, can we use `exception.type` and `Class#getCanonicalName()` instead? It's what `Span#recordException()` uses.

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Timed.java
```diff
@@ -0,0 +1,101 @@
...
+   *
+   * <p>By default, the instrument will have an attribute named {@code exception}. If an Exception
+   * is thrown by the annotated method, the value of the {@code exception} attribute is the {@link
+   * Class#getSimpleName() simple name} of that Exception. If the annotated method terminates
+   * without throwing an Exception, the value will be {@code None}.
+   *
+   * <p>Setting {@code disableExceptionAttribute = true} will disable the {@code exception}
+   * attribute.
+   */
+  boolean disableExceptionAttribute() default false;
```
mateuszrzeszutek, line 100: Is this method necessary? The same thing can be achieved by a `View` if somebody wants that; and turning off the exception capturing does not seem that useful.

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Timed.java
fstab: There should be an option to use an attribute for a span but not for a metric. For metrics, each set of attributes creates a new time series, and users must be careful to use only attributes that have a limited number of possible different values. For spans that is not an issue. So it might make sense to use arbitrary parameters, return values, and exceptions as Span attributes but not use them as metric attributes.

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/TimerAttribute.java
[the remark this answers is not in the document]
fstab: @jkwatson true, it does not make sense to use both `@Timed` and `@Counted` since timers already include the count. I'm happy to merge `TimerAttribute` and `CounterAttribute` into a single `MetricAttribute`.

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Timed.java
fstab: Thanks for the heads-up. If that's the standard attribute names from the Spec we should use them.

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Timed.java
fstab: I'm not sure if there is a situation where Exceptions can result in cardinality explosion because a large number of different Exception classes is thrown. If this can happen it would make sense to have a way to disable the exception attribute. If we are sure that Exception classes are always low cardinality we could remove this.

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Timed.java
fstab: Thanks for the heads-up. If that's the standard attribute names from the Spec we should use them.

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Timed.java
mateuszrzeszutek: It's _technically_ possible to have hight cardinality here (like autogenerating exception classes at runtime 🙈 ), but it's very low likely to happen -- and again, you can use Views to filter this attribute out if you somehow happen to run into such scenario.

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Timed.java
fstab: I guess if we want to have an option to use the return value as a span attribute but not as a metric attribute we don't need an extra annotation, because then we cannot reuse it for spans and metrics anyway. In that case we could just leave the `returnValueAttribute` parameter.

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Timed.java
fstab: I pushed an update renaming this to the standard attributes.

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Timed.java
fstab: I pushed an update using `exception.type` and the canonical name.

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Timed.java
fstab: I pushed an update removing the `disableExceptionAttribute` parameter.

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Counted.java
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * This annotation creates a {@link io.opentelemetry.api.metrics.LongCounter Counter} instrument
+ * recording the number of invocations of the annotated method or constructor.
+ *
+ * <p>By default, the Counter instrument will have the following attributes:
+ *
+ * <ul>
+ *   <li><b>code.namespace:</b> The fully qualified name of the class whose method is invoked.
+ *   <li><b>code.function:</b> The name of the annotated method, or "new" of the annotation is on a
```
trask, line 21:
```suggestion
 *   <li><b>code.function:</b> The name of the annotated method, or "new" if the annotation is on a
```

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Counted.java
```diff
@@ -0,0 +1,88 @@
...
+ *       constructor.
+ *   <li><b>exception.type:</b> The {@link Class#getCanonicalName()} () canonical name} of the
```
trask, line 23:
```suggestion
 *   <li><b>exception.type:</b> The {@link Class#getCanonicalName() canonical name} of the
```

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Counted.java
```diff
@@ -0,0 +1,88 @@
...
+
+  /**
+   * Name of the Counter instrument.
+   *
+   * <p>The name should follow the instrument naming rule: <a
+   * href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-naming-rule">https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-naming-rule</a>
+   *
+   * <p>The default name is {@code method.invocations.total}.
+   */
+  String value() default "method.invocations.total";
```
trask, line 46: since `method.invocations.total` hasn't been spec'd, maybe we shouldn't have a default, or maybe we should move the metrics annotations to an alpha package for now?

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Counted.java
```diff
@@ -0,0 +1,88 @@
...
+   *
+   * <pre>
+   * {@literal @}Counted(
+   *     additionalAttributes = {
+   *       "key1", "value1",
+   *       "key2", "value2",
+   * })
+   * </pre>
+   */
+  String[] additionalAttributes() default {};
```
trask, line 77: just curious, do you have a particular use case in mind for these additional attributes?

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Counted.java
```diff
@@ -0,0 +1,88 @@
...
+ * recording the number of invocations of the annotated method or constructor.
+ *
+ * <p>By default, the Counter instrument will have the following attributes:
+ *
+ * <ul>
+ *   <li><b>code.namespace:</b> The fully qualified name of the class whose method is invoked.
+ *   <li><b>code.function:</b> The name of the annotated method, or "new" of the annotation is on a
+ *       constructor.
+ *   <li><b>exception.type:</b> The {@link Class#getCanonicalName()} () canonical name} of the
+ *       Exception if an Exception is thrown, or "None" if the method did not throw an Exception.
```
trask, line 24: on the instrumentation side, we don't populate metric attributes when they are n/a, I definitely don't understand some of the nuances of metric design, is populating "None" needed for some reason?

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Timed.java
trask: I think @mateuszrzeszutek was suggesting something like this:

```
  @Timed
  @ReturnAttribute("process.result")
  public int process() {
    ...
  }
```

mateuszrzeszutek: Yeah, exactly that -- and this annotation could be used to fill in attributes both in spans and metrics.

fstab: I'm not quite sure whether we should have common attributes for both spans and metrics, or whether we should separate them. The difference is that with metrics each new attribute value creates a new time series. Therefore you only want to choose attributes that have a limited number of possible values. With Spans that's not the case.

An example would be a method returning a unique identifier: It would be valid to use the return value as a Span attribute, but it would be a bad idea to use it as a metric attribute.

For method parameters the current proposal is to have two separate annotations: `@MetricAttribute` and `@SpanAttribute`. If you want a parameter to be used in both, metric and span, you would need to add both annotations.

If we want to have the same separation for return values, we would end up with something like `@MetricReturnAttribute` and `@SpanReturnAttribute`. That's fine. I find it shorter to have this as a parameter to the `@Counted`, `@Timed`, and maybe `@WithSpan`, but I have no strong opinion about this.

However, having a common `@ReturnAttribute` for both metrics and spans would make it impossible to use a high cardinality return value as a Span attribute without using it as a metric attribute at the same time.

What do you think?

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Counted.java
fstab: Moving to an alpha package sounds good. Could you let me know how to do this? I don't know the conventions too well.

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Counted.java
fstab: Something that comes to my mind would be this:
```java
  @Timed(attributes = {"paymentType", "CreditCard"})
  public void processPayment(CreditCard creditCard) {
    // ...
  }

  @Timed(attributes = {"paymentType", "DebitCard"})
  public void processPayment(DebitCard debitCard) {
    // ...
  }
```
Does this make sense?

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Counted.java
fstab: Good catch. The reason why it's there is because that's what Micrometer does, but I guess not populating it would be better.

In Prometheus missing attributes are treated the same way as empty attributes. I pushed an update saying that the attribute is only present if an Exception is thrown. Is this ok, or would it be better to always keep the attribute but with an empty value?

extensions/annotations/src/main/java/io/opentelemetry/extension/annotations/Timed.java
trask: Oh, I understand now, yeah that makes sense that you would want control over which attributes are tagged on the span vs tagged on the metric

tylerlance:
Do we have a planned version for this to be merge for? This seems like a valuable feature. Hoping it wasn't abandoned.

jack-berg:
It feels strange accepting additions to the stable extension annotations module without knowing what an implementation of a corresponding annotation processor looks like.

Can `opentelemetry-extension-annotations` move to `opentelemetry-java-instrumentation` to live alongside the annotation processor implementation?

If not, can we we put this in `opentelemetry-extension-incubator` while we build confidence in the API / implementation?

fstab:
My initial impulse is that `opentelemetry-extension-incubator` is a better fit because users would explicitly use these annotations in their code. Should I move it?

fstab:
I moved it to `extensions/incubator/`. The new annotations are now in package `io.opentelemetry.extension.incubator.annotations`.

From the discussion above I think the only unresolved point is how to use return values as attributes.

Current implementation:
```java
  @Timed(returnValueAttribute = "process.result")
  public int process() {
    return 3;
  }
```
Alternative proposal:
```java
  @Timed
  @ReturnAttribute("process.result")
  public int process() {
    return 3;
  }
```
I think the idea behind the alternative proposal is that `@ReturnAttribute` could work for both, spans and metrics. However, there is a fundamental difference between spans and metrics: For metrics, each attribute value creates a new time series. If you use a high cardinality return value as a metric attribute, you might overload the monitoring backend. For spans, high cardinality attributes are not an issue.

For that reason I think we should provide a way to say that a return value will be a span attribute, but not a metric attribute. With `returnValueAttribute` being a property of the annotation, this is clear.

Anyway, I'm happy to change this if consensus is that we should go with a dedicated `@ReturnAttribute` annotation.

jkwatson:
@fstab sorry about the delay on this. We're going to discuss at today's OTel java office hours and will respond with a recommendation on next steps ASAP.

jack-berg:
Hi @fstab just following up on @jkwatson's comment. Here's where we're at:

- We don't think opentelemetry-java is the right long term home for these annotations, given that the primary implementation lives in opentelemetry-java-instrumentation. We'd like to make opentelemetry-java-instrumentation the long term home of annotations. 
- Modules in opentelemetry-java-instrumentation generally live in `io.opentelemetry.instrumentation.*` package, versus `io.opentelemetry.*` for opentelemetry-java projects. In order to continue following this pattern while not breaking compatibility for the annotations module, we're going to start a new annotations module in opentelemetry-java-instrumentation, deprecating the opentelemetry-java `opentelemetry-extension-annotations` module until we ultimately stop publishing it starting with the next major version.
- We'd like to not get further away from this target, and merging this until the span annotations have been moved to opentelemetry-java-instrumentation. At that point, we can open a PR adding these annotations to that new module, and simultaneously include an implementation for the annotations.

Again, sorry for the delay.

trask:
I've kicked off the discussion in the instrumentation repo https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/6245, and added the topic of naming to the SIG agenda for tomorrow. Once we close on naming, I'll do the initial copy of existing annotations over to the instrumentation repo.

extensions/incubator/src/main/java/io/opentelemetry/extension/incubator/annotations/Counted.java
```diff
@@ -0,0 +1,88 @@
...
+   */
+  String description() default "";
+
+  /**
+   * Unit of the instrument.
+   *
+   * <p>Unit strings should follow the instrument unit rules: <a
+   * href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-unit">https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-unit</a>
+   */
+  String unit() default "";
```
jsuereth, line 62: Should this default to "1" ?

extensions/incubator/src/main/java/io/opentelemetry/extension/incubator/annotations/Timed.java
```diff
@@ -0,0 +1,88 @@
...
+   * href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-description">https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-description</a>
+   */
+  String description() default "";
+
+  /**
+   * The unit for the instrument.
+   *
+   * <p>Default is seconds.
+   */
+  TimeUnit unit() default TimeUnit.SECONDS;
```
jsuereth, line 62: OTEL usually defaults to milliseconds, e.g. https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md

trask:
hey @fstab! we've finished moving the `@WithSpan` instrumentation annotations to the instrumentation repo (https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation-annotations). we're keeping the existing annotations around in this repository (and still supporting them in auto-instrumentation for backwards compatibility), but would like to add new features going forwards in the new module now

jack-berg:
I'm going to close this because now that we've moved forward with the decision to move annotations to `opentelemetry-java-instrumentation`.

cptkng:
@fstab Could you summarize what the current status of your (great) idea with annotations for metrics is?

Issue #3882, opened 2021-11-16
MetricReaderFactory is not a "Factory". Find a better name or remove if not needed?


bogdandrutu:
It may be better if this is not a "post-construction" function and accept "next" reader/consumer in the constructor. Avoids any thread-safety issue.

jack-berg:
`MetricReaderFactory` was dropped in #4268, closing this issue.

PR #4268, opened 2022-03-15
Simplify MetricReader

jack-berg:
Proposal to simplify the `MetricReader`, `MetricReaderFactory`, `MetricProducer` setup:

- Remove `MetricReaderFactory`
- Move `MetricProducer` to internal class
- `MetricReader` instances are directly registered with `SdkMeterProviderBuilder`
- Custom implementations of `MetricReader` are prohibited for now. `SdkMeterProviderBuilder` only accepts instances of `AbstractMetricReader`, which is an internal class which obfuscates the process of a registering `MetricProducer` with the reader. Later, we can also custom readers by making `MetricProducer` public and adding a registration method to `MetricReader`. 

I think it's more intuitive to register a reader than a reader factory. It also makes the reader builders like `PeriodicMetricReaderBuilder` more intuitive since they now have a `build()` method that returns `PeriodicMetricReader`, rather than the rather confusing `newMetricReaderFactory()`.

Related to #3882.


anuraaga (commented):
Thanks! I was curious if we could remove MetricProducer and it looks like you found a nice approach

jack-berg:
Discussed another deprecation / RC cycle in the 3/17 Office Hours and current thinking is that we will not and therefore the 1.13.0 release will include some breaking changes. 

Reverting the deprecation cycle code in this PR.

anuraaga (approved):

PR #4271, opened 2022-03-17
Add SpanDataAssert.hasAttribute

zeitlinger:
to be able to assert a single attribute in a fluent way

### DIFF
--- a/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/SpanDataAssert.java
+++ b/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/SpanDataAssert.java
@@ -252,6 +252,13 @@ public final class SpanDataAssert extends AbstractAssert<SpanDataAssert, SpanDat
     return startsAt(toNanos(timestamp));
   }
 
+  /** Asserts the span attributes. */
+  public SpanDataAssert assertAttributes(Consumer<AttributesAssert> attributesAssert) {
+    isNotNull();
+    attributesAssert.accept(assertThat(actual.getAttributes()));
+    return this;
+  }
+
   /** Asserts the span has the given attributes. */
   public SpanDataAssert hasAttributes(Attributes attributes) {
     isNotNull();
--- a/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/SpanDataAssert.java
+++ b/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/SpanDataAssert.java
@@ -252,10 +252,24 @@ public final class SpanDataAssert extends AbstractAssert<SpanDataAssert, SpanDat
     return startsAt(toNanos(timestamp));
   }
 
-  /** Asserts the span attributes. */
-  public SpanDataAssert assertAttributes(Consumer<AttributesAssert> attributesAssert) {
+  /** Asserts a single span attributes. */
+  public <T> SpanDataAssert hasAttribute(AttributeKey<T> key, T value) {
+    return hasAttribute(OpenTelemetryAssertions.equalTo(key, value));
+  }
+
+  /** Asserts a single span attributes. */
+  public SpanDataAssert hasAttribute(AttributeAssertion attributeAssertion) {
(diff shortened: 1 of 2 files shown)


anuraaga:
Hi @zeitlinger - we recently added `hasAttributesSatisfyingExactly`, have you tried using it? We have many ways of asserting attributes already so hesitant on adding yet another one.

zeitlinger:
> Hi @zeitlinger - we recently added `hasAttributesSatisfyingExactly`, have you tried using it? We have many ways of asserting attributes already so hesitant on adding yet another one.

yes, I saw that one - it has the problem that you need to assert every attribute.
In my setup, the auto instrumentation already defines some  attributes, which I don't want to mention - there're out of scope for the metric I define.

this is the best setup I have right now:

```java
                    span.hasAttributesSatisfying(attributes -> {
                        assertThat(attributes)
                                .containsEntry("stock_group", "primary");
                    });
```

This PR improves this scenario, because it's fluent (i.e. it avoid that static `assertThat`).

I think the new assert idiom is to go from an outer to an inner assertion - and this idiom is already support in `public final TracesAssert hasTracesSatisfyingExactly(Consumer<TraceAssert>... assertions)` and `public final TraceAssert hasSpansSatisfyingExactly(Consumer<SpanDataAssert>... assertions)`. I wasn't able to use `hasAttributesSatisfyingExactly` in this PR, because it clashes with a method that doesn't follow the same idiom (because it doesn't prove an assert object).

anuraaga:
I think this unfortunately a result of the current `hasAttributesSatisfying` being one of the earlier assertions before some of our patterns got settled - if adding it now it probably would be `Consumer<AttributesAssert>` and not its current form.

The naming `assertAttributes` is quite different from others and does have the issue of adding yet another similar method so do still want to avoid it, we can consider the new method to be mostly syntactic sugar, it is useful sugar but I think it would be better replacing one of the existing methods in a future v2. If the main use case is checking a single or couple of attributes, would adding `hasAttribute` be enough? That would be new useful functionality and have a naming consistent with the others.

zeitlinger:
> would adding `hasAttribute` be enough? 

Yes, that's exactly what I was looking for :bulb: - together with `hasTotalAttributeCount` it can even be an alternative to the current `hasAttributesSatisfyingExactly` (if one prefers).

zeitlinger:
> would adding `hasAttribute` be enough? 

Yes, that's exactly what I was looking for :bulb: - together with `hasTotalAttributeCount` it can even be an alternative to the current `hasAttributesSatisfyingExactly` (if one prefers).

anuraaga (approved):

kubawach (approved):
Thanks!

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/SpanDataAssert.java
```diff
@@ -252,6 +252,27 @@ public SpanDataAssert startsAt(Instant timestamp) {
     return startsAt(toNanos(timestamp));
   }
 
+  /** Asserts a single span attributes. */
```
jack-berg, line 255:
```suggestion
  /** Asserts the span has the given attribute. */
```

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/SpanDataAssert.java
```diff
@@ -252,6 +252,27 @@ public SpanDataAssert startsAt(Instant timestamp) {
...
+  public <T> SpanDataAssert hasAttribute(AttributeKey<T> key, T value) {
+    return hasAttribute(OpenTelemetryAssertions.equalTo(key, value));
+  }
+
+  /** Asserts a single span attributes. */
```
jack-berg, line 260:
```suggestion
  /** Asserts the span has an attributes matching the {@code assertions}. */
```

jack-berg (approved):
Small #nit on the javadoc but looks good.

anuraaga (approved):

PR #4273, opened 2022-03-17
Metrics SDK odds and ends

jack-berg:


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/InstrumentSelector.java
```diff
@@ -38,6 +38,8 @@ static InstrumentSelector create(
         meterSchemaUrlFilter);
   }
 
+  InstrumentSelector() {}
```
jack-berg, line 41: Its a bit strange that `InstrumentSelector` is an abstract class while `View` is an interface, given that we don't expect anyone to provide their own implementation of `View` / `ViewBuilder`.

This appears to be a consequence of the metrics SDK internals needing access to pieces of the view that should not be publicly accessible, like `SourceInfo` and `AttributesProcessor`.

But I'm not sure this is necessary. The view could be delivered to the SDK as a simple data carrier of its configurable parts: name, description, and aggregation. The translation of that configuration to things like `SourceInfo` and `AttributesProcessor` could be done upon registration and kept an internal implementation detail.

If we did this then view could be an abstract class with `@AutoValue` and be consistent with `InstrumentSelector`.

anuraaga: Another reason for making `ViewBuilder` specifically an interface was to allow relatively easily incubating internal features such as baggage appending. While doing this, I sort of came to a general idea that interfaces may just "be better" in general for these APIs to give us this sort of flexibility, though it's not a hard conclusion. Indeed it would mean it is a good idea to migrate `InstrumentSelector` to interface to be consistent if this approach is actually a good idea.

jack-berg: Yeah I later noticed the additional internal-only capabilities of `ViewBuilderImpl`. 

One advantage of the abstract class / `@AutoValue` approach is that it makes it clear that the configuration is a simple immutable data carrier, rather than something implementable, with responses that could change over time. 

Do you (or anyone else) know of any use cases for baggage appenders and general attribute appenders?

anuraaga (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProviderBuilder.java
```diff
@@ -29,7 +29,7 @@
   private Resource resource = Resource.getDefault();
   private final ViewRegistryBuilder viewRegistryBuilder = ViewRegistry.builder();
   private final List<MetricReaderFactory> metricReaders = new ArrayList<>();
-  // Default the sampling strategy.
+  // Set the default exemplar filter strategy.
   private ExemplarFilter exemplarFilter = ExemplarFilter.sampleWithTraces();
```
kubawach, line 33: Perhaps we should document these defaults (sampleWithTraces and min coll interval) in the class JavaDoc? Looks to be a better place than field doc.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/ViewBuilder.java
```diff
@@ -11,15 +11,15 @@
...
+   * Sets the name of the resulting metric.
    *
    * @param name metric name or {@code null} if the underlying instrument name should be used.
    * @return this Builder.
    */
   ViewBuilder setName(String name);
 
   /**
-   * sets the name of the resulting metric.
+   * Sets the description of the resulting metric.
```
kubawach, line 22: Nice catch :)

kubawach (approved):

PR #4277, opened 2022-03-18
Refactor view API, make AttributesProcessor internal

jack-berg:
Originally discussed [here](https://github.com/open-telemetry/opentelemetry-java/pull/4273#discussion_r829559732).

Notes:
- Drop `io.opentelemetry.sdk.metrics.view` package. 
- Move `Aggregation`, `View`, `ViewBuilder`, `InstrumentSelector`, `InstrumentSelectorBuilder` to `io.opentelemetry.sdk.metrics`. Remaining classes move to internal package. Note, `AttributesProcessor` is no longer part of public API.
- Change `View`, `ViewBuilder` from interfaces to classes.
- `View#getAttributesProcessor()` is package private.
- `ViewUtil` is an internal class with static methods that use reflection to allow baggage attributes to be appended to metrics.


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/view/AttributesProcessor.java
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.metrics.view;
+
+/** An {@code AttributesProcessor} is used by {@code View}s to define attribute processing. */
```
anuraaga, line 8: Can we keep this class private? I think it's a bit early for it and shouldn't be needed with the configuration methods we have right now

jack-berg: However, there would be another option if the `io.opentelemetry.sdk.metrics.view` package was collapsed into `io.opentelemetry.sdk.metrics`. If `SdkMeterProvider` was in the same package as `View`, then `View` could have package private methods for accessing configuration parameters, including any experimental features like appending baggage attributes.

anuraaga: Hmm - yeah moving the code into the same package seems like a nice idea in general

jack-berg: I've done this in my latest commit and think it makes sense. I've updated the PR title and description to reflect the current changes.

anuraaga (approved):

PR #4280, opened 2022-03-20
Add option in BatchSpanProcessor to support multiple pending exports

### LINKED ISSUE #4264: How to implement concurrent exporter?

I'd like to increase export throughput, and the most obvious way (at least in my case with a remote ingestion service) is to not block waiting for a response from the ingestion service.

However the spec explicitly [disallows calling the exporter concurrently](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#exportbatch):

> Export() will never be called concurrently for the same exporter instance. Export() can be called again only after the current call returns.

So exporters need to implement the non-blocking behavior themselves.

But the only way to tell the batch span processor not to block, is to return an immediately completed `CompletableResultCode`, instead of a real `CompletableResultCode`.

Which sort of defeats the purpose of the async-friendly `CompletableResultCode` return type from `export`.

In any case though, if this is a supported pattern (returning an immediately completed `CompletableResultCode` from the exporter), it would be nice for the batch span processor to flush the span exporter after if flushes itself, or to document on the batch span processor's `flush()` method that you may still need to flush the underlying span exporter after calling that method.

trask:
Resolves #4264

This is a third option that I like better than both #4245 and #4246.


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -165,17 +170,22 @@ public String toString() {
     private volatile boolean continueWork = true;
     private final ArrayList<SpanData> batch;
 
+    private final Set<CompletableResultCode> activeExports =
+        Collections.newSetFromMap(new ConcurrentHashMap<>());
```
kubawach, line 174: NIT: ConcurrentHashMap.newKeySet()

kubawach (approved):
Looks good!

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
trask: animal sniffer said no, must not be available in android 21

trask, line 174: consider `pendingExports` similar to SimpleSpanProcessor

trask:
discussed with @anuraaga, I'm not in a rush for this change, will open a spec issue to discuss adding this setting to the spec

trask:
@open-telemetry/java-maintainers can you apply the `prototype` label to this (if you agree)?

jack-berg:
Hey @trask I'm going to go ahead and close this because its been stale for a while. Happy to re-open if / when the conversation re-opens.

PR #4287, opened 2022-03-21
Add markdown link check

trask:


.github/workflows/main-build.yml
```diff
@@ -68,9 +68,37 @@ jobs:
...
+                   -not -path './.github/*' \
+                   -not -path './node_modules/*' \
+                   -print0 \
+              | xargs -0 -n1 markdown-link-check --config .github/scripts/markdown_link_check_config.json
+
   publish-snapshots:
     name: Publish snapshots
     if: ${{ github.event_name == 'push' && github.repository == 'open-telemetry/opentelemetry-java' }}
+    # not blocking snapshot publishing on markdown-link-check because links to external urls can
+    # break at any time
```
jack-berg, line 101: This comment pertains to the `if` online 99, correct? If so, can we place the comment before line 99?

jack-berg, line 101: This comment pertains to the `if` online 99, correct? If so, can we place the comment before line 99?

jack-berg (approved):

.github/workflows/main-build.yml
trask: it's related to the `needs` below, reworded it to try to make it clearer

anuraaga (approved):

PR #4293, opened 2022-03-22
Common github action practices

trask:
I think I've gone far enough with exploring all the github actions stuff, no need to keep scattering PRs all over the three repos, and time to document the desired state and then sync everything up.

The major change for this repository (and contrib too) would be moving off of nebula and adopting the release branch automation from the instrumentation repository. @anuraaga was receptive to this idea if we implement all of the same release automation across all three repos (and we discussed an additional automation for PR backports, which I documented here but have not implemented anywhere yet).

(and yes, I am volunteering to do this work if/once we establish consensus)

cc @open-telemetry/java-instrumentation-maintainers

### DIFF
--- a/docs/common-github-actions-practices.md
+++ b/docs/common-github-actions-practices.md
@@ -1,0 +1,387 @@
+# Common GitHub Actions practices
+
+For now this is documenting desired state across the three Java repositories.
+
+Once we agree and implement, will share more broadly across OpenTelemetry
+(and surely learn and make more changes).
+
+<details>
+<summary>Table of Contents</summary>
+
+<!-- toc -->
+
+- [Have a single required status check for PRs](#have-a-single-required-status-check-for-prs)
+- [Configure "cancel-in-progress" on PR workflows](#configure-cancel-in-progress-on-pr-workflows)
+- [Prefer `gh` cli over third-party github actions for simple tasks](#prefer-gh-cli-over-third-party-github-actions-for-simple-tasks)
+- [Use github action cache to make builds faster and less flaky](#use-github-action-cache-to-make-builds-faster-and-less-flaky)
+- [Run CodeQL daily](#run-codeql-daily)
+- [Additional checks](#additional-checks)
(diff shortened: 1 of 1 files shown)


docs/common-github-actions-practices.md
```diff
@@ -0,0 +1,387 @@
...
+
+If you have multiple workflows that run on pull requests, there are a couple of options:
+
+* If they have the same `on` triggers, they can be merged into a single workflow.
+* Otherwise turn all but one of them into
+  [reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows),
+  remove the `pull_request` trigger from the reusable workflows,
+  and call the reusable workflows from the single PR workflow.
+
+## Configure "cancel-in-progress" on PR workflows
```
mateuszrzeszutek, line 63: Nit: should we enable it for the CI build too?

docs/common-github-actions-practices.md
```diff
@@ -0,0 +1,387 @@
...
+
+### Markdown lint
+
+Specification repo uses https://github.com/DavidAnson/markdownlint.
+
+Go repo uses https://github.com/avto-dev/markdown-lint.
+
+TODO
+
+### Running checks against changed files only
```
mateuszrzeszutek, line 211: Do we have any examples of doing this in any of the three repos?

docs/common-github-actions-practices.md
trask: I sort of like to have a record of CI for each commit on `main`

docs/common-github-actions-practices.md
trask: no, the go repository does though, and I thought it would be cool for link checks which are slow and slightly flaky, but then realized the fallacy of incremental link checks 😂

anuraaga (approved):

mateuszrzeszutek (approved):

jack-berg (approved):

PR #4301, opened 2022-03-24
Drop deprecations in metrics sdk

jack-berg:
Cleaning up deprecations in the metrics SDK.


exporters/otlp/metrics/src/main/java/io/opentelemetry/exporter/otlp/metrics/OtlpGrpcMetricExporterBuilder.java
```diff
@@ -51,21 +50,6 @@
...
-   * @param channel the channel to use
-   * @return this builder's instance
-   * @deprecated Use {@link #setEndpoint(String)}. If you have a use case not satisfied by the
-   *     methods on this builder, please file an issue to let us know what it is.
-   */
-  @Deprecated
-  public OtlpGrpcMetricExporterBuilder setChannel(ManagedChannel channel) {
-    delegate.setChannel(channel);
-    return this;
-  }
```
jack-berg, line 67: I would understand the argument to keep this to maintain symmetry with `OtlpGrpcSpanExporterBuilder`.

anuraaga: Yeah let's go ahead and keep this one

anuraaga (approved):

PR #4302, opened 2022-03-24
Mark metrics components as stable

jack-berg:
Drafting up marking the metrics artifacts as stable.

- Mark `:sdk:metrics`, `:exporters:otlp:metrics`, and `:exporters:otlp-http:metrics` as stable
- Include `:exporters:otlp:metrics` in `:exporters:otlp:all`
- Enable `otlp` metrics exporter by default in autoconfigure


docs/apidiffs/current_vs_latest/opentelemetry-sdk-metrics.txt
```diff
@@ -0,0 +1,266 @@
...
++++  NEW INTERFACE: PUBLIC(+) ABSTRACT(+) io.opentelemetry.sdk.metrics.export.MetricReader  (not serializable)
+	+++  CLASS FILE FORMAT VERSION: 52.0 <- n.a.
+	+++  NEW SUPERCLASS: java.lang.Object
+	+++  NEW METHOD: PUBLIC(+) ABSTRACT(+) io.opentelemetry.sdk.common.CompletableResultCode flush()
+	+++  NEW METHOD: PUBLIC(+) ABSTRACT(+) io.opentelemetry.sdk.metrics.data.AggregationTemporality getAggregationTemporality(io.opentelemetry.sdk.metrics.InstrumentType)
+	+++  NEW METHOD: PUBLIC(+) ABSTRACT(+) io.opentelemetry.sdk.common.CompletableResultCode shutdown()
++++  NEW CLASS: PUBLIC(+) FINAL(+) io.opentelemetry.sdk.metrics.export.PeriodicMetricReader  (not serializable)
+	+++  CLASS FILE FORMAT VERSION: 52.0 <- n.a.
+	+++  NEW INTERFACE: io.opentelemetry.sdk.metrics.export.MetricReader
+	+++  NEW SUPERCLASS: io.opentelemetry.sdk.metrics.internal.export.AbstractMetricReader
```
jack-berg, line 151: Could we change / remove this superclass without breaking binary compatibility? 🤔

jack-berg: I've been trying to adjust the `MetricReader` class / interface structure to try to get rid of this and I'm kind of stuck.

Some observations:
- `MetricReader` has a couple of methods we know we need, like `flush`, `shutdown`, and `getAggregationTemporality`.
- A `MetricReader` also needs some way for a meter provider to register a collection handle to it so that it can trigger collections. We've been calling this collection handle `MetricProducer`. 
- We need to hide the details of a registering a `MetricProducer` because the spec may provide additional options that change the registration contract in the future. 
- We do all that today, with `AbstractMetricReader` hiding the `MetricProducer` registration. 
- The problem is that `AbstractMetricReader` becomes part of the class hierarchy of reader implementations that we need to maintain.

How can we keep `MetricProducer` registration hidden / internal?

jack-berg: Ok, might have gotten unstuck. We shouldn't actually need to hide that meter provider needs to register something with metric reader as long as we keep the definition of what that thing is opaque.

So we could extend `MetricReader` with:
```
  void register(SomeEmptyInterface);
```
And then call register on metric reader implementations with an internal class `MetricProducer` that extends `SomeEmptyInterface` with the actual collect method:
```
public interface MetricProducer extends SomeEmptyInterface {
  public Collection<MetricData> collectAllMetrics();
}
```
Metric reader implementations of `register` need to cast `SomeEmptyInterface` to `MetricProducer`.

anuraaga (approved):

jack-berg:
Closing in favor of #4330 per discussion at the 3/31 Office Hours SIG.

PR #4310, opened 2022-03-25
Make setMinimumCollectionInterval package private

jack-berg:
I don't think the use case for `SdkMeterProviderBuilder#setMinimumCollectionInterval(Duration)` is well understood enough for us to include it in our public API.

This PR also centralizes all the utility methods for experimental features in an internal class called `SdkMeterProviderUtil`. I think this is preferable to having them scattered in several places and potentially losing track of them.


anuraaga (commented):
Have we looked into removing the minimum? I think both other repos set it to zero in tests so while not for production use cases, the API itself seems to be too important right now

jack-berg:
The minimum collection interval is a strange concept IMO, and we could consider removing it.

When collections occur more frequently than the minimum, collections of synchronous instruments are suppressed, but asynchronous instrument callbacks continue to be invoked. Presumably this is to reduce contention in systems that are performing frequent collects and also recording many synchronous measurements, but @jsuereth would know more about the original intent. 

Thinking about it more, if we do want to retain this functionality, it should default to 0. The current default behavior of suppressing collection of synchronous instruments during collections more frequent than 100ms is actually surprising and feels arbitrary. Users should have to opt into it, and I think opting in through an experimental internal API call is reasonable.

anuraaga (approved):

PR #4315, opened 2022-03-28
PMR don't export if metrics collection is empty

### LINKED ISSUE #4313: Flaky test PeriodicMetricReaderTest
https://github.com/open-telemetry/opentelemetry-java/runs/5702200370?check_suite_focus=true

```
PeriodicMetricReaderTest > periodicExport() FAILED
    java.lang.AssertionError: 
    Expecting actual:
      [[]]
    to contain exactly (and in same order):
      [[ImmutableMetricData{resource=Resource{schemaUrl=null, attributes={}}, instrumentationScopeInfo=InstrumentationScopeInfo{name=PeriodicMetricReaderTest, version=null, schemaUrl=null}, name=my metric, description=my metric description, unit=us, type=LONG_SUM, data=ImmutableSumData{points=[ImmutableLongPointData{startEpochNanos=1000, epochNanos=3000, attributes={}, exemplars=[], value=1234567}], monotonic=true, aggregationTemporality=CUMULATIVE}}]]
    but some elements were not found:
      [[ImmutableMetricData{resource=Resource{schemaUrl=null, attributes={}}, instrumentationScopeInfo=InstrumentationScopeInfo{name=PeriodicMetricReaderTest, version=null, schemaUrl=null}, name=my metric, description=my metric description, unit=us, type=LONG_SUM, data=ImmutableSumData{points=[ImmutableLongPointData{startEpochNanos=1000, epochNanos=3000, attributes={}, exemplars=[], value=1234567}], monotonic=true, aggregationTemporality=CUMULATIVE}}]]
    and others were not expected:
      [[]]
        at io.opentelemetry.sdk.metrics.export.PeriodicMetricReaderTest.periodicExport(PeriodicMetricReaderTest.java:104)
```

@jack-berg The metric producer change affected the test so curious if that could be related to this or just coincidence

jack-berg:
Resolves #4313.

After the changes I made to remove `MetricReaderFactory`, the `PeriodicMetricReader` schedule starts once `PeriodicMetricReaderBuilder#build()` is called. This can cause `PeriodicMetricReader` to pull from a no-op `MetricProducer` until the `SdkMeterProvider` registers a proper `MetricProducer`. I believe the test flake in #4313 is caused by this.

To fix it, I've adjusted the `PeriodicMetricReader` to only start its scheduled collection after `SdkMeterProvider` registers a proper `MetricProducer`. This makes sense because there's no point in running a background collection until a producer has been registered. 

I also noticed that `PeriodicMetricReader` calls `MetricExporter#export(..)` even when the collection did not produce metrics. I think this is strange for exporter implementations so I've adjusted it to skip calling export if there are no metrics. This is consistent with the behavior of `BatchSpanProcessor` and `BatchLogProcessor`.

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/PeriodicMetricReader.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/PeriodicMetricReader.java
@@ -128,7 +128,7 @@ public final class PeriodicMetricReader extends AbstractMetricReader {
       if (exportAvailable.compareAndSet(true, false)) {
         try {
           Collection<MetricData> metricData = getMetricProducer().collectAllMetrics();
-          if (metricData.size() == 0) {
+          if (metricData.isEmpty()) {
             logger.log(Level.FINE, "No metric data to export - skipping export.");
             flushResult.succeed();
             exportAvailable.set(true);


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/PeriodicMetricReader.java
```diff
@@ -115,15 +127,22 @@ CompletableResultCode doRun() {
...
-          result.whenComplete(
-              () -> {
-                if (!result.isSuccess()) {
-                  logger.log(Level.FINE, "Exporter failed");
-                }
-                flushResult.succeed();
-                exportAvailable.set(true);
-              });
+          Collection<MetricData> metricData = getMetricProducer().collectAllMetrics();
+          if (metricData.size() == 0) {
```
anuraaga, line 131:
```suggestion
          if (metricData.isEmpty()) {
```

anuraaga (approved):

Issue #4267, opened 2022-03-15
Expose histogram min and max


jack-berg:
Originally discussed in CNCF [slack](https://cloud-native.slack.com/archives/C014L2KCTE3/p1647365096522509).

- [ ] Add getters for min and max to `HistogramPointData`. Note, this is useful on its own since non-OTLP exporters can take advantage of these fields. 
- [ ] Export min and max via OTLP once [available](https://github.com/open-telemetry/opentelemetry-proto/pull/279) in the OTLP proto

PR #4317, opened 2022-03-29
Add min and max to HistogramPointData

jack-berg:
Related to #4267.

Figure it would be good to make these part of the metric data model before stabilization. Not strictly necessary to do now because they're optional and could be later added with a default of `null`.

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/HistogramPointData.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/HistogramPointData.java
@@ -6,7 +6,6 @@
 package io.opentelemetry.sdk.metrics.data;
 
 import java.util.List;
-import javax.annotation.Nullable;
 
 /**
  * A histogram metric point.
@@ -29,21 +28,17 @@ public interface HistogramPointData extends PointData {
    */
   long getCount();
 
-  /**
-   * The min of all measurements recorded.
-   *
-   * @return the min of recorded measurements, or {@code null} if not available.
-   */
-  @Nullable
-  Double getMin();
+  /** Return {@code true} if {@link #getMin()} is set. */
+  boolean hasMin();
 
-  /**
-   * The max of all measurements recorded.
-   *
-   * @return the max of recorded measurements, or {@code null} if not available.
-   */
-  @Nullable
-  Double getMax();
+  /** The min of all measurements recorded, if {@link #hasMin()} is {@code true}. */
+  double getMin();
+
+  /** Return {@code true} if {@link #getMax()} is set. */
+  boolean hasMax();
+
+  /** The max of all measurements recorded, if {@link #hasMax()} is {@code true}. */
+  double getMax();
 
   /**
    * The bucket boundaries. For a Histogram with N defined boundaries, e.g, [x, y, z]. There are N+1
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/HistogramPointData.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/HistogramPointData.java
@@ -31,13 +31,19 @@ public interface HistogramPointData extends PointData {
   /** Return {@code true} if {@link #getMin()} is set. */
   boolean hasMin();
 
-  /** The min of all measurements recorded, if {@link #hasMin()} is {@code true}. */
+  /**
+   * The min of all measurements recorded, if {@link #hasMin()} is {@code true}. If {@link
+   * #hasMin()} is {@code false}, the response should be ignored.
+   */
   double getMin();
 
   /** Return {@code true} if {@link #getMax()} is set. */
   boolean hasMax();
 
-  /** The max of all measurements recorded, if {@link #hasMax()} is {@code true}. */
+  /**
+   * The max of all measurements recorded, if {@link #hasMax()} is {@code true}. If {@link
+   * #hasMax()} is {@code false}, the response should be ignored.
+   */
   double getMax();
 
   /**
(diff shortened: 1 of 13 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/HistogramPointData.java
```diff
@@ -28,6 +29,22 @@
    */
   long getCount();
 
+  /**
+   * The min of all measurements recorded.
+   *
+   * @return the min of recorded measurements, or {@code null} if not available.
+   */
+  @Nullable
```
jack-berg, line 37: Arguably we should add `@Nullable` to `getSum()` as wel since its optional according to the [proto](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L462).

Although, its not (currently) optional in the [data model](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md?plain=1#L499). Wondering if that's an oversight or intentional.

anuraaga: I guess we don't need to make these nullable either - we always populate it anyways, I'm not sure if we get any practical flexibility by it being nullable (i.e., we could potentially decide to not populate it in the future, but why would we?)

anuraaga: I think I would treat the optional on whether an SDK supports it or not - but if we're adding support, we don't need to match that with a Nullable, we are supporting the optional feature all the time

anuraaga: Also tangentially related, I brought up the issue of understanding the relationship between the proto and data model in https://github.com/open-telemetry/opentelemetry-specification/pull/2429#issuecomment-1082585856. I don't think it necessarily always needs to be the case that both match up, but there can probably be some more documentation to understand these cases better.

jack-berg: There are a couple of situations where we may want to leave the min / max values unset:
- When bridging metrics from another system. For example, opencensus histograms dont have a min / max so I've set min / max to `null` in the opencensus shim.
- Taking the diff between two histograms. The asynchronous instruments are assumed to observe cumulative values, and so if we need to export in delta format, we need to take the diff. Min and max have no useful values when computing from cumulative to delta. Practically speaking, I don't think it would make sense to configure the aggregation to be histogram for any of the current async instruments instruments  (and in fact, the current output when you do such a configuration is surprising and arguably wrong). Maybe in the future there might be other observable instruments that do make more sense as a histogram, and for which a diff is necessary. 

Along similar lines, if we wanted to be be able to offer any sort of open metrics bridge, or in general support metrics systems that allow negative recordings for histograms, we'd need some way to indicate that `sum` is not set, else we break the monotonicity assumption of `sum`.

jack-berg: We discussed in the 3/31 Office Hours using `hasMin()` and `hasMax()` instead of `@Nullable`. I prefer this and have added a commit to switch to this strategy. However, I've kept `@Nullable` on the internal `HistogramAccumulation` `getMin` and `getMax` methods because its makes the code a bit lit verbose. Don't feel strongly about it.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/HistogramPointData.java
```diff
@@ -28,6 +28,18 @@
...
+  /** Return {@code true} if {@link #getMin()} is set. */
+  boolean hasMin();
+
+  /** The min of all measurements recorded, if {@link #hasMin()} is {@code true}. */
+  double getMin();
+
+  /** Return {@code true} if {@link #getMax()} is set. */
+  boolean hasMax();
+
+  /** The max of all measurements recorded, if {@link #hasMax()} is {@code true}. */
```
jkwatson, line 40: worth a comment about what you get if `hasMax()` isn't true?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/HistogramAccumulation.java
```diff
@@ -37,6 +44,22 @@ static HistogramAccumulation create(double sum, long[] counts, List<ExemplarData
    */
   abstract double getSum();
 
+  /**
+   * The min of all measurements recorded.
+   *
+   * @return the min of recorded measurements, or {@code null} if not available.
+   */
+  @Nullable
```
anuraaga, line 52: Since this is internal, it's probably worth having it match our expectations. They appear to be `boolean hasMinMax()` and unboxed for the others, since we always have both min/max or neither. Probably same or less code than the `applyToNullable` pattern if doing that too

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/data/HistogramPointData.java
jack-berg: I've been giving more thought to whether or not we should add `boolean hasSum()` to reflect that `sum` is required in the data model, but optional in the [proto](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L462), and I don't think its necessary:

- `hasSum()` would be false if bridging a histogram from a metrics system that allows negative measurements. We don't have this bridging use case today, and if we did, we could add a new `default boolean hasSum() { return true; }` and be ok.
- In the future, otel histograms instruments may support negative measurements, but for now that's an unsolved problem and sum is monotonic. If histograms change to support negative measurements, there will be need to be some sort of indication that sum is non-monotonic. In this case, we wouldn't want a `hasSum()` method since sum would always be present, but a `isMonotonic()` method. 

So we have a reasonable path forward in both scenarios without having to solve it now.

anuraaga (approved):

jkwatson:
> @jkwatson any additional thoughts before merging?

:shipit:

PR #4328, opened 2022-03-31
Ensure callbacks are not called concurrently

jack-berg:
The [metrics sdk spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#observations-inside-asynchronous-callbacks) says:

> The implementation MUST complete the execution of all callbacks for a given instrument before starting a subsequent round of collection.

It's currently possible for concurrent collections to invoke a callback concurrently. This resolves that.

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/CallbackRegistration.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/CallbackRegistration.java
@@ -27,6 +27,7 @@ public class CallbackRegistration<T> {
   private static final Logger logger = Logger.getLogger(CallbackRegistration.class.getName());
 
   private final ThrottlingLogger throttlingLogger = new ThrottlingLogger(logger);
+  private final Object lock = new Object();
   private final InstrumentDescriptor instrumentDescriptor;
   private final Consumer<T> callback;
   private final T measurement;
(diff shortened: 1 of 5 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/CallbackRegistration.java
```diff
@@ -69,20 +70,23 @@ public InstrumentDescriptor getInstrumentDescriptor() {
...
-    } catch (Throwable e) {
-      propagateIfFatal(e);
-      throttlingLogger.log(
-          Level.WARNING,
-          "An exception occurred invoking callback for instrument "
-              + instrumentDescriptor.getName()
-              + ".",
-          e);
+    // A callback MUST not be called concurrently
+    synchronized (lock) {
```
breedx-splk, line 74: What do you think about tightening it up so that the `synchronized(lock){ ... }` only surrounds the `callback.accept()` call? The exception handling wouldn't have to be within the lock would it?

jack-berg: No strong preference. Seems reasonable.

anuraaga:
As we chatted offline, it seems reasonable to lock the entire collection instead of only the async callbacks.

jack-berg:
After thinking about it a bit more, I believe we want to actually prevent two types of concurrent collection:
- A single reader shouldn't be able to collect concurrently
- Two different readers should be able to collect concurrently, but callbacks should not be invoked concurrently when this happens

To accomplish this I've kept the lock on `CallbackRegistration`, and added on to `SdkMeterProvider.LeasedMetricProducer#collectAllMetrics()` (there is one LeasedMetricProducer per reader).

An alternative would be to add a single lock at `MeterSharedState#collectAll`, which would force collections across all readers to be sequential.

anuraaga:
> An alternative would be to add a single lock at MeterSharedState#collectAll, which would force collections across all readers to be sequential.

This was what I was leaning towards. It seems easiest to reason about and the potential downsides (concurrent Prometheus pull) seem to not be a huge problem unless we get a report in the future.

Notably I can't reason out at all what it means to have concurrent collection for delta metrics.

anuraaga:
> the potential downsides (concurrent Prometheus pull) seem to not be a huge problem unless we get a report in the future.

Also just came to mind, if we think collection could potentially be so expensive that we definitely can't have concurrent reads waiting on each other I wonder if the metrics are supposed to be returned asynchronously through callback instead of synchronously. Probably not but it would be a reasonable pattern for a slow code.

jack-berg:
> This was what I was leaning towards. It seems easiest to reason

Agreed. Adjusted - if we get feedback we can consider changing in a future version. 

> Notably I can't reason out at all what it means to have concurrent collection for delta metrics.

Metric readers should be able to read without impacting each other. Suppose you have two periodic metric readers exporting to two different backends. Each backend should get the full set of metrics, rather than the collection of the first periodic metric reader reseting metric state and impacting the second. 

> Also just came to mind, if we think collection could potentially be so expensive that we definitely can't have concurrent reads

Only situation I can think of that could take a lot of time are async calls, but the guidance on those is that they [should complete quickly](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#asynchronous-instrument-api): "Callback functions SHOULD NOT take an indefinite amount of time.

anuraaga (approved):

PR #4329, opened 2022-04-01
Only log grpc unimplemented error once.

anuraaga:
Unavailable indicates a configuration error, which may be just not disabling the signal exporter when it's not needed. There's no need to continue to spam logs for such a scenario.

### DIFF
--- a/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/grpc/DefaultGrpcExporter.java
+++ b/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/grpc/DefaultGrpcExporter.java
@@ -16,6 +16,7 @@ import io.opentelemetry.exporter.internal.marshal.Marshaler;
 import io.opentelemetry.sdk.common.CompletableResultCode;
 import io.opentelemetry.sdk.internal.ThrottlingLogger;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.checkerframework.checker.nullness.qual.Nullable;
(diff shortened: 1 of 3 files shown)


anuraaga:
Oops wrong log message

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/grpc/DefaultGrpcExporter.java
```diff
@@ -33,6 +34,9 @@
 
   private final ThrottlingLogger logger = new ThrottlingLogger(internalLogger);
 
+  // We only log unavailable once since it's a configuration issue that won't be recovered.
```
jkwatson, line 37: couldn't someone fix their collector and things would start working again?

jack-berg (approved):

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/grpc/DefaultGrpcExporter.java
jack-berg: Yes, although probably rare. Spamming the logs is probably not useful in most situations.

jkwatson: well..it won't be spammed, exactly, since we're throttling the logging. I just wonder about the one message getting lost and people not seeing it vs. having a few messages per minute that remind them things are broken with their collector and/or their configuration.

anuraaga: IMO since it's SEVERE, I think a single log is better even if throttled - it's similar to any sort of logging we'd do in the SDK configuration codepath, and for SEVERE I guess we'd expect it to get noticed.

anuraaga:
I accidentally set the upstream for this PR to open-telemetry it seems, closing this to resend.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/grpc/DefaultGrpcExporter.java
jkwatson: I guess we'll wait and see how many confused users we get and we can adjust as needed.

PR #4330, opened 2022-04-01
Enable OTLP metric export by default in autoconfigure

jack-berg:
We discussed in the 3/31 SIG that the next release would be a metrics SDK RC. This alternative to #4302 enables OTLP metric export by default in the autoconfigure module, but keeps the metrics artifacts as `alpha`.

### DIFF
--- a/sdk-extensions/autoconfigure/build.gradle.kts
+++ b/sdk-extensions/autoconfigure/build.gradle.kts
@@ -77,7 +77,6 @@ testing {
       targets {
         all {
           testTask {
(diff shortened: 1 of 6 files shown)


jkwatson (approved):

trask (approved):
heads up @open-telemetry/java-instrumentation-approvers, the current thinking is that we will do nothing and this behavior will be carried over to javaagent users in 1.13.0

#4329 (and highlighting this change in the release notes) should help mitigate this disruption for javaagent users who do not have a metrics pipeline

PR #4331, opened 2022-04-01
Remove instrument selector predicates, add instrument name wildcards

### LINKED ISSUE #4326: Consider limiting View InstrumentSelector API
`InstrumentSelectorBuilder` offers more selection criteria than is required by the [view spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#view).

The view spec describes wildcard selection by instrument name using a limited syntax of `?` and `*`. The other selection fields are exact match.

Our InstrumentSelectorBuilder offers generic selection by `Predicate<String>` for instrument name, meter name, meter version, and meter schema URL.

This is (slightly) problematic because it impacts our ability to reject view configurations that change the name AND select more than one instrument. We don't know if a view has selected more than one instrument since we only have `Predicate<String> InstrumentSelector#getInstrumentNameFilter()`. 

Are there known use cases that require the richer selection options we offer? If not, I propose we reduce scope.

jack-berg:
Resolves #4326.

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/ViewRegistry.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/ViewRegistry.java
@@ -14,6 +14,7 @@ import io.opentelemetry.sdk.metrics.internal.descriptor.InstrumentDescriptor;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.regex.Pattern;
 import javax.annotation.concurrent.Immutable;
 
 /**
@@ -28,9 +29,7 @@ public final class ViewRegistry {
   static final View DEFAULT_VIEW = View.builder().build();
   static final RegisteredView DEFAULT_REGISTERED_VIEW =
       RegisteredView.create(
-          // InstrumentSelector requires some selection criteria but the default view is a special
-          // case, so we trick it by setting a select all criteria manually.
-          InstrumentSelector.builder().setName((unused) -> true).build(),
+          InstrumentSelector.builder().setName("*").build(),
           DEFAULT_VIEW,
           AttributesProcessor.NOOP,
           SourceInfo.noSourceInfo());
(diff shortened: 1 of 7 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/ViewRegistry.java
```diff
@@ -71,17 +70,79 @@ private static boolean matchesSelector(
...
+        case '|':
+          builder.append("\\").append(c);
+          break;
+        default:
+          builder.append(c);
+          break;
+      }
+    }
+    String regex = builder.append('$').toString();
+    return Pattern.matches(regex, instrumentName);
```
jack-berg, line 132: Please take a close look at this translation logic from wildcard pattern to regex.

jkwatson: it looks ok to me. have you considered building this Pattern in the InstrumentSelector itself, rather than every time its evaluated?

jack-berg: Good point - currently, for each instrument, its computed once for each view registered. Definitely should consider caching minimally, even if its not on the hot path.

I considered adding it to InstrumentSelector, but don't love it. We want to know the raw instrument name selection when a view is configured so that we can disallow registering views that change the instrument name and select multiple instrument on the same meter, as [required by the spec](https://github.com/open-telemetry/opentelemetry-specification/blame/main/specification/metrics/sdk.md#L197-L200). Its more difficult to tell if a pattern represents an exact match or has wildcards. This sort of implies having two methods on InstrumentSelector, one for exact name match, and one for a pattern match. I figure its better to limit the API surface area and just have one.

Technically we don't need InstrumentSelector to have any public methods. Instead, it could have a package private method that returns an internal class with getters to all the values. If we did this, then we could add whatever getters we wanted without worrying about API surface area. Same thing could be applied to `View`.

jack-berg: > Technically we don't need InstrumentSelector to have any public methods. Instead, it could have a package private method that returns an internal class with getters to all the values.

Although this is inconsistent with our our configuration data classes like `SpanLimits`.

jkwatson: yeah, that would be a little weird. It probably doesn't matter much, since this *shouldn't* be on the hot path (although users are gonna user, of course).

jkwatson (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/ViewRegistry.java
```diff
@@ -67,21 +72,92 @@ public static ViewRegistryBuilder builder() {
...
+   * <p>{@code namePattern} may contain the wildcard characters {@code *} and {@code ?} with the
+   * following matching criteria:
+   *
+   * <ul>
+   *   <li>{@code *} matches 0 or more instances of any character
+   *   <li>{@code ?} matches exactly one instance of any character
+   * </ul>
+   */
+  // Visible for testing
+  boolean matchesName(String namePattern, String instrumentName) {
```
anuraaga, line 103: Ah sorry for not linking this before, can you copy this in?

Oops sorry for not linking to this earlier, can you swap this in?

https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplier.java#L358

Notable minor improvements(?) are optimizing `*`, only iterating once to find wildcard characters, and using `Pattern.quote` instead of huge switch

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/ViewRegistry.java
```diff
@@ -67,21 +72,92 @@ public static ViewRegistryBuilder builder() {
...
+    return Pattern.matches(regex, instrumentName);
+  }
+
+  /**
+   * Transform the {@code namePattern} to a regex by converting {@code *} to {@code .*}, {@code ?}
+   * to {@code .}, and escaping other regex special characters.
+   *
+   * <p>Based off implementation from: https://www.rgagnon.com/javadetails/java-0515.html
+   */
+  private static String namePatternToRegex(String namePattern) {
```
anuraaga, line 119: This should return `Pattern` (compiled), not `String`.

anuraaga: And we can consider to cache `matchesName` by having it return a `Predicate`, instead of only caching the `Pattern` codepath.

anuraaga: Also we can consider just not caching :P IMO a cache signals wanting to fully optimize the codepath so it brings up that sort of point, but for this code we may just want to mark it as not worth optimizing since not on hot path, in such a case I'd recommend not having a cache actually

jack-berg: I'm happy to add a cache later if we learn its needed.

I think it would only really make a difference in a situations with lots of views utilizing patterns, and with lots of instruments being added throughout runtime.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/ViewRegistry.java
jack-berg: Nice! I've substituted this in, but just use `Predicate` instead of the `Matcher` interface.

anuraaga (approved):

PR #4333, opened 2022-04-02
Sync release workflow

trask:
Updates the release workflow to match instrumentation and contrib release workflows.

Removes nebula plugin.

Changes the release branch naming to `release/v*`, which matches the collector branch naming, and as I understand, some tools will display branch naming like this hierarchically.


jkwatson (approved):

anuraaga (approved):

mateuszrzeszutek (approved):

trask:
Keep me in the loop for timing of running through the new release so I can be on standby for any issues I have caused.

jack-berg (approved):

PR #4334, opened 2022-04-04
Only log grpc unimplemented error once.

anuraaga:


anuraaga:
Resend of #4329

trask (approved):
I think this will be nice as we roll out otlp metric exporter on by default in the javaagent in 1.13.0

jkwatson:
I'm still worried that this will result in users who will miss the initial log entry and not understand why they are getting no metrics being reported. But, we can always revert this if we get enough confused users that it becomes a support burden.

PR #4338, opened 2022-04-04
Disallow aggregations for incompatible instruments

### LINKED ISSUE #4327: Drop views that aggregate asynchronous instruments as histograms
As discussed in this [comment](https://github.com/open-telemetry/opentelemetry-specification/pull/2462#discussion_r840004121) we can't (and currently don't) produce semantically correct histograms when views are used to change async instruments (async counter, async up down counter, async gauge) to aggregate as histograms.

I think we should get ahead of the spec. Instead of producing bad histograms we should log a warning and not produce the view.

jack-berg:
Resolves #4327.

Generalizes the idea that histograms are valid aggregations for async instruments. Each aggregation now has a `boolean isCompatibleWithInstrument(InstrumentDescriptor)` argument which is called when metric storages are being created for an  instrument. If a view is configured who's aggregation is incompatible with an instrument it matches, a warning is logged and the view is skipped.


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/DefaultAggregation.java
```diff
@@ -56,6 +56,13 @@ private static Aggregation resolve(InstrumentDescriptor instrument) {
         .createAggregator(instrumentDescriptor, exemplarFilter);
   }
 
+  @Override
+  public boolean isCompatibleWithInstrument(InstrumentDescriptor instrumentDescriptor) {
+    // This should always return true
```
jkwatson, line 61: any reason not to just return `true` then?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/LastValueAggregation.java
```diff
@@ -45,6 +46,11 @@ private LastValueAggregation() {}
     throw new IllegalArgumentException("Invalid instrument value type");
   }
 
+  @Override
+  public boolean isCompatibleWithInstrument(InstrumentDescriptor instrumentDescriptor) {
+    return instrumentDescriptor.getType() == InstrumentType.OBSERVABLE_GAUGE;
```
jkwatson, line 51: this seems overly restrictive to me. Why not allow this for other instruments? There certainly won't be harm in it, will there?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/SumAggregation.java
```diff
@@ -53,6 +53,20 @@ private SumAggregation() {}
...
+  public boolean isCompatibleWithInstrument(InstrumentDescriptor instrumentDescriptor) {
+    switch (instrumentDescriptor.getType()) {
+      case COUNTER:
+      case OBSERVABLE_COUNTER:
+      case UP_DOWN_COUNTER:
+      case OBSERVABLE_UP_DOWN_COUNTER:
+      case HISTOGRAM:
+        return true;
+      default:
+        return false;
```
jkwatson, line 66: I suspect there will be people who want to `sum` their gauge values. Not sure why we need to restrict this at this point.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/DefaultAggregation.java
jack-berg: If it returns false, then that means a maintainer of the SDK has configured an incompatible default aggregation for an instrument. It could simply return `true`, but this offers a little more protection and prevents measurements from flowing to the aggregator even in the case of an SDK bug, which hopefully never happens.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/LastValueAggregation.java
jack-berg: > There certainly won't be harm in it, will there?

I don't know, but here's where my head is at: Suppose the metrics SDK spec comes out with a table that expresses instrument - aggregation compatibility. Would it be a breaking change if we were to initially not restrict which instruments can use last value, and later add restrictions? Possibly.

I figure its easier to start off with a restrictive set of behaviors and later loosen things up, than it is to add new restrictions later.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/SumAggregation.java
jack-berg: Reasoning for this is same as from this [comment](https://github.com/open-telemetry/opentelemetry-java/pull/4338#discussion_r843056505). If we relax gauge, we should relax sum as well.

jkwatson (approved):

anuraaga (approved):

PR #4342, opened 2022-04-06
Another round of log suppression

jack-berg:
Another round of log suppression. I believe that the only remaining logs to standard error are coming from libraries that log to SLF4J, which our `@SuppressLogger` annotation isn't able to suppress. Would be great to get those suppressed as well as they're all the same SSL exceptions that look like:
```
[armeria-common-worker-nio-2-21] WARN com.linecorp.armeria.server.HttpServerPipelineConfigurator - [id: 0xad7721fe, L:/127.0.0.1:55103 ! R:/127.0.0.1:55150] Unexpected exception:
    io.netty.handler.codec.DecoderException: javax.net.ssl.SSLHandshakeException: error:10000438:SSL routines:OPENSSL_internal:TLSV1_ALERT_INTERNAL_ERROR
```

### DIFF
--- a/sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/CardinalityTest.java
+++ b/sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/CardinalityTest.java
@@ -11,6 +11,7 @@ import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.metrics.LongCounter;
 import io.opentelemetry.api.metrics.Meter;
 import io.opentelemetry.api.metrics.ObservableLongMeasurement;
+import io.opentelemetry.internal.testing.slf4j.SuppressLogger;
 import io.opentelemetry.sdk.testing.exporter.InMemoryMetricReader;
 import java.time.Duration;
 import java.util.concurrent.atomic.AtomicLong;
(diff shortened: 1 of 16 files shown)


sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/CardinalityTest.java
```diff
@@ -11,13 +11,17 @@
...
 import io.opentelemetry.sdk.testing.exporter.InMemoryMetricReader;
 import java.time.Duration;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Consumer;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
+@SuppressLogger(
+    loggerName = "io.opentelemetry.sdk.metrics.internal.state.AsynchronousMetricStorage")
+@SuppressLogger(loggerName = "io.opentelemetry.sdk.metrics.internal.state.DeltaMetricStorage")
```
jack-berg, line 24: I also extended the `@SuppressLoggert` to accept the name of logger to suppress, for cases where the class is package private.

jkwatson (approved):

anuraaga:
@jack-berg Yeah I think the HttpServerPipelineConfigurator having a log message we would actually be able to use is quite low so we can disable it here

https://github.com/open-telemetry/opentelemetry-java/blob/main/testing-internal/src/main/resources/simplelogger.properties#L3

anuraaga (approved):

PR #4346, opened 2022-04-07
Update errorprone

anuraaga:
http://errorprone.info/bugpattern/LongDoubleConversion


micrometer1-shim/src/main/java/io/opentelemetry/micrometer1shim/OpenTelemetryTimer.java
```diff
@@ -149,10 +150,10 @@ public double totalTime(TimeUnit unit) {
   private static final class MicrometerHistogramMeasurements implements Measurements {
 
     private final LongAdder count = new LongAdder();
-    private final LongAdder totalTime = new LongAdder();
+    private final DoubleAdder totalTime = new DoubleAdder();
```
anuraaga, line 153: Since micrometer bridge is all about recording doubles, figured this is more appropriate than multiple conversions

mateuszrzeszutek (approved):

jack-berg (approved):

PR #4359, opened 2022-04-08
Fix prepare release branch workflow

trask:
~will rebase this once #4357 is merged~

### DIFF
--- a/version.gradle.kts
+++ b/version.gradle.kts
@@ -1,8 +1,13 @@
+val snapshot = true
+
 allprojects {
   val release = findProperty("otel.release")
   if (release != null) {
-    version = "1.14.0-" + release + "-SNAPSHOT"
+    version = "1.14.0-" + release
   } else {
-    version = "1.14.0-SNAPSHOT"
+    version = "1.14.0"
+  }
+  if (snapshot) {
+    version += "-SNAPSHOT"
   }
 }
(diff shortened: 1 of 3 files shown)


jack-berg (approved):

version.gradle.kts
```diff
@@ -1,8 +1,13 @@
...
   if (release != null) {
-    version = "1.14.0-" + release + "-SNAPSHOT"
-  } else {
-    version = "1.14.0-SNAPSHOT"
+    v += "-" + release
   }
+  if (snapshot) {
+    v += "-SNAPSHOT"
+  }
+  version = v
```
trask, line 12: @anuraaga if you know of a more kotlin-y way to write this let me know

anuraaga: An extra line but guess this might be more kotlin-y (priority on not using `var`).

```kotlin
val ver = "1.14.0"

version = if (release != null) {
  "${ver}-${release}"
} else if (snapshot) {
  "${ver}-SNAPSHOT"
} else {
  ver
}
```

`ver + if (release != null)` ... is also an option that some may like, but I personally don't

anuraaga (approved):

mateuszrzeszutek (approved):

version.gradle.kts
trask: there's also 😢 

```
"${ver}-${release}-SNAPSHOT"
```

anuraaga: Oh! Then current seems good

PR #4369, opened 2022-04-10
More github action syncing and doc

trask:
Also adds retry for markdown link check to help reduce sporadic failures


.github/scripts/markdown-link-check-with-retry.sh
```diff
@@ -0,0 +1,13 @@
...
+  for i in 1 2 3; do
+    if markdown-link-check --config "$(dirname "$0")/markdown-link-check-config.json" \
+                           "$file"; then
+      break
+    elif [[ $i == 3 ]]; then
+      exit 1
+    fi
+    sleep 5
+  done
+done
```
mateuszrzeszutek, line 13: Nit: WDYT about extracting the retry count to a variable?
```suggestion
retry_count=3

for file in "$@"; do
  for i in $(seq 1 $retry_count); do
    if markdown-link-check --config "$(dirname "$0")/markdown-link-check-config.json" \
                           "$file"; then
      break
    elif [[ $i == $retry_count ]]; then
      exit 1
    fi
    sleep 5
  done
done
```

mateuszrzeszutek (approved):

anuraaga (approved):

jack-berg (approved):

jkwatson (approved):

jkwatson:
didn't help with the 403s on the markdown link check. :(

trask:
> didn't help with the 403s on the markdown link check. :(

that's so weird, wonder why docs.github.com is returning 403

I think I'll add 403 to markdown-link-check's `aliveStatusCodes`, seems better than just excluding docs.github.com entirely

.github/scripts/markdown-link-check-config.json
```diff
@@ -4,5 +4,6 @@
       "pattern": "^https://mvnrepository\\.com/artifact/"
     }
   ],
-  "retryOn429": true
+  "retryOn429": true,
+  "aliveStatusCodes": [ 200, 403 ]
```
jkwatson, line 8: this might deserve a comment, since it definitely looks funny

trask: yes, but where to put the comment (damn you json!)

jkwatson: oh. yeah. :(

trask:
(I'm all done with changes)

trask:
> (I'm all done with changes)

wait, now I'm wondering if we need retries with the new `403` suppression

trask:
ok since retries alone didn't solve the 403 problem, let's go with just suppressing 403s for now, and we can add retries back in the future if it proves needed

anuraaga:
For 403, we can try

```
"httpHeaders": [
    {
      "urls": ["https://github.com"],
      "headers": {
        "Authorization": "Bearer ${{ secrets.GITHUB_TOKEN }}",
      }
    }
  ],
```

GitHub often rate limits requests from the same IP when unauthenticated but just adding the token (doesn't need any particular permissions) fixes it

anuraaga:
We can also scope down the permissions on the token for just the markdown-link-check job I think (doesn't matter on PR builds where token is read only but is a risk on main branch)

https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/#:~:text=GitHub%20Actions%20now%20lets%20you,token%20when%20a%20job%20completes.

trask:
@anuraaga either I'm not doing it right, or this doesn't work for docs.github.com:

```
  ERROR: 2 dead links found!
  [✖] https://docs.github.com/en/actions/using-workflows/reusing-workflows → Status: 403
  [✖] https://docs.github.com/en/organizations/managing-access-to-your-organizations-repositories/repository-roles-for-an-organization#permissions-for-each-role → Status: 403
```

anuraaga (commented):
Aww :( Oh well

.github/scripts/markdown-link-check-config.json
```diff
@@ -1,8 +1,13 @@
...
-  "ignorePatterns": [
+  "retryOn429": true,
+  "httpHeaders": [
     {
-      "pattern": "^https://mvnrepository\\.com/artifact/"
+      "urls": [
+        "https://docs.github.com"
+      ],
+      "headers": {
+        "Authorization": "Bearer {{ env.GITHUB_TOKEN }}"
```
anuraaga, line 9: BTW any reference for this env pattern? Don't see it in the docs. Found this technique searching on github. Can try these in separate PR of course

https://github.com/ctreffs/SwiftFSOMonitor/blob/541426bdbe073080ad103edde9083350db56016a/.github/workflows/markdown-link-check.yml#L17

trask: this is where I found that: https://github.com/tcort/markdown-link-check/pull/146

anuraaga: Not sure if I'm missing any regex flag but looks like the spaces around the braces could cause a problem

https://github.com/tcort/markdown-link-check/pull/146/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R10

trask: will try 👍

anuraaga:
It would probably be an improvement anyways to switch those mvnrepository links to the actual bom files in Maven including the version :)

PR #4383, opened 2022-04-14
Return noop instrument if name is invalid

### LINKED ISSUE #4378: Enforce metric instrument naming rule
We do no validation on instrument names, but should according to the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-naming-rule).

Maybe a question to be answered by the spec, but if we add rules enforcing instrument naming rules and they are violated, what should we do? Some options:

- Throw an exception
- Log a warning but allow the invalid name to be used
- Log a warning and transform the name to a valid name
- Log a warning and return a noop instrument

jack-berg:
Resolves #4378.

I think returning a noop instrument is the least wrong solution:
- Throwing an exception is not in line with API patterns
- If we log a warning but allowed the invalid name to be used, we'd put the burden on exporters to figure out how to handle metrics that break the rules.
- If we log a warning an transform the name to be valid, we'd be making up unspecified rules, and could cause unexpected conflicts with other metrics.

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/internal/ValidationUtil.java
+++ b/api/all/src/main/java/io/opentelemetry/api/internal/ValidationUtil.java
@@ -7,6 +7,7 @@ package io.opentelemetry.api.internal;
 
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.regex.Pattern;
 import javax.annotation.concurrent.Immutable;
 
 /**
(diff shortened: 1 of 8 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeter.java
```diff
@@ -45,21 +68,44 @@ Collection<MetricData> collectAll(
...
   public DoubleGaugeBuilder gaugeBuilder(String name) {
-    return new SdkDoubleGaugeBuilder(meterProviderSharedState, meterSharedState, name);
+    return !isValidName(name)
+        ? MeterProvider.noop().meterBuilder(name).build().gaugeBuilder(name)
+        : new SdkDoubleGaugeBuilder(meterProviderSharedState, meterSharedState, name);
+  }
+
+  // Visible for testing
+  static boolean isValidName(String name) {
+    if (name == null || !VALID_INSTRUMENT_NAME.matcher(name).matches()) {
```
jkwatson, line 99: is it worth putting this warning in the API as well, so instrumentation authors will know that they are doing something wrong, even without an SDK in place?

jack-berg: By API, do you mean the noop implementation `DefaultMeter`?

jkwatson: yes

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeter.java
```diff
@@ -10,15 +10,38 @@
...
+   *
+   * <ul>
+   *   <li>They are not null or empty strings.
+   *   <li>They are case-insensitive, ASCII strings.
+   *   <li>The first character must be an alphabetic character.
+   *   <li>Subsequent characters must belong to the alphanumeric characters, '_', '.', and '-'.
+   *   <li>They can have a maximum length of 63 characters.
+   * </ul>
+   */
+  private static final String VALID_INSTRUMENT_PATTERN =
```
anuraaga, line 38: Nit that it's a bit mysterious for _PATTERN to not be the Pattern. I think this is split out mainly to use in the log message but I feel this regex-fu will not be all that helpful in a log message and would probably write prose instead, maybe "must be a non-empty string up to 63 characters including alphanumeric, _, ., -, and starting with a letter". Or something like that

mateuszrzeszutek: (Also, it this is about a log message you could simply use `Pattern#pattern()` or `Pattern#toString()`)

api/all/src/main/java/io/opentelemetry/api/internal/ValidationUtil.java
```diff
@@ -18,12 +19,50 @@
 @Immutable
 public final class ValidationUtil {
 
-  private ValidationUtil() {}
+  public static final String API_USAGE_LOGGER_NAME = "io.opentelemetry.ApiUsageLogging";
```
jack-berg, line 22: Seemed appropriate to log to the same logger as other API usage violations. However, instrument name violations are logged at level warning instead of at finest. I've also omitted the stack trace for now, though I could be convinced to include it.

anuraaga: I'd add stack trace - may as well make it even easier to find the registration within potential string mangling through variables.

So can probably add a method to `ValidationUtil` that accepts level

api/all/src/main/java/io/opentelemetry/api/metrics/DefaultMeter.java
```diff
@@ -51,21 +51,25 @@ static Meter getInstance() {
 
   @Override
   public LongCounterBuilder counterBuilder(String name) {
+    ValidationUtil.isValidInstrumentName(name);
```
anuraaga, line 54: There doesn't seem to be a perfect name for this method but I think `checkValidInstrumentationName` returning a `boolean` is still a bit better than this seemingly-noop statement.

anuraaga (approved):

PR #4384, opened 2022-04-14
Set unit to 1 if invalid

### LINKED ISSUE #4379: Enforce metric instrument unit requirements
We do no validation on instrument units, but should according to the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-unit).

Similar to the metric name validation issue #4378, we could do a number of different things if a unit violates the rule. However, we may choose to do something different given that unit is optional and doesn't play as crucial a role in the identity of a metric. For example, we could decide to omit the unit if invalid and continue to export data - the same option isn't available for instrument name.

jack-berg:
Resolves #4379.

### DIFF
--- a/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractInstrumentBuilder.java
+++ b/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractInstrumentBuilder.java
@@ -12,11 +12,18 @@ import io.opentelemetry.sdk.metrics.internal.state.CallbackRegistration;
 import io.opentelemetry.sdk.metrics.internal.state.MeterProviderSharedState;
 import io.opentelemetry.sdk.metrics.internal.state.MeterSharedState;
 import io.opentelemetry.sdk.metrics.internal.state.WriteableMetricStorage;
+import java.nio.charset.StandardCharsets;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /** Helper to make implementing builders easier. */
 abstract class AbstractInstrumentBuilder<BuilderT extends AbstractInstrumentBuilder<?>> {
+
+  static final String DEFAULT_UNIT = "1";
+
+  private static final Logger logger = Logger.getLogger(AbstractInstrumentBuilder.class.getName());
 
   private final MeterProviderSharedState meterProviderSharedState;
   private String description;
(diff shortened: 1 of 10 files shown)


jkwatson (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractInstrumentBuilder.java
```diff
@@ -51,7 +58,32 @@ public BuilderT setDescription(String description) {
...
+  /**
+   * Determine if the instrument unit is valid. Units must be non-null, non-blank, strings of 63 or
+   * less ASCII characters.
+   */
+  // Visible for testing
+  static boolean isValidUnit(String unit) {
+    return unit != null
+        && !unit.equals("")
+        && unit.length() < 64
+        && StandardCharsets.US_ASCII.newEncoder().canEncode(unit);
```
anuraaga, line 86: This probably needs clarification in the spec but I wouldn't allow control characters be allowed if restricting. Alternatively, given description is descriptive, I'm wondering why it's not a UTF-8 string.

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/SdkMeterTest.java
```diff
@@ -341,9 +346,53 @@ void testDoubleGauge() {
...
+    sdkMeter.gaugeBuilder("name").setUnit(unit).buildWithCallback(unused -> {});
+    sdkMeter.gaugeBuilder("name").setUnit(unit).ofLongs().buildWithCallback(unused -> {});
+
+    assertThat(instrumentBuilderLogs.getEvents())
+        .hasSize(12)
+        .allSatisfy(
+            log ->
+                assertThat(log.getMessage())
+                    .isEqualTo(
+                        "Instrument name unit \"日\" is invalid, using 1 instead. Instrument unit must be 63 or less ASCII characters."));
```
anuraaga, line 377: While the instrument name in this case is exacerbating the fact, the log message might be a bit hard to parse out. "Unit \"熊\" for instrument \"name\" is invalid"?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractInstrumentBuilder.java
jack-berg: Perhaps I misunderstand the comment. UTF-8 characters are allowed for description, but unit is more restrictive. Agree with point about the control characters. Opened this [issue](https://github.com/open-telemetry/opentelemetry-specification/issues/2496) in the spec to address.

jack-berg:
I've gone ahead and added validation in `DefaultMeter` as well. This allows you to identify instrumentation with invalid unit without the SDK installed, and is symmetric to the [instrument name validation](https://github.com/open-telemetry/opentelemetry-java/pull/4383).

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/AbstractInstrumentBuilder.java
anuraaga: Ooops, got unit and description mixed up

anuraaga (approved):

PR #4387, opened 2022-04-15
Retry markdown link check on failure

trask:
It looks like we need this after all: https://github.com/open-telemetry/opentelemetry-java/runs/6033601095?check_suite_focus=true


jkwatson (approved):
who would have guessed the hardest part of the build would be checking links?

anuraaga (approved):

PR #4395, opened 2022-04-20
Move gauge assertions into stable testing artifact.

anuraaga:
Significant differences from current API

- `hasPointsSatisfying` instead of `points()`. I found the later in practice just needs to be followed with satisfiesExactly and additional `assertThat` so we can save

https://github.com/open-telemetry/opentelemetry-java-contrib/blob/0dd50e3ab7dd48b1422f2368a58c3657f44ab685/jmx-metrics/src/test/java/io/opentelemetry/contrib/jmxmetrics/OtelHelperSynchronousMetricTest.java#L61

would become

```java
assertThat(metric)
  .hasPointsSatisfying(point -> point.hasValue(123.456))
```

- `hasDoubleGaugeSatisfying` instead of `hasDoubleGauge()`. In this case, the latter does have some savings in typing. I would like to add it later if desired and initially keep a baseline where assertion methods can be called in any order

- Attributes methods from SpanData replicated almost as-is. However, in the metric version I have removed the version that accepts `Consumer<Attributes>`, only having the specialized versions. This also means removing the word `Exactly`. Unfortunately while iterating on tracing, the more useful pattern came up later after the former version and we needed a new method for it, but ideally we would have only had one and went with it here. It means metrics assertions are not completely symmetric with tracing, but are close and I suspect the change will be better than worse. Open to changing this though to match tracing exactly though

Will followup with remaining instruments

### DIFF
--- a/sdk/metrics-testing/build.gradle.kts
+++ b/sdk/metrics-testing/build.gradle.kts
@@ -1,6 +1,5 @@
 plugins {
   id("otel.java-conventions")
-  id("otel.publish-conventions")
 }
 
 description = "OpenTelemetry Metrics SDK Testing utilities"
--- a/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/LongGaugeAssert.java
+++ b/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/LongGaugeAssert.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.sdk.testing.assertj;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.opentelemetry.sdk.metrics.data.GaugeData;
+import io.opentelemetry.sdk.metrics.data.LongPointData;
+import java.util.Arrays;
+import java.util.function.Consumer;
+import javax.annotation.Nullable;
+import org.assertj.core.api.AbstractAssert;
+
+/** Assertions for an exported long {@link GaugeData}. */
+public final class LongGaugeAssert
+    extends AbstractAssert<LongGaugeAssert, GaugeData<LongPointData>> {
+
+  LongGaugeAssert(@Nullable GaugeData<LongPointData> actual) {
+    super(actual, LongGaugeAssert.class);
+  }
+
+  /**
+   * Asserts the gauge has points matching all of the given assertions and no more, in any order.
+   */
+  @SafeVarargs
+  @SuppressWarnings("varargs")
+  public final LongGaugeAssert hasPointsSatisfying(Consumer<LongPointDataAssert>... assertions) {
+    return hasPointsSatisfying(Arrays.asList(assertions));
+  }
+
+  /**
+   * Asserts the gauge has points matching all of the given assertions and no more, in any order.
+   */
+  public LongGaugeAssert hasPointsSatisfying(
+      Iterable<? extends Consumer<LongPointDataAssert>> assertions) {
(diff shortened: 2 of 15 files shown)


sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/LongGaugeAssert.java
```diff
@@ -0,0 +1,43 @@
...
+  @SafeVarargs
+  @SuppressWarnings("varargs")
+  public final LongGaugeAssert hasPointsSatisfying(Consumer<LongPointDataAssert>... assertions) {
+    return hasPointsSatisfying(Arrays.asList(assertions));
+  }
+
+  /**
+   * Asserts the gauge has points matching all of the given assertions and no more, in any order.
+   */
+  public LongGaugeAssert hasPointsSatisfying(
```
anuraaga, line 37: I had originally combined long/double versions of assertions into one class since they were equivalent, but I wasn't able to get generic-fu to work with this sort of method so went with splitting again.

sdk/metrics-testing/build.gradle.kts
```diff
@@ -1,6 +1,5 @@
 plugins {
   id("otel.java-conventions")
-  id("otel.publish-conventions")
```
anuraaga, line 3: @jack-berg I think we may have moved packages in this artifact to match the future to help with migration. Because API changes, actually it seems to have had the opposite effect :) With the classes being named the same in old and new, the build currently have issues with collecting coverage, but anyways with the API changes having same named classes with different APIs could cause any amount of weirdness. I think we just need to live with a hard break with this release.

jkwatson:
I assume we'll also be deprecating the corresponding code in the non-stable metrics testing artifact?

sdk/metrics-testing/build.gradle.kts
jack-berg: Are you planning on merging the rest of the stuff from `metrics-testing` is subsequent PRs?

jack-berg:
I don't have enough experience with assertj to have good intuition on whether the patterns we've introduced in the API match standards. As a result, I've been hesitant to move these as stable. Will differ to @jkwatson and @anuraaga to make the call.

jkwatson:
> I don't have enough experience with assertj to have good intuition on whether the patterns we've introduced in the API match standards. As a result, I've been hesitant to move these as stable. Will differ to @jkwatson and @anuraaga to make the call.

Heh. I'm in the same boat, @jack-berg . I had never used it before this project, TBH. I trust @anuraaga do have done reasonable things here. We'll only know for sure when we get feedback from real users, I guess.

jkwatson (approved):

anuraaga:
> I assume we'll also be deprecating the corresponding code in the non-stable metrics testing artifact?

I'm not sure we can deprecate because of https://github.com/open-telemetry/opentelemetry-java/pull/4395#discussion_r853749137 unfortunately

jkwatson:
> > I assume we'll also be deprecating the corresponding code in the non-stable metrics testing artifact?
> 
> I'm not sure we can deprecate because of [#4395 (comment)](https://github.com/open-telemetry/opentelemetry-java/pull/4395#discussion_r853749137) unfortunately

I was just thinking a blanket deprecation of the old artifact. Are you thinking we just stop publishing it altogether? (I see you removed the publishing conventions, so I guess 'yes').

anuraaga:
@jkwatson Yeah stopping publishing was the easiest way to work around our build having an issue with collecting coverage for the same class from multiple artifacts (we only collect for published). I can probably fix the build for this but I figure downstream users may have the same problem with javac and it's best to just not publish for the next. Not ideal but it's alpha

PR #4402, opened 2022-04-21
fix: fix BatchSpanProcessor with non-runtime exception

### LINKED ISSUE #4348: opentelemetry-sdk-trace: BatchSpanExporter gets stuck when non-runtime exceptions are thrown (kotlin)
**Describe the bug**

We are using opentelemtry-java with kotlin and a custom exporter.
We faced an issue that our exporter threw a java.net.ConnectException which wasn't catched in the exporter.
This caused the wrapping BatchSpanExporter's export thread being killed:

```
Exception in thread "BatchSpanProcessor_WorkerThread-1" java.net.ConnectException: Connection refused (Connection refused)
  at java.base/java.net.PlainSocketImpl.socketConnect(Native Method)
    at java.base/java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:412)
    at java.base/java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:255)
    at java.base/java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:237)
    at java.base/java.net.Socket.connect(Socket.java:609)
    at java.base/java.net.Socket.connect(Socket.java:558)
    at java.base/sun.net.NetworkClient.doConnect(NetworkClient.java:182)
    at java.base/sun.net.www.http.HttpClient.openServer(HttpClient.java:474)
    at java.base/sun.net.www.http.HttpClient.openServer(HttpClient.java:569)
    at java.base/sun.net.www.http.HttpClient.<init>(HttpClient.java:242)
    at java.base/sun.net.www.http.HttpClient.New(HttpClient.java:341)
    at java.base/sun.net.www.http.HttpClient.New(HttpClient.java:362)
    at java.base/sun.net.www.protocol.http.HttpURLConnection.getNewHttpClient(HttpURLConnection.java:1253)
    at java.base/sun.net.www.protocol.http.HttpURLConnection.plainConnect0(HttpURLConnection.java:1187)
    at java.base/sun.net.www.protocol.http.HttpURLConnection.plainConnect(HttpURLConnection.java:1081)
    at java.base/sun.net.www.protocol.http.HttpURLConnection.connect(HttpURLConnection.java:1015)
    at java.base/sun.net.www.protocol.http.HttpURLConnection.getOutputStream0(HttpURLConnection.java:1367)
    at java.base/sun.net.www.protocol.http.HttpURLConnection.getOutputStream(HttpURLConnection.java:1342)
    [custom exporter stack]
    at io.opentelemetry.sdk.trace.export.BatchSpanProcessor$Worker.exportCurrentBatch(BatchSpanProcessor.java:320)
    at io.opentelemetry.sdk.trace.export.BatchSpanProcessor$Worker.run(BatchSpanProcessor.java:238)
    at java.base/java.lang.Thread.run(Thread.java:829)
```

**Steps to reproduce**

- Implement a io.opentelemetry.sdk.trace.export.SpanExporter in kotlin and throw and uncatched java.net.ConnectException in export()
- Try export any span using this exporter

https://github.com/open-telemetry/opentelemetry-java/blob/429afa1c991f2162cf4aa3e1e8c33ac8e580e825/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java#L327 should catch and handle Throwable

**What did you see instead?**

Only RuntimeExceptions are catched. This is sufficient for java, as non-runtime exceptions are ensured to be catched by compiler.

Artifacts: 
        - "io.opentelemetry:opentelemetry-api:${groupedVersions.io_opentelemetry}"
        - "io.opentelemetry:opentelemetry-sdk:${groupedVersions.io_opentelemetry}"
        - "io.opentelemetry:opentelemetry-exporter-zipkin:${groupedVersions.io_opentelemetry}"
        - "io.opentelemetry:opentelemetry-semconv:${groupedVersions.io_opentelemetry}-alpha"
        - "io.opentelemetry:opentelemetry-opencensus-shim:${groupedVersions.io_opentelemetry}-alpha"

Version: groupedVersions.io_opentelemetry = 1.10.1
How did you reference these artifacts? gradle java-platform plugin

**Environment**

Runtime: gcr.io/distroless/java

**Additional context**

We also handle errors now decently in our exporter, which is IMHO also the right place to do so.
Still, it would be great that if such an error happens, it only results in a warning log than in a stuck exporter.

wallezhang:
Fix `BatchSpanProcessor` worker thread is killed by non-runtime exception.

close #4348

### DIFF
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
@@ -122,6 +122,11 @@ public final class BatchSpanProcessor implements SpanProcessor {
     return worker.batch;
   }
 
+  // Visible for testing
+  Queue<ReadableSpan> getQueue() {
+    return worker.queue;
+  }
+
   @Override
   public String toString() {
     return "BatchSpanProcessor{"
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
@@ -324,7 +324,7 @@ public final class BatchSpanProcessor implements SpanProcessor {
         } else {
           logger.log(Level.FINE, "Exporter failed");
         }
-      } catch (RuntimeException e) {
+      } catch (Throwable e) {
         logger.log(Level.WARNING, "Exporter threw an Exception", e);
       } finally {
         batch.clear();
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
@@ -13,6 +13,7 @@ import io.opentelemetry.api.metrics.MeterProvider;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.sdk.common.CompletableResultCode;
 import io.opentelemetry.sdk.internal.DaemonThreadFactory;
+import io.opentelemetry.sdk.internal.ThrowableUtil;
 import io.opentelemetry.sdk.trace.ReadWriteSpan;
 import io.opentelemetry.sdk.trace.ReadableSpan;
 import io.opentelemetry.sdk.trace.SpanProcessor;
@@ -324,8 +325,9 @@ public final class BatchSpanProcessor implements SpanProcessor {
         } else {
           logger.log(Level.FINE, "Exporter failed");
         }
-      } catch (Throwable e) {
-        logger.log(Level.WARNING, "Exporter threw an Exception", e);
+      } catch (Throwable t) {
+        ThrowableUtil.propagateIfFatal(t);
+        logger.log(Level.WARNING, "Exporter threw an Exception", t);
       } finally {
         batch.clear();
       }
--- a/sdk/trace/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorTest.java
+++ b/sdk/trace/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorTest.java
@@ -505,6 +505,26 @@ class BatchSpanProcessorTest {
                 + "exporterTimeoutNanos=30000000000}");
   }
 
+  @Test
+  @Timeout(5)
+  void exporterThrowsNonRuntimeException() {
+    when(mockSpanExporter.export(anyList()))
+        .thenAnswer(
+            invocation -> {
+              throw new Exception("No export for you.");
+            });
+    BatchSpanProcessor batchSpanProcessor =
+        BatchSpanProcessor.builder(mockSpanExporter)
+            .setScheduleDelay(MAX_SCHEDULE_DELAY_MILLIS, TimeUnit.MILLISECONDS)
+            .build();
+    sdkTracerProvider = SdkTracerProvider.builder().addSpanProcessor(batchSpanProcessor).build();
+    createEndedSpan(SPAN_NAME_1);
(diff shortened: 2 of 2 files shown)


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -324,7 +324,7 @@ private void exportCurrentBatch() {
         } else {
           logger.log(Level.FINE, "Exporter failed");
         }
-      } catch (RuntimeException e) {
+      } catch (Throwable e) {
```
anuraaga, line 327: Can you call `ThrowableUtil.propagateIfFatal` here?

https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/common/src/main/java/io/opentelemetry/sdk/internal/ThrowableUtil.java#L19

anuraaga (commented):
Thanks @wallezhang

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
wallezhang: I think we should not propagate the exception, the worker thread will be killed as usual if call `ThrowableUtil.propagateIfFatal`.

anuraaga: Yes for fatal exceptions like OutOfMemory this needs to happen to let the runtime know of the unrecoverable problem. We always propagateIfFatal in this sort of pattern.

wallezhang: OK, it sounds reasonable.

sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java
```diff
@@ -324,8 +325,9 @@ private void exportCurrentBatch() {
         } else {
           logger.log(Level.FINE, "Exporter failed");
         }
-      } catch (RuntimeException e) {
-        logger.log(Level.WARNING, "Exporter threw an Exception", e);
+      } catch (Throwable t) {
```
anuraaga, line 328: Can you add a test case with a non-runtime exception? I think Mockito will allow throwing a non-runtime exception from a mock even if not declared, or otherwise there is this pattern

https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/common/util/Exceptions.java#L235

wallezhang: I have tried to add a unit test that made exporter throw a non-runtime exception. However mockito alert that `checked exception is invalid for this method!`. Thanks for your suggestion, I will try again.

wallezhang: Done. I add a unit test with a non-runtime exception.

jack-berg (approved):

sdk/trace/src/test/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorTest.java
```diff
@@ -505,6 +505,26 @@ void stringRepresentation() {
                 + "exporterTimeoutNanos=30000000000}");
   }
 
+  @Test
+  @Timeout(5)
+  void exporterThrowsNonRuntimeException() {
+    when(mockSpanExporter.export(anyList()))
+        .thenAnswer(
+            invocation -> {
+              throw new Exception("No export for you.");
```
anuraaga, line 514: Ah that was even easier than expected, nice

anuraaga (approved):

PR #4403, opened 2022-04-21
Have long aggregators export long exemplars

anuraaga:
While writing the assertion code, I noticed that we have a public long exemplar type but don't actually use it in the SDK. I'm not so comfortable releasing a public type without us having experience using it in internal code. It also seemed weird that a `LongPointData` can have a `DoubleExemplarData`.

This PR parameterizes many components to allow recording long exemplars from long aggregations. It ended up needing to be a huge PR - I guess we have some tight coupling between exemplars and storage, and perhaps with some time we can find a creative solution to make this a bit simpler. For now, the main change I want to get in by next month is the 3 lines of public API tweaks (ABI compatible, slightly API-incompatible) which notably makes sure long points export with long exemplars, double points with double exemplars. All the other changes were required for these 3 lines, or at least I couldn't find a workaround to get the PR sized down.

### DIFF
--- a/docs/apidiffs/current_vs_latest/opentelemetry-sdk-metrics.txt
+++ b/docs/apidiffs/current_vs_latest/opentelemetry-sdk-metrics.txt
@@ -10,6 +10,7 @@ Comparing source compatibility of  against
 	+++  NEW METHOD: PUBLIC(+) STATIC(+) io.opentelemetry.sdk.metrics.Aggregation sum()
 +++  NEW ENUM: PUBLIC(+) FINAL(+) io.opentelemetry.sdk.metrics.data.AggregationTemporality  (compatible)
 	+++  CLASS FILE FORMAT VERSION: 52.0 <- n.a.
+	+++  NEW INTERFACE: java.lang.constant.Constable
 	+++  NEW INTERFACE: java.lang.Comparable
 	+++  NEW INTERFACE: java.io.Serializable
 	+++  NEW SUPERCLASS: java.lang.Enum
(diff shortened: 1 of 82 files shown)


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/exemplar/LongFixedSizeExemplarReservoir.java
```diff
@@ -0,0 +1,144 @@
...
+import java.util.Set;
+import java.util.function.Supplier;
+import javax.annotation.Nullable;
+
+/**
+ * Base implementation for fixed-size reservoir sampling of Exemplars.
+ *
+ * <p>Additionally this implementation ONLY exports double valued exemplars.
+ */
+// TODO(anuraaga): Reduce copy-paste from AbstractDoubleFixedSizeExemplarReservoir.
```
anuraaga, line 31: I copied this for now to keep things straight forward to get the public API in and hope for cleanup in the future in this internal code

docs/apidiffs/current_vs_latest/opentelemetry-sdk-metrics.txt
```diff
@@ -10,6 +10,7 @@ Comparing source compatibility of  against
 	+++  NEW METHOD: PUBLIC(+) STATIC(+) io.opentelemetry.sdk.metrics.Aggregation sum()
 +++  NEW ENUM: PUBLIC(+) FINAL(+) io.opentelemetry.sdk.metrics.data.AggregationTemporality  (compatible)
 	+++  CLASS FILE FORMAT VERSION: 52.0 <- n.a.
+	+++  NEW INTERFACE: java.lang.constant.Constable
```
jkwatson, line 13: wut

trask: oh that is weird

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/exemplar/LongFixedSizeExemplarReservoir.java
jack-berg: I have a few ideas on how we could clean this up. I'm fine addressing it in a followup.

jack-berg (approved):

PR #4405, opened 2022-04-21
Decrease logging level for no metric data exported

laurit:
For testing agent configures `PeriodicMetricReader` to run with a short interval. Due to this it may log `No metric data to export - skipping export.` multiple times for each test. Decreasing the log level to `FINEST` should keep it out from agent debug logs.


jack-berg (approved):

mateuszrzeszutek (approved):

jkwatson (approved):

anuraaga:
@laurit Wouldn't it be better for agent to configure the logging level of this class instead of needing to change the SDK like this? `FINE` is our common log level so it's a bit weird to change just this one

laurit:
@anuraaga `FINEST` seems to be used in a couple of more places. I can revert this and suppress this logging in the agent for test runs if you believe it is better that way.

anuraaga:
@laurit I think our intent is to use FINEST when logs could be extremely spammy like on API calls. I think it is better here to stick to FINE since we know a way to deal with it in the agent.

PR #4411, opened 2022-04-22
More resilient change log merging

trask:
this would have made this succeed: https://github.com/open-telemetry/opentelemetry-java-instrumentation/runs/6133898396


jkwatson (approved):
Only one suggestion: maybe use the long form `--3way` rather than short form, for additional  clarity.

jack-berg (approved):

PR #4412, opened 2022-04-22
Change ubuntu-20.04 to ubuntu-latest

trask:
`ubuntu-latest` currently is the same as `ubuntu-20.04`, I'm not sure if this is intentionally using `ubuntu-20.04` because of previous issues when `ubuntu-latest` was bumped from 18.04 to 20.04


jkwatson:
It was intentional so we didn't get mysterious build failures if/when it gets updated. Especially since it rolls out to the workers over time and it was unpredictable what was going to be the version running the builds.

PR #4419, opened 2022-04-27
Migrate sum, histogram, summary assert to stable testing.

anuraaga:
Didn't migrate exponential histogram yet since it's still internal. Will probably migrate the assertions themselves to internal in the future.


kubawach (approved):
Nice code!

jkwatson (approved):
Big PR. I'm assuming it's just porting over the existing stuff.

PR #4420, opened 2022-04-27
adding warning to XRay extractor when traceId/spanId missing

### LINKED ISSUE #4259: AwsXrayPropagator fails to extract SpanId for actual HTTP header
**Describe the bug**
AwsXrayPropagator fails to extract Context for actual HTTP header 
`X-Amzn-Trace-Id: Root=1-622422bf-59625fe25708d4660735d8ef`.

**Steps to reproduce**
AwsXrayPropagatorTest:
```java
  @Test
  void extract_actualTraceId() {
    Map<String, String> carrier = new LinkedHashMap<>(1);
    // Actual header captured:
    carrier.put("X-Amzn-Trace-Id", "Root=1-622422bf-59625fe25708d4660735d8ef");

    Context context = xrayPropagator.extract(Context.current(), carrier, getter);
    assertThat(context).as("Context should not be root")
        .isNotEqualTo(Context.root()); // BOOM!

    SpanContext spanContext = getSpanContext(context);
    assertThat(spanContext).as("SpanContext should be valid")
        .satisfies(SpanContext::isValid);
    assertThat(spanContext).as("SpanContext should not be equal to invalid")
        .isNotEqualTo(SpanContext.getInvalid());
  }
```

I expected `xrayPropagator.extract` to return a non-root Context.

**What did you see instead?**
 `xrayPropagator.extract` returns the root Context.

> io.opentelemetry:opentelemetry-api:1.12.0
> io.opentelemetry:opentelemetry-context:1.12.0
> io.opentelemetry:opentelemetry-exporter-otlp-common:1.12.0
> io.opentelemetry:opentelemetry-exporter-otlp-trace:1.12.0
> io.opentelemetry:opentelemetry-extension-aws:1.12.0
> io.opentelemetry:opentelemetry-sdk:1.12.0
> io.opentelemetry:opentelemetry-sdk-common:1.12.0
> io.opentelemetry:opentelemetry-sdk-extension-aws:1.12.0
> io.opentelemetry:opentelemetry-sdk-logs:1.12.0-alpha
> io.opentelemetry:opentelemetry-sdk-metrics:1.12.0-alpha
> io.opentelemetry:opentelemetry-sdk-trace:1.12.0
> io.opentelemetry:opentelemetry-semconv:1.12.0-alpha
> io.opentelemetry.contrib:opentelemetry-aws-xray:1.11.0

I use SBT, but I can reproduce the problem with the unit test above directly in the `opentelemetry-java` build. 

**Environment**
Compiler: Amazon Corretto 11.0.14
OS: Windows 10 Version 21H1
Runtime (if different from JDK above): -
OS (if different from OS compiled on): -

**Additional context**
It seems that no `SpanId` can be extracted from the header, and the "default" `SpanId` `"0000000000000000"` is considered invalid.

kubawach:
As agreed, fixes #4259


extensions/aws/src/main/java/io/opentelemetry/extension/aws/AwsXrayPropagator.java
```diff
@@ -228,6 +228,10 @@ private static <C> Context getContextFromHeader(
       return context;
     }
 
+    if (spanId == null || traceId == null) {
+      logger.warning("Both traceId and spanId are required to extract a valid span context. ");
```
jkwatson, line 232: this seems like it might spam the logs under some circumstances. Perhaps even the normal circumstances if you're running in the environment where this would happen at all. We should definitely not spam the logs in the hot path of people's code.

kubawach: This is to address https://github.com/open-telemetry/opentelemetry-java/issues/4259 issue. 

TLDR: currently users are creating issues expecting that traceId is sufficient (as per XRay behaviour) to createe otel context, which isn't the case. To avoid further confusion, the idea was to add this log warning. Do you have sth better in mind? Perhaps not warning but info level?

anuraaga (commented):
Let's use FINEST log for this - we generally don't enable logs by default when the app has no practical fix, eg since it doesn't have control over what header is sent. In this case since AWS is sending the header there is literally nothing a user can do so even FINE is too high IMO

extensions/aws/src/main/java/io/opentelemetry/extension/aws/AwsXrayPropagator.java
jack-berg: [ThrottlingLogger](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/common/src/main/java/io/opentelemetry/sdk/internal/ThrottlingLogger.java)?

anuraaga: This log message will happen for every request from ELB unfortunately, and with no recourse since ELB actually needs to be updated. ThrottlingLogger may actually be good in addition to a lower log level, but do think we need to not log by default for something with no actionable item.

jkwatson: The existing implementation is in the SDK, so not available to propagator implementations. We'd need to put that into some part of the API, which is possible, but I'm not sure where it would live, exactly.

kubawach: Switched to FINEST as suggested by @anuraaga 
Whole change is really only for end user's convenience, so I believe FINEST is a good solution here.

anuraaga (approved):

PR #4421, opened 2022-04-27
Parse 0-probability sampling strategies correctly (fixes: open-telemetry/opentelemetry-java-instrumentation#5899)

chrismgrayftsinc:


jkwatson (approved):
I don't have the context to judge the correctness of this code, but it looks fine to me otherwise.

chrismgrayftsinc:
I am not authorized to merge this PR -- should I change the destination branch or will an authorized person merge it later?

jkwatson:
> I am not authorized to merge this PR -- should I change the destination branch or will an authorized person merge it later?

I'm going to let the other maintainers take a look before merging.

trask:
@chrismgrayftsinc have you confirmed that this fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5899?

another possibility @laurit brought up for https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5899 is that the SERVER span name ("operation name") is usually something very generic (like servlet path) when the server span is created (which is when sampling is applied), and then the javaagent updates the SERVER span name later on, e.g. in spring webmvc instrumentation, when we are able to identify the "route" to use for the SERVER span name

chrismgrayftsinc:
@trask I haven't confirmed that it fixes the problem completely, but running the integration test before and after my patch to the unmarshaler does confirm that it does fix a problem with operation strategies that have probability exactly 0.  So yeah, there may be other issues at play that are preventing the remote sampler from working but this at least fixes the probability 0 issue.

sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/SamplingStrategyResponseUnMarshaler.java
```diff
@@ -184,7 +184,16 @@ private static SamplingStrategyResponse.OperationSamplingStrategy parseOperation
...
+          int length = input.readRawVarint32();
+          // Default probabilistic strategy.
+          if (length == 0) {
+            SamplingStrategyResponse.ProbabilisticSamplingStrategy.Builder
+                defaultProbabilisticStrategy =
+                    new SamplingStrategyResponse.ProbabilisticSamplingStrategy.Builder();
+            defaultProbabilisticStrategy.setSamplingRate(0.0);
+            builder.setProbabilisticSamplingStrategy(defaultProbabilisticStrategy.build());
+            break;
+          }
```
jack-berg, line 196: Let's move this logic to `parseProbabilistic`.

jack-berg (approved):

PR #4423, opened 2022-04-28
Apply attribute limits to exception events

laurit:
Resolves open-telemetry/opentelemetry-java#4424


jkwatson (approved):
Seems like a reasonable solution to me. Thanks!

jack-berg (approved):

PR #4426, opened 2022-04-28
Flaky retry interceptor test

### LINKED ISSUE #4407: Flaky Test RetryInterceptorTest
Luckily have a seemingly easy to debug stack trace
```
> Task :exporters:otlp:common:test

RetryInterceptorTest > connectTimeout() FAILED
    java.lang.NullPointerException: Cannot invoke "String.equalsIgnoreCase(String)" because the return value of "java.lang.Throwable.getMessage()" is null
        at io.opentelemetry.exporter.internal.retry.RetryInterceptorTest.lambda$connectTimeout$2(RetryInterceptorTest.java:156)
        at org.assertj.core.api.AbstractAssert.matches(AbstractAssert.java:992)
        at org.assertj.core.api.AbstractAssert.matches(AbstractAssert.java:755)
        at io.opentelemetry.exporter.internal.retry.RetryInterceptorTest.connectTimeout(RetryInterceptorTest.java:154)
```

https://github.com/open-telemetry/opentelemetry-java/runs/6120597180?check_suite_focus=true

jack-berg:
Resolves #4407.

### DIFF
--- a/exporters/otlp/common/src/test/java/io/opentelemetry/exporter/internal/retry/RetryInterceptorTest.java
+++ b/exporters/otlp/common/src/test/java/io/opentelemetry/exporter/internal/retry/RetryInterceptorTest.java
@@ -26,7 +26,6 @@ import java.net.SocketTimeoutException;
 import java.time.Duration;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
-import java.util.function.Predicate;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.Response;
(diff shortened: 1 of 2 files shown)


exporters/otlp/common/src/test/java/io/opentelemetry/exporter/internal/retry/RetryInterceptorTest.java
```diff
@@ -150,10 +149,7 @@ void connectTimeout() throws Exception {
     assertThatThrownBy(
             () ->
                 client.newCall(new Request.Builder().url("http://10.255.255.1").build()).execute())
-        .isInstanceOf(SocketTimeoutException.class)
-        .matches(
-            (Predicate<Throwable>)
-                throwable -> throwable.getMessage().equalsIgnoreCase("connect timed out"));
```
anuraaga, line 156: Weird this was only null sporadically 🤷

anuraaga (approved):

exporters/otlp/common/src/test/java/io/opentelemetry/exporter/internal/retry/RetryInterceptorTest.java
laurit: The way I understand it is that there are 2 places where `SocketTimeoutException` is thrown. `SocketTimeoutException` with null message is thrown when timeout is reached before it actually attempts to connect. `SocketTimeoutException` with message is thrown a bit later. Which you get depends on how fast the test is executed.

PR #4430, opened 2022-05-02
Refactor exemplars to reduce code duplication

jack-berg:
Followup from #4403.

Main change is introduction of abstract `FixedSizeExemplarReservoir`, with `ReservoirCellSelector` interface that allows histogram and random reservoirs to use common code. 

Many files are changed because of the removal of `LongExemplarReservoir` and `DoubleExemplarReservoir`. There's now just a parameterized `ExemplarReservoir<T extends ExemplarData>` which has methods for accepting both long and double measurements. It's up to the implementation to decide whether it wants to record the long or double measurements.


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/exemplar/ReservoirCell.java
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * <p>We only allocate new objects during collection. This class should NOT cause allocations during
+ * sampling or within the synchronous metric hot-path.
+ *
+ * <p>Allocations are acceptable in the {@link #getAndResetDouble(Attributes)} and {@link
+ * #getAndResetLong(Attributes)} collection methods.
+ */
+class ReservoirCell {
+  private final Clock clock;
+  private long longValue;
```
anuraaga, line 32: Ah hadn't thought of having both fields and both getAndReset methods, it seems to cut a lot of code without significant overhead (an extra few bytes per reservoir isn't that bad in practice). Let's add some comments that SDK instruments will only populate one of long or double depending on the type of the instrument.

jack-berg: Considered a generic version of `ReservoirCell<T>` where `T` is either `Double` or `Long`, but opted for primitives to avoid the boxing / unboxing.

anuraaga (approved):

PR #4433, opened 2022-05-03
Remove Data from metric assertion class names

anuraaga:
Trace assertions have some mix of `Data` or not on whether there is a corresponding SDK class for it, but this isn't really very helpful. An assertion is an assertion and there's not much need for the word - for metrics we can clean up a bit


jkwatson (approved):
I'm fine with this change.

jack-berg (approved):

PR #4435, opened 2022-05-04
Add JSON export option

jonahaapala:
I am working on adding JSON support for the New Relic OTLP Metrics consumer and while writing functional tests (written in Java) to exercise the new protocol I noticed that there wasn't an OOTB way to configure the OkHttpExporter to send payloads as JSON. For now I've copied the classes I need and changed them like I have in this PR, but figured it would be a nice addition to the opentelemetry-java project as it's not too much code and makes testing more convenient. 

I recognize that clients will want to configure themselves to send things as Protobuf, but it might be nice to have the JSON option if only for testing. I wanted to just add the option to the OkHttpExporter (which is what we use directly in our tests) so that it wouldn't need to be exposed in the OtlpHttpMetricExporterBuilder, but the existing tests seem to exercise the OkHttpExporter via the OtlpHttpMetricExporterBuilder so I added the extra wiring to keep testing similar.


docs/apidiffs/current_vs_latest/opentelemetry-sdk-metrics.txt
```diff
@@ -10,6 +10,7 @@ Comparing source compatibility of  against
 	+++  NEW METHOD: PUBLIC(+) STATIC(+) io.opentelemetry.sdk.metrics.Aggregation sum()
 +++  NEW ENUM: PUBLIC(+) FINAL(+) io.opentelemetry.sdk.metrics.data.AggregationTemporality  (compatible)
 	+++  CLASS FILE FORMAT VERSION: 52.0 <- n.a.
+	+++  NEW INTERFACE: java.lang.constant.Constable
```
jonahaapala, line 13: Not sure where this is coming from 🤔

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/okhttp/JsonRequestBody.java
```diff
@@ -0,0 +1,43 @@
...
+
+import io.opentelemetry.exporter.internal.marshal.Marshaler;
+import io.opentelemetry.exporter.internal.marshal.MarshalerUtil;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import okhttp3.MediaType;
+import okhttp3.RequestBody;
+import okio.BufferedSink;
+
+public class JsonRequestBody extends RequestBody {
```
jonahaapala, line 16: Pretty much a copy paste of ProtoResponseBody

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/okhttp/JsonRequestBody.java
```diff
@@ -0,0 +1,43 @@
...
+  private static final MediaType JSON_MEDIA_TYPE = MediaType.parse("application/json");
+
+  private final byte[] preserializedJson;
+  private final int contentLength;
+
+  /** Creates a new {@link JsonRequestBody}. */
+  public JsonRequestBody(Marshaler marshaler) {
+    preserializedJson =
+        MarshalerUtil.preserializeJsonFields(marshaler).getBytes(StandardCharsets.UTF_8);
```
jonahaapala, line 25: Initially I wanted to save off the marshaler like ProtoResponseBody does and substitute `writeBinaryTo()` with `writeJsonTo()` in the overridden `writeTo()` method below. However there wasn't an equivalent method for getting the `contentLength` like ProtoResponseBody does in its constructor, so this seemed like the simplest way to do this with the existing code at my disposal.

jonahaapala:
I have signed the CLA, not sure when the PR will update with that fact though.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/okhttp/JsonRequestBody.java
```diff
@@ -0,0 +1,51 @@
...
+import okio.BufferedSink;
+
+public class JsonRequestBody extends RequestBody {
+  private static final MediaType JSON_MEDIA_TYPE = MediaType.parse("application/json");
+
+  private final byte[] preserializedJson;
+
+  /** Creates a new {@link JsonRequestBody}. */
+  public JsonRequestBody(Marshaler marshaler) {
+    preserializedJson = preserializeJson(marshaler);
```
jonahaapala, line 23: Initially I wanted to save off the marshaler like ProtoResponseBody does and substitute writeBinaryTo() with writeJsonTo() in the overridden writeTo() method below. However there wasn't an equivalent method for getting the contentLength like ProtoResponseBody does in its constructor, so this seemed like the simplest way to do this with the existing code at my disposal. 

I copied the implementation from `MarshalerUtil.preserializeJsonFields()` but omit the part that trims the beginning and ending brackets.

jonahaapala:
Using these changes locally I was able to verify that it exports as proper JSON that can be parsed by our consumers.

jonahaapala:
\easycla

jonahaapala:
easycla

exporters/otlp-http/metrics/src/main/java/io/opentelemetry/exporter/otlp/http/metrics/OtlpHttpMetricExporterBuilder.java
```diff
@@ -117,6 +117,11 @@ public OtlpHttpMetricExporterBuilder setAggregationTemporality(
     return this;
   }
 
+  OtlpHttpMetricExporterBuilder exportAsJson() {
```
jonahaapala, line 120: not exposing as a public API. Keeping the one on OkHttpMetricExporterBuilder public though for testing purposes.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/ExporterMetrics.java
```diff
@@ -70,4 +70,9 @@ public static ExporterMetrics createGrpcOkHttp(String type, MeterProvider meterP
   public static ExporterMetrics createHttpProtobuf(String type, MeterProvider meterProvider) {
     return new ExporterMetrics(meterProvider.get("io.opentelemetry.exporters.otlp-http"), type);
   }
+
+  public static ExporterMetrics createHttpJson(String type, MeterProvider meterProvider) {
```
anuraaga, line 74: Since this is only really for testing anyways, we probably don't need a separate metrics for it

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/okhttp/JsonRequestBody.java
anuraaga: How about just sticking with `-1` for `contentLength` without preserializing?

jonahaapala: Interesting. I initially didn't think we could leave the contentLength unset because we do some validation on our end to see that it is a positive value. But I don't see anything in the spec that indicates the `Content-Length` header must be set and I don't see anything in our code that relies on it other than for monitoring. Asking my team for any context around this choice, but so far I'm inclined to do as you suggest and return -1.

jonahaapala: Though, might I ask, is there a downside to leaving this as is? I _think_ it's just a matter of when this serialization logic gets called. It's a bit wasteful to do it this way as there is an extra copy cost to transfer the bytes into the final buffer, but I don't think by very much.

README.md
```diff
@@ -189,21 +189,21 @@ This is a **current** feature status list:
...
+| Zipkin Trace Exporter       | v<!--VERSION_STABLE-->1.15.0-SNAPSHOT<!--/VERSION_STABLE-->           |
+| OTLP Exporter (Spans)       | v<!--VERSION_STABLE-->1.15.0-SNAPSHOT<!--/VERSION_STABLE-->           |
+| Metrics API                 | v<!--VERSION_STABLE-->1.15.0-SNAPSHOT<!--/VERSION_STABLE-->           |
+| OTLP Exporter (Metrics)     | v<!--VERSION_UNSTABLE-->1.15.0-SNAPSHOT-alpha<!--/VERSION_UNSTABLE--> |
+| Metrics SDK                 | v<!--VERSION_UNSTABLE-->1.15.0-SNAPSHOT-alpha<!--/VERSION_UNSTABLE--> |
+| OTLP Exporter (Logs)        | v<!--VERSION_UNSTABLE-->1.15.0-SNAPSHOT-alpha<!--/VERSION_UNSTABLE--> |
+| Logs SDK                    | v<!--VERSION_UNSTABLE-->1.15.0-SNAPSHOT-alpha<!--/VERSION_UNSTABLE--> |
+| Prometheus Metrics Exporter | v<!--VERSION_UNSTABLE-->1.15.0-SNAPSHOT-alpha<!--/VERSION_UNSTABLE--> |
+| OpenTracing Bridge          | v<!--VERSION_UNSTABLE-->1.15.0-SNAPSHOT-alpha<!--/VERSION_UNSTABLE--> |
+| OpenCensus Bridge           | v<!--VERSION_UNSTABLE-->1.15.0-SNAPSHOT-alpha<!--/VERSION_UNSTABLE--> |
```
jonahaapala, line 206: yikes, idk what's going on with this and all of the `apidiffs` updates. This is what the `./gradlew check` did...

anuraaga: Sorry about that @jonahaapala, you got caught in a race condition with our release. I pushed a commit to this branch to pull in HEAD to fix it

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/okhttp/JsonRequestBody.java
anuraaga: It's not a big deal either way if only for tests - but OkHttp's intended usage is to "do work" on its I/O thread, so it's a bit more correct to do serialization within `writeTo`, whereas preserialization in the constructor is serializing on the hot path in principle (not if it's a test, but otherwise it would be).

anuraaga (approved):

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/okhttp/JsonRequestBody.java
jack-berg: I believe this can be package private, and final. (`ProtoResponseBody` can be package private as well.)

jack-berg (approved):
Just the minor comment about package privacy.

Issue #4400, opened 2022-04-20
Unexpected behavior for DELTA metric interval


trask:
The following code:

```
    OpenTelemetry openTelemetry =
        OpenTelemetrySdk.builder()
            .setMeterProvider(
                SdkMeterProvider.builder()
                    .registerMetricReader(
                        PeriodicMetricReader.builder(
                                LoggingMetricExporter.create(AggregationTemporality.DELTA))
                            .setInterval(Duration.ofSeconds(10))
                            .build())
                    .build())
            .build();

    openTelemetry.getMeter("test").histogramBuilder("histo1").build().record(1);
    SECONDS.sleep(15);
    openTelemetry.getMeter("test").histogramBuilder("histo2").build().record(2);
    SECONDS.sleep(15);
```

reports the following metrics:

```
INFO: metric: ImmutableMetricData{name=histo1, description=, unit=1, type=HISTOGRAM,
  data=ImmutableHistogramData{aggregationTemporality=DELTA,
    points=[ImmutableHistogramPointData{getStartEpochNanos=1650484261414000000, getEpochNanos=1650484271431000000}]
  }
}
INFO: metric: ImmutableMetricData{name=histo2, description=, unit=1, type=HISTOGRAM,
  data=ImmutableHistogramData{aggregationTemporality=DELTA,
    points=[ImmutableHistogramPointData{getStartEpochNanos=1650484261414000000, getEpochNanos=1650484281425000000}]
  }
}
```

what is unexpected (to me at least) is that the metric "interval" (getEpochNanos - getStartEpochNanos) is ~20 seconds for `histo2`, instead of approximately 10 seconds as configured.

I can see how the current behavior is technically correct, but I'm worried about user questions asking why the metric interval that we report isn't always approximately the value that they configured (e.g. via `OTEL_METRIC_EXPORT_INTERVAL`).

jack-berg:
Yeah that is a bit strange isn't it? Here's the [code](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/TemporalMetricStorage.java#L56-L57) that produces it. 

If we have previous recordings we use the start time from it. However if its the first collection, we need to set the start time to some value. We don't currently track the time of the last collection for each reader, so we take the overall start time of the sdk instead (which we do track because its needed for cumulative metrics).

anuraaga:
[...]

PR #4436, opened 2022-05-04
Fix metric sdk when multiple readers are present

jack-berg:
I’ve found a rather important conceptual flaw in the metrics SDK that prevents proper function when multiple metric readers are registered.

The problem is rooted in that there is 1 metric storage per instrument and matching view. This single metric storage is shared among multiple metric readers which each reset the storage after collection. The result is readers interfering with each other and each receiving partial measurements.

The solution in this PR is to create 1 metric storage per reader per instrument and matching view. When measurements are recorded to an instrument, they accumulate to each registered readers storage. When a reader collects metrics, it reads and resets only from the storages associated with it. Along the way, I was able to remove a fair amount of complexity.

I discovered this issue while investigating adding support for allowing metric readers (and exporters) to specify their own default aggregation for each instrument type. This solution paves the way for that as well.


sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/SdkMeterProviderTest.java
```diff
@@ -748,22 +748,21 @@ void viewSdk_capturesBaggageFromContext() {
   }
 
   @Test
-  void sdkMeterProvider_supportsMultipleCollectorsCumulative() {
```
jack-berg, line 751: These two tests were written in a way that allowed the bug to hide. Recording additional measurements with attributes reveals the true behavior. These tests as currently written fail on `main`.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
```diff
@@ -41,10 +35,8 @@ public final class SdkMeterProvider implements MeterProvider, Closeable {
 
   private final ComponentRegistry<SdkMeter> registry;
   private final MeterProviderSharedState sharedState;
-  private final Map<CollectionHandle, CollectionInfo> collectionInfoMap;
```
jack-berg, line 44: CollectionHandle and CollectionInfo had overlapping responsibilities. I've replaced them with a single class with a name which more clearly conveys its responsibilities: `RegisteredReader`.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
```diff
@@ -41,10 +35,8 @@ public final class SdkMeterProvider implements MeterProvider, Closeable {
...
+  private final List<RegisteredReader> registeredReaders;
   private final AtomicBoolean isClosed = new AtomicBoolean(false);
-  private final AtomicLong lastCollectionTimestamp;
-  private final long minimumCollectionIntervalNanos;
```
jack-berg, line 47: The `minimumCollectionIntervalNanos` we've previously talked about turned out to be central to the bug. It's true purpose appears to be trying to ensure that when multiple readers are present they each receive the same data if they collect within a narrow enough interval of time. However, I believe the mechanism to be flawed as it didn't account for correctness when the readers are on different schedules (a perfectly reasonable scenario).

The refactor negates the need for it and assists in reducing complexity.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/export/RegisteredReader.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.UUID;
+import javax.annotation.Nullable;
+
+/**
+ * Represents a {@link MetricReader} registered with {@link SdkMeterProvider}.
+ *
+ * <p>This class is internal and is hence not for public use. Its APIs are unstable and can change
+ * at any time.
+ */
+public class RegisteredReader {
```
jack-berg, line 19: A simple wrapper of `MetricReader` which is assigned a UUID to allows internal code to differentiate readers. 

Later, we may choose to use this class to track when a reader has last collected, which would assist in solving #4400.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/AsynchronousMetricStorage.java
```diff
@@ -39,17 +39,25 @@
...
+      RegisteredReader registeredReader,
       MetricDescriptor metricDescriptor,
       Aggregator<T, U> aggregator,
       AttributesProcessor attributesProcessor) {
+    this.registeredReader = registeredReader;
     this.metricDescriptor = metricDescriptor;
+    this.aggregationTemporality =
+        registeredReader
+            .getReader()
+            .getAggregationTemporality(metricDescriptor.getSourceInstrument().getType());
```
jack-berg, line 60: A small optimization to calculate the temporality one instead of each time a collection occurs.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageRegistry.java
```diff
@@ -38,7 +38,7 @@ public class MetricStorageRegistry {
   private final Map<MetricDescriptor, MetricStorage> registry = new HashMap<>();
 
   /** Returns a {@link Collection} of the registered {@link MetricStorage}. */
-  public Collection<MetricStorage> getMetrics() {
+  public Collection<MetricStorage> getStorages() {
```
jack-berg, line 41: Rename for improved clarify.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/export/RegisteredReader.java
jsuereth: As an FYI - you could continue to use `CollectionHandle` if you wanted, which is GUID (from the standpoint of the Metrics SDK), and performs similarly to using an integer ID.

They may be less needed given the overall scope here.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/CallbackRegistration.java
```diff
@@ -30,50 +33,56 @@
...
+        asyncMetricStorages,
+        callbackRegistration -> callbackRegistration.new ObservableLongMeasurementImpl());
   }
 
   public InstrumentDescriptor getInstrumentDescriptor() {
     return instrumentDescriptor;
   }
 
-  void invokeCallback() {
+  void invokeCallback(RegisteredReader reader) {
```
jsuereth, line 78: Is this method synchronized in some fashion?

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DefaultSynchronousMetricStorage.java
```diff
@@ -27,20 +27,28 @@
  */
 public final class DefaultSynchronousMetricStorage<T, U extends ExemplarData>
     implements SynchronousMetricStorage {
+  private final RegisteredReader registeredReader;
   private final MetricDescriptor metricDescriptor;
+  private final AggregationTemporality aggregationTemporality;
   private final DeltaMetricStorage<T, U> deltaMetricStorage;
```
jsuereth, line 33: If you're having ONE storage per-reader, you can actually remove the need for Delta vs. Temporal storage completely.

The assumption before was that DELTA + TEMPORAL => I can reconstitute state for any number of readers.

Given the direction this is going, you really want to GUT these, as they're only causing memory overhead (likely large GC cycles) in a more consistent manner in this change.

While i'm OOO travelling this week, would be happy to set up some time brainstorm/discuss alternatives here.

FWIW - Prior to my personal situation going sour in January, I was working on a refactor VERY similar to this one, just a bit more aggressive

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/CallbackRegistration.java
jack-berg: At the moment all collections across all readers are synchronized such that only one happens at a time. This is controlled in `MeterSharedState`, which obtains `collectLock` during collection.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DefaultSynchronousMetricStorage.java
jack-berg: Nice observation! I've pushed a commit to merge the remaining minimal logic of DeltaMetricStorage into DefaultSynchronousMetricStorage

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/export/RegisteredReader.java
jack-berg: If I understand correctly, `CollectionHandle` and `CollectionInfo` are both somewhat involved in identifying a unique reader. I didn't see the need to have two classes for that concept, and figured renaming to `RegisteredReader` was more representative of what the class does. That is, it represents a unique registered reader. Internal components can rely on `hashCode()` and `equals()` that are unique among readers, and any meta data that needs to be stored with the reader can be associated with the `RegisteredReader`.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkMeterProvider.java
```diff
@@ -56,38 +48,26 @@ public static SdkMeterProviderBuilder builder() {
...
       Resource resource,
       ViewRegistry viewRegistry,
-      ExemplarFilter exemplarFilter,
-      long minimumCollectionIntervalNanos) {
+      ExemplarFilter exemplarFilter) {
     this.sharedState =
         MeterProviderSharedState.create(clock, resource, viewRegistry, exemplarFilter);
+    this.registeredReaders = registeredReaders;
+    for (RegisteredReader registeredReader : registeredReaders) {
+      registeredReader.getReader().register(new LeasedMetricProducer(registeredReader));
```
anuraaga, line 60: Compared to the previous code, it became less obvious why we would call this seemingly circular registration. Maybe a bit clearer to make `LeasedMetricProducer` static and pass in what it uses (or even `this`, just to make it explicit what state is being registered)

anuraaga (approved):

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/export/RegisteredReader.java
```diff
@@ -0,0 +1,58 @@
...
+
+  private final UUID id = UUID.randomUUID();
```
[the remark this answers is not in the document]
kubawach: Yeah, that makes sense for me as well.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageRegistry.java
kubawach: Makes sense. It could be even something like "getAllStorages" to make it clear that all registered ones will be returned.
It may also make sense to change JavaDocs - eg line 48 to "Registers the storage..." (I can't add review comment to these lines directly).

kubawach (approved):
Looks solid, some minor comments from my side.

sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/export/RegisteredReader.java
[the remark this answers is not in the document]
jack-berg: Possible, but astronomically low probability. Still tho, an atomically incremented int produces the same effect, is more human readable, and uses less memory 😁.

Issue #4372, opened 2022-04-11
Consider removing MetricExporter.alwaysCumulative


anuraaga:
It was a bit confusing since it's easy enough to say `AggregationTemporality.Cumulative`.

Also I recall the discussion maybe coming up but how about having aggregation temporality `default` to always cumulative?

anuraaga:
> any exporter implementation would want to think through carefully

Yeah I think this is definitely a concern, but I have a notion that in a vast majority of cases, cumulative will be desired. For example, if a user just wants to implement a very simple exporter, write out all the metrics to a csv file, delta is mostly inconceivable, and this sort of use case will be by end users, not vendors, so it's not really expected for them to understand what temporality means.

> From that point, I figured it was good to have alwaysCumulative for symmetry with the deltaPreferred option.

I would probably lean towards not worrying about symmetry for this one, returning `CUMULATIVE` seems to be more intuitive than using a helper method that does the same thing.

trask:
my 2 cents:

* removing `alwaysCumulative` makes sense to me (I don't find that method makes things any less confusing)
* I've seen delta cpu metrics written to csv files, so lean towards forcing users to make a choice by not having a default implementation for getAggregationTemporality
* add more doc on getAggregationTemporality explaining the two most common implementations (this will give users somewhere to start in lieu of a default method)

jkwatson:
I actually like this method and where it is. It is going to be somewhat confusing for exporter authors to implement the interface, and providing some stock answers like this seems like a good thing for the developer experience, even if this particular one has a trivial implementation.

anuraaga:
Just as an anecdote, when I implemented the interface I found the method to provide more confusion than help for me. Why is there a method that just returns a value? And when I tried just `return alwaysCumulative()`, it didn't work since it's an interface, not an abstract class. `return MetricExporter.alwaysCumulative()` didn't really save on keystrokes and made the logic more indirect so went with `return AggregationTemporality.CUMULATIVE`.

jack-berg:
Resolved in #4438.

PR #4438, opened 2022-05-05
Add AggregationTemporalitySelector functional interface

jack-berg:
My solution to #4372.

Ensures uniformity between `MetricReader` and `MetricExporter`'s `AggregationTemporality getAggregationTemporality(InstrumentType)` methods. 

By treating `AggregationTemporality getAggregationTemporality(InstrumentType)` as a functional interface, the `alwaysCumulative()` and `deltaPreferred()` methods are more natural because you don't need to know to use them as a method reference. Also, its more natural to have setters on `OtlpHttpMetricExporterBuilder` and `OtlpGrpcMetricExporterBuilder` that accept a concrete type instead of a generic `Function<InstrumentType, AggregationTemporality>` (plus better `@Nullable` guarantees).


anuraaga (approved):
LGTM and solves my issue I think

PR #4444, opened 2022-05-07
Convert tests to from :sdk:metrics-testing to :sdk:testing

jack-berg:
Wanted to increase confidence that `:sdk:testing` has all the tools needed for making metrics assertions after realizing that internally we're still using `:sdk:metrics-testing` everywhere. So I switched all our internal usage to `:sdk:testing`.

For the most part, the tools in `:sdk:testing` suffice. In rare cases like asserting the size of point data it was possible but more awkward. In cases where we test directly test low level features like exemplar we don't have the accessors to make assertions directly (everything flows through `assertThat(MetricData)`). 

I found that `HistogramPointAssert` was missing methods to assert exemplars, so I added those.

I deleted everything in `:sdk:metrics-testing` that was no longer being used. What's left are some utilities for asserting exemplars directly, and for asserting exponential histograms.


jsuereth:
My only comment here is that some of the tests used `satisifiesExactly` and some used `anySatisfy` for points.   It seems the new testing library either doesn't support that use case, or we need some other assertions in addition here.

Specifically, some of the "satisfiesExactly" are in place to prevent certain bad behaviors, while the "anySatisfy" are because testing exact behavior is hard.

anuraaga:
We can add assertions that seem useful indeed. I wouldn't expect to need `satisfiesExactly` - our data model / OTLP doesn't specify anything about the order of points within a metric AFAIK, so presumably we wouldn't want assertions with strict ordering. `anySatisfy` on the flipside seems very imprecise as it only matches one point so also not sure about adding this - it's always possible to break out of fluent assertions to use normal assertj matching, and if we see a lot of that in the wild then we'd want to add it.

anuraaga (approved):
Thanks @jack-berg completely missed doing this

PR #4453, opened 2022-05-09
Post release 1.14.0

jack-berg:


build.gradle.kts
```diff
@@ -53,6 +53,7 @@ tasks {
   register("updateVersionInDocs") {
     group = "documentation"
     doLast {
+      val version = findProperty("release.version")
```
jack-berg, line 56: The release doc calls to run `./gradlew updateVersionInDocs -Prelease.version=x.y.z`, which doesn't work if reading the version from `version.gradle.kts`.

anuraaga (approved):

anuraaga:
@trask Even with automation would we need to approve a PR for this every time? I guess we would generally approve quickly enough (since it would be effectively self approve if the author of the PR is the bot) but just wanted to call out some risk of a timing race condition confusing contributors https://github.com/open-telemetry/opentelemetry-java/pull/4435#discussion_r868557310

trask:
@anuraaga would it solve the issue if we update the api diff files in the same autogenerated PR that updates the SNAPSHOT version on main?

anuraaga:
@trask IIUC, probably not, the diffs are generated against the `latest.release` from Maven Central, I don't think the SNAPSHOT version is used currently for it. But perhaps it can be used instead

https://github.com/open-telemetry/opentelemetry-java/blob/main/buildSrc/src/main/kotlin/otel.japicmp-conventions.gradle.kts#L26

PR #4459, opened 2022-05-11
Add metrics to OpenTelemetryRule and OpenTelemetryExtension

jack-berg:
Still need to find a way to clear metrics between tests to make metrics testing practical.


sdk/testing/src/main/java/io/opentelemetry/sdk/testing/junit4/OpenTelemetryRule.java
```diff
@@ -22,23 +27,28 @@
  * tests. This rule cannot be used with {@link org.junit.ClassRule}.
  *
  * <pre>{@code
- * // public class CoolTest {
```
anuraaga, line 25: While we may not diligently do it everywhere, the `// ` are unfortunately needed to have annotations render properly in Javadoc. Not needed when switching to snippet tag :)

anuraaga (approved):

kubawach (approved):

PR #4460, opened 2022-05-11
Add experimental method to reset SdkMeterProvider

jack-berg:
Adds a testing utility method that allows `SdkMeterProvider` to be reset, which clears all registered instruments and storages. The initially configured resource, views, and readers are left alone. Any calls to record measurements to existing instruments after reset don't do anything, and those resource can be garbage collected as soon as there are no remaining referenced.

Originally discussed in the Java SIG for assisting in testing `opentelemetry-java-instrumentation`, but also necessary for a complete solution in #4459.


anuraaga (approved):

breedx-splk (approved):
👍🏻

PR #4466, opened 2022-05-15
feat: `otel.metrics.exporter` setting support multiple values

### LINKED ISSUE #4195: Add multiple value to `otel.metrics.exporter`
`otel.traces.exporter` can be added two or more value ,but `otel.metrics.exporter` can be added only one.

wallezhang:
`otel.metrics.exporter` setting support multiple values. 

close #4195

### DIFF
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MeterProviderConfiguration.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MeterProviderConfiguration.java
@@ -6,10 +6,13 @@
 package io.opentelemetry.sdk.autoconfigure;
 
 import io.opentelemetry.sdk.autoconfigure.spi.ConfigProperties;
+import io.opentelemetry.sdk.autoconfigure.spi.ConfigurationException;
 import io.opentelemetry.sdk.metrics.SdkMeterProviderBuilder;
 import io.opentelemetry.sdk.metrics.export.MetricExporter;
 import io.opentelemetry.sdk.metrics.internal.SdkMeterProviderUtil;
 import io.opentelemetry.sdk.metrics.internal.exemplar.ExemplarFilter;
+import java.util.Collections;
+import java.util.Set;
 import java.util.function.BiFunction;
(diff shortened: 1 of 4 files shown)


sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MeterProviderConfiguration.java
```diff
@@ -40,14 +43,26 @@ static void configureMeterProvider(
...
+      exporterNames = Collections.singleton("otlp");
     }
+    exporterNames.forEach(
+        exporterName ->
+            MetricExporterConfiguration.configureExporter(
+                exporterName,
+                config,
+                serviceClassLoader,
+                meterProviderBuilder,
+                metricExporterCustomizer));
```
jack-berg, line 65: Let's break this out into its own function that returns `List<MetricReader>`:
```
  static List<MetricReader> configureMetricReaders(ConfigProperties config, ClassLoader serviceClassLoader,
      BiFunction<? super MetricExporter, ConfigProperties, ? extends MetricExporter>
          metricExporterCustomizer) {
    Set<String> exporterNames = DefaultConfigProperties.getSet(config, "otel.metrics.exporter");
    if (exporterNames.contains("none")) {
      if (exporterNames.size() > 1) {
        throw new ConfigurationException(
            "otel.metrics.exporter contains none along with other exporters");
      }
      return Collections.emptyList();
    }

    if (exporterNames.isEmpty()) {
      exporterNames = Collections.singleton("otlp");
    }

    return exporterNames.stream().map(exporterName ->
        MetricExporterConfiguration.configureExporter(exporterName, config, serviceClassLoader, metricExporterCustomizer))
            .collect(toList());
  }
```
And adjust `MetricExporterConfiguration.configureExporter` to return a `MetricReader`.

This will make it simpler to test.

wallezhang: Make sense. I'll modify it.

anuraaga (approved):

PR #4469, opened 2022-05-16
Add helper to SpanDataAssert to assert that span status satisfies a condition

HaloFour:
Allows test to assert that the status of a span meets a specific criteria instead of having to exactly match a given `StatusData`.  This is to help with shared instrumentation tests where the status description may represent the exception that is thrown, but that exception might be different depending on the exact library being instrumented.


sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/SpanDataAssert.java
```diff
@@ -455,6 +455,13 @@ public SpanDataAssert hasStatus(StatusData status) {
     return this;
   }
 
+  /** Asserts the span has a status satisfying the given condition. */
+  public SpanDataAssert hasStatusSatisfying(Consumer<StatusData> condition) {
```
anuraaga, line 459: Thanks @HaloFour - can you instead add `StatusAssert` with things like `hasCode`, `hasDescription`, and other helpers possibly `hasDescriptionMatching(String regex)`? And this can be `Consumer<StatusAssert>`. We've settled on using that pattern and find it is simpler and more fluent then just consumers on the data objects.

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/OpenTelemetryAssertions.java
```diff
@@ -41,6 +42,11 @@ public static SpanDataAssert assertThat(@Nullable SpanData spanData) {
     return new SpanDataAssert(spanData);
   }
 
+  /** Returns an assertion for {@link StatusData}. */
+  public static StatusDataAssert assertThat(@Nullable StatusData statusData) {
+    return new StatusDataAssert(statusData);
+  }
```
jack-berg, line 48: I don't think we should add a top level assertion for `StatusData`. These assertions will always be done in the context of asserting against a particular span.

sdk/testing/src/main/java/io/opentelemetry/sdk/testing/assertj/StatusDataAssert.java
```diff
@@ -0,0 +1,64 @@
...
+
+  /** Asserts that the status has a description matching the given pattern. */
+  public StatusDataAssert hasDescriptionMatching(Pattern pattern) {
+    isNotNull();
+    assertThat(actual.getDescription()).matches(pattern);
+    return this;
+  }
+
+  /** Asserts that the status has a description that satisfies the given condition. */
+  public StatusDataAssert hasDescriptionSatisfying(Consumer<StringAssert> condition) {
```
jack-berg, line 58: Do we need 3 variants for asserting the description contents? Would prefer to keep only the exact match and pattern match until there is more demand.

jack-berg (approved):

anuraaga (approved):

jkwatson:
@HaloFour can you rebase and regenerate the apidiffs? Thanks!

Issue #4467, opened 2022-05-16
SimpleSpanProcessor without sampling


jeffreye:
**Is your feature request related to a problem? Please describe.**
We want to have a SimpleSpanProcessor without sampling

**Describe the solution you'd like**
We have existing logics. Just need to add another factory method.

**Describe alternatives you've considered**
A new class

**Additional context**
We are having two different span exporters:
gRPC -> sending to a trace store ( need sampling to reduce storage requirement)
customized exporter ->  sending to ETL ( doesn't need sampling, but it may take a while to be used by devs)

jack-berg:
I believe this is solved in #6057, which added the ability for SimpleSpanProcessor / BatchSpanProcessor to export unsampled spans:
```
SimpleSpanProcessor simpleSpanProcessor = SimpleSpanProcessor.builder(spanExporter)
        .setExportUnsampledSpans(true)
        .build();

BatchSpanProcessor batchSpanProcessor = BatchSpanProcessor.builder(spanExporter)
        .setExportUnsampledSpans(true)
        .build();
```

PR #4470, opened 2022-05-16
SimpleSpanProcessor without sampling 

jeffreye:
https://github.com/open-telemetry/opentelemetry-java/issues/4467
https://github.com/open-telemetry/opentelemetry-specification/issues/2543

### DIFF
--- a/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/Sampler.java
+++ b/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/Sampler.java
@@ -90,6 +90,25 @@ public interface Sampler {
   }
 
   /**
+   * Returns a new TraceIdRatioBased {@link Sampler}. The ratio of sampling a trace is equal to that
+   * of the specified ratio.
+   *
+   * <p>The algorithm used by the {@link Sampler} is undefined, notably it may or may not use parts
+   * of the trace ID when generating a sampling decision. Currently, only the ratio of traces that
+   * are sampled can be relied on, not how the sampled traces are determined. As such, it is
+   * recommended to only use this {@link Sampler} for root spans using {@link
+   * Sampler#parentBased(Sampler)}.
+   *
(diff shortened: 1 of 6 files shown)


sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/Sampler.java
```diff
@@ -89,6 +89,25 @@ static Sampler traceIdRatioBased(double ratio) {
...
+   * are sampled can be relied on, not how the sampled traces are determined. As such, it is
+   * recommended to only use this {@link Sampler} for root spans using {@link
+   * Sampler#parentBased(Sampler)}.
+   *
+   * @param ratio The desired ratio of sampling. Must be within [0.0, 1.0].
+   * @param negativeSamplingResult The sampling result when it's a negative decision
+   * @return a new TraceIdRatioBased {@link Sampler}.
+   * @throws IllegalArgumentException if {@code ratio} is out of range
+   */
+  static Sampler traceIdRatioBased(double ratio, SamplingResult negativeSamplingResult) {
```
anuraaga, line 107: Hi @jeffreye - we originally had a knob for exporting spans regardless of sampling decision but removed it not seeing a use case for it (but forgot to remove the logic related to it). I think I follow your use case, but it seems to be predicated on this sort of sampling strategy - we would need to add it in the specification before adding it here though. Can you file an issue at https://github.com/open-telemetry/opentelemetry-specification and change this PR to draft pending spec for this?

jeffreye: I just realized it's a spec change too. Submitted here https://github.com/open-telemetry/opentelemetry-specification/issues/2543

jeffreye: Just out of curious, was `SamplingDecision.RECORD_ONLY` designed for this?

jeffreye:
on hold until https://github.com/open-telemetry/opentelemetry-specification/issues/2543

PR #4472, opened 2022-05-17
MetricReader and MetricExporter can determine default aggregation by instrument

jack-berg:
As indicated in the metrics SDK [specification](https://github.com/open-telemetry/opentelemetry-specification/blame/main/specification/metrics/sdk.md#L757), metric readers should have the ability to include the default aggregation. The default aggregation dictates the aggregation to use when an instrument is selected and no views match. As discussed in a SIG a while back, this selection is extended to the metric exporter, which `PeriodicMetricReader` passes through. 

The idea being that you can configure an exporter (e.g. the otlp grpc metric exporter) to use exponential histograms as the default histogram aggregation.

This also sets the stage for reader specific view configuration, as the implementation gives each registered metric reader its own view registry. 

I'll follow up with options to extend the build in metric exporters with corresponding configuration options.


sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/export/RegisteredReaderTest.java
```diff
@@ -20,8 +21,8 @@ class RegisteredReaderTest {
 
   @Test
   void create_UniqueIdentity() {
-    RegisteredReader registeredReader1 = RegisteredReader.create(reader);
-    RegisteredReader registeredReader2 = RegisteredReader.create(reader);
+    RegisteredReader registeredReader1 = RegisteredReader.create(reader, ViewRegistry.create());
```
jsuereth, line 24: Does this mean each reader gets its own set of views?

It seems like the "ViewRegistry"is actually the new proxy for "in-memory storage" of metrics with this PR.  I'm not sure if a different name would help here (or description).

jsuereth (approved):
This seems like a very reasonable approach to tie the in-memory storage of metrics per-reader (or exporter), and allow them to define defaults.

I think we can / should follow this up with some refactorings to optimise the new direction of per-reader storage.   There's a LOT of cruft in the original design around having ONE storage of metrics that is shared by all readers.   If we undo that, there's also a lot of cleanup we can do (and optimisation potential).

sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/internal/export/RegisteredReaderTest.java
jack-berg: > Does this mean each reader gets its own set of views?

Yes. This helps accommodate the differences that the SDK spec already allows for from reader to reader, like each being able to dictate its own aggregation temporality and (as of this PR) its own default aggregation. But it also sets us up neatly for future (currently proposed) spec changes where each reader could have its own set of views. One real world application of this would be to have different collection intervals for different instruments - maybe you want 5s granularity of JVM runtime metrics like memory usage, but 60s granularity is enough for other instruments. If each reader has its own view configuration you register two periodic metric readers. The first has a 5s interval and uses views to drop all metrics except JVM runtime. The second has a 60s interval and uses views to drop all JVM runtime metrics.

> It seems like the "ViewRegistry"is actually the new proxy for "in-memory storage" of metrics with this PR. I'm not sure if a different name would help here (or description).

The key API of view registry is `List<RegisteredView> findViews(InstrumentDescriptor, InstrumentationScopeInfo)`. This is called once per registered reader when an instrument is created. Each `RegisteredView` in the response is used to create a `MetricStorage`. Each measurement recorded from an instrument is sent to all the associated `MetricStorages`. So effectively each instrument results metric storages equal to `num readers * num views`. 

I think `ViewRegistry` is still the correct name, as it reflects its core function of defining which views should be created for each instrument. Those views ultimately result in metric storages being created, but that's not a concern of `ViewRegistry`.

jkwatson:
> Hey @anuraaga / @jkwatson wondering if you have plans to take a look at this, or opinions about merging this. It would be nice to get this and its followup #4557 merged before the 1.16.0 release.

If Josh is good with this, I'm fine with it. The addition to the SDK's APIs are minimal, and default implementations are provided in all cases. 👍🏽

jkwatson (approved):

anuraaga (approved):

PR #4486, opened 2022-05-20
Make ArrayBackedAttributes be public

JoeCqupt:
when i use jdk1.6 , i can not get Attributes instance with method `Attributes.of(...)` 
image

solution 1 :  i custom implements myself `Attributes` ;  it make code redundant

solution2： i want new instance of ArrayBackedAttributes ; but it not public class

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/common/ArrayBackedAttributes.java
+++ b/api/all/src/main/java/io/opentelemetry/api/common/ArrayBackedAttributes.java
@@ -12,7 +12,7 @@ import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
 @Immutable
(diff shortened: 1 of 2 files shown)


Oberon00:
OpenTelemetry Java has Java 8 as minimum requirements. You would run into many more problems if you tried using it on Java 6.

anuraaga:
Hi @JoeCqupt - as @Oberon00 says this project requires java 8. If you need to add tracing to a java 6 app you will probably want to use zipkin brave with the OpenTelemetry collector to convert the format.

JoeCqupt:
I'm maintaining an old project,  i can not upgrade jdk version . I can only find a way to make opentelemetry-sdk run 。 😂

jkwatson:
I'm going to close this. We aren't going to support java 6, and we don't want to expose this class unless there's a really compelling reason to do it.

PR #4487, opened 2022-05-20
Add setter for MeterProvider to JaegerGrpcSpanExporterBuilder

Donnerbart:
`GrpcExporterBuilder` supports to set a custom `MeterProvider`, which is exposed in the [OtlpGrpcSpanExporterBuilder](https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/otlp/trace/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterBuilder.java#L140). The `JaegerGrpcSpanExporterBuilder` should have the same functionality.

I could not find a dedicated tests for the `OtlpGrpcSpanExporterBuilder` using a `MeterProvider` implementation beside NOOP. I guess this is part of the underlying `GrpcExporterBuilder`.

### DIFF
--- a/exporters/jaeger/src/main/java/io/opentelemetry/exporter/jaeger/JaegerGrpcSpanExporterBuilder.java
+++ b/exporters/jaeger/src/main/java/io/opentelemetry/exporter/jaeger/JaegerGrpcSpanExporterBuilder.java
@@ -33,6 +33,7 @@ public final class JaegerGrpcSpanExporterBuilder {
   JaegerGrpcSpanExporterBuilder() {
     delegate =
         GrpcExporter.builder(
+            "jaeger",
             "span",
             DEFAULT_TIMEOUT_SECS,
             DEFAULT_ENDPOINT,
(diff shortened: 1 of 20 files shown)


jack-berg:
Collecting metrics on exporting is an experimental feature. As its currently built it won't work for the jaeger exporter because the [ExporterMetrics](https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/ExporterMetrics.java) class that's used to collect metrics assumes its collecting for otlp exporters. To support this, we need to:
1. Agree that the jaeger exporter should have metrics instrumentation.
2. Adjust `ExporterMetrics` to accept additional arguments reflecting the exporter its instrumenting.

Donnerbart:
You mean that `io.opentelemetry.exporters.otlp-grpc`, `otlp.exporter.seen`, ... in `ExporterMetrics` must be configurable to `io.opentelemetry.exporters.jaeger-grpc`, `jaeger.exporter.seen`, ..., right? I could work on that if wanted.

How do we solve the "agree" part? Is this a topic for the SIG?

Donnerbart:
The `ExporterMetrics` are adapted now, including the whole builder chain. I hope code-wise this is acceptable now. Please give me a heads-up if I have to do something to move the decision process forward.

anuraaga (approved):

anuraaga:
This seems like a good idea to me. Thanks!

jack-berg (approved):

PR #4489, opened 2022-05-23
Add experimental option to disable autoconfigured opentelemetry SDK

### LINKED ISSUE #4483: Consider a method (env. variable/property) to disable autoconfiguration

I have applications that are manually instrumented using the Java Otel API but also include the Autoconfiguration SDK extension class so that it's easy to configure the SDK via environment variables and '-D' switches.

However, there are times when we want to completely disable OpenTelemetry (e.g. to check whether any issues in our application are caused by OpenTelemetry, or compare performance with/without OpenTelemetry enabled).

For auto-instrumentation via the javaagent, there is a single "kill switch", namely `-Dotel.javaagent.enabled=false` that disables any SDK initialization, etc. However, there is no equivalent for disabling autoconfiguration SDK completely if the "autoconfiguration" SDK extension class exists on the application classpath.

A property or environment variable that can be used to disable OpenTelemetry autoconfiguration completely, e.g. `-Dotel.autoconfiguration.enabled=false`.

Even if we have our own flag to prevent any calls to the Otel autoconfigure API, any library we include which makes use of `GlobalOpenTelemetry.get()` will end up autoconfiguring the Otel SDK if the autoconfig extension class exists. By default, the autoconfigured SDK is a "working" SDK with OTLP exporters enabled for metrics and traces, rather than a "noop" OpenTelemetry instance.

**Additional context**
This issue was brought up at the May 10th 2022 Java SIG meeting. https://docs.google.com/document/d/1WK9h4p55p8ZjPkxO75-ApI9m0wfea6ENZmMoFRvXSCw/edit#heading=h.wtapn135xim7

jack-berg:
Resolves #4483.

### DIFF
--- a/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdk.java
+++ b/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdk.java
@@ -6,10 +6,11 @@
 package io.opentelemetry.sdk.autoconfigure;
 
 import com.google.auto.value.AutoValue;
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.sdk.OpenTelemetrySdk;
 import io.opentelemetry.sdk.autoconfigure.spi.ConfigProperties;
 import io.opentelemetry.sdk.resources.Resource;
-import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
 /**
@@ -25,8 +26,7 @@ public abstract class AutoConfiguredOpenTelemetrySdk {
    * Returns an {@link AutoConfiguredOpenTelemetrySdk} automatically initialized through recognized
    * system properties and environment variables.
    *
-   * <p>This will automatically set the resulting SDK as the {@link
-   * io.opentelemetry.api.GlobalOpenTelemetry} instance.
+   * <p>This will automatically set the resulting SDK as the {@link GlobalOpenTelemetry} instance.
    */
   public static AutoConfiguredOpenTelemetrySdk initialize() {
     return builder().build();
@@ -41,15 +41,23 @@ public abstract class AutoConfiguredOpenTelemetrySdk {
   }
 
   static AutoConfiguredOpenTelemetrySdk create(
-      @Nullable OpenTelemetrySdk sdk, Resource resource, ConfigProperties config) {
-    return new AutoValue_AutoConfiguredOpenTelemetrySdk(sdk, resource, config);
+      boolean isSdkEnabled, OpenTelemetrySdk sdk, Resource resource, ConfigProperties config) {
+    return new AutoValue_AutoConfiguredOpenTelemetrySdk(isSdkEnabled, sdk, resource, config);
   }
+
+  /**
+   * Returns true if the SDK is enabled. If false, {@link #getOpenTelemetrySdk()} should be ignored.
+   *
+   * <p>When the SDK is disabled and {@link
+   * AutoConfiguredOpenTelemetrySdkBuilder#setResultAsGlobal(boolean)} is {@code true}, {@link
+   * GlobalOpenTelemetry#set(OpenTelemetry)} is set to a noop.
+   */
+  public abstract boolean isSdkEnabled();
 
   /**
    * Returns the {@link OpenTelemetrySdk} that was auto-configured, or {@code null} if the SDK has
    * been disabled.
    */
-  @Nullable
   public abstract OpenTelemetrySdk getOpenTelemetrySdk();
 
   /** Returns the {@link Resource} that was auto-configured. */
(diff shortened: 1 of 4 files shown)


jkwatson (approved):
Seems reasonable. Thanks!

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdk.java
```diff
@@ -40,11 +41,15 @@ public static AutoConfiguredOpenTelemetrySdkBuilder builder() {
...
+      @Nullable OpenTelemetrySdk sdk, Resource resource, ConfigProperties config) {
     return new AutoValue_AutoConfiguredOpenTelemetrySdk(sdk, resource, config);
   }
 
-  /** Returns the {@link OpenTelemetrySdk} that was auto-configured. */
+  /**
+   * Returns the {@link OpenTelemetrySdk} that was auto-configured, or {@code null} if the SDK has
+   * been disabled.
+   */
+  @Nullable
```
anuraaga, line 52: What do you think about putting the flag in the API only for the implocit autoconfigure instead of autoconfigure actually?

https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java#L214

The nullable here seems to reduce ux of autoconfigure significantly

anuraaga: Also I believe the javaagent currently is designed with the assumption that it is either completely disabled, or has an SDK. Changing that by making this nullable will be fairly tricky actually. Probably (not certainly) many apps using autoconfigure would have similar downstream effects if autoconfiguration was requested but disabled.

I think it's probably best to leave this flag guarding of autoconfiguration itself to callers, though adding the flag for the implicit magic seems good. Alternatively, we would want to return an SDK without any exporters, etc, though we know that does still do more work than may be expected.

anuraaga: Converting the null to no-op API isn't that bad but it's still a huge difference in UX. Before no flags really affected how to use autoconfigure, now this one requires the caller to do more work even if they're not interested in the feature. That work could have been defining a flag instead without a lot more effort and not affecting the default experience. This is an experimental flag that may not make it through spec anyways (perhaps other languages have even bigger problems than we have here). So it doesn't seem appropriate for it to affect the UX like this.

jack-berg: It does feel strange to change the UX in such a significant way to accommodate an experimental property. But it does seem like the value of the feature is eroded quite a bit if it doesn't work for users of `AutoConfiguredOpenTelemetrySdk`. I'll see if I can think of any other options.

jack-berg: I tried something different in a new commit. Instead of `getOpenTelemetrySdk()` returning null, I've added a new method called `AutoConfiguredOpenTelemetrySdk#isSdkEnabled()`. The ergonomics are unchanged for someone who is not using the experiemental property, but for someone who is, `GlobalOpenTelemetry` is still set to noop and they have a way to tell if the property was set to disable the SDK.

When `isSdkEnabled() == false`, the response of `getOpenTelemetrySdk()` should be ignored, but returns `OpenTelemetrySdk.builder().build()` for lack of a better option. 

Still not a perfect solution, but maybe good enough while balancing the various arguments.

anuraaga: I'd say let's leave the isConfigured method out for now as it's new API for an experimental flag. We could add it in the future of the flag stabilizes, but I think the main problem of exporting data to a non-existent collector automatic is resolved by returning the empty SDK

jack-berg: I've removed the method in the latest commit 👍

anuraaga (approved):

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdk.java
```diff
@@ -44,7 +44,10 @@ static AutoConfiguredOpenTelemetrySdk create(
     return new AutoValue_AutoConfiguredOpenTelemetrySdk(sdk, resource, config);
   }
 
-  /** Returns the {@link OpenTelemetrySdk} that was auto-configured. */
+  /**
+   * Returns the {@link OpenTelemetrySdk} that was auto-configured, or {@code null} if the SDK has
+   * been disabled.
+   */
   public abstract OpenTelemetrySdk getOpenTelemetrySdk();
```
NthPortal, line 51: This PR does not appear to have or add a code path that can return `null`, and after several hours of poking at the current code, that fact does not seem to have changed since. Is this a bug? or did the design get changed after the doc was written?

jack-berg: Oops. The response of this method was originally nullable but [per this conversation](https://github.com/open-telemetry/opentelemetry-java/pull/4489#discussion_r879949618), we went in a different direction. Looks like the javadoc never got updated - sorry about the confusion.

I've opened #6109 to fix the comment.

Issue #4482, opened 2022-05-19
Reduce set of published artifacts


jack-berg:
We publish a lot of artifacts, and it's cumbersome for users to sort through which are needed. There's opportunity to reduce the sprawl, partially made possible by the recent stabilization of metrics. Here are some simplification opportunities:

- ~~Stop publishing `opentelemetry-context`, instead merging it into `opentelemetry-api`.~~ Won't do - this separation is intentional.
- Stop publishing signal specific sdk artifacts. We still need a separate artifact for the unstable `opentelemetry-sdk-logs` and `opentelemetry-sdk-testing`, but we can merge `opentelemetry-sdk-common`, `opentelemetry-sdk-metrics`, and `opentelemetry-sdk-metrics` into `opentelemetry-sdk`. 
- ~~Stop publishing signal specific otlp grpc artifacts. We still need a separate artifact for the unstable `opentelemetry-exporter-otlp-logs`, but we can merge `opentelemetry-exporter-otlp-trace` and `opentelemetry-exporter-otlp-metrics` into `opentelemetry-exporter-otlp`.~~ #4494.
- ~~Stop publishing signal specific otlp http artifacts, preferring a combined artifact. We still need a separate artifact for the unstable `opentelemetry-exporter-otlp-http-logs`, but can start publishing `opentelemetry-exporter-otlp-http` combining `opentelemetry-exporter-http-trace` and `opentelemetry-exporter-http-metrics`.~~ #4499 
- ~~Stop publishing `opentelemetry-exporter-jaeger-proto`. Its for internal use only for `opentelemetry-exporter-jaeger` and I believe that the jaeger exporter doesn't rely on it being published.~~ #4496.

One of the themes of simplifications is to remove artifacts for individual signals. In the past we've argued that part of the value of otel is a single set of APIs / SDKs for all three signals. This is embodied in `opentelemetry-api`, which includes the APIs for tracing and metrics. We should follow that precedent with other artifacts.

jkwatson:
> * Stop publishing `opentelemetry-context`, instead merging it into `opentelemetry-api`.

There was an explicit ask early-on to have this be a separate artifact that could be used outside the scope (heh) of the API.

[...]

PR #4492, opened 2022-05-24
Stop publishing jaeger protos

jack-berg:
Related to #4482.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,7 +208,7 @@ dependency as follows, replacing `{{artifact-id}}` with the value from the "Arti
   implementation('io.opentelemetry:{{artifact-id}}')
 ```
 
-### Bill of Material
+### Bill of Materials
 
 | Component                                    | Description                            | Artifact ID               | Version                                                     |
 |----------------------------------------------|----------------------------------------|---------------------------|-------------------------------------------------------------|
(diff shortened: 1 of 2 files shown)


jkwatson (approved):

anuraaga:
Oh missed Jaeger on the original issue. The only reason we publish them is we had published them as non-alpha mistakenly already. It's not semver compliant to stop publishing them I think. It's fine to stop documenting them though.

jkwatson:
> Oh missed Jaeger on the original issue. The only reason we publish them is we had published them as non-alpha mistakenly already. It's not semver compliant to stop publishing them I think. It's fine to stop documenting them though.

One could argue if you never change an artifact, it will never have any breaking changes!

anuraaga:
It would mean one artifact that people may be using happens to not be managed by the BOM though looking very similar to our others. I feel the cost of publishing an artifact is not high enough to create this inconsistency if it's just removed from documentation.

anuraaga:
I think there is also some mindset / philosophy - if we publish Java code as stable, we commit do continuing to publish it until a next major version. I guess we don't document this particular case in VERSIONING, but doing that just seems to be more robust in terms of promising semver.

jkwatson:
I agree that the cost of this artifact is quite low to us, and I suspect it will never end up actually needing any maintenance. We should put a disclaimer in the docs that it will not have any changes in the future, but will continue to be published as-is.

PR #4493, opened 2022-05-24
Improve metrics API javadoc

jack-berg:
Over time I've noticed some opportunities to improve the metrics API javadoc. This PR:
- Fixes various types
- Standardizes language (i.e. how we refer to meters, instruments, measurements, etc)
- Includes notes about parameter constraints (e.g. histograms values must be non-negative

### DIFF
--- a/api/all/src/main/java/io/opentelemetry/api/metrics/DefaultMeter.java
+++ b/api/all/src/main/java/io/opentelemetry/api/metrics/DefaultMeter.java
@@ -11,7 +11,11 @@ import io.opentelemetry.context.Context;
 import java.util.function.Consumer;
 import javax.annotation.concurrent.ThreadSafe;
 
-/** No-op implementation of {@link Meter}. */
+/**
+ * No-op implementation of {@link Meter}.
+ *
+ * <p>This implementation should induce as close to zero overhead as possible.
+ */
 @ThreadSafe
 class DefaultMeter implements Meter {
 
--- a/api/all/src/main/java/io/opentelemetry/api/metrics/DefaultMeter.java
+++ b/api/all/src/main/java/io/opentelemetry/api/metrics/DefaultMeter.java
@@ -11,28 +11,7 @@ import io.opentelemetry.context.Context;
 import java.util.function.Consumer;
 import javax.annotation.concurrent.ThreadSafe;
 
-/**
- * No-op implementations of {@link Meter}.
- *
- * <p>This implementation should induce as close to zero overhead as possible.
- *
- * <p>A few notes from the specification on allowed behaviors leading to this design [<a
- * href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument">Instrument
- * Spec</a>]:
- *
- * <ul>
- *   <li>Multiple Instruments with the same name under the same Meter MUST return an error
- *   <li>Different Meters MUST be treated as separate namespaces
- *   <li>Implementations MUST NOT require users to repeatedly obtain a Meter again with the same
- *       name+version+schema_url to pick up configuration changes. This can be achieved either by
- *       allowing to work with an outdated configuration or by ensuring that new configuration
- *       applies also to previously returned Meters.
- *   <li>A MeterProvider could also return a no-op Meter here if application owners configure the
- *       SDK to suppress telemetry produced by this library
- *   <li>In case an invalid name (null or empty string) is specified, a working Meter implementation
- *       MUST be returned as a fallback rather than returning null or throwing an exception,
- * </ul>
- */
+/** No-op implementation of {@link Meter}. */
 @ThreadSafe
 class DefaultMeter implements Meter {
 
--- a/api/all/src/main/java/io/opentelemetry/api/metrics/DoubleHistogram.java
+++ b/api/all/src/main/java/io/opentelemetry/api/metrics/DoubleHistogram.java
@@ -9,7 +9,11 @@ import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.context.Context;
 import javax.annotation.concurrent.ThreadSafe;
 
-/** A Histogram instrument that records {@code double} values. */
+/**
+ * A Histogram instrument that records {@code double} values.
+ *
+ * @since 1.10.0
+ */
 @ThreadSafe
 public interface DoubleHistogram {
 
--- a/api/all/src/main/java/io/opentelemetry/api/metrics/DoubleHistogram.java
+++ b/api/all/src/main/java/io/opentelemetry/api/metrics/DoubleHistogram.java
@@ -9,7 +9,7 @@ import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.context.Context;
 import javax.annotation.concurrent.ThreadSafe;
 
-/** A histogram instrument that records {@code long} values. */
+/** A histogram instrument that records {@code double} values. */
 @ThreadSafe
 public interface DoubleHistogram {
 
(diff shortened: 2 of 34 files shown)


api/all/src/main/java/io/opentelemetry/api/metrics/DefaultMeter.java
```diff
@@ -11,28 +11,7 @@
...
- *       name+version+schema_url to pick up configuration changes. This can be achieved either by
- *       allowing to work with an outdated configuration or by ensuring that new configuration
- *       applies also to previously returned Meters.
- *   <li>A MeterProvider could also return a no-op Meter here if application owners configure the
- *       SDK to suppress telemetry produced by this library
- *   <li>In case an invalid name (null or empty string) is specified, a working Meter implementation
- *       MUST be returned as a fallback rather than returning null or throwing an exception,
- * </ul>
- */
+/** No-op implementation of {@link Meter}. */
```
breedx-splk, line 14: Minor idea: Maybe retain a small mention of "minimal overhead"?

api/all/src/main/java/io/opentelemetry/api/metrics/DoubleHistogram.java
```diff
@@ -9,7 +9,7 @@
 import io.opentelemetry.context.Context;
 import javax.annotation.concurrent.ThreadSafe;
 
-/** A histogram instrument that records {@code long} values. */
+/** A histogram instrument that records {@code double} values. */
```
breedx-splk, line 12: 👍🏻

api/all/src/main/java/io/opentelemetry/api/metrics/DoubleUpDownCounter.java
```diff
@@ -9,7 +9,7 @@
 import io.opentelemetry.context.Context;
 import javax.annotation.concurrent.ThreadSafe;
 
-/** An up-down-counter instrument that records {@code double} values. */
+/** An up down counter instrument that records {@code double} values. */
```
breedx-splk, line 12:
```suggestion
/** An UpDownCounter instrument that records {@code double} values. */
```
(that's how the spec has it I think)

api/all/src/main/java/io/opentelemetry/api/metrics/LongUpDownCounter.java
```diff
@@ -9,7 +9,7 @@
 import io.opentelemetry.context.Context;
 import javax.annotation.concurrent.ThreadSafe;
 
-/** An up-down-counter instrument that records {@code long} values. */
+/** An up down counter instrument that records {@code long} values. */
```
breedx-splk, line 12:
```suggestion
/** An UpDownCounter instrument that records {@code long} values. */
```

api/all/src/main/java/io/opentelemetry/api/metrics/Meter.java
```diff
@@ -8,71 +8,110 @@
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
- * Provides instruments used to produce metrics.
+ * Provides instruments used to record measurements which are aggregated to metrics.
```
breedx-splk, line 11: 👍🏻

api/all/src/main/java/io/opentelemetry/api/metrics/Meter.java
```diff
@@ -8,71 +8,110 @@
...
  */
 @ThreadSafe
 public interface Meter {
   /**
    * Constructs a counter instrument.
    *
-   * <p>This is used to build both synchronous (in-context) instruments and asynchronous (callback)
-   * instruments.
+   * <p>This is used to build both synchronous instruments and asynchronous instruments (i.e.
+   * callbacks).
```
breedx-splk, line 64:
```suggestion
   * <p>This is used to build both synchronous (in-context) instruments and asynchronous instruments (i.e.
   * callbacks).
```
Minor: I like the original inclusion of "in-context" because I think it's an important aspect of the difference between the two. The spec notes the distinction about context too.

api/all/src/main/java/io/opentelemetry/api/metrics/Meter.java
```diff
@@ -8,71 +8,110 @@
...
+   *     may be changed.
+   * @see <a
+   *     href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-naming-rule">Instrument
+   *     Naming Rule</a>
    */
   LongCounterBuilder counterBuilder(String name);
 
   /**
-   * Constructs an up-down-counter instrument.
+   * Constructs an up down counter instrument.
```
breedx-splk, line 77:
```suggestion
   * Constructs an UpDownCounter instrument.
```

api/all/src/main/java/io/opentelemetry/api/metrics/Meter.java
```diff
@@ -8,71 +8,110 @@
...
    *
-   * <p>This is used to build both synchronous (in-context) instruments and asynchronous (callback)
-   * instruments.
+   * <p>This is used to build both synchronous instruments and asynchronous instruments (i.e.
```
breedx-splk, line 79: same comment as above about liking in-context.

api/all/src/main/java/io/opentelemetry/api/metrics/MeterBuilder.java
```diff
@@ -5,19 +5,12 @@
 
 package io.opentelemetry.api.metrics;
 
-/**
- * Builder class for creating {@link Meter} instances.
- *
- * @since 1.4.0
```
breedx-splk, line 11: Maybe this is a bigger discussion, but what's our current stance on `@since <version>` strings? I know there was an OLD issue to normalize it across the repo, but I've long lost track of it.

breedx-splk (approved):
Thanks for this. The corrections and consistencies and clarifications here are huge. I had a few minor suggestions, but overall this is solid.

api/all/src/main/java/io/opentelemetry/api/metrics/Meter.java
jack-berg: Measurements _can_ be recorded in context for synchronous instruments but don't need to be. I omitted this because this method is simply the entry point for a builder - it needs to make clear that the builder builds both synchronous and async varieties, but that's it. I think the "i.e. callbacks" clause for asynchronous instruments is different because the method name refers to callback `buildWithCallback`, so we need to make clear that callback = asynchronous.

api/all/src/main/java/io/opentelemetry/api/metrics/MeterBuilder.java
jack-berg: I made the same observation but forgot to submit the comment:

I removed these @since annotations because they don't reflect when the metrics API was published. If we're interested in keeping this, we should: 1. Add it at the class level for each public class in the metrics package. 2. Be more consistent about ensuring this is added to new APIs.

api/all/src/main/java/io/opentelemetry/api/metrics/Meter.java
breedx-splk: Oh yeah good point about "in-context" not being being a given, just a possibility. I hadn't thought about it suggesting such...but you're totally right. Looks good with your change.

api/all/src/main/java/io/opentelemetry/api/metrics/MeterBuilder.java
jkwatson: Our original plan was to assume everything was `@since 1.0.0`, and then only add explicit `@since` tags when we added new APIs. I think we have not done this consistently or well so far, though.

anuraaga (approved):

PR #4494, opened 2022-05-25
Merge otlp grpc trace and metric exporter into :exporters:otlp:all

jack-berg:
Related to #4482.

If approved I'd like to do the same otlp http by merging `:exporters:otlp-http:trace` and `:exporters:otlp-http:metrics` into a new module called `:exporters:otlp-http:all`.


anuraaga:
This seems reasonable to me, it's a tiny amount of code in each artifact. What do you think of merging the HTTP exporters into this same artifact too?

jack-berg:
Was planning on doing a separate artifact for the http exporters, but come to think of it, both grpc and the http versions depend on okhttp, and share the same serialization code, so why not?

anuraaga (approved):

PR #4498, opened 2022-05-27
Use the HTTP_TARGET attribute for sampling decisions in the Jaeger Remote Sampler

chrismgrayftsinc:
The starting span name for spans that derive from HTTP requests is
always `HTTP <method>`, which is not very useful for sampling
decisions.  Using the http target if the name is not matched allows
for things like health checks to be suppressed.

### DIFF
--- a/sdk-extensions/jaeger-remote-sampler/build.gradle.kts
+++ b/sdk-extensions/jaeger-remote-sampler/build.gradle.kts
@@ -12,6 +12,7 @@ otelJava.moduleName.set("io.opentelemetry.sdk.extension.trace.jaeger")
 
 dependencies {
   api(project(":sdk:all"))
+  implementation(project(":semconv"))
   compileOnly(project(":sdk-extensions:autoconfigure"))
 
   implementation(project(":sdk:all"))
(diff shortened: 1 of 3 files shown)


jkwatson:
Interesting. I don't know much about the jaeger remote sampler, personally. Is this something that we should have go through the specification before making the change?

chrismgrayftsinc:
I believe that this would conform to the spec.  There is conversation on open-telemetry/opentelemetry-java-contrib#65 and open-telemetry/opentelemetry-specification#1597 that lead me to this belief, but I may have misinterpreted.

jkwatson:
@anuraaga do you have much exposure to the jaeger remote sampler or know someone who does? 
@jpkrohling or @yurishkuro can you provide feedback on this?

anuraaga:
I don't know much about jaeger, but it would indeed be good to add language to the spec if changing behavior.

That being said, most instrumentation would set a span name with a route URL, so perhaps @chrismgrayftsinc your problem is a bug in instrumentation. What framework are you using?

yurishkuro:
Agree with @anuraaga, if a more descriptive name is available, why is the span name set to `HTTP <method>`, which is only meant for the worst case?

The change itself is not bad, but a) it introduces unintuitive / unexpected behavior into the SDK, b) it would be different from other implementations, and c) it's fairly hacky since it singles out a specific attribute.

A cleaner solution would be to support attribute-based sampling in the remote sampling configuration, but that would be a much larger change (including extending Jaeger's sampling config data model, something we always wanted to do).

chrismgrayftsinc:
I am using the javaagent with Spring Boot.  The name does get changed somewhere down the line, but at the time that the sampling decision is made it seems to be always `HTTP <method>`.

anuraaga:
Thanks @chrismgrayftsinc - I filed https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/6188 to see if there's a way to improve this.

PR #4499, opened 2022-05-27
Merge otlp exporters

jack-berg:
- Merge otlp http trace and metric exporters into :exporters:otlp:all
- Merge otlp http log exporter into :exporters:otlp:logs


sdk-extensions/autoconfigure/src/testOtlp/java/io/opentelemetry/sdk/autoconfigure/OtlpGrpcConfigTest.java
```diff
@@ -324,4 +327,29 @@ void configuresGlobal() {
               assertThat(server.metricRequests).isNotEmpty();
             });
   }
+
+  static void shutdownGlobalSdk() {
```
jack-berg, line 331: I merged the `testOtlpGrpc` and `testOtlpHttp` test sets, but ran into some problems because both test suites had a test that sets up `GlobalOpenTelemetry` with a periodic metric reader that continues to export after the test is over and the test grpc / http server is shutdown.

The solution I came up with is to use reflection to obtain `OpenTelemetrySdk` from `GlobalOpenTelemetry` and call shutdown. 

We might consider adding a utility to this to the `sdk-testing` in case users need to set up `GlobalOpenTelemetry` for some reason, and need to cleanup after.

exporters/otlp/all/build.gradle.kts
```diff
@@ -6,7 +6,7 @@ plugins {
   id("otel.animalsniffer-conventions")
 }
 
-description = "OpenTelemetry Protocol gRPC Exporters"
+description = "OpenTelemetry Protocol OTLP Exporters"
```
jkwatson, line 9:
```suggestion
description = "OpenTelemetry Protocol (OTLP) Exporters"
```

jkwatson (approved):
seems fine to me

anuraaga (approved):

PR #4500, opened 2022-05-27
Add default getters to ConfigProperties

breedx-splk:
Related to https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4925

This adds several default-capable get* methods to `ConfigProperties`. This can help to reduce the number of null checks. These defaults are then wired up in several places.

### DIFF
--- a/sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/ConfigProperties.java
+++ b/sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/ConfigProperties.java
@@ -23,6 +23,8 @@ public interface ConfigProperties {
   String getString(String name);
 
   /**
+   * Returns a string-valued configuration property.
+   *
    * @return a string-valued configuration property or {@code defaultValue} if a property with
    *     {@code name} has not been configured.
    * @throws ConfigurationException if the property is not a valid string.
@@ -42,6 +44,8 @@ public interface ConfigProperties {
   Boolean getBoolean(String name);
 
   /**
+   * Returns a boolean-valued configuration property.
+   *
    * @return a Boolean-valued configuration property or {@code defaultValue} if a property with
    *     {@code name} has not been configured.
    * @throws ConfigurationException if the property is not a valid string.
@@ -60,6 +64,8 @@ public interface ConfigProperties {
   Integer getInt(String name);
 
   /**
+   * Returns an Integer-valued configuration property.
+   *
    * @return an Integer-valued configuration property or {@code defaultValue} if a property with
    *     {@code name} has not been configured.
    * @throws ConfigurationException if the property is not a valid string.
@@ -78,6 +84,8 @@ public interface ConfigProperties {
   Long getLong(String name);
 
   /**
+   * Returns a Long-valued configuration property.
+   *
    * @return a Long-valued configuration property or {@code defaultValue} if a property with {@code
    *     name} has not been configured.
    * @throws ConfigurationException if the property is not a valid string.
@@ -96,6 +104,8 @@ public interface ConfigProperties {
   Double getDouble(String name);
 
   /**
+   * Returns a double-valued configuration property.
+   *
    * @return a Double-valued configuration property or {@code defaultValue} if a property with
    *     {@code name} has not been configured.
    * @throws ConfigurationException if the property is not a valid string.
@@ -129,6 +139,8 @@ public interface ConfigProperties {
   Duration getDuration(String name);
 
   /**
+   * Returns a Duration value configuration property.
+   *
    * @see ConfigProperties#getDuration(String name)
    * @return a Double-valued configuration property or {@code defaultValue} if a property with name
    *     {@code name} has not been configured.
@@ -148,6 +160,8 @@ public interface ConfigProperties {
   List<String> getList(String name);
 
   /**
+   * Returns a List value configuration property.
+   *
    * @see ConfigProperties#getList(String name)
    * @return a List configuration property or {@code defaultValue} if a property with {@code name}
    *     has not been configured.
@@ -169,6 +183,8 @@ public interface ConfigProperties {
   Map<String, String> getMap(String name);
 
   /**
+   * Returns a Map value configuration property.
+   *
    * @see ConfigProperties#getMap(String name)
    * @return a Double-valued configuration property or {@code defaultValue} if a property with
    *     {@code name} has not been configured.
--- a/sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/ConfigProperties.java
+++ b/sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/ConfigProperties.java
@@ -169,7 +169,7 @@ public interface ConfigProperties {
    */
   default List<String> getList(String name, List<String> defaultValue) {
     List<String> value = getList(name);
-    return (value == null || value.isEmpty()) ? defaultValue : value;
+    return value.isEmpty() ? defaultValue : value;
   }
 
   /**
@@ -192,7 +192,7 @@ public interface ConfigProperties {
    */
   default Map<String, String> getMap(String name, Map<String, String> defaultValue) {
     Map<String, String> value = getMap(name);
-    return (value == null || value.isEmpty()) ? defaultValue : value;
+    return value.isEmpty() ? defaultValue : value;
   }
 
   /**
--- a/sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/ConfigProperties.java
+++ b/sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/ConfigProperties.java
@@ -30,7 +30,7 @@ public interface ConfigProperties {
    * @throws ConfigurationException if the property is not a valid string.
    */
   default String getString(String name, String defaultValue) {
-    return _defaultIfNull(getString(name), defaultValue);
+    return defaultIfNull(getString(name), defaultValue);
   }
 
   /**
@@ -51,7 +51,7 @@ public interface ConfigProperties {
    * @throws ConfigurationException if the property is not a valid string.
    */
   default Boolean getBoolean(String name, Boolean defaultValue) {
-    return _defaultIfNull(getBoolean(name), defaultValue);
+    return defaultIfNull(getBoolean(name), defaultValue);
   }
 
   /**
@@ -71,7 +71,7 @@ public interface ConfigProperties {
    * @throws ConfigurationException if the property is not a valid string.
    */
   default Integer getInt(String name, Integer defaultValue) {
-    return _defaultIfNull(getInt(name), defaultValue);
+    return defaultIfNull(getInt(name), defaultValue);
   }
 
   /**
@@ -91,7 +91,7 @@ public interface ConfigProperties {
    * @throws ConfigurationException if the property is not a valid string.
    */
   default Long getLong(String name, Long defaultValue) {
-    return _defaultIfNull(getLong(name), defaultValue);
+    return defaultIfNull(getLong(name), defaultValue);
   }
 
   /**
@@ -111,7 +111,7 @@ public interface ConfigProperties {
    * @throws ConfigurationException if the property is not a valid string.
    */
   default Double getDouble(String name, Double defaultValue) {
-    return _defaultIfNull(getDouble(name), defaultValue);
+    return defaultIfNull(getDouble(name), defaultValue);
   }
 
   /**
@@ -147,7 +147,7 @@ public interface ConfigProperties {
    * @throws ConfigurationException if the property is not a valid string.
    */
   default Duration getDuration(String name, Duration defaultValue) {
-    return _defaultIfNull(getDuration(name), defaultValue);
+    return defaultIfNull(getDuration(name), defaultValue);
   }
 
   /**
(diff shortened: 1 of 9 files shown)


sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/ConfigProperties.java
```diff
@@ -33,7 +44,18 @@ public interface ConfigProperties {
...
 
   /**
-   * Returns a integer-valued configuration property.
+   * Returns a boolean-valued configuration property.
+   *
+   * @return a Boolean-valued configuration property or {@code defaultValue} if a property with
+   *     {@code name} has not been configured.
+   * @throws ConfigurationException if the property is not a valid string.
+   */
+  default Boolean getBoolean(String name, Boolean defaultValue) {
```
trask, line 53:
```suggestion
  @Nullable
  default Boolean getBoolean(String name, @Nullable Boolean defaultValue) {
```

or 

```suggestion
  default boolean getBoolean(String name, boolean defaultValue) {
```

breedx-splk: Hmmm....this is super interesting. Do you think we should do that for all the primitive/boxable types?
I worry a little in the inconsistency/mismatch between the primitive one, but I think this is an improvement. I'm gonna go for it for all of them. ❤️

sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/ConfigProperties.java
```diff
@@ -93,12 +160,46 @@ public interface ConfigProperties {
...
+   * Returns a List value configuration property.
+   *
+   * @see ConfigProperties#getList(String name)
+   * @return a List configuration property or {@code defaultValue} if a property with {@code name}
+   *     has not been configured.
+   * @throws ConfigurationException if the property is not a valid string.
+   */
+  default List<String> getList(String name, List<String> defaultValue) {
+    List<String> value = getList(name);
+    return (value == null || value.isEmpty()) ? defaultValue : value;
```
anuraaga, line 172: `getList` and `getMap` can't return `null`

sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/ConfigProperties.java
```diff
@@ -93,12 +160,46 @@ public interface ConfigProperties {
...
+  default Map<String, String> getMap(String name, Map<String, String> defaultValue) {
+    Map<String, String> value = getMap(name);
+    return (value == null || value.isEmpty()) ? defaultValue : value;
+  }
+
+  /**
+   * Returns defaultValue if value is null, otherwise value. This is an internal method that should
+   * not be broadly used.
+   */
+  default <T> T defaultIfNull(@Nullable T value, T defaultValue) {
```
anuraaga, line 202: Let's move this to a package-private class and then wouldn't need the scary ignorable javadoc. Or consider just inlining - it's ok, but slightly harder to reason about the behavior of this method with the primitive types / autoboxing.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtil.java
```diff
@@ -32,11 +32,9 @@ final class OtlpConfigUtil {
   static final String PROTOCOL_HTTP_PROTOBUF = "http/protobuf";
 
   static String getOtlpProtocol(String dataType, ConfigProperties config) {
-    String protocol = config.getString("otel.exporter.otlp." + dataType + ".protocol");
-    if (protocol == null) {
-      protocol = config.getString("otel.exporter.otlp.protocol");
-    }
-    return (protocol == null) ? PROTOCOL_GRPC : protocol;
+    String configKey = "otel.exporter.otlp." + dataType + ".protocol";
+    String fallback = config.getString("otel.exporter.otlp.protocol", PROTOCOL_GRPC);
```
anuraaga, line 36: This seems harder to read then previously as usually you don't pre-initialize a fallback and the variable names are not semantic anymore. I'd probably use the nullable version for the first lookup, for all the similar fallback type of lookups here.

```java
String protocol = config.getString("otel.exporter.otlp." + dataType + ".protocol);
if (protocol != null) {
  return protocol;
}
return config.getString("otel.exporter.otlp.protocol", PROTOCOL_GRPC);
```

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/PropagatorConfiguration.java
```diff
@@ -20,16 +20,15 @@
 
 final class PropagatorConfiguration {
 
+  public static final List<String> DEFAULT_PROPAGATORS = Arrays.asList("tracecontext", "baggage");
```
anuraaga, line 23:
```suggestion
  private static final List<String> DEFAULT_PROPAGATORS = Arrays.asList("tracecontext", "baggage");
```

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java
```diff
@@ -27,6 +27,9 @@
 
 final class TracerProviderConfiguration {
 
+  private static final double DEFAULT_SAMPLE_RATIO = 1.0d;
```
anuraaga, line 30:
```suggestion
  private static final double DEFAULT_TRACEIDRATIO_SAMPLE_RATIO = 1.0d;
```

Stuttering isn't great, but without the sampler type here the variable seems mysterious

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java
```diff
@@ -157,22 +157,16 @@ static Sampler configureSampler(
...
         {
-          Double ratio = config.getDouble("otel.traces.sampler.arg");
-          if (ratio == null) {
-            ratio = 1.0d;
-          }
+          double ratio = config.getDouble("otel.traces.sampler.arg", DEFAULT_SAMPLE_RATIO);
           return Sampler.traceIdRatioBased(ratio);
         }
-      case "parentbased_always_on":
+      case DEFAULT_SAMPLER:
```
anuraaga, line 163: Whether `parentbased_always_on` is the default or not isn't important here, let's stick with the string since we're just parsing strings in this switch statement

sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/ConfigProperties.java
breedx-splk: Good point, I'll remove the null checks.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtil.java
breedx-splk: Yeah, the starting point had both the conditional and the ternary. This seems like a good compromise. 👍🏻

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java
breedx-splk: Ah ok, the goal was to eliminate having the string constant in two places. I'll rename it to reflect the contents instead.

sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OtlpConfigUtil.java
```diff
@@ -73,10 +73,9 @@ static void configureOtlpExporterBuilder(
       setEndpoint.accept(endpoint.toString());
     }
 
-    Map<String, String> headers = config.getMap("otel.exporter.otlp." + dataType + ".headers");
-    if (headers.isEmpty()) {
-      headers = config.getMap("otel.exporter.otlp.headers");
-    }
+    Map<String, String> fallbackHeaders = config.getMap("otel.exporter.otlp.headers");
```
anuraaga, line 76: Let's revert these two lines for similar reasons to above. When reading two properties in priority order, I think it is a lot easier to read when the code also goes in that order.

sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/NullDefaultUtility.java
```diff
@@ -0,0 +1,22 @@
...
+
+package io.opentelemetry.sdk.autoconfigure.spi;
+
+import javax.annotation.Nullable;
+
+/**
+ * Holder for the non-public defaultIfNull method. This serves only to mitigate the method being on
+ * a public interface.
+ */
+final class NullDefaultUtility {
```
anuraaga, line 14: Can call this `ConfigUtil` to reduce friction when adding more methods for the same reason

anuraaga (approved):

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/ConfigPropertiesTest.java
```diff
@@ -216,4 +220,31 @@ void durationUnitParsing() {
                 .getDuration("duration"))
         .isEqualTo(Duration.ofMillis(8));
   }
+
+  @Test
+  void defaultMethods() {
+    ConfigProperties properties = DefaultConfigProperties.get(emptyMap());
+    assertEquals(true, properties.getBoolean("foo", true));
```
anuraaga, line 227: Use assertj for all the assertions.

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/ConfigPropertiesTest.java
```diff
@@ -216,4 +220,31 @@ void durationUnitParsing() {
...
+    assertEquals(65.535, properties.getDouble("foo", 65.535));
+    assertEquals(21, properties.getInt("foo", 21));
+    assertEquals(123L, properties.getLong("foo", 123L));
+    assertEquals(Duration.ofDays(13), properties.getDuration("foo", Duration.ofDays(13)));
+  }
+
+  @Test
+  void defaultCollectionTypes() {
+    ConfigProperties properties = DefaultConfigProperties.get(emptyMap());
+    assertEquals(
```
anuraaga, line 238: Don't forget to use list and map assertions when switching to assertj

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/ConfigPropertiesTest.java
jkwatson: Yes, we want to use assertJ for everything. If you know of tests that aren't, please create an issue pointing to them.

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/ConfigPropertiesTest.java
```diff
@@ -216,4 +218,31 @@ void durationUnitParsing() {
                 .getDuration("duration"))
         .isEqualTo(Duration.ofMillis(8));
   }
+
+  @Test
+  void defaultMethods() {
+    ConfigProperties properties = DefaultConfigProperties.get(emptyMap());
+    assertThat(properties.getBoolean("foo", true)).isTrue();
```
anuraaga, line 225: Noticed it looks like there's missing test coverage that the default methods will return the present value, these are just checking the fallback

sdk-extensions/autoconfigure/src/test/java/io/opentelemetry/sdk/autoconfigure/ConfigPropertiesTest.java
```diff
@@ -216,4 +218,31 @@ void durationUnitParsing() {
...
+    assertThat(properties.getList("foo", Arrays.asList("1", "2", "3")))
+        .containsExactly("1", "2", "3");
+    assertThat(properties.getList("foo")).isEmpty();
+    Map<String, String> defaultMap = new HashMap<>();
+    defaultMap.put("one", "1");
+    defaultMap.put("two", "2");
+    Map<String, String> expected = new HashMap<>();
+    expected.put("one", "1");
+    expected.put("two", "2");
+    assertThat(properties.getMap("foo", defaultMap)).containsExactlyEntriesOf(expected);
```
anuraaga, line 245:
```suggestion
    assertThat(properties.getMap("foo", defaultMap)).containsExactly(
      entry("one", "1"), entry("two", "2"));
```

Alternatively `.containsExactlyEntriesOf(defaultMap)` seems pretty readable too

jack-berg (approved):

jkwatson (approved):

PR #4501, opened 2022-05-28
Add metrics to ZipkinSpanExporter

Donnerbart:
This is a follow-up to https://github.com/open-telemetry/opentelemetry-java/pull/4487 to complete the metrics for span exporters story.

I'm not sure if you like the distinct transport name for HTTP/JSON encoding in the `instrumentationScopeName`. This will change the existing metric names for `OkHttpExporter` when JSON encoding is used. I made that part an extra commit which can easily be dropped, if that change is not wanted.


Donnerbart:
I think the field `baseLogger` and the method `produceLocalIp()` in `ZipkinSpanExporter` should be `private`. This change technically breaks binary compatibility, but I think those should never have been exposed. Is there a way to do this change without failing the `exporters:zipkin:jApiCmp` check or can't this be changed anymore?

jkwatson:
> I think the public field `baseLogger` and the package-private method `produceLocalIp()` in `ZipkinSpanExporter` should be `private`. This change would break the binary compatibility, but I think it's very unlikely that someone is using these anyway. Is there a way to do this change without failing the `exporters:zipkin:jApiCmp` check or can't this be changed anymore?

Unfortunately, we're stuck with that mistake. :(

jack-berg:
Sorry for not responding sooner. I've been hesitant to accept this because it introduces a new transitive dependency for the `:exporters:zipkin` module on `com.squareup.okhttp3:okhttp`. More generally, the `:exporters:otlp:common` artifact is becoming a home for utility methods used by a variety of exporters, which feels sloppy.

Maybe we can move the truly common classes to a new module `:exporters:common`, and merge remaining shared otlp utilities into `:exporters:otlp:all`. Thoughts @anuraaga / @jkwatson?

jack-berg:
It shades in okhttp [io.zipkin.reporter2:zipkin-sender-okhttp3](https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/zipkin/build.gradle.kts#L18).

jack-berg:
This is unblocked now that #4575 is merged. I've pushed a commit to rebase. Thanks @Donnerbart!

jack-berg (approved):

exporters/common/src/main/java/io/opentelemetry/exporter/internal/ExporterMetrics.java
```diff
@@ -58,21 +58,39 @@ public void addFailed(long value) {
     exported.add(value, failedAttrs);
   }
 
-  /** Create an instance for recording gRPC exporter metrics. */
+  /**
+   * Create an instance for recording exporter metrics under the meter "{@code
+   * io.opentelemetry.exporters.}" + {@code exporterName} + "-grpc".
```
jkwatson, line 63: does this javadoc work???

jack-berg: It certainly leaves room for improvement! Fixed.

jack-berg:
@jkwatson any additional thoughts before I merge this?

jkwatson (approved):
🚢

PR #4514, opened 2022-06-03
Allow overriding OTLP gRPC authority.

### LINKED ISSUE #4513: Allow overriding authority and GrpcChannel used for GRPC exporter stub

We're using Envoy proxy for all GRPC requests, which relies on `authority` field to distinguish requests to different servers going through the same proxy (same `endpoint`).
We're using OTLP GRPC exporter (`OtlpGrpcSpanExporter`), which doesn't allow overriding `authority` or any other `CallOptions`, since it's being created with default empty `CallOptions` and it's not being exposed anywhere ([OtlpGrpcSpanExporterBuilder.java:42](https://github.com/open-telemetry/opentelemetry-java/blob/77f5a174d4d6f735ee7fe4717cab720844f77395/exporters/otlp/all/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterBuilder.java#L42)).

Also, we need to pass `GrpcChannel` because of the proxy to re-use the same proxy connection. This can be achieved by using `OtlpGrpcSpanExporterBuilder.setChannel` ([OtlpGrpcSpanExporterBuilder.java:57](https://github.com/open-telemetry/opentelemetry-java/blob/77f5a174d4d6f735ee7fe4717cab720844f77395/exporters/otlp/all/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterBuilder.java#L57)), but the method is deprecated hence it's probably not wise to use it.

1. Undeprecate `OtlpGrpcSpanExporterBuilder.setChannel`
2. allow overriding authority on `OtlpGrpcSpanExporterBuilder` via one of these:
a. allowing overriding `authority` directly, or
b. allowing overriding `CallOptions`, which would allows overriding `authority` among other things

Tried extending builder and/or exporter, but that doesn't seem feasible, because most of relevant classes aren't public and/or are final, or methods of public classes are private. Alternative solution might be to relax visibility by making related classes and methods extensible, but it would expose more internal things so the main solution seems better.

anuraaga:
Fixes #4513

OkHttp already allows setting the host header to override host on HTTP/1 or :authority on HTTP/2. For the gRPC codepath we need to propagate it through though.

### DIFF
--- a/docs/apidiffs/current_vs_latest/opentelemetry-sdk-extension-autoconfigure-spi.txt
+++ b/docs/apidiffs/current_vs_latest/opentelemetry-sdk-extension-autoconfigure-spi.txt
@@ -1,4 +1,14 @@
 Comparing source compatibility of  against 
+***! MODIFIED INTERFACE: PUBLIC ABSTRACT io.opentelemetry.sdk.autoconfigure.spi.ConfigProperties  (not serializable)
+	===  CLASS FILE FORMAT VERSION: 52.0 <- 52.0
+	+++! NEW METHOD: PUBLIC(+) boolean getBoolean(java.lang.String, boolean)
+	+++! NEW METHOD: PUBLIC(+) double getDouble(java.lang.String, double)
+	+++! NEW METHOD: PUBLIC(+) java.time.Duration getDuration(java.lang.String, java.time.Duration)
+	+++! NEW METHOD: PUBLIC(+) int getInt(java.lang.String, int)
+	+++! NEW METHOD: PUBLIC(+) java.util.List getList(java.lang.String, java.util.List)
+	+++! NEW METHOD: PUBLIC(+) long getLong(java.lang.String, long)
+	+++! NEW METHOD: PUBLIC(+) java.util.Map getMap(java.lang.String, java.util.Map)
+	+++! NEW METHOD: PUBLIC(+) java.lang.String getString(java.lang.String, java.lang.String)
 +++  NEW INTERFACE: PUBLIC(+) ABSTRACT(+) io.opentelemetry.sdk.autoconfigure.spi.metrics.ConfigurableMetricExporterProvider  (not serializable)
 	+++  CLASS FILE FORMAT VERSION: 52.0 <- n.a.
 	+++  NEW SUPERCLASS: java.lang.Object
--- a/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/grpc/DefaultGrpcExporterBuilder.java
+++ b/exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/grpc/DefaultGrpcExporterBuilder.java
@@ -20,7 +20,7 @@ import io.opentelemetry.exporter.internal.retry.RetryPolicy;
 import java.net.URI;
 import java.time.Duration;
 import java.util.concurrent.TimeUnit;
-import java.util.function.Function;
+import java.util.function.BiFunction;
 import javax.annotation.Nullable;
 import javax.net.ssl.SSLException;
 
@@ -35,7 +35,7 @@ public final class DefaultGrpcExporterBuilder<T extends Marshaler>
 
   private final String exporterName;
   private final String type;
-  private final Function<ManagedChannel, MarshalerServiceStub<T, ?, ?>> stubFactory;
+  private final BiFunction<ManagedChannel, String, MarshalerServiceStub<T, ?, ?>> stubFactory;
   private final String grpcServiceName;
 
   @Nullable private ManagedChannel channel;
@@ -47,6 +47,7 @@ public final class DefaultGrpcExporterBuilder<T extends Marshaler>
   @Nullable private byte[] privateKeyPem;
   @Nullable private byte[] certificatePem;
   @Nullable RetryPolicy retryPolicy;
+  @Nullable String authorityOverride;
   private MeterProvider meterProvider = MeterProvider.noop();
 
   /** Creates a new {@link DefaultGrpcExporterBuilder}. */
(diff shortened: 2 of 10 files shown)


exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/grpc/DefaultGrpcExporterBuilder.java
```diff
@@ -167,7 +172,7 @@ public GrpcExporter<T> build() {
 
     Codec codec = compressionEnabled ? new Codec.Gzip() : Codec.Identity.NONE;
     MarshalerServiceStub<T, ?, ?> stub =
-        stubFactory.apply(channel).withCompression(codec.getMessageEncoding());
+        stubFactory.apply(channel, authorityOverride).withCompression(codec.getMessageEncoding());
```
anuraaga, line 175: BTW, while many call options are available on the stub itself, e.g. `stub.withDeadline`, authority isn't probably because it's an experimental API.

Oberon00:
This scenario seems very relevant for the HTTP/net.peer.name/proxy semantic convention discussions here: https://github.com/open-telemetry/opentelemetry-specification/pull/2469#issuecomment-1143728234 (and open-telemetry/opentelemetry-specification#260)
I would be glad if you could chime in in the discussion there to check if the suggested new HTTP semantic conventions (or the current ones for that matter) would allow for a sensible representation of that scenario.

donce:
It's a bit hard to follow the discussion about HTTP semantic conventions, but for our use-case, we're building exporter manually, via `OtlpGrpcSpanExporter.builder()` and we're currently lacking 2 things:
- Allow re-using existing `GrpcChannel` for gRPC exporter (which is possible, but method is marked as deprecated)
- Allow setting authority (which is usually set via `CallOptions`)

Oberon00:
The question for the semantic conventions would rather be: if you were to create a span for the export operation (which we obviously don't want in practice, since it would create a feedback loop), how would you represent this call with the custom authority/host? How would you, in theory, set net.peer.name, http.url, http.host.

docs/apidiffs/current_vs_latest/opentelemetry-sdk-extension-autoconfigure-spi.txt
```diff
@@ -1,4 +1,14 @@
...
+***! MODIFIED INTERFACE: PUBLIC ABSTRACT io.opentelemetry.sdk.autoconfigure.spi.ConfigProperties  (not serializable)
+	===  CLASS FILE FORMAT VERSION: 52.0 <- 52.0
+	+++! NEW METHOD: PUBLIC(+) boolean getBoolean(java.lang.String, boolean)
+	+++! NEW METHOD: PUBLIC(+) double getDouble(java.lang.String, double)
+	+++! NEW METHOD: PUBLIC(+) java.time.Duration getDuration(java.lang.String, java.time.Duration)
+	+++! NEW METHOD: PUBLIC(+) int getInt(java.lang.String, int)
+	+++! NEW METHOD: PUBLIC(+) java.util.List getList(java.lang.String, java.util.List)
+	+++! NEW METHOD: PUBLIC(+) long getLong(java.lang.String, long)
+	+++! NEW METHOD: PUBLIC(+) java.util.Map getMap(java.lang.String, java.util.Map)
+	+++! NEW METHOD: PUBLIC(+) java.lang.String getString(java.lang.String, java.lang.String)
```
jkwatson, line 11: oh man. I had forgotten this was in a stable artifact.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/grpc/DefaultGrpcExporterBuilder.java
```diff
@@ -110,6 +111,10 @@ public GrpcExporterBuilder<T> setClientTls(byte[] privateKeyPem, byte[] certific
 
   @Override
   public DefaultGrpcExporterBuilder<T> addHeader(String key, String value) {
+    if (key.equals("host")) {
+      authorityOverride = value;
+      return this;
```
jkwatson, line 116: I don't know enough about this, but will people be surprised if it doesn't end up as an actual header?

docs/apidiffs/current_vs_latest/opentelemetry-sdk-extension-autoconfigure-spi.txt
jack-berg: Opened an #4516 to make committing these API diffs a requirement for merging PRs.

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/grpc/DefaultGrpcExporterBuilder.java
jack-berg: [Host](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host) is a header in HTTP - not sure if it plays any role in gRPC.

anuraaga: It does actually end up in a header, `: authority`, just grpc doesn't support setting it through the metadata API but has this separate API.

anuraaga: Ah but it wouldn't be named `host`. AFAIK it is not expected to set host on a http/2 request. For example, if proxies get both with different values, it will ignore the host value when converting to HTTP/1, moving the value from authority into host, from what I've seen.

In practice I couldn't find any way to set both with okhttp anyways, it will always accept host and either set it as host on http/1 or authority on http/2 with no way of setting host on the latter, probably for the above reason. So we'll probably want to just keep that behavior consistent across all the exporters.

jack-berg: Is there any problem with using `if (key.equals(":authority")) { ... }` instead?

anuraaga: OkHttp doesn't accept `: authority` and silently ignores it

jack-berg (approved):

PR #4529, opened 2022-06-13
Verify the TLS we use in tests are in PEM format and document PCKS8 r…

### LINKED ISSUE #4528: mtls Client key file does not follow the specification to read pem file, it reads der instead
**Describe the bug**
while the specification said the OTEL_EXPORTER_OTLP_CLIENT_KEY will read pem format, it actually read DER only which is the binary form of pem. If you also feed with pem file, you will get wrong key format error.
Also the code only read PCKS8 encoded key which should be mentioned in the readme at least, PCKS1 is still widely being used.

**Steps to reproduce**
OTEL_EXPORTER_OTLP_CLIENT_KEY=aPEMKeyfile.pem

should read pem file without error, supporting DER is something good to have.

**What did you see instead?**
it cannot read pem file.

v0.13.1, while the code did not change since then

**Environment**
opentelemetry operator

**Additional context**
the spec and code are from the the same issue, so we can also change the spec either, but some other languages support both.

anuraaga:
…estriction on key

Fixes #4528


exporters/otlp/all/src/main/java/io/opentelemetry/exporter/otlp/trace/OtlpGrpcSpanExporterBuilder.java
```diff
@@ -105,7 +105,7 @@ public OtlpGrpcSpanExporterBuilder setCompression(String compressionMethod) {
 
   /**
    * Sets the certificate chain to use for verifying servers when TLS is enabled. The {@code byte[]}
-   * should contain an X.509 certificate collection in PEM format. If not set, TLS connections will
+   * should contain a X.509 certificate collection in PEM format. If not set, TLS connections will
```
jkwatson, line 108: This was correct before. English is weird, but since you pronounce it "ex", you need an "an" here.

exporters/otlp/testing-internal/src/main/java/io/opentelemetry/exporter/otlp/testing/internal/AbstractGrpcTelemetryExporterTest.java
```diff
@@ -170,6 +170,24 @@ protected AbstractGrpcTelemetryExporterTest(String type, U resourceTelemetryInst
   @BeforeAll
   void setUp() {
     exporter = exporterBuilder().setEndpoint(server.httpUri().toString()).build();
+
+    // Sanity check that TLS files are in PEM format.
+    assertThat(certificate.certificateFile())
```
pureklkl, line 175: The issue is that the test encoded the key before it sets to the client, but the sdk read the key file and set to the client without encoding.

anuraaga: Thanks a lot for clarifying this!

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/TlsUtil.java
```diff
@@ -130,4 +136,27 @@ public static X509TrustManager trustManager(byte[] trustedCertificatesPem) throw
       throw new SSLException("Could not build TrustManagerFactory from trustedCertificatesPem.", e);
     }
   }
+
+  // We catch linkage error to provide a better exception message on Android.
+  // https://github.com/open-telemetry/opentelemetry-java/issues/4533
```
anuraaga, line 141: Filed this issue and punted for now

jkwatson (approved):

jack-berg (approved):

PR #4538, opened 2022-06-16
Use OkHttpGrpcExporter unless upstream Channel provided by user

### LINKED ISSUE #4515: Use OkHttpGrpcExporter unless setChannel is called
We originally used classpath check to determine whether to use the upstream gRPC-based exporter vs OkHttp-based one to make the change conservative, in case there are issues with the latter in production. The latter has been the only option for javaagent users for a long time and there aren't any known performance or stability issues so we can probably make it the default regardless of classpath.

The gRPC codepath would still exist to support our `setChannel` method we're stuck with for semver reasons. So we can change the logic, probably with a bit of tedium, to use the OkHttp exporter unless `setChannel` is called, in which case return a gRPC exporter. The main tedium is that this _probably_ requires copying from one builder to the other, I don't think our reliance on lazy classpath linking will work if attempting to do this with a single builder in `build()`.

anuraaga:
I was able to use a inner class to handle the classloading barrier, so merged both exporter builders into one, as well as some testing utilities.

Fixes #4515


exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/grpc/UpstreamGrpcExporter.java
```diff
@@ -121,9 +117,6 @@ public void onFailure(Throwable t) {
 
   @Override
   public CompletableResultCode shutdown() {
-    if (managedChannel.isTerminated()) {
-      return CompletableResultCode.ofSuccess();
-    }
-    return ManagedChannelUtil.shutdownChannel(managedChannel);
```
anuraaga, line 127: This was a bit fishy previously - we were closing user-provided channels which likely shouldn't have been happening. Now since this class is only for user-provided channels, we can simply never close it

exporters/otlp/testing-internal/src/main/java/io/opentelemetry/exporter/otlp/testing/internal/ManagedChannelTelemetryExporterBuilder.java
```diff
@@ -0,0 +1,149 @@
...
+
+  @Override
+  public TelemetryExporterBuilder<T> addHeader(String key, String value) {
+    delegate.addHeader(key, value);
+    return this;
+  }
+
+  // When a user provides a Channel, we are not in control of TLS or retry config and reimplement it
+  // here for use in tests. Technically we don't have to test them since they are out of the SDK's
+  // control, but it's probably worth verifying the baseline functionality anywaysl
```
jack-berg, line 83:
```suggestion
  // control, but it's probably worth verifying the baseline functionality anyways
```

jack-berg (approved):

exporters/otlp/common/src/main/java/io/opentelemetry/exporter/internal/grpc/GrpcExporterBuilder.java
```diff
@@ -5,39 +5,211 @@
...
+        type,
+        clientBuilder.build(),
+        meterProvider,
+        endpoint,
+        headers.build(),
+        compressionEnabled);
+  }
+
+  // Use an inner class to ensure GrpcExporterBuilder does not have classloading dependencies on
+  // upstream gRPC.
```
JesseEstum, line 187: @anuraaga I think there is a bug here.  At the time UpstreamGrpcExporterFactory inner class is called on line 133, the current ClassLoader is the `AgentClassLoader` (io.opentelemetry.javaagent.bootstrap.AgentClassLoader), which does not have access to `io.grpc.Codec`.  `io.grpc.Codec` may be available in an ExtensionClassloader, but not here when line 133 is called.  Thus, everything blows up with a `NoClassDefFoundError`.

JesseEstum: I'm speaking in the context of the [java instrumentation agent](https://github.com/open-telemetry/opentelemetry-java-instrumentation) and an extension that tries to provide a `Channel`.

trask: hi @JesseEstum! can you open an issue for this? otherwise likely to get lost on this old PR

JesseEstum: Good point, @trask .  After reflecting over the weekend, I realized this would be a non-issue if there was a way to control the class path of the `AgentClassLoader` and perhaps I just "wasn't doing it right".

As such, I opened https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/9029 as an alternative option.  I'll let that one play out first then if necessary open an issue against this repo.

PR #4557, opened 2022-06-23
Otlp exporter default aggregation

### LINKED ISSUE #4540: Make exponential histograms more available
We have an exponential histogram implementation but because exponential histograms aren't yet stable in the metrics spec, the implementation is obscure / difficult to use. This issue is tracking steps to make histograms easier to use, even while they remain experimental. If they're easier to use, we can get feedback that may help stabilize the specification.

- [x] Align exponential histogram implementation with specification #4536 
- [ ] MetricReader can influence default aggregation by instrument kind #4472
- [ ] MetricExporter can influence default aggregation of PeriodicMetricReader
- [ ] Add autoconfigure option for using exponential histogram aggregation as default aggregation for otlp metric exporters. Implementation of spec [PR #2619](https://github.com/open-telemetry/opentelemetry-specification/pull/2619).

jack-berg:
Resolves #4540.

- Add `Otlp*MetricExporterBuilder#setDefaultAggregationSelector(DefaultAggregationSelector)`
- Add `DefaultAggregationSelector#compose(InstrumentType, Aggregation)` method for easier composition of aggregation selector rules.
- Add autoconfigure support for `OTLP_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION` (from spec PR [2619](https://github.com/open-telemetry/opentelemetry-specification/pull/2619))


sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/DefaultAggregationSelector.java
```diff
@@ -24,6 +26,33 @@ static DefaultAggregationSelector getDefault() {
...
+   * <pre>{@code
+   * // DefaultAggregationSelector selector =
+   * //   DefaultAggregationSelector.getDefault()
+   * //     .compose(InstrumentType.HISTOGRAM, Aggregation.drop());
+   * }</pre>
+   *
+   * @since 1.16.0
+   */
+  default DefaultAggregationSelector compose(
+      InstrumentType instrumentType, Aggregation aggregation) {
```
jack-berg, line 45: This isn't needed for this change, but I think it really improves the ergonomics of defining `DefaultAggregationSelector` instances. Inspired by [Function#compose](https://docs.oracle.com/javase/8/docs/api/java/util/function/Function.html#compose-java.util.function.Function-).

anuraaga: Perhaps can be named `withAggregation` or something, this feels more like a "mutation" of the selector rather than composition (usually you compose two of the same type)

jack-berg: I like that, but how about just `with`? The class is annotated with a `FunctionalInterface` so its likely to stay close to its current form and not need to disambiguate between with "aggregation" and with "something else".

anuraaga (approved):

PR #4558, opened 2022-06-25
Add javadoc about autoconfigure pitfall

trask:


sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/AutoConfigurationCustomizer.java
```diff
@@ -77,6 +77,14 @@ AutoConfigurationCustomizer addPropertiesSupplier(
    * customization. The return value of the {@link BiFunction} will replace the passed-in argument.
    *
    * <p>Multiple calls will execute the customizers in order.
+   *
+   * <p>Note: calling {@link SdkTracerProviderBuilder#setSampler(Sampler)} inside of your
+   * configuration function will cause any sampler customizers to be ignored that were configured
+   * via {@link #addSamplerCustomizer(BiFunction)}.
```
trask, line 83: I was going to add something like:

> If you want to replace the default sampler, check out ConfigurableSamplerProvider and ConfigCustomizer instead.

but realized ConfigCustomizer is javaagent spi.

Just checking, is there any interest in adding ConfigCustomizer to the SDK autoconfigure?

anuraaga: That's what `addPropertiesSupplier` in this file is for. Is it possible to deprecate `ConfigCustomizer`?

trask: oh, I thought I looked for this, but obviously not very well 🤦‍♂️

I updated the javadoc to reference this.

@mateuszrzeszutek after https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/6202, maybe we can undo my suggestion https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/6010#pullrequestreview-970295866 and move `ConfigCustomizer` back to `agent-tooling`

mateuszrzeszutek: Hmm, I think we could -- however, we should get the `ConfigProperties` story straight first. If moving it into API is an option, it would most likely solve most of our problems with it.

trask: Resolving this conversation, I think the javadoc comment is good now, and we have other issues / PRs tracking getting the `ConfigProperties` story straight.

anuraaga (approved):

jack-berg (approved):

